PR #22, opened 2021-09-23
Refactor form definition to config file

baarkerlounger:
This PR starts to consider that a form may have multiple questions per page and that we will have multiple form definitions.

It moves the definition of a form (what sections, subsections, pages and questions it has) to a JSON config file with a parsing class. It also generates the required views for the pages and questions dynamically.

Assumptions made by the JSON definition:
- Section, Subsection, Page and Question keys are unique within a form
- All sections have at least one subsection
- All subsections have at least one page
- All pages have at least one question
- All questions have a type [text, numeric, radio or select]

Next steps:

- [x] Create initial config definition of form
- [x] Load form definition from single point of entry 
- [x] Generate View for Pages and corresponding questions dynamically
- [x] Ensure answers are saved for the questions submitted for each page
- [ ] Complete the JSON for the remaining sections
- [ ] Data migrations for fields for remaining questions
- [ ] Clean up form parser class and make more performant


app/models/form.rb
```diff
@@ -1,46 +1,63 @@
...
+  end
+
+  # Returns a hash with pages as keys
+  def all_pages
+    @all_pages ||= all_subsections.map { |_subsection_key, subsection_value|
+      subsection_value["pages"]
+    }.reduce(:merge)
+  end
+
+  # Returns a hash with the pages as keys
```
dushan-madetech:
```suggestion
  # Returns a hash with the pages of a subsection as keys
```

app/models/form.rb
```diff
@@ -1,46 +1,63 @@
...
-  end
-
-  def self.previous_question(current_question)
-    Hash[QUESTIONS.to_a.map(&:reverse)][current_question]
+class Form
+  attr_reader :form_definition
+
+  def initialize(start_year, end_year)
+    form_json = "config/forms/#{start_year}_#{end_year}.json"
+    raise "No form definition file exists for given year".freeze unless File.exist?(form_json)
```
dushan-madetech, line 6: Should raising an error be part of the initialisation or would it be better to extract this into its own method? Asking because I feel like the second matches the Single Responsibility Principle better but I'm keen to hear what you think 🙂

app/views/form/_radio_question.html.erb
```diff
@@ -0,0 +1,6 @@
+<% answer_options = question["answer_options"].map { |k, v| OpenStruct.new(id: k.to_i, value: v) } %>
+
+<%= f.govuk_collection_radio_buttons question_key, answer_options, :id, :value, legend: { text: question["header"], size: "l" } %>
+<%= f.hidden_field :previous_page, value: question_key %>
+<%= f.hidden_field :case_log_id, value: case_log_id %>
+<%= f.govuk_submit "Save and continue" %>
```
dushan-madetech, line 6: I like the idea of splitting these into reusable erb files 👍

config/forms/2021_2022.json
```diff
@@ -0,0 +1,306 @@
+{
+  "form_type": "lettings",
+  "start_year": 2021,
+  "end_year": 2022,
```
dushan-madetech, line 4: I like this approach, I can see why it would make the form easy to change since it has to be changed on a yearly basis 👍

dushan-madetech (approved):
Good changes, I definitely think this is a better approach for making the form more dynamic given that it's going to have to change every year

app/models/form.rb
baarkerlounger: I think this is reasonable here because it's a small method that I still intend to change significantly and currently it's the only place it applies anyway (nothing else raises an error yet). Might be worth extracting out later though.

dushan-madetech: Sounds good to me 👍

PR #25, opened 2021-09-30
CLDC-277: create check answer view

dushan-madetech:
This PR has the following important changes:
- Creates a generalised check answer view (a template page for check answers)
- Updates the case log controller for check answers pages
- Changes questions to save values to the database instead of keys/numbers
- Adds an ADR about saving values directly to the database
- Tests for the content and functionality of the check answers page

### DIFF
--- a/app/controllers/case_logs_controller.rb
+++ b/app/controllers/case_logs_controller.rb
@@ -40,7 +40,11 @@ class CaseLogsController < ApplicationController
 
   def check_answers
     @case_log = CaseLog.find(params[:case_log_id])
-    render "form/check_answers", locals: { case_log_id: @case_log.id }
+    form = Form.new(2021, 2022)
+    current_url = request.env['PATH_INFO']
+    subsection = current_url.split('/')[-2]
+    subsection_pages = form.pages_for_subsection(subsection)
+    render "form/check_answers", locals: { case_log_id: @case_log.id, subsection_pages: subsection_pages }
   end
 
   form = Form.new(2021, 2022)
--- a/app/controllers/case_logs_controller.rb
+++ b/app/controllers/case_logs_controller.rb
@@ -28,7 +28,14 @@ class CaseLogsController < ApplicationController
     answers_for_page = page_params(questions_for_page).select { |k, _v| questions_for_page.include?(k) }
     @case_log.update!(answers_for_page)
     next_page = form.next_page(previous_page)
-    redirect_to(send("case_log_#{next_page}_path", @case_log))
+    redirect_path = if next_page == :check_answers 
+      subsection = form.subsection_for_page(previous_page)
+      "case_log_#{subsection}_check_answers_path"
+    else  
+      "case_log_#{next_page}_path"
+    end
+      
+    redirect_to(send(redirect_path, @case_log))
   end
 
   def check_answers
--- a/app/controllers/case_logs_controller.rb
+++ b/app/controllers/case_logs_controller.rb
@@ -44,7 +44,7 @@ class CaseLogsController < ApplicationController
     current_url = request.env["PATH_INFO"]
     subsection = current_url.split("/")[-2]
     subsection_pages = form.pages_for_subsection(subsection)
-    render "form/check_answers", locals: { case_log: @case_log, subsection_pages: subsection_pages }
+    render "form/check_answers", locals: { case_log: @case_log, subsection_pages: subsection_pages, subsection: subsection }
   end
 
   form = Form.new(2021, 2022)
--- a/app/controllers/case_logs_controller.rb
+++ b/app/controllers/case_logs_controller.rb
@@ -44,7 +44,7 @@ class CaseLogsController < ApplicationController
     current_url = request.env["PATH_INFO"]
     subsection = current_url.split("/")[-2]
     subsection_pages = form.pages_for_subsection(subsection)
-    render "form/check_answers", locals: { case_log: @case_log, subsection_pages: subsection_pages, subsection: subsection }
+    render "form/check_answers", locals: { case_log: @case_log, subsection_pages: subsection_pages, subsection: subsection.humanize(capitalize: false) }
   end
 
   form = Form.new(2021, 2022)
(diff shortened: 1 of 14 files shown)


doc/adr/adr-005.md
```diff
@@ -0,0 +1,23 @@
...
+
+- It doesn't require additional dev work to map keys/numbers to values, we can just pull the values out directly and use them in the code, for example on the check answers page.
+
+
+
+### Drawbacks
+
+- changing the wording/casing of the answers could result in discrepancies in the database
+
+- There is a small risk that if the database is accessed by someone unauthorised they would have access to personally identifiable information.
```
baarkerlounger: Worth clarifying the data at present doesn't contain any personally identifiable information so this would only be the case if we added additional information later. Also our production database will be encrypted as well so they would also need to get past that.

spec/features/case_log_spec.rb
```diff
@@ -58,7 +59,7 @@
           when "text"
             fill_in(question.to_s, with: answer)
           when "radio"
-            choose("#{question.to_s.tr('_', '-')}-#{answer}-field")
+            choose("#{question.to_s.tr('_', '-')}-#{answer.parameterize}-field")
```
baarkerlounger: Could make this more "railsy" by doing `question.to_s.underscore` instead of the string substitution I think but this is fine too

dushan-madetech: Thanks for the tip 😄

app/helpers/check_answers_helper.rb
```diff
@@ -0,0 +1,23 @@
+module CheckAnswersHelper
+  def get_answered_questions_total(subsection_pages, case_log)
+    questions = []
+    subsection_pages.keys.each do |page|
```
baarkerlounger: Could be wrong but I _think_  this will only work for pages that only have a single question on them? (Which is most pages but not strictly all).

Think you might need to loop over `questions_for_page` as well

app/controllers/case_logs_controller.rb
```diff
@@ -28,13 +28,29 @@ def next_page
...
     params.permit(questions_for_page)
   end
 
+  def check_answers
+    @case_log = CaseLog.find(params[:case_log_id])
+    form = Form.new(2021, 2022)
+    current_url = request.env['PATH_INFO']
+    subsection = current_url.split('/')[-2]
+    subsection_pages = form.pages_for_subsection(subsection)
+    render "form/check_answers", locals: { case_log_id: @case_log.id, case_log: @case_log, subsection_pages: subsection_pages }
```
kosiakkatrina: is it worth at this point getting rid of case_log_id here any just use the case_log?

app/views/form/check_answers.html.erb
```diff
@@ -0,0 +1,24 @@
...
+          <%= page_info["check_answer_label"].to_s %>
+        <dt>
+        <dd class="govuk-summary-list__value">
+          <%= case_log[page] %>
+        </dd>
+        <dd class="govuk-summary-list__actions">
+          <%= create_update_answer_link(case_log[page], case_log_id, page)%>
+        </dd>
+      </div>
+    </dl>
```
kosiakkatrina: would it make to extract this into a partial?

dushan-madetech: Probably 😄

app/helpers/check_answers_helper.rb
kosiakkatrina: Can we TDD that out at this stage?

spec/features/case_log_spec.rb
kosiakkatrina: Oh, we had to use dasherize 😅

app/helpers/check_answers_helper.rb
baarkerlounger: Might be worth checking this by writing a test that checks it counts the correct number of questions for a section where some pages have multiple questions per age like household_needs or rent.

dushan-madetech: I think that's roughly the idea I had for checking this. Thanks for the advice 😄

kosiakkatrina: @baarkerlounger I've updated this, but could you please check if I correctly understood how the json schema is structured, before I update the rest of the helper methods based on the same idea? It's [this](https://github.com/communitiesuk/mhclg-data-collection-beta/pull/25/commits/bbf32a2ebd6f9d4b9f71c57e53a18c313a33e6f1) commit

app/helpers/check_answers_helper.rb
```diff
@@ -0,0 +1,22 @@
+module CheckAnswersHelper
+  def get_answered_questions_total(subsection_pages, case_log)
+    questions = subsection_pages.values.map do |page|
+      page["questions"].keys
+    end
+    return questions.count {|questions_for_page| questions_for_page.none? { |question| case_log[question].blank?}}
```
baarkerlounger: How about

```suggestion
   questions = subsection_pages.values.flat_map do |page| 
       page["questions"].keys 
    end
   questions.count { |question| case_log[question].present? }
```

kosiakkatrina: Hmm, that fails the tests, is it meant to be just a refactor or a different logic ?

baarkerlounger: Different logic I think.

describe "Get answered questions total" do

it "returns 0 if only 1 question on a page gets answered" do

Doesn't make sense to me? I'd expect this to return 1 if 1 question has been answered but maybe I'm missing what this is supposed to do?

kosiakkatrina: So from what I understand, we've got check_answers page which displays a list of pages.

The 'number of questions' wording is confusing, because I think we're actually trying to say how many pages out of all in subsection have been filled in. So if one page has only half the questions completed this method would return 0.

Is that not how it's meant to work?

kosiakkatrina: Ignore that, we're just chatting about it now

baarkerlounger (approved):

PR #26, opened 2021-09-30
Add ADR record for JSON config approach

baarkerlounger:


dushan-madetech (approved):
Well explained, no comments from me 👍. 🚀 it

PR #29, opened 2021-10-04
CLDC-341: Conditional showing and hiding of question on a page

baarkerlounger:
This uses Stimulus JS to organise our javascript. By having a generic controller for showing and hiding conditional questions we can drive changes to that behaviour purely from the JSON form definition with no further code changes. Currently it only works for radio questions. The current form needs it to be replicated for numeric questions also but I'll do that in a separate PR.

- [x] Conditionally show or hide questions on a page based on the answer to an earlier question on the same page
- [x] Clear answers to conditional questions, if we revert the earlier question

Currently passing JSON with arrays of strings as HTML attributes on Form Elements doesn't work well because of https://github.com/DFE-Digital/govuk-formbuilder/issues/318

Need to resolve that before this could be merged


app/javascript/controllers/conditional_question_controller.js
```diff
@@ -0,0 +1,27 @@
...
+    this.displayConditional()
+  }
+
+  displayConditional() {
+    if(this.element.checked) {
+      let selected = this.element.value
+      let conditional_for = JSON.parse(this.element.dataset.info)
+
+      Object.entries(conditional_for).forEach(([key, values]) => {
+        let el = document.getElementById(key + "_div")
```
dushan-madetech: I think this would be nicer if it was `element` or `div`

Gemfile
```diff
@@ -19,7 +19,7 @@ gem "jbuilder", "~> 2.7"
 gem "bootsnap", ">= 1.4.4", require: false
 # Gov.UK frontend components
 gem "govuk-components"
-gem "govuk_design_system_formbuilder"
+gem "govuk_design_system_formbuilder", git: "https://github.com/DFE-Digital/govuk-formbuilder.git", branch: "whitelist-mergeable-html-attributes"
```
dushan-madetech: I'm assuming this will be updated when the referenced branch gets merged in to master?

app/views/form/_radio_question.html.erb
```diff
@@ -6,7 +6,13 @@
...
-      <% else %> 
+      <% elsif question["conditional_for"] %>
+        <%= f.govuk_radio_button question_key, val, label: { text: val },
+            "data-controller": "conditional-question",
+            "data-action": "conditional-question#displayConditional",
+            "data-info": question["conditional_for"].to_json
+        %>
+      <% else %>
         <%= f.govuk_radio_button question_key, val, label: { text: val } %>
       <% end %>
```
dushan-madetech, line 17: This seems like a lot of conditional logic in a view. I feel like this should be extracted to a helper so that the view is just focused on presenting stuff. I think it would also allow you to add some unit tests for the extracted code

dushan-madetech: I'm happy for a note of some kind to be made (like a todo comment in the code or a ticket in the backlog) to do this in the future since this is a fairly large PR already 🙂

app/views/form/page.html.erb
```diff
@@ -14,7 +14,13 @@
...
 
     <% page_info["questions"].map do |question_key, question| %>
-      <%= render partial: "form/#{question["type"]}_question", locals: { question_key: question_key, question: question, f: f } %>
+      <% if conditional_questions_for_page(page_info).include?(question_key) %>
+        <div id=<%= "#{question_key}_div" %> style="display:none;">
+      <% else %>
+        <div id=<%= "#{question_key}_div" %> >
+      <% end %>
+        <%= render partial: "form/#{question["type"]}_question", locals: { question_key: question_key, question: question, f: f } %>
+      </div>
```
dushan-madetech: I also think this should be in a helper since it's conditional logic so should belong outside of a view. Though also happy for a note to be made for this to be done in a future PR as this one is already fairly large 🙂

spec/features/case_log_spec.rb
```diff
@@ -218,4 +218,26 @@ def fill_in_number_question(case_log_id, question, value)
...
+  describe "Conditional questions" do
+    context "given a page where some questions are only conditionally shown, depending on how you answer the first question" do
+      it "initially hides conditional questions" do
+        visit("/case_logs/#{id}/armed_forces")
+        expect(page).not_to have_selector("#armed_forces_injured_div")
+      end
+
+      it "shows conditional questions if the required answer is selected and hides it again when a different answer option is selected", js: true do
+        visit("/case_logs/#{id}/armed_forces")
+        # Someting about our styling makes the selenium webdriver think the actual radio buttons are not visible so we allow label click here
```
dushan-madetech:
```suggestion
        # Something about our styling makes the selenium webdriver think the actual radio buttons are not visible so we allow label click here
```

spec/features/case_log_spec.rb
```diff
@@ -218,4 +218,26 @@ def fill_in_number_question(case_log_id, question, value)
...
   end
+
+  describe "Conditional questions" do
+    context "given a page where some questions are only conditionally shown, depending on how you answer the first question" do
+      it "initially hides conditional questions" do
+        visit("/case_logs/#{id}/armed_forces")
+        expect(page).not_to have_selector("#armed_forces_injured_div")
+      end
+
+      it "shows conditional questions if the required answer is selected and hides it again when a different answer option is selected", js: true do
```
dushan-madetech, line 229: I feel like this may benefit being two tests, one that tests the conditional questions appear given selection of the appropriate option and another that tests that the conditional questions are hidden when you select a different option given that your current selection is revealing the conditional questions. Currently it feels like this tests is doing two sets of acts and asserts or is that fine since this is a feature test?

dushan-madetech (approved):
Left a couple of comments for discussion/consideration but other than that this PR is looking good 👍

Gemfile
baarkerlounger: Yep, once they release we'll just revert this and bump.

spec/features/case_log_spec.rb
baarkerlounger: It is doing multiple things yes. My reasoning for making it a "single test" was mostly that this is sort of a single feature/user journey, so you can't really test the second step here without doing the first. Duplicating the first just to have it be it's own test didn't really seem worth it?

dushan-madetech: That's fair enough, I think if it was a unit test I'd be more concerned about it being a single test 🙂

app/views/form/_radio_question.html.erb
baarkerlounger: That might be hard to do without circumventing the benefits of the form builder gem?

PR #31, opened 2021-10-05
CLDC-341 - Conditional Questions

baarkerlounger:
- [x] Add remaining conditional questions that depend on radio question answers
- [x] Add conditional Other Household Members questions
- [x] Fix section progress count/status based on conditional questions


app/helpers/check_answers_helper.rb
```diff
@@ -1,39 +1,67 @@
...
+      question.fetch("conditional_for", []).map do |conditional_question_key, condition|
+        if condition_not_met(case_log, question_key, question, condition)
+          questions_not_applicable << conditional_question_key
+        end
+      end
+      questions_not_applicable.include?(question_key)
+    end
+  end
+
+  def condition_not_met(case_log, question_key, question, condition)
```
baarkerlounger, line 26: It's not ideal that we're now evaluating these conditions in two places (Ruby side and JS side), and that the evaluation of them is relatively brittle, but given how few use cases we currently have I think it's OK to leave refactoring this out until we have more and/or a better abstraction becomes clear.

app/helpers/stimulus_controller_attribute_helper.rb
```diff
@@ -0,0 +1,39 @@
+module StimulusControllerAttributeHelper
```
kosiakkatrina: Extracting these methods like this sounds very dependent on the technology we're using. I'm aware that there are very stimulus specific things going on here but I think I'd rename it to something more generic, like questions attribute helper etc. What do you think?

app/javascript/controllers/conditional_question_controller.js
```diff
@@ -6,13 +6,25 @@ export default class extends Controller {
...
   displayConditional() {
+    switch(this.element.type) {
+      case "number":
+        this.displayConditionalNumeric()
+      case "radio":
+        this.displayConditionalRadio()
+      default:
+        console.log("Not yet implemented for " + this.element.type)
+        break;
+    }
```
kosiakkatrina, line 17: This will likely grow, I take it there's no option to pass a method through `data-info` 😅 Have you had any ideas of how this could be structured differently?

app/javascript/controllers/conditional_question_controller.js
```diff
@@ -24,4 +36,18 @@ export default class extends Controller {
       })
     }
   }
+
+  displayConditionalNumeric() {
+    let value = this.element.value
+    let conditional_for = JSON.parse(this.element.dataset.info)
+
+    Object.entries(conditional_for).forEach(([key, values]) => {
```
kosiakkatrina: Maybe would be easier to read if 'key' was 'question' and 'values' would be whatever values are (answer?)

app/javascript/controllers/conditional_question_controller.js
```diff
@@ -24,4 +36,18 @@ export default class extends Controller {
...
+      let div = document.getElementById(key + "_div")
+      if(eval((value + values))) {
```
kosiakkatrina: Just by reading this I'm not entirely sure what's going on in this line. Seems extreme, but maybe extracting it and naming it for readability would be good?

app/javascript/controllers/conditional_question_controller.js
baarkerlounger: I suppose it could grow to as many paths as we have question types, which is currently 5 (text, numeric, date, checkbox, radio). I suppose there could be a couple more but it doesn't feel like an unreasonable amount?

Possibly the method for showing/hiding could be refactored to be more generic so you don't need the split?

app/javascript/controllers/conditional_question_controller.js
baarkerlounger: Renamed the variables which hopefully makes this clearer. What do you think?

kosiakkatrina: Way better!

app/javascript/controllers/conditional_question_controller.js
kosiakkatrina: Yeah, maybe. Happy to leave it as is, just a general hate for switch statements

kosiakkatrina (approved):
Happy for this to go in!

PR #32, opened 2021-10-05
Cldc 275 focus styling 2

kosiakkatrina:
Copy of https://github.com/communitiesuk/mhclg-data-collection-beta/pull/30 with clean git changes


app/javascript/packs/tasklist.js
```diff
@@ -0,0 +1,3 @@
+window.addHighlight = function(elementId) {
```
baarkerlounger: This file isn't need anymore right?

kosiakkatrina: 🤦 You're right!

baarkerlounger (approved):

PR #33, opened 2021-10-05
Render html from JSON questions

kosiakkatrina:
Render html given in question headers: 
image


app/views/form/_radio_question.html.erb
```diff
@@ -1,5 +1,5 @@
 <%= f.govuk_radio_buttons_fieldset question_key,
-    legend: { text: question["header"], size: "l" },
+    legend: { text: question["header"].html_safe, size: "l" },
```
baarkerlounger, line 2: Should we add this to the other question type partials so that they're capable of rendering this too?

baarkerlounger (approved):

app/views/form/_radio_question.html.erb
kosiakkatrina: great idea! 😅

PR #34, opened 2021-10-06
Cldc 340 read only questions

kosiakkatrina:
* Make number input fields readonly based on the value in form json
* Set total value in rent section to be readonly
* Add missing check answer labels to the form json


config/forms/2021_2022.json
```diff
@@ -1168,7 +1176,8 @@
                   "hint_text": "This is the total of rent and all charges",
                   "type": "numeric",
                   "min": 0,
-                  "step": 1
+                  "step": 1,
+                  "readonly": true
```
baarkerlounger, line 1188: Think this is missing some way of saying that the field needs to display the total of the other rent/charge fields on this page? Currently it would just be an empty non-editable field right?

kosiakkatrina: Yup! I thought that wasn't part of the ticket 😅 
Will add that

app/views/form/_numeric_question.html.erb
```diff
@@ -2,5 +2,9 @@
     hint: { text: question["hint_text"] },
     label: { text: question["header"].html_safe, size: "l"},
     min: question["min"], max: question["max"], step: question["step"],
-    width: 20, :readonly => question["readonly"]
+    width: 20, :readonly => question["readonly"],
+    :"data-controller" => "number-question",
+    :"data-action"=> "number-question#calculateFields",
+    :"data-affected" => "#{question["result-field"].to_s.dasherize}-field",
```
baarkerlounger: Could call this `data-target` to match the intent to Stimulus naming?

baarkerlounger: Could also put these attributes in a helper like:

https://github.com/communitiesuk/mhclg-data-collection-beta/pull/31/files#diff-1071361721dc5b6fcfa551cb18a7a1e555de469ef8e7193470c2090e08077c0eR5

baarkerlounger (approved):

app/views/form/_numeric_question.html.erb
kosiakkatrina: Ah nice! Thanks, I think I've done that now and it works fine!

kosiakkatrina: Will merge unless any objections

baarkerlounger (approved):
This looks great. :rocket:

PR #35, opened 2021-10-06
Update README.md

paulrobertlloyd:
Started off by updating the installation instructions. I had to run `yarn install` first else I would get an error when first starting the app. Also needed to prefix the `rails s` command with `bundle exec` otherwise I got a Bundler error.

Then I got a bit carried away a bit and:
* Tweaked the Markdown (to quieten my IDEs linter),
* Used the correct capitalisation for products and services,
* Attempted to update the JSON example so that it a.) it uses valid JSON and b.) works with syntax highlighting.

A pedantic PR? Almost certainly! A useful PR? I hope so 🙂


baarkerlounger (approved):

baarkerlounger:
:rocket:

PR #36, opened 2021-10-06
CLDC-344: Validations

baarkerlounger:
Adding validations properly requires a bit of refactoring here:

1. We need to use `form_with @model` helper to bind our form to our ActiveModel object as that's where the Form Builder expects to find error messages (form.object.errors.messages)
2. That requires a bit of refactoring of tests and JS since it means form fields now get generated as `case-log-question-field` rather than just `question`
3. We also refactor out a bit of logic to determine next page paths from the controller into the form class and rename the `next_page` method to submit form since it now does more than just determine the next page
4. Finally, in order for Turbo to realise it needs to re-render the same form with an error rather than expecting a redirect, we need to return status 422

This PR lays the groundwork for displaying validations but doesn't yet implement any (testing example in CaseLog model is commented out).

image

### DIFF
--- a/app/controllers/case_logs_controller.rb
+++ b/app/controllers/case_logs_controller.rb
@@ -20,24 +20,16 @@ class CaseLogsController < ApplicationController
     render :edit
   end
 
-  def next_page
+  def submit_form
     form = Form.new(2021, 2022)
     @case_log = CaseLog.find(params[:id])
     previous_page = params[:case_log][:previous_page]
     questions_for_page = form.questions_for_page(previous_page).keys
     answers_for_page = page_params(questions_for_page).select { |k, _v| questions_for_page.include?(k) }
-    begin
-      @case_log.update!(answers_for_page)
-      next_page = form.next_page(previous_page)
-      redirect_path = if next_page == :check_answers
-                        subsection = form.subsection_for_page(previous_page)
-                        "case_log_#{subsection}_check_answers_path"
-                      else
-                        "case_log_#{next_page}_path"
-                      end
-
+    if @case_log.update(answers_for_page)
+      redirect_path = form.next_page_redirect_path(previous_page)
       redirect_to(send(redirect_path, @case_log))
-    rescue ActiveRecord::RecordInvalid
+    else
       page_info = form.all_pages[previous_page]
       render "form/page", locals: { form: form, page_key: previous_page, page_info: page_info }, status: :unprocessable_entity
     end
--- a/app/controllers/case_logs_controller.rb
+++ b/app/controllers/case_logs_controller.rb
@@ -22,20 +22,25 @@ class CaseLogsController < ApplicationController
 
   def next_page
     form = Form.new(2021, 2022)
-    @case_log = CaseLog.find(params[:case_log_id])
-    previous_page = params[:previous_page]
+    @case_log = CaseLog.find(params[:id])
+    previous_page = params[:case_log][:previous_page]
     questions_for_page = form.questions_for_page(previous_page).keys
     answers_for_page = page_params(questions_for_page).select { |k, _v| questions_for_page.include?(k) }
-    @case_log.update!(answers_for_page)
-    next_page = form.next_page(previous_page)
-    redirect_path = if next_page == :check_answers
-                      subsection = form.subsection_for_page(previous_page)
-                      "case_log_#{subsection}_check_answers_path"
-                    else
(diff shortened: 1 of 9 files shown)


app/javascript/controllers/conditional_question_controller.js
```diff
@@ -16,7 +16,7 @@ export default class extends Controller {
           div.style.display = "block"
         } else {
           div.style.display = "none"
-          let buttons = document.getElementsByName(key)
+          let buttons = document.getElementsByName("case_log[" + key + "]")
```
MatthewPhelanMadeTech: Potentially too picky but would this be better with string interpolation rather than appending strings - such as `` `case_log[${key}]` ``

app/helpers/conditional_questions_helper.rb
```diff
@@ -6,10 +6,6 @@ def conditional_questions_for_page(page)
   end
 
   def display_question_key_div(page_info, question_key)
-    if conditional_questions_for_page(page_info).include?(question_key)
-      "<div id=#{question_key}_div style='display:none;'>".html_safe
-    else
-      "<div id=#{question_key}_div>".html_safe
-    end
+    "style='display:none;'".html_safe if conditional_questions_for_page(page_info).include?(question_key)
```
MatthewPhelanMadeTech, line 9: Love this

spec/features/case_log_spec.rb
```diff
@@ -17,12 +17,12 @@
...
-    choose("net-income-frequency-yearly-field")
+    fill_in("case-log-net-income-field", with: 18_000)
+    choose("case-log-net-income-frequency-yearly-field")
     click_button("Save and continue")
-    choose("net-income-uc-proportion-all-field")
+    choose("case-log-net-income-uc-proportion-all-field")
     click_button("Save and continue")
-    choose("housing-benefit-housing-benefit-but-not-universal-credit-field")
+    choose("case-log-housing-benefit-housing-benefit-but-not-universal-credit-field")
     click_button("Save and continue")
```
MatthewPhelanMadeTech, line 26: should we have an expect here?

MatthewPhelanMadeTech (commented):
Looking great! Just a couple of potentially personal preference comments. All working well

app/controllers/case_logs_controller.rb
```diff
@@ -20,22 +20,19 @@ def edit
...
-    next_page = form.next_page(previous_page)
-    redirect_path = if next_page == :check_answers
-                      subsection = form.subsection_for_page(previous_page)
-                      "case_log_#{subsection}_check_answers_path"
-                    else
-                      "case_log_#{next_page}_path"
-                    end
-
-    redirect_to(send(redirect_path, @case_log))
+    if @case_log.update(answers_for_page)
```
MatthewPhelanMadeTech, line 29: This will mean it will validate for tenant_age even when submitting for tenant_code - this is something I can work out on my branch however thought it worth mentioning

app/controllers/case_logs_controller.rb
```diff
@@ -20,22 +20,19 @@ def edit
     render :edit
   end
 
-  def next_page
+  def submit_form
```
dushan-madetech, line 23: Is this submitting the form or technically just the question(s) you've answered on the current page?

spec/features/case_log_spec.rb
baarkerlounger: No it's not a spec in itself, it's a method other specs are calling.

app/controllers/case_logs_controller.rb
baarkerlounger: Yeah this isn't meant to be a complete implementation by any means, just the basis for your work to start really.

spec/features/case_log_spec.rb
dushan-madetech: No, this method used to answer all the questions in a subsection as part of the test setup for any test relying on a subsection being completed e.g. check answers page tests.

dushan-madetech (commented):
Just a few thoughts, wasn't sure if you wanted an approval for this or not

app/controllers/case_logs_controller.rb
baarkerlounger: I suppose this confusion comes from a slight overloading of the term form. Was referring here to the html form, which every page has, as opposed to the `Form` class that's doing the config parsing. Happy to change this though if you have a better name.

PR #38, opened 2021-10-07
CLDC-512: Display answered fields

kosiakkatrina:
Answered fields are now displaying by default, I assume due to [this PR](https://github.com/communitiesuk/mhclg-data-collection-beta/pull/36)

In current PR: add tests for prepopulating answered questions

Will likely need more work after CLDC-342 ticket which allows saving values for multiple choice questions.

### DIFF
--- a/spec/features/case_log_spec.rb
+++ b/spec/features/case_log_spec.rb
@@ -15,6 +15,12 @@ RSpec.describe "Test Features" do
     household_number_of_other_members: { type: "numeric", answer: 2 },
   }
 
+  def fill_in_number_question(case_log_id, question, value)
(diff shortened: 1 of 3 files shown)


spec/features/case_log_spec.rb
```diff
@@ -15,6 +15,12 @@
     household_number_of_other_members: { type: "numeric", answer: 2 },
   }
 
+  def fill_in_number_question(case_log_id, question, value)
```
baarkerlounger, line 18: Rather than filling in the values through Capybara (slower) you could just create a test object that has the values you need with FactoryBot (faster)?

PR #39, opened 2021-10-08
Cldc 344/implement a few validations

baarkerlounger:


app/models/case_log.rb
```diff
@@ -1,5 +1,32 @@
-class CaseLog < ApplicationRecord
-  enum status: { "in progress" => 0, "submitted" => 1 }
+class CaseLogValidator < ActiveModel::Validator
+
+  def validate_tenant_age(record, validation_regex)
+    regexp = Regexp.new validation_regex
+    if !record.tenant_age?
+      record.errors.add :base, "Tenant age can't be blank"
```
baarkerlounger: Possibly you've already fixed this but wondering if this is why it's only being shown on the summary and not next to the field? Because it's being added to :base here rather than the field?

MatthewPhelanMadeTech: yeah I updated this, no change unfortunately

config/forms/2021_2022.json
```diff
@@ -32,7 +32,8 @@
                   "type": "numeric",
                   "min": 0,
                   "max": 150,
-                  "step": 1
+                  "step": 1,
+                  "regex_validation" : "^[1-9][0-9]?$|^100$"
```
baarkerlounger: Given that I think for some validations the logic is likely to be quite complex/not suitable for a regex, I think I'd be inclined to not put it into the JSON for now, or if you do, just put in a method name, where that method contains the validation to be run?

Might be prematurely optimising here though so feel free to ignore that.

app/models/case_log.rb
```diff
@@ -1,5 +1,32 @@
-class CaseLog < ApplicationRecord
-  enum status: { "in progress" => 0, "submitted" => 1 }
+class CaseLogValidator < ActiveModel::Validator
```
baarkerlounger, line 1: Might be useful to have validator classes somewhat grouped by domain?

Like one that validates person/tenant information, one for property information etc?

app/controllers/case_logs_controller.rb
```diff
@@ -26,6 +26,7 @@ def submit_form
     previous_page = params[:case_log][:previous_page]
     questions_for_page = form.questions_for_page(previous_page).keys
     answers_for_page = page_params(questions_for_page).select { |k, _v| questions_for_page.include?(k) }
+    @case_log.custom_validator_options = { previous_page: previous_page, validation: form.questions_for_page(previous_page)[previous_page]["regex_validation"] }
```
baarkerlounger: This feels like a slight OOP anti-pattern, could we make this a method on CaseLog?

app/models/case_log.rb
```diff
@@ -1,5 +1,32 @@
...
+
+  def validate_tenant_age(record, validation_regex)
+    regexp = Regexp.new validation_regex
+    if !record.tenant_age?
```
dushan-madetech: Could you use `if record.tenant_age.blank?` here instead? I think it might be clearer than the `!`

app/models/case_log.rb
```diff
@@ -1,5 +1,32 @@
...
+      record.errors.add :tenant_age, "Tenant age must be between 0 and 100"
+    end
+  end
 
-  # validates :tenant_age, presence: true
+  def validate(record)
+    question_to_validate = options[:previous_page]
+    validation_regex = options[:validation]
+    if question_to_validate == "tenant_code"
+      if !record.tenant_code?
```
dushan-madetech: Same question as line 5, can we use `if record.tenant_code.blank?` here?

baarkerlounger: Think rubocop would insist on that as well

app/models/case_log.rb
```diff
@@ -1,5 +1,32 @@
...
+        record.errors.add :tenant_code, "Tenant code can't be blank"
+      end
+    elsif question_to_validate == "tenant_age"
+        validate_tenant_age(record, validation_regex)
+    end
```
dushan-madetech: I imagine this method might grow as more validations are added so you could potentially rewrite it to make it a switch statement so that it's easier to add to

baarkerlounger: Could also do some metaprogramming like we did with routes and controller methods so every question has a `validate_#{question}` method that gets called automatically from submit_form or something like that...

dushan-madetech: Thinking something like 
```
switch question_to_validate
  case: tenant_age
    age validation method
  case: tenant_code
    code validation method
   default
     some default behaviour
```
Not the cleanest but should give you an idea

dushan-madetech: I'd also say extract lines 16-17 to their own method you can call e.g. `validate_tenant_code` then when future validation methods are added it might be possible to refactor them all down to a single method

config/forms/2021_2022.json
MatthewPhelanMadeTech: My hope was to keep all the bits of the form that are subject to change in the same domain. However, in the interest of keeping it as simple as possible I will leave it in it's respective validation method.

app/models/case_log.rb
MatthewPhelanMadeTech: This feels like prematurely optimising as well I think this abstraction would be better in a later ticket when it's necessary.

app/controllers/case_logs_controller.rb
MatthewPhelanMadeTech: I tried this extensively, I couldn't get a setter to work, will have another look with fresh eyes this morning

app/models/case_log.rb
MatthewPhelanMadeTech: Love this metaprogramming idea, I'm on it

app/models/case_log.rb
MatthewPhelanMadeTech: good call!

app/controllers/case_logs_controller.rb
MatthewPhelanMadeTech: Actually fairly simple now I'm fresh, thanks

app/models/case_log.rb
```diff
@@ -1,5 +1,41 @@
...
+    question_to_validate = options[:previous_page]
+    if respond_to?("validate_#{question_to_validate}")
+      public_send("validate_#{question_to_validate}", record)
+    end
+  end
+end
+
+class CaseLog < ApplicationRecord 
+  validate :instance_validations
+  @previous_page
```
baarkerlounger: You'd typically use `attr_writer` here to add a setter method (or accessor if you want a getter and setter)

app/controllers/case_logs_controller.rb
```diff
@@ -26,6 +26,7 @@ def submit_form
     previous_page = params[:case_log][:previous_page]
     questions_for_page = form.questions_for_page(previous_page).keys
     answers_for_page = page_params(questions_for_page).select { |k, _v| questions_for_page.include?(k) }
+    @case_log.previous_page(previous_page)
```
baarkerlounger: Any reason not to validate for page directly here? Something like:

`@case_log.validate_page(previous_page)`

app/models/case_log.rb
```diff
@@ -1,5 +1,41 @@
...
+  def validate_tenant_age(record)
+    if record.tenant_age.blank?
+      record.errors.add :tenant_age, "Tenant age can't be blank"
+    elsif !(record.tenant_age.to_s =~ /^[1-9][0-9]?$|^100$/)
+      record.errors.add :tenant_age, "Tenant age must be between 0 and 100"
+    end
+  end
+
+  def validate(record)
+    question_to_validate = options[:previous_page]
```
baarkerlounger, line 22: I think this might have to be a loop? If you had multiple things on a page to validate?

baarkerlounger:
Probably need a couple of specs for this before merging (and linting), but otherwise looking good!

app/controllers/case_logs_controller.rb
MatthewPhelanMadeTech: The ActiveModel::Validator pattern came out of the box so I stuck with it

app/models/case_log.rb
MatthewPhelanMadeTech: Calling validate in the model class will loop over all the fields for us, we shouldn't have to write one ourselves

app/models/case_log.rb
MatthewPhelanMadeTech: hahah I tried this on Friday and couldn't get it going, all good now

app/models/case_log.rb
```diff
@@ -1,5 +1,41 @@
...
+  end
+end
+
+class CaseLog < ApplicationRecord 
+  validate :instance_validations
+  @previous_page
   enum status: { "in progress" => 0, "submitted" => 1 }
 
-  # validates :tenant_age, presence: true
+  def previous_page(value)
```
baarkerlounger: If you have the `attr_writer` above it'll autogenerate this for you (rails magic)

PR #40, opened 2021-10-08
CLDC-342: Save checkbox answers

kosiakkatrina:
In case log controller individually build and save a hash of answers for checkbox questions.
Change the keys for answer_options within checkbox JSON questions from numbers to text representing a table column.
Add new checkbox fields into the database.


spec/controllers/case_logs_controller_spec.rb
```diff
@@ -43,4 +43,41 @@
...
+        "previous_page" => "accessibility_requirements" }
+    end
+
+    it "sets checked items to true" do
+      get :submit_form, params: { id: id, case_log: case_log_to_submit }
+      CaseLog.find(id)
+
+      expect(CaseLog.find(id)["accessibility_requirements_fully_wheelchair_accessible_housing"]).to eq(true)
+      expect(CaseLog.find(id)["accessibility_requirements_wheelchair_access_to_essential_rooms"]).to eq(true)
+      expect(CaseLog.find(id)["accessibility_requirements_level_access_housing"]).to eq(true)
```
baarkerlounger: I think you can just do `case_log.reload` after your get request and then you can remove the 3 look ups here? That way you load the object once rather than 4 times so should be faster.

kosiakkatrina: Oh, I mean, would that be any different from extracting CaseLog.find(id) into a variable?

baarkerlounger: Not really no, pretty much the same thing. Just makes it a bit more explicit it's definitely the same object I guess rather than needing to check id is the same.

kosiakkatrina: Ahh, ok I see!

app/controllers/case_logs_controller.rb
```diff
@@ -20,12 +20,32 @@ def edit
...
+    case_log.each do |question, answer|
+      if question == "previous_page"
+        result[question] = answer
+      elsif questions_for_page[question]["type"] == "checkbox"
+        questions_for_page[question]["answer_options"].keys.reject { |x| x.match(/divider/) }.each do |option|
+          result[option] = case_log[question].include?(option) ? true : false
+        end
+      else
+        result[question] = answer
+      end
```
dushan-madetech: I'm wondering if there's a way we can shorten this? I'm noticing that the condition on like 26 gives the same result as the else on line 32

dushan-madetech (approved):
Left a single comment but this looks good to me, merge when you're ready 👍

db/migrate/20211011115946_rename_economic_status_fields.rb
```diff
@@ -0,0 +1,14 @@
...
+  def change
+    change_table :case_logs, bulk: true do |t|
+      t.rename :person_2_economic, :person_2_economic_status
+      t.rename :person_3_economic, :person_3_economic_status
+      t.rename :person_4_economic, :person_4_economic_status
+      t.rename :person_5_economic, :person_5_economic_status
+      t.rename :person_6_economic, :person_6_economic_status
+      t.rename :person_7_economic, :person_7_economic_status
+      t.rename :person_8_economic, :person_8_economic_status
+      t.rename :postcode, :property_postcode
```
kosiakkatrina, line 11: We already had property postcode, it was added in "add_tenancy_fields_to_case_log" migration, I think this should have been drop :postcode. 
wouldn't this destroy the data we had in property postcode as that's the one listed in the form?

baarkerlounger: Yes it should have been. Apologies for that.

PR #41, opened 2021-10-11
Navigate straight to check answers page if a section has been started…

baarkerlounger:
… already

### DIFF
--- a/app/helpers/tasklist_helper.rb
+++ b/app/helpers/tasklist_helper.rb
@@ -37,11 +37,12 @@ module TasklistHelper
   end
 
   def get_first_page_or_check_answers(subsection, case_log, form, questions)
-    if is_started?(subsection, case_log, questions)
-      send("case_log_#{subsection}_check_answers_path", case_log)
-    else
-      send("case_log_#{form.first_page_for_subsection(subsection)}_path", case_log)
-    end
+    path = if is_started?(subsection, case_log, questions)
+             "case_log_#{subsection}_check_answers_path"
+           else
+             "case_log_#{form.first_page_for_subsection(subsection)}_path"
+           end
+    send(path, case_log)
   end
 
 private
--- a/app/helpers/tasklist_helper.rb
+++ b/app/helpers/tasklist_helper.rb
@@ -36,6 +36,14 @@ module TasklistHelper
     subsections.count { |subsection| get_subsection_status(subsection, case_log, form.questions_for_subsection(subsection).keys) == status }
   end
 
+  def get_first_page_or_check_answers(subsection, case_log, form, questions)
+    if is_started?(subsection, case_log, questions)
+      send("case_log_#{subsection}_check_answers_path", case_log)
+    else
+      send("case_log_#{form.first_page_for_subsection(subsection)}_path", case_log)
+    end
+  end
+
 private
 
   def all_questions_completed(case_log)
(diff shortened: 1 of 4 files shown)


app/helpers/tasklist_helper.rb
```diff
@@ -36,6 +36,14 @@ def get_sections_count(form, case_log, status = :all)
     subsections.count { |subsection| get_subsection_status(subsection, case_log, form.questions_for_subsection(subsection).keys) == status }
   end
 
+  def get_first_page_or_check_answers(subsection, case_log, form, questions)
+    if is_started?(subsection, case_log, questions)
+      send("case_log_#{subsection}_check_answers_path", case_log)
+    else
+      send("case_log_#{form.first_page_for_subsection(subsection)}_path", case_log)
+    end
+  end
```
kosiakkatrina: Not sure if it's nicer, but we could have 1 send statement if we did it like this?

```
path = is_started?(subsection, case_log, questions) ? "case_log_#{subsection}_check_answers_path" :
 "case_log_#{form.first_page_for_subsection(subsection)}_path"
 send(path, case_log)
 ```

Comment (kosiakkatrina) [reply app/helpers/tasklist_helper.rb]: Long first line tho 🤷

File: app/helpers/tasklist_helper.rb [hunk 2]
```diff
@@ -46,4 +54,9 @@ def is_incomplete?(subsection, case_log, questions)
     status = get_subsection_status(subsection, case_log, questions)
     %i[not_started in_progress].include?(status)
   end
+
+  def is_started?(subsection, case_log, questions)
+    status = get_subsection_status(subsection, case_log, questions)
+    %i[in_progress completed].include?(status)
+  end
```
kosiakkatrina, line 62: Would it make sense to turn this method and the method above (is_incomplete?) into a single method and pass in the statuses we expect the subsection to have?
It might mean less duplication if we have other status combinations in the future, but it might make it less readable as `is_incomplete?` and `is_started?` are very clear. I could be swayed either way

spec/helpers/tasklist_helper_spec.rb
```diff
@@ -1,12 +1,15 @@
...
   describe "get subsection status" do
-    @form = Form.new(2021, 2022)
-    income_and_benefits_questions = @form.questions_for_subsection("income_and_benefits").keys
-    declaration_questions = @form.questions_for_subsection("declaration").keys
-    local_authority_questions = @form.questions_for_subsection("local_authority").keys
-    let!(:case_log) { FactoryBot.create(:case_log) }
+    let(:section) { "income_and_benefits" }
+    let(:income_and_benefits_questions) { form.questions_for_subsection("income_and_benefits").keys }
+    let(:declaration_questions) { form.questions_for_subsection("declaration").keys }
+    let(:local_authority_questions) { form.questions_for_subsection("local_authority").keys }
```
kosiakkatrina, line 12: nice!

kosiakkatrina (approved):
Looks good, left a few non-crucial comments. Love the test cleanup!

app/helpers/tasklist_helper.rb
baarkerlounger: How about

```
path = if is_started?(subsection, case_log, questions)
      "case_log_#{subsection}_check_answers_path"
    else
      "case_log_#{form.first_page_for_subsection(subsection)}_path"
    end
    send(path, case_log)
```

app/helpers/tasklist_helper.rb
baarkerlounger: Think it's alright to defer that to if/when we do in the future. Think this way is quite readable for now?

PR #43, opened 2021-10-12
CLDC-531: Accept a basic create request with JSON params

baarkerlounger:
- [x] Enable formt: json API request to create case log
- [x] Add Basic Auth
- [x] Set ENV variables

### DIFF
--- a/spec/requests/case_log_controller_spec.rb
+++ b/spec/requests/case_log_controller_spec.rb
@@ -1,11 +1,11 @@
-require 'rails_helper'
+require "rails_helper"
 
 RSpec.describe CaseLogsController, type: :request do
   describe "POST #create" do
     let(:headers) do
       {
-       "Content-Type" => "application/json",
-       "Accept" => "application/json"
+        "Content-Type" => "application/json",
+        "Accept" => "application/json",
       }
     end
 
@@ -17,7 +17,7 @@ RSpec.describe CaseLogsController, type: :request do
       {
         "tenant_code": tenant_code,
         "tenant_age": 35,
-        "property_postcode": property_postcode
+        "property_postcode": property_postcode,
       }
     end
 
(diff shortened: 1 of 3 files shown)


spec/requests/case_log_controller_spec.rb
```diff
@@ -0,0 +1,64 @@
...
+    let(:params) do
+      {
+        "tenant_code": tenant_code,
+        "tenant_age": 35,
+        "property_postcode": property_postcode,
+      }
+    end
+
+    before do
+      allow(ENV).to receive(:[])
```
dushan-madetech, line 32: What does this line do? I can tell what 33 and 34 do but curious about this one 🙂

dushan-madetech (approved):
I can't see anything that I think needs to be changed so I'm happy for you to 🚢 this

spec/requests/case_log_controller_spec.rb
baarkerlounger: If you don't set it then having 33 and 34 means Rails thinks those are the _only_ arguments ENV is allowed to receive for `:[]` so any other ENV vars that are looked up stop working. That means you can't for example put a `binding.pry` in your test because that relies on `ENV["DISABLE_PRY"]`

spec/requests/case_log_controller_spec.rb
```diff
@@ -0,0 +1,64 @@
...
+        "Content-Type" => "application/json",
+        "Accept" => "application/json",
+        "Authorization" => basic_credentials,
+      }
+    end
+
+    let(:params) do
+      {
+        "tenant_code": tenant_code,
+        "tenant_age": 35,
```
kosiakkatrina: `"tenant_age": 35` -> `"tenant_age": tenant_age` ?

baarkerlounger: Good catch. Only fixed that in my other branch :face_palm:

spec/requests/case_log_controller_spec.rb
dushan-madetech: Thanks for the explanation 😄

PR #45, opened 2021-10-13
CLDC-531: Part 2 - Validate API created case logs

baarkerlounger:
Part 1: #43

Here we try to have CaseLog validations run on API request params. Standard Rails presence checks can't be used here because we do want the ability to create records without those fields, we just don't want to be able to "complete" or "submit" records until they're there, so instead we use a `before_save` hook that runs any time save is called on the ActiveRecord object (i.e. on `create` or `update`), that checks whether all the required data has been completed with no errors, and sets the corresponding status.


app/controllers/case_logs_controller.rb
```diff
@@ -1,17 +1,23 @@
 class CaseLogsController < ApplicationController
-  skip_before_action :verify_authenticity_token, only: [:create], if: :json_request?
-  before_action :authenticate, only: [:create], if: :json_request?
+  skip_before_action :verify_authenticity_token, if: :json_create_request?
```
baarkerlounger, line 2: Needed because using `only` and `if` means `only` gets precedence and `if` gets ignored

MatthewPhelanMadeTech:
I'm happy to merge this

PR #46, opened 2021-10-13
CLDC-530 reduce test dependency on form definition

dushan-madetech:
This PR aims to reduce the specs dependency on the form definition as the form will change each year. A test from file has been created along with a FormHandler class that will ensure that each form is only created once rather than created the form multiple times as was previously the case


app/models/form_handler.rb
```diff
@@ -0,0 +1,18 @@
...
+  include Singleton
+
+  @forms = {}
+  # todo: add singleton
+  def self.get_form(form)
+    @forms[form] ||= Form.new(form)
+  end
+
+  def self.get_all_forms
+    Dir.foreach('config/forms') do |filename|
```
baarkerlounger: I think `Dir.glob("config/forms/*.json").each` would avoid needing to check for `.` and `..` below

app/models/form_handler.rb
```diff
@@ -0,0 +1,18 @@
...
+      next if filename == '.' or filename == '..'
+      form_name = filename.sub('.json', '')
```
baarkerlounger: Could leave the file extension in and then use this path directly in the Form class?

app/controllers/case_logs_controller.rb
```diff
@@ -49,13 +52,14 @@ def submit_form
...
+    form = use_form
     @case_log = CaseLog.find(params[:case_log_id])
     current_url = request.env["PATH_INFO"]
     subsection = current_url.split("/")[-2]
-    render "form/check_answers", locals: { case_log: @case_log, subsection: subsection }
+    render "form/check_answers", locals: { case_log: @case_log, subsection: subsection, form: form }
   end
 
-  form = Form.new(2021, 2022)
+  form = ENV["RAILS_ENV"] == "test" ? @@form_handler.get_form("test_form") : @@form_handler.get_form("2021_2022")
```
baarkerlounger: This seems like a leaky abstraction. Would suggest moving the ENV check into the form handler class and setting the form directory based on it there?

baarkerlounger: That way I think you don't need the `use_form` method, or the class var, you can just you `FormHandler.instance.each` everywhere you need the collection?

We probably also need the CaseLog itself to have some way of deriving what form it was created from, for the singular/specific case.

Gemfile
```diff
@@ -21,6 +21,7 @@ gem "bootsnap", ">= 1.4.4", require: false
 gem "govuk-components"
 gem "govuk_design_system_formbuilder"
 gem "hotwire-rails"
+gem "singleton"
```
baarkerlounger: Worth double checking if this is actually needed or if it's in the standard library now?

app/models/form_handler.rb
```diff
@@ -0,0 +1,27 @@
+class FormHandler
+  include Singleton
+
+  def initialize
+    @forms = {}
+    get_all_forms
```
baarkerlounger: Might be clearer to just say `@forms = get_all_forms`?

config/routes.rb
```diff
@@ -5,7 +5,8 @@
 
   post "/case_logs/:id", to: "case_logs#submit_form"
 
-  form = Form.new(2021, 2022)
+  form_handler = FormHandler.instance
+  form = ENV["RAILS_ENV"] == "test" ? form_handler.get_form("test_form") : form_handler.get_form("2021_2022")
```
baarkerlounger: Same as above on the leaky abstraction. I think if you move the ENV var check into the FormHandler class you only need to do it once.

spec/features/case_log_spec.rb
```diff
@@ -52,7 +49,7 @@ def answer_all_questions_in_income_subsection
       it "displays a section status" do
         visit("/case_logs/#{empty_case_log.id}")
 
-        assert_selector ".govuk-tag", text: /Not started/, count: 8
+        assert_selector ".govuk-tag", text: /Not started/, count: 7
```
baarkerlounger, line 52: I'm assuming these are changing because the test form is slightly more minimal than the original?

kosiakkatrina: Yup!

app/models/form_handler.rb
kosiakkatrina: We're directly changing @forms in get_all_forms method, so it fails when we do that and doing it a different way (creating a new var in the method and returning it) might not make sense in case get_all_forms gets called from somewhere else and we don't want to create forms again 🤔 
What do you think?

baarkerlounger: I think I'd be inclined to make `get_all_forms` more functional by just returning a {} rather than modifying @forms directly, and then possibly make it a `private` method to mark that it shouldn't really be called directly?

kosiakkatrina: Yup, sounds good!

app/controllers/case_logs_controller.rb
kosiakkatrina: Moved the check to form handler but I don't think we ever need to access the collection just yet

app/models/form_handler.rb
kosiakkatrina: Nice!

app/controllers/case_logs_controller.rb
```diff
@@ -1,6 +1,9 @@
 class CaseLogsController < ApplicationController
   skip_before_action :verify_authenticity_token, if: :json_create_request?
   before_action :authenticate, if: :json_create_request?
+  # rubocop:disable Style/ClassVars
+  @@form_handler = FormHandler.instance
```
baarkerlounger, line 5: Given it's hardly any shorter I think I'd be inclined to scrap the class var and just call `FormHandler.instance` directly but purely a style preference.

baarkerlounger (approved):

PR #48, opened 2021-10-14
Api update

baarkerlounger:
- [x] Add PATCH case log
- [x] Add PUT case log
- [x] Add DELETE (soft) case log

As well as adding additional API methods this PR also does a bit of a refactoring of our CaseLog state machine to make it easier to work with. This does mean breaking changes for data (status enum changes), which will need to be recreated or adjusted locally if this is merged. Think it's worth it to have a coherent status though.


app/helpers/tasklist_helper.rb
```diff
@@ -15,7 +15,7 @@ module TasklistHelper
 
   def get_subsection_status(subsection_name, case_log, questions)
     if subsection_name == "declaration"
-      return all_questions_completed(case_log) ? :not_started : :cannot_start_yet
+      return case_log.completed? ? :not_started : :cannot_start_yet
```
kosiakkatrina, line 18: is completed != submitted?

app/views/form/check_answers.html.erb
```diff
@@ -2,11 +2,11 @@
...
+      <%= display_answered_questions_summary(subsection, @case_log, form) %>
       <% form.pages_for_subsection(subsection).each do |page, page_info| %>
         <% page_info["questions"].each do |question_title, question_info| %>
-          <% if total_questions(subsection, case_log, form).include?(question_title) %>
-            <%= render partial: 'form/check_answers_table', locals: { question_title: question_title, question_info: question_info, case_log: case_log, page: page } %>
+          <% if total_questions(subsection, @case_log, form).include?(question_title) %>
+            <%= render partial: 'form/check_answers_table', locals: { question_title: question_title, question_info: question_info, case_log: @case_log, page: page } %>
           <%end %>
         <%end %>
       <% end %>
```
kosiakkatrina, line 12: nice!

app/helpers/tasklist_helper.rb
baarkerlounger: Yeah in terms of what they're doing in the code here they're the same. I just thought `completed` is a clearer name than `submitted` because you can "submit" an incomplete log, so I've just renamed this everywhere.

kosiakkatrina: Yup, makes sense

kosiakkatrina (approved):
I'm still trying to make the API requests locally, but the code looks sound, no reason not to go in

dushan-madetech (approved):
Looks good to me, 🚢 it

PR #49, opened 2021-10-15
CLDC-358 conditional page routing

dushan-madetech:
Add logic for conditional page routing.

Add `get_next_page_path` method in `case_log_controller` which considers `conditional_route_to` question field when selecting the next relevant page.

Change `next_page` form model method to  consider `default_next_page` page field when selecting the next page.

Change `total_questions` method in `check_answers_helper` to step through the form from the user's perspective and accumulates relevant questions that have been shown to the user.

Change back button to work the same way as the browser navigation.


app/controllers/case_logs_controller.rb
```diff
@@ -96,4 +96,18 @@ def create_params
 
     params.require(:case_log).permit(CaseLog.editable_fields)
   end
+
+  def get_next_page_path(form, previous_page, responses_for_page={})
```
baarkerlounger: I think I'd keep this being a `form` method since that's where we mostly encapsulate form flow logic at the moment?

app/views/form/page.html.erb
```diff
@@ -1,8 +1,5 @@
-<% previous_page = form.previous_page(page_key) %>
 <% content_for :before_content do %>
-  <%= govuk_back_link href: "/case_logs/#{@case_log.id}/#{previous_page}" do %>
-    Back
-  <% end %>
+  <%= link_to 'Back', 'javascript:history.back()', class: "govuk-back-link" %>
```
baarkerlounger, line 2: @roobottom if this button is just going to replicate the browser back button does it actually make any sense to have it at all?

roobottom: @baarkerlounger - yeah, good question. The [guidance on the back link](https://design-system.service.gov.uk/components/back-link/) suggests it should act a lot like the browser back. It also says you _can_ use javascript to do so - but should disable it if js is not enabled on the user's device. It would be better, if possible, to do this server side?

app/controllers/case_logs_controller.rb
kosiakkatrina: I think we considered that, but decided on putting it here as it deals with responses and the model should probably only know about itself

app/helpers/check_answers_helper.rb
```diff
@@ -10,16 +10,46 @@ def total_number_of_questions(subsection, case_log, form)
...
-          questions_not_applicable << conditional_question_key
+    total_questions = {}
+    page_name = form.pages_for_subsection(subsection).keys.first
+
+    while page_name != "check_answers" && page_name != :check_answers
+      questions = form.questions_for_page(page_name)
+      question_key = questions.keys[0]
+      question_value = questions.values[0]
+
+      appliccable_questions = filter_conditional_questions(questions, case_log)
```
baarkerlounger: Spelling on `applicable`

app/helpers/check_answers_helper.rb
```diff
@@ -10,16 +10,46 @@ def total_number_of_questions(subsection, case_log, form)
...
 
   def total_questions(subsection, case_log, form)
-    questions = form.questions_for_subsection(subsection)
-    questions_not_applicable = []
-    questions.reject do |question_key, question|
-      question.fetch("conditional_for", []).map do |conditional_question_key, condition|
-        if condition_not_met(case_log, question_key, question, condition)
-          questions_not_applicable << conditional_question_key
+    total_questions = {}
+    page_name = form.pages_for_subsection(subsection).keys.first
```
baarkerlounger, line 14: Could do a `to_s` (or `to_sym`) here and avoid needing to check two conditions on the line below?

baarkerlounger:
Am I right in thinking there's basically two possible approaches here:

1) Have the question define what next page each of it's answers route to
2) Have each question define a list of conditions that must be true for it to be shown

This PR takes approach 1?

I suspect approach 2 might be more scalable if we get situations where showing a page depends on the answer to multiple questions on different pages (you can just list all those conditions in one place, and evaluate them together), but maybe that's not needed here yet?

app/views/form/_check_answers_table.html.erb
```diff
@@ -1,7 +1,7 @@
 <dl class="govuk-summary-list govuk-!-margin-bottom-9">
   <div class="govuk-summary-list__row">
     <dt class="govuk-summary-list__key">
-      <%= question_info["check_answer_label"].to_s %>
+      <%= question_info["check_answer_label"].to_s != "" ?  question_info["check_answer_label"].to_s : question_info["header"].to_s%>
```
baarkerlounger:
```suggestion
      <%= question_info["check_answer_label"].to_s.present? ?  question_info["check_answer_label"].to_s : question_info["header"].to_s%>
```

app/controllers/case_logs_controller.rb
```diff
@@ -130,14 +130,14 @@ def api_case_log_params
...
   def get_next_page_path(form, previous_page, responses_for_page = {})
-    # binding.pry
     questions_for_page = form.questions_for_page(previous_page)
     questions_for_page.each do |question, content|
       if content.key?("conditional_route_to")
         content["conditional_route_to"].each do |route, answer|
-          # binding.pry
-          if answer.include?(responses_for_page[question])
-            return "case_log_#{route}_path"
+          if !responses_for_page[question].nil?
```
baarkerlounger: `.present?` Empty strings probably aren't valid responses?

app/helpers/check_answers_helper.rb
kosiakkatrina: good shout!

app/controllers/case_logs_controller.rb
```diff
@@ -128,4 +128,19 @@ def api_case_log_params
 
     params.require(:case_log).permit(CaseLog.editable_fields)
   end
+
+  def get_next_page_path(form, previous_page, responses_for_page = {})
+    questions_for_page = form.questions_for_page(previous_page)
+    questions_for_page.each do |question, content|
+      next unless content.key?("conditional_route_to")
```
baarkerlounger, line 135: Hypothetically if multiple questions on the same page have a conditional_route_to, am I right in thinking this will just return the result for the 1st one?

That's probably fine but it might be more explicit to raise an error if that's been defined?

baarkerlounger (approved):

app/controllers/case_logs_controller.rb
kosiakkatrina: Yes. It'll look only at the 1st one. Will look at this as a separate PR

PR #50, opened 2021-10-15
Clcd 495/times prev offered

MatthewPhelanMadeTech:
https://digital.dclg.gov.uk/jira/browse/CLDC-495


app/models/case_log.rb
```diff
@@ -1,14 +1,20 @@
...
+  # followed by field name this is how the metaprogramming of the method
+  # name being call in the validate method works.
 
   def validate_tenant_age(record)
     if record.tenant_age && !/^[1-9][0-9]?$|^120$/.match?(record.tenant_age.to_s)
       record.errors.add :tenant_age, "must be between 0 and 120"
     end
   end
 
+  def validate_property_number_of_times_relet(record)
```
baarkerlounger, line 12: Could we add tests to `model/case_log_spec.rb` as well that make it explicit what all the anticipated cases are similar to the age validations?

baarkerlounger: That way we can use that for validation unit tests, and the request spec can stay a test of the general functionality

MatthewPhelanMadeTech: All set 👍

baarkerlounger (approved):

PR #51, opened 2021-10-18
WIP - CLDC-541: API request bulk actions

baarkerlounger:
- [ ] Add background job queuing/processing
- [ ] Create job definition for create/update Case Log
- [ ] Update controller actions
- [ ] Decide how to communicate status for async actions (Callback? GET endpoint?)

Additional thoughts to consider:

- API interface: Do we have a single approach here (like always passing an array of params, and always expecting async result, or do we make an exception for passing a single case and processing that synchronously on the assumption that it's likely to be the most common use case?)
- Do we even have a real user need for bulk actions? Do data providers ever enter data into their own software systems in bulk or is it always one at a time? And if, always one at a time is there any real reason not to send individual API requests as that's done? Are bulk updates a thing?

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -25,6 +25,8 @@ gem "govuk_design_system_formbuilder"
 gem "hotwire-rails"
 # Soft delete ActiveRecords objects
 gem "discard"
+# Multi-threaded, Postgres-based, ActiveJob backend for Ruby on Rails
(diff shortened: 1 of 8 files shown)


dushan-madetech:
Looks good so far, let me know when you're after a full review/approval 👍

PR #52, opened 2021-10-19
Why reasonable preference? validation added

MatthewPhelanMadeTech:


app/javascript/controllers/conditional_question_controller.js
```diff
@@ -28,7 +28,10 @@ export default class extends Controller {
           div.style.display = "block"
         } else {
           div.style.display = "none"
-          let buttons = document.getElementsByName(`case_log[${targetQuestion}]`)
+          let buttons = document.getElementsByName(`case_log[${targetQuestion}]`);
+          if (buttons.length == 0){
```
baarkerlounger, line 32: What case does this happen in?

app/models/case_log.rb
```diff
@@ -15,13 +15,33 @@ def validate_property_number_of_times_relet(record)
     end
   end
 
+  def validate_api_reasonable_preference(record)
+    if record.reasonable_preference == "No"
+      if record.reasonable_preference_reason_homeless || record.reasonable_preference_reason_unsatisfactory_housing || record.reasonable_preference_reason_medical_grounds || record.reasonable_preference_reason_avoid_hardship || record.reasonable_preference_reason_do_not_know
+        record.errors.add :reasonable_preference_reason, "- no resasons can be set to true, if reasonable preference is No"
```
baarkerlounger: Typo on "reasons". Might be clearer to say:

"if reasonable preference is no, no reasons should be given"?

app/models/case_log.rb
```diff
@@ -15,13 +15,33 @@ def validate_property_number_of_times_relet(record)
     end
   end
 
+  def validate_api_reasonable_preference(record)
```
baarkerlounger: I suppose technically someone could modify their browser request/turn off the javascript that shows/hides this so probably worth including it in the method below as well?

app/javascript/controllers/conditional_question_controller.js
MatthewPhelanMadeTech: If the hidden field is a checkbox the name gets an extra "[]" on the end and I couldn't work out where it came from

app/models/case_log.rb
MatthewPhelanMadeTech: good catch!

app/models/case_log.rb
MatthewPhelanMadeTech: Sorry you've lost me, what need to be added to the method?

app/javascript/controllers/conditional_question_controller.js
baarkerlounger: Ah I see, so we probably need case handling for all the different target types (in the long run). Might be worth adding a comment or a TODO there that that'll currently only clear the answer for radio and checkbox, and probably won't work yet for other fields?

app/models/case_log.rb
baarkerlounger: As in combine these validations into a single `validate_reasonable_preference` method since they probably do all need to run each time?

app/javascript/controllers/conditional_question_controller.js
MatthewPhelanMadeTech: it seems to be fine on text, numeric and radio just checkboxes

baarkerlounger: Interesting. Well :+1: then.

app/models/case_log.rb
MatthewPhelanMadeTech: No they won't the UI and the stimulus controller handles it normally and we shouldn't error - the api one only need to be run when not submitted through the UI

baarkerlounger: I'm thinking if someone had js turned off, they'd _potentially_ be able to bypass that (ideally you always validate server side even if you have front end validations). Also bulk upload this will likely apply to too?

MatthewPhelanMadeTech: Ahh I see what you mean, I'll have a mess around with it.

spec/fixtures/complete_case_log.json
```diff
@@ -38,7 +38,7 @@
       "person_8_age": 2,
       "person_8_gender": "Prefer not to say",
       "person_8_economic_status": "Child under 16",
-      "homelessness": "No",
+      "homelessness": "Yes - other homelessness",
```
baarkerlounger, line 41: How come this fixed the spec? 🤔

MatthewPhelanMadeTech: The data in the complete log file was not valid - you don't get reasonable preference without previous homelessness so the response body just included the errors and the spec could not find the status making it fail. With valid data it works fine.

app/models/case_log.rb
```diff
@@ -15,13 +15,29 @@ def validate_property_number_of_times_relet(record)
     end
   end
 
+  def validate_reasonable_preference(record)
+    if record.homelessness == "No" && record.reasonable_preference == "Yes"
+      record.errors.add :reasonable_preference, "can not be Yes if Not Homesless imediately prior to this letting has been selected"
```
baarkerlounger: Typo on immediately

spec/fixtures/complete_case_log.json
baarkerlounger: Ahhh I see it now. Thanks.

app/models/case_log.rb
baarkerlounger: Also on homeless

MatthewPhelanMadeTech: Is there a Grammarly extension for vs code? 😂 I'll fix this now

app/models/case_log.rb
```diff
@@ -15,13 +15,29 @@ def validate_property_number_of_times_relet(record)
...
+    elsif record.reasonable_preference == "Yes"
+      if !record.reasonable_preference_reason_homeless && !record.reasonable_preference_reason_unsatisfactory_housing && !record.reasonable_preference_reason_medical_grounds && !record.reasonable_preference_reason_avoid_hardship && !record.reasonable_preference_reason_do_not_know
+        record.errors.add :reasonable_preference_reason, "- if reasonable preference is Yes, a reason must be given"
+      end
+    elsif record.reasonable_preference == "No"
+      if record.reasonable_preference_reason_homeless || record.reasonable_preference_reason_unsatisfactory_housing || record.reasonable_preference_reason_medical_grounds || record.reasonable_preference_reason_avoid_hardship || record.reasonable_preference_reason_do_not_know
+        record.errors.add :reasonable_preference_reason, "- if reasonable preference is no, no reasons should be given"
```
baarkerlounger: Can we make this "is No" so the answer option capitalization is consistent with the message above?

baarkerlounger (approved):

PR #54, opened 2021-10-19
add other reason left settled home validation

dushan-madetech:
Added validation for the other reason for leaving settled home field along with unit tests. I was experimenting with trying to refactor the validation unit tests into a single unit test which loops through the fields and the values to test those fields with but was unable to get it working so discarded the attempt.


dushan-madetech:
Rubocop is flagging an issue but it appears to have come from main and I wasn't sure of the best way to deal with it

spec/models/case_log_spec.rb
```diff
@@ -25,6 +25,22 @@
     it "validates number of relets is over 0" do
       expect { CaseLog.create!(property_number_of_times_relet: 0) }.to raise_error(ActiveRecord::RecordInvalid)
     end
+
+    context "other reason for leaving last settled home validation" do
+      it "must be provided if main reason for leaving last settled home was given as other" do
+        expect {
+          CaseLog.create!(reason_for_leaving_last_settled_home: "Other",
+                          other_reason_for_leaving_last_settled_home: nil)
+        }.to raise_error(ActiveRecord::RecordInvalid)
```
dushan-madetech, line 34: I would like to find a way to refactor the tests since they're all following the same cycle of creating a Case Log with invalid fields and the expectation of all the tests is the same.

baarkerlounger (approved):

PR #55, opened 2021-10-20
CLDC-474 receive income frequency validation

dushan-madetech:
This didn't need any validation since the question just needs to be made optional if the net income is refused (given as 0) and no errors need to be displayed


app/models/case_log.rb
```diff
@@ -105,6 +105,10 @@ def mandatory_fields
       dynamically_not_required << "other_reason_for_leaving_last_settled_home"
     end
 
+    if net_income.to_i.zero?
```
baarkerlounger, line 108: I wonder if we should differentiate between zero and refused (i.e. have an actual refused option for text/numeric input fields) since this way it's impossible to tell whether they genuinely have 0 income or wanted to refuse. @roobottom what do you think?

baarkerlounger (approved):

app/models/case_log.rb
dushan-madetech: If we want to have a refused option I'd be quite happy to create a ticket to have that option for all of the relevant questions and to edit the validations for those questions as appropriate 😄

PR #57, opened 2021-10-22
Full error messages

baarkerlounger:
**UI**

image

**API response:**
```
{
  "errors": {
    "reasonable_preference_reason": [
      "If reasonable preference is Yes, a reason must be given"
    ],
    "tenant_age": [
      "Tenant age must be between 0 and 120"
    ]
  }
}
```


MatthewPhelanMadeTech (commented):
This all looks good!

PR #59, opened 2021-10-22
Cldc 477/reason for leaving because of the benefit cap

MatthewPhelanMadeTech:


db/schema.rb
```diff
@@ -131,6 +131,7 @@
     t.boolean "reasonable_preference_reason_medical_grounds"
     t.boolean "reasonable_preference_reason_avoid_hardship"
     t.boolean "reasonable_preference_reason_do_not_know"
+    t.integer "property_number_of_times_relet"
```
baarkerlounger: Probably shouldn't be committing schema changes without a migration?

app/models/case_log.rb
```diff
@@ -38,6 +38,12 @@ def validate_other_reason_for_leaving_last_settled_home(record)
       record.errors.add :other_reason_for_leaving_last_settled_home, "The other reason must not be provided if the reason for leaving settled home was not other"
     end
   end
+  
+  def validate_reason_for_leaving_last_settled_home(record)
```
baarkerlounger, line 42: Can we add a spec for this?

MatthewPhelanMadeTech: We absolutely can, knew I was forgetting something

db/schema.rb
MatthewPhelanMadeTech: very good point - I assume this is why the tests failed - will fix now

baarkerlounger (approved):

PR #62, opened 2021-10-25
CLDC-473: Add net income uc proportion validation

kosiakkatrina:
Must be completed
Cannot be All if person 1 works part or full time
Cannot be All if any of persons 2-8 are person 1's partner and work part or full time


app/models/case_log.rb
```diff
@@ -65,6 +74,14 @@ def validate(record)
       validation_methods.each { |meth| public_send(meth, record) }
     end
   end
+
+private
```
baarkerlounger: :+1:

app/models/case_log.rb
```diff
@@ -49,6 +49,15 @@ def validate_armed_forces_injured(record)
     end
   end
 
+  def validate_net_income_uc_proportion(record)
+    if ["Full-time - 30 hours or more", "Part-time - Less than 30 hours"].include?(record.tenant_economic_status) && record.net_income_uc_proportion == "All"
+      record.errors.add :net_income_uc_proportion, "income is from Universal Credit, state pensions or benefits cannot be All if person works part or full time"
+    end
+    check_partner_net_income_uc_proportion(record.person_2_economic_status, record.person_2_relationship, record)
```
baarkerlounger: Do we not need to check all the `person_n_` fields? If so maybe `.any?` or `.all?` block? Could also potentially genericize it to include the above condition.

kosiakkatrina: Ah, `any` could be a good shout, from the specification in jira we only need to check person 2-4

baarkerlounger: Hmm that's weird. I suspect there's an implicit assumption there that any person listed below 4 must be a child then and not a partner but as we're not enforcing that in any way it seems shaky... maybe confirm with Rachel?

app/models/case_log.rb
```diff
@@ -49,6 +49,20 @@ def validate_armed_forces_injured(record)
...
+      record.errors.add :net_income_uc_proportion, "income is from Universal Credit, state pensions or benefits cannot be All if person works part or full time"
+    end
+
+    fields = [{ status: record.person_2_economic_status, relationship: record.person_2_relationship },
+              { status: record.person_3_economic_status, relationship: record.person_3_relationship },
+              { status: record.person_4_economic_status, relationship: record.person_4_relationship }]
+
+    if (fields.any? { |field| ["Full-time - 30 hours or more", "Part-time - Less than 30 hours"].include?(field[:status]) && field[:relationship] == "Partner" }) && record.net_income_uc_proportion == "All"
+      record.errors.add :net_income_uc_proportion, "income is from Universal Credit, state pensions or benefits cannot be All if the partner works part or full time"
+    end
```
baarkerlounger: What about something like:

```
  def validate_net_income_uc_proportion(record)
    if ["Full-time - 30 hours or more", "Part-time - Less than 30 hours"].include?(record.tenant_economic_status) && record.net_income_uc_proportion == "All"
      record.errors.add :net_income_uc_proportion, "income is from Universal Credit, state pensions or benefits cannot be All if person works part or full time"
    end
    (2..4).any? { |n| check_partner_net_income_uc_proportion(n, record) }
  end

private

  EMPLOYED_STATUSES = ["Full-time - 30 hours or more", "Part-time - Less than 30 hours"].freeze

  def check_partner_net_income_uc_proportion(person_num, record)
    economic_status = record["person_#{person_num}_economic_status"]
    relationship = record["person_#{person_num}_relationship"]
    if EMPLOYED_STATUSES.include?(economic_status) && relationship == "Partner" && record.net_income_uc_proportion == "All"
      record.errors.add :net_income_uc_proportion, "income is from Universal Credit, state pensions or benefits cannot be All if the partner works part or full time"
    end
  end
```

Could potentially refactor that further by changing the main tenants fields to `record.person_1_economic_status` etc and then:

```
EMPLOYED_STATUSES = ["Full-time - 30 hours or more", "Part-time - Less than 30 hours"].freeze

  def validate_net_income_uc_proportion(record)
    (1..4).any? do |n|
      economic_status = record["person_#{n}_economic_status"]
      employed? = EMPLOYED_STATUSES.include?(economic_status)
      relationship = record["person_#{n}_relationship"]
      partner_or_main? = relationship == "Partner" || (relationship.nil && economic_status.present?)
      if employed? && partner_or_main? && record.net_income_uc_proportion == "All"
        record.errors.add :net_income_uc_proportion, "income is from Universal Credit, state pensions or benefits cannot be All if the partner works part or full time"
      end
    end
  end
```

This might not be any better than what you have so feel free to ignore... :grin:

magicmilo:
In CLDC-471 pregnancy validation I changed tenant_age/gender to person_1_age/gender. Should be able to any? now I've done a person loop like Dan showed in women_of_child_bearing_age_in_household()

app/models/case_log.rb
kosiakkatrina: Refactored this with your first suggestion, not sure if changing the field names would be beneficial in terms of the rest of the tool? 🤔

kosiakkatrina:
> In CLDC-471 pregnancy validation I changed tenant_age/gender to person_1_age/gender. Should be able to any? now I've done a person loop like Dan showed in women_of_child_bearing_age_in_household()

Ahh, well maybe it makes sense to do it too

baarkerlounger (approved):

baarkerlounger:
We will also need to update the API docs to reflect the field change

kosiakkatrina:
> We will also need to update the API docs to reflect the field change

I think that's done

PR #63, opened 2021-10-25
Cldc 471 pregnancy validation

magicmilo:
Not a fan of the while loop for checking all tenants


app/models/case_log.rb
```diff
@@ -49,6 +49,30 @@ def validate_armed_forces_injured(record)
...
 
+  def validate_household_pregnancy(record)
+    if (record.pregnancy == "Yes" || record.pregnancy == "Prefer not to say") && !women_of_child_bearing_age_in_household(record)
+      record.errors.add :pregnancy, "You must answer no as there are no female tenants aged 16-50 in the property"
+    end
+  end
+
+  def women_of_child_bearing_age_in_household(record)
+    unless record.tenant_gender.nil? || record.tenant_age.nil?
+      if record.tenant_gender == "Female" && record.tenant_age >= 16 && record.tenant_age <= 50
```
baarkerlounger: This can just be:

```suggestion
      record.tenant_gender == "Female" && record.tenant_age >= 16 && record.tenant_age <= 50
```

Don't need the `if` or the `return`

app/models/case_log.rb
```diff
@@ -49,6 +49,30 @@ def validate_armed_forces_injured(record)
...
+        return true
+      end
+    end
+
+    p = 2
```
baarkerlounger: Would be more idiomatic in Ruby to do something like:

```suggestion
(2..8).any? do |n|
  next if record["person_#{n}_gender"].nil? || record["person_#{n}_age"].nil?
  
  record["person_#{n}_gender"] == "Female" && record["person_#{n}_age"] >= 16 && record["person_#{n}_age"] <= 50
end
```

baarkerlounger: Could also refactor the main tenant field names to follow the same pattern (i.e. `tenant_gender` becomes `person_1_gender` or something like that) and then it can all be part of the same loop rather than being a special case

app/models/case_log.rb
```diff
@@ -49,6 +49,30 @@ def validate_armed_forces_injured(record)
     end
   end
 
+  def validate_household_pregnancy(record)
+    if (record.pregnancy == "Yes" || record.pregnancy == "Prefer not to say") && !women_of_child_bearing_age_in_household(record)
+      record.errors.add :pregnancy, "You must answer no as there are no female tenants aged 16-50 in the property"
+    end
+  end
+
+  def women_of_child_bearing_age_in_household(record)
```
baarkerlounger: Think we should probably make any methods that aren't direct validations (i.e. `validate_#{question}`) private so the assumption above about which methods to run holds.

app/models/case_log.rb
baarkerlounger: Think this might be quite worth it because the same pattern came up on https://github.com/communitiesuk/mhclg-data-collection-beta/pull/62 and probably will elsewhere too

baarkerlounger:
Think this just needs API docs updating (for the new field names) and rubocop linting and then it looks good to go

magicmilo:
> Think this just needs API docs updating (for the new field names) and rubocop linting and then it looks good to go

All done, I wish Stoplight Studio had been around a few years ago or if I'd known of it.

baarkerlounger (approved):

PR #64, opened 2021-10-26
added armed forces active response validation

magicmilo:


docs/api/DLUHC-CORE-Data.v1.json
```diff
@@ -1086,8 +1086,8 @@
         },
         "required": [
           "tenant_code",
-          "tenant_age",
```
baarkerlounger: Think we also need to add allowed values for the armed forces questions here?

magicmilo: Missed that, replaced now in three places

PR #65, opened 2021-10-27
added simple validation for outstanding rent

magicmilo:


app/models/case_log.rb
```diff
@@ -55,6 +55,15 @@ def validate_armed_forces_injured(record)
     end
   end
 
+  def validate_outstanding_rent_amount(record)
+    if record.outstanding_rent_or_charges == "Yes" && record.outstanding_amount.blank?
+      record.errors.add :outstanding_amount, "You must answer the oustanding amout question if you have outstanding rent or charges."
+    end
+    if record.outstanding_rent_or_charges == "No" && !record.outstanding_amount.blank?
```
baarkerlounger: `.present?` is the opposite of `.blank?` which I think rubocop will want here

magicmilo: Oops dam cops. Done

baarkerlounger:
Got some merge conflicts that need resolving and need to update API docs but otherwise looks good to me

docs/api/DLUHC-CORE-Data.v1.json
```diff
@@ -1207,6 +1207,17 @@
         ]
       }
     },
-    "securitySchemes": {}
+    "securitySchemes": {
```
baarkerlounger, line 1210: Was this intended to be committed or should it have been updating the outstanding amount fields?

PR #66, opened 2021-10-27
CLDC-466: Add fixed term tenancy validation

kosiakkatrina:
Add fixed term tenancy validation:
- Must not be completed if "Type of main tenancy" is not responded with either Secure or Assured shorthold 
- Must be completed and between 2 and 99 if "Type of main tenancy" is Assured shorthold
- Must be empty or between 2 and 99 if "Type of main tenancy" is Secure

### DIFF
--- a/app/models/case_log.rb
+++ b/app/models/case_log.rb
@@ -186,6 +186,10 @@ private
       dynamically_not_required << "net_income_frequency"
     end
 
+    if tenancy_type == "Fixed term – Secure"
+      dynamically_not_required << "fixed_term_tenancy"
+    end
+
     required.delete_if { |key, _value| dynamically_not_required.include?(key) }
   end
 end
--- a/app/models/case_log.rb
+++ b/app/models/case_log.rb
@@ -75,17 +75,17 @@ class CaseLogValidator < ActiveModel::Validator
   end
 
   def validate_fixed_term_tenancy(record)
-    if !(record.tenancy_type == "Fixed term – Secure" || record.tenancy_type == "Fixed term – Assured Shorthold Tenancy (AST)") && record.fixed_term_tenancy.present?
-      record.errors.add :fixed_term_tenancy, "You must only answer the fixed term tenancy length question if the tenancy type is fixed term"
-    end
+    is_present = record.fixed_term_tenancy.present?
+    is_in_range = record.fixed_term_tenancy.to_i.between?(2, 99)
+    is_secure = record.tenancy_type == "Fixed term – Secure"
+    is_ast = record.tenancy_type == "Fixed term – Assured Shorthold Tenancy (AST)"
+    conditions = [
+      { condition: !(is_secure || is_ast) && is_present, error: "You must only answer the fixed term tenancy length question if the tenancy type is fixed term" },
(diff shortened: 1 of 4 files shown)


app/models/case_log.rb
```diff
@@ -74,6 +74,20 @@ def validate_household_pregnancy(record)
     end
   end
 
+  def validate_fixed_term_tenancy(record)
+    is_present = record.fixed_term_tenancy.present?
```
baarkerlounger, line 78: I don't think we need this check as the general pattern is that all presence checks happen as part of setting the form to `complete` or `in_progress` rather than when submitting individual questions

app/models/case_log.rb
```diff
@@ -74,6 +74,20 @@ def validate_household_pregnancy(record)
...
+    is_in_range = record.fixed_term_tenancy.to_i.between?(2, 99)
+    is_secure = record.tenancy_type == "Fixed term – Secure"
+    is_ast = record.tenancy_type == "Fixed term – Assured Shorthold Tenancy (AST)"
+    conditions = [
+      { condition: !(is_secure || is_ast) && is_present, error: "You must only answer the fixed term tenancy length question if the tenancy type is fixed term" },
+      { condition: is_ast && !is_in_range,  error: "Fixed term – Assured Shorthold Tenancy (AST) should be between 2 and 99 years" },
+      { condition: is_secure && (!is_in_range && is_present), error: "Fixed term – Secure should be between 2 and 99 years or not specified" },
+    ]
+
+    conditions.each { |condition| condition[:condition] ? (record.errors.add :fixed_term_tenancy, condition[:error]) : nil }
```
baarkerlounger, line 88: This is a nice pattern

app/models/case_log.rb
kosiakkatrina: Oh, well, there's this condition that doesn't require it to be filled `Must be empty or between 2 and 99 if "Type of main tenancy" is Secure`, how do I disable this check on the db? 🤔

baarkerlounger: Currently the way to do it would be to add the field to the dynamically not required list here: https://github.com/communitiesuk/mhclg-data-collection-beta/blob/main/app/models/case_log.rb#L165 based on that condition

spec/fixtures/complete_case_log.json
```diff
@@ -51,7 +51,7 @@
       "tenancy_code": "BZ757",
       "tenancy_start_date": "12/03/2019",
       "starter_tenancy": "No",
-      "fixed_term_tenancy": "No",
+      "fixed_term_tenancy": "5",
```
baarkerlounger, line 54: Also need to update the API docs to show the allowed values

kosiakkatrina: Ah, yes!

app/models/case_log.rb
kosiakkatrina: I see! You mentioned this before, thanks!
We'd still need this check here tho, cause it'd throw error otherwise

baarkerlounger: Ah yes, you're right of course

baarkerlounger (approved):

PR #67, opened 2021-10-27
CLDC-568: Database Cleaner

baarkerlounger:
https://digital.dclg.gov.uk/jira/browse/CLDC-568

Attempt to solve the intermittent test failures, which appear to be database related using https://github.com/DatabaseCleaner/database_cleaner


kosiakkatrina (approved):
Hmm, seems like it's a common issue if there's a gem for it 🤔 
Nice one!

PR #70, opened 2021-10-28
Cldc 497 shared accomodation bedrooms

magicmilo:


app/models/case_log.rb
```diff
@@ -143,6 +163,15 @@ def women_of_child_bearing_age_in_household(record)
       record["person_#{n}_gender"] == "Female" && record["person_#{n}_age"] >= 16 && record["person_#{n}_age"] <= 50
     end
   end
+
+  def people_in_household(record)
```
baarkerlounger: Why would we need to derive this? Can't we just check `record.household_number_of_other_members`?

magicmilo: Good spot thanks, switched it. Don't think the test should be dependent on this field being filled.

app/models/case_log.rb
```diff
@@ -1,10 +1,136 @@
 class CaseLogValidator < ActiveModel::Validator
+  # Methods to be used on save and continue need to be named 'validate_'
+  # followed by field name this is how the metaprogramming of the method
+  # name being call in the validate method works.
+
+  def validate_person_1_age(record)
```
baarkerlounger: This looks like a merge conflict. Think you want to check out this file from `main` and then put all your new validation methods in `/app/validators/household_validations.rb`

app/models/case_log.rb
```diff
@@ -1,10 +1,136 @@
...
 
+  def validate_shared_housing_rooms(record)
+    unless record.property_unit_type.nil?
+      if record.property_unit_type == "Bed-sit" && record.property_number_of_bedrooms != 1
+        record.errors.add :property_unit_type, "A bedsit can only have one bedroom"
+      end
+
+      unless record.household_number_of_other_members.nil?
+        if record.household_number_of_other_members > 0
+          if record.property_unit_type.include?("Shared") && (record.property_number_of_bedrooms.to_i == 0 || record.property_number_of_bedrooms.to_i > 7)
```
baarkerlounger: How about `record.property_number_of_bedrooms.between?(0, 7)`

magicmilo: went with !record.property_number_of_bedrooms.to_i.between?(1, 7)

app/models/case_log.rb
```diff
@@ -1,4 +1,53 @@
 class CaseLogValidator < ActiveModel::Validator
+  # Methods to be used on save and continue need to be named 'validate_'
+  # followed by field name this is how the metaprogramming of the method
+  # name being call in the validate method works.
+  def validate_property_number_of_times_relet(record)
```
baarkerlounger: This looks like it's still coming from merge? Don't think this file needs to change at all?

baarkerlounger (approved):

PR #71, opened 2021-10-28
CLDC-513: Refused income validation

kosiakkatrina:
only require the income if it is known
the validation in the ticket is only for the `net_income`, but also added `net_income_frequency` to the `conditional_for` as it wouldn't make sense without `net_income` value present

### DIFF
--- a/config/forms/2021_2022.json
+++ b/config/forms/2021_2022.json
@@ -1458,6 +1458,21 @@
               "header": "",
               "description": "",
               "questions": {
+                "net_income_known": {
+                  "check_answer_label": "Income known",
+                  "header": "Do you know the tenant and their partner's net income?",
+                  "hint_text": "",
+                  "type": "radio",
+                  "answer_options": {
+                    "0": "Yes",
+                    "1": "No",
+                    "2": "Tenant prefers no to say"
+                  },
+                  "conditional_for": {
+                    "net_income": ["Yes"],
+                    "net_income_frequency": ["Yes"]
+                  }
+                },
                 "net_income": {
                   "check_answer_label": "Income",
                   "header": "What is the tenant’s /and partner’s combined income after tax?",
--- a/config/forms/2021_2022.json
+++ b/config/forms/2021_2022.json
@@ -1466,7 +1466,7 @@
                   "answer_options": {
                     "0": "Yes",
                     "1": "No",
-                    "2": "Tenant prefers no to say"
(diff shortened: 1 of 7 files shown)


spec/fixtures/complete_case_log.json
```diff
@@ -68,6 +68,7 @@
       "property_major_repairs_date": "05/05/2020",
       "property_number_of_times_relet": 2,
       "property_wheelchair_accessible": true,
+      "net_income_known": "Yes",
       "net_income": 0,
       "net_income_frequency": null,
```
baarkerlounger, line 73: Shouldn't this fail the spec since `net_income_known` is `Yes` so this should be required?

kosiakkatrina: 🤔 That's a very good point, let me check it out

kosiakkatrina: Ah, it's cause net_income is 0, and that makes net_income_frequency dynamically not required

baarkerlounger: Ah I think that's left over from when 0 was refused. Now that we're adding a pre-question I guess we should consider 0 to be a real answer.

config/forms/2021_2022.json
```diff
@@ -1458,6 +1458,21 @@
...
               "questions": {
+                "net_income_known": {
+                  "check_answer_label": "Income known",
+                  "header": "Do you know the tenant and their partner's net income?",
+                  "hint_text": "",
+                  "type": "radio",
+                  "answer_options": {
+                    "0": "Yes",
+                    "1": "No",
+                    "2": "Tenant prefers no to say"
```
baarkerlounger: typo on `not`

baarkerlounger: Think this is another example of where we need https://digital.dclg.gov.uk/jira/browse/CLDC-591

`Tenant prefers not to say` should probably be equivalent to `Prefer not to say` at the data storage level. (Outside the scope of this PR though).

spec/fixtures/complete_case_log.json
kosiakkatrina: I mean 0 is a real answer, but it doesn't make sense to ask for frequency if it's 0 right?

baarkerlounger: That's also a good point!

baarkerlounger (approved):

PR #72, opened 2021-10-28
CLDC-510: Income range soft validations

baarkerlounger:
This PR changes the http caching behaviour of the form page view so that soft error conditions are re-evaluated when you use back navigation.

Beyond the scope of this PR for now is considering what would happen if we had multiple soft validations on a single page


app/views/form/page.html.erb
```diff
@@ -17,7 +19,11 @@
       </div>
     <% end %>
 
-    <%= f.hidden_field :previous_page, value: page_key %>
+    <% if @case_log.soft_errors.present? && @case_log.soft_errors.keys.first == page_info["soft_validations"]&.keys&.first %>
```
baarkerlounger, line 22: I wonder if doing an ajax request here for whether or not this should show would avoid the need to hijack the back button... would probably also eliminate the override confirm question momentarily showing before being replaced.

baarkerlounger: On the other hand, if you've come straight from a bookmark or just typed the URL for a specific question into your browser - what should the back link do there? Still default browser back?

app/controllers/case_logs_controller.rb
```diff
@@ -56,16 +56,17 @@ def edit
...
-    previous_page = params[:case_log][:previous_page]
-    questions_for_page = form.questions_for_page(previous_page)
-    responses_for_page = question_responses(questions_for_page)
-    @case_log.previous_page = previous_page
-    if @case_log.update(responses_for_page)
-      redirect_path = get_next_page_path(form, previous_page, @case_log)
+    page = params[:case_log][:page]
+    @case_log.current_page = page
+    responses_for_page = responses_for_page(page)
+    if @case_log.update(responses_for_page) && (@case_log.soft_errors.empty? || @case_log.soft_errors_overridden?)
```
dushan-madetech: I feel like this line is quite complex, is there some way we can extract or shorten the conditions?

app/controllers/case_logs_controller.rb
```diff
@@ -56,16 +56,17 @@ def edit
...
+      @case_log.previous_page = page
+      redirect_path = get_next_page_path(form, page, @case_log)
       redirect_to(send(redirect_path, @case_log))
     else
-      page_info = form.all_pages[previous_page]
-      render "form/page", locals: { form: form, page_key: previous_page, page_info: page_info }, status: :unprocessable_entity
+      page_info = form.all_pages[page]
+      back_path = get_previous_page_path(form, page, @case_log)
+      render "form/page", locals: { form: form, back_path: back_path, page_key: page, page_info: page_info }, status: :unprocessable_entity
```
dushan-madetech: It feels like this is increasing in cyclomatic complexity as we build out the case log functionality. I wonder if there's some way we can reduce this? Nothing springs to mind for me right now but might be worth keeping an eye on

app/controllers/case_logs_controller.rb
```diff
@@ -129,8 +138,8 @@ def api_case_log_params
     params.require(:case_log).permit(CaseLog.editable_fields)
   end
 
-  def get_next_page_path(form, previous_page, case_log = {})
-    content = form.all_pages[previous_page]
+  def get_next_page_path(form, page, case_log = {})
```
dushan-madetech, line 133: In this context is page the current page that the user would be on?

app/controllers/case_logs_controller.rb
```diff
@@ -93,17 +94,25 @@ def check_answers
...
+      render "form/page", locals: { form: form, back_path: back_path, page_key: page_key, page_info: page_info }
     end
   end
 
 private
 
   API_ACTIONS = %w[create show update destroy].freeze
 
-  def question_responses(questions_for_page)
+  def responses_for_page(page)
```
dushan-madetech, line 102: I feel like this method is getting more bloated and worth trimming down somehow if possible

app/controllers/case_logs_controller.rb
```diff
@@ -93,17 +94,25 @@ def check_answers
...
+    form = FormHandler.instance.get_form("2021_2022")
```
dushan-madetech, line 103: What if we made the form year an environment variable or something like that? That way we could define this line as `form = FormHandler.instance.get_form("FORM_YEAR)` or something like that so we only have to update it in a single place when/if the form changes in the future.

app/controllers/case_logs_controller.rb
```diff
@@ -139,6 +148,12 @@ def get_next_page_path(form, previous_page, case_log = {})
...
       end
     end
-    form.next_page_redirect_path(previous_page)
+    form.next_page_redirect_path(page)
+  end
+
+  def get_previous_page_path(form, current_page, case_log = {})
+    return case_log.previous_page if case_log.previous_page
+    return request.referer.split("/")[-2..-1].join("/") if request.referer&.ends_with?("check_answers")
+    return form.previous_page(current_page) if current_page
```
dushan-madetech: I like the use of guard clauses 👍

app/models/form.rb
```diff
@@ -41,6 +41,11 @@ def questions_for_subsection(subsection)
     pages_for_subsection(subsection).map { |title, _value| questions_for_page(title) }.reduce(:merge)
   end
 
+  # Returns a hash with soft validation questions as keys
+  def soft_validations_for_page(page)
```
dushan-madetech, line 45: maybe rename the method to `soft_validations_questions_for_page` and then you might be able to delete the comment above

app/validations/soft_validations.rb
```diff
@@ -0,0 +1,41 @@
...
+      )
+    elsif net_income_in_soft_max_range?
+      net_income_errors["override_net_income_validation"] = OpenStruct.new(
+        message: "Net income is higher than expected based on the main tenant's working situation. Are you sure this is correct?",
+        hint_text: "This is based on the tenant's work situation: #{person_1_economic_status}",
+      )
+    else
+      update_column(:override_net_income_validation, nil)
+    end
+    net_income_errors
```
dushan-madetech, line 31: Could a switch statement work here instead?

dushan-madetech (commented):
Left some comments for you to have a look at, let me know when you think this is ready to be looked at for merging

app/views/form/page.html.erb
dushan-madetech: It might be worth speaking to Jon about how we want the back link to behave in certain scenarios. My gut feeling is that it should behave like the browser back button. It might also be worth seeing what other gov services do for back button behaviour

app/controllers/case_logs_controller.rb
baarkerlounger: The page they've just submitted yes. I guess you're kind of in between pages at the literal point this is run?

app/controllers/case_logs_controller.rb
baarkerlounger: Do you have a suggestion?

app/controllers/case_logs_controller.rb
baarkerlounger: Happy to try that but would suggest that happens in a separate PR since this is already quite long/fiddly.

app/models/form.rb
baarkerlounger: Comment is really there for consistency with the methods above - it's about return type (hash) rather than what the thing is. Could add that to the method name I guess but possibly a bit clunky?

app/views/form/page.html.erb
baarkerlounger: Reverted hijacking the back link and made the form page not cache instead. Still sub-optimal but slightly less so...

app/controllers/case_logs_controller.rb
dushan-madetech: I feel like `(@case_log.soft_errors.empty? || @case_log.soft_errors_overridden?)` could be extracted to a method like `case_log_has_no_errors?`

app/controllers/case_logs_controller.rb
dushan-madetech: I'm also happy for that to a be a separate PR 🙂

app/models/form.rb
dushan-madetech: In that case no need to change it 🙂

dushan-madetech (approved):

PR #73, opened 2021-10-28
Refactor validations into modules

baarkerlounger:


app/models/case_log.rb
```diff
@@ -127,9 +15,7 @@ def validate(record)
         public_send("validate_#{question_to_validate}", record)
       end
     else
-      # This assumes that all methods in this class other than this one are
-      # validations to be run
-      validation_methods = public_methods(false) - [__callee__]
+      validation_methods = public_methods.select { |method| method.starts_with?("validate") } - [__callee__]
```
kosiakkatrina: If we have starts_with?("validate_"), we wouldn't need to remove the callee?

kosiakkatrina (approved):
nice one!

app/models/case_log.rb
baarkerlounger: That's a good point

PR #74, opened 2021-10-29
Cldc 467/validate other household members

MatthewPhelanMadeTech:
Could whomever picks up this code review take a pass over the ticket to make sure all conditions are included in case Dan and I have missed any. The more eyes on it the better. Any questions - let me know.

### DIFF
--- a/app/models/case_log.rb
+++ b/app/models/case_log.rb
@@ -117,6 +117,81 @@ class CaseLogValidator < ActiveModel::Validator
     validate_other_field(record, "tenancy_type", "other_tenancy_type")
   end
 
+  def validate_other_household_members(record)
+    index = 0
+    number_of_other_members = record.household_number_of_other_members
+    partner = false
+
+    while index < number_of_other_members
+      
+      member_number = index+2
+      relationship = record["person_#{member_number}_relationship"]
+      age = record["person_#{member_number}_age"]
+      gender = record["person_#{member_number}_gender"]
+      economic_status = record["person_#{member_number}_economic_status"]
+
+      binding.pry
+      if relationship || age || gender || economic_status
+        if relationship.nil? || age.nil? || gender.nil? || economic_status.nil?
+          record.errors.add "person_#{member_number}_age", "If any of the person is filled out it must all be filled"
+        end
+      end
+
+      if age<1 || age>120
+        record.errors.add "person_#{member_number}_age", "Tenant #{member_number} age must be between 1 and 120 (i.e. infants must be entered as 1)"
+      end
+
+      if age>70 && economic_status != "Retired"
+        record.errors.add "person_#{member_number}_economic_status", "Tenant #{member_number} must be retired if over 70"
+      end
+
+      if gender=="Male" && economic_status == "Retired" && age<65
+        record.errors.add "person_#{member_number}_age", "Male tenant who is retired must be 65 or over"
+      end
+
+      if gender=="Female" && economic_status == "Retired" && age<60
+        record.errors.add "person_#{member_number}_age", "Female tenant who is retired must be 60 or over"
+      end
+
+      if age>70 && economic_status != "Retired"
+        record.errors.add "person_#{member_number}_economic_status", "Tenant #{member_number} must be retired if over 70"
+      end
+
+      if age<16 
+        if relationship != "Child - includes young adult and grown-up"
+          record.errors.add "person_#{member_number}_relationship", "Tenant #{member_number}'s relationship to tenant 1 must be Child if their age is under 16"
+        end
+        if economic_status != "Child under 16"
+          record.errors.add "person_#{member_number}_economic_status", "Tenant #{member_number} economic status must be Child under 16 if their age is under 16"
+        end
+      end
+
+      if relationship == "Partner"
+        if partner
+          record.errors.add "person_#{member_number}_relationship", "Tenant can not have multiple partners"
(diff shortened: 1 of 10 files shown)


docs/api/DLUHC-CORE-Data.v1.json
```diff
@@ -110,7 +110,7 @@
                     "value": {
                       "errors": {
                         "person_1_age": [
-                          "Tenant age must be between 0 and 120"
+                          "Tenant age must be an integer between 0 and 120"
```
kosiakkatrina: person 1 age should be over 16 methinks

app/validations/household_validations.rb
```diff
@@ -65,4 +75,70 @@ def women_of_child_bearing_age_in_household(record)
       record["person_#{n}_gender"] == "Female" && record["person_#{n}_age"] >= 16 && record["person_#{n}_age"] <= 50
     end
   end
+
+  def validate_person_age(record, person_num)
+    age = record.public_send("person_#{person_num}_age")
+    return unless age
+
+    if !age.is_a?(Integer) || age < 1 || age > 120
```
kosiakkatrina, line 110: for person 2 on the ticket it's `DO NOT DO: Must be 100 or younger` I'm not sure what do not do means and if it's valid tho

app/validations/household_validations.rb
```diff
@@ -65,4 +75,70 @@ def women_of_child_bearing_age_in_household(record)
...
+      record.errors.add "person_#{person_num}_age".to_sym, "Tenant age must be an integer between 0 and 120"
+    end
+  end
+
+  def validate_person_age_matches_economic_status(record, person_num)
```
kosiakkatrina, line 115: is this going to run for person 1? or only 2-8?

kosiakkatrina (commented):
Are person 1 validations supposed to be a part of this PR?
- Person 1 age refused: Cannot be selected if P1 Age is entered (Needs Design)
- If person 1 is female and retired they must be 60 or older 
- If person 1 is male and retired they must be 65 or older
- Person 1 economic status: Must not have status 9 "child under 16"

app/validations/household_validations.rb
```diff
@@ -56,6 +51,21 @@ def validate_household_pregnancy(record)
     end
   end
 
+  def validate_household_number_of_other_members(record)
```
kosiakkatrina, line 54: the method name sounds as if we're validating that the count of other household members is correct, but we're doing much more here

app/models/case_log.rb
```diff
@@ -135,6 +134,14 @@ def mandatory_fields
       dynamically_not_required << "net_income_frequency"
     end
 
+    start_range = (household_number_of_other_members || 0) + 2
```
kosiakkatrina, line 137: why would this ever not be 2?

app/validations/household_validations.rb
baarkerlounger: Badly written ticket. Just means validate to 120 not to 100.

app/validations/household_validations.rb
baarkerlounger: Yeah it's not great. What it's actually doing is expecting there to be a `validate_<page_name>` method for every page. This page happens to be called `household_number_of_other_members`. We should maybe make the pattern something more like `validate_<page_name>_page`

app/models/case_log.rb
baarkerlounger: If the answer to "how many other household members are there?" is 3 then you have 4 people total, so only fields `person_5_age` onwards etc are not required

kosiakkatrina: ohhhh 🤦 I've mistaken this with the part that runs the validations. This makes sense

app/validations/household_validations.rb
kosiakkatrina: riiiight, yeah, makes sense

baarkerlounger: Shall we refactor that before this goes in to make that clearer?

app/validations/household_validations.rb
```diff
@@ -56,6 +51,21 @@ def validate_household_pregnancy(record)
...
+      validate_person_age(record, n)
+      validate_person_age_matches_economic_status(record, n)
+      validate_person_age_matches_relationship(record, n)
+      validate_person_age_and_gender_match_economic_status(record, n)
+      validate_person_age_and_relationship_matches_economic_status(record, n)
+    end
+    validate_partner_count(record)
+  end
+
+  def validate_person_1_age(record)
```
baarkerlounger, line 65: Think Katrina's right this also needs to validate `validate_person_age_matches_economic_status` for person 1

app/validations/household_validations.rb
kosiakkatrina: sure, if changing the pattern isn't too difficult

app/validations/household_validations.rb
MatthewPhelanMadeTech: just for 2-8

app/validations/household_validations.rb
MatthewPhelanMadeTech: How do you both think would be the best way to refactor?

baarkerlounger: Actually I think it makes sense to consider that out of scope for this PR since this is how it's already working. Can do that as a separate thing if needed?

PR #75, opened 2021-11-03
CLDC-639: Active admin

baarkerlounger:
Adds a basic admin interface with a dashboard with basic charting ability, and a case log view that allows CSV downloads

image

image


dushan-madetech:
Is there any non-test data on the Gov PAAS deployment that this would give access to?

baarkerlounger:
@dushan-madetech nope there are no real users yet. Will also put this behind auth as soon as your work goes in, just didn't want to duplicate it and get huge merge conflicts.

dushan-madetech:
Awesome, I just wanted to check if maybe it should be locked behind something temporary like a feature flag if there was something to hide away, will go on and approve now

dushan-madetech (approved):
This seems good to me since there's no live data in the deployment, 🚢 it

PR #76, opened 2021-11-04
[WIP] JsonValidate_Initial

magicmilo:


app/helpers/json_schema_validation.rb
```diff
@@ -0,0 +1,49 @@
+require "json-schema"
```
baarkerlounger: I would put this into a rake task (`lib/tasks/json_schema_validation`) so that you can run it from terminal like `rake form_definition:validate` or similar.

baarkerlounger: Probably doesn't really belong in the app since I don't think we'd ever expect to run it at runtime?

baarkerlounger (commented):
This is really cool.

app/helpers/json_schema_validation.rb
magicmilo: Getting frustrating error where run from rake it won't load the json-schema library

lib/tasks/form_definition.rake
```diff
@@ -0,0 +1,6 @@
+namespace :form_definition do
+    desc "Validate JSON against Generic Form Schema"
+    task :validate do
+        ruby "app/helpers/json_schema_validation.rb"
```
baarkerlounger: Ah sorry I think this wasn't very clear. I'd move that whole files logic into here rather than keeping a file in the app dir for this.

app/helpers/json_schema_validation.rb
baarkerlounger: What happens if you require it in the rakefile?

magicmilo: Fixed thanks to Dushan. was doing gem install rather than adding to the gemfile and will this ran when called directly. didn't find it from rake.

lib/tasks/form_definition.rake
magicmilo: I moved it into lib along with the rake file but think its cleaner as seperate file?

magicmilo: Ok all in one now

PR #77, opened 2021-11-04
Soft validations with controllers

baarkerlounger:
This PR changes the implementation of soft validation override questions so that they're shown or hidden by Javascript (stimulus controller) rather than server side. This has the advantage of meaning the check will run on every page load, including when it's cached (i.e. when you've navigated to it using the back button), without needing to override back button behaviour. We've avoided duplicating the checking logic across front and backend by having the check implemented as a get request to a dedicated controller.

Still TBC whether this UI pattern is wanted at all vs making all questions 1 question per page and doing this purely through routing, but for the current UI pattern this seems like a better approach.


app/javascript/controllers/conditional_question_controller.js
```diff
@@ -12,7 +12,6 @@ export default class extends Controller {
       case "radio":
         this.displayConditionalRadio()
       default:
-        console.log("Not yet implemented for " + this.element.type)
```
dushan-madetech, line 15: Do we not want anything to happen on the default case?

app/javascript/controllers/soft_validations_controller.js
```diff
@@ -0,0 +1,28 @@
...
+          let innerHTML = div.innerHTML
+          innerHTML = innerHTML.replace("soft-validations-placeholder-message", response["label"])
+          innerHTML = innerHTML.replace("soft-validations-placeholder-hint-text", response["hint"])
+          div.innerHTML = innerHTML
+        } else {
+          div.style.display = "none"
+          let buttons = document.getElementsByName(`case_log[override_net_income_validation][]`)
+          Object.entries(buttons).forEach(([idx, button]) => {
+            button.checked = false
+          })
```
dushan-madetech: I'm assuming this is covered by the existing case log feature specs and the new one you added so there's no need to unit test this?

dushan-madetech (approved):
Left some comments for discussion but nothing worth blocking this for. 🚢 it when you feel it's ready to go in main 👍

app/javascript/controllers/conditional_question_controller.js
baarkerlounger: I don't think so, it's more there as a guard but we shouldn't have any conditionals that don't fall into one of the explicit cases (since we wouldn't know how to action them anyway)

app/javascript/controllers/soft_validations_controller.js
baarkerlounger: Exactly, this is just a refactoring of approach, not a change of outcomes. It does solve a bug with the back link but there is a test for that added.

MatthewPhelanMadeTech: using a .map here may be quicker than a .forEach

PR #78, opened 2021-11-05
Cldc 576 reproduce about this log

magicmilo:


app/helpers/check_answers_helper.rb
```diff
@@ -28,6 +28,7 @@ def total_questions(subsection, case_log, form)
   def filter_conditional_questions(questions, case_log)
     applicable_questions = questions
 
+    # puts applicable_questions.count
```
baarkerlounger: This looks like it was for debugging?

app/views/form/_select_question.html.erb
```diff
@@ -0,0 +1,11 @@
+
+<%= answers = question["answer_options"].map {|key, value| OpenStruct.new(id:key, name: value)}
```
baarkerlounger, line 2: Needs linting I think

baarkerlounger:
Am I right in thinking this PR just bring in the about this log pages, it doesn't yet enable complex routing? (i.e. things like, if it's a relet 4-5 pages become unnecessary to show?)

app/helpers/check_answers_helper.rb
```diff
@@ -56,11 +56,12 @@ def condition_not_met(case_log, question_key, question, condition)
...
       operand = condition[/\d+/].to_i
       case_log[question_key].blank? || !case_log[question_key].send(operator, operand)
+    when "text"
+      case_log[question_key].blank? || !condition.include?(case_log[question_key])
     when "radio"
       case_log[question_key].blank? || !condition.include?(case_log[question_key])
     else
-      raise "Not implemented yet"
-    end
+      raise "Not implemented yet" end
```
baarkerlounger: Don't think you need the `end` here do you?

app/views/form/_select_question.html.erb
baarkerlounger: (Spaces on the `}`)

app/helpers/check_answers_helper.rb
magicmilo: Yep sorry got it

> Am I right in thinking this PR just bring in the about this log pages, it doesn't yet enable complex routing? (i.e. things like, if it's a relet 4-5 pages become unnecessary to show?)

It's to prove that the routing we have already is capable of any level of complexity

app/views/form/_select_question.html.erb
magicmilo: ran rubocop on it. Didn't show any errors.

app/helpers/check_answers_helper.rb
magicmilo: Do need it but moved it onto correct line

app/views/form/_select_question.html.erb
baarkerlounger: Interesting, wonder if it doesn't lint erbs

magicmilo:
> Am I right in thinking this PR just bring in the about this log pages, it doesn't yet enable complex routing? (i.e. things like, if it's a relet 4-5 pages become unnecessary to show?)

It is really just to prove that the logic we have works. We already have complex routing. Any scenario you can think of I'd be happy to prove that.

baarkerlounger (approved):

PR #79, opened 2021-11-05
Db migration

kosiakkatrina:
**What**
Change field names and types to more closely resemble CORE data.
For checkbox answers, pass through "Yes"/"No" instead of true/false from the form in order to map it using enums.

**Why**
To replicate CORE schema and make it easier to run existing queries.

**Anything else**
Display checkbox answers


app/models/case_log.rb
```diff
@@ -47,6 +48,60 @@ class CaseLog < ApplicationRecord
 
   enum status: { "not_started" => 0, "in_progress" => 1, "completed" => 2 }
 
+  enum ethnic: DbEnums.ethnic
```
baarkerlounger, line 51: Could we put the enums into the module as well or do they have to be here?

baarkerlounger (approved):
Great stuff!

app/models/case_log.rb
kosiakkatrina: I don't think we can put it in module, it's something to do with being part of ActiveModel

app/controllers/case_logs_controller.rb
```diff
@@ -107,7 +107,7 @@ def responses_for_page(page)
 
       if question_info["type"] == "checkbox"
         question_info["answer_options"].keys.reject { |x| x.match(/divider/) }.each do |option|
-          result[option] = question_params.include?(option)
+          result[option] = question_params.include?(option) ? "Yes" : "No"
```
baarkerlounger, line 110: This looks a little weird. It won't work for actual boolean fields (which will save "No" as `true`, and probably not for integer fields either?) are checkbox questions consistently stored as strings? Looks like some are ints and some booleans from the schema?

kosiakkatrina: Ah, just noticed the boolean checkboxes, they should really be integer. I'll fix that

PR #81, opened 2021-11-08
WIP: Cldc 571 allow users to login and change password

dushan-madetech:


config/initializers/devise.rb
```diff
@@ -0,0 +1,311 @@
...
+
+  # ==> Controller configuration
+  # Configure the parent class to the devise controllers.
+  # config.parent_controller = 'DeviseController'
+
+  # ==> Mailer Configuration
+  # Configure the e-mail address which will be shown in Devise::Mailer,
+  # note that it will be overwritten if you use your own mailer class
+  # with default "from" parameter.
+  config.mailer_sender = "please-change-me-at-config-initializers-devise@example.com"
```
baarkerlounger: Do we know who this should be coming from?

spec/support/controller_macros.rb
```diff
@@ -0,0 +1,16 @@
+module ControllerMacros
+  # def login_admin
+  #   before(:each) do
+  #     @request.env["devise.mapping"] = Devise.mappings[:admin]
+  #     sign_in FactoryBot.create(:admin) # Using factory bot as an example
+  #   end
+  # end
+
+  def login_user
```
baarkerlounger, line 9: This is very nice. Would have expected the namespace to be more like `/spec/helpers/login_helper` or something like that though maybe?

config/initializers/devise.rb
dushan-madetech: No but I can find out. I assume it's the email address we want registration and password related emails to be sent from so I suspect that Sona or Sarah might have an answer or will be able to point me in the right direction

spec/support/controller_macros.rb
dushan-madetech: I was following the advice in this article https://yuta-san.medium.com/a-simple-login-test-with-rspec-devise-factorybot-in-rails-29aeb2ebc4ab which I think is based on the information contained in here https://github.com/heartcombo/devise/wiki/How-To:-Test-controllers-with-Rails-%28and-RSpec%29

dushan-madetech: I'm happy to change this if you think it should be changed 😄

lib/tasks/form_definition.rake
```diff
@@ -1,50 +1,51 @@
 require "json"
 require "json-schema"
-
+# rubocop:disable Lint/ShadowingOuterLocalVariable
```
baarkerlounger: Would suggest reverting this, it's been resolved in main now

spec/rails_helper.rb
```diff
@@ -113,4 +113,10 @@
   # Silence capybara logging puma start up messages to stdout on first js: true
   # spec
   Capybara.server = :puma, { Silent: true }
+
+  # For Devise > 4.1.1
+  config.include Devise::Test::ControllerHelpers, type: :controller
+  config.include Devise::Test::IntegrationHelpers, type: :request
+  # Use the following instead if you are on Devise <= 4.1.1
```
baarkerlounger: Can probably remove these comments? Not ever going to be relevant to us I think

spec/support/devise.rb
```diff
@@ -0,0 +1,9 @@
+require_relative "./controller_macros"
+
+RSpec.configure do |config|
+  # For Devise > 4.1.1
```
baarkerlounger: Same for these

baarkerlounger (approved):

baarkerlounger (approved):

PR #83, opened 2021-11-09
Cldc 591/further data changes

kosiakkatrina:
Add derived fields for postcodes:
- `property_postcode` == `postcode` + `postcod2`
- `previous_postcode` == `ppostc1` + `ppostc2`

Change location field names and store ECode instead of a string location value:
- `property_location` == `la`
- `previous_la` == `prevloc`

Remove `condition_effects_prefer_not_to_say` as part of the data requirements

Rename and map values for the following fields:
`reason_for_leaving_last_settled_home` to `reason`
`property_reference` to `propcode`
`property_major_repairs` to `majorrepairs`
`housing_benefit` to `hb`
`outstanding_rent_or_charges` to `hbrentshortfall`
`outstanding_amount` to `tshortfall`

Add Slovakia, Slovenia and Croatia options to `national` field
Remove `Prefer not to say` from `illness`
Add `Do not know` to `reasonpref`


dushan-madetech (approved):
Forgot to click approve earlier, good stuff! 🚢 it

app/controllers/case_logs_controller.rb
```diff
@@ -116,6 +117,23 @@ def responses_for_page(page)
     end
   end
 
+  def get_inferred_answers(responses_for_page)
```
baarkerlounger: I think this should probably go in the model since it's logic about how to handle data rather than request/response logic?

db/migrate/20211108143649_further_change_field_types.rb
```diff
@@ -0,0 +1,21 @@
+class FurtherChangeFieldTypes < ActiveRecord::Migration[6.1]
+  def up
+    change_table :case_logs, bulk: true do |t|
+      t.change :reason, "integer USING reason::integer"
```
baarkerlounger: These sometimes fail if there's data that can't be changed so worth checking if that's likely to be the case. Otherwise slightly safer might be removing and adding.

app/controllers/case_logs_controller.rb
kosiakkatrina: Probably, I'll see how I can move that out

kosiakkatrina: Does it look better now?

app/constants/db_enums.rb
```diff
@@ -249,4 +259,439 @@ def self.latime
...
+      "Discharged from long stay hospital or similar institution" => 45,
+      "Other" => 20,
+      "Do not know" => 28,
+      "Prefer not to say" => 100,
+    }
+  end
+
+  def self.la
+    {
+      "Hartlepool" => "E06000001",
```
baarkerlounger, line 315: Ouch. Nice work getting hold of this data!

app/models/case_log.rb
```diff
@@ -38,6 +38,7 @@ class CaseLog < ApplicationRecord
   include Discard::Model
   include SoftValidations
   include DbEnums
+  require "uk_postcode"
```
baarkerlounger: Don't think we need this? Rails should autoload this no?

db/migrate/20211108143649_further_change_field_types.rb
kosiakkatrina: oooh, that's fair. I'll change that

kosiakkatrina: Updated that now

db/migrate/20211108134601_further_core_migrations.rb
```diff
@@ -0,0 +1,58 @@
...
+      t.column :postcode, :string
+      t.column :postcod2, :string
+      t.column :ppostc1, :string
+      t.column :ppostc2, :string
+      t.remove :property_relet
+      t.column :property_relet, :integer
+    end
+  end
+
+  def down
```
baarkerlounger, line 30: This isn't currently reversible, I think because of the order columns are being added and removed in

```
$ rake db:rollback STEP=1
== 20211108134601 FurtherCoreMigrations: reverting ============================
-- change_table(:case_logs, {:bulk=>true})
rake aborted!
StandardError: An error has occurred, this and all later migrations canceled:

PG::DuplicateColumn: ERROR:  column "condition_effects_prefer_not_to_say" of relation "case_logs" already exists
/home/dan/Documents/mhclg-data-collection-beta/db/migrate/20211108134601_further_core_migrations.rb:31:in `down'

Caused by:
ActiveRecord::StatementInvalid: PG::DuplicateColumn: ERROR:  column "condition_effects_prefer_not_to_say" of relation "case_logs" already exists
/home/dan/Documents/mhclg-data-collection-beta/db/migrate/20211108134601_further_core_migrations.rb:31:in `down'

Caused by:
PG::DuplicateColumn: ERROR:  column "condition_effects_prefer_not_to_say" of relation "case_logs" already exists
/home/dan/Documents/mhclg-data-collection-beta/db/migrate/20211108134601_further_core_migrations.rb:31:in `down'
Tasks: TOP => db:rollback

```

kosiakkatrina: fixed

baarkerlounger (approved):

docs/api/DLUHC-CORE-Data.v1.json
```diff
@@ -309,7 +309,7 @@
             "tenancy": "Fixed term – Secure",
             "lettype": "Affordable Rent - General Needs",
             "landlord": "This landlord",
-            "property_location": "Barnet",
+            "la": "Barnet",
```
baarkerlounger, line 312: Looks like we missed the update to `property_major_repairs` here

PR #84, opened 2021-11-09
CLDC-638: Bulk upload

baarkerlounger:
image

Things for subsequent tickets:

- [ ] How do we handle data type errors in the template (i.e. a value that rails can't save at all so the whole `CaseLog.create` fails, currently we silently fail/drop those rows but we probably need to handle it in a way that's explicit to the user).


app/models/bulk_upload.rb
```diff
@@ -0,0 +1,187 @@
...
+  include ActiveModel::Model
+  include ActiveModel::Validations
+  include ActiveModel::Conversion
+
+  SPREADSHEET_CONTENT_TYPES = %w[
+    application/vnd.ms-excel
+    application/vnd.openxmlformats-officedocument.spreadsheetml.sheet
+  ].freeze
+
+  FIRST_DATA_ROW = 7
```
dushan-madetech, line 11: Is the 7 likely to change or will it always be a constant?

dushan-madetech (approved):
It looks good to me, 🚢 it

app/models/bulk_upload.rb
baarkerlounger: My assumption is it will be constant as long as we use this template because 1-6 are the template headers. TBC though.

app/models/bulk_upload.rb
```diff
@@ -0,0 +1,193 @@
...
+    sheet = xlsx.sheet(0)
+    last_row = sheet.last_row
+    if last_row < FIRST_DATA_ROW
+      errors.add(:case_log_bulk_upload, "No data found")
+    else
+      data_range = FIRST_DATA_ROW..last_row
+      data_range.map do |row_num|
+        case_log = CaseLog.create
+        map_row(sheet.row(row_num)).each do |attr_key, attr_val|
+          begin
```
baarkerlounger, line 31: This is horrendously slow but the idea is to save all valid data, and not error when you hit an invalid bit. Not sure it can be easily parallelized (like this) because all the things here will be acting on the same table row... Anyone got any better ideas?

PR #85, opened 2021-11-11
CLDC-677: Save mcrdate as datetime

kosiakkatrina:
- Rename `property_major_repairs_date` to `mcrdate`
- Change `mcrdate` from string to datetime
- Add inferred `mcrday`, `mcrmonth` and `mcryear` fields
- Fix saving of the date input. In the UI we're asking for a date as 3 separate inputs and that comes through as `name_of_the_question(1i)`, `name_of_the_question(2i)` and `name_of_the_question(3i)` which previously wasn't saving to the db


dushan-madetech (approved):
Good stuff 👍, feel free to 🚢 it

PR #86, opened 2021-11-11
CLDC-678: Store total household members

kosiakkatrina:
Store other housdehold members (`other_hhmemb`) and infer total household members (`hhmemb`) by adding 1 to `other_hhmemb`


dushan-madetech (approved):
🚢  it

PR #87, opened 2021-11-11
Add incref column

kosiakkatrina:


dushan-madetech (approved):
🚢 it

app/models/case_log.rb
```diff
@@ -247,6 +251,10 @@ def mandatory_fields
       dynamically_not_required << "ecstat#{n}"
     end
 
+    if net_income_known != "Prefer not to say"
```
baarkerlounger, line 258: Think we also need `No` here?

kosiakkatrina: I wouldn't think that income not known == refused to say (?)

baarkerlounger: You're right, sorry misread this!

baarkerlounger (approved):

PR #90, opened 2021-11-15
Bug fixes

baarkerlounger:
- [x] Use the correct blue

image

### DIFF
--- a/app/javascript/styles/_application-custom.scss
+++ b/app/javascript/styles/_application-custom.scss
@@ -1,6 +1,0 @@
-$govuk-global-styles: true;
-
-// turbo-frame {
-//   display: block;
-//   border: 1px solid blue
-// }
--- a/app/javascript/styles/_application-custom.scss
+++ b/app/javascript/styles/_application-custom.scss
@@ -1,21 +1,18 @@
-.a:link,
-.govuk-link:link {
+.a,
+.govuk-link {
   color: #1d70b8;
-}
-
-.a:visited
-.govuk-link:visited {
-  color: #4c2c92;;
-}
-
-.a:focus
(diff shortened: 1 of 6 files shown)


app/javascript/styles/_application-custom.scss
```diff
@@ -0,0 +1,21 @@
...
+  &__visited {
+    color: #4c2c92;
+  }
+  &__link:focus {
+    color: #0b0c0c;
+  }
+  &__link:hover {
+    color: #003078;
+  }
+}
```
paulrobertlloyd: Not sure you need these styles as they exist in `govuk-frontend` CSS. If the intention is to automatically style any link correctly, there is a setting that can be added to our application SCSS file:

`$govuk-global-styles: true;`

This will automatically style links as if they had the `govuk-link` class (and paragraphs as if they had the `govuk-body` class). Opinions on this setting are mixed; possibly bit of a blunt instrument, but it works (we use this setting in the prototype).

paulrobertlloyd: Oh wait… where are `.govuk-link__link` and `.govuk-link__visited` coming from?

baarkerlounger: Was taking it directly from browser inspector on the prototype, but have replaced it with the global setting now.

paulrobertlloyd: Cool. I don’t think you need this separate stylesheet; this setting can go at the top of `app/javascript/styles/application.scss`

baarkerlounger: Done

paulrobertlloyd (approved):

PR #92, opened 2021-11-15
CLDC-709: Conditional section status bugfix

baarkerlounger:


app/models/form.rb
```diff
@@ -97,6 +97,36 @@ def all_questions
     }.reduce(:merge)
   end
 
+  def filter_conditional_questions(questions, case_log)
```
kosiakkatrina, line 100: Is this being used in multiple places/ non check_answers page?
Should the form know about the case_log?

kosiakkatrina: Ah, I see it's in the tasklist

baarkerlounger: It's being used to calculate which questions should show/still need to be answered in the Check Answers page, and it's used on the Tasklist page to show the status of each section.

I'd agree that having the form know about the case log here is possibly a slightly leaky abstraction but I think this is the most logical place for it to sit for now? Open to suggestion though.

kosiakkatrina (approved):
Happy for this to go in, looks sensible. Only not sure about case_log being the model. I'm guilty of doing that in one of the methods, but still unsure if that's the correct way to do it. Seems like tight coupling

app/models/form.rb
kosiakkatrina: I haven't quite figured out how this could be done differently, so happy for this to be there till a time when a logical refactor comes our way

kosiakkatrina: `So happy` == `therefore happy` and not == `very happy`

PR #93, opened 2021-11-16
Bugfixes

baarkerlounger:
In order for page validation methods to run on Save and Continue, they need to match the name of the page in the JSON. Ticket here for a more long term fix https://digital.dclg.gov.uk/jira/browse/CLDC-711.

### DIFF
--- a/config/forms/2021_2022.json
+++ b/config/forms/2021_2022.json
@@ -196,7 +196,7 @@
                 }
               }
             },
-            "age1": {
+            "person_1_age": {
               "header": "",
               "description": "",
               "questions": {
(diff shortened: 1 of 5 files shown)


kosiakkatrina (approved):

config/forms/2021_2022.json
```diff
@@ -196,7 +196,7 @@
                 }
               }
             },
-            "age1": {
+            "person_1_age": {
```
MatthewPhelanMadeTech, line 199: I'm not sure how this conflicts with @kosiakkatrina 's data storage ticket

baarkerlounger: It shouldn't do, this is changing the `page` name which is what the validations depend on, whereas @kosiakkatrina is changing `questions` which is what the data storage depends on.

PR #95, opened 2021-11-16
CLDC-687 implement property information

magicmilo:


db/migrate/20211119104835_add_property_info_fields.rb
```diff
@@ -0,0 +1,12 @@
+class AddPropertyInfoFields < ActiveRecord::Migration[6.1]
+  def change
+    change_table :case_logs, bulk: true do |t|
+      t.column :do_you_know_the_postcode, :integer
+      t.column :do_you_know_the_local_authority, :integer
+      t.column :why_dont_you_know_la, :string
+      t.column :first_time_property_let_as_social_housing, :integer
+      t.column :type_property_most_recently_let_as, :integer
+      t.column :builtype, :integer
```
baarkerlounger: `buildtype`? (Typo)

db/migrate/20211119154133_change_column_types.rb
```diff
@@ -0,0 +1,21 @@
+class ChangeColumnTypes < ActiveRecord::Migration[6.1]
+  def up
+    change_table :case_logs, bulk: true do |t|
+      t.change :first_time_property_let_as_social_housing, :string
```
baarkerlounger: Given that this hasn't gone into main yet you can probably combine these migrations (i.e. just a single migration that creates these columns with the right types in the first place)?

db/migrate/20211119104835_add_property_info_fields.rb
```diff
@@ -0,0 +1,12 @@
+class AddPropertyInfoFields < ActiveRecord::Migration[6.1]
+  def change
+    change_table :case_logs, bulk: true do |t|
+      t.column :do_you_know_the_postcode, :integer
```
baarkerlounger: Am I right in thinking these fields don't need adding to the API docs because they're UI only and we're not expecting API clients to pass them? If so they might need to be added to the optional fields in the CaseLog model so that submission validation doesn't fail when they're not provided?

db/migrate/20211119154133_change_column_types.rb
magicmilo: Done this all in one

db/migrate/20211119104835_add_property_info_fields.rb
```diff
@@ -0,0 +1,13 @@
+class AddPropertyInfoFields < ActiveRecord::Migration[6.1]
+  def change
+    change_table :case_logs, bulk: true do |t|
+      t.column :do_you_know_the_postcode, :string
```
baarkerlounger: Should these be `ints` (Yes/No enums) so it's consistent with how we're doing it for other fields?

db/migrate/20211119104835_add_property_info_fields.rb
magicmilo: This was from the data dictionary general needs lettings.  

24 | BUILTYPE | Building type | Type of building

baarkerlounger: My bad :)

db/migrate/20211119104835_add_property_info_fields.rb
magicmilo: Done this for three fields

db/migrate/20211119104835_add_property_info_fields.rb
```diff
@@ -0,0 +1,11 @@
+class AddPropertyInfoFields < ActiveRecord::Migration[6.1]
+  def change
+    change_table :case_logs, bulk: true do |t|
+      t.column :first_time_property_let_as_social_housing, :int
+      t.column :why_dont_you_know_la, :string
+      t.column :type_property_most_recently_let_as, :string
+      t.column :builtype, :string
+      t.rename :tenant_same_property_renewal, :renewal
```
baarkerlounger: Renewal should also be an int/enum right? I think that applies to basically all radio button questions so probably also `buildtype` and `type_property_most_recently_let_as` (is that the field name the analysts have in the data dictionary btw?)

spec/fixtures/complete_case_log.json
```diff
@@ -1,12 +1,15 @@
 {
   "case_log": {
     "tenant_code": "T657",
+    "tenanceother": "",
```
baarkerlounger: What is `tenanceother`?

spec/fixtures/complete_case_log.json
```diff
@@ -57,13 +60,13 @@
...
     "property_relet": "No",
-    "rsnvac": "Relet - tenant abandoned property",
+    "rsnvac": "First let of newbuild property",
     "property_reference": "P9876",
     "unittype_gn": "House",
     "property_building_type": "dummy",
     "beds": 3,
     "property_void_date": "03/11/2019",
-    "majorrepairs": "Yes",
+    "majorrepairs": "No",
```
baarkerlounger, line 68: How come this one needs to change?

spec/fixtures/complete_case_log.json
magicmilo: Must've been bad find and replace

spec/fixtures/complete_case_log.json
magicmilo: Fails if not changed looking at it.
 1) Form Features Soft Validation given a weekly net income that is above the expected amount for the given economic status but below the hard max clears the confirmation question if the amount was amended and the page is returned to using the back button
     Failure/Error: @case_log = CaseLog.find(params[:case_log_id])
     
     ActiveRecord::RecordNotFound:
       Couldn't find CaseLog with 'id'=50 [WHERE "case_logs"."discarded_at" IS NULL]
     # ./app/controllers/case_logs_controller.rb:95:in `block (2 levels) in <class:CaseLogsController>'
     # ------------------
     # --- Caused by: ---
     # Capybara::CapybaraError:
     #   Your application server raised an error - It has been raised in your test code because Capybara.raise_server_errors == true

db/migrate/20211119104835_add_property_info_fields.rb
magicmilo: Both ints yes type_property_most_recently_let_as is UNITLETAS and changed to reflect

db/migrate/20211124152204_change_recently_let_as_to_enum.rb
```diff
@@ -0,0 +1,19 @@
+class ChangeRecentlyLetAsToEnum < ActiveRecord::Migration[6.1]
```
baarkerlounger, line 1: Worth merging these two migrations again as well?

baarkerlounger (approved):

spec/fixtures/forms/test_aboutthislog.json
```diff
@@ -1,179 +0,0 @@
-{
```
baarkerlounger, line 1: Really nice we could DRY this up!

config/forms/2021_2022.json
```diff
@@ -1537,7 +1680,21 @@
...
+              "description": "",
+              "questions": {
+                "beds": {
+                  "check_answer_label": "void/renewal date",
+                  "header": "What is the void or renewal date?",
+                  "hint_text": "",
+                  "type": "date"
+                }
+              },
+              "depends_on": { "rsnvac": "First let of newbuild property", "tenant_same_property_renewal": "No" }
```
kosiakkatrina, line 1697: Why does this depend on rsnvac? edit: nevermind, it's newpop in the designs

PR #97, opened 2021-11-17
CLDC-700: date validation 

kosiakkatrina:
Add date validation for mrcdate column
This  is also done for the remainder of the dates, but they need to be migrated to a date time format in order to work

### DIFF
--- a/app/controllers/case_logs_controller.rb
+++ b/app/controllers/case_logs_controller.rb
@@ -107,9 +107,13 @@ private
         day = params["case_log"]["#{question_key}(3i)"]
         month = params["case_log"]["#{question_key}(2i)"]
         year = params["case_log"]["#{question_key}(1i)"]
-        next unless day.present? && month.present? && year.present?
+        next unless [day, month, year].any?(&:present?)
 
-        result[question_key] = Date.new(year.to_i, month.to_i, day.to_i)
+        result[question_key] = if day.to_i.between?(1, 31) && month.to_i.between?(1, 12) && year.to_i.between?(2000, 2200)
+                                 Date.new(year.to_i, month.to_i, day.to_i)
+                               else
+                                 Date.new(0, 1, 1)
+                               end
       end
       next unless question_params
 
--- a/app/models/case_log.rb
+++ b/app/models/case_log.rb
@@ -5,6 +5,7 @@ class CaseLogValidator < ActiveModel::Validator
   include PropertyValidations
   include FinancialValidations
   include TenancyValidations
+  include DateValidations
 
   def validate(record)
     # If we've come from the form UI we only want to validate the specific fields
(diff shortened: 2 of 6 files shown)


app/controllers/case_logs_controller.rb
```diff
@@ -107,9 +107,13 @@ def responses_for_page(page)
...
         month = params["case_log"]["#{question_key}(2i)"]
         year = params["case_log"]["#{question_key}(1i)"]
-        next unless day.present? && month.present? && year.present?
+        next unless [day, month, year].any?(&:present?)
 
-        result[question_key] = Date.new(year.to_i, month.to_i, day.to_i)
+        result[question_key] = if day.to_i.between?(1, 31) && month.to_i.between?(1, 12) && year.to_i.between?(2000, 2200)
+                                 Date.new(year.to_i, month.to_i, day.to_i)
+                               else
+                                 Date.new(0, 1, 1)
```
kosiakkatrina, line 115: this is a bit hacky, but I couldn't figure out a way to indicate that the attempt to fill in the date happened vs users just ignoring the question

baarkerlounger: What about saving the `day, month, year` fields here and then deriving the datetime in the validation?

baarkerlounger (approved):

app/controllers/case_logs_controller.rb
kosiakkatrina: I thought about that, it would work for mrcdate, but for other dates I'm not sure if we're partitioning them into separate fields

baarkerlounger: Fair enough. Would be nice to have a consistent way to treat them I suppose but happy for this to go in first anyway.

app/controllers/case_logs_controller.rb
```diff
@@ -107,9 +107,13 @@ def responses_for_page(page)
         day = params["case_log"]["#{question_key}(3i)"]
         month = params["case_log"]["#{question_key}(2i)"]
         year = params["case_log"]["#{question_key}(1i)"]
-        next unless day.present? && month.present? && year.present?
+        next unless [day, month, year].any?(&:present?)
```
baarkerlounger, line 110: Don't you need all of these?

kosiakkatrina: Yeah, so that gets checked on the next line. This check right here is to see if any fields at all have been filled in. This indicates that there was an attempt to fill out the date. After that we check if the values were valid and set it to a fake (01, 01, 0) date in order to display the error.

We wouldn't want it to fail if there was no attempt to fill in the fields to begin with, but we'd want it to fail if only 1 or 2 fields are filled in instead of just not saving it silently

baarkerlounger: Oh yeah, of course. Thanks

PR #98, opened 2021-11-17
Refactor routing

baarkerlounger:
This PR makes some pretty major changes to how page routing works. It changes the logic from 
```
current question -> next question
```

to

```
next_question -> evaluate if all conditions have been met to show
```

The advantages of this are that

- We don't need large amounts of duplication in the form definition (repeating page definitions for each possible route)
- It allows us to easily calculate whether a page was shown or not
- This fixes some bugs with showing the right questions in the check answers section

### DIFF
--- a/app/models/form.rb
+++ b/app/models/form.rb
@@ -108,22 +108,15 @@ class Form
   end
 
   def page_routed_to?(page, case_log)
-    return true unless pages_with_routing_conditions.include?(page)
+    return true unless conditions = page_dependencies(page)
 
-    applicable_conditions = all_routing_conditions.flat_map { |conditions|
-      conditions.select { |k, _condition| k == page }
-    }.flat_map(&:values)
-    applicable_conditions.all? do |condition|
-      condition.all? { |k, v| case_log[k].present? && case_log[k] == v }
+    conditions.all? do |question, value|
+      case_log[question].present? && case_log[question] == value
     end
   end
 
-  def pages_with_routing_conditions
-    @pages_with_routing_conditions ||= all_routing_conditions.flat_map(&:keys)
-  end
-
-  def all_routing_conditions
-    @all_routing_conditions ||= all_pages.map { |_page_key, page| page["conditional_route_to"] }.compact
+  def page_dependencies(page)
+    all_pages[page]["depends_on"]
   end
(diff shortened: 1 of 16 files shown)


app/models/form.rb
```diff
@@ -110,6 +107,25 @@ def filter_conditional_questions(questions, case_log)
     applicable_questions
   end
 
+  def page_routed_to?(page, case_log)
```
baarkerlounger, line 110: This could be simplified quite a lot if we changed the form definition schema so that the conditions for showing a page were attached to the question it affects rather than to the question providing the data. Given the size of this PR I think that should be a subsequent step though.

spec/features/case_log_spec.rb
```diff
@@ -466,27 +466,17 @@ def answer_all_questions_in_income_subsection
...
       click_button("Save and continue")
+      expect(page).to have_current_path("/case_logs/#{id}/household_number_of_other_members")
       visit("/case_logs/#{id}/conditional_question")
-      choose("case-log-preg-occ-yes-field", allow_label_click: true)
+      choose("case-log-preg-occ-no-field", allow_label_click: true)
       click_button("Save and continue")
-      expect(page).to have_current_path("/case_logs/#{id}/rent")
+      expect(page).to have_current_path("/case_logs/#{id}/conditional_question_no_page")
+      click_button("Save and continue")
+      expect(page).to have_current_path("/case_logs/#{id}/conditional_question/check_answers")
```
kosiakkatrina, line 492: Any reason for changing this test instead of adding a new one? Surely the routing should have stayed the same?

baarkerlounger: Routing doesn't necessarily stay the same (behaviour wise), because this PR removes default next page and instead steps through the form definition page by page and assumes the next page is the next page unless it's conditions aren't met.

I think this particular feature spec might be better off being unit tests (which are elsewhere on form_model and check_answers_helper specs) but kept this for now.

app/models/form.rb
baarkerlounger: Now done

magicmilo (approved):
Certainly ends up cleaner and more elegant. Perfect

PR #100, opened 2021-11-17
CLDC-706: Check answers should be a single summary list

baarkerlounger:
image

Will update PR target once the fix routing PR has been merged.


paulrobertlloyd (approved):
Perfect – that’s the ticket!

PR #101, opened 2021-11-18
CLDC-681-armed-forces: Change armed forces question

kosiakkatrina:
Merged armed forces and armed forces question into one as per reviewed household needs section flow.
Save armed forces answer as enums.


baarkerlounger (approved):
:tada:

PR #102, opened 2021-11-18
CLDC-696: Fix heading sizes

kosiakkatrina:
Update the headings. Set top level headings from `xl` to `l` and use descending sizes for subheadings:

image

image

image

image

image


baarkerlounger:
This is such a nice PR to review

app/views/case_logs/_tasklist.html.erb
```diff
@@ -1,7 +1,7 @@
 <ol class="app-task-list app-task-list--no-numbers">
   <% @form.all_sections.map do |section_key, section_value| %>
     <li>
-      <h2 class="app-task-list__section">
+      <h2 class="app-task-list__section govuk-heading-m">
```
paulrobertlloyd: I’m not sure this is needed here, as `app-task-list__section` should set the correct size.

app/javascript/styles/_task-list.scss
```diff
@@ -13,7 +13,6 @@
 
 .app-task-list__section {
   display: table;
-  @include govuk-font($size:24, $weight: bold);
```
paulrobertlloyd: Oh, I see. I think we can leave this, at least for now, as the task list section heading sizes shouldn’t need to change.

app/views/case_logs/edit.html.erb
```diff
@@ -1,10 +1,10 @@
 <%= turbo_frame_tag "case_log_form", target: "_top" do %>
   <div class="govuk-grid-row">
     <div class="govuk-grid-column-two-thirds-from-desktop">
-      <h1 class="govuk-heading-xl">Tasklist for log
+      <h1 class="govuk-heading-l">Tasklist for log
         <%= @case_log.id %></h1>
 
-      <h2 class="govuk-heading-s govuk-!-margin-bottom-2">This submission is
+      <h2 class="govuk-heading-m govuk-!-margin-bottom-2">This submission is
```
paulrobertlloyd: I’m going to be horribly inconsistent here and say that in this particular case, the heading should maintain the small size. Semantically it is a second level heading, but should be slightly deemphasised from the section headings used in the task list below.

paulrobertlloyd (changes requested):
This is brilliant! Just a few small nits.

Will there be a second PR for adding captions to headings showing the section a question is in?

kosiakkatrina:
> This is brilliant! Just a few small nits.
> 
> Will there be a second PR for adding captions to headings showing the section a question is in?

Ooh, maybe I misunderstood. So should it be:
Section heading (stays the same for all the questions in the section, for example `housing needs`) -> `l`
Page heading (if exists) -> `m`
Question label (if there is page heading) -> `s`
Question label (if there is no page heading) -> `m`
?

app/javascript/styles/_task-list.scss
kosiakkatrina: I think that the size gets set by adding `govuk-heading-m` to the element, so this `@include` is redundant. Did you mean that we should bring it back or is it ok as is?

paulrobertlloyd:
@kosiakkatrina So the heading sizes for question pages are now correct. What’s missing is the corresponding captions, the grey text that appears above the heading:

Screenshot 2021-11-18 at 17 12 23

Both the caption and the heading should use the `l` size.

app/javascript/styles/_task-list.scss
paulrobertlloyd: Yup, I would revert this change

app/views/case_logs/_tasklist.html.erb
paulrobertlloyd: …and revert this change too.

kosiakkatrina: I'm happy to change it, but it sorta make it inconsistent the way we set heading styles, as `app-task-list__section` is something we've defined

kosiakkatrina:
@paulrobertlloyd 
Added captions now, looks like this:
image

app/views/case_logs/_tasklist.html.erb
paulrobertlloyd: There’s not a GDS component for the task list yet (frustratingly), but they do provide the code for it via the prototype kit and it [uses this pattern for headings](https://govuk-prototype-kit.herokuapp.com/docs/templates/task-list). When/if it becomes a component, it would likely be designed very similarly.

kosiakkatrina: Ah I see, okay! I'll revert it then!

kosiakkatrina:
@paulrobertlloyd I reverted the change from the last comment, is this okay to deploy or is there anything else I should fix up?

app/views/form/page.html.erb
```diff
@@ -6,15 +6,18 @@
   <div class="govuk-grid-row">
     <div class="govuk-grid-column-two-thirds-from-desktop">
       <% if page_info["header"].present? %>
-        <h1 class="govuk-heading-xl">
+        <h1 class="govuk-caption-l">
+          <%= subsection %>
+        </h1>
+        <h1 class="govuk-heading-l">
```
paulrobertlloyd: The caption should be in a `span`:

```suggestion
        <h1 class="govuk-heading-l">
          <span class="govuk-caption-l">
            <%= subsection %>
          </span>
```

kosiakkatrina: done!

paulrobertlloyd (approved):
Thanks @kosiakkatrina!

PR #103, opened 2021-11-18
Add section dependencies

baarkerlounger:
so that you can mark that a section "Cannot be started" until it's dependencies have been met.

image


MatthewPhelanMadeTech:
I'm happy to approve this as the functionality seems to work well, I am however getting some failing tests around 'supcharg' and 'Capybara::ElementNotFound:Unable to find field "case-log-earnings-field" that is not disabled' (for example)

MatthewPhelanMadeTech (approved):

PR #104, opened 2021-11-19
CLDC-724: Admin Panel requires authentication

baarkerlounger:
- [x] ActiveAdmin requires authentication to access
- [x] Admin Users can be created or removed using ActiveAdmin UI

image
image
image


dushan-madetech (approved):
Excellent stuff, 🚢 it

PR #106, opened 2021-11-22
Refactor the form parsing and navigation logic into OOP domain objects

baarkerlounger:
This PR refactors the form parsing and navigation logic. It takes it from being split across various helpers and a very large form class that deals mostly in Hash map structures that are hard to reason about at a glance without REPLing, and moves it into smaller classes (section, subsection, page, question).

This should make it easier to TDD/change in the future and easier to compare to the Form definition and see at a glance what options are present and what functionality they control.

On this branch removing Case Log feature specs (acceptance tests) reduces Spec Coverage by 1.41% from 94.01% to 93.6%. On main the same change reduces spec coverage by 4.33% from 93.38% to 89.05%. With increased confidence in unit testing we could hopefully reduce reliance on (slower) acceptance testing over time.

### DIFF
--- a/app/controllers/case_logs_controller.rb
+++ b/app/controllers/case_logs_controller.rb
@@ -57,8 +57,8 @@ class CaseLogsController < ApplicationController
   def submit_form
     form = FormHandler.instance.get_form("2021_2022")
     @case_log = CaseLog.find(params[:id])
-    @case_log.page = params[:case_log][:page]
-    page = form.get_page(@case_log.page)
+    @case_log.page_id = params[:case_log][:page]
+    page = form.get_page(@case_log.page_id)
     responses_for_page = responses_for_page(page)
     if @case_log.update(responses_for_page) && @case_log.has_no_unresolved_soft_errors?
       redirect_path = form.next_page_redirect_path(page, @case_log)
@@ -101,7 +101,7 @@ private
   API_ACTIONS = %w[create show update destroy].freeze
 
   def responses_for_page(page)
-    page.expected_responses.each_with_object({}) do |(question), result|
+    page.expected_responses.each_with_object({}) do |question, result|
       question_params = params["case_log"][question.id]
       if question.type == "date"
         day = params["case_log"]["#{question.id}(3i)"]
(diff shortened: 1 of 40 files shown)


dushan-madetech (approved):
A really good refactor 😄. I liked the ADR explanation and the PR description giving context behind this change, 🚢 it

app/controllers/case_logs_controller.rb
```diff
@@ -101,29 +101,28 @@ def check_answers
...
 
   def responses_for_page(page)
-    form = FormHandler.instance.get_form("2021_2022")
-    form.expected_responses_for_page(page).each_with_object({}) do |(question_key, question_info), result|
-      question_params = params["case_log"][question_key]
-      if question_info["type"] == "date"
-        day = params["case_log"]["#{question_key}(3i)"]
-        month = params["case_log"]["#{question_key}(2i)"]
-        year = params["case_log"]["#{question_key}(1i)"]
+    page.expected_responses.each_with_object({}) do |(question), result|
```
kosiakkatrina: do we need `()` around `question`?

app/models/form.rb
```diff
@@ -1,171 +1,51 @@
...
+    subsections.find { |s| s.pages.find { |p| p.id == page.id } }
   end
 
   def next_page(page, case_log)
-    subsection = subsection_for_page(page)
-    page_idx = pages_for_subsection(subsection).keys.index(page)
-    nxt_page = pages_for_subsection(subsection).keys[page_idx + 1]
+    page_ids = subsection_for_page(page).pages.map(&:id)
+    page_idx = page_ids.index(page.id)
+    nxt_page = get_page(page_ids[page_idx + 1])
```
kosiakkatrina: I think these should be `page-index` and `next_page`

kosiakkatrina (approved):
Big fan of this!
I maybe would have liked to see more commits, so that I could follow the refactor step by step, but other than that, absolutely beautiful

app/models/form.rb
baarkerlounger: Have changed `page_index`. `nxt_page` was to avoid shadowing the method name so left that one for now.

PR #110, opened 2021-11-25
Your account page

dushan-madetech:
Your account page allows users to see and edit the data held about them

https://digital.dclg.gov.uk/jira/browse/CLDC-571


config/routes.rb
```diff
@@ -1,14 +1,19 @@
 Rails.application.routes.draw do
   devise_for :admin_users, ActiveAdmin::Devise.config
-  devise_for :users, controllers: { passwords: "users/passwords" }
+  devise_for :users, controllers: { passwords: "users/passwords" }, :skip => [:registrations] 
   devise_scope :user do
     get "confirmations/reset", to: "users/passwords#reset_confirmation"
-  end
+    get 'users/edit' => 'devise/registrations#edit', :as => 'edit_user_registration'
```
baarkerlounger: Do you actually need this one? Is it not the default?

app/views/users/account/index.html.erb
```diff
@@ -0,0 +1,73 @@
+<div class="govuk-grid-row">
```
baarkerlounger, line 1: Can we prettify this one? Indentation looks quite different from the others.

app/controllers/users/account_controller.rb
```diff
@@ -0,0 +1,21 @@
...
+  def index
+    check_logged_in
+  end
+
+  def personal_details
+    check_logged_in
+  end
+
+  def update
+    if current_user.update('name': params[:user][:name], 'email': params[:user][:email])
```
baarkerlounger: The `permit_params` pattern might be better here? Something like:

```
def update
  if current_user.update(user_params)
    redirect_to(users_account_path)
  else
  # What do we do here?
  end
end

def user_params
  params.require(:user).permit(:email, :name, :password)
end
```

app/helpers/account_helper.rb
```diff
@@ -0,0 +1,13 @@
+module AccountHelper
+  def resource_name
+    :user
+  end
+
+  def resource
+    @resource = current_user
+  end
+
+  def devise_mapping
```
baarkerlounger: Is this being used anywhere?

config/routes.rb
baarkerlounger: Yes you do...

app/helpers/account_helper.rb
MatthewPhelanMadeTech: The views wouldn't get the resource with out it

app/controllers/users/account_controller.rb
MatthewPhelanMadeTech: I'll have a look at it now

PR #112, opened 2021-11-26
CLDC-711: Refactor Validations

baarkerlounger:
Originally we thought that we'd need to have a way of determining which validations should run for a page on form submit, and separate that from validations running for an API or Bulk Upload create, because a form page submission, wouldn't have all the fields yet and so wouldn't be able to validate everything.

Our new design means this isn't actually the case anymore though because:

- We don't check for "presence" of data as an ActiveModel validation anymore, we do that as part of determining the Case Log's `status`. This means that validating records with missing mandatory data is now fine.
- The form submit controller method only permits the response params expected for each page based on the questions that page is defined to have, so there's no danger of it trying to set other random data.

As a result I think we can simplify our code a bit with this PR. We no longer need to couple validation method names to the JSON form definition (`validate_<page_name>`), we can just validate the entire record on each page submit. This also means, people writing the Form Definition don't need to know anything about the validations that will run.

### DIFF
--- a/app/controllers/case_logs_controller.rb
+++ b/app/controllers/case_logs_controller.rb
@@ -57,8 +57,7 @@ class CaseLogsController < ApplicationController
   def submit_form
     form = FormHandler.instance.get_form("2021_2022")
     @case_log = CaseLog.find(params[:id])
-    @case_log.page_id = params[:case_log][:page]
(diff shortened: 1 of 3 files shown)


dushan-madetech (approved):
Love the explanation behind the refactor 🙌

PR #114, opened 2021-11-26
CLDC-673: Organisation page

baarkerlounger:
This PR adds the pre-requisites for organisations. All case logs now have two associations to an Organisation and through that to a User, i.e. you can do:

```
case_log.owning_organisation
case_log.managing_organisation

organisation.owned_case_logs
organisation.managed_case_logs
organisation.case_logs

user.owned_case_logs
user.managed_case_logs
user.case_logs
```

This means we can now start restricting resource views to only case logs that are associated with a Users organisation, which this PR does for all existing web pages.

This PR doesn't limit that for API use yet, which can still act on all Case Logs. That will come when we add API users/auth.

- [x] Create Organisation model
- [x] Create Organisation/User associations
- [x] Create Organisation/CaseLogs/User associations
- [x] Create admin panel to create/update Organisations
- [x] Create admin panel to create/update Users
- [x] Restrict Case log index view to organisation members
- [x] Restrict Case log edit view to organisation members
- [x] Restrict Case log submit form path to organisation members
- [x] Restrict Case log check answers page to organisation members
- [x] Enable updating user model from admin panel without inputting password
- [ ] Add organisation link to navigation bar (TODO in separate PR)
- [ ] Add organisation page view (TODO in separate PR)


db/migrate/20211126142105_user_belongs_to_organisation.rb
```diff
@@ -0,0 +1,15 @@
+class UserBelongsToOrganisation < ActiveRecord::Migration[6.1]
+  def up
+    change_table :users, bulk: true do |t|
+      t.remove :organisation
+      t.belongs_to :organisation
```
baarkerlounger: Could make this a foreign key by doing `t.belongs_to :organisation, foreign_key: true` but not done so for now since we might already have user records and it's not yet clear if this is necessary or even correct (must all users belong to an organisation?). For now I think using plain ActiveRecord association here is fine.

app/admin/case_logs.rb
```diff
@@ -2,7 +2,122 @@
...
+                   reasonable_preference_reason_homeless
+                   reasonable_preference_reason_unsatisfactory_housing
+                   reasonable_preference_reason_medical_grounds
+                   reasonable_preference_reason_avoid_hardship
+                   reasonable_preference_reason_do_not_know
+                   other_tenancy_type
+                   override_net_income_validation
+                   net_income_known
+                   owning_organisation_id
+                   managing_organisation_id]
```
dushan-madetech, line 120: This is a lot nicer 👍

spec/controllers/admin/organisations_controller_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+      expect { post :create, session: valid_session, params: params }.to change(Organisation, :count).by(1)
+    end
+  end
+
+  describe "Update organisation" do
+    before do
+      get :edit, session: valid_session, params: { id: organisation.id }
+    end
+
+    it "creates a new admin users" do
```
dushan-madetech, line 37: `creates new admin users` or `creates a new admin user`? The current text reads a bit weird

spec/controllers/admin/users_controller_spec.rb
```diff
@@ -0,0 +1,71 @@
...
+        user: {
+          email: "somethin5@example.com",
+          name: "Jane",
+          password: "pAssword1",
+          organisation_id: organisation.id,
+        },
+      }
+    end
+
+    it "creates a new users" do
```
dushan-madetech: This one too, `creates new users` or `creates a new user`?

spec/factories/user.rb
```diff
@@ -2,6 +2,7 @@
   factory :user do
     sequence(:email) { |i| "test#{i}@example.com" }
     password { "pAssword1" }
+    organisation
```
dushan-madetech, line 5: Would this not be ` organisation { FactoryBot.create(:organisation) }`?

dushan-madetech (approved):
Really nice stuff here 👍, it looks read to be 🚢ped

spec/factories/user.rb
baarkerlounger: This is shorthand for that. So `organisation` tells FactoryBot to try and look up a factory with that name.

dushan-madetech: Ah I see 😄

PR #118, opened 2021-11-29
Formbuilder fixes

paulrobertlloyd:
* Make `GOVUKDesignSystemFormBuilder` the default form builder
* Update devise forms to use `GOVUKDesignSystemFormBuilder`
* Remove `autofocus` attribute on forms (`autofocus` can be problematic for people with motor control disabilities, as it may create extra work for them to navigate out from the autofocused area and to other locations on the page.)

Tests pass, but I can’t log in with the credentials (shown in `.env.example`), so unable to test all the different views. 

Additionally, there is no server side validation to understand what value may be incorrect (see [this bug](https://digital.dclg.gov.uk/jira/browse/CLDC-760); submitting the sign in form only generates a `401` in the background.)

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -1,4 +1,6 @@
 class ApplicationController < ActionController::Base
+  default_form_builder GOVUKDesignSystemFormBuilder::FormBuilder
+
   def render_not_found_html
     render file: Rails.root.join("public/404.html"), status: :not_found
   end
--- a/app/views/case_logs/bulk_upload.html.erb
+++ b/app/views/case_logs/bulk_upload.html.erb
@@ -1,5 +1,5 @@
 <div class="govuk-form-group">
-  <%= form_for @bulk_upload, url: bulk_upload_case_logs_path, method: "post", builder: GOVUKDesignSystemFormBuilder::FormBuilder do |f| %>
+  <%= form_for @bulk_upload, url: bulk_upload_case_logs_path, method: "post" do |f| %>
     <%= f.govuk_error_summary %>
     <%= f.govuk_file_field :case_log_bulk_upload,
         label: { text: "Bulk Upload", size: "l" },
--- a/app/views/devise/confirmations/new.html.erb
+++ b/app/views/devise/confirmations/new.html.erb
@@ -3,14 +3,13 @@
 <%= form_for(resource, as: resource_name, url: confirmation_path(resource_name), html: { method: :post }) do |f| %>
   <%= render "devise/shared/error_messages", resource: resource %>
 
-  <div class="field">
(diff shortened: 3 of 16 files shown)


app/controllers/application_controller.rb
```diff
@@ -1,2 +1,3 @@
 class ApplicationController < ActionController::Base
+  default_form_builder GOVUKDesignSystemFormBuilder::FormBuilder
```
baarkerlounger, line 2: Would expect this to be in a `config/initializer` rather than the base controller?

paulrobertlloyd: Was following [the instructions on the form builder site](https://govuk-form-builder.netlify.app/introduction/getting-started/), but more than happy to move to a different location if you can show me how/where 🙂

baarkerlounger: Appears I'm wrong and this is now the recommended place to put it. Nice :)

baarkerlounger:
The credentials in `.env.example` are not intended to be actual credentials they're just placeholders for real ones. Those also refer to the email address notifications are sent from rather than user credentials.

To test sign in you can run `rake db:seed` which will create a user `test@example.com` with password `password` (https://github.com/communitiesuk/mhclg-data-collection-beta/blob/main/db/seeds.rb)

baarkerlounger:
Not entirely sure what you mean by no server side validation. The 401 is server side, returned by the post request.

If you check your rails server you'll see something like:

```
Started POST "/users/sign_in" for 127.0.0.1 at 2021-11-29 19:57:10 +0000
Processing by Devise::SessionsController#create as TURBO_STREAM
  Parameters: {"authenticity_token"=>"[FILTERED]", "user"=>{"email"=>"test@example.com", "password"=>"[FILTERED]"}}
  User Load (1.2ms)  SELECT "users".* FROM "users" WHERE "users"."email" = $1 ORDER BY "users"."id" ASC LIMIT $2  [["email", "test@example.com"], ["LIMIT", 1]]
Completed 401 Unauthorized in 6ms (ActiveRecord: 1.2ms | Allocations: 1687)
```

paulrobertlloyd:
> Not entirely sure what you mean by no server side validation. The 401 is server side, returned by the post request.

Sorry, that was poorly worded (was a bit thrown by the way devise seems to be doing something client side in terms of form submission/validation). What I meant was that there is no error summary or error messages, i.e. the ones used by the GOV.UK Design System. If you disable JavaScript, some validation is shown, but using a success flash message.

baarkerlounger:
Added a PR to fix that.

Have you checked  `# app/views/users/account/index.html.erb` as well? Looks like that one might be handcrafted html rather than gem components?

baarkerlounger (approved):

PR #119, opened 2021-11-29
Devise doesn't play nice with Turbo yet

baarkerlounger:
This points at a branch of Devise that fixes it for Turbo (Rails 7 is expected before the end of the year so hopefully not too long until this is merged or supported in main) https://github.com/heartcombo/devise/pull/5340.

It also swaps devises default flash message on error for Gov UK errors. Needs some tests still but otherwise looks like it works.

Could also fix this by turning off Turbo for the login form specifically by adding `html: {"data-turbo" => "false"}` but then you lose the SPA-like goodness:

```
# app/views/devise/sessions/new.html.erb

<%= form_for(resource, as: resource_name, html: {"data-turbo" => "false"}, url: session_path(resource_name)) do |f| %>
```

image
image
image

image
image

### DIFF
--- a/app/controllers/users/sessions_controller.rb
+++ b/app/controllers/users/sessions_controller.rb
@@ -2,7 +2,7 @@ class Users::SessionsController < Devise::SessionsController
   def create
     self.resource = resource_class.new
     if params.dig("user", "email").empty?
-      resource.errors.add :email, "Please enter email address"
+      resource.errors.add :email, "Enter an email address"
     elsif !email_valid?(params.dig("user", "email"))
       resource.errors.add :email, "Enter an email address in the correct format, like name@example.com"
     end
--- a/app/controllers/users/sessions_controller.rb
+++ b/app/controllers/users/sessions_controller.rb
@@ -1,0 +1,16 @@
+class Users::SessionsController < Devise::SessionsController
+  def create
+    self.resource = resource_class.new
+    if params.dig("user", "email").empty?
+      resource.errors.add :email, "Please enter email address"
+    end
+    if params.dig("user", "password").empty?
+      resource.errors.add :password, "Please enter password"
+    end
+    if resource.errors.present?
+      render :new, status: :unprocessable_entity
+    else
+      super
+    end
+  end
+end
--- a/app/controllers/users/sessions_controller.rb
+++ b/app/controllers/users/sessions_controller.rb
@@ -7,7 +7,7 @@ class Users::SessionsController < Devise::SessionsController
       resource.errors.add :email, "Enter an email address in the correct format, like name@example.com"
     end
     if params.dig("user", "password").empty?
-      resource.errors.add :password, "Please enter password"
+      resource.errors.add :password, "Enter a password"
     end
     if resource.errors.present?
       render :new, status: :unprocessable_entity
--- a/app/controllers/users/sessions_controller.rb
+++ b/app/controllers/users/sessions_controller.rb
@@ -4,7 +4,7 @@ class Users::SessionsController < Devise::SessionsController
     if params.dig("user", "email").empty?
       resource.errors.add :email, "Please enter email address"
     elsif !email_valid?(params.dig("user", "email"))
-      resource.errors.add :email, "Email addess is not valid"
+      resource.errors.add :email, "Enter an email address in the correct format, like name@example.com"
     end
     if params.dig("user", "password").empty?
       resource.errors.add :password, "Please enter password"
--- a/app/controllers/users/sessions_controller.rb
+++ b/app/controllers/users/sessions_controller.rb
@@ -3,6 +3,8 @@ class Users::SessionsController < Devise::SessionsController
     self.resource = resource_class.new
     if params.dig("user", "email").empty?
       resource.errors.add :email, "Please enter email address"
+    elsif !email_valid?(params.dig("user", "email"))
+      resource.errors.add :email, "Email addess is not valid"
     end
     if params.dig("user", "password").empty?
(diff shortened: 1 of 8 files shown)


app/views/devise/sessions/new.html.erb
```diff
@@ -3,6 +3,9 @@
     <div class="govuk-grid-column-two-thirds">
       <h1 class="govuk-heading-l">Sign in to your account to submit CORE data</h1>
 
+      <% flash_to_model_errors(resource) %>
+      <%= f.govuk_error_summary %>
+
```
paulrobertlloyd, line 6: The error summary should appear above the `h1`. We also need to show error messages on the respective fields that have been validated.

baarkerlounger: Moved the errors above the header and updated the screenshot. 

I'd say you generally shouldn't add the error to specific fields because it allows malicious attackers to check whether an email exists in the system or not, which is not usually recommended?

paulrobertlloyd: If I don’t enter any value into any field, I’d expect validation to tell me to enter an email address and password.

If I enter an email in the incorrect format, I’d expect validation to tell me the input is not using a valid email format.

That sort of thing.

baarkerlounger: Added those two

kosiakkatrina (approved):

app/controllers/users/sessions_controller.rb
```diff
@@ -0,0 +1,24 @@
+class Users::SessionsController < Devise::SessionsController
+  def create
+    self.resource = resource_class.new
+    if params.dig("user", "email").empty?
+      resource.errors.add :email, "Please enter email address"
+    elsif !email_valid?(params.dig("user", "email"))
+      resource.errors.add :email, "Email addess is not valid"
```
paulrobertlloyd:
```suggestion
      resource.errors.add :email, "Enter an email address in the correct format, like name@example.com"
```

app/controllers/users/sessions_controller.rb
```diff
@@ -0,0 +1,24 @@
+class Users::SessionsController < Devise::SessionsController
+  def create
+    self.resource = resource_class.new
+    if params.dig("user", "email").empty?
+      resource.errors.add :email, "Please enter email address"
```
paulrobertlloyd:
```suggestion
      resource.errors.add :email, "Enter an email address"
```

app/controllers/users/sessions_controller.rb
```diff
@@ -0,0 +1,24 @@
...
+    elsif !email_valid?(params.dig("user", "email"))
+      resource.errors.add :email, "Email addess is not valid"
+    end
+    if params.dig("user", "password").empty?
+      resource.errors.add :password, "Please enter password"
```
paulrobertlloyd:
```suggestion
      resource.errors.add :password, "Enter a password"
```

spec/features/user_spec.rb
```diff
@@ -80,6 +81,34 @@
...
+      expect(page).to have_selector("#user-password-field-error")
+    end
+
+    it "show specific field error messages if an invalid email address is entered" do
+      visit("/case_logs")
+      fill_in("user[email]", with: "thisisn'tanemail")
+      click_button("Sign in")
+      expect(page).to have_selector("#error-summary-title")
+      expect(page).to have_selector("#user-email-field-error")
+      expect(page).to have_content(/Email addess is not valid/)
```
paulrobertlloyd:
```suggestion
      expect(page).to have_content(/Enter an email address in the correct format, like name@example.com/)
```

paulrobertlloyd (changes requested):

paulrobertlloyd:
This message should use the notification banner (information aka blue) component, not the error summary component.

Screenshot 2021-12-01 at 11 32 32

paulrobertlloyd:
This is possibly a separate PR as likely effects multiple pages in the service, but page titles should be unique (i.e. `[Short page heading] - [Service name] - GOV.UK`).

When forms are validated, the page title should be prefixed with `Error: `.

paulrobertlloyd:
> This message should use the notification banner (information aka blue) component, not the error summary component.

To be honest, removing this banner would be simplest, as not sure this message is necessary.

baarkerlounger:
@paulrobertlloyd have removed that error message entirely.

On page title - I think they are all unique? Just that showing an error doesn't show a different page it shows the same page... I think that's the standard way of doing this but if it needs to change that definitely should be a separate PR.

paulrobertlloyd:
> On page title - I think they are all unique?

They are not, sadly. All pages have a `title` with the value ‘DLUHC CORE Data Collection’.

It should be something like:

`Sign in – [Service name] - GOV.UK`
`Case logs – [Service name] - GOV.UK`
`Case log EX001 – [Service name] - GOV.UK`
etc. etc.

> Just that showing an error doesn't show a different page it shows the same page... I think that's the standard way of doing this but if it needs to change that definitely should be a separate PR.

That’s correct, but the page needs to have `Error: ` prefix the title, i.e.:

`Error: Sign in – [Service name] - GOV.UK`

On my previous project, we created a little helper to achieve this: https://github.com/DFE-Digital/apply-for-teacher-training/blob/e7c4fa34f1e3fae294a9992d2113d1897c9f82d2/app/helpers/view_helper.rb#L40

I’ll create a bug card.

baarkerlounger:
Oh sorry titles! Was thinking of URLs. Cool yep, new bug please.

PR #121, opened 2021-11-30
Uses dashes to separate words in URLs

paulrobertlloyd:
The [URL standards for GOV.UK](https://www.gov.uk/guidance/content-design/url-standards-for-gov-uk) state that:

> dashes should be used to separate words within URLs so they are easy to read - for example, <https://www.gov.uk/set-up-business> (this might not apply if the URL is designed to be read aloud)

We should follow the same URL standards as used on GOV.UK and many other government services.

## Todo
- [x] Devise routes
- [x] Case log routes
- [x] Bulk upload routes
- [ ] Sub section names that appear in routing
- [ ] Soft validations route
- [ ] Page fragments (aka `id`s)

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -1,6 +1,6 @@
 Rails.application.routes.draw do
   devise_for :admin_users, ActiveAdmin::Devise.config
-  devise_for :users, controllers: { passwords: "users/passwords" }, skip: [:registrations]
+  devise_for :users, controllers: { passwords: "users/passwords" }, path_names: { sign_in: 'sign-in', sign_out: 'sign-out' }, skip: [:registrations]
   devise_scope :user do
     get "confirmations/reset", to: "users/passwords#reset_confirmation"
     get "users/edit" => "devise/registrations#edit", :as => "edit_user_registration"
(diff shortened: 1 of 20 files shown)


paulrobertlloyd:
@baarkerlounger I have got the tests passing, but when creating a new log, I get the following error:

> NoMethodError (undefined method `id' for nil:NilClass):
>
> app/controllers/case_logs_controller.rb:166:in \`org_params'
> app/controllers/case_logs_controller.rb:158:in \`case_log_params'
> app/controllers/case_logs_controller.rb:12:in `create'

```diff
def org_params
  {
-   "owning_organisation_id" => current_user.organisation.id,
    "managing_organisation_id" => current_user.organisation.id,
  }
end
```

Any ideas why this is failing?

spec/controllers/case_logs_controller_spec.rb
```diff
@@ -162,10 +162,10 @@
 
       it "routes to the appropriate conditional page based on the question answer of the current page" do
         post :submit_form, params: { id: id, case_log: case_log_form_conditional_question_yes_params }
-        expect(response).to redirect_to("/case_logs/#{id}/conditional_question_yes_page")
```
baarkerlounger, line 165: Shouldn't all the pages be dashes as well? i.e. `"/case_logs/#{id}/conditional-question-yes-page"`

baarkerlounger: Oh sorry you have a TODO for that already

config/routes.rb
```diff
@@ -13,15 +13,15 @@
...
-  resources :case_logs do
+  resources :case_logs, :path => '/case-logs' do
     collection do
-      post "/bulk_upload", to: "bulk_upload#bulk_upload"
-      get "/bulk_upload", to: "bulk_upload#show"
+      post "/bulk-upload", to: "bulk_upload#bulk_upload"
+      get "/bulk-upload", to: "bulk_upload#show"
     end
 
     member do
```
baarkerlounger, line 27: You'd want to change `get page.id.to_s, to: "case_logs##{page.id}"` here to:

`get page.id.to_s.dasherize, to: "case_logs##{page.id}"` I think. Or else change the page ids in the form definition (`# config/forms/2021_2022.json`

baarkerlounger:
@paulrobertlloyd that failure will be because you don't have an organisation for your user currently. Run:

```
rails c
```

```
org = Organisation.create!(
  name: "DLUHC",
  address_line1: "2 Marsham Street",
  address_line2: "London",
  postcode: "SW1P 4DF",
  local_authorities: "None",
  holds_own_stock: false,
  other_stock_owners: "None",
  managing_agents: "None"
)

User.all.each do |u| 
    u.organisation = org
    u.save!
end
```
```

paulrobertlloyd:
Closed in favour of #128.

PR #123, opened 2021-12-01
Cldc 648 derived rent type

kosiakkatrina:
Add `renttype` column and infer rent type value based on `rent_type`.
Persist inferred field values to the db


db/schema.rb
```diff
@@ -168,6 +168,8 @@
     t.bigint "owning_organisation_id"
     t.bigint "managing_organisation_id"
     t.datetime "property_void_date"
+
```
dushan-madetech: I don't think there should be a newline here

spec/models/case_log_spec.rb
```diff
@@ -843,4 +835,73 @@
...
+    require "date"
+    let(:organisation) { FactoryBot.create(:organisation) }
+    let!(:case_log) do
+      CaseLog.create({
+        managing_organisation: organisation,
+        owning_organisation: organisation,
+        property_postcode: "M1 1AE",
+        previous_postcode: "M2 2AE",
+        # rubocop:disable Style/DateTime
+        mrcdate: DateTime.new(2021, 5, 4),
```
dushan-madetech, line 849: Could you not use Time.new(2021, 5, 4) instead of disabling the Rubocop?

dushan-madetech (approved):

spec/models/case_log_spec.rb
kosiakkatrina: That fails the test, it sets the month to 3 instead of 4, I assume because it starts indexing at 0 🤔

app/models/case_log.rb
```diff
@@ -212,6 +161,31 @@ def update_status!
...
+    self.ppostc2 = UKPostcode.parse(previous_postcode).incode if previous_postcode.present?
+    if mrcdate.present?
+      self.mrcday = mrcdate.day
+      self.mrcmonth = mrcdate.month
+      self.mrcyear = mrcdate.year
+    end
+    self.incref = 1 if net_income_known == "Prefer not to say"
+    self.hhmemb = other_hhmemb + 1 if other_hhmemb.present?
+
+    rent_type_mapping = {
```
baarkerlounger: This could be a constant outside this method?

app/models/case_log.rb
```diff
@@ -212,6 +161,31 @@ def update_status!
                   else
                     "in_progress"
                   end
+    set_derived_fields
+  end
+
+  def set_derived_fields
+    self.postcode = UKPostcode.parse(property_postcode).outcode if property_postcode.present?
```
baarkerlounger: Probably better to wrap these in a single `if` since it's the same condition for all 4?

spec/models/case_log_spec.rb
baarkerlounger: Weird. But so couldn't you just increment it? Like `Time.new(2021, 5, 5)`?

app/models/case_log.rb
kosiakkatrina: it's checking `property_postcode` for 2 of them and `previous_postcode` for the other 2, but I'll extract them anyway

spec/models/case_log_spec.rb
kosiakkatrina: Yeah, I though it'd be confusing to read

PR #125, opened 2021-12-01
Reset password validation

baarkerlounger:


config/webpacker.yml
```diff
@@ -16,7 +16,7 @@ default: &default
   cache_manifest: false
 
   # Extract and emit a css file
-  extract_css: true
+  extract_css: false
```
paulrobertlloyd, line 19: Was this change intentional?

baarkerlounger: Yeah, it causes issues for rspec (think you need to compile assets before running tests?) so wanted to revert this until checking it properly. Left the `inline: false` for dev setting though.

app/views/devise/passwords/new.html.erb
```diff
@@ -9,7 +9,7 @@
   <div class="govuk-grid-row">
     <div class="govuk-grid-column-two-thirds">
       <h1 class="govuk-heading-l">Reset password</h1>
-      <%= render "devise/shared/error_messages", resource: resource %>
+      <%= f.govuk_error_summary %>
```
paulrobertlloyd: Error summary should appear before the page title.

config/webpacker.yml
paulrobertlloyd: 👍

paulrobertlloyd (approved):

PR #127, opened 2021-12-01
Cldc 647 lettype derived variable

kosiakkatrina:
Derive `lettype` value from the `renttype`, `needstype` and `providertype` fields


app/admin/organisations.rb
```diff
@@ -17,7 +17,7 @@
     selectable_column
     id_column
     column :name
-    column :org_type
+    column :providertype
```
baarkerlounger: Should probably give this a readable name?

app/models/organisation.rb
```diff
@@ -3,6 +3,9 @@ class Organisation < ApplicationRecord
   has_many :owned_case_logs, class_name: "CaseLog", foreign_key: "owning_organisation_id"
   has_many :managed_case_logs, class_name: "CaseLog", foreign_key: "managing_organisation_id"
 
+  include DbEnums
+  enum providertype: DbEnums.providertype, _suffix: true
```
baarkerlounger: Think the `display_attributes` method here will also need updating since it refers to `org_type` still?

baarkerlounger:
Does anything need changing on the bulk upload?

app/admin/organisations.rb
kosiakkatrina: Yeah, was just wondering if we need it to be the same as in the data dictionary for core?

baarkerlounger: We don't but I think consistency with the data dictionary is fine anyway. Can give this column a different name from the field name by passing a string though. Look at the one below.

app/models/organisation.rb
kosiakkatrina: Yes! Missed it, thanks!

app/admin/organisations.rb
kosiakkatrina: Ah, that's interesting! Let me check it out

kosiakkatrina:
> Does anything need changing on the bulk upload?

@baarkerlounger Don't think so 🤔

app/admin/organisations.rb
```diff
@@ -17,7 +17,7 @@
     selectable_column
     id_column
     column :name
-    column :org_type
+    column "org_type", :providertype
```
baarkerlounger: Would suggest "Org type" or "Organisation type" so it's humanized, consistent with other table labels.

baarkerlounger (approved):

PR #128, opened 2021-12-01
Uses dashes to separate words in URLs

paulrobertlloyd:
The [URL standards for GOV.UK](https://www.gov.uk/guidance/content-design/url-standards-for-gov-uk) state that:

> dashes should be used to separate words within URLs so they are easy to read - for example, <https://www.gov.uk/set-up-business> (this might not apply if the URL is designed to be read aloud)

We should follow the same URL standards as used on GOV.UK and many other government services.

### DIFF
--- a/app/models/form.rb
+++ b/app/models/form.rb
@@ -12,11 +12,11 @@ class Form
   end
 
   def get_subsection(id)
-    subsections.find { |s| s.id == id }
+    subsections.find { |s| s.id == id.to_s.underscore }
   end
 
   def get_page(id)
-    pages.find { |p| p.id == id }
+    pages.find { |p| p.id == id.to_s.underscore }
   end
 
   def subsection_for_page(page)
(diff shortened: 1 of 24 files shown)


app/models/form.rb
```diff
@@ -12,11 +12,11 @@ def initialize(form_path)
   end
 
   def get_subsection(id)
-    subsections.find { |s| s.id == id }
+    subsections.find { |s| s.id == id.to_s.underscore }
```
baarkerlounger, line 15: Might be better to just expect them to be dasherized in the form definition JSON? We do control that spec after all.

baarkerlounger (approved):

app/models/form.rb
paulrobertlloyd: I tried that approach initially, but that involved changing more areas of the codebase (i.e. things like `question.hint_text` would need to be `question["hint-text"]`). But am happy with either approach.

baarkerlounger: Wouldn't have to be consistent for all keys in the JSON, just page ids (although I appreciate that makes a find and replace fiddly). Not particularly important I think this is fine to go in as is.

paulrobertlloyd: Underscores versus dashes is something I go back and forwards on… pros and cons to each approach, as is so often the case 🤪

PR #129, opened 2021-12-01
Show correct headings and captions on question pages, add width option to schema

paulrobertlloyd:
* Allow schema to provide a width value for a text or numeric input – @baarkerlounger have I implemented this correctly?
* Fix display of captions and headings on form questions
* Correct page title for check your answers pages
* Content tweaks for 2021/22 schema – ‘About this log’ is correct, other sections have largely been left as still in flux

| Before | After |
| - | - |
| Screenshot 2021-12-01 at 17 50 49 | Screenshot 2021-12-01 at 17 50 24 |
| Screenshot 2021-12-01 at 17 50 59 | Screenshot 2021-12-01 at 17 50 00 |

### DIFF
--- a/app/views/form/check_answers.html.erb
+++ b/app/views/form/check_answers.html.erb
@@ -1,7 +1,10 @@
 <%= turbo_frame_tag "case_log_form", target: "_top" do %>
   <div class="govuk-grid-row">
     <div class="govuk-grid-column-three-quarters-from-desktop">
-      <h1 class="govuk-heading-l">Check the answers you gave for <%= subsection.id.humanize(capitalize: false) %></h1>
+      <h1 class="govuk-heading-l">
+        <span class="govuk-caption-l"><%= subsection.id.humanize %></span>
+        Check your answers
+      </h1>
       <%= display_answered_questions_summary(subsection, @case_log) %>
       <dl class="govuk-summary-list govuk-!-margin-bottom-9">
         <% subsection.applicable_questions(@case_log).each do |question| %>
--- a/.github/workflows/pipeline.yml
+++ b/.github/workflows/pipeline.yml
@@ -57,6 +57,10 @@ jobs:
         with:
           # runs 'bundle install' and caches installed gems automatically
           bundler-cache: true
+      - name: Set up node
+        uses: actions/setup-node@v2
+        with:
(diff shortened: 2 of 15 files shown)


app/views/form/check_answers.html.erb
```diff
@@ -1,7 +1,10 @@
 <%= turbo_frame_tag "case_log_form", target: "_top" do %>
   <div class="govuk-grid-row">
     <div class="govuk-grid-column-three-quarters-from-desktop">
-      <h1 class="govuk-heading-l">Check the answers you gave for <%= subsection.id.humanize(capitalize: false) %></h1>
+      <h1 class="govuk-heading-l">
+        <span class="govuk-caption-l"><%= subsection.id.humanize %></span>
+        Check your answers
```
baarkerlounger, line 6: Are these changes Dan has asked for? Think that wording was in the original prototype because it was asked for as part of the content review Stu did?

baarkerlounger:
@paulrobertlloyd yep implementation looks good. Would be good to have a spec for it as well.

paulrobertlloyd:
> yep implementation looks good. Would be good to have a spec for it as well.

For the schema change?

paulrobertlloyd:
> Are these changes Dan has asked for?

This is pretty standard and has been in the prototype throughout testing without being changed, but I’ll double-check with Dan.

baarkerlounger:
For checking that when a width value is passed in the JSON definition, it's rendered correctly by the form building. So in `spec/fixtures` there's a smaller test form definition, be good to include it there on 1 question and have a test to spec it works.

app/views/form/check_answers.html.erb
paulrobertlloyd: Also to note: I had the before and after images for check your answers the wrong way round, if that’s why you asked this question. Images are now showing in the correct order ☺️

baarkerlounger (approved):

baarkerlounger:
@paulrobertlloyd you can cherry-pick this for test to run or wait for it to be merged to main https://github.com/communitiesuk/mhclg-data-collection-beta/pull/134/commits/86baa53972aa87676bb720a9f772620afa9a1be9

PR #132, opened 2021-12-01
Update service name

paulrobertlloyd:
* Update service name to one of the preferred (and shorter!) options
* Add it locales so encourage using a single source of truth; only need to change it one place when we decide on a name
* Update README.md to match


baarkerlounger:
Need to update the spec that's checking for this, otherwise LGTM

baarkerlounger (approved):

PR #135, opened 2021-12-02
Add page titles

kosiakkatrina:
image


dushan-madetech (approved):

paulrobertlloyd:
To save having to repeat the service name and `GOV.UK` each time, I wonder if we can use a helper, and just have pages provide a `:title`

Here’s an example: https://github.com/DFE-Digital/apply-for-teacher-training/blob/1a37fb2cce6c933529eddfa818e4283c56f90966/app/helpers/application_helper.rb#L7

PR #136, opened 2021-12-02
CLDC-763

kosiakkatrina:
Refactor creation of page titles into a method

### DIFF
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -1,7 +1,7 @@
 <!DOCTYPE html>
 <html lang="en" class="govuk-template">
   <head>
-    <title><%= yield(:title).present? ? browser_title(yield(:title)) : t('service_name') %></title>
+    <title><%= browser_title(yield(:title)) %></title>
     <%= csrf_meta_tags %>
     <%= csp_meta_tag %>
     <%= tag :meta, name: 'viewport', content: 'width=device-width, initial-scale=1' %>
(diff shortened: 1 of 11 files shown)


paulrobertlloyd:
@kosiakkatrina I think we can move the helper to the application layout, then pages only need to give the title text.

kosiakkatrina:
> @kosiakkatrina I think we can move the helper to the application layout, then pages only need to give the title text.

@paulrobertlloyd is this what you meant https://github.com/communitiesuk/mhclg-data-collection-beta/pull/136/commits/53b76d38c2760a2736201aaa671e6cbd5a470e87 ?

app/views/layouts/application.html.erb
```diff
@@ -1,7 +1,7 @@
 <!DOCTYPE html>
 <html lang="en" class="govuk-template">
   <head>
-    <title><%= yield(:title).present? ? yield(:title) : t('service_name') %></title>
+    <title><%= yield(:title).present? ? browser_title(yield(:title)) : t('service_name') %></title>
```
paulrobertlloyd: This is close, but if there’s no `:title`, the browser title will be `[Service name]` not `[Service name] - GOV.UK`.

I wonder if we can use the helper like this here:

```suggestion
    <title><%= browser_title(yield(:title)) %></title>
```

And then in the helper, if a title is present, output:

`[title] - [Service name] - GOV.UK`

and if no title is given, output:

`[Service name] - GOV.UK`

paulrobertlloyd:
❤️ it!

PR #137, opened 2021-12-02
Move model modules to models

baarkerlounger:
This PR moves the constants and validations modules to Models since they're actually all exclusively to do with the data layer and not the app in general. 

It also makes the constants actual constants rather than methods.


dushan-madetech (approved):

kosiakkatrina (approved):
This is beautiful!

PR #138, opened 2021-12-02
User roles

baarkerlounger:
This PR adds user roles and restricts the viewing of an organisations user and inviting of new users to data coordinators only.

Data Provider:
image

Data Coordinator:
image


dushan-madetech (approved):
🚢 it

PR #139, opened 2021-12-02
Remove unneeded flash notices and add a notices on password change

kosiakkatrina:


baarkerlounger (approved):

app/controllers/users_controller.rb
```diff
@@ -6,6 +6,7 @@ class UsersController < ApplicationController
   def update
     if current_user.update(user_params)
       bypass_sign_in current_user
+      flash[:notice] = I18n.t("devise.passwords.updated")
```
baarkerlounger, line 9: Should this fire only on password change rather than on any change (which I think it currently does?)

kosiakkatrina: ahh, good point. Let me fix that

PR #140, opened 2021-12-02
Add role to invite

baarkerlounger:
Data coordinators now set the role of the user they're adding:
image

### DIFF
--- a/app/views/users/new.html.erb
+++ b/app/views/users/new.html.erb
@@ -20,14 +20,14 @@
         autocomplete: "name"
       %>
 
-      <%= roles = User::ROLES.map { |key, value| OpenStruct.new(id:key, name: key.humanize) }
-        f.govuk_collection_select :role, roles, :id, :name, label: { text: "Role" }
-      %>
-
       <%= f.govuk_email_field :email,
         label: { text: "Email address" },
         autocomplete: "email",
         value: @resource.email
+      %>
+
+      <%= roles = User::ROLES.map { |key, value| OpenStruct.new(id:key, name: key.humanize) }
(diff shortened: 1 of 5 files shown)


app/views/users/new.html.erb
```diff
@@ -16,6 +16,10 @@
         autocomplete: "name"
       %>
 
+      <%= roles = User::ROLES.map { |key, value| OpenStruct.new(id:key, name: key.humanize) }
+        f.govuk_collection_select :role, roles, :id, :name, label: { text: "Role" }
```
paulrobertlloyd: We can use `govuk_collection_radio_buttons` for these options.

paulrobertlloyd:
Ideally role would be set on a separate page (this is because we can then adjust the content hints shown which refer to organisation, provided in answer to previous question). See: https://core-beta-prototype.herokuapp.com/users/DC001/role

For now however, as we start simple, the role should come last on the page, so:

* Name
* Email address
* Role

baarkerlounger:
@paulrobertlloyd prototype link throws an error for me but I've changed the field to radio buttons and moved them to the bottom

kosiakkatrina (approved):

paulrobertlloyd:
What is a ‘Data accessor’ - is this the ‘Administrator’ role?

paulrobertlloyd:
> prototype link throws an error for me but I've changed the field to radio buttons and moved them to the bottom

Ooops! Here’s the correct link: https://core-beta-prototype.herokuapp.com/users/DCOO1/role

baarkerlounger:
@paulrobertlloyd data accessor is the read only role (they can view/download case logs but not create them and not do any org/user management). We've not really implemented that role yet.

paulrobertlloyd (approved):

PR #141, opened 2021-12-02
Tweak page titles

paulrobertlloyd:
* Adds pages titles to a few pages where they were missing
* Ensures title for question pages are always unique
* Ensures title for check answers pages are always unique
* Plus a few markup/whitespace tweaks


baarkerlounger (approved):
Nice

PR #142, opened 2021-12-02
CLDC-498: property type validation

kosiakkatrina:
Update mappings based on content changes and add tests for property type validation


app/models/case_log.rb
```diff
@@ -157,12 +157,12 @@ def applicable_income_range
 private
 
   RENT_TYPE_MAPPING = {
```
baarkerlounger: This might be better living in the `constants` module now since it seems like it's doing the same job as all the others (mapping values)?

baarkerlounger (approved):

app/models/case_log.rb
kosiakkatrina: Ohh, good shout!

PR #143, opened 2021-12-02
CLDC-769: Scope authentication to the correct resources

baarkerlounger:
This PR refactors out the Case Logs controller (which was getting rather large) into:
1. A Case Logs controller that purely deals with the Case Logs resource
2. A Form controller that deals with the additional "form" UI pages

- [x] A User can only see case log pages for case logs that belong to their organisation
- [x] A User can only submit case log pages for case logs that belong to their organisation
- [x] A User can only see organisation pages (details/users) for their organisation
- [x] A user can only see their own account details
- [x] A user can only bulk upload cases for their own org


docs/adr/adr-012-controller-http-return-statuses.md
```diff
@@ -0,0 +1,15 @@
+### ADR - 012: Controller HTTP return statuses
+
+Controllers assess authentication by 3 criteria:
+
+1. Are you signed in at all?
+2. Are you signed in and requesting an action that your role/user type has access to?
+3. Are you signed in, requesting an action that your role/user type has access to and requesting a resource that your user has access to.
+
+When these aren't met the fail with the following response types:
```
dushan-madetech:
```suggestion
When these aren't met they fail with the following response types:
```

spec/requests/case_log_controller_spec.rb
```diff
@@ -180,84 +180,63 @@
           allow(FormHandler.instance).to receive(:get_form).and_return(form)
         end
 
-        context "case logs that are owned or managed by your organisation" do
-          before do
-            sign_in user
+        context "a user that is not signed in" do
+          it "does not let you get case log tasklist pages you don't have access to" do
```
dushan-madetech:
```suggestion
        context "a user that is not signed in" do
          it "does not let the user get case log tasklist pages they don't have access to" do
```

otherwise it might read as if a user can block someone else from accessing tasklist pages

dushan-madetech (approved):
left a few suggestions but nothing to block this PR over. Nice stuff here 🚢 it

PR #145, opened 2021-12-03
Titles prefixed with error

MatthewPhelanMadeTech:


app/views/layouts/application.html.erb
```diff
@@ -1,7 +1,15 @@
 <!DOCTYPE html>
 <html lang="en" class="govuk-template">
   <head>
-    <title><%= browser_title(yield(:title)) %></title>
+    <% if @resource.present? && @resource.errors.present? %>
```
baarkerlounger: Would these be better being in their specific pages/layouts?

paulrobertlloyd: Here’s a good example of using a helper, which you can pass it the title and error: https://github.com/DFE-Digital/apply-for-teacher-training/blob/e7c4fa34f1e3fae294a9992d2113d1897c9f82d2/app/helpers/view_helper.rb#L40

You only then need to use that helper on pages which have error handling.

MatthewPhelanMadeTech: @paulrobertlloyd just for clarification - should the title be prefixed with just the word "error" or the whole error message?

paulrobertlloyd: Just the word Error. So for example if you had a validation message on the sign in page, the page title would become:

`Error: Sign in – [Service name] - GOV.UK`

MatthewPhelanMadeTech: okay, fantastic - all set

PR #146, opened 2021-12-03
CLDC-768: Add 2FA for Admin Users

baarkerlounger:
When signing into the ActiveAdmin panel AdminUsers now receive an SMS OTP code via Gov Notify that they need to enter before they're authenticated.

- [x] Require phone numbers for all AdminUsers
- [x] Fix OTP code view
- [x] Use real Gov Notify API key and Template ID
- [x] Add resend code page
- [x] Dasherize URLs

Current flow:
image
OTP code sent by Notify:
image
OTP code prompt:
image
Resend OTP code:
image
image

Successful sign in:
image

### DIFF
--- a/app/views/devise/two_factor_authentication/show.html.erb
+++ b/app/views/devise/two_factor_authentication/show.html.erb
@@ -13,6 +13,7 @@
       <%= f.govuk_number_field :code,
           label: { text: "Security code" },
           width: 5,
+          autocomplete: 'one-time-code',
           autofocus: true
       %>
 
--- a/app/views/devise/two_factor_authentication/show.html.erb
+++ b/app/views/devise/two_factor_authentication/show.html.erb
@@ -1,6 +1,6 @@
 <% content_for :title, "Check your phone" %>
 
-<%= form_with(url: "/admin/two_factor_authentication", html: { method: :put }) do |f| %>
+<%= form_with(url: "/admin/two-factor-authentication", html: { method: :put }) do |f| %>
   <div class="govuk-grid-row">
     <div class="govuk-grid-column-two-thirds">
 
@@ -22,5 +22,5 @@
 <% end %>
 
 <p class="govuk-body">
-  <%= govuk_link_to "Not received a text message?", "#" %>
+  <%= govuk_link_to "Not received a text message?", admin_two_factor_authentication_resend_path %>
 </p>
--- a/app/views/devise/two_factor_authentication/show.html.erb
+++ b/app/views/devise/two_factor_authentication/show.html.erb
@@ -1,0 +1,26 @@
+<% content_for :title, "Check your phone" %>
+
+<%= form_with(url: "/admin/two_factor_authentication", html: { method: :put }) do |f| %>
+  <div class="govuk-grid-row">
+    <div class="govuk-grid-column-two-thirds">
+
+      <h1 class="govuk-heading-l">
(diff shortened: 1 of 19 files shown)


paulrobertlloyd:
Can we CORE customise CORE the CORE text CORE message at all?

app/views/devise/two_factor_authentication/show.html.erb
```diff
@@ -0,0 +1,26 @@
...
+      <h1 class="govuk-heading-l">
+        <%= content_for(:title) %>
+      </h1>
+
+      <p class="govuk-body">We’ve sent you a text message with a security code.</p>
+
+      <%= f.govuk_number_field :code,
+          label: { text: "Security code" },
+          width: 5,
+          autofocus: true
```
paulrobertlloyd:
```suggestion
          width: 5,
          autocomplete: 'one-time-code',
          autofocus: true
```

paulrobertlloyd:
Haven’t tested yet, but looks great from a code point of view.

baarkerlounger:
@paulrobertlloyd if you don't like my content you only have to say so... :stuck_out_tongue: . Have updated the template and screenshot. cc @Dan13L3

app/views/devise/two_factor_authentication/show.html.erb
baarkerlounger: Why would you want that?

paulrobertlloyd: * https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/autocomplete
* https://dev.to/maciejtrzcinski/html-autocomplete-one-time-code-357g

Where supported, makes inputting number a bit quicker.

baarkerlounger: In general sure but I'm not sure that should apply to an SMS OTP code (what help could/should the browser provide there?). That second link you have there seems to be assuming you're browser is on the same device as the OTP code is being sent to which seems like circumventing the intention of the 2FA.

I'd say that falls more into the category of:

> Set autocomplete="off" for sensitive form fields, like credit card number
> Why: So other users of the browser can’t access this saved information

(https://github.com/ankane/secure_rails)

paulrobertlloyd: It only works on mobile browsers, and when a text message is received, it shows the code as a hint above the keyboard so you can select that rather type the numbers:



https://support.apple.com/en-gb/guide/iphone/iphc89a3a3af/ios

paulrobertlloyd: I guess unlike credit cards, this value has a limited lifespan.

paulrobertlloyd (approved):
Love it!

Dan13L3:
@baarkerlounger Thanks - looks great. Just a couple of small things:

- ‘Resend security code’ page

I’d shorten to: ‘If you don’t receive the text message with your security code after a few minutes, you can request a new one.’

- ‘Success’ banner on final screen

I’d change ‘authentication code’ to ‘security code’ to keep it consistent.

baarkerlounger:
@Dan13L3 have updated the resent flash content (and screenshot), have left the resend page text the same as the design guidance as discussed.

Dan13L3:
perfect - thanks @baarkerlounger

dushan-madetech (commented):
🚢 it

PR #148, opened 2021-12-06
CLDC-454-la-validation

kosiakkatrina:
Add local authority validation if rent type is `London affordable rent` or `london living rent`


baarkerlounger (approved):
Nice

PR #150, opened 2021-12-06
CLDC-730: Data Coordinators can edit org details

baarkerlounger:
**Data Coordinator:**
image
image
image

**Data Provider:**
image

### DIFF
--- a/app/views/organisations/edit.html.erb
+++ b/app/views/organisations/edit.html.erb
@@ -19,12 +19,12 @@
       %>
 
       <%= f.govuk_text_field :address_line1,
-        label: { text: "Address line 1 of 2", hidden: true },,
+        label: { text: "Address line 1" },
         autocomplete: "address-line1"
       %>
 
       <%= f.govuk_text_field :address_line2,
-        label: { text: "Address line 2 of 2", hidden: true },
+        label: { text: "Address line 2" },
         autocomplete: "address-line2"
       %>
 
--- a/app/views/organisations/edit.html.erb
+++ b/app/views/organisations/edit.html.erb
@@ -19,12 +19,12 @@
       %>
 
       <%= f.govuk_text_field :address_line1,
-        label: -> { safe_join(["Building and street", " ", tag.span("line 1 of 2", class: "govuk-visually-hidden")]) },
+        label: { text: "Address line 1 of 2", hidden: true },,
         autocomplete: "address-line1"
       %>
 
       <%= f.govuk_text_field :address_line2,
-        label: { text: "Building and street line 2 of 2", hidden: true },
+        label: { text: "Address line 2 of 2", hidden: true },
         autocomplete: "address-line2"
       %>
 
--- a/app/views/organisations/edit.html.erb
+++ b/app/views/organisations/edit.html.erb
@@ -1,0 +1,40 @@
+<% content_for :title, "Change #{@organisation.name}'s details" %>
+
+<% content_for :before_content do %>
+  <%= govuk_back_link(
+    text: 'Back',
+    href: :back,
+  ) %>
+<% end %>
+
+<%= form_for(@organisation, as: :organisation, html: { method: :patch }) do |f| %>
+  <div class="govuk-grid-row">
+    <div class="govuk-grid-column-two-thirds">
+      <h1 class="govuk-heading-l">
+        <%= content_for(:title) %>
+      </h1>
+
+      <%= f.govuk_text_field :name,
+        autocomplete: "name"
+      %>
+
+      <%= f.govuk_text_field :address_line1,
+        autocomplete: "address-line1"
+      %>
+
+      <%= f.govuk_text_field :address_line2,
+        autocomplete: "address-line2"
+      %>
+
+      <%= f.govuk_text_field :postcode,
+        autocomplete: "postal-code"
+      %>
+
+      <%= f.govuk_phone_field :phone,
+        autocomplete: "tel"
+      %>
+
+      <%= f.govuk_submit "Save changes" %>
+    </div>
+  </div>
+<% end %>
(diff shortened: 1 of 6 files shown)


app/views/organisations/show.html.erb
```diff
@@ -7,11 +7,22 @@
...
+            row.value { simple_format(attr[:value].to_s, {}, wrapper_tag: "div") }
+            row.action(visually_hidden_text: 'name', href: edit_organisation_path, html_attributes: { 'data-qa': "change-#{attr[:name]}" })
+          end %>
+        <% else %>
+          <%= summary_list.row do |row|
+            row.key { attr[:name].to_s.humanize }
+            row.value { simple_format(attr[:value].to_s, {}, wrapper_tag: "div") }
+            row.action()
+          end %>
+        <% end %>
```
dushan-madetech, line 24: Should this logic not be in some kind of helper instead of the view?

dushan-madetech (approved):
Left a comment but it looks good. 🚢 it

app/views/organisations/show.html.erb
baarkerlounger: Happy to put it in one if you think I should. My reasoning for not was that doing so would pretty much move the entire view into a helper (since this block is pretty much all it has and it's more presentation than logic), so wouldn't really be isolating anything to unit test that isn't already being tested I don't think, not a strongly held opinion though!

app/views/organisations/edit.html.erb
```diff
@@ -0,0 +1,40 @@
+<% content_for :title, "Change #{@organisation.name}'s details" %>
```
paulrobertlloyd:
```suggestion
<% content_for :title, "Change #{@organisation.name}’s details" %>
```

app/views/organisations/edit.html.erb
```diff
@@ -0,0 +1,40 @@
...
+      <%= f.govuk_text_field :address_line1,
+        autocomplete: "address-line1"
+      %>
+
+      <%= f.govuk_text_field :address_line2,
+        autocomplete: "address-line2"
+      %>
+
+      <%= f.govuk_text_field :postcode,
+        autocomplete: "postal-code"
```
paulrobertlloyd:
```suggestion
        autocomplete: "postal-code",
        width: 10
```

app/views/organisations/edit.html.erb
```diff
@@ -0,0 +1,40 @@
...
+      %>
+
+      <%= f.govuk_phone_field :phone,
+        autocomplete: "tel"
```
paulrobertlloyd:
```suggestion
        label: { text: "Telephone number" },
        autocomplete: "tel",
        width: 20
```

app/views/organisations/edit.html.erb
```diff
@@ -0,0 +1,40 @@
...
+      <h1 class="govuk-heading-l">
+        <%= content_for(:title) %>
+      </h1>
+
+      <%= f.govuk_text_field :name,
+        autocomplete: "name"
+      %>
+
+      <%= f.govuk_text_field :address_line1,
+        autocomplete: "address-line1"
```
paulrobertlloyd, line 23: I haven’t tested this, but something like:

```suggestion
        label: -> { safe_join(["Building and street", " ", tag.span("line 1 of 2", class: "govuk-visually-hidden")]) },
        autocomplete: "address-line1"
```

Output should be to display a label with visible text ‘Building and street’ and hidden text ‘ line 1 of 2’.

app/views/organisations/edit.html.erb
```diff
@@ -0,0 +1,40 @@
...
+      %>
+
+      <%= f.govuk_text_field :address_line2,
+        autocomplete: "address-line2"
```
paulrobertlloyd, line 28: Again, not tested this, but I think this should work:

```suggestion
        label: { text: "Building and street line 2 of 2", hidden: true },
        autocomplete: "address-line2"
```

Output should be a visually hidden label that says ‘Building and street line 2 of 2’

paulrobertlloyd (changes requested):
Just some tweaks to the edit organisation details form.

app/views/organisations/show.html.erb
paulrobertlloyd: Yeah, I think this is fine, at least for now given the page and its current state.

app/views/organisations/edit.html.erb
baarkerlounger: Do you not want a visible label here? It also sort of includes city at the moment?

paulrobertlloyd: Do the addresses that we have, have a standard format. Probably not.

Perhaps the more fail safe version, as much as I hate it, is just to have:

```
Address line 1
[                            ]

Address line 2
[                            ]
```

baarkerlounger: Not currently. I'm not sure if the old CORE system had any rules around how org addresses had to be saved. We'd need to see some data to check that I suppose so for now here it's just Line 1 and Line 2.

app/views/organisations/edit.html.erb
```diff
@@ -0,0 +1,45 @@
...
+      <h1 class="govuk-heading-l">
+        <%= content_for(:title) %>
+      </h1>
+
+      <%= f.govuk_text_field :name,
+        autocomplete: "name"
+      %>
+
+      <%= f.govuk_text_field :address_line1,
+        label: { text: "Address line 1 of 2", hidden: true },,
```
paulrobertlloyd:
```suggestion
        label: { text: "Address line 1" },
```

app/views/organisations/edit.html.erb
```diff
@@ -0,0 +1,45 @@
...
+        autocomplete: "address-line1"
+      %>
+
+      <%= f.govuk_text_field :address_line2,
+        label: { text: "Address line 2 of 2", hidden: true },
```
paulrobertlloyd:
```suggestion
        label: { text: "Address line 2" },
```

baarkerlounger:
@paulrobertlloyd anything left on this?

paulrobertlloyd (approved):
:shipit:

PR #155, opened 2021-12-08
Add lint stage to pipeline

baarkerlounger:
Help ensure Rubocop violations aren't merged so linting the codebase happens automatically and stays easy.

### DIFF
--- a/.github/workflows/pipeline.yml
+++ b/.github/workflows/pipeline.yml
@@ -74,7 +74,22 @@ jobs:
         run: |
           bundle exec rspec
 
+  lint:
+    name: Lint
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
 
+      - name: Set up Ruby
+        uses: ruby/setup-ruby@v1
+        with:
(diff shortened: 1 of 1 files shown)


kosiakkatrina (approved):

.github/workflows/pipeline.yml
```diff
@@ -74,15 +74,30 @@ jobs:
...
 
+      - name: Set up Ruby
+        uses: ruby/setup-ruby@v1
+        with:
+          # runs 'bundle install' and caches installed gems automatically
+          bundler-cache: true
+
+      - name: Rubocop
+        run: |
+          bundle exec rubocop
```
dushan-madetech, line 92: Is it possible to have this do it with the `-A` flag to fix any automatically fixable linting issues?

dushan-madetech (approved):
This is fine with me

.github/workflows/pipeline.yml
baarkerlounger: Don't think a pipeline is the best place for that since it would need to also commit code changes and then re-run. Would work as a local pre-commit hook though?

PR #158, opened 2021-12-09
Validate that password match when updating and show errors if they don't

baarkerlounger:
image

### DIFF
--- a/app/views/users/reset_password.html.erb
+++ b/app/views/users/reset_password.html.erb
@@ -11,6 +11,8 @@
   <%= f.hidden_field :reset_password_token %>
   <div class="govuk-grid-row">
     <div class="govuk-grid-column-two-thirds">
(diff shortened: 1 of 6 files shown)


app/views/users/reset_password.html.erb
```diff
@@ -11,6 +11,8 @@
   <%= f.hidden_field :reset_password_token %>
   <div class="govuk-grid-row">
     <div class="govuk-grid-column-two-thirds">
+      <%= f.govuk_error_summary %>
+
```
kosiakkatrina, line 15: are these new lines rubocops?

kosiakkatrina (approved):
Works for me!

app/views/users/reset_password.html.erb
baarkerlounger: No just looked easier to scan to me

PR #159, opened 2021-12-09
Add errors controller

paulrobertlloyd:
Inspired by [govuk-rails-boilerplate](https://github.com/DFE-Digital/govuk-rails-boilerplate/), moves errors from `public` folder and instead uses an error controller.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -1,8 +1,8 @@
 class ApplicationController < ActionController::Base
   default_form_builder GOVUKDesignSystemFormBuilder::FormBuilder
 
-  def render_not_found_html
-    render file: Rails.root.join("public/404.html"), status: :not_found
+  def render_not_found
+    render "errors/not_found", status: :not_found
   end
 
   def render_not_found_json(class_name, id)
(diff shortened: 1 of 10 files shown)


app/controllers/application_controller.rb
```diff
@@ -2,7 +2,7 @@ class ApplicationController < ActionController::Base
   default_form_builder GOVUKDesignSystemFormBuilder::FormBuilder
 
   def render_not_found_html
-    render file: Rails.root.join("public/404.html"), status: :not_found
+    render "errors/not_found", status: :not_found
```
baarkerlounger, line 5: We could just replace this (and the method below) entirely with the errors controller implementation?

app/controllers/errors_controller.rb
```diff
@@ -0,0 +1,23 @@
+class ErrorsController < ApplicationController
```
baarkerlounger, line 1: Probably need `skip_before_action :verify_authenticity_token` here as well if we're expecting this to handle json (api) requests as well?

paulrobertlloyd: Done!

app/controllers/application_controller.rb
paulrobertlloyd: Tried a few things, and not able to get that working. Might need a hand figuring this out in the new year!

baarkerlounger (approved):

PR #160, opened 2021-12-09
Use ‘log‘ not ‘case log’

paulrobertlloyd:
Removes ‘case logs’ from user-visible parts of the site:

* Updates `/case-logs/*` routes to use `/logs/*`
* Changes references to ‘case log(s)’ to ‘log(s)’

Lots of mention of case logs in the code, comments and API spec, but can decide if we want to address that later.


baarkerlounger (approved):
For consistency might be nice if there's a chance to change the backend references as well but happy to merge this first.

PR #162, opened 2021-12-13
CLDC-492: After housing benefit and/or housing element of UC payment is received, will there be an outstanding amount for basic rent and/or benefit eligible charges?

MatthewPhelanMadeTech:
validate_hbrentshortfall


kosiakkatrina (commented):
Any tests coming for this?

PR #164, opened 2021-12-13
Cldc 650 derived startdate

dushan-madetech:
Deriving the day, month, year from the startdate since we already collect the startdate. This matches the approach taken with mrcdate


kosiakkatrina (approved):
Looks good to me 👍

PR #176, opened 2021-12-21
Iterate over all forms

baarkerlounger:
The form questions change from year to year. The data collection window overlaps by several months to allow for late submissions of data from the previous year. This means that it's necessary for us to be able to run at least 2 versions of the form simultaneously rather than just having a single form definition as before.

This PR moves towards that.


app/controllers/case_logs_controller.rb
```diff
@@ -107,4 +106,9 @@ def api_case_log_params
   def find_resource
     @case_log = CaseLog.find_by(id: params[:id])
   end
+
+  def find_form
+    year = @case_log.year || 2021
```
baarkerlounger: Defaulting to this year here is equivalent to the current behaviour where we hardcoded the form name. We probably need to work out how to default this more dynamically or separate out "About this log" into a separate form, so that this isn't needed until the start date has definitely been completed.

PR #180, opened 2021-12-22
Bump node & fix docker image

baarkerlounger:
To be merged once cloudfoundry/ruby-buildpack#341 has been resolved


smeny:
As discussed, currently the ruby buildpack does not support nodeJS version 16.
One solution is to use the base image ruby:3.0.3-alpine3.14 (Alpine 3.14) which depends on node 14 as well as adding git as a package.

I tested this solution successfully on local.

baarkerlounger:
Closing in favour of https://github.com/communitiesuk/mhclg-data-collection-beta/pull/186. Can re-open bumping node when/if the CloudFoundry buildpack is updated,

PR #181, opened 2021-12-22
Accessible autocomplete for select questions

baarkerlounger:
Without JS:
image

With JS:
image

image


paulrobertlloyd (changes requested):
This looks good, but there are few issues:

1. On first appearance, the first value is selected/shown ('Adur'), when it should be blank.
2. I’m seeing both the autocomplete and select showing in some cases:
Screenshot 2021-12-22 at 15 27 39


Am investigating.

paulrobertlloyd:
2. I think this may be down to Webpacker’s dev environment with inline styles/JS being applied 🙄

paulrobertlloyd (approved):

PR #187, opened 2022-01-04
Tenant's MUST be shown the privacy notice otherwise you can't complete the rest of the form

baarkerlounger:
If the GDPR privacy notice question is answered as no, the About this Log section cannot be completed and the rest of the form cannot be completed until the answer is changed to yes.

### DIFF
--- a/app/models/form/subsection.rb
+++ b/app/models/form/subsection.rb
@@ -30,14 +30,9 @@ class Form::Subsection
 
     qs = applicable_questions(case_log)
     return :not_started if qs.all? { |question| case_log[question.id].blank? }
-    return :completed if qs.all? { |question| case_log[question.id].present? } && !invalidated?(case_log)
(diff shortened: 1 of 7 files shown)


app/models/form/subsection.rb
```diff
@@ -30,11 +30,16 @@ def status(case_log)
 
     qs = applicable_questions(case_log)
     return :not_started if qs.all? { |question| case_log[question.id].blank? }
-    return :completed if qs.all? { |question| case_log[question.id].present? }
+    return :completed if qs.all? { |question| case_log[question.id].present? } && !invalidated?(case_log)
 
     :in_progress
   end
 
+  def invalidated?(case_log)
```
smeny: I think this is a bit too generic, we could go with something like 'can_use_service' since this is what the submission says when refusing the GDPR statement.

smeny (approved):
The new version is definitely clearer 👍🏼

PR #188, opened 2022-01-04
CLDC-819: Move property reference question into About this log section

baarkerlounger:
https://digital.dclg.gov.uk/jira/browse/CLDC-819


smeny (approved):
Does what it says on the tin

PR #189, opened 2022-01-05
Content tweaks for About this log section

paulrobertlloyd:
* Use more recent year for date hint examples
* Use correct title for property reference question

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -395,6 +395,7 @@ PLATFORMS
   arm64-darwin-21
   x86_64-darwin-19
   x86_64-darwin-20
+  x86_64-darwin-21
   x86_64-linux
 
 DEPENDENCIES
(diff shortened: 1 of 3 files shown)


Gemfile.lock
```diff
@@ -395,6 +395,7 @@ PLATFORMS
   arm64-darwin-21
   x86_64-darwin-19
   x86_64-darwin-20
+  x86_64-darwin-21
```
baarkerlounger, line 398: Back to your old mac?

baarkerlounger (approved):

Gemfile.lock
paulrobertlloyd: Don’t ask. Same Mac, but `eventmachine` on another Ruby project was moaning about ARM. Turns out you can run Terminal in Rosetta. How was your Christmas 😂

PR #190, opened 2022-01-05
Cldc 670 property information inferred answers

kosiakkatrina:
Display inferred LA underneath the postcode:
image

Display postcode as not known:
image

### DIFF
--- a/app/models/form/question.rb
+++ b/app/models/form/question.rb
@@ -34,10 +34,9 @@ class Form::Question
 
   def get_inferred_answers(case_log)
     if inferred_answers
-      inferred_answers.keys.map { |x| case_log[x].to_s }
-    else
-      []
+      return filter_inferred_answers(inferred_answers, case_log).keys.map { |x| case_log[x].to_s }
     end
+    []
   end
 
   def read_only?
@@ -83,5 +82,9 @@ private
     else
       raise "Not implemented yet"
     end
+  end
+
+  def filter_inferred_answers(inferred_answers, case_log)
+    inferred_answers.filter { |_key, value| value.all? { |condition_key, condition_value| case_log[condition_key] == condition_value } }
   end
 end
--- a/app/models/form/question.rb
+++ b/app/models/form/question.rb
@@ -34,11 +34,11 @@ class Form::Question
 
     return case_log[id].to_s if case_log[id].present?
 
-    inferred_check_answers_value.present? ? inferred_check_answers_value["value"] : ""
+    has_inferred_check_answers_value?(case_log) ? inferred_check_answers_value["value"] : ""
   end
 
   def get_inferred_answers(case_log)
-    return filter_inferred_answers(inferred_answers, case_log).keys.map { |x| case_log[x].to_s } if inferred_answers
+    return enabled_inferred_answers(inferred_answers, case_log).keys.map { |x| case_log[x].to_s } if inferred_answers
 
     []
   end
@@ -98,7 +98,7 @@ private
     end
   end
 
-  def filter_inferred_answers(inferred_answers, case_log)
+  def enabled_inferred_answers(inferred_answers, case_log)
     inferred_answers.filter { |_key, value| value.all? { |condition_key, condition_value| case_log[condition_key] == condition_value } }
   end
 end
--- a/app/models/form/question.rb
+++ b/app/models/form/question.rb
@@ -2,7 +2,7 @@ class Form::Question
   attr_accessor :id, :header, :hint_text, :description, :questions,
                 :type, :min, :max, :step, :width, :fields_to_add, :result_field,
                 :conditional_for, :readonly, :answer_options, :page, :check_answer_label,
-                :inferred_answers
+                :inferred_answers, :hidden_in_check_answers
 
   def initialize(id, hsh, page)
     @id = id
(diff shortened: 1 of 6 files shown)


app/models/form/question.rb
```diff
@@ -41,6 +53,16 @@ def enabled?(case_log)
     conditional_on.map { |condition| evaluate_condition(condition, case_log) }.all?
   end
 
+  def hidden_in_check_answers?
+    hidden_in_check_answers
+  end
```
kosiakkatrina, line 58: Could this just be attr accessor?

config/forms/2021_2022.json
```diff
@@ -1121,14 +1144,20 @@
...
                   "header": "",
                   "hint_text": "",
                   "type": "text",
-                  "width": 5
+                  "width": 5,
+                  "inferred_answers": { "la": { "is_la_inferred": true } },
+                  "inferred_check_answers_value": {
+                    "condition": { "postcode_known": "No" },
+                    "value": "Not known"
+                  }
```
kosiakkatrina, line 1160: This is the only actual change, the rest is formatting..

dushan-madetech (approved):
I don't think there any major/urgent changes necessary so I'm approving it, though feel free to do any improvements/refactors before merging 👍

app/models/form/question.rb
baarkerlounger: I guess this way you get the `?` boolean return naming convention? Although I suppose current implementation doesn't _definitely_ return a boolean?

app/models/form/question.rb
```diff
@@ -75,4 +97,8 @@ def evaluate_condition(condition, case_log)
       raise "Not implemented yet"
     end
   end
+
+  def filter_inferred_answers(inferred_answers, case_log)
```
baarkerlounger: Think it might be better to name this after what it filters by since it isn't a generic filter method? Something that tells you what is actually returned along the lines of:

`enabled_inferred_answers` ?

app/models/form/question.rb
```diff
@@ -41,6 +53,16 @@ def enabled?(case_log)
     conditional_on.map { |condition| evaluate_condition(condition, case_log) }.all?
   end
 
+  def hidden_in_check_answers?
+    hidden_in_check_answers
+  end
+
+  def has_inferred_check_answers_value?(case_log)
+    return inferred_check_answers_value["condition"].values[0] == case_log[inferred_check_answers_value["condition"].keys[0]] if inferred_check_answers_value.present?
```
baarkerlounger, line 61: Is this definitely always a hash with only a single key/value?

app/models/form/subsection.rb
```diff
@@ -62,4 +62,8 @@ def answered_questions(case_log)
   def unanswered_questions(case_log)
     applicable_questions(case_log) - answered_questions(case_log)
   end
+
+  def displayed_to_user?(case_log, question)
```
baarkerlounger, line 66: Is this a little confusing? The question is still displayed to the user right? Just not the answer in the check answers page?

config/forms/2021_2022.json
baarkerlounger: Could we add this to the documentation? Possibly also to the JsonSchema so it validates (not sure if anyone is using that/finding it useful though?)

app/models/form/subsection.rb
kosiakkatrina: So the 'what is the postcode?' question would not be displayed to the user if postcode known == "No". This would return false for the postcode, instead we would display 'postcode' question in the check answers page because `has_inferred_check_answers_value?` on line 55 would return true

app/models/form/question.rb
kosiakkatrina: It is right now, cause there's only one such occurrence

app/models/form/subsection.rb
baarkerlounger: The only question `question.hidden_in_check_answers?` would currently return `True` for is the `Do you know the postcode?` question right? And that question is always displayed to the user?

kosiakkatrina: Oh sorry, I thought you were raising a different question.
Let's chat about it. I'm getting muddled in if statements

app/models/form/question.rb
baarkerlounger: I think that's fine then but worth documenting that assumption in the definition schema doc in the readme?

app/models/form/subsection.rb
```diff
@@ -52,7 +52,7 @@ def answered_questions_count(case_log)
   end
 
   def applicable_questions(case_log)
```
baarkerlounger, line 54: It looks like this method is only used by check answers page. Might be worth renaming it to make that clearer?

kosiakkatrina: Hmm, it's also used in `status` method by the looks of it, which is used in the `tasklist`

baarkerlounger: In that case does this change break that? (Should maybe have a test for it?). I'm thinking for a section that includes a question like this (hidden from check answers), will the status be set correctly?

kosiakkatrina: Which now needs to be fixed, because it shows up as `in progress` even if it's `completed` if postcode is selected as not known. Thanks for flagging!

kosiakkatrina: This is now described in separate tickets (CLDC-830,CLDC-831,CLDC-832), so will merge this branch

PR #192, opened 2022-01-06
CLDC-827: Add custom guidance partials

baarkerlounger:
- [x] Form schema supports custom guidance for questions between the page header and question header
- [x] Integration test
- [x] Unit tests for each question type

image
image

Given that the HTML here is potentially quite large and this is likely to be functionality present in multiple questions it seems better to break this out into actual partials and reference them than to try to include the html in the JSON directly.


config/initializers/default_form_builder.rb
```diff
@@ -0,0 +1 @@
+Rails.application.config.action_view.default_form_builder = GOVUKDesignSystemFormBuilder::FormBuilder
```
baarkerlounger, line 1: This is moving here so that it is loaded for view specs which don't load controllers. I think that allows us to better unit test our templates.

app/views/form/guidance/_what_counts_as_income.html.erb
```diff
@@ -0,0 +1,25 @@
...
+  <p class="govuk-body">Don’t include:</p>
+  <ul class="govuk-list govuk-list--bullet">
+    <li>National Insurance (NI) contributions and tax</li>
+    <li>housing benefit</li>
+    <li>child benefit</li>
+    <li>council tax support</li>
+  </ul>
+
+  </div>
+</details>
```
paulrobertlloyd: Does this have to be HTML as it gets rendered as JSON? Be nice if could use a rails component(s) here, if possible.

paulrobertlloyd:
Smart solution! Just wondering if we can use `govuk_details` component rather than pure HTML for guidance.

app/views/form/guidance/_what_counts_as_income.html.erb
baarkerlounger: What do you mean it gets rendered as JSON? How so?

paulrobertlloyd: I might have been overthinking my comment! We can do the following, no?

```erb
<%= govuk_details(summary_text: 'What counts as income?') do %>
  <p class="govuk-body">You should include any income from:</p>
  <ul class="govuk-list govuk-list--bullet">
    <li>employment</li>
    <li>pensions</li>
    <li>Universal Credit</li>
  </ul>

  <p class="govuk-body">Don’t include:</p>
  <ul class="govuk-list govuk-list--bullet">
    <li>National Insurance (NI) contributions and tax</li>
    <li>housing benefit</li>
    <li>child benefit</li>
    <li>council tax support</li>
  </ul>
<% end %>
```

baarkerlounger: Ah yes sorry, the rendering thing threw me but we should definitely use the component. Have updated.

config/forms/2021_2022.json
```diff
@@ -1796,18 +1796,21 @@
...
                   "type": "radio",
                   "answer_options": {
-                    "0": "Yes",
-                    "1": "No",
-                    "2": "Tenant prefers not to say"
+                    "0": "Yes – the household has a weekly income",
+                    "1": "Yes – the household has a monthly income",
+                    "2": "Yes – the household has a yearly income",
+                    "divider_a": true,
+                    "3": "Tenant prefers not to say"
```
kosiakkatrina, line 1813: We got some inferred variables that rely on existing values like 
`self.incref = 1 if net_income_known == "Prefer not to say"` in `models/case_log.rb:217` (of course this would've failed before too, but might be a good place to fix)
Also same file lines `319` `332`

kosiakkatrina (approved):
This looks good, very nice PR!
Just left one comment, but can get merged in after that is fixed

config/forms/2021_2022.json
baarkerlounger: Good catch. Have updated and made it an enum so you can't save unexpected values.

app/models/case_log.rb
```diff
@@ -316,9 +317,11 @@ def mandatory_fields
       dynamically_not_required << "tenancyother"
     end
 
-    if net_income_known == "No"
+    if net_income_known == "Tenant prefers not to say"
       dynamically_not_required << "earnings"
       dynamically_not_required << "incfreq"
+    else
+      dynamically_not_required << "incref"
```
kosiakkatrina, line 324: Ooh, very nice!

PR #193, opened 2022-01-06
WIP: Cldc 600 reduce user burden 

dushan-madetech:


smeny (approved):
Everything looks fine to me but I don't really if there is a better way to add new fields given the existing structure. So you might want a second opinion from @baarkerlounger or @kosiakkatrina

spec/features/form/page_routing_spec.rb
```diff
@@ -72,4 +72,28 @@
       expect(page).to have_current_path("/logs/#{id}/property-wheelchair-accessible")
     end
   end
+
+  context "routing based on start date period", js: true do
```
baarkerlounger, line 76: I think this functionality is well tested here already https://github.com/communitiesuk/mhclg-data-collection-beta/blob/main/spec/requests/form_controller_spec.rb#L55 so these feature ones might not be necessary?

spec/requests/case_log_controller_spec.rb
```diff
@@ -240,6 +240,31 @@
...
+          FactoryBot.create(
+            :case_log,
+            :about_completed,
+            startdate: Time.zone.local(2022, 10, 10),
+            owning_organisation: organisation,
+            managing_organisation: organisation,
+          )
+        end
+
+        it "has a joint tenancy page that can be visited if the case log is in the 2022-2023 period" do
```
baarkerlounger, line 256: Same with this spec. Not sure it's adding much over the one that's there already? Also I think it would probably belong in `form_controller_spec` (form navigation) rather than `case_log_controller_spec` (case log data API)

baarkerlounger (approved):
Didn't realise we were starting work on next years form already. Couple of comments on tests but nothing blocking.

app/helpers/tasklist_helper.rb
```diff
@@ -29,7 +29,7 @@ def first_page_or_check_answers(subsection, case_log)
     path = if subsection.is_started?(case_log)
              "case_log_#{subsection.id}_check_answers_path"
            else
-             "case_log_#{subsection.pages.first.id}_path"
+             "case_log_#{subsection.applicable_questions(case_log).first.page.id}_path"
```
baarkerlounger, line 32: Should there be a test for the case this change is covering? Or have I missed one?

baarkerlounger:
I wonder if creating next year's form now (before completing this year's form) is going to make it harder to keep up to date/correct since every change PR now needs to be made against both form definition files.

It might be easier to complete this years and then do next years changes as a single PR?

spec/features/form/page_routing_spec.rb
dushan-madetech: This was originally written prior to those tests I'm happy to remove since I think this is also covered by some request specs

spec/requests/case_log_controller_spec.rb
dushan-madetech: I think there's a slight difference where one is checking the difference in headers for the same question whereas this is checking the difference in presence of entire pages in a section. Happy to remove/move this though

baarkerlounger: In both cases the behaviour actually being checked is that the correct form for that year is being routed to though right?

app/helpers/tasklist_helper.rb
dushan-madetech: I think this was originally covered by the tests added in this PR - I did recently have to rebase it and update the tests due to all the changes that happened while I was away. It's not breaking any existing tests at the moment so I'd be willing to say that's a promising sign

dushan-madetech:
> I wonder if creating next year's form now (before completing this year's form) is going to make it harder to keep up to date/correct since every change PR now needs to be made against both form definition files.
> 
> It might be easier to complete this years and then do next years changes as a single PR?

@baarkerlounger I think I'm inclined to agree with you - I'll have a look at splitting the ticket into separate parts so the that the 2021-2022 part could be shipped now

spec/requests/case_log_controller_spec.rb
dushan-madetech: Yep, that's a fair point. Happy to remove these in that case

app/helpers/tasklist_helper.rb
baarkerlounger: Might be worth adding a unit test for the specific edge case here https://github.com/communitiesuk/mhclg-data-collection-beta/blob/main/spec/helpers/tasklist_helper_spec.rb#L44 but not a blocker.

PR #195, opened 2022-01-06
Content and design audit (Property information)

paulrobertlloyd:
Content and design issues found during audit of Property information section. Any common issues corrected too.

### DIFF
--- a/app/models/constants/case_log.rb
+++ b/app/models/constants/case_log.rb
@@ -4,7 +4,7 @@ module Constants::CaseLog
     "Yes - removal of the spare room subsidy" => 4,
     "Yes - both the benefit cap and the removal of the spare room subsidy" => 6,
     "No" => 2,
-    "Do not know" => 3,
+    "Don’t know" => 3,
     "Prefer not to say" => 100,
   }.freeze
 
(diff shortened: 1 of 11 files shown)


app/models/constants/case_log.rb
```diff
@@ -4,15 +4,15 @@ module Constants::CaseLog
     "Yes - removal of the spare room subsidy" => 4,
     "Yes - both the benefit cap and the removal of the spare room subsidy" => 6,
     "No" => 2,
-    "Do not know" => 3,
+    "Don’t know" => 3,
```
baarkerlounger, line 7: I think this might be worth an ADR or some sort of decision record because I think we've changed this back and forth a couple of times now?

baarkerlounger (approved):
@Dan13L3 have you reviewed this?

Dan13L3:
@baarkerlounger Looks good to me!

app/models/constants/case_log.rb
paulrobertlloyd: @Dan13L3 @baarkerlounger GDS style guide says:

> Avoid negative contractions like can’t and don’t. Many users find them harder to read, or misread them as the opposite of what they say. Use cannot, instead of can’t.

Though people have differing opinions on this, suggesting that `don’t` is one of those words that people recognise/understand more than `do not` (which also sounds more forceful). You said use `don’t`, but happy to go either way!

Dan13L3: @paulrobertlloyd Yeah, there's always a bit of a debate around when to use contractions, so I tend to go on a case-by-case basis. 'Don't' seems widely understood, whereas things like 'should've' / 'would've' can cause problems and are often tricky to scan. We've not seen any issues with it in testing so far, but it's one we can always keep an eye on...

PR #196, opened 2022-01-07
CDLC-783: Errors when a postcode is known but not provided

smeny:
The scope of this change is to add a validation to prevent a property postcode to be saved empty when the "postcode is known" question is answered with "Yes".

* Add a validation with an error message
* Add a header for the postcode field
* Update the related tests
* Do not clear the LA when the postcode lookup hasn't returned any data

### DIFF
--- a/app/models/case_log.rb
+++ b/app/models/case_log.rb
@@ -222,7 +222,9 @@ private
     if property_postcode.blank? || postcode_known == "No"
       reset_location_fields
     else
-      self.la = get_la(property_postcode)
+      inferred_la = get_inferred_la(property_postcode)
+      self.is_la_inferred = inferred_la.present?
+      self.la = inferred_la if inferred_la.present?
     end
     if property_postcode.present?
       self.postcode = UKPostcode.parse(property_postcode).outcode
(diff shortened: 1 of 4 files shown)


app/models/case_log.rb
```diff
@@ -254,15 +256,12 @@ def get_totadult
     end
   end
 
-  def get_la(postcode)
+  def get_inferred_la(postcode)
     postcode_lookup = nil
     Timeout.timeout(5) { postcode_lookup = PIO.lookup(postcode) }
```
baarkerlounger, line 261: Am I right in thinking this API call still happens on every `save!`? Might be better to check whether `postcode` is a field that's changed as a condition for sending it?

baarkerlounger (approved):

app/models/case_log.rb
```diff
@@ -254,15 +256,12 @@ def get_totadult
...
     if postcode_lookup && postcode_lookup.info.present?
-      self.is_la_inferred = true
```
kosiakkatrina, line 261: I think this might be needed (probably line 265 too), cause if we consider this situation:
1. Enter a valid postcode
2. LA gets inferred
3. Go back and change `postcode_known` to `No`

I don't think that LA would get reset

kosiakkatrina: Hmm, never mind that. It actually gets reset 👀

kosiakkatrina (approved):
Nice one! Seems to work as expected!

app/models/case_log.rb
smeny: This private method only gets called when this condition is false:
`property_postcode.blank? || postcode_known == "No"`

So yeah we end up making a call every time the field is saved with a postcode populated.
Is there an easy way in AR to detect that an attribute has changed (other than manually loading from DB)?

baarkerlounger: Yes you can do something like:

```
before_save :infer_postcode, if: postcode_changed?
```

smeny:
Merging this and will tackle the API called on each save with the next set of changes.

PR #198, opened 2022-01-07
CLDC-825: Support suffixes and prefixes for numeric fields

baarkerlounger:
- [x] Income frequency is derived rather than asked
- [x] Prefix and suffix can be added to JSON and is displayed
- [x] Prefix/suffix/header updates based on previous answer
- [x] Readme is updated

Weekly:
image
image

Monthly:
image
image

Yearly:
image
image


app/models/case_log.rb
```diff
@@ -215,7 +215,16 @@ def set_derived_fields
...
+    case net_income_known
+    when "Yes – the household has a weekly income"
+      self.incfreq = "Weekly"
+    when "Yes – the household has a monthly income"
+      self.incfreq = "Monthly"
+    when "Yes – the household has a yearly income"
+      self.incfreq = "Yearly"
+    when "Tenant prefers not to say"
+      self.incref = 1
+    end
```
kosiakkatrina, line 227: I don't know if it'd look better or not, but what if we have a mapping instead?
```
INCREF_MAPPINGS =  {"Yes – the household has a weekly income" : "Weekly",  "Yes – the household has a monthly income": "Monthly" ...}

self.incfrec = INCREF_MAPPINGS[net_income_known]
```

baarkerlounger: Happy to change but note there's two different fields being set there depending on the case (`incref` and `incfreq` so might not get rid of that much)

kosiakkatrina: Oh, yeah, that's fair. I don't have a strong preference

kosiakkatrina (approved):
Looks good! Just one thing that is not part of this PR but should probably be fixed. We're saving the amount as integers, so £34.15 would round to £34

baarkerlounger:
@kosiakkatrina that's expected isn't it? We only want to the nearest pound (although that's no longer mentioned in the question text, cc. @Dan13L£)

kosiakkatrina:
> @kosiakkatrina that's expected isn't it? We only want to the nearest pound (although that's no longer mentioned in the question text, cc. @dan13l£)

Oh, I'm not sure anymore..

PR #199, opened 2022-01-07
CLDC-828: Total field should only be a progressive enhancement if JS is enabled

baarkerlounger:
- [x] Total charge field is hidden by default and only shown if JS is enabled
- [x] Total charge value is always calculated on the backend


dushan-madetech (approved):
looks ready for 🚢 ping to me

PR #200, opened 2022-01-07
Cldc 705 cross section navigation

kosiakkatrina:
Add breadcrumbs

image

image


app/views/form/_check_answers_breadcrumb.html.erb
```diff
@@ -0,0 +1,11 @@
+<div class="govuk-breadcrumbs">
+  <ol class="govuk-breadcrumbs__list">
+    <li class="govuk-breadcrumbs__list-item">
+      <a class="govuk-breadcrumbs__link" href="/logs">Logs</a>
+    </li>
+    <li class="govuk-breadcrumbs__list-item">
+      <a class="govuk-breadcrumbs__link" href=" /logs/<%= @case_log.id %>">Log</a>
```
baarkerlounger: Could possibly DRY this into a single breadcrumbs partial with a variable?

baarkerlounger (approved):

app/views/form/_check_answers_breadcrumb.html.erb
paulrobertlloyd: There’s also the GOV.UK Components helper we can use for breadcrumbs: https://govuk-components.netlify.app/components/breadcrumbs/

kosiakkatrina: That is actually quite perfect! Changed that now, let me know if it looks okay!

db/schema.rb
```diff
@@ -162,9 +162,9 @@
     t.string "why_dont_you_know_la"
     t.integer "unitletas"
     t.integer "builtype"
+    t.datetime "property_void_date"
```
baarkerlounger: Doesn't look like this should really be here since there's no migration?

kosiakkatrina: Yeah, I'll move it back

baarkerlounger (approved):

app/views/case_logs/edit.html.erb
```diff
@@ -3,6 +3,8 @@
 <%= turbo_frame_tag "case_log_form", target: "_top" do %>
   <div class="govuk-grid-row">
     <div class="govuk-grid-column-two-thirds-from-desktop">
+      <% breadcrumbs = { "Logs" => "/logs", content_for(:title) => "" } %>
+      <%= govuk_breadcrumbs(breadcrumbs: breadcrumbs) %>
```
paulrobertlloyd, line 7: The breadcrumbs should appear directly after the phase banner, resulting code looking something like this:

```html
    <div class="govuk-phase-banner">…</div>

    <div class="govuk-breadcrumbs">…</div>

    <main class="govuk-main-wrapper govuk-main-wrapper--auto-spacing" id="main-content" role="main">…</main>
</div>
```

PR #202, opened 2022-01-10
Don’t require JS to create new log

paulrobertlloyd:
Use `govuk_button_to` to use form submission rather than rely on a link which uses Turbolinks.

### DIFF
--- a/app/views/case_logs/index.html.erb
+++ b/app/views/case_logs/index.html.erb
@@ -6,7 +6,7 @@
       <%= content_for(:title) %>
     </h1>
 
-    <%= govuk_button_link_to "Create new log", case_logs_path, method: :post %>
+    <%= govuk_button_to "Create new log", case_logs_path %>
 
     <% if @in_progress_case_logs.present? %>
       <%= render partial: "log_list", locals: { case_logs: @in_progress_case_logs, title: "Logs you need to complete", date_title: "Last Changed" } %>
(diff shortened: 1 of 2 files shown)


db/schema.rb
```diff
@@ -162,9 +162,9 @@
     t.string "why_dont_you_know_la"
     t.integer "unitletas"
     t.integer "builtype"
+    t.datetime "property_void_date"
     t.bigint "owning_organisation_id"
     t.bigint "managing_organisation_id"
-    t.datetime "property_void_date"
```
paulrobertlloyd: Grrr, is this a problem, or should I remove from this PR?

baarkerlounger: Preferable would be to remove from this PR I think and check whether it's there on a fresh `rake db:drop; rake db:create; rake db:migrate`. i.e. is the repo schema out of date or is this a local inconsistency.

baarkerlounger:
Does this fix the issue on Safari?

paulrobertlloyd:
> Does this fix the issue on Safari?

Yup! Although when I disabled JS, it also disabled CSS, but I suspect that is a dev environment issue.

baarkerlounger (approved):

db/schema.rb
paulrobertlloyd: Still there when do the above, but have removed change from this PR.

PR #203, opened 2022-01-10
Migrate webpacker to esbuild

baarkerlounger:
**What is Esbuild?**
tl:dr an alternative to webpack, i.e. a Javascript bundler written in Go that's significantly faster. https://blog.logrocket.com/fast-javascript-bundling-with-esbuild/

This PR migrates us from using Webpacker to using Esbuild. See this for more information on the general direction of bundling JS in rails https://world.hey.com/dhh/rails-7-will-have-three-great-answers-to-javascript-in-2021-8d68191b.

The advantages are primarily speed: (Test suite for example runs in ~44 s vs ~90s) and it's a lot less code.

Downsides are it may be less well tested/documented, particularly across Gov repos and workflows are slightly different.

**Do we need this at all?**
Short answer: no. It's really just a case of which tool we prefer to work with and get the most benefit from.

Edit: While we still don't need it Webpacker is now officially considered "retired" by Rails upstream community so moving to at least `js-bundling` seems worthwhile long term. https://github.com/rails/webpacker#webpacker-has-been-retired-

**Caveats:**

This ditches Babel - possibly that means if our stimulus controllers use ES6 it may not work correctly on IE (which is unsupported by MS but we know from server logs some users are still on). Probably need to test what the experience would be like there. https://world.hey.com/dhh/modern-web-apps-without-javascript-bundling-or-transpiling-a20f2755


app/assets/stylesheets/_accessible-autocomplete.scss
```diff
@@ -1,16 +1,16 @@
 .autocomplete__wrapper,
 .autocomplete__input,
 .autocomplete__hint {
-  font-family: $govuk-font-family;
+  font-family: govuk-font-family;
```
paulrobertlloyd: Why this change? This Sass variable should update the font stack to the one used by the design system (`"GDS Transport", arial, sans-serif;`). I’d be wary of changing it.

app/assets/stylesheets/_related-navigation.scss
```diff
@@ -1,23 +1,23 @@
 .app-related-navigation {
-  @include govuk-text-colour;
+  @include all.govuk-text-colour;
```
paulrobertlloyd, line 2: Ooooh, what’s this? Is this syntax specific to the new bundling tool?

app/assets/stylesheets/_tab-navigation.scss
```diff
@@ -37,25 +37,25 @@
...
-  @include govuk-link-style-no-underline;
-  @include govuk-typography-weight-bold;
+  @include all.govuk-link-common;
+  @include all.govuk-link-style-no-visited-state;
+  @include all.govuk-link-style-no-underline;
+  @include all.govuk-typography-weight-bold;
 
   &:not(:focus):hover {
-    color: $govuk-link-colour;
+    color: govuk-link-colour;
```
paulrobertlloyd: Again, this seems incorrect. We may end up using other components in the GOV.UK ecosystem, which would rely on these variables.

paulrobertlloyd (commented):
This is all very curious, need to do some reading up!

app/assets/stylesheets/_related-navigation.scss
baarkerlounger: Module specific syntax. @use vs @import https://sass-lang.com/documentation/at-rules/use

app/assets/stylesheets/_accessible-autocomplete.scss
baarkerlounger: You're right this was a mistake, should be fixed now though.

dushan-madetech:
I have a few questions:

1) How easy would it be to revert to using webpacker if we try this and find that it doesn't work for us?

2) Should we document this swapping in an ADR? I'm asking because I like the PR description and want to it be visible without having to dig up this PR down the line.

3) Since this isn't commonly used across Gov repos is there some way we can document/share our experiences with this so other Gov projects can benefit from our experiences?

dushan-madetech (commented):
I have a few questions:

1) How easy would it be to revert to using webpacker if we try this and find that it doesn't work for us?

2) Should we document this swapping in an ADR? I'm asking because I like the PR description and want to it be visible without having to dig up this PR down the line.

3) Since this isn't commonly used across Gov repos is there some way we can document/share our experiences with this so other Gov projects can benefit from our experiences?

dushan-madetech (approved):
Left some general questions but since this currently improves our test suite speed by reducing build times I'm happy to give this a go

kosiakkatrina (approved):

baarkerlounger:
It seems ES5/IE 11 isn't currently supported and there are no immediate plans to support it so this is unlikely to be an option for at the moment. https://github.com/evanw/esbuild/issues/297.

Closing this for now.

paulrobertlloyd:
Farewell little buddy #203, you had a good run! 👋

paulrobertlloyd:
Do any of the new  asset compilation options in Rails 7 help, or still early days for those?

baarkerlounger:
@paulrobertlloyd :point_up: for your evening reading enjoyment :stuck_out_tongue:

PR #206, opened 2022-01-10
CLDC-883: Clear dependent answers when the dependency value changes

baarkerlounger:
https://digital.dclg.gov.uk/jira/browse/CLDC-883

**Bugfixes:**

1. When you answer a question 1 that then routes you to another question 2, answer that second question (2), then go back to question 1 and change your answer such that you would no longer be routed to question 2, we weren't clearing the saved answer to question 2 (no longer valid)
2. Fixed start date incoming from bulk upload

**Design changes:**

1. Rather than hard code which fields are required and which aren't based on the answers given, we now derive it from the JSON since that already contains all the required information.

This also removes a potential source of bugs where the hardcoding doesn't match the JSON, and makes the test suite easier to work with since the test JSON is now less dependent on the model fields.

**Caveats to watch for:**
1. Since this PR clears answers for questions whose page you wouldn't currently be routed to, going back and removing "About this log" section answers would currently re-disable all other sections and therefore also clear any answers saved for those sections (which may not be what we want). We should probably consider how we want to handle that case. cc. @paulrobertlloyd


paulrobertlloyd:
> Since this PR clears answers for questions whose page you wouldn't currently be routed to, going back and removing "About this log" section answers would currently re-disable all other sections and therefore also clear any answers saved for those sections

@baarkerlounger This shouldn’t be an issue once we prevent you from changing answers to (most) questions in About this log, right?

baarkerlounger:
@paulrobertlloyd depends if it's "most" or "all" I think. In either case it may be the desired behaviour (since changing "About this log" *does* invalidate large parts), but whether it's what users would expect I don't know.

paulrobertlloyd:
So the only questions you should be able to change in About this log after you have completed it are:

* Tenant/purchaser code
* Property reference

These 2 values are independent of any other question, so we should be good, right?

baarkerlounger:
@paulrobertlloyd sounds like it. Think we'll just need to remember to consider this when designing how that works.

paulrobertlloyd (approved):
Haven’t tested, but let’s live dangerously. Seems like a better solution.

dushan-madetech (approved):
Looks good to me from a code perspective, 🚢 it

kosiakkatrina (approved):

PR #207, opened 2022-01-10
CDLC-504: Introduce regexp validation of property postcodes

smeny:
Changes:
* Split model changes into conditional callbacks called before validation
* Changes tests to support new postcode validation rules
* Validate postcodes against agreed regular experession

### DIFF
--- a/app/models/case_log.rb
+++ b/app/models/case_log.rb
@@ -251,7 +251,6 @@ private
     self.postcod2 = UKPostcode.parse(property_postcode).incode
   end
 
-  # Called when the postcode is not known
   def reset_location_fields!
     if is_la_inferred == true
       self.la = nil
--- a/app/models/case_log.rb
+++ b/app/models/case_log.rb
@@ -35,7 +35,9 @@ class CaseLog < ApplicationRecord
   default_scope -> { kept }
 
   validates_with CaseLogValidator
-  before_save :update_status!
+  before_validation :update_status!
+  before_validation :process_postcode_changes!, if: :property_postcode_changed?
+  before_validation :reset_location_fields!, unless: :postcode_known?
 
   belongs_to :owning_organisation, class_name: "Organisation"
   belongs_to :managing_organisation, class_name: "Organisation"
@@ -162,6 +164,10 @@ class CaseLog < ApplicationRecord
     status == "in_progress"
   end
 
+  def postcode_known?
+    postcode_known == "Yes"
+  end
+
   def weekly_net_income
     return unless earnings && incfreq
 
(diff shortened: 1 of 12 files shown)


app/models/case_log.rb
```diff
@@ -35,7 +35,9 @@ class CaseLog < ApplicationRecord
   default_scope -> { kept }
 
   validates_with CaseLogValidator
-  before_save :update_status!
+  before_validation :update_status!
```
baarkerlounger: Should this not still happen `before_save`? If validation fails we wouldn't have expected the status to change?

app/models/case_log.rb
```diff
@@ -219,25 +225,32 @@ def set_derived_fields
...
     self.totadult = get_totadult
   end
 
+  def process_postcode_changes!
+    return if property_postcode.blank?
+
+    self.postcode_known = "Yes"
+    inferred_la = get_inferred_la(property_postcode)
+    self.is_la_inferred = inferred_la.present?
+    self.la = inferred_la if inferred_la.present?
```
baarkerlounger, line 249: Is the `if` clause needed here? For empty string case?

app/models/validations/property_validations.rb
```diff
@@ -35,8 +37,10 @@ def validate_unitletas(record)
   end
 
   def validate_property_postcode(record)
-    if record.postcode_known == "Yes" && record.property_postcode.blank?
-      record.errors.add :property_postcode, "Enter a postcode in the correct format, for example AA1 1AA"
+    postcode = record.property_postcode
+    if record.postcode_known == "Yes" && (postcode.blank? || !postcode.match(POSTCODE_REGEXP))
```
baarkerlounger, line 41: I _think_ we don't want the blank check here so that users aren't blocked on this question? (i.e. they can skip and come back to it?)

spec/features/form/accessible_autocomplete_spec.rb
```diff
@@ -39,7 +39,7 @@
   end
 
   it "has the correct option selected if one has been saved" do
-    case_log.update!(property_postcode: nil, la: "Oxford")
+    case_log.update!(postcode_known: "No", la: "Oxford")
```
baarkerlounger, line 42: If `postcode_known == "No"` doesn't `la` now get reset? And in that case is this working correctly?

app/models/case_log.rb
smeny: Good point, this needs to be changed.

app/models/case_log.rb
smeny: We want to keep the previously set LA if the new postcode doesn't infer a new one.

baarkerlounger: That seems counter-intuitive? This only runs if the `postcode` value has changed and if that's changed surely the previously inferred `la` value is likely to be incorrect?

app/models/validations/property_validations.rb
smeny: I don't think they are, at least this was working when I tested.

spec/features/form/accessible_autocomplete_spec.rb
smeny: The LA is not cleared if it was not inferred. Otherwise, we would lose a previously set value.

app/models/validations/property_validations.rb
baarkerlounger: The use case I'm thinking of is:

1. Do you know the postcode answered as "Yes" (I know I have it, I just don't have it in front of me right now)
2. Don't fill in the postcode just click save and continue (I'll go look it up in a minute when I've filled out the rest of the section)
3. Am I now blocked on postcode by this validation?

smeny: I think this is the behaviour we expect as defined in[ CLDC-783](https://digital.dclg.gov.uk/jira/browse/CLDC-783)

app/models/case_log.rb
smeny: It could be incorrect, but the alternative would be removing an answer which we added previously. When doing some tests with Kat, it is weird when the LA suddenly disappears when you change the postcode. Given we can answer any question, in any order, I don't think we should ever delete something entered. The case where the postcode is valid, but can't be inferred, while the user also previously selected the wrong LA sounds quite far fetched to me. Happy to chat :)

app/models/validations/property_validations.rb
baarkerlounger: Looks like this is agreed now so will mark this as resolved. Might be worth updating this https://github.com/communitiesuk/mhclg-data-collection-beta/blob/main/docs/adr/adr-007-data-validations.md to reflect that.

app/models/case_log.rb
```diff
@@ -36,6 +36,9 @@ class CaseLog < ApplicationRecord
 
   validates_with CaseLogValidator
   before_save :update_status!
+  before_validation :process_postcode_changes!, if: :property_postcode_changed?
+  before_validation :reset_location_fields!, unless: :postcode_known?
+  before_validation :set_derived_fields!
```
baarkerlounger, line 41: Now that these are set before validation we likely don't need to remove them from the fields checked by the status update (i.e. they don't need to be in the auto generated list now). Not important to do here though.

PR #208, opened 2022-01-10
Remove all ampersands from the service

paulrobertlloyd:
It does what it says on the tin!

(Dan says we should keep it for the department name)


Gemfile
```diff
@@ -19,7 +19,7 @@ gem "bootsnap", ">= 1.4.4", require: false
 gem "govuk-components"
 # GOV UK component form builder DSL
 gem "govuk_design_system_formbuilder"
-# Turbo & Stimulus
+# Turbo and Stimulus
```
baarkerlounger, line 22: :joy: Thorough!

paulrobertlloyd: Always.

config/initializers/active_admin.rb
```diff
@@ -201,7 +201,7 @@
   #
   # config.create_another = true
 
-  # == Register Stylesheets & Javascripts
```
baarkerlounger: I would revert the ones in this file so as not to cause unnecessary conflicts when upgrading versions.

paulrobertlloyd: Good idea! Done.

baarkerlounger (approved):

PR #211, opened 2022-01-12
Point to custom devise

baarkerlounger:
Screenshot 2022-01-11 at 16 43 40

Upstream devise currently has this incorrectly capitalised string in it's error messages. A fix was previously merged into a V5 release candidate branch (https://github.com/heartcombo/devise/pull/4834) but development on that branch appears to have stalled in 2019 and doesn't look likely to be picked up again. A new PR has been opened for it against current main: https://github.com/heartcombo/devise/pull/5454.

Since we were already pointing to a fork of Devise to make it compatible with Turbo until upstream has full support (see https://github.com/heartcombo/devise/pull/5340) we may as well point to a fork that contains both fixes. Which is what this PR does. This also means we can ensure the fork branch stays up to date with upstream.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -35,7 +35,7 @@ gem "roo"
 gem "json-schema"
 # Authentication
 # Point at branch until devise is compatible with Turbo, see https://github.com/heartcombo/devise/pull/5340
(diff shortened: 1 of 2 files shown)


paulrobertlloyd (approved):
❤️ this!

PR #212, opened 2022-01-12
CLDC-509: Previous postcode validation

smeny:
* Add a validation test separated from the main case_log_spec
* Create a new validator for the "Local Authority" section
* Changes the postcode regexp to be case insensitive

### DIFF
--- a/spec/models/validations/local_authority_validations_spec.rb
+++ b/spec/models/validations/local_authority_validations_spec.rb
@@ -1,7 +1,7 @@
 require "rails_helper"
 require_relative "../../request_helper"
 
-RSpec.describe Form, type: :model do
+RSpec.describe CaseLog do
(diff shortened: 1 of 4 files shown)


spec/models/validations/local_authority_validations_spec.rb
```diff
@@ -0,0 +1,56 @@
+require "rails_helper"
+require_relative "../../request_helper"
+
+RSpec.describe Form, type: :model do
```
baarkerlounger: Why is this describing `Form`? Should be `CaseLog` I think?

baarkerlounger (approved):
Aside from the one comment I like this approach a lot! Much more unit friendly.

spec/models/validations/local_authority_validations_spec.rb
smeny: Fixed

PR #213, opened 2022-01-12
Cldc 632 income and benefits section

kosiakkatrina:
- [x] Update benefits and income section content
- [x] Adjust the route for 2021/2022 lettings benefits and income section
- [x] Allow for multiple condition sets with depends_on

Do do in a different PR for the same ticket:

- [ ] update check answers page
- [ ] update 2022/2023 form
- [ ] add validations
- [ ] add sales version


app/models/case_log.rb
```diff
@@ -294,6 +301,13 @@ def get_inferred_la(postcode)
     end
   end
 
+  def get_has_benefits
+    return "yes" if ["1. Housing benefit",
```
baarkerlounger: Does this need to be "Yes"?

app/models/constants/case_log.rb
```diff
@@ -198,23 +203,22 @@ module Constants::CaseLog
   }.freeze
 
   BENEFITS = {
-    "All" => 1,
-    "Some" => 2,
-    "None" => 3,
-    "Don’t know" => 4,
+    "1. All" => 1,
```
baarkerlounger: @paulrobertlloyd / @roobottom can confirm but I think the decision was not to include the numbers in answer options until a further spike has been done on the paper form?

app/models/case_log.rb
baarkerlounger: Array itself might also be worth extracting out into a named constant?

config/forms/2021_2022.json
```diff
@@ -2022,40 +2232,403 @@
...
                   "type": "numeric",
                   "min": 0,
                   "step": 1,
                   "prefix": "£",
                   "suffix": "every week",
                   "readonly": true,
                   "requires_js": true
+                }
+              },
+              "depends_on" : [{
```
kosiakkatrina, line 2246: Might need to add more logic to evaluating depends_on instead of writing out every condition

app/models/constants/case_log.rb
```diff
@@ -144,6 +144,11 @@ module Constants::CaseLog
     "Yes" => 1,
   }.freeze
 
+  POLAR3 = {
```
baarkerlounger: What's the purpose/distinction between of `POLAR`, `POLAR2` and `POLAR3`? Shouldn't they all be the same (a mapping of `0: "No" and 1: "Yes"` since order doesn't matter for these?

app/models/constants/case_log.rb
paulrobertlloyd: Correct. 🤞we never have to include the numbers on the service.

config/forms/2021_2022.json
baarkerlounger: Think this is fine. It's easily understandable.

config/forms/2021_2022.json
```diff
@@ -1903,79 +1903,289 @@
...
+            },
+            "rent_weekly": {
+              "header": "",
+              "description": "",
+              "questions": {
                 "brent": {
                   "check_answer_label": "Basic Rent",
-                  "header": "What is the basic rent?",
-                  "hint_text": "Eligible for housing benefit or Universal Credit",
+                  "header": "What is the basic rent every week?",
```
baarkerlounger, line 2187: This looks like it also covers https://digital.dclg.gov.uk/jira/browse/CLDC-787

app/models/constants/case_log.rb
kosiakkatrina: When was this decided and how come it's not on the map?

app/models/case_log.rb
baarkerlounger: Any particular reason not to make `has_benefits` an enum?

kosiakkatrina: I think this is just for us to track, but yeah, makes sense to do it the way we do it everywhere else

app/models/constants/case_log.rb
paulrobertlloyd: Yesterday afternoon! Removed from the prototype, still need to update the map (that might be a while!)

app/models/constants/case_log.rb
kosiakkatrina: Well, POLAR and POLAR2 are different because depending on the field we save answers differently in the db (based on the data dictionary, which is a bit inconsistent)

POLAR3 is for `nocharge` column where the question sort of gets asked inversely ("Does the household pay rent or other charges for the accommodation?"), so if answered "No" it would mean that `nocharge` is true and therefore based on the data dictionary should be saved as 1 in the db. 

Would it make sense to instead add a new field for this question and consider `nocharge` as a derived variable?

app/models/constants/case_log.rb
kosiakkatrina: 😭 Alright, I'll get rid of them. I take it the wording still stays the same?

paulrobertlloyd: Yup. Just not showing the numbers at the beginning of each question.

kosiakkatrina: Sweet, thank you thank you!

app/models/constants/case_log.rb
baarkerlounger: Ah right, if the inconsistency is there already in the data dictionary then yeah I guess we need to keep it :disappointed: 

`nocharge` as a derived variable seems more intuitive to me if we're not asking it directly but either way works.

db/schema.rb
```diff
@@ -19,8 +19,8 @@
     t.string "email", default: "", null: false
     t.string "encrypted_password", default: "", null: false
     t.string "reset_password_token"
-    t.datetime "reset_password_sent_at"
-    t.datetime "remember_created_at"
+    t.datetime "reset_password_sent_at", precision: 6
```
baarkerlounger: Where is this coming from?

kosiakkatrina: No idea, just running the migrations does that

app/models/constants/case_log.rb
kosiakkatrina: changed that now as I do agree it's probably more intuitive. Could you give that a glance?

db/schema.rb
baarkerlounger: Curious, seems to have happened on all date fields... anything underlying changed?

kosiakkatrina: Not that I know of, I'll manually remove them cause it's a bit odd

kosiakkatrina: Maybe Postgres version is different (? dunno if that'd impact it)

baarkerlounger (approved):

PR #214, opened 2022-01-12
CLDC-896: Integer Validation

baarkerlounger:
This fixes:
image


dushan-madetech (approved):
Looks good to me from a code perspective 👍

PR #215, opened 2022-01-12
CLDC-836: Don't make subsection labels links if they're not enabled

baarkerlounger:
After:

image
image

### DIFF
--- a/app/helpers/tasklist_helper.rb
+++ b/app/helpers/tasklist_helper.rb
@@ -35,11 +35,11 @@ module TasklistHelper
   end
 
   def subsection_link(subsection, case_log)
-    next_page_path = if subsection.status(case_log) != :cannot_start_yet
-                       first_page_or_check_answers(subsection, case_log)
-                     else
-                       "#"
-                     end
-    govuk_link_to(subsection.label, next_page_path.to_s.dasherize, class: "task-name")
+    if subsection.status(case_log) != :cannot_start_yet
+      next_page_path = first_page_or_check_answers(subsection, case_log).to_s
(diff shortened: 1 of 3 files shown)


paulrobertlloyd:
This is great, but looking at the HTML, there are a few other issues we should address.

Here’s the current markup:

```html
<li class="app-task-list__item" id="about_this_log">
  <a class="govuk-link task-name" href="/logs/1/gdpr-acceptance">About this log</a>
  <strong class="govuk-tag app-task-list__tag govuk-tag--grey">
    Not started
  </strong>
</li>
```

* `id` should be on the tag
* `id` should be kebab-case
* `a` should be wrapped by a `span` with class `app-task-list__task-name`
* `a` should have `aria-describedby` attribute with name of the `id` on the tag

This gives the correct markup:

```html
<li class="app-task-list__item">
  <span class="app-task-list__task-name">
    <a href="/logs/1/gdpr-acceptance" aria-describedby="about-this-log">About this log</a>
  </span>
  <strong class="govuk-tag app-task-list__tag govuk-tag--grey" id="about-this-log">
    Not started
  </strong>
</li>
```

If the item isn’t linked:

```html
<li class="app-task-list__item">
  <span class="app-task-list__task-name">
    About this log
  </span>
  <strong class="govuk-tag app-task-list__tag govuk-tag--grey" id="about-this-log">
    Not started
  </strong>
</li>
```

See: https://govuk-prototype-kit.herokuapp.com/docs/templates/task-list

paulrobertlloyd (approved):
Cheers!

PR #218, opened 2022-01-13
CLDC-913: Export validation messages into i18n language file

smeny:


baarkerlounger:
I think we check against these strings in various specs too. Would be great to point those at the same file so we don't need to update specs if the strings change. Otherwise this looks fantastic! :+1:

app/models/validations/household_validations.rb
```diff
@@ -21,33 +21,33 @@ def validate_other_reason_for_leaving_last_settled_home(record)
...
     if record.reason == "Don’t know" && record.underoccupation_benefitcap != "Don’t know"
-      record.errors.add :underoccupation_benefitcap, "must be don’t know if tenant’s main reason for leaving is don’t know"
+      record.errors.add :underoccupation_benefitcap, I18n.t("validations.household.underoccupation_benefitcap.dont_know_required")
     end
   end
 
   def validate_armed_forces_injured(record)
     if (record.armedforces == "A current or former regular in the UK Armed Forces (excluding National Service)" || record.armedforces == "A current or former reserve in the UK Armed Forces (excluding National Service)") && record.reservist.blank?
-      record.errors.add :reservist, "You must answer the armed forces injury question if the tenant has served in the armed forces"
+      record.errors.add :reservist, I18n.t("validations.household.reservist.inquiry_required")
```
kosiakkatrina: Should this be injury_required?

app/models/validations/household_validations.rb
```diff
@@ -21,33 +21,33 @@ def validate_other_reason_for_leaving_last_settled_home(record)
...
     end
 
     if (record.armedforces == "No" || record.armedforces == "Prefer not to say") && record.reservist.present?
-      record.errors.add :reservist, "You must not answer the armed forces injury question if the tenant has not served in the armed forces or prefer not to say was chosen"
+      record.errors.add :reservist, I18n.t("validations.household.reservist.inquiry_not_required")
```
kosiakkatrina: same here

app/models/validations/household_validations.rb
smeny: good catch!

kosiakkatrina (approved):

PR #228, opened 2022-01-19
Cldc 632 check answers

kosiakkatrina:
Update check answers labels and add suffix and prefix to answers
Display readonly questions in check answers
Format currency
Display `you didn't answer this question` for unanswered questions

image


app/models/case_log.rb
```diff
@@ -241,18 +241,20 @@ def set_derived_fields!
...
       self.incref = nil
-    when "Yes – the household has a yearly income"
+    when "Annually"
       self.incfreq = "Yearly"
       self.incref = nil
     when "Tenant prefers not to say"
       self.incref = 1
       self.incfreq = nil
+    when "Don’t know"
+      self.incfreq = nil
```
smeny, line 257: Not a bid deal but we could factor out the `self.incfreq = nil`

kosiakkatrina: Wouldn't we need to set it?

db/migrate/20220117104521_change_currency_types.rb
```diff
@@ -0,0 +1,22 @@
+class ChangeCurrencyTypes < ActiveRecord::Migration[7.0]
+  def up
+    change_column :case_logs, :earnings, :decimal, precision: 10, scale: 2
```
smeny: Just wondering, is there a case for using `bigint` instead? A scale of 2 is fine as long as we don't do any operations with rounding.

kosiakkatrina: How would using bigint work?

app/models/case_log.rb
kosiakkatrina: We're only setting it to nil in 2 cases, not sure how we'd refactor it to be clearer 🤔

app/models/form/subsection.rb
```diff
@@ -31,7 +31,7 @@ def status(case_log)
     end
 
     qs = applicable_questions(case_log)
-    return :not_started if qs.all? { |question| case_log[question.id].blank? }
+    return :not_started if qs.all? { |question| case_log[question.id].blank? || question.read_only? }
```
baarkerlounger, line 34: This is fine because we're now setting the derived fields before validation so it _should_ always be there right?

db/migrate/20220117104521_change_currency_types.rb
baarkerlounger: Don't we collect all financial amounts to the nearest £ anyway? Why do we need to store a decimal? (But I also think we should be using int or bigint for all currency amounts)

app/models/case_log.rb
smeny: I count 5 so let's catch up 😄

config/forms/2021_2022.json
```diff
@@ -2423,9 +2430,10 @@
                   "step": 1,
                   "width": 5,
                   "prefix": "£",
-                  "suffix": "every 2 weeks",
+                  "suffix": " every 2 weeks",
```
smeny, line 2433: Sorry little detail, but why do we need a space for suffixes?

kosiakkatrina: In the check answers page we construct the answer from prefix and suffix. I could join the prefix, answer and suffix with an empty space, but then there would be a space between `£` and amount. I could also join prefix and answer without space and join the suffix with space, but that then doesn't account for non currency questions that might have prefixes/suffixes. We don't have such questions yet, but in case we do in the future this was the easiest way. I can change it if you convince me to tho!

app/models/case_log.rb
smeny: clarified

baarkerlounger (approved):

PR #230, opened 2022-01-19
Cldc 938 multiple page questions

kosiakkatrina:
Update headers, add a divider before total rent and fix a typo

Before:
image

After:
image

### DIFF
--- a/app/views/form/page.html.erb
+++ b/app/views/form/page.html.erb
@@ -30,6 +30,9 @@
       <%= form_with model: @case_log, url: form_case_log_path(@case_log), method: "post" do |f| %>
         <%= f.govuk_error_summary %>
         <% @page.questions.map do |question| %>
+          <% if question.read_only? %>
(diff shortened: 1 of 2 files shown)


app/views/form/page.html.erb
```diff
@@ -30,6 +30,9 @@
       <%= form_with model: @case_log, url: form_case_log_path(@case_log), method: "post" do |f| %>
         <%= f.govuk_error_summary %>
         <% @page.questions.map do |question| %>
+          <% if question.read_only? %>
+            <hr class="govuk-section-break govuk-section-break--visible govuk-section-break--m">
```
baarkerlounger: So I think this question/field only shows if you have JS enabled. Would we still want the divider to show if the field wasn't?

kosiakkatrina: That's a good question, probably not, I'll fix it

baarkerlounger (approved):

PR #231, opened 2022-01-19
CLDC-831: Remove link to complete all answers from check your answers page

paulrobertlloyd:
This link is not in the prototype, and we have not observed a need for it so far. We can revisit this feature should we have evidence for it.

| Before | After |
| - | - |
| Screenshot 2022-01-19 at 17 23 20 | Screenshot 2022-01-19 at 17 23 33 |


baarkerlounger (approved):
Always happy to remove code :)

PR #234, opened 2022-01-21
Set grey background if the field is readonly

kosiakkatrina:
image

### DIFF
--- a/app/helpers/question_attribute_helper.rb
+++ b/app/helpers/question_attribute_helper.rb
@@ -10,6 +10,7 @@ module QuestionAttributeHelper
 private
 
   def numeric_question_html_attributes(question)
+    return { "style": "background-color: #f3f2f1;" } if question.read_only?
     return {} if question.fields_to_add.blank? || question.result_field.blank?
 
     {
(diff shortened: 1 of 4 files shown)


app/helpers/question_attribute_helper.rb
```diff
@@ -10,6 +10,7 @@ def stimulus_html_attributes(question)
 private
 
   def numeric_question_html_attributes(question)
+    return { "style": "background-color: #f3f2f1;" } if question.read_only?
```
baarkerlounger, line 13: Might be better to add a CSS class? (Just a nit though)

baarkerlounger (approved):

app/helpers/question_attribute_helper.rb
kosiakkatrina: I thought that, but it's just one thing. What are the benefits?

baarkerlounger: Not huge so I think either way is ok. Benefits are consistent place you know to look for styling and colours (it's probably where I'd check first), and quicker to reuse if we did have a second thing at some point?).

kosiakkatrina: I can't see us using it somewhere else right now. I'll just merge it in as is

PR #235, opened 2022-01-21
Render radio conditional questions inline rather than below

baarkerlounger:
After this PR:

- When the top level question is a radio button and the conditional question is a text, numeric or date field
   - Then the conditional question is shown inline
- When the top level question is a radio button and the conditional question is a checkbox, select or radio button
  -  Then the conditional question is shown on a new page (not enforced by code)

- When the top level question is a numeric or text question for now the behaviour is unchanged (other household members)

**When the top level question is a radio question**

Before:
image

After:
image

When conditional is not a text box:

image

--------------------------------------------------------------------------------------------------------------------------------------------------------------

**When the top level question is not a radio type the behaviour remain unchanged for now:**
image

Radios with further conditional radio should now be on there own page:
image

image

image

image

### DIFF
--- a/app/helpers/question_attribute_helper.rb
+++ b/app/helpers/question_attribute_helper.rb
@@ -1,8 +1,7 @@
 module QuestionAttributeHelper
   def stimulus_html_attributes(question)
     attribs = [
-      numeric_question_html_attributes(question),
-      conditional_html_attributes(question),
+      numeric_question_html_attributes(question)
     ]
     merge_controller_attributes(*attribs)
   end
@@ -18,16 +17,6 @@ private
       "data-action": "numeric-question#calculateFields",
       "data-target": "case-log-#{question.result_field.to_s.dasherize}-field",
       "data-calculated": question.fields_to_add.to_json,
-    }
-  end
-
-  def conditional_html_attributes(question)
-    return {} if question.conditional_for.blank?
-
-    {
-      "data-controller": "conditional-question",
-      "data-action": "conditional-question#displayConditional",
-      "data-info": question.conditional_for.to_json,
     }
   end
 end
--- a/app/helpers/question_attribute_helper.rb
+++ b/app/helpers/question_attribute_helper.rb
@@ -1,7 +1,8 @@
 module QuestionAttributeHelper
   def stimulus_html_attributes(question)
     attribs = [
-      numeric_question_html_attributes(question)
+      numeric_question_html_attributes(question),
+      conditional_html_attributes(question),
     ]
     merge_controller_attributes(*attribs)
   end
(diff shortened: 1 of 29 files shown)


app/helpers/question_attribute_helper.rb
```diff
@@ -20,16 +19,6 @@ def numeric_question_html_attributes(question)
       "data-calculated": question.fields_to_add.to_json,
     }
   end
-
-  def conditional_html_attributes(question)
```
baarkerlounger: Can't actually remove this until we also change the Other Household Members conditional questions but after that we can.

baarkerlounger:
@paulrobertlloyd  Few things to confirm here:

1. For the Postcode Known question currently there is no validation for the case `Postcode Known = "No"`, `Postcode entered = "Something"`. Instead we *assume* that if a postcode value has been entered, Postcode Known should have been "Yes" and fix it for them. (This is only really possible via the UI if JS is turned off)

2. Are you happy with conditional check box questions appearing inline as above? If so should there be a max limit on how many answer options it can have or should it just be a new page? (Also still need to make sure input answers are cleared, when invalidated if so)

3. Are you happy with date questions appearing inline as above? 

4. Should there be validations for the care home charges question? I don't see any, nor a ticket for it?

paulrobertlloyd:
1. Let’s validate based on the answer to the radios:
    * If you select ‘Yes’, and have entered a text value for ‘Yes’, validate that text value.
    * If you select ‘No’, but have entered a text value for ‘Yes’, don’t perform validation on that text value.
  
    (The non-JS use case definitely is worth keeping mind)

2. Yes, but I don’t we have or will ever show checkboxes as a conditional field.

3. Yes.

4. Yes. Perhaps check with Mohammed?

baarkerlounger:
@paulrobertlloyd  for (2) we do have that case already - the screenshots above are from the actual form not made up examples

paulrobertlloyd:
2. Sure, but not for long… it’s a separate page in the prototype. 😉

baarkerlounger:
@paulrobertlloyd  for 2 have updated to match.

For 1:

> Let’s validate based on the answer to the radios:
>
>   If you select ‘Yes’, and have entered a text value for ‘Yes’, validate that text value.
>    If you select ‘No’, but have entered a text value for ‘Yes’, don’t perform validation on that text value.

This is how it currently works with the addition that "If you select 'No' it will assume that value is a mistake and clear it for postcodes. 

Will leave it to you to decide if you want to change that in a new ticket?

paulrobertlloyd:
> This is how it currently works with the addition that "If you select 'No' it will assume that value is a mistake and clear it for postcodes.

Perfect.

dushan-madetech (approved):
Looks good 👍, does the json schema (the one sitting in `config/forms/schema/...`) need to be updated as a result of this change?

baarkerlounger:
@dushan-madetech I don't think so. Nothing that was previously valid isn't valid anymore, we're just documenting that you *shouldn't* do it anymore.

PR #237, opened 2022-01-21
Do not show privacy notice question in check answers page

kosiakkatrina:


baarkerlounger:
This means if the answer is no they just see a blank Check Answers page? Is that expected here?

kosiakkatrina:
> This means if the answer is no they just see a blank Check Answers page? Is that expected here?

ooh, that's a good point, @paulrobertlloyd, what do you think?

paulrobertlloyd:
I think we ignore this ticket for now and close this PR – where the privacy notice appears is something we’re looking at again in design.

PR #241, opened 2022-01-24
Use Notify rather than Gmail for devise emails

baarkerlounger:
This PR switches Devise from using Gmail SMTP via ActionMailer to Gov UK Notify to send emails for password resets and user confirmations.

- [x] Use Notify for new user emails
- [x] Use Notify for password reset emails
- [x] Create templates in Notify account
- [x] Remove no longer needed Devise mailer templates since the templates now live in Notify
- [x] Set API key env var in secret store

image
image

### DIFF
--- a/app/mailers/devise_notify_mailer.rb
+++ b/app/mailers/devise_notify_mailer.rb
@@ -1,0 +1,49 @@
+class DeviseNotifyMailer < Devise::Mailer
+  require "notifications/client"
+
+  RESET_PASSWORD_TEMPLATE_ID = "4593417c-500f-452c-8111-0f9d311aad0e".freeze
+  SET_PASSWORD_TEMPLATE_ID   = "00cd7163-4213-4596-b4f9-9e72796e0d76".freeze
+
+  def notify_client
+    @notify_client ||= ::Notifications::Client.new(ENV["GOVUK_NOTIFY_API_KEY"])
+  end
+
+  def host
+    @host ||= ENV["APP_HOST"]
+  end
+
+  def send_email(email, template_id, personalisation)
+    notify_client.send_email(
+      email_address: email,
+      template_id: template_id,
+      personalisation: personalisation,
+    )
+  end
+
+  def reset_password_instructions(record, token, _opts = {})
+    template_id = record.last_sign_in_at ? RESET_PASSWORD_TEMPLATE_ID : SET_PASSWORD_TEMPLATE_ID
+    personalisation = {
+      name: record.name,
+      email: record.email,
+      organisation: record.organisation.name,
+      link: "https://#{host}/users/password/edit?reset_password_token=#{token}",
+    }
+    send_email(record.email, template_id, personalisation)
+  end
+
+  # def confirmation_instructions(record, token, _opts = {})
+  #   super
+  # end
+  #
+  # def unlock_instructions(record, token, opts = {})
+  #   super
+  # end
+  #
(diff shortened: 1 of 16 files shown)


app/mailers/devise_notify_mailer.rb
```diff
@@ -0,0 +1,49 @@
...
+  #   super
+  # end
+  #
+  # def email_changed(record, opts = {})
+  #   super
+  # end
+  #
+  # def password_change(record, opts = {})
+  #   super
+  # end
```
dushan-madetech, line 48: Was there a specific reason for leaving all this commented out code in here?

config/initializers/devise.rb
```diff
@@ -24,10 +24,11 @@
   # Configure the e-mail address which will be shown in Devise::Mailer,
   # note that it will be overwritten if you use your own mailer class
   # with default "from" parameter.
-  config.mailer_sender = ENV["CORE_EMAIL_USERNAME"]
+  # config.mailer_sender = ENV["CORE_EMAIL_USERNAME"]
```
dushan-madetech, line 27: This could just be deleted right? Since this was only set up for the Gmail

dushan-madetech (approved):
Left some comments but nothing worth blocking this over. It looks much better now! 👍

dushan-madetech:
Also a comment outside of the approval. Should we have an ADR for using Gov Notify and why we have chosen/prefer this to the previous solution?

app/mailers/devise_notify_mailer.rb
baarkerlounger: Yeah was mainly thinking that if we used other devise notification paths in the future this makes it nice and obvious which methods need implementing in the subclass to get notify rather than action mailer without having to go look in the super class. Happy to delete though if it's not helpful.

config/initializers/devise.rb
baarkerlounger: Possibly. Might still be useful as a fallback for local dev or other things? At the moment only devise emails are routed through Notify.

baarkerlounger:
@dushan-madetech happy to write an ADR but not sure if that's the right place since this isn't really a technical decision at all but rather justa  DLUHC departmental one?

dushan-madetech:
@baarkerlounger No need in that case, figured it was worth asking just in case 😄

app/mailers/devise_notify_mailer.rb
```diff
@@ -0,0 +1,49 @@
+class DeviseNotifyMailer < Devise::Mailer
+  require "notifications/client"
+
+  RESET_PASSWORD_TEMPLATE_ID = "4593417c-500f-452c-8111-0f9d311aad0e".freeze
+  SET_PASSWORD_TEMPLATE_ID   = "00cd7163-4213-4596-b4f9-9e72796e0d76".freeze
```
kosiakkatrina: Should this be in Constants?

kosiakkatrina: Or env vars or smth

kosiakkatrina (approved):

app/mailers/devise_notify_mailer.rb
baarkerlounger: They could be ENV vars I suppose but they're not expected to ever be used anywhere outside this file and also not sensitive so maybe not a ton of benefit?

kosiakkatrina: absolutely fair 👍

PR #242, opened 2022-01-24
CLDC 502: rsnvac validations

kosiakkatrina:
- Add the following validations:
Rsnvac cannot be 9 (relet to tenant who occupied same property as temp accom) if Q11 (prev household situation) is 4, 6, 8, 24 or 25 (various non temp accom options)
Rsnvac cannot be 9 (relet to tenant who occupied same property as temp accom) if Q16 (source of letting referral) is one of 8, 10, 12, 13, 14 or 15

- Add referral field/question
- Filter out errors in view leaving only errors relevant for the page and base errors. This allows different messages to be displayed for composite validations


baarkerlounger (approved):
Very nice

PR #244, opened 2022-01-25
Cldc 956 household needs subsection incomplete bugfix

dushan-madetech:
Fixed the methods for counting answered questions and subsection status

Screenshot 2022-01-25 at 16 52 56
Screenshot 2022-01-25 at 16 57 09

### DIFF
--- a/app/models/form/subsection.rb
+++ b/app/models/form/subsection.rb
@@ -32,7 +32,7 @@ class Form::Subsection
 
     qs = applicable_questions(case_log)
     return :not_started if qs.all? { |question| case_log[question.id].blank? || question.read_only? }
-    return :completed if qs.all? { |question| question.completed?(case_log) }
+    return :completed if qs.all? { |question| question.type == "checkbox" ? question.answer_options.keys.any? { |key| case_log[key] == "Yes" } : question.completed?(case_log) }
 
     :in_progress
   end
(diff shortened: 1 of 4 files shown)


app/models/form/subsection.rb
```diff
@@ -32,7 +32,7 @@ def status(case_log)
 
     qs = applicable_questions(case_log)
     return :not_started if qs.all? { |question| case_log[question.id].blank? || question.read_only? }
-    return :completed if qs.all? { |question| question.completed?(case_log) }
+    return :completed if qs.all? { |question| question.type == "checkbox" ? question.answer_options.keys.any? { |key| case_log[key] == "Yes" } : question.completed?(case_log) }
```
baarkerlounger: I think this logic should go into the `question.completed?` method. i.e. I should be able to ask a question whether it's completed or not without having to care what type it is.

app/models/form/subsection.rb
```diff
@@ -58,7 +58,7 @@ def applicable_questions(case_log)
   end
 
   def answered_questions(case_log)
-    applicable_questions(case_log).select { |question| case_log[question.id].present? }
+    applicable_questions(case_log).select { |question| question.type == "checkbox" ? question.answer_options.keys.any? { |key| case_log[key] == "Yes" } : case_log[question.id].present? }
```
baarkerlounger: This could probably also be switched to using `question.completed?`?

app/models/form/subsection.rb
dushan-madetech: That's actually a more elegant solution so will get that sorted 🙂

app/models/form/subsection.rb
dushan-madetech: Will update this after updating the completed method, especially since there are tests now 🙂

baarkerlounger (approved):
Nice!

PR #248, opened 2022-01-26
Ensures DB migration only happens on first instance

smeny:


manifest.yml
```diff
@@ -1,10 +1,10 @@
 ---
 defaults: &defaults
   buildpacks:
-    - https://github.com/cloudfoundry/ruby-buildpack.git
+    - ruby_buildpack
```
smeny, line 4: I changed the buildpack to the list supported by PaaS.
See `cf buildpacks`

lib/tasks/cf.rake
```diff
@@ -0,0 +1,12 @@
+namespace :cf do
+  desc "Only run on the first application instance"
+  task on_first_instance: :environment do
+    instance_index = -1
+    begin
+      instance_index = JSON.parse(ENV["VCAP_APPLICATION"])["instance_index"]
+    rescue StandardError
```
baarkerlounger: I wonder what scenario this is wanted in? If you couldn't derive an index and therefore couldn't run migrations wouldn't you rather have a loud crash?

baarkerlounger (approved):

lib/tasks/cf.rake
smeny: I am not sure myself to be honest, should I remove the rescue altogether?
If the JSON is malformed or VCAP_APPLICATION missing, I think indeed we should break.

Otherwise, we might just stop migrating silently.

baarkerlounger: Might be a good place to insert a Sentry call once we have that. Otherwise yeah I'd probably say let's just fail.

lib/tasks/cf.rake
```diff
@@ -0,0 +1,8 @@
+namespace :cf do
+  desc "Only run on the first application instance"
+  task on_first_instance: :environment do
+    # We expect this information to be always available or break otherwise
+    instance_index = JSON.parse(ENV["VCAP_APPLICATION"])["instance_index"]
```
baarkerlounger, line 5: Probably didn't mean to remove this `instance_index = -1`?

smeny: I actually did because otherwise I get:
`lib/tasks/cf.rake:4:5: W: Lint/UselessAssignment: Useless assignment to variable - instance_index. (https://rubystyle.guide#underscore-unused-vars)
    instance_index = -1
    ^^^^^^^^^^^^^^
`

PR #252, opened 2022-01-26
CLDC-731 bypass tasklist from check answers

dushan-madetech:
**Save and go to next incomplete section when the case log still has subsections to complete**
Screenshot 2022-01-27 at 09 32 30

**Save and go to submit when the case log only has the declaration left to complete**
Screenshot 2022-01-27 at 14 57 59

### DIFF
--- a/app/models/form.rb
+++ b/app/models/form.rb
@@ -67,7 +67,7 @@ class Form
   def first_question_in_last_subsection(subsection_ids)
     next_subsection = get_subsection(subsection_ids[subsection_ids.length - 1])
     first_question_in_subsection = next_subsection.pages.first.id
-    return first_question_in_subsection.to_s.dasherize
+    first_question_in_subsection.to_s.dasherize
   end
 
   def next_subsection(subsection, case_log, subsection_ids)
@@ -78,7 +78,7 @@ class Form
       next_subsection = get_subsection(subsection_ids[0])
     end
 
-    return next_subsection
+    next_subsection
   end
 
   def all_subsections_except_declaration_completed?(case_log)
--- a/app/models/form.rb
+++ b/app/models/form.rb
@@ -46,27 +46,27 @@ class Form
       "case_log_#{nxt_page}_path"
     end
   end
-  
+
   def next_incomplete_section_redirect_path(subsection, case_log)
     subsection_ids = subsections.map(&:id)
-    
-    if(case_log.status == "completed")
+
+    if case_log.status == "completed"
       next_subsection = get_subsection(subsection_ids[subsection_ids.length - 1])
       first_question_in_subsection = next_subsection.pages.first.id
-      return "#{first_question_in_subsection}".dasherize
+      return first_question_in_subsection.to_s.dasherize
     end
 
     next_subsection_id_index = subsection_ids.index(subsection.id) + 1
     next_subsection = get_subsection(subsection_ids[next_subsection_id_index])
 
-    if(next_subsection.id == "declaration" && case_log.status != "completed")
+    if next_subsection.id == "declaration" && case_log.status != "completed"
       next_subsection = get_subsection(subsection_ids[0])
     end
-    if(next_subsection.status(case_log) == :completed)
+    if next_subsection.status(case_log) == :completed
       next_incomplete_section_redirect_path(next_subsection, case_log)
     else
       first_question_in_subsection = next_subsection.pages.first.id
-      return "#{first_question_in_subsection}".dasherize
+      first_question_in_subsection.to_s.dasherize
     end
   end
 
--- a/app/models/form.rb
+++ b/app/models/form.rb
@@ -55,10 +55,13 @@ class Form
       first_question_in_subsection = next_subsection.pages.first.id
       return "#{first_question_in_subsection}".dasherize
     end
-    
+
     next_subsection_id_index = subsection_ids.index(subsection.id) + 1
     next_subsection = get_subsection(subsection_ids[next_subsection_id_index])
 
+    if(next_subsection.id == "declaration" && case_log.status != "completed")
+      next_subsection = get_subsection(subsection_ids[0])
+    end
     if(next_subsection.status(case_log) == :completed)
       next_incomplete_section_redirect_path(next_subsection, case_log)
     else
--- a/app/models/form.rb
+++ b/app/models/form.rb
@@ -61,9 +61,11 @@ class Form
       next_incomplete_section_redirect_path(next_subsection, case_log)
     when :in_progress
       "#{next_subsection.id}/check_answers".dasherize
-    else
+    when :not_started
       first_question_in_subsection = next_subsection.pages.first.id
       first_question_in_subsection.to_s.dasherize
+    else
+      "error"
     end
   end
 
--- a/app/models/form.rb
+++ b/app/models/form.rb
@@ -56,8 +56,11 @@ class Form
 
     next_subsection = next_subsection(subsection, case_log, subsection_ids)
 
-    if next_subsection.status(case_log) == :completed
+    case next_subsection.status(case_log)
+    when :completed
       next_incomplete_section_redirect_path(next_subsection, case_log)
+    when :in_progress
+      "#{next_subsection.id}/check_answers".dasherize
     else
       first_question_in_subsection = next_subsection.pages.first.id
       first_question_in_subsection.to_s.dasherize
--- a/app/models/form.rb
+++ b/app/models/form.rb
@@ -51,17 +51,10 @@ class Form
     subsection_ids = subsections.map(&:id)
 
     if case_log.status == "completed" || all_subsections_except_declaration_completed?(case_log)
-      next_subsection = get_subsection(subsection_ids[subsection_ids.length - 1])
-      first_question_in_subsection = next_subsection.pages.first.id
-      return first_question_in_subsection.to_s.dasherize
+      return first_question_in_last_subsection(subsection_ids)
     end
 
-    next_subsection_id_index = subsection_ids.index(subsection.id) + 1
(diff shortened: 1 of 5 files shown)


app/models/form.rb
```diff
@@ -47,6 +47,29 @@ def next_page_redirect_path(page, case_log)
...
+    if next_subsection.id == "declaration" && case_log.status != "completed"
+      next_subsection = get_subsection(subsection_ids[0])
+    end
+    if next_subsection.status(case_log) == :completed
+      next_incomplete_section_redirect_path(next_subsection, case_log)
+    else
+      first_question_in_subsection = next_subsection.pages.first.id
+      first_question_in_subsection.to_s.dasherize
+    end
+  end
```
dushan-madetech: This is really ugly but I don't have any ideas about how to make it nicer. Open to suggestions or a brief pairing session to tidy this up

spec/factories/case_log.rb
```diff
@@ -153,6 +153,7 @@
       is_carehome { "No" }
       chcharge { 7 }
       letting_in_sheltered_accomodation { "No" }
+      la_known { "Yes" }
```
dushan-madetech, line 154: I had to add this line for the test here: `./spec/features/form/check_answers_page_spec.rb:194` to work as it relies on the completed case log. I'm not sure why only this test failed but the other tests relying on the completed case log passed when this was missing from the case log 🤔

baarkerlounger: Probably because whether this is required or not depends on what other fields you're setting? _Might_ be safer to set this in the specific test you need it in so you can be sure you haven't accidentally stopped any other tests from legitimately failing if their functionality changes?

spec/models/form_spec.rb
```diff
@@ -32,6 +32,48 @@
...
+      it "returns the first page of the next incomplete subsection (skipping completed subsections)" do
+        case_log.armedforces = "No"
+        case_log.illness = "No"
+        case_log.housingneeds_a = "Yes"
+        case_log.la = "York"
+        case_log.illness_type_1 = "Yes"
+        expect(form.next_incomplete_section_redirect_path(subsection, case_log)).to eq("tenancy-code")
+      end
+
+      it "returns the next incomplete section by cycling back around if next subsections are completed" do
```
baarkerlounger: Am I understanding this right that if I have:

Section A - Incomplete
Section B - Complete
Section C - Incomplete

I'm on Check Answers page for Section B - the next incomplete section link should take me to C not A?

app/models/form.rb
baarkerlounger: Summarising previous discussion slightly:

I think part of the reason this is odd is that `sections` don't really have paths, pages do. If a section is incomplete (in progress) but the first question **has** been answered - should we really direct them to that first page anyway or rather to the first unanswered question's page?

I think we want to be finding the next page to direct them to (and there is a method for that, it just doesn't work for when the next page is in a different section). Now that you've got tests for your use case though I think it would make sense to refactor it to do that?

spec/factories/case_log.rb
dushan-madetech: I'm not setting any other fields, the test in question just use the exact ones defined by the completed case log factory because I'm less interested in the fields and more interested is testing the redirection with a completed case log so all l do in that test is use a completed case log created using the factory, no other setup involved

spec/models/form_spec.rb
dushan-madetech: Not quite, I couldn't think of a nice way to phrase this but if you had:

Section A - Incomplete
Section B - Incomplete
Section C - Complete
Declaration Section - Incomplete (requires all other sections completed)

And you're on Check Answers page for Section C instead of going to the declaration you should loop back around to section A

dushan-madetech: The test before this one is checking for the behaviour you described

baarkerlounger: Cool, that makes sense - was just checking that the behaviour I described _is_ the intended behaviour.

dushan-madetech:
@paulrobertlloyd Could you have a play around with this and see if the following scenarios behave like you'd expect

Scenario 1: About this log completed, go to check answers about this log, "Save and go to next incomplete section" is on page and redirects to the first question in household characteristics (even if that first question has been completed)

Scenario 2: About this log completed, household characteristics completed, go to check answers for about this log, "Save and go to next incomplete section" is on page and redirects to the first question in household situation (even if that first question has been completed).

Scenario 3: About this log completed, household characteristics not completed, household situation completed, go to check answers for household situation, "Save and go to next incomplete section" is on page and redirects to the first question in household needs (even if that first question has been completed) and not back to household characteristics.

Scenario 4: All the sections completed (except the declaration), go to check answers for any section, "Save and go to submit" is on page and redirects to the first question in the declaration subsection.

app/views/form/check_answers.html.erb
```diff
@@ -22,7 +22,13 @@
       </dl>
 
       <%= form_with model: @case_log, method: "get" do |f| %>
-        <%= f.govuk_submit "Save and continue"  %>
+        <%= f.govuk_submit 'Save and continue' do %>
+          <% if @case_log.status == "in_progress" && @case_log.status == "completed" || @case_log.form.all_subsections_except_declaration_completed?(@case_log) == false %>
+            <a class="govuk-button govuk-button--secondary" href=<%= "/logs/#{@case_log.id}/#{@case_log.form.next_incomplete_section_redirect_path(subsection, @case_log)}" %>>Save and go to next incomplete section</a>
```
paulrobertlloyd: Can use `govuk_link` helper for these links. See https://govuk-components.netlify.app/helpers/button/#link-styled-like-a-button

paulrobertlloyd (changes requested):
Will test behaviour in the app in a minute…

app/views/form/check_answers.html.erb
```diff
@@ -22,7 +22,13 @@
       </dl>
 
       <%= form_with model: @case_log, method: "get" do |f| %>
-        <%= f.govuk_submit "Save and continue"  %>
+        <%= f.govuk_submit 'Save and continue' do %>
```
paulrobertlloyd:
```suggestion
        <%= f.govuk_submit 'Save and return to log' do %>
```

paulrobertlloyd (commented):
Apart from the changes mentioned, works great! 🎉

app/views/form/check_answers.html.erb
dushan-madetech: Thanks for the tip, I originally got the above from https://govuk-form-builder.netlify.app/form-elements/submit/#a-group-of-buttons do you think it might be worth suggesting that they update the page to reference the helpers too or link out to the helpers?

paulrobertlloyd: Ah, interesting, and actually makes me realise… shouldn’t these be buttons, not links, if they submit a form? What happens if JS is unavailable?

Otherwise, maybe it’s worth a PR on the GOV.UK Form Builder repo, asking if pages should use or mention components from the GOV.UK Components library.

dushan-madetech: I can answer that! 😄 Under the hood these links/buttons on the check answers page aren't actually submitting anything. By the time you reach the check answers page you've already submitted all the questions you've answered since we're actually submitting after each question. These buttons/links just act as redirects to elsewhere in the case log process without performing any kind of submission so theoretically they shouldn't be affected by JS being unavailable (might be worth @baarkerlounger confirming this)

baarkerlounger: Yeah they are links rather than real buttons. Don't think they should be affected by JS at all but, as always, worth testing.

paulrobertlloyd (approved):
Perfect! Thanks @dushan-madetech.

spec/models/form_spec.rb
```diff
@@ -32,6 +32,97 @@
...
+      end
+
+      before do
+        case_log.tenant_code = "123"
+        case_log.age1 = 35
+        case_log.sex1 = "Male"
+        case_log.other_hhmemb = 0
+      end
+
+      it "returns the first page of the next incomplete subsection" do
```
baarkerlounger: Is this the correct behaviour even if the first page's questions have been answered already?

baarkerlounger: Answer to this is yes

baarkerlounger: @paulrobertlloyd can you confirm that this is what we want rather than redirecting to the first unanswered page?

paulrobertlloyd: If a section has been partially completed, I imagine we’d take you to the ‘Check your answers’ page for that section. otherwise, the first question in that section? Is that possible?

paulrobertlloyd: (For the record, not a big issue either way, we can polish as we go).

baarkerlounger: Summarising expected behaviour from discussion here:

**Scenario A. User is on Section 2 Check Answers page:**

Section 1: Not started
Section 2: Completed
Section 3: In Progress

-> Link should take the user to Section 3 Check answers page

**Scenario B. User is on Section 2 Check Answers page:**

Section 1: Not started
Section 2: Completed
Section 3: Not started

-> Link should take the user to Section 3 first question page

**Scenario C. User is on Section 2 Check Answers page:**

Section 1: Not started
Section 2: Completed
Section 3: Completed

-> Link should take the user to Section 1 first question page

baarkerlounger (changes requested):
When the next incomplete section in the cycle is "in progress" the link should go to the check answers page rather than the first page as shown by the scenarios left on the comment. Otherwise let's ship and then refactor!

dushan-madetech:
@baarkerlounger Will write a test for that behaviour and then make sure it's implemented. Will ping you for a quick re-review when that's done 🙂

baarkerlounger (approved):

PR #254, opened 2022-01-27
CLDC-791: Rename ‘About this log‘ section

paulrobertlloyd:
After testing ‘Tailor your log’, we’re going to use ‘Set up your [lettings|sales] log’.

### DIFF
--- a/config/forms/2021_2022.json
+++ b/config/forms/2021_2022.json
@@ -3,11 +3,11 @@
   "start_year": 2021,
   "end_year": 2022,
   "sections": {
-    "about_this_log": {
-      "label": "About this log",
+    "setup": {
+      "label": "Before you start",
       "subsections": {
-        "about_this_log": {
-          "label": "About this log",
+        "setup": {
+          "label": "Set up your lettings log",
           "pages": {
             "gdpr_acceptance": {
               "header": "",
@@ -230,7 +230,7 @@
       "subsections": {
         "household_characteristics": {
           "label": "Household characteristics",
-          "depends_on": [{ "about_this_log": "completed" }],
+          "depends_on": [{ "setup": "completed" }],
           "pages": {
             "person_1_age": {
               "header": "",
@@ -779,7 +779,7 @@
         },
         "household_situation": {
           "label": "Household situation",
-          "depends_on": [{ "about_this_log": "completed" }],
+          "depends_on": [{ "setup": "completed" }],
           "pages": {
             "previous_housing_situation": {
               "header": "",
@@ -907,7 +907,7 @@
         },
         "household_needs": {
           "label": "Household needs",
-          "depends_on": [{ "about_this_log": "completed" }],
+          "depends_on": [{ "setup": "completed" }],
           "pages": {
             "armed_forces": {
               "header": "Experience of the UK Armed Forces",
@@ -1051,7 +1051,7 @@
       "subsections": {
         "tenancy_information": {
           "label": "Tenancy information",
-          "depends_on": [{ "about_this_log": "completed" }],
+          "depends_on": [{ "setup": "completed" }],
           "pages": {
             "starter_tenancy": {
               "header": "",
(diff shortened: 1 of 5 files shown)


db/schema.rb
```diff
@@ -1,245 +0,0 @@
...
-#
-# This file is the source Rails uses to define your schema when running `bin/rails
-# db:schema:load`. When creating a new database, `bin/rails db:schema:load` tends to
-# be faster and is potentially less error prone than running all of your
-# migrations from scratch. Old migrations may fail to apply correctly if those
-# migrations use external dependencies or application code.
-#
-# It's strongly recommended that you check this file into your version control system.
-
-ActiveRecord::Schema.define(version: 2022_01_26_101743) do
```
baarkerlounger: This looks accidental.

config/forms/2021_2022.json
```diff
@@ -3,11 +3,11 @@
...
   "sections": {
-    "about_this_log": {
-      "label": "About this log",
+    "setup": {
+      "label": "Before you start",
       "subsections": {
-        "about_this_log": {
-          "label": "About this log",
+        "setup": {
+          "label": "Set up your lettings log",
```
baarkerlounger, line 10: This might need further thought/discussion. How are we going to differentiate between sales and lettings at this point?

paulrobertlloyd: Given the form has `"type": "lettings"` I wouldn’t have thought this would be a problem; sales would be in a different definition.

db/schema.rb
paulrobertlloyd: Gah!

config/forms/2021_2022.json
baarkerlounger: Think that was an initial idea but that key doesn't actually do anything right now (should maybe be removed). Possible that we put sales in it's own definition file - but I think you still have the problem of how do you know which this is at this first point since we only ask if it's a sale or letting half way through About this Log?

paulrobertlloyd: Given all the questions are lettings related for now, and we will launch private beta only with lettings, think this is fine. Also, in the prototype we’re looking to ask if want to create a sales or lettings before setting up the new log, partly because the form groupings and labels are so radical different (cf. references to ‘tenancy’).

Which makes me think… maybe we should remove the Lettings/Sale question in this JSON form?

Screenshot 2022-01-27 at 15 55 13

baarkerlounger: Ok I think that's fine if that's the approach we're taking but then we should probably also remove the sales/lettings questions as you suggested in this PR.

Gemfile.lock
```diff
@@ -1,468 +0,0 @@
-GIT
```
baarkerlounger: This one looks accidental too.

paulrobertlloyd: I thought I removed these files from this commit… my git foo is not strong.

paulrobertlloyd: Ooooh, I know what I’ve done. 🤦

config/forms/2021_2022.json
paulrobertlloyd: Have removed the sales related question and its dependencies.

spec/factories/case_log.rb
```diff
@@ -4,7 +4,6 @@
     managing_organisation { FactoryBot.create(:organisation) }
     trait :about_completed do
       gdpr_acceptance { "Yes" }
-      sale_or_letting { "Letting" }
```
paulrobertlloyd, line 7: Not sure if this is correct, as `sale_or_letting` might still be required for bulk uploads?

spec/fixtures/complete_case_log.json
```diff
@@ -126,7 +126,6 @@
     "gdpr_declined": "",
     "property_owner_organisation": "",
     "property_manager_organisation": "",
-    "sale_or_letting": "",
```
paulrobertlloyd, line 129: Not sure if this is correct, as `sale_or_letting` might still be required for bulk uploads?

spec/factories/case_log.rb
baarkerlounger: Fine for now I think

spec/fixtures/complete_case_log.json
baarkerlounger: Same

baarkerlounger:
Needs conflicts resolving but otherwise lgtm

PR #256, opened 2022-01-27
CLDC-633: Remove ‘Local authority’ section

paulrobertlloyd:
Removes the no longer needed ‘Local authority’ section. Notes:

* I have updated the existing JSON scheme to move the questions that were in this section to the respective section, but not edited any of the questions themselves.
* I considered updating the tests, whose fixture has a Local authority section, but it was causing all sorts of issues. Given that the tests test the JSON, not the content, this should be fine?


baarkerlounger:
Needs conflicts resolving but otherwise if you've tested :+1:

baarkerlounger (approved):

PR #261, opened 2022-01-31
Cldc 946 output field

kosiakkatrina:
Change numeric output field to `<output>` tag


app/views/form/_numeric_output_question.html.erb
```diff
@@ -0,0 +1,22 @@
+<%= render partial: "form/guidance/#{question.guidance_partial}" if question.guidance_partial %>
```
baarkerlounger, line 1: Might be worth checking whether this would be accepted upstream? @paulrobertlloyd  what do you think?

baarkerlounger (approved):
Really nice!

app/webpacker/styles/application.scss
```diff
@@ -15,6 +15,7 @@ $govuk-image-url-function: frontend-image-url;
 @import 'task-list';
 @import 'tab-navigation';
 @import 'accessible-autocomplete';
+@import "questions";
```
paulrobertlloyd: Can we call this `input` and rename `_questions.scss` to `_input.scss` – best to keep file names the same as the namespace they contain.

app/views/form/_numeric_output_question.html.erb
paulrobertlloyd: I think it might! Something to think about after the DAC audit. But in the meantime, might be worth starting a thread on the design system backlog. I’ll add a task to do this.

PR #263, opened 2022-01-31
Add a storage service backed by the AWS S3 SDK

smeny:


app/services/storage_service.rb
```diff
@@ -0,0 +1,68 @@
...
+    )
+  end
+end
+
+class StorageConfiguration
+  attr_reader :access_key_id, :secret_access_key, :bucket_name, :region
+
+  def initialize(credentials)
+    @access_key_id = credentials[:aws_access_key_id]
+    @secret_access_key = credentials[:aws_secret_access_key]
```
baarkerlounger, line 57: I was expecting that Gov PaaS woudl abstract away this level of config in the same way it does for RDS. I take it it doesn't?

smeny: Unfortunately no, it is literally like this in the environment variable

baarkerlounger (approved):

PR #267, opened 2022-02-03
Setup rake task to import organisations from S3

smeny:


app/services/storage_service.rb
```diff
@@ -8,6 +8,14 @@ def initialize(paas_config_service, paas_instance_name)
     @client = create_client
   end
 
+  def list_files(folder)
+    file_list = []
+    @client.list_objects_v2(bucket: @configuration.bucket_name, prefix: folder).each do |response|
```
baarkerlounger: Is this equivalent?
 
```
@client.list_objects_v2(bucket: @configuration.bucket_name, prefix: folder).flat_map do |response|
   response.contents.map(&:key) 
end
```

db/migrate/20220203104800_add_unique_index_to_old_visible_id.rb
```diff
@@ -0,0 +1,3 @@
+class AddUniqueIndexToOldVisibleId < ActiveRecord::Migration[7.0]
+  add_index :organisations, :old_visible_id, unique: true
```
baarkerlounger, line 2: Why do we want an index on that?

app/services/import_service.rb
```diff
@@ -0,0 +1,49 @@
...
+      cbl_proportion_percentage: field_value(doc, "cbl-proportion-percentage"),
+      enter_affordable_logs: to_boolean(field_value(doc, "enter-affordable-logs")),
+      owns_affordable_logs: to_boolean(field_value(doc, "owns-affordable-rent")),
+      housing_registration_no: field_value(doc, "housing-registration-no"),
+      general_needs_units: field_value(doc, "general-needs-units"),
+      supported_housing_units: field_value(doc, "supported-housing-units"),
+      unspecified_units: field_value(doc, "unspecified-units"),
+      old_org_id: field_value(doc, "id"),
+      old_visible_id: field_value(doc, "visible-id"),
+    }, unique_by: "old_visible_id")
```
baarkerlounger: Nice, this is clever.

app/services/import_service.rb
```diff
@@ -0,0 +1,49 @@
+class ImportService
+  def initialize(storage_service)
+    @storage_service = storage_service
+  end
+
+  def update_organisations(folder)
```
baarkerlounger, line 7: Scenario:

We import Organisations 1 & 2 and they start using the new service. Organisation 1 updates their phone number in the new service. Organisation 3 is added to the S3 bucket and we run `update_organisations` again. 

Does Organisation 1's phone number get reset back to it's value in the old system? I think that's what would happen with the upsert right?

smeny: Correct, we assumed that we would have potential updates via the rake task. Happy to prevent updates instead if you think this is our normal use case.

app/services/storage_service.rb
smeny: Should be, I am not a huge fan of `map` to @kosiakkatrina despair 😂 
If you think this is more readable, I will update and run the tests.

app/services/import_service.rb
baarkerlounger: I think we need to prevent updates because the assumption is that once we've imported them into the new system they stop using the old system and use the new one exclusively (though we probably need to confirm this will be enforced by softwire).

Alternatively we could use the last edit date and take the most recent?

app/services/storage_service.rb
baarkerlounger: :scream: I prefer `map` to each as being more functional/idiomatic but purely personal style preference so happy either way :slightly_smiling_face:

kosiakkatrina: We'd need to flatten it tho if that's how we do it, right? Cause we'd get an array of arrays?

smeny: done

app/services/import_service.rb
smeny: Decided to prevent duplicates since we don't have a date in the XML (we would have to trust the modification date on the file instead).

baarkerlounger (approved):

kosiakkatrina (approved):

PR #268, opened 2022-02-03
CLDC-967: count inferred check answers answer as completed

kosiakkatrina:
Count inferred check answers answer as an answered answer :) and remove la unknown reason field

### DIFF
--- a/app/models/form/question.rb
+++ b/app/models/form/question.rb
@@ -83,12 +83,12 @@ class Form::Question
     return false if id == "gdpr_acceptance" && case_log[id] == "No"
     return answer_options.keys.any? { |key| case_log[key] == "Yes" } if type == "checkbox"
 
-    case_log[id].present? || !case_log.respond_to?(id.to_sym) || has_inferred_value?(case_log)
+    case_log[id].present? || !case_log.respond_to?(id.to_sym) || has_inferred_display_value?(case_log)
   end
 
 private
 
-  def has_inferred_value?(case_log)
+  def has_inferred_display_value?(case_log)
     inferred_check_answers_value.present? && case_log[inferred_check_answers_value["condition"].keys.first] == inferred_check_answers_value["condition"].values.first
   end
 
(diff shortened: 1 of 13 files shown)


app/models/form/question.rb
```diff
@@ -83,11 +83,15 @@ def completed?(case_log)
     return false if id == "gdpr_acceptance" && case_log[id] == "No"
     return answer_options.keys.any? { |key| case_log[key] == "Yes" } if type == "checkbox"
 
-    case_log[id].present? || !case_log.respond_to?(id.to_sym)
+    case_log[id].present? || !case_log.respond_to?(id.to_sym) || has_inferred_value?(case_log)
```
baarkerlounger: If it `has_inferred_value` shouldn't that value have been set and so shouldn't `case_log[id].present?` return true?

kosiakkatrina: oh, okay, maybe a weird wording. It's not actually an inferred value, it's an inferred value that gets displayed in check answers page, but doesn't get saved. In this case if postcode is known our check answers line would look like this:
Postcode: AB1 2CD

If the postcode is not known, we **won't** display this:
Postcode known?: No

Instead we will display:
Postcode: Not known


This caused an issue, cause `postcode` field doesn't actually get displayed to the user, nor is it inferred, but we do want to display it as not known.

Happy to walk through this if confusing

baarkerlounger: Right ok. I think I understand the scenario but the code is maybe not very self-describing there in terms of what case that's catching? 

A spec and a comment might be helpful there to demonstrate? Or else maybe renaming things a bit so it's clear it's a check answers displayed value rather than the same as other inferred ones?

kosiakkatrina: Renamed it to `inferred_display_value`, is that somewhat clearer or would we need some sort of comments still?

spec/models/form_spec.rb
```diff
@@ -108,7 +112,9 @@ def answer_local_authority(case_log)
       end
 
       it "returns the next incomplete section by cycling back around if next subsections are completed" do
-        answer_local_authority(case_log)
+        # answer_local_authority(case_log)
```
baarkerlounger: Is this comment needed?

baarkerlounger (approved):

spec/models/form_spec.rb
kosiakkatrina: no..

PR #269, opened 2022-02-03
Update styles on task list page

paulrobertlloyd:
* Only show the task list page’s skip to next section link on focus
* Include name of next incomplete section in that skip link
* Update task list styles to remove unneeded styles and sync with latest prototype classes
* ~~Tidy up imports in `application.scss`~~ #272

No focus:
Screenshot 2022-02-03 at 15 17 06

With focus:
Screenshot 2022-02-03 at 15 17 19

### DIFF
--- a/app/views/case_logs/_tasklist.html.erb
+++ b/app/views/case_logs/_tasklist.html.erb
@@ -1,10 +1,8 @@
-<ol class="app-task-list app-task-list--no-numbers">
+<ol class="app-task-list govuk-!-margin-top-8">
   <% @case_log.form.sections.map do |section| %>
     <li>
-      <h2 class="app-task-list__section">
-        <span class="app-task-list__section-number">
-          <%= section.label %>
-        </span>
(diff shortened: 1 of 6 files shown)


baarkerlounger:
> Only show skip section to next section link on focus

Where is that happening and what's the reasoning behind it?

Do you think you could split out the CSS import tidy up changes into a separate PR? Want to see how those affect the PR that moves away from webpacker which will be easier if it's atomic I think.

paulrobertlloyd:
> Do you think you could split out the CSS import tidy up changes into a separate PR?

Will do!

paulrobertlloyd:
> Where is that happening and what's the reasoning behind it?

On the task list page. This link is an accessibility feature, to help users orientate to the next incomplete section and inform them which section that is. This link doesn’t need to be visibly shown because this information can be gleaned from looking at the task list.

baarkerlounger (approved):

PR #270, opened 2022-02-03
2fa fixes

baarkerlounger:
- [x] Change 2FA expiry time from 5 minutes to 15 (consistent with Notify OTP)
- [x] Change template to add 15 min expiry time (consistent with Notify OTP)
- [x] Change OTP code length from 6 to 5 (consistent with Notify OTP)
- [x] Require 2FA on every sign in
- [x] Show errors when invalid or no OTP code is entered

When 2FA code is not entered:
image

When 2FA code is incorrect:

image

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -20,7 +20,7 @@ GIT
 
 GIT
   remote: https://github.com/baarkerlounger/two_factor_authentication.git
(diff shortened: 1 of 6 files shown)


kosiakkatrina (approved):
Looks good! Would we want to log failed code entries in the future for stats? 👀

baarkerlounger:
@kosiakkatrina possibly but I think that's out of scope here for now

PR #271, opened 2022-02-03
Enable section descriptions and add note about privacy notice

paulrobertlloyd:
* Adds a new field for a section: `description` – this is shown under the respective section label in the task list
* Adds test for the above
* Adds description for the ‘About the household’ section which shows a note about the tenant needing to see the privacy notice.

Screenshot 2022-02-03 at 15 46 29

### DIFF
--- a/app/views/case_logs/_tasklist.html.erb
+++ b/app/views/case_logs/_tasklist.html.erb
@@ -6,6 +6,9 @@
           <%= section.label %>
         </span>
       </h2>
+      <% if section.description %>
+        <p class="govuk-body"><%= section.description.html_safe %></p>
+      <% end %>
       <ul class="app-task-list__items">
         <% section.subsections.map do |subsection| %>
           <% subsection_status = subsection.status(@case_log) %>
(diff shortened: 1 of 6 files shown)


app/views/case_logs/_tasklist.html.erb
```diff
@@ -6,6 +6,9 @@
           <%= section.label %>
         </span>
       </h2>
+      <% if section.description %>
+        <p class="govuk-body"><%= section.description.html_safe %></p>
```
baarkerlounger, line 10: Should this not be slightly smaller/hint text coloured? Looks very prominent in the screenshot. (Gets back into own lane...)

baarkerlounger (approved):

app/views/case_logs/_tasklist.html.erb
paulrobertlloyd: Hint text should be avoided, where possible (the dark grey colour has a known issues), and anyway, we want users to see this message.

PR #273, opened 2022-02-04
CLDC-933 update household income questions

dushan-madetech:
**New options for do you know income**
Screenshot 2022-02-07 at 16 33 36

**New page with earnings and frequency questions on the same page**
Screenshot 2022-02-07 at 16 44 16

**Check answers page when the earnings & frequency questions are unanswered**
Screenshot 2022-02-07 at 16 46 26

**Check answers page when the earnings & frequency questions are answered**
Screenshot 2022-02-04 at 15 20 03

**Validation for earnings & frequency questions to prevent submitting one but not the other**
Screenshot 2022-02-07 at 16 47 43
Screenshot 2022-02-07 at 16 47 58


app/helpers/check_answers_helper.rb
```diff
@@ -12,12 +12,25 @@ def display_answered_questions_summary(subsection, case_log)
   end
 
   def get_answer_label(question, case_log)
-    answer = question.prefix == "£" ? ActionController::Base.helpers.number_to_currency(question.answer_label(case_log), delimiter: ",", format: "%n") : question.answer_label(case_log)
-
-    if answer.present?
-      [question.prefix, answer, question.suffix].join("")
+    if question.id == "earnings"
```
baarkerlounger: Is there a more generic way of doing this? I think if we start hard coding special cases like this in, then pretty soon the whole JSON definition idea becomes a bit useless and we might as well make every page a bespoke view?

app/models/form/question.rb
```diff
@@ -75,6 +75,9 @@ def update_answer_link_name(case_log)
                 else
                   case_log[id].blank? ? "Answer" : "Change"
                 end
+    if id == "earnings"
```
baarkerlounger: Same here really. Is there some property of this question that means it gets treated differently that we can support in the JSON DSL?

dushan-madetech: The way it gets displayed in the check answers - I wasn't sure of how to get link the earnings to the frequency but modify the suffix based on the options. Since the options are `weekly`, `monthly` and `yearly` and I need that to become `every week`, `every month` and `every year` on the check answers page. Suffix is currently just coded to write out whatever gets put in the suffix field as a string but I guess it could be done without this if we could modify the suffix bit

dushan-madetech: I might need some help with that though since I couldn't really think of a nice way to do it

baarkerlounger: You could add a `depends_on` key to the suffix similar to how it works for pages and sections? So you'd have something like:

```
"scharge": {
                  "check_answer_label": "Service charge",
                  "header": "What is the service charge?",
                  "hint_text": "For example, for cleaning. Households may receive housing benefit or Universal Credit towards the service charge.",
                  "type": "numeric",
                  "min": 0,
                  "step": 1,
                  "width": 5,
                  "prefix": "£",
                  "suffix": [ 
                    { "label": " every week", "depends_on" : [{ "incfreq": "weekly"}]},
                    { "label": " every month", "depends_on" : [{ "incfreq": "monthly"}]}
                  ],
                  "fields-to-add": ["brent", "scharge", "pscharge", "supcharg"],
                  "result-field": "tcharge",
                  "hidden_in_check_answers": true
                },
```

dushan-madetech: I'll have a crack at this suggestion, could I potentially look at with you on Monday if I haven't got it working by then?

baarkerlounger (approved):

paulrobertlloyd (approved):
Great stuff – thanks @dushan-madetech!

Dan13L3:
@dushan-madetech This looks good - thanks. The only things that need updating are the error messages (I'm only just getting to review quite a few of them, so they could be old versions). They should be:
1) ‘Select how often the household receives income’
2) 'Enter how much income the household has in total'

dushan-madetech:
@Dan13L3 They were newly added to this change. I've updated them now as per your recommendation 😄 
1.Screenshot 2022-02-08 at 09 33 23
2.Screenshot 2022-02-08 at 09 33 33

baarkerlounger (approved):

Dan13L3:
@dushan-madetech perfect - thanks

PR #277, opened 2022-02-07
CLDC-961: Model data versioning

baarkerlounger:
This creates a new record in the Versions table every time a model with `has_paper_trail` is updated. https://github.com/paper-trail-gem/paper_trail.

This does potential mean an explosion in data size since CaseLog is updated every time a form question is submitted. Given the form currently has ~70 pages, assuming some question are amended - we're likely to see ~100 records (tracking each change) for each case log where previously we only had 1. Worth considering whether we want to track this aggressively or whether we should be more selective on versioning. For example we could choose to only version if the change is >24 hours after the last version was created or only version on specific field etc. Alternatively we could only keep x number of version per model, or versions up to a certain age.

This PR removes Discard (gem for soft delete functionality), as we don't really need it any more now that deletes are versioned and can be restored from Paper Trail.

- [x] Use PaperTrail to version data models (User, Admin User, Organisation, Case Log)
- [x] Track whodunnit on form UI changes
- [x] Track whodunnit on Admin Panel changes

Out of scope: 
1. Track whodunnit on API changes (since we don't have individual API users yet)
2. Tracking whodunnit on changes made from Rails console or via Rake tasks
3. Display versions in Active Admin (Subsequent PR if required to keep this small and atomic)

### DIFF
--- a/spec/requests/case_logs_controller_spec.rb
+++ b/spec/requests/case_logs_controller_spec.rb
@@ -401,9 +401,8 @@ RSpec.describe CaseLogsController, type: :request do
         expect(response).to have_http_status(:success)
       end
 
-      it "soft deletes the case log" do
+      it "deletes the case log" do
         expect { CaseLog.find(id) }.to raise_error(ActiveRecord::RecordNotFound)
-        expect(CaseLog.with_discarded.find(id)).to be_a(CaseLog)
       end
 
       context "with an invalid case log id" do
@@ -428,7 +427,7 @@ RSpec.describe CaseLogsController, type: :request do
     context "when a case log deletion fails" do
       before do
         allow(CaseLog).to receive(:find_by).and_return(case_log)
-        allow(case_log).to receive(:discard).and_return(false)
+        allow(case_log).to receive(:delete).and_return(false)
         delete "/logs/#{id}", headers: headers
       end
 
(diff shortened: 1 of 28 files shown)


spec/requests/case_logs_controller_spec.rb
```diff
@@ -34,84 +34,104 @@
...
+        it "returns 401" do
+          expect(response).to have_http_status(:unauthorized)
+        end
       end
     end
 
-    context "with a request containing invalid credentials" do
-      let(:basic_credentials) do
-        ActionController::HttpAuthentication::Basic.encode_credentials(api_username, "Oops")
+    context "when UI" do
```
smeny, line 120: Not a big deal but I think we should export our UI tests outside this file at some point (not now).
The "when API" and "when UI" are not really contexts, more how we test the application (at least to me). But I like that it logically splits the code which would make a refactor easier.

smeny (approved):

spec/requests/case_logs_controller_spec.rb
baarkerlounger: They kind of are contexts (I could also have said: "when request header is json" and "when request header is html") but happy to refactor. Will leave that out of scope for this PR though.

PR #279, opened 2022-02-08
Cldc 972 submitting lettings log

kosiakkatrina:
Display error message is the declaration box isn't ticked
Route user back to the your logs page (skipping check your answers)
Change the button for submit your log section from `Save and continue` to `submit lettings log`

image


app/controllers/form_controller.rb
```diff
@@ -8,8 +8,12 @@ def submit_form
       @page = @case_log.form.get_page(params[:case_log][:page])
       responses_for_page = responses_for_page(@page)
       if @case_log.update(responses_for_page) && @case_log.has_no_unresolved_soft_errors?
-        redirect_path = @case_log.form.next_page_redirect_path(@page, @case_log)
-        redirect_to(send(redirect_path, @case_log))
+        if @page.id == "declaration"
```
baarkerlounger: I don't think we should ever hard code page ids if we can avoid it. Effectively couples behavior to form definition again. Could we instead check if the page is the last page in the form?

Should maybe also be in the `next_page_redirect_path` method itself?

app/views/form/page.html.erb
```diff
@@ -47,7 +47,11 @@
         <% end %>
 
         <%= f.hidden_field :page, value: @page.id %>
-        <%= f.govuk_submit "Save and continue"  %>
+        <% if @page.id == "declaration" %>
```
baarkerlounger: Same here, maybe a helper that checks if it's the last page?

app/controllers/form_controller.rb
kosiakkatrina: Hmm, agree with the first point. But if we put it in the `next_page_redirect_path` we then use whatever is returned within `redirect_to(send(redirect_path, @case_log))` which sends us to /logs.#{log_id}, I assume because we pass `@case_log` through

spec/factories/case_log.rb
```diff
@@ -152,6 +152,128 @@
       chcharge { 7 }
       letting_in_sheltered_accomodation { "No" }
       la_known { "Yes" }
+      declaration { "Yes" }
+    end
+    trait :completed_without_declaration do
```
baarkerlounger: Nested factory/traits might be better here to avoid duplicating all the data just to change one field? (or even just flipping the field in your test?)

kosiakkatrina: That's very fair, was struggling to do that yesterday cause the validation kept on going off. Clearly was missing something cause it seems okay now. Thank you thank you!

app/controllers/form_controller.rb
kosiakkatrina: Is it okay to use ID in the model? we're using it `next_subsection` method already
edit: ignore me

baarkerlounger: Ideally I think we should remove it there too

config/forms/2021_2022.json
```diff
@@ -3226,9 +3226,12 @@
               "questions": {
                 "declaration": {
                   "check_answer_label": "",
-                  "header": "What is the tenant code?",
+                  "header": "Submit your lettings log ",
                   "hint_text": "",
-                  "type": "text"
+                  "type": "checkbox",
+                  "answer_options": {
+                    "declaration": "The tenant has seen the Department for Levelling Up, Housing & Communities (DLUHC) privacy notice"
```
baarkerlounger, line 3195: Does this now need to be removed from the About this log section?

app/models/form/question.rb
```diff
@@ -80,7 +80,6 @@ def update_answer_link_name(case_log)
 
   def completed?(case_log)
     # Special case as No is a valid answer but doesn't let you progress and use the service
-    return false if id == "gdpr_acceptance" && case_log[id] == "No"
```
baarkerlounger, line 83: :heart:

baarkerlounger: Can we remove the comment also?

spec/models/form_spec.rb
```diff
@@ -111,11 +107,6 @@ def answer_local_gdpr_acceptance(case_log)
         expect(form.next_incomplete_section_redirect_path(subsection, case_log)).to eq("tenancy-code")
       end
 
-      it "returns the next incomplete section by cycling back around if next subsections are completed" do
```
baarkerlounger, line 114: Does this test need replacing? Looks like it was a valid test case.

baarkerlounger (approved):

PR #280, opened 2022-02-08
CLDC-835: support hint on radio/checkbox questions

kiddhustle:
amend JSON to support hint

https://digital.dclg.gov.uk/jira/browse/CLDC-835

@paulrobertlloyd please see screenshots below
CLDC-835_checkbox_radio
CLDC-835_checkbox

### DIFF
--- a/app/models/form/question.rb
+++ b/app/models/form/question.rb
@@ -40,7 +40,6 @@ class Form::Question
 
     answer = case_log[id].to_s if case_log[id].present?
     answer_label = [prefix, format_value(answer), suffix_label(case_log)].join("") if answer
-
     return answer_label if answer_label
 
     has_inferred_check_answers_value?(case_log) ? inferred_check_answers_value["value"] : ""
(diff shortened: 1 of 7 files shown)


db/schema.rb
```diff
@@ -10,34 +10,33 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 202202071123100) do
-
+ActiveRecord::Schema[7.0].define(version: 202202071123100) do
```
baarkerlounger: Can you revert these schema changes since there's no migration here?

kiddhustle: yea I think these migration changes came in when I rebased main.

Wasn't sure if I should commit them or not I can undo

baarkerlounger (approved):

db/schema.rb
baarkerlounger: General rule of thumb is commit them if you've intended to change the schema (by writing a migration), but not if it's unintended (for example because your local database was in a different state etc).

paulrobertlloyd:
LGTM! Thanks @kiddhustle! 🚢

paulrobertlloyd (approved):

PR #281, opened 2022-02-08
Implement exclusive checkboxes

kosiakkatrina:
Set all the checkboxes after a divider to be exclusive

### DIFF
--- a/app/views/form/_checkbox_question.html.erb
+++ b/app/views/form/_checkbox_question.html.erb
@@ -4,14 +4,17 @@
     caption: caption(caption_text, page_header, conditional),
     legend: legend(question, page_header, conditional),
     hint: { text: question.hint_text&.html_safe } do %>
+    <% after_divider = false %>
 
     <% question.answer_options.map do |key, val| %>
       <% if key.starts_with?("divider") %>
-        <%= f.govuk_check_box_divider %>
+      <% after_divider = true %>
+      <%= f.govuk_check_box_divider %>
       <% else %>
         <%= f.govuk_check_box question.id, key,
           label: { text: val },
           checked: @case_log[key] == "Yes",
+          exclusive: after_divider,
           **stimulus_html_attributes(question)
         %>
       <% end %>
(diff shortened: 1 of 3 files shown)


app/views/form/_checkbox_question.html.erb
```diff
@@ -4,14 +4,17 @@
     caption: caption(caption_text, page_header, conditional),
     legend: legend(question, page_header, conditional),
     hint: { text: question.hint_text&.html_safe } do %>
+    <% after_divider = false %>
 
     <% question.answer_options.map do |key, val| %>
       <% if key.starts_with?("divider") %>
-        <%= f.govuk_check_box_divider %>
+      <% after_divider = true %>
```
baarkerlounger, line 11: Interesting approach! Is it definitely always the case that anything separated by a divider is mutually exclusive? @paulrobertlloyd? If so, then :very_nice:

kosiakkatrina: Hmm, well our dividers put "or" between the options, so I'd assume so 👀

baarkerlounger:
Think we also need to add backend validations for this for when JS is turned off and/or API submission?

kosiakkatrina:
> Think we also need to add backend validations for this for when JS is turned off and/or API submission?

@baarkerlounger Yeah, I think we've got those defined in separate tickets. One of them is already implemented, but thanks for bringing it up, cause it actually needed fixing (done now)

app/views/form/_checkbox_question.html.erb
paulrobertlloyd: Yup, we’d want to separate out those options that are mutually exclusive after a divider. See https://design-system.service.gov.uk/components/checkboxes/#add-an-option-for-none-

baarkerlounger (approved):

app/views/form/_checkbox_question.html.erb
paulrobertlloyd: Whether we’d use a divider for other use cases with checkboxes I don’t know, but I suspect not. This is a good assumption to go forward with, for now.

PR #282, opened 2022-02-08
CLDC-966-optional-setup-fields

kosiakkatrina:
Add `tenant_code` and `propcode` as optional fields and consider subsection completed it they are not filled in
This will allow to start the following sections without filling out the whole setup section.


app/models/form/subsection.rb
```diff
@@ -30,7 +32,7 @@ def status(case_log)
       return :cannot_start_yet
     end
 
-    qs = applicable_questions(case_log)
+    qs = applicable_questions(case_log).reject { |q| OPTIONAL_FIELDS.include?(q.id) }
```
baarkerlounger, line 35: Worth adding a spec for this?

baarkerlounger (approved):

PR #287, opened 2022-02-09
Production deployment action

baarkerlounger:
These actions make the following assumptions about workflow:

1. Staging has continuous CI/CD - all pushes to main automatically run the lint, test, deploy pipeline
2. Production deploy is triggered automatically when a release is made or manually 
3. The production action still needs CI (lint, test) in case of hotfixes (where the tag has been branched off)


.github/workflows/production_pipeline.yml
```diff
@@ -0,0 +1,133 @@
...
+    env:
+      RAILS_ENV: test
+      GEMFILE_RUBY_VERSION: 3.0.0
+      DB_HOST: localhost
+      DB_DATABASE: data_collector
+      DB_USERNAME: postgres
+      DB_PASSWORD: password
+
+      # Rails verifies the time zone in DB is the same as the time zone of the Rails app
+      TZ: "Europe/London"
```
smeny: Do you know why is this an issue for our tests, when we have UTC for the DB in staging?

.github/workflows/production_pipeline.yml
```diff
@@ -0,0 +1,133 @@
...
+          CF_API_ENDPOINT: ${{ secrets.CF_API_ENDPOINT }}
+          CF_SPACE: ${{ secrets.CF_SPACE }}
+          CF_ORG: ${{ secrets.CF_ORG }}
+          API_USER: ${{ secrets.API_USER }}
+          API_KEY: ${{ secrets.API_KEY }}
+          APP_NAME: dluhc-core-staging
+          GOVUK_NOTIFY_API_KEY: ${{ secrets.GOVUK_NOTIFY_API_KEY }}
+          APP_HOST: ${{ secrets.APP_HOST }}
+          RAILS_MASTER_KEY: ${{ secrets.RAILS_MASTER_KEY }}
+          IMPORT_PAAS_INSTANCE: ${{ secrets.IMPORT_PAAS_INSTANCE }}
```
smeny, line 173: you might want to add the extra env variables added to staging

.github/workflows/production_pipeline.yml
baarkerlounger: Good question, I'm actually not sure that's doing anything here...

.github/workflows/production_pipeline.yml
```diff
@@ -0,0 +1,133 @@
...
+    name: Test
+    runs-on: ubuntu-latest
+
+    services:
+      postgres:
+        image: postgres:13.5
+        env:
+          POSTGRES_PASSWORD: password
+          POSTGRES_USER: postgres
+          POSTGRES_DB: data_collector
```
dushan-madetech, line 28: I'm assuming we don't need to hide these?

.github/workflows/production_pipeline.yml
```diff
@@ -0,0 +1,133 @@
...
+        # needed because the postgres container does not provide a healthcheck
+        # tmpfs makes DB faster by using RAM
+        options: >-
+          --mount type=tmpfs,destination=/var/lib/postgresql/data
+          --health-cmd pg_isready
+          --health-interval 10s
+          --health-timeout 5s
+          --health-retries 5
+    env:
+      RAILS_ENV: test
```
dushan-madetech, line 40: What's the reason for using the test env in the production pipeline?

dushan-madetech (approved):
I left some questions mainly for my own understanding, otherwise this seems fine to me 👍

.github/workflows/production_pipeline.yml
baarkerlounger: Not really, they're only used within this container to run the test stage, they're not copied to prod environment or anything like that

.github/workflows/production_pipeline.yml
baarkerlounger: This is just running rspec - which assumes test environment

.github/workflows/production_pipeline.yml
```diff
@@ -0,0 +1,133 @@
...
+
+      - name: Rubocop
+        run: |
+          bundle exec rubocop
+
+    deploy:
+      name: Deploy
+
+      runs-on: ubuntu-latest
+      environment: 'production'
```
baarkerlounger, line 138: @dushan-madetech you're maybe looking for this?

dushan-madetech: Yep, that makes sense 👍

.github/workflows/production_pipeline.yml
```diff
@@ -0,0 +1,185 @@
...
+      - name: Checkout tag
+        uses: actions/checkout@v2
+        with:
+          ref: ${{ steps.latestrelease.outputs.releasetag }}
+
+      - name: Install cf client
+        env:
+          CF_CLI_VERSION: 7.0.0-beta.30
+        run: |
+          curl -L "https://packages.cloudfoundry.org/stable?release=linux64-binary&version=${CF_CLI_VERSION}" | tar -zx -C /tmp
```
smeny: should we update the cloud foundry install based on the monitoring work?

baarkerlounger: Good idea. Done.

.github/workflows/production_pipeline.yml
```diff
@@ -0,0 +1,185 @@
...
+      DB_DATABASE: data_collector
+      DB_USERNAME: postgres
+      DB_PASSWORD: password
+
+    steps:
+      - name: Get latest release with tag
+        id: latestrelease
+        run: |
+          echo "::set-output name=releasetag::$(curl -s https://api.github.com/repos/weide-zhou/ticket13/releases/latest | jq '.tag_name' | sed 's/\"//g')"
+      - name: Confirm release tag
```
smeny, line 52: Looking at https://docs.github.com/en/rest/reference/releases#get-the-latest-release, should we use /repos/{owner}/{repo}/releases/latest ?

smeny: Work checking https://docs.github.com/en/actions/learn-github-actions/environment-variables as well (API reference etc)

baarkerlounger: We are using that first one - I just had the wrong repo in there

smeny (approved):

PR #288, opened 2022-02-09
Case logs export

smeny:


spec/spec_helper.rb
```diff
@@ -108,4 +109,7 @@
   #   Kernel.srand config.seed
 
   config.include FactoryBot::Syntax::Methods
+
+  config.before { RequestHelper.stub_http_requests }
```
baarkerlounger, line 113: This is a nice refactor

baarkerlounger (approved):

baarkerlounger:
Couple of thoughts:

- Think we'll need zipping before this gets used so we have the file encrypted at rest in S3
- Is there any danger here of losing data that's written to DB after the manifest has been created but before the export to XML has completed?

PR #289, opened 2022-02-09
Add sentry configuration

smeny:


Gemfile
```diff
@@ -49,6 +49,9 @@ gem "aws-sdk-s3"
 gem "paper_trail"
 # Store active record objects in version whodunnits
 gem "paper_trail-globalid"
+# Receive exceptions and configure alerts
+gem "sentry-rails", "~> 5.0", ">= 5.0.2"
```
baarkerlounger: Do we need the min and the twiddle waka?

smeny: Gone

config/initializers/sentry.rb
```diff
@@ -0,0 +1,7 @@
+Sentry.init do |config|
+  config.breadcrumbs_logger = %i[active_support_logger http_logger]
+
+  # To activate performance monitoring, set one of these options.
+  # We recommend adjusting the value in production:
+  config.traces_sample_rate = 0.5
```
baarkerlounger, line 4: Do we want to be configuring environments in here? Something like:

```
config.environments = %w[staging production]
```

smeny: It is not needed, a lot is actually fetched from Rails.
I would leave the default which uses RAILS_ENV.
Happy to add the other one.

baarkerlounger (approved):

config/initializers/sentry.rb
baarkerlounger: Sorry that was just a question. Happy to leave the default as well.

PR #291, opened 2022-02-10
Derive providertype from org

kosiakkatrina:


db/schema.rb
```diff
@@ -21,8 +20,8 @@
     t.string "reset_password_token"
     t.datetime "reset_password_sent_at"
     t.datetime "remember_created_at"
-    t.datetime "created_at", precision: 6, null: false
```
baarkerlounger: These make me nervous. Can you find out where they come from or revert?

app/models/constants/case_log.rb
```diff
@@ -1125,4 +1125,9 @@ module Constants::CaseLog
                             "Other"].freeze
 
   OPTIONAL_FIELDS = %w[postcode_known la_known first_time_property_let_as_social_housing tenant_code propcode].freeze
+
+  PROVIDER_TYPE = {
```
smeny: We could reuse the organisation constants here

db/schema.rb
```diff
@@ -191,6 +190,7 @@
     t.decimal "tshortfall", precision: 10, scale: 2
     t.decimal "chcharge", precision: 10, scale: 2
     t.integer "declaration"
+    t.integer "providertype"
```
smeny, line 193: Why is this not fetched from the organisation directly? I think that adding it like this creates a duplication where we could have the organisation and the case logs having different values.

db/schema.rb
kosiakkatrina: It had precision when it was created I'll revert it

PR #292, opened 2022-02-10
Remove checkbox parent fields

baarkerlounger:
For checkbox questions (Accessbility requirements and condition effects) we have individual boolean fields set for each option:

```
housingneeds_a
housingneeds_b
housingneeds_c
housingneeds_f
housingneeds_g
housingneeds_h

illness_type_1
illness_type_2
illness_type_3
illness_type_4
illness_type_5
illness_type_6
illness_type_7
illness_type_8
illness_type_9
illness_type_10
```

So the parent field is never set, but is still checked against for determining Case Log status. Meaning case logs incorrectly don't get set as completed because those two fields aren't, even though they'll never be. So removing them since they're not needed at all.

The schema changes are a result of Rails now versioning schemas by Rails version as off 7.0.2 (https://rubyonrails.org/2022/2/8/Rails-7-0-2-has-been-released - https://github.com/rails/rails/pull/44286).


db/migrate/20220210132254_remove_checkbox_parent_fields.rb
```diff
@@ -0,0 +1,15 @@
+class RemoveCheckboxParentFields < ActiveRecord::Migration[7.0]
+  def up
+    change_table :case_logs, bulk: true do |t|
+      t.remove :accessibility_requirements
+      t.remove :condition_effects
```
kosiakkatrina, line 5: Would it be the same for `reasonable_preference_reason`?

baarkerlounger: Yes, good catch!

db/schema.rb
```diff
@@ -10,34 +10,33 @@
...
+    t.datetime "direct_otp_sent_at", precision: nil
+    t.datetime "totp_timestamp", precision: nil
     t.string "phone"
     t.index ["encrypted_otp_secret_key"], name: "index_admin_users_on_encrypted_otp_secret_key", unique: true
   end
 
   create_table "case_logs", force: :cascade do |t|
     t.integer "status", default: 0
-    t.datetime "created_at", precision: 6, null: false
-    t.datetime "updated_at", precision: 6, null: false
```
kosiakkatrina, line 40: 🤔

baarkerlounger: It looks like this is Rails 7.0.2 removing something that was only ever meant to target MySQL https://github.com/rails/rails/issues/37031

kosiakkatrina (approved):

PR #293, opened 2022-02-10
CLDC-900: Update order of questions in ‘Set up your lettings log’ section

paulrobertlloyd:
* ‘What is the needs type’ is asked after ‘Organisation details’, separated from rent type. (In future releases, selecting ‘Supported housing’ will route users onto a supported housing branch)
* ‘Tell us about this letting’ is renamed ‘What is the rent type?’ and only asks for rent type
* The ‘Check your answers’ page is updated to reflect these changes and revised ordering

Also…
* Correct ordering and capitalisation of options.

### DIFF
--- a/config/forms/2021_2022.json
+++ b/config/forms/2021_2022.json
@@ -43,6 +43,26 @@
                 }
               }
             },
+            "needs_type": {
+              "header": "",
+              "description": "",
+              "questions": {
+                "needstype": {
+                  "check_answer_label": "Needs type",
+                  "header": "What is the needs type?",
+                  "hint_text": "",
+                  "type": "radio",
+                  "answer_options": {
+                    "1": {
+                      "value": "General needs"
+                    },
+                    "0": {
+                      "value": "Supported housing"
+                    }
+                  }
+                }
+              }
+            },
             "renewal": {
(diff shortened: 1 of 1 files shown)


config/forms/2021_2022.json
```diff
@@ -85,20 +105,20 @@
...
                     "2": {
-                      "value": "London affordable rent"
+                      "value": "London Affordable Rent"
+                    },
+                    "4": {
+                      "value": "London Living Rent"
                     },
                     "3": {
-                      "value": "Rent to buy"
+                      "value": "Rent to Buy"
```
baarkerlounger, line 118: Isn't this inconsistent with how we're capitalising things elsewhere? I thought it was supposed to be regular sentence case everywhere? @Dan13L3

Dan13L3: @baarkerlounger In most cases it should be sentence case, yes - but we should capitalise any specific government schemes like this

baarkerlounger (approved):

config/forms/2021_2022.json
paulrobertlloyd: This is the name of a scheme, so I believe.

PR #294, opened 2022-02-10
Bugfix: Default charge values to 0

baarkerlounger:
Currently if you answer any of the charge fields the subsection assumes you've answered that question and marks it as complete but the form doesn't. This defaults all unanswered values on that page to 0 if any have been answered so that the behaviour is consistent. i.e. so that you don't get all sections complete but form incomplete because it thinks those fields still need answering.


kosiakkatrina (approved):
Nice!

PR #295, opened 2022-02-10
Don't validate bedroom number if not answered yet

baarkerlounger:
Fixes bug where if you select Shared house or Shared bungalow you can't progress because it expects a bedroom number between 1 and 7 and you haven't answered that question yet.

### DIFF
--- a/app/models/validations/household_validations.rb
+++ b/app/models/validations/household_validations.rb
@@ -70,22 +70,6 @@ module Validations::HouseholdValidations
     end
   end
 
-  def validate_shared_housing_rooms(record)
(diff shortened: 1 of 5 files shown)


kosiakkatrina (approved):
I take it this is just adding .present? to the second validation and the rest is refactoring?

baarkerlounger:
@kosiakkatrina yes exactly. The idea of the refactoring is to break it up into better unit tests bit by bit. Eventually the case log model spec should just check that the right validation methods are called and the validator specs can test the specific individual logic rules.

PR #296, opened 2022-02-10
Fix tenancy validations

baarkerlounger:
Fixes two bugs:

1. Tenancy validation is a compound validation - it validates that the tenancy type and tenancy length are valid together but we were only adding the error to the length field so if you made the incompatible change on the tenancy type page you wouldn't see the error. This PR adds it to both.

2. For assured shorthold tenancy it was trying to validate the length before you answered length (meaning if you picked assured shorthold you couldn't click save and continue to progress to the next question). This PR fixes that.


kosiakkatrina (approved):
Smashing it

smeny (approved):

PR #297, opened 2022-02-11
Refactor validation specs

baarkerlounger:
The idea here is to make the tests smaller units since the Case Log spec at the moment does a lot and is large. 

This would mean that when testing validations, the case log spec is only responsible for checking that the correct validation methods are called while the individual Validator class specs are responsible for testing the actual logic and edge cases of those validations.

### DIFF
--- a/app/models/validations/household_validations.rb
+++ b/app/models/validations/household_validations.rb
@@ -6,6 +6,7 @@ module Validations::HouseholdValidations
   def validate_reasonable_preference(record)
     if record.homeless == "No" && record.reasonpref == "Yes"
       record.errors.add :reasonpref, I18n.t("validations.household.reasonpref.not_homeless")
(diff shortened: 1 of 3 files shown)


smeny (approved):

app/models/validations/household_validations.rb
```diff
@@ -6,6 +6,7 @@ module Validations::HouseholdValidations
   def validate_reasonable_preference(record)
     if record.homeless == "No" && record.reasonpref == "Yes"
       record.errors.add :reasonpref, I18n.t("validations.household.reasonpref.not_homeless")
+      record.errors.add :homeless, I18n.t("validations.household.reasonpref.not_homeless")
```
smeny, line 9: Is this a duplicate?

baarkerlounger: Yes, same error message but added on a different field (since those fields appear on different pages we want the error linked no matter which page you've triggered the error from).

PR #298, opened 2022-02-11
Cldc 629 household situation

kosiakkatrina:
Update routing and content for household situation not renewal questions


app/models/form/page.rb
```diff
@@ -49,7 +49,7 @@ def depends_on_met(case_log)
 
     depends_on.any? do |conditions_set|
       conditions_set.all? do |question, value|
-        value.nil? ? case_log[question] == value : !case_log[question].nil? && case_log[question] == value
+        value.nil? ? eval("case_log.#{question}") == value : !eval("case_log.#{question}").nil? && eval("case_log.#{question}") == value
```
kosiakkatrina: Not sure about this one, but couldn't think of another way of getting nested/org field

baarkerlounger: Why does this work but the original didn't? I don't think rubocop will like eval.

kosiakkatrina: Yeah, it didn't.
Well, the original way to access the field value would be:
`case_log[key]`
but if we want to access `case_log.managing_organisation.provider_type` that would translate to `case_log["managing_organisation.provider_type"]` maybe I'm missing an obvious way to make it work here 🤔

baarkerlounger: Something along the lines of:

```
case_log_value = if question.include?(".")
  parts = question.split(".")
  case_log.public_send(parts.first.to_sym).public_send(parts.second.to_sym)
else
  case_log[question]
end

value.nil? ? case_log_value == value : !case_log_value.nil? && case_log_value == value
```

baarkerlounger: Can probably refactor that nicer to make it accept any number of chained methods but should work as an approach?

kosiakkatrina: That's great! Better now?

app/models/form/page.rb
```diff
@@ -44,12 +44,19 @@ def conditional_question_ids
     }.compact
   end
 
+  def send_chain(arr, case_log)
+    Array(arr).inject(case_log) { |o, a| o.public_send(*a) }
```
baarkerlounger, line 48: You got to love Ruby

spec/models/case_log_spec.rb
```diff
@@ -1087,6 +1087,64 @@ def check_rsnvac_referral_validation(referral)
...
+        })
+      end
+
+      it "correctly infers prevloc" do
+        record_from_db = ActiveRecord::Base.connection.execute("select prevloc from case_logs where id=#{address_case_log.id}").to_a[0]
+        expect(address_case_log.prevloc).to eq("Manchester")
+        expect(record_from_db["prevloc"]).to eq("E08000003")
+      end
+
+      it "errors if the previous postcode is emptied" do
```
baarkerlounger, line 1111: This seems wrong? What if you realise you input an incorrect one and actually don't know it?

kosiakkatrina: hmm, this is how we're doing it for the current postcode. I think the point is that `No` should be selected for `postcode known`

baarkerlounger: Ah I see, makes sense

baarkerlounger (approved):

PR #299, opened 2022-02-11
Start date validation

baarkerlounger:
Validate that tenancy start date is within collection windows (i.e. we actually have a form for it)

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -35,7 +35,9 @@ en:
     updated: "Organisation details updated"
 
   validations:
-    date: "Please enter a valid date"
+    date:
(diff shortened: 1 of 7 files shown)


config/locales/en.yml
```diff
@@ -35,7 +35,9 @@ en:
     updated: "Organisation details updated"
 
   validations:
-    date: "Please enter a valid date"
+    date:
+      invalid_date: "Please enter a valid date"
+      outside_collection_window: "Date must be within the current collection windows"
```
baarkerlounger, line 40: @DAN13L3 this is the date validation error message.

PR #303, opened 2022-02-14
Refactor Reason for leaving last settled home validations

baarkerlounger:


Gemfile
```diff
@@ -54,17 +54,18 @@ gem "sentry-rails"
 gem "sentry-ruby"
 
 group :development, :test do
-  # Call 'byebug' anywhere in the code to stop execution and get a debugger console
+  # Check gems for known vulnerabilities
```
dushan-madetech, line 57: Does this also remedy the situation by bumping the affected gems to fixed versions if one is available?

dushan-madetech (approved):
Nice change 👍

Gemfile
baarkerlounger: No it's the same as the rubocop stage, it's an alert that manual intervention is required. You could probably chain on dependabot or something but I think depending on the vulnerability a human check is probably a good thing.

PR #308, opened 2022-02-16
Cldc 629 household situation renewal

kosiakkatrina:
Implement routing for household situation section
Fix inferring local authority (previously was cleared by `reset_invalidated_dependent_fields`)
Infer relevant fields for the subsection and update the content

### DIFF
--- a/app/helpers/tasklist_helper.rb
+++ b/app/helpers/tasklist_helper.rb
@@ -29,7 +29,8 @@ module TasklistHelper
     path = if subsection.is_started?(case_log)
              "case_log_#{subsection.id}_check_answers_path"
            else
-             "case_log_#{subsection.pages.first.id}_path"
+             next_page = subsection.pages.first.routed_to?(case_log) ? subsection.pages.first.id : case_log.form.next_page(subsection.pages.first, case_log)
+             "case_log_#{next_page}_path"
            end
     send(path, case_log)
   end
--- a/app/helpers/tasklist_helper.rb
+++ b/app/helpers/tasklist_helper.rb
@@ -25,7 +25,7 @@ module TasklistHelper
     case_log.form.subsections.count { |subsection| subsection.status(case_log) == status }
   end
 
-  def first_page_or_check_answers(subsection, case_log)
+  def next_page_or_check_answers(subsection, case_log)
     path = if subsection.is_started?(case_log)
              "case_log_#{subsection.id}_check_answers_path"
            else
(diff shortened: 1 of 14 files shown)


db/migrate/20220210132254_remove_checkbox_parent_fields.rb
```diff
@@ -11,7 +11,7 @@ def down
     change_table :case_logs, bulk: true do |t|
       t.column :accessibility_requirements, :string
       t.column :condition_effects, :string
-      t.remove :reasonable_preference_reason, :string
+      t.column :reasonable_preference_reason, :string
```
baarkerlounger, line 14: Why is this coming back?

app/helpers/tasklist_helper.rb
```diff
@@ -29,7 +29,8 @@ def first_page_or_check_answers(subsection, case_log)
     path = if subsection.is_started?(case_log)
              "case_log_#{subsection.id}_check_answers_path"
            else
-             "case_log_#{subsection.pages.first.id}_path"
+             next_page = subsection.pages.first.routed_to?(case_log) ? subsection.pages.first.id : case_log.form.next_page(subsection.pages.first, case_log)
```
baarkerlounger, line 32: This looks like it probably wants a spec?

db/migrate/20220210132254_remove_checkbox_parent_fields.rb
kosiakkatrina: This is for `down` method. I ran a rollback and it failed because of this methinks (?)

app/helpers/tasklist_helper.rb
kosiakkatrina: Hm, I thought it would already have one, lemme check

db/migrate/20220210132254_remove_checkbox_parent_fields.rb
baarkerlounger: Sorry stupid question

app/helpers/tasklist_helper.rb
baarkerlounger: It looks like you've added functionality for a new use case (the first page in the section isn't routed to), so should probably add a test for that specifically?

kosiakkatrina: You're right

kosiakkatrina: Added one now

baarkerlounger (approved):

PR #312, opened 2022-02-17
Cldc 465 tenancy type validation

kosiakkatrina:


app/models/validations/household_validations.rb
```diff
@@ -73,6 +73,15 @@ def validate_previous_housing_situation(record)
     end
   end
 
+  def validate_referral(record)
+    if record.rsnvac == "Relet to tenant who occupied same property as temporary accommodation" && REFERRAL_INVALID_TMP.include?(record.referral)
+      record.errors.add :referral, I18n.t("validations.household.referral.rsnvac_non_temp")
+    end
+    if record.tenancy == "Secure (including flexible)" && record.referral.present? && record.referral != "Internal transfer"
+      record.errors.add :referral, I18n.t("validations.household.referral.secure_tenancy")
+    end
```
dushan-madetech: I wonder if you can set this up to use something like the case log constants instead of the text - it might make the code less brittle if you update the content of the answers

app/models/validations/tenancy_validations.rb
```diff
@@ -25,4 +25,10 @@ def validate_fixed_term_tenancy(record)
   def validate_other_tenancy_type(record)
     validate_other_field(record, :tenancy, :tenancyother)
   end
+
+  def validate_tenancy(record)
+    if record.tenancy.present? && record.tenancy != "Secure (including flexible)" && record.referral == "Internal transfer"
```
dushan-madetech, line 30: Same point here

app/models/validations/household_validations.rb
```diff
@@ -73,6 +73,15 @@ def validate_previous_housing_situation(record)
     end
   end
 
+  def validate_referral(record)
+    if record.rsnvac == "Relet to tenant who occupied same property as temporary accommodation" && REFERRAL_INVALID_TMP.include?(record.referral)
+      record.errors.add :referral, I18n.t("validations.household.referral.rsnvac_non_temp")
+    end
+    if record.tenancy == "Secure (including flexible)" && record.referral.present? && record.referral != "Internal transfer"
+      record.errors.add :referral, I18n.t("validations.household.referral.secure_tenancy")
```
baarkerlounger, line 78: Should this add the error to `rsnvac` as well? i.e. could you answer them in either order?

spec/models/validations/tenancy_validations_spec.rb
```diff
@@ -106,6 +106,25 @@
...
+            record.tenancy = "Assured"
+            record.referral = "Internal transfer"
+            tenancy_validator.validate_tenancy(record)
+            expect(record.errors["tenancy"])
+              .to include(match I18n.t("validations.tenancy.internal_referral"))
+          end
+        end
+
+        context "when referral is internal transfer" do
+          it "does not add an error" do
```
dushan-madetech: The test description here is a little unclear - I had to look at your code under test to realise that the error should only apply if the referral is an internal transfer but the tenancy type is not given as secure (including flexible)

dushan-madetech (approved):
Left you some comments but nothing worth blocking the PR over. 🚢 it when you're happy with it

app/models/validations/household_validations.rb
baarkerlounger: I think we'll need to do that on all validations and tests at some point very soon (either use constants or the DB value directly). Wouldn't worry about that in this PR specifically though.

app/models/validations/household_validations.rb
kosiakkatrina: Hmm, so this is a validation that I just moved. Seems like there's a validation for it on line 70 of the same file: `validate_previous_housing_situation`, do you remember if they will run on both pages if they're in one method?

app/models/validations/household_validations.rb
kosiakkatrina: Seems like the general pattern in our codebase right now. Might be worth fixing, is now the right time to do it? Maybe, but I'd like to discuss if it's a good use of time!

app/models/validations/household_validations.rb
baarkerlounger: Yeah all validations run on all pages (on all saves) so shouldn't need to duplicate any.

app/models/validations/household_validations.rb
dushan-madetech: @kosiakkatrina Based on Dan's reply we can save this task for a future PR (or PRs) 🙂

spec/models/validations/tenancy_validations_spec.rb
kosiakkatrina: `context "when type of tenancy is secure"` > `context "when referral is not internal transfer"` >  `it "adds an error"`
`context "when type of tenancy is secure"` > `context "when referral is internal transfer"` >  `it "does not add an error"`

Is that because the top most context is too high up?

app/models/validations/household_validations.rb
kosiakkatrina: I was wrong about the `validate_previous_housing_situation` validation, it's actually checking a different field 🤦 
I located the correct one in the property validation tho and moved the referral validation there

baarkerlounger: So I think you can technically answer `referral` first and then `rsnvac` right? In which case we'd need to show the error message on the `rsnvac` page?

kosiakkatrina: Yup! That's happening

config/locales/en.yml
```diff
@@ -110,16 +110,16 @@ en:
         one_or_two_choices: "Only one box must be ticked or 'other disabilities' plus one of mobility disabilities"
       prevten:
         non_temp_accommodation: "Answer cannot be non-temporary accommodation as you already told us this is a re-let to tenant who occupied the same property as temporary accommodation"
+      referral:
+        secure_tenancy: "Answer must be internal transfer as you already told us this is a secure tenancy"
+        rsnvac_non_temp: "Answer cannot be this source of referral as you already told us this is a re-let to tenant who occupied the same property as temporary accommodation"
```
baarkerlounger, line 115: @@dan13L3 do these error messages work given that they can answer these in either order?

Dan13L3: @baarkerlounger Would you or one of the other devs be able to run these screens with me on a call so I can make sure I've understood the full context?

kosiakkatrina: Yup, happy to do that. Will set up some time with you

Dan13L3: @baarkerlounger Reviewed this with @kosiakkatrina and should be good to go now

baarkerlounger (approved):

PR #313, opened 2022-02-17
Fix sending reset email for admins

baarkerlounger:
- [x] Enable Admin Users to reset forgotten passwords
- [x] Prevent admin email addresses from being enumerable
- [x] Ensure password reset doesn't let you bypass 2FA
- [x] Use Gov style template for password reset view


app/views/devise/sessions/new.html.erb
```diff
@@ -1,4 +1,10 @@
-<% content_for :title, "Sign in to your account to submit CORE data" %>
+<% if resource_name == :admin_user %>
+  <% title = "CORE Admin Sign in" %>
```
paulrobertlloyd:
```suggestion
  <% title = "Sign in to your CORE administration account" %>
```

PR #315, opened 2022-02-17
Correct order of options in ‘Property information’ sub-section

paulrobertlloyd:


config/forms/2021_2022.json
```diff
@@ -654,15 +654,18 @@
...
                     "1": {
                       "value": "Internal transfer",
                       "hint": "Excluding renewals of a fixed-term tenancy"
                     },
+                    "10": {
+                      "value": "Previous tenant died with no succession"
+                    },
                     "2": {
-                      "value": "Relet to tenant who occupied same property as temporary accommodation"
+                      "value": "Re-let to tenant who occupied same property as temporary accommodation"
```
baarkerlounger, line 665: This will need updating here too: https://github.com/communitiesuk/mhclg-data-collection-beta/blob/main/app/models/constants/case_log.rb#L176

config/forms/2021_2022.json
```diff
@@ -654,15 +654,18 @@
...
                   "answer_options": {
-                    "0": {
-                      "value": "Renewal of fixed-term tenancy"
-                    },
                     "1": {
                       "value": "Internal transfer",
                       "hint": "Excluding renewals of a fixed-term tenancy"
                     },
+                    "10": {
+                      "value": "Previous tenant died with no succession"
```
baarkerlounger, line 662: This will need adding to the same hash as below

PR #316, opened 2022-02-17
Correct logic for void/renewal/handover dates

paulrobertlloyd:
* Correct Yes/No order for renewal question
* Fixes void/renewal/new-build question not being asked
* Shows the new-build handover date question if reason for vacancy is ‘First let of new-build property’
* Shows the void/renewal date and major repairs questions if reason for vacancy is NOT ‘First let of new-build property’

Screenshot 2022-02-17 at 14 56 09

### DIFF
--- a/config/forms/2021_2022.json
+++ b/config/forms/2021_2022.json
@@ -73,11 +73,11 @@
                   "hint_text": "A renewal is a letting to the same tenant in the same property.",
                   "type": "radio",
                   "answer_options": {
-                    "0": {
-                      "value": "No"
-                    },
                     "1": {
                       "value": "Yes"
+                    },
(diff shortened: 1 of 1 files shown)


config/forms/2021_2022.json
```diff
@@ -902,10 +905,7 @@
               "depends_on": [
                 {
                   "renewal": "No",
-                  "rsnvac": [
-                    "First let of conversion, rehabilitation or acquired property?",
```
paulrobertlloyd, line 906: The `?` in this option was causing a bug where none of the following questions were being shown 🙈

dushan-madetech (approved):
Nicely fixed 😄

PR #320, opened 2022-02-18
Feature/cldc 959 multiple tenant questions

kiddhustle:
TODO fix person_x_info string interpolation

https://github.com/communitiesuk/mhclg-data-collection-beta/blob/feature/CLDC-959-multiple-tenatnt-questions/config/forms/2021_2022.json#L1569


app/models/constants/case_log.rb
```diff
@@ -77,23 +85,23 @@ module Constants::CaseLog
 
   NATIONAL = {
     "UK national resident in UK" => 1,
-    "A current or former reserve in the UK Armed Forces (excluding National Service)" => 100,
+    # "A current or former reserve in the UK Armed Forces (excluding National Service)" => 100,
```
kosiakkatrina: Is this meant to be commented out?

db/migrate/20220211155138_add_age_known_to_case_logs.rb
```diff
@@ -0,0 +1,14 @@
...
+  def change
+    change_table :case_logs, bulk: true do |t|
+      t.column :age1_known, :string
+      t.column :age2_known, :string
+      t.column :age3_known, :string
+      t.column :age4_known, :string
+      t.column :age5_known, :string
+      t.column :age6_known, :string
+      t.column :age7_known, :string
+      t.column :age8_known, :string
```
kosiakkatrina: All of our other `_known` columns (like `la_known`, `postcode_known`) are integers, I wonder if we would want to keep it the same across these ones as well? 🤔

spec/models/case_log_spec.rb
```diff
@@ -459,6 +459,7 @@
 
       it "correctly derives and saves totelder" do
         record_from_db = ActiveRecord::Base.connection.execute("select totelder from case_logs where id=#{household_case_log.id}").to_a[0]
+        puts record_from_db
```
kosiakkatrina: Is this meant to be here?

config/forms/2021_2022.json
```diff
@@ -1172,85 +1187,252 @@
                 }
               }
             },
-            "tenant_ethnic_group": {
+            "person_1_ethnic_group": {
               "header": "",
               "description": "",
               "questions": {
-                "ethnic": {
+                "ethnic_group": {
```
baarkerlounger, line 1194: Do we need to change the question key? These are generally chosen specifically to match ADDs data dictionary (since the question name matches the DB field name)

config/forms/2021_2022.json
```diff
@@ -1134,10 +1134,25 @@
               "header": "",
               "description": "",
               "questions": {
+                "age1_known": {
+                  "header": "Do you know the lead tenant’s age?",
```
kosiakkatrina, line 1138: I think the first question should be `How many people live in the household for this letting?` ?

config/forms/2021_2022.json
```diff
@@ -1134,10 +1134,25 @@
...
+                  },
+                  "conditional_for": {
+                    "age1": ["Yes"]
+                  }
+                },
                 "age1": {
-                  "check_answer_label": "Tenant’s age",
-                  "header": "What is the tenant’s age?",
-                  "hint_text": "",
+                  "header": "Do you know the lead tenant’s age?",
```
kosiakkatrina, line 1154: Should this be just `Age` or something like that?

config/forms/2021_2022.json
```diff
@@ -1975,20 +2627,38 @@
...
+            },
+            "person_8_sex": {
+              "header": "",
+              "description": "",
+              "questions": {
                 "sex8": {
                   "check_answer_label": "Person 8’s gender",
                   "header": "Which of these best describes person 8’s gender identity?",
-                  "hint_text": "",
+                  "hint_text": "The lead tenant is the person in the household who does the most paid work. If several people do the same paid work, the lead tenant is whoever is the oldest.",
```
baarkerlounger: Is this hint text relevant to Person 8's gender?

config/forms/2021_2022.json
```diff
@@ -2004,9 +2674,18 @@
...
+                "details_known_8": "Yes"
+              }]
+            },
+            "person_8_economic_status": {
+              "header": "",
+              "description": "",
+              "questions": {
                 "ecstat8": {
-                  "check_answer_label": "Person 8’s Work",
+                  "check_answer_label": "Person 8's Work",
```
baarkerlounger: This probably needs reverting? I think we want `’` consistently in all output

config/forms/2021_2022.json
```diff
@@ -1378,39 +1561,56 @@
...
-                    "relat8": ">6",
-                    "age8": ">6",
-                    "sex8": ">6",
-                    "ecstat8": ">6"
+                  "step": 1
+                }
+              }
+            },
+            "person_2_info": {
+              "header": "You’ve given us the details for 1 of #{@case_log.other_hhmemb + 1} people in the household",
```
kosiakkatrina: This doesn't seem to work for me

spec/fixtures/exports/case_logs.xml
```diff
@@ -158,6 +158,23 @@
     <previous_postcode_known>1</previous_postcode_known>
     <previous_la_known/>
     <is_previous_la_inferred>false</is_previous_la_inferred>
+    <age1_known/>
```
baarkerlounger: I don't think we want to be exporting any of these since they're just specific to the form UI not the other entry methods.

config/forms/2021_2022.json
baarkerlounger: Oh I see, sorry.

The saved answer we want to export as `ethnic` now comes as a follow on question. :heavy_check_mark:

spec/fixtures/exports/case_logs.xml
kiddhustle: OK I think I saw some errors from the the tests/linting but I think they might have been around diff I can remove these

spec/models/case_log_spec.rb
kiddhustle: nope!

spec/fixtures/exports/case_logs.xml
kiddhustle: @baarkerlounger Is it all of the new fields that need omitting of just the **_known** fields

baarkerlounger: I think just the `_known` fields.

spec/fixtures/exports/case_logs.xml
```diff
@@ -158,6 +158,23 @@
...
+    <age2_known/>
+    <age3_known/>
+    <age4_known/>
+    <age5_known/>
+    <age6_known/>
+    <age7_known/>
+    <age8_known/>
+    <ethnic_group/>
```
baarkerlounger: Should this be `ethnic` rather than `ethnic_group`? That's the field that stores the final information ADD want right?

app/models/constants/case_log.rb
kiddhustle: we don't have it as an option any more. now removed

baarkerlounger: Just to confirm is that change intended to be for this years form or next years form? (@paulrobertlloyd ) because changing that this year surely changes data continuity?

db/migrate/20220211155138_add_age_known_to_case_logs.rb
kiddhustle: amended

app/models/constants/case_log.rb
paulrobertlloyd: As far as I can tell, this option shouldn’t be in this list, for either collection year.

spec/fixtures/exports/case_logs.xml
kiddhustle: Do I need to amend  the export service to omit these fields then?

**app/services/exports/case_log_export_service.rb**

app/models/constants/case_log.rb
paulrobertlloyd: These are the nationality options from the 2021/22 form:

Screenshot 2022-02-21 at 11 01 19

spec/models/case_log_spec.rb
kiddhustle: removed

config/forms/2021_2022.json
kiddhustle: reverted

config/forms/2021_2022.json
kiddhustle: aah no. it should just be on the lead tenant questions not the additional tenants

app/models/constants/case_log.rb
baarkerlounger: :man_facepalming:  Totally missed that this was the nationality question :+1:

spec/fixtures/exports/case_logs.xml
baarkerlounger: Yes I think so

config/forms/2021_2022.json
kiddhustle: amended

config/forms/2021_2022.json
kiddhustle: You're right it should

db/migrate/20220211155138_add_age_known_to_case_logs.rb
kiddhustle: @kosiakkatrina @baarkerlounger I changed these to integers but now the depends_on conditional page doesn't work. I set the depends_on value to integers to match (0 = No, 1 = Yes)

Do we have any examples of it working for integers values?

config/forms/2021_2022.json
kiddhustle: removed

config/forms/2021_2022.json
kiddhustle: For the lead tenant the text has changed:

https://miro.com/app/board/o9J_lqoxxRo=/

spec/fixtures/exports/case_logs.xml
kiddhustle: this have been split across two fields now but it could be omitted if necessary

kiddhustle: now omitted

baarkerlounger: Right. Those two field will need to be recombined into the original field then (and that exported) so the data sent to CDS remains the same.

spec/fixtures/exports/case_logs.xml
kiddhustle: amended

config/forms/2021_2022.json
kiddhustle: content  changed

spec/fixtures/exports/case_logs.xml
kiddhustle: amended

app/models/constants/case_log.rb
```diff
@@ -0,0 +1,1134 @@
+module Constants::CaseLog
```
dushan-madetech: I think this file was removed in main so it shouldn't be in this PR? 🤔

Gemfile.lock
```diff
@@ -23,10 +23,10 @@ GIT
 
 GIT
   remote: https://github.com/tagliala/activeadmin.git
-  revision: 000dd9425860fbadc72d4e08b893be098f5dff9b
+  revision: 871c6af8dec41f764a9f08bf04b5422971a23e32
```
baarkerlounger: Think this disappears if you rebase?

app/services/exports/case_log_export_service.rb
```diff
@@ -45,13 +45,27 @@ def retrieve_case_logs
...
-          value = case_log.read_attribute_before_type_cast(key)
-          form << doc.create_element(key, value)
+          # omit _known fileds from export
+          if key.starts_with?("details_known_") || pattern_age.match(key) || omitted_attrs.include?(key)
+            next
+          else
+            # combine ethnic fields
+            # rubocop:disable Style/ConditionalAssignment
+            if key == "ethnic"
+              value = case_log.read_attribute_before_type_cast("ethnic") || case_log.read_attribute_before_type_cast("ethnic_group")
```
baarkerlounger: I think we should be combining these in the model (like we do for other derived variables), not just in the export?

app/models/constants/case_log.rb
kiddhustle: removed

app/services/exports/case_log_export_service.rb
kiddhustle: sorry only just saw this. So via a `get_` method?

app/models/constants/case_log.rb
baarkerlounger: Looks like it's still here? (This whole file)

app/services/exports/case_log_export_service.rb
baarkerlounger: Something like:

`self.ethnic = ethnic || ethnic_group`

In the before validation. (Assuming `ethnic` is the fieldname that should be exported).

kiddhustle: oh OK so in this method:

`def set_derived_fields!`

baarkerlounger: Yep

app/services/exports/case_log_export_service.rb
```diff
@@ -45,13 +45,20 @@ def retrieve_case_logs
...
+      pattern_age = /age\d_known/
+      omitted_attrs = %w[ethnic_group]
 
       case_logs.each do |case_log|
         form = doc.create_element("form")
         doc.at("forms") << form
         case_log.attributes.each do |key, _|
-          value = case_log.read_attribute_before_type_cast(key)
-          form << doc.create_element(key, value)
+          # omit _known fileds from export
```
baarkerlounger: Spelling on fields. Could also possibly wrap that in a method to avoid the need for a comment? Something like:

```
if omitted_field?(key)
```

baarkerlounger (approved):

kiddhustle:
> Convention in Ruby is generally to have method name end in `?` if it returns a boolean

I get this lint error when I do this.

Is this down to my IDE needing configuration?

```
app/services/exports/case_log_export_service.rb:39:37: W: Lint/Syntax: invalid character syntax; use ?\n
(Using Ruby 3.1 parser; configure using TargetRubyVersion parameter, under AllCops)
    def is_omitted_field(field_name)? ...
                                    ^
app/services/exports/case_log_export_service.rb:39:37: E: Lint/Syntax: unexpected token tEH
(Using Ruby 3.1 parser; configure using TargetRubyVersion parameter, under AllCops)
    def is_omitted_field(field_name)?
```

@baarkerlounger I realised the `?` should be before the braces

baarkerlounger (approved):

PR #321, opened 2022-02-21
Refactor enums

baarkerlounger:
### **What problem is this PR trying to solve?**

- Currently every change to an answer option label requires changing code in **at least two** places, the form JSON and the model constant. If validations rely on the answer option, it can require changing in many places. This is brittle, easy to miss and easy to introduce bugs that result in data not being saved properly

- Currently this makes it impossible to have multiple forms with different labels for the same answer option since the enum is statically defined on the class

- Requiring string data to be sent also makes for a poor API as it's easy to introduce errors through subtle differences, for example character encoding or `'` type.

### **How?**

- We scrap enums and use the stored values (mostly integers) directly in validations

- Where we need to map them (for example on Form save, or in Check Answers Display) we look up the required values from the JSON such that we have a single source of truth per form

### **What new problems does this give us?**

- Because we're trying to match the existing data dictionary our values are poorly mapped. Answers like "No", "Yes", "Prefer not to say", "Don't know" and "Other", would ideally all have a consistent key across different questions such that's easy to reason about them in validations. To mitigate that we could introduce convenience methods for these common ones like `question.answer(case_log).is_other?`, `question.answer(case_log).is_prefer_not_to_say?`. That will probably be left for a subsequent PR though.

- It makes it harder to decouple testing functionality from the form definition since for some cases (for example local authority), the functionality is closely coupled to the value

### **TODO**

- [x] Make all validation and derived variable comparisons check values not labels
- [x] Update JSON to make all depends on and conditional for comparisons based on values rather than labels
- [x] For radio buttons ensure key/value is passed as value (and saved) rather than value label
- [x] For select questions ensure key/value is passed as value (and saved) rather than value label
- [x] Add answer value lookup from JSON so that form answers get displayed correctly in check answers (labels)
- [x] Make sure LA returned by API request gets converted to value key
- [x] Update API documentation - since the API docs will now just have integers will need to consider how we document their meaning for users (https://digital.dclg.gov.uk/jira/browse/CLDC-1025)


db/migrate/20220221181805_integer_fields.rb
```diff
@@ -0,0 +1,83 @@
+class IntegerFields < ActiveRecord::Migration[7.0]
+  def up
+    change_table :case_logs, bulk: true do |t|
+      t.remove :sex1
+      t.column :sex1, :integer
```
kosiakkatrina: Not an integer? Isn't it like "F, M, X ..."

baarkerlounger: We have integers in JSON but you're right it should be a char!

baarkerlounger: Done

app/models/form/question.rb
```diff
@@ -73,19 +75,45 @@ def has_inferred_check_answers_value?(case_log)
 
   def update_answer_link_name(case_log)
     link_type = if type == "checkbox"
-                  answer_options.keys.any? { |key| case_log[key] == "Yes" } ? "Change" : "Answer"
+                  answer_options.keys.any? { |key| case_log[key] == 1 } ? "Change" : "Answer"
```
smeny: Is this always true? I thought some answers could be Yes with a different value than 1.

baarkerlounger: Not for checkbox type questions (because we set those from the main question. Those are all 1.

smeny (approved):

PR #322, opened 2022-02-22
Change how we're reseting answers

kosiakkatrina:
Reset values for dependant derived questions
Only reset question values if the selected answer is not present in the answer options

### DIFF
--- a/app/models/case_log.rb
+++ b/app/models/case_log.rb
@@ -230,7 +230,7 @@ private
       condition_key = conditions.first[:key]
       condition_value = conditions.first[:value]
       if public_send("#{condition_key}_changed?") && condition_value == public_send(condition_key) && !public_send("#{dependent}_changed?")
-        self.layear = nil
+        self[dependent] = nil
       end
     end
   end
--- a/app/models/case_log.rb
+++ b/app/models/case_log.rb
@@ -219,6 +219,16 @@ private
     form.invalidated_page_questions(self).each do |question|
       public_send("#{question.id}=", nil) if respond_to?(question.id.to_s)
     end
+
+    dependent_questions = { layear: [{ key: :renewal, value: "No" }] }
+
+    dependent_questions.each do |dependent, conditions|
+      condition_key = conditions.first[:key]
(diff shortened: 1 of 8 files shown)


app/models/case_log.rb
```diff
@@ -213,12 +213,33 @@ def update_status!
...
+      enabled = form.enabled_page_questions(self)
+      contains_selected_answer_option = enabled.map(&:id).include?(question.id) && enabled.find { |q| q.id == question.id }.answer_options.values.map { |x| x["value"] }.include?(public_send(question.id))
+      unless contains_selected_answer_option
+        public_send("#{question.id}=", nil) if respond_to?(question.id.to_s)
+      end
+    end
+  end
+
+  def reset_derived_questions
+    dependent_questions = { layear: [{ key: :renewal, value: "No" }], homeless: [{ key: :renewal, value: "No" }] }
```
baarkerlounger, line 227: Can we derive this from the JSON rather than hardcoding it?

kosiakkatrina: We don't really have anything that indicates that these are inferred in the JSON. Would it make sense to add it just for this reason?

baarkerlounger: hmm we have:

```
"layear": {
    "depends_on":  [{ "renewal": "No" }]
},
"homeless": {
    "depends_on":  [{ "renewal": "No" }]
}
```

If we collect those and apply this does it break for other fields?

kosiakkatrina: So we're already getting the depends on to determine if something is not routed to and if it's not we clear it.
In this case we want to clear something that is routed to, but only if it's one of the fields we're inferring. We can get the questions that are routed to from this JSON, but we would still need to filter them down to the ones that get inferred, no?

kosiakkatrina: Otherwise we'd just clear everything?

baarkerlounger: Ah yeah I remember now.  :+1:

baarkerlounger:
Think you'll need to bump nokogiri on main before you can merge?

PR #323, opened 2022-02-22
Cldc 810 outside of the uk addresses

kosiakkatrina:
Put LA as conditional questions vs conditional pages
Outside of uk addresses worked already
image
image

### DIFF
--- a/spec/models/case_log_spec.rb
+++ b/spec/models/case_log_spec.rb
@@ -396,15 +396,15 @@ RSpec.describe CaseLog do
       end
 
       it "correctly resets la if la is not known" do
-        address_case_log.update!({ previous_la_known: "Yes", previous_la: "Scotland" })
-        record_from_db = ActiveRecord::Base.connection.execute("select previous_la from case_logs where id=#{address_case_log.id}").to_a[0]
-        expect(record_from_db["previous_la"]).to eq(nil)
-        expect(address_case_log.previous_la).to eq("Scotland")
+        address_case_log.update!({ previous_la_known: "Yes", prevloc: "Scotland" })
+        record_from_db = ActiveRecord::Base.connection.execute("select prevloc from case_logs where id=#{address_case_log.id}").to_a[0]
+        expect(record_from_db["prevloc"]).to eq("S92000003")
+        expect(address_case_log.prevloc).to eq("Scotland")
 
         address_case_log.update!({ previous_la_known: "No" })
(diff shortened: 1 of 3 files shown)


spec/models/case_log_spec.rb
```diff
@@ -395,9 +395,22 @@
...
+        address_case_log.update!({ previous_la_known: "No" })
+        record_from_db = ActiveRecord::Base.connection.execute("select prevloc from case_logs where id=#{address_case_log.id}").to_a[0]
+        expect(address_case_log.prevloc).to eq(nil)
+        expect(record_from_db["prevloc"]).to eq(nil)
+      end
+
       it "changes the prevloc if previous postcode changes from not known to known and provided" do
         address_case_log.update!({ previous_postcode_known: "No" })
-        address_case_log.update!({ prevloc: "Westminster" })
+        address_case_log.update!({ previous_la_known: "Yes", prevloc: "Westminster" })
```
kosiakkatrina, line 413: doesn't reset it if we update it like this, but not sure if that would ever be the case:
`address_case_log.update!({ previous_postcode_known: "No",  previous_la_known: "Yes", prevloc: "Westminster" })`

baarkerlounger: Could be attempted via the API but I think in that case a validation error would be more appropriate than a reset?

kosiakkatrina: Can we update using API?

baarkerlounger: Yes (Patch/Put request)

kosiakkatrina: Okay, that might need some fixing then. So for location validations so far I think we've got:
If postcode known is yes then a valid postcode needs to be provided.

I guess it would make sense to have:
If la known is yes then a valid la needs to be provided.

What would a validation for this case be tho? Cause it's valid to have `previous_postcode_known: "No", previous_la_known: "Yes", prevloc: "Scotland"`

but if it was previously inferred it'd clear the values of LA

baarkerlounger: I think clearing the previously inferred values of LA actually is expected if the dependency changes (but looks like we got this wrong initially):

https://communities-govuk.slack.com/archives/C01QGR78H9S/p1645532079426639

kosiakkatrina: Yeah, we can clear previously inferred value, but not the one that is provided in the current request

baarkerlounger: Ah right yes, actually nothing to validate or reset there I think

baarkerlounger (approved):

PR #328, opened 2022-02-23
Reset more values

kosiakkatrina:
When we select non renewal and select reasonable preference as "Yes"
Then when we try to select renewal
It tries to set homeless to No which doesn't pass the validation as reasonable preference cannot be yes anymore.

This PR resets resonablepref if it's yes and it's a renewal and also resets referral

### DIFF
--- a/app/models/case_log.rb
+++ b/app/models/case_log.rb
@@ -227,7 +227,7 @@ private
   def reset_derived_questions
     dependent_questions = { layear: [{ key: :renewal, value: "No" }],
                             homeless: [{ key: :renewal, value: "No" }],
-                            referral: [{ key: :renewal, value: "No" }]}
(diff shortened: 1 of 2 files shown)


app/models/case_log.rb
```diff
@@ -283,6 +285,7 @@ def set_derived_fields!
       self.homeless = "No"
       self.referral = "Internal transfer"
       self.layear = "Less than 1 year"
+      self.reasonpref = nil if reasonpref == "Yes"
```
baarkerlounger, line 288: Worth adding a test or already covered?

kosiakkatrina: Good point. Thank you!

baarkerlounger (approved):

PR #334, opened 2022-02-24
Set providertype before save

kosiakkatrina:


baarkerlounger:
Is this still needed now that Provider Type is an enum? If it is can we have a test for it?

kosiakkatrina:
> Is this still needed now that Provider Type is an enum? If it is can we have a test for it?

@baarkerlounger doesn't work without it works with 🤷‍♀️

smeny (approved):

PR #338, opened 2022-02-28
Cldc 788 rent outside accepted range

dushan-madetech:


spec/features/form/validations_spec.rb
```diff
@@ -124,18 +124,19 @@ def fill_in_date(case_log_id, question, day, month, year, path)
       let(:income_over_soft_limit) { 750 }
       let(:income_under_soft_limit) { 700 }
 
-      it "prompts the user to confirm the value is correct", js: true do
+      it "prompts the user to confirm the value is correct on an interruption screen", js: true do
```
baarkerlounger: Does this still require js?

app/controllers/form_controller.rb
```diff
@@ -7,7 +7,7 @@ def submit_form
     if @case_log
       @page = @case_log.form.get_page(params[:case_log][:page])
       responses_for_page = responses_for_page(@page)
-      if @case_log.update(responses_for_page) && @case_log.has_no_unresolved_soft_errors?
```
baarkerlounger, line 10: :heart:

app/helpers/interuption_screen_helper.rb
```diff
@@ -0,0 +1,16 @@
+module InteruptionScreenHelper
+  def display_informative_text(informative_text, case_log)
+    translation_questions = informative_text["argument"].map { |x| case_log.form.get_question(x) }
+    begin
+      case translation_questions.count
```
baarkerlounger, line 5: Can this never be more than 2?

app/helpers/interuption_screen_helper.rb
```diff
@@ -0,0 +1,16 @@
...
+      when 2
+        translation = I18n.t(informative_text["translation"], informative_text["argument"][0].to_sym => translation_questions[0].answer_label(case_log), informative_text["argument"][1].to_sym => translation_questions[1].answer_label(case_log))
+      when 1
+        translation = I18n.t(informative_text["translation"], informative_text["argument"][0].to_sym => translation_questions[0].answer_label(case_log))
+      end
+    rescue StandardError
```
baarkerlounger, line 11: What are we really catching here? Is it the `count` not being 1 or 2? Or is the lookup failing? Or something else?

app/models/form/page.rb
```diff
@@ -1,24 +1,21 @@
...
+                :depends_on, :title_text, :informative_text, :subsection, :hide_subsection_label
 
   def initialize(id, hsh, subsection)
     @id = id
     @header = hsh["header"]
     @description = hsh["description"]
     @questions = hsh["questions"].map { |q_id, q| Form::Question.new(q_id, q, self) }
     @depends_on = hsh["depends_on"]
+    @title_text = hsh["title_text"]
+    @informative_text = hsh["informative_text"]
```
baarkerlounger, line 12: Is this different from `description`?

app/models/form/page.rb
```diff
@@ -1,24 +1,21 @@
...
     @hide_subsection_label = hsh["hide_subsection_label"]
-    @soft_validations = hsh["soft_validations"]&.map { |sv_id, s| Form::Question.new(sv_id, s, self) }
     @subsection = subsection
   end
 
   def expected_responses
```
baarkerlounger: Is this method still needed at all? If it's just a wrapper for questions now can probably use `questions` directly wherever this is used?

app/helpers/interuption_screen_helper.rb
dushan-madetech: It potentially could be but realistically probably won't be - I suggested that we just go with the YAGNI principle here and don't extend it beyond 2 until we actually end up needing to. Happy to change this if you think we should

app/models/form/page.rb
dushan-madetech: the `@title_text` is for passing to the `title_text` param of the GovUK Panel component in  app/views/form/_interruption_screen_question.html.erb 

`@informative_text` was the the text under the panel `title_text` - the thinking behind it was to keep the params distinct and easily identifiable and also to have these as things only relevant to the `interruption_screen` question type

app/models/form/page.rb
dushan-madetech: Probably not - we're still in the process of removing all the unneeded bits and tidying up the resulting code. We'll probably end up removing the method 😄

spec/features/form/validations_spec.rb
dushan-madetech: It didn't - this has been fixed 😄

app/helpers/interuption_screen_helper.rb
baarkerlounger: Would the behaviour be different for 2 vs 3?

i.e. rather than a case statement just a `translation_questions.count > 1`?

PR #339, opened 2022-02-28
Update referral validation numbers and clearing of the answers

kosiakkatrina:
We were clearing out the answers we didn't want to clear upon routing changes because of the enum update

### DIFF
--- a/app/models/case_log.rb
+++ b/app/models/case_log.rb
@@ -156,7 +156,7 @@ class CaseLog < ApplicationRecord
   end
 
   def is_internal_transfer?
-    !!(referral && referral == 1)
+    referral == 1
   end
 
   def is_relet_to_temp_tenant?
--- a/app/models/case_log.rb
+++ b/app/models/case_log.rb
@@ -156,7 +156,7 @@ class CaseLog < ApplicationRecord
   end
 
   def is_internal_transfer?
-    !!(referral && referral.zero?)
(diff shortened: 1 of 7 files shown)


config/forms/2021_2022.json
```diff
@@ -3222,37 +3222,37 @@
                   "hint_text": "",
                   "type": "radio",
                   "answer_options": {
-                    "0": {
+                    "1": {
```
baarkerlounger, line 3225: How come this is changing? These should match the data dictionary right?

kosiakkatrina: Yes, I think that these were pre this PR 0,1,2 .... 10 and now they are matching the dictionary? Are we looking at different dictionary files?

app/models/case_log.rb
```diff
@@ -156,7 +156,7 @@ def is_assured_shorthold_tenancy?
   end
 
   def is_internal_transfer?
-    !!(referral && referral.zero?)
+    !!(referral && referral == 1)
```
baarkerlounger: This can just be `referral == 1`, the presence check is just there for 0 checks because `nil` doesn't respond to `.zero?`

kosiakkatrina: Nice! Will change that

baarkerlounger (approved):

PR #347, opened 2022-03-02
remove nullifying resonpref

kosiakkatrina:
Removing this reset of reasonpref. Not sure why it was there as we're displaying the question in both cases 🤔 

Removing it because it prevents reasonpref to be set to Yes if the renewal is Yes, which doesn't make sense

### DIFF
--- a/app/models/case_log.rb
+++ b/app/models/case_log.rb
@@ -302,7 +302,6 @@ private
       self.homeless = 2
       self.referral = 0
       self.layear = 1
-      self.reasonpref = nil if given_reasonable_preference?
     end
     if is_general_needs?
       self.prevten = 2 if managing_organisation.provider_type == "PRP"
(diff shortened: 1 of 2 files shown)


smeny (changes requested):
Especially if we took the time to write a test for this, it hints that this is a feature that was expected at some point. I would check with Jon or Paul before merging this in.

kosiakkatrina:
> Especially if we took the time to write a test for this, it hints that this is a feature that was expected at some point. I would check with Jon or Paul before merging this in.

From the slack convo:
It’s because of the wrong validation. Previously if homeless was selected as No we couldn’t have reasonable preference as Yes, and because we were inferring Homeless as No if renewal changed to Yes and reasonable preference was previously selected as Yes that would trigger the validation, We’ve since changed that validation, so Homeless as No and reasonable preference as Yes are now allowed to be selected at the same time

smeny (approved):
Approved as we retrieved the original reason for the feature

PR #349, opened 2022-03-03
Content and big fixes for household characteristics

paulrobertlloyd:
* Ensures options and answer labels use the correct content
* Only show one row for tenant/person’s age
* Bug fixes (spotted while editing JSON)
* Update tests

NOTE: This PR does not address the issue around deriving the correct number of people in the household, or adjust the answer label that relates to this.

### DIFF
--- a/spec/fixtures/forms/2021_2022.json
+++ b/spec/fixtures/forms/2021_2022.json
@@ -31,7 +31,7 @@
               "person_1_age": {
                 "questions": {
                   "age1": {
-                    "check_answer_label": "Tenant’s age",
+                    "check_answer_label": "Lead tenant’s age",
                     "header": "What is the tenant’s age?",
                     "type": "numeric",
                     "min": 0,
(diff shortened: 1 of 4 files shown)


spec/fixtures/forms/2021_2022.json
```diff
@@ -31,7 +31,7 @@
               "person_1_age": {
```
kosiakkatrina, line 31: We don't care too much about the wording in our test fixtures. The idea behind it being, if we change content for the actual form that doesn't impact the tests. It doesn't hurt having this as a change, but for the next time, it's not necessary

paulrobertlloyd: Yeah, I’m always unsure how important/useful it is to keep the fixtures and tests up to date with the actual form. OCD usually gets the better of me! 🤪

kosiakkatrina (approved):
Looks good. It seems like the mappings for ecstat(1-8) still need to be updated, but I assume it's not part of this PR

paulrobertlloyd:
> Looks good. It seems like the mappings for ecstat(1-8) still need to be updated, but I assume it's not part of this PR

Thanks @kosiakkatrina; not sure what needs to be updated changed with regards to this issue, so left values as is for now.

dushan-madetech (approved):
Looks good to me 👍. Happy for you to 🚢 it when you're ready for it to go into main

PR #354, opened 2022-03-03
Content tweaks for household situation section

paulrobertlloyd:
* Adding ‘or’ dividers
* Order of options
* Question URL slugs

### DIFF
--- a/config/forms/2021_2022.json
+++ b/config/forms/2021_2022.json
@@ -3344,32 +3344,32 @@
                 "reason": {
                   "check_answer_label": "Reason for leaving last settled home",
                   "header": "What is the tenant’s main reason for the household leaving their last settled home?",
-                  "hint_text": "The tenant’s ’last settled home’ is their last long-standing home. For tenants who were in temporary accommodation or sleeping rough, their last settled home is where they were living previously.",
+                  "hint_text": "The tenant’s ‘last settled home’ is their last long-standing home. For tenants who were in temporary accommodation or sleeping rough, their last settled home is where they were living previously.",
                   "type": "radio",
                   "answer_options": {
                     "40": {
-                      "value": "End of assured shorthold tenancy – no fault"
+                      "value": "End of assured shorthold tenancy (no fault)"
                     },
                     "41": {
-                      "value": "End of assured shorthold tenancy – eviction or tenant at fault"
+                      "value": "End of assured shorthold tenancy (eviction or tenant at fault)"
                     },
                     "42": {
-                      "value": "End of fixed term tenancy – no fault"
+                      "value": "End of fixed term tenancy (no fault)"
                     },
                     "43": {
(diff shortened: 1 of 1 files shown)


config/forms/2021_2022.json
```diff
@@ -4365,10 +4365,10 @@
                       "value": "Internal transfer"
                     },
                     "2": {
-                      "value": "Tenant applied directly (no referral or nomination)"
+                      "value": "Tenant applied directly (no referral)"
                     },
                     "3": {
-                      "value": "Private registered provider (PRP) supported lettings only – referred by local authority housing department"
+                      "value": "Referred by local authority housing department"
```
paulrobertlloyd, line 4371: @roobottom Do we need to include the text ‘Private registered provider (PRP) supported lettings only’ given the option is only shown in this case anyway?

config/forms/2021_2022.json
```diff
@@ -4425,10 +4425,10 @@
                       "value": "Tenant applied directly (no referral or nomination)"
                     },
                     "3": {
-                      "value": "Private registered provider (PRP) lettings only – nominated by a local housing authority"
+                      "value": "Nominated by a local housing authority"
                     },
                     "4": {
-                      "value": "Private registered provider (PRP) supported lettings only – referred by local authority housing department"
+                      "value": "Referred by local authority housing department"
```
paulrobertlloyd, line 4431: @roobottom Do we need to include the text ‘Private registered provider (PRP) [supported] lettings only’ given these options are only shown in this case anyway?

config/forms/2021_2022.json
```diff
@@ -4308,10 +4308,10 @@
                       "value": "Internal transfer"
                     },
                     "2": {
-                      "value": "Tenant applied directly (no referral or nomination)"
+                      "value": "Tenant applied directly (no nomination)"
                     },
                     "3": {
-                      "value": "Private registered provider (PRP) lettings only – nominated by a local housing authority"
+                      "value": "Nominated by a local housing authority"
```
paulrobertlloyd, line 4314: @roobottom Do we need to include the text ‘Private registered provider (PRP) lettings only’ given the option is only shown in this case anyway?

paulrobertlloyd (commented):
@roobottom I have some questions about the labels we use for source of referral.

config/forms/2021_2022.json
```diff
@@ -3611,7 +3611,7 @@
                 }
               ]
             },
-            "underoccupation_benefitcap": {
+            "under_occupation_benefit_cap": {
```
baarkerlounger, line 3614: Why is this changing? Looks like the original was correct in the General Needs Data Dictionary?

paulrobertlloyd: This is just changing the URL slug, the data key stays the same.

PR #355, opened 2022-03-03
CLDC-1009: Logs summary

paulrobertlloyd:
This PR implements most of the changes outlined in [CLDC-1009](https://digital.dclg.gov.uk/jira/browse/CLDC-1009) besides calculating and showing the completion status of each log (for now it just shows a tag with the log status)

* Implement wider application layout
* Replace instances/mentions of ‘your log’ with ‘[this] log’
* Add `app-figure` styles
* Add `app-table-group` styles
* Link to bulk upload from logs page
* Show logs of all statuses in one table on logs page
* Show log id, tenant code, property reference, created data and placeholder status columns
* Update tests

| Before | After |
| - | - |
| before | after |

### DIFF
--- a/app/views/case_logs/_log_list.html.erb
+++ b/app/views/case_logs/_log_list.html.erb
@@ -1,29 +1,46 @@
-<table class="govuk-table">
-  <caption class="govuk-table__caption govuk-table__caption--m"><%= title %></caption>
-  <thead class="govuk-table__head">
-    <tr class="govuk-table__row">
-      <th scope="col" class="govuk-table__header">Log reference</th>
-      <th scope="col" class="govuk-table__header">Postcode</th>
-      <th scope="col" class="govuk-table__header">Tenant code</th>
-      <th scope="col" class="govuk-table__header"><%= date_title %></th>
-    </tr>
-  </thead>
-  <tbody class="govuk-table__body">
-    <% case_logs.map do |log| %>
-    <tr class="govuk-table__row">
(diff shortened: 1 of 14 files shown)


smeny (approved):

app/views/case_logs/_log_list.html.erb
```diff
@@ -1,29 +1,46 @@
...
-      <td id="last-changed" class="govuk-table__cell">
-        <%= log.updated_at.to_formatted_s(:govuk_date) %>
-      </td>
-    </tr>
-    <% end %>
-  </tbody>
-</table>
+<figure class="app-figure">
+  <figcaption id="<%= title.dasherize %>" class="app-figure__caption">
+    <b><%= case_logs.count %></b> <%= title.downcase %>
```
baarkerlounger, line 3: You probably want `case_logs.size` here rather than `.count` since you've already loaded them. `.count` will end up doing two SQL queries:

```
CaseLog Load (2.0ms)  SELECT "case_logs".* FROM "case_logs" WHERE ("case_logs"."owning_organisation_id" = $1 OR "case_logs"."managing_organisation_id" = $2) 

CaseLog Count (1.7ms)  SELECT COUNT(*) FROM "case_logs" WHERE ("case_logs"."owning_organisation_id" = $1 OR "case_logs"."managing_organisation_id" = $2)
```

Whereas `.size` here would just do the first.

PR #356, opened 2022-03-07
CLDC-1015: limit password reset

kosiakkatrina:
When password reset email is being sent to the same email address (from `/users/password/new` or post to `/users/password`) more than 5 times per minute.
We redirect to `/429` and return 429 (too many requests) status code
image

### DIFF
--- a/config/initializers/rack_attack.rb
+++ b/config/initializers/rack_attack.rb
@@ -11,3 +11,10 @@ Rack::Attack.throttle("password reset requests", limit: 5, period: 60.seconds) d
     request.params["user"]["email"].to_s.downcase.gsub(/\s+/, "")
   end
 end
+
+Rack::Attack.throttled_responder = lambda do |_env|
+  headers = {
+    "Location" => "/429",
+  }
+  [301, headers, []]
+end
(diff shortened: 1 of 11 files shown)


dushan-madetech (approved):
Looks good 👍

config/initializers/rack_attack.rb
```diff
@@ -0,0 +1,20 @@
+if Rails.env.development? || Rails.env.test?
+  Rack::Attack.cache.store = ActiveSupport::Cache::MemoryStore.new
+  Rack::Attack.enabled = false
+else
+  redis_url = PaasConfigurationService.new.redis_uris[:"dluhc-core-#{Redis.env}-redis-rate-limit"]
```
kosiakkatrina, line 5: @smeny Do we need to set up any of the env vars etc before we deploy this? Also, is the Redis instance on staging now?

smeny: * No, because when the redis instance is bound to our CORE app, it adds the section in VCAP_SERVICES (which we read from). Note that we should update the deployment manifest to ensure this link is enforced.
* Yes, the redis instance is available on staging.

PR #358, opened 2022-03-07
Use less expensive query for counting logs

paulrobertlloyd:


app/views/case_logs/_log_list.html.erb
```diff
@@ -1,6 +1,6 @@
 <figure class="app-figure">
   <figcaption id="<%= title.dasherize %>" class="app-figure__caption">
-    <b><%= case_logs.count %></b> <%= title.downcase %>
+    <b><%= case_logs.size %></b> <%= title.downcase %>
```
smeny, line 3: Adding some context (if someone is wondering):
* count will always result in a select query in database
* size will return the array length if already loaded or make the count call

The only risk I see is that we could miss an update since the case_logs were loaded, but unlikely for this page?

smeny (approved):

baarkerlounger (approved):

PR #362, opened 2022-03-08
Generically validate all numeric question min and maxes

baarkerlounger:
This no validates all numeric questions (including ages) against the min and max values defined for them in the JSON config.


kosiakkatrina (approved):
Very nice!

PR #372, opened 2022-03-10
CLDC-788 interruption screen net income soft validation

dushan-madetech:
This PR does the following:

- Removes the old soft validation behaviour
- Adds a new interruption screen for soft validations
- Converts the existing net income soft validation to the new design

**Net income soft validation converted to the new interruption screen design**
Screenshot 2022-03-10 at 15 08 03


config/locales/en.yml
```diff
@@ -144,7 +143,7 @@ en:
 
   soft_validations:
     net_income:
-      hint_text: "This is based on the tenant’s work situation: %{ecstat1}"
+      hint_text: "This is based on the tenant’s work situation: <strong>%{ecstat1}</strong><br><br>The household income you have entered is <strong>%{earnings}</strong>"
```
paulrobertlloyd: Not for this PR, but we really ought to have a way of enabling paragraphs in hint text. (Not sure adding `<p>` in JSON would work… needs some thought).

dushan-madetech: This hint text does get wrapped in a `<p>` in line 14 of this file: `app/helpers/interuption_screen_helper.rb`

paulrobertlloyd: Ooooh, actually hint text is a `div`, not a `span` as I thought. Brilliant! In which case, you should be able to do this:

```yml
hint_text: "<p>This is based on the tenant’s work situation: <strong>%{ecstat1}</strong></p><p>The household income you have entered is <strong>%{earnings}</strong></p>"
```

(Also just noticed this is a YAML config, not the JSON form config 😅)

dushan-madetech: Yeah, the old soft validation code put the text in the yaml so we looked at a way to re-use that so that we could display the values the user provided for previous questions. We couldn't come up with an elegant way of doing that using the JSON. The upside is that I can easily add the HTML you'd like to this YAML file 😄

config/locales/en.yml
```diff
@@ -143,7 +143,7 @@ en:
 
   soft_validations:
     net_income:
-      hint_text: "You told us the main tenant's working situation is: <strong>%{ecstat1}</strong><br><br>The household income you have entered is <strong>%{earnings}</strong>"
+      hint_text: "<p class='govuk-body'>You told us the main tenant's working situation is: <strong>%{ecstat1}</strong></p><p class='govuk-body'>The household income you have entered is <strong>%{earnings}</strong></p>"
```
paulrobertlloyd: What happens if you don’t add the classes?

```suggestion
      hint_text: "<p>You told us the main tenant's working situation is: <strong>%{ecstat1}</strong></p><p>The household income you have entered is <strong>%{earnings}</strong></p>"
```

dushan-madetech: **With the class**
Screenshot 2022-03-10 at 16 53 36

**Without the class**
Screenshot 2022-03-10 at 16 54 10

paulrobertlloyd: Ah, I’ll just make a quick tweak the CSS and add it to this PR…

paulrobertlloyd: Tweaked!

paulrobertlloyd (approved):

app/models/form.rb
```diff
@@ -35,7 +35,9 @@ def subsection_for_page(page)
   def next_page(page, case_log)
     page_ids = subsection_for_page(page).pages.map(&:id)
     page_index = page_ids.index(page.id)
-    nxt_page = get_page(page_ids[page_index + 1])
+    page_id = page.id.include?("value_check") && case_log[page.id] == 1 ? page_ids[page_index - 1] : page_ids[page_index + 1]
```
baarkerlounger, line 38: Is there a test for this specific logic? i.e. one asserting what should be returned for each of the cases?

Also when is `case_log[page.id] == 1` true? What is 1 there?

app/helpers/interuption_screen_helper.rb
```diff
@@ -0,0 +1,16 @@
+module InteruptionScreenHelper
+  def display_informative_text(informative_text, case_log)
+    translation_questions = informative_text["argument"].map { |x| case_log.form.get_question(x) }
+    begin
+      case translation_questions.count
+      when 2
+        translation = I18n.t(informative_text["translation"], informative_text["argument"][0].to_sym => translation_questions[0].answer_label(case_log), informative_text["argument"][1].to_sym => translation_questions[1].answer_label(case_log))
```
baarkerlounger, line 7: Might be nicer to map over the questions here such that it works for any number > 1 but just a nitpick. No need to change.

baarkerlounger (approved):
Great stuff @dushan-madetech @kosiakkatrina !

app/models/form.rb
dushan-madetech: This is my bad I should have added a comment here! `case_log[page.id]` in this case becomes `case_log[net_income_value_check]` and the 1 refers to choosing no on the radio question

baarkerlounger: That makes sense! Might be worth extracting a method to make that a little more obvious.

dushan-madetech: But the idea is that any interruption screen page will have a page name containing `value_check` and the accompanying radio question will also have `value_check` in the name so `case_log[page.id]` will become `case_log[*field_name*_value_check]`

dushan-madetech: Hopefully I've explained that clearly! 😅

PR #373, opened 2022-03-10
Sentry fixes

baarkerlounger:
Fixes:

1. Exception thrown when email already exists by showing validation error instead: 
https://sentry.io/organizations/dluhc-core/issues/3069237247/?project=6196323&query=is%3Aunresolved

2. Exception thrown when saved value is invalid (can't be mapped) by displaying the saved value in check answers directly:
 https://sentry.io/organizations/dluhc-core/issues/3053838734/?project=6196323&query=is%3Aunresolved

3. Exception thrown when provider type is nil by enforcing presence:
https://sentry.io/organizations/dluhc-core/issues/3085913725/?project=6196323&query=is%3Aunresolved
https://sentry.io/organizations/dluhc-core/issues/3085912158/?project=6196323&query=is%3Aunresolved

### DIFF
--- a/app/admin/organisations.rb
+++ b/app/admin/organisations.rb
@@ -2,7 +2,7 @@ ActiveAdmin.register Organisation do
   permit_params do
     permitted = %i[name
                    phone
-                   providertype
+                   provider_type
                    address_line1
                    address_line2
                    postcode
@@ -27,9 +27,5 @@ ActiveAdmin.register Organisation do
     column :other_stock_owners
     column :managing_agents
     actions
-  end
-
-  before_save do |org|
-    org.provider_type = params[:organisation][:provider_type]
   end
 end
--- a/app/models/form/question.rb
+++ b/app/models/form/question.rb
@@ -113,7 +113,7 @@ class Form::Question
             else
               value.to_s
             end
-    label || value
+    label || value.to_s
   end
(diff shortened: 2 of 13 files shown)


app/admin/organisations.rb
```diff
@@ -2,7 +2,7 @@
   permit_params do
     permitted = %i[name
                    phone
-                   providertype
+                   provider_type
```
baarkerlounger, line 5: @kosiakkatrina I had this wrong, so fixing means the before save isn't needed anymore (in case you're interested)

kosiakkatrina: Have you tested it without the before save?

baarkerlounger: Yep

app/models/form/question.rb
```diff
@@ -104,14 +104,16 @@ def value_from_label(label)
...
+    label = case type
+            when "radio"
+              labels = answer_options[value.to_s]
+              labels["value"] if labels
+            when "select"
+              answer_options[value.to_s]
+            else
+              value.to_s
+            end
+    label || value
```
kosiakkatrina: should this be `label || value.to_s` ?

kosiakkatrina (approved):

PR #375, opened 2022-03-11
Remove tenancy/referral validations

kosiakkatrina:
Remove validations that only let internal referral to be selected for secure tenancy and vice versa. This validation seemed to be created by mistake and could not be located in the validations sheet


baarkerlounger (approved):
That's super weird, where did this come from?

kosiakkatrina:
> That's super weird, where did this come from?

@baarkerlounger No idea, but there's a reasoning behind the decision to remove it in the CLDC-465 comments

PR #376, opened 2022-03-11
Add reason and referral validations

kosiakkatrina:
reason for leaving last settled home cannot be "permanently decanted from another property owned by this landlord" if Q16 (What was the source of referral for this letting) is not  "internal transfer"

### DIFF
--- a/app/models/validations/household_validations.rb
+++ b/app/models/validations/household_validations.rb
@@ -20,7 +20,7 @@ module Validations::HouseholdValidations
     end
     validate_other_field(record, 31, :reason, :other_reason_for_leaving_last_settled_home)
 
-    if record.reason == 1 && record.referral.present? && !record.is_internal_transfer?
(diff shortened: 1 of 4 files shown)


app/models/validations/household_validations.rb
```diff
@@ -19,6 +19,11 @@ def validate_reason_for_leaving_last_settled_home(record)
       record.errors.add :reason, I18n.t("validations.household.underoccupation_benefitcap.dont_know_required")
     end
     validate_other_field(record, 31, :reason, :other_reason_for_leaving_last_settled_home)
+
+    if record.reason == 1 && record.referral.present? && !record.is_internal_transfer?
```
baarkerlounger: Worth a wrapper method to help make it obvious what 1 is?

baarkerlounger (approved):

app/models/validations/household_validations.rb
kosiakkatrina: Yo, you speedy! Will do!

PR #377, opened 2022-03-11
CLDC-996 Login lockout

smeny:


config/initializers/devise.rb
```diff
@@ -195,27 +195,27 @@
...
+  config.unlock_strategy = :none
 
   # Number of authentication tries before locking an account if lock_strategy
   # is failed attempts.
-  # config.maximum_attempts = 20
+  config.maximum_attempts = 5
 
   # Time interval to unlock the account if :time is enabled as unlock_strategy.
-  # config.unlock_in = 1.hour
+  config.unlock_in = 1.hour
```
baarkerlounger, line 215: Do you not need to set the strategy to time for this to take effect?

smeny: I don't think so, happy to remove it.

spec/features/auth/user_lockout_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+    before do
+      visit("/admin/sign-in")
+      attempt_number.times do
+        fill_in("admin_user[email]", with: admin.email)
+        fill_in("admin_user[password]", with: "wrong_password")
+        click_button("Sign in")
+      end
+    end
+
+    it "locks the admin account" do
```
baarkerlounger, line 40: Should probably also have a test for entering the 2FA code incorrectly 5 times?

config/initializers/devise.rb
baarkerlounger: Do we want the account to unlock after 1.hour? At the moment I don't think it will unless we also set the `unlock_strategy = :time` but I could be wrong.

spec/features/auth/user_lockout_spec.rb
smeny: I thought this was already tested in its own feature but it seems not. Adding an extra test.

baarkerlounger: Entering an incorrect code is, but not locking after 5 incorrect tries

config/initializers/devise.rb
smeny: I initially removed it, but I think it is a good idea to start with a default unlock by default and revisit.

baarkerlounger (approved):

PR #379, opened 2022-03-11
Derived rent and charges variables

baarkerlounger:


app/models/case_log.rb
```diff
@@ -85,6 +85,31 @@ def weekly_net_income
...
+      field_value / 52 * 49
+    when "Weekly for 48 weeks"
+      field_value / 52 * 48
+    when "Weekly for 47 weeks"
+      field_value / 52 * 47
+    when "Weekly for 46 weeks"
+      field_value / 52 * 46
+    when "Weekly for 52 weeks"
+      field_value
+    end
```
dushan-madetech: I'm surprised that Rubocop didn't throw this one at you https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Style/HashLikeCase unless we've disabled it

app/models/case_log.rb
```diff
@@ -231,6 +256,31 @@ def is_reason_permanently_decanted?
...
+      "Weekly for 49 weeks"
+    when 5
+      "Weekly for 48 weeks"
+    when 6
+      "Weekly for 47 weeks"
+    when 7
+      "Weekly for 46 weeks"
+    when 8
+      "Weekly for 52 weeks"
+    end
```
dushan-madetech: Same comment as above

dushan-madetech (approved):
I spotted something but it's not particularly important. Looks good enough to 🚢

app/models/case_log.rb
baarkerlounger: Interesting. I suspect it's because we're returning an operation rather than a static value so your hash would need to contain lambdas?

dushan-madetech (approved):

PR #380, opened 2022-03-11
Remove dead code

baarkerlounger:
This code doesn't get run anymore now that the design pattern is for only radio, select and date questions to have inline conditionals (and those are provided by the design system), so we can remove this nasty eval.


dushan-madetech (approved):
Love it 🎉

PR #385, opened 2022-03-14
CLDC-663: Weekly total shortfall derived variable

baarkerlounger:
https://digital.dclg.gov.uk/jira/browse/CLDC-663

WTSHORTFALL is the weekly total shortfall between housing-related benefits and the rent and charges combined.

WTSHORTFALL should only be calculated if HB=1 or HB=6 or HB=8 (note these categories will be changing in the 22/23 forms). WTSHORTFALL should be blank in all other situations, even if the user has submitted a value for TSHORTFALL.

TSHORTFALL = Total rent shortfall charge

PERIOD = Rent and other charge period

WTSHORTFALL = (TSHORTFALL / 52) x (no. of weeks from PERIOD)

### DIFF
--- a/app/models/case_log.rb
+++ b/app/models/case_log.rb
@@ -38,7 +38,7 @@ class CaseLog < ApplicationRecord
   OPTIONAL_FIELDS = %w[postcode_known la_known first_time_property_let_as_social_housing tenant_code propcode].freeze
   RENT_TYPE_MAPPING = { 0 => 1, 1 => 2, 2 => 2, 3 => 3, 4 => 3, 5 => 3 }.freeze
   RENT_TYPE_MAPPING_LABELS = { 1 => "Social Rent", 2 => "Affordable Rent", 3 => "Intermediate Rent" }.freeze
-  HAS_BENEFITS_OPTIONS = [0, 1, 2, 3].freeze
+  HAS_BENEFITS_OPTIONS = [1, 6, 8, 7].freeze
   STATUS = { "not_started" => 0, "in_progress" => 1, "completed" => 2 }.freeze
   NUM_OF_WEEKS_FROM_PERIOD = { 0 => 26, 1 => 13, 2 => 12, 3 => 50, 4 => 49, 5 => 48, 6 => 47, 7 => 46, 8 => 52 }.freeze
   enum status: STATUS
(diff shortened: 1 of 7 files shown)


app/models/case_log.rb
```diff
@@ -327,6 +356,9 @@ def set_derived_fields!
       self.wtcharge = weekly_value(tcharge) if tcharge.present?
     end
     self.has_benefits = get_has_benefits
+    self.wtshortfall = if tshortfall && receives_housing_related_benefits?
+                         weekly_value(tshortfall)
+                       end
```
kosiakkatrina, line 361: Do we want to reset it to nil in the `else` case?

app/models/validations/shared_validations.rb
```diff
@@ -21,7 +21,7 @@ def validate_numeric_min_max(record)
       field = question.check_answer_label || question.id
 
       begin
-        answer = Integer(record.public_send("#{question.id}_before_type_cast"))
+        answer = Float(record.public_send("#{question.id}_before_type_cast"))
```
kosiakkatrina: Nice!

app/models/case_log.rb
baarkerlounger: It will do that already since it'll assign whatever is returned which will implicitly be nil (there is a spec proving that).

app/models/validations/shared_validations.rb
baarkerlounger: Oops, this was from the other PR. Will merge that one first (once reviewed) and then rebase this one.

spec/models/case_log_spec.rb
```diff
@@ -299,6 +299,15 @@
...
         end
 
+        context "when the tenant is not in receipt of applicable benefits" do
+          it "correctly resets total shortfall" do
+            case_log.update!(wtshortfall: 100, hb: 9)
+            record_from_db = ActiveRecord::Base.connection.execute("select wtshortfall from case_logs where id=#{case_log.id}").to_a[0]
+            expect(case_log.wtshortfall).to be_nil
+            expect(record_from_db["wtshortfall"]).to be_nil
+          end
+        end
```
kosiakkatrina, line 309: Would this still pass if we first set benefits to yes?

kosiakkatrina: Nevermind!

kosiakkatrina (approved):

PR #386, opened 2022-03-14
CLDC-1050: re-add ethnic question

kiddhustle:
https://digital.dclg.gov.uk/jira/browse/CLDC-1050

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -284,7 +284,6 @@ ActiveRecord::Schema[7.0].define(version: 202202071123100) do
     t.integer "failed_attempts", default: 0
     t.string "unlock_token"
     t.datetime "locked_at", precision: nil
-    t.string "phone"
     t.index ["email"], name: "index_users_on_email", unique: true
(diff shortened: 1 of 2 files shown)


kosiakkatrina (approved):
Nice!

db/schema.rb
```diff
@@ -277,10 +277,10 @@
     t.string "last_sign_in_ip"
     t.integer "role"
     t.string "old_user_id"
+    t.string "phone"
```
baarkerlounger: Not sure if this should be here

db/schema.rb
```diff
@@ -284,7 +284,6 @@
     t.integer "failed_attempts", default: 0
     t.string "unlock_token"
     t.datetime "locked_at", precision: nil
-    t.string "phone"
```
baarkerlounger, line 287: @kiddhustle don't think this should have been here either. As a general rule of thumb should only really have schema changes if you have a migration.

PR #387, opened 2022-03-14
CLDC-815-long-term-vision-validation

dushan-madetech:
Prevents a submission if a condition effect has been selected but the answer to anyone in the household having a health condition is not yes. This validation might not have been necessary for the online form since the routing for the section is set up to prevent the condition effects question page from being visited is the answer to the previous question was not yes

### DIFF
--- a/app/models/validations/household_validations.rb
+++ b/app/models/validations/household_validations.rb
@@ -65,6 +65,13 @@ module Validations::HouseholdValidations
     end
   end
 
+  def validate_condition_effects(record)
+    all_options = [record.illness_type_1, record.illness_type_2, record.illness_type_3, record.illness_type_4, record.illness_type_5, record.illness_type_6, record.illness_type_7, record.illness_type_8, record.illness_type_9, record.illness_type_10]
+    if all_options.count(1) >= 1 && record.illness != 0
+      record.errors.add :condition_effects, I18n.t("validations.household.condition_effects.no_choices")
+    end   
+  end
+
   def validate_previous_housing_situation(record)
     if record.is_relet_to_temp_tenant? && !record.previous_tenancy_was_temporary?
       record.errors.add :prevten, I18n.t("validations.household.prevten.non_temp_accommodation")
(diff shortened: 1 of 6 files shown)


app/models/validations/household_validations.rb
```diff
@@ -65,6 +65,13 @@ def validate_accessibility_requirements(record)
     end
   end
 
+  def validate_condition_effects(record)
+    all_options = [record.illness_type_1, record.illness_type_2, record.illness_type_3, record.illness_type_4, record.illness_type_5, record.illness_type_6, record.illness_type_7, record.illness_type_8, record.illness_type_9, record.illness_type_10]
+    if all_options.count(1) >= 1 && record.illness != 0
```
baarkerlounger: Might be nice to have a wrapper around this for readability? Something like:

```
record.has_illness?
```

spec/models/validations/household_validations_spec.rb
```diff
@@ -482,6 +482,88 @@
...
+    end
+
+    it "validates other can't be selected if answer to anyone in household with health condition is not yes" do
+      record.illness = 1
+      record.illness_type_10 = 1
+      household_validator.validate_condition_effects(record)
+      expect(record.errors["condition_effects"])
+        .to include(match I18n.t("validations.household.condition_effects.no_choices"))
+    end
+  end
```
baarkerlounger, line 574: Be good to add a test for the positive case too, i.e. an expected answer not raising an error

app/models/validations/household_validations.rb
```diff
@@ -65,6 +65,14 @@ def validate_accessibility_requirements(record)
     end
   end
 
+  def validate_condition_effects(record)
+    all_options = [record.illness_type_1, record.illness_type_2, record.illness_type_3, record.illness_type_4, record.illness_type_5, record.illness_type_6, record.illness_type_7, record.illness_type_8, record.illness_type_9, record.illness_type_10]
+    all_options.count(1) >= 1
```
baarkerlounger: Think this line is redundant?

dushan-madetech: Removed it, was a bad replacement 😅

baarkerlounger (approved):

PR #388, opened 2022-03-14
Cldc 491 rent and charges validations (part 1)

kosiakkatrina:
- Move age validation tests to shared validation spec file

- Add `brent`/`tshortfall` validation: 
_Basic rent cannot be less than shortfall in basic rent
Basic rent cannot be less than double shortfall in basic rent_

- Update `period` mappings

- Add `scharge` validation for `this landlord`
_If Landlord is 1 This Landlord, this must be between 0 and 55 per week (GN only)_
_If Landlord is 1 This Landlord, this must be between 0 and 280 per week (SH only)_

- Add tests for float values in `weekly_value`

Error message content TBC
image
image

### DIFF
--- a/app/models/case_log.rb
+++ b/app/models/case_log.rb
@@ -266,6 +266,8 @@ class CaseLog < ApplicationRecord
 
   def benefits_unknown?
     hb == 3
+  end
+
   def this_landlord?
     landlord == 1
   end
--- a/app/models/case_log.rb
+++ b/app/models/case_log.rb
@@ -266,6 +266,8 @@ class CaseLog < ApplicationRecord
 
   def benefits_unknown?
     hb == 3
+  def this_landlord?
+    landlord == 1
   end
 
 private
--- a/app/models/case_log.rb
+++ b/app/models/case_log.rb
@@ -272,6 +272,10 @@ class CaseLog < ApplicationRecord
     landlord == 1
   end
 
+  def other_landlord?
+    landlord == 2
+  end
+
 private
 
   PIO = Postcodes::IO.new
--- a/app/models/case_log.rb
+++ b/app/models/case_log.rb
@@ -40,7 +40,7 @@ class CaseLog < ApplicationRecord
   RENT_TYPE_MAPPING_LABELS = { 1 => "Social Rent", 2 => "Affordable Rent", 3 => "Intermediate Rent" }.freeze
   HAS_BENEFITS_OPTIONS = [1, 6, 8, 7].freeze
   STATUS = { "not_started" => 0, "in_progress" => 1, "completed" => 2 }.freeze
-  NUM_OF_WEEKS_FROM_PERIOD = { 0 => 26, 1 => 13, 2 => 12, 3 => 50, 4 => 49, 5 => 48, 6 => 47, 7 => 46, 8 => 52 }.freeze
+  NUM_OF_WEEKS_FROM_PERIOD = { 2 => 26, 3 => 13, 4 => 12, 5 => 50, 6 => 49, 7 => 48, 8 => 47, 9 => 46, 1 => 52 }.freeze
   enum status: STATUS
 
   def form
--- a/app/models/case_log.rb
+++ b/app/models/case_log.rb
@@ -90,7 +90,7 @@ class CaseLog < ApplicationRecord
     num_of_weeks = NUM_OF_WEEKS_FROM_PERIOD[period]
     return unless field_value && num_of_weeks
 
-    field_value / 52 * num_of_weeks
+    (field_value / 52 * num_of_weeks).round(2)
   end
 
   def applicable_income_range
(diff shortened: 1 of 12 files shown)


app/models/validations/financial_validations.rb
```diff
@@ -58,4 +58,19 @@ def validate_tshortfall(record)
...
   end
+
+  def validate_rent_amount(record)
+    if record.brent.present? && record.tshortfall.present? && record.brent < record.tshortfall * 2
+      record.errors.add :brent, I18n.t("validations.financial.rent.less_than_double_shortfall", tshortfall: record.tshortfall * 2)
+      record.errors.add :tshortfall, I18n.t("validations.financial.tshortfall.more_than_rent")
+    end
+
+    if record.scharge.present? && record.this_landlord? && record.weekly_value(record.scharge).present?
+      if !record.weekly_value(record.scharge).between?(0, 55) && record.is_general_needs?
```
baarkerlounger: Do these numbers mean anything? Could they be named constants?

app/models/case_log.rb
```diff
@@ -268,6 +268,14 @@ def benefits_unknown?
     hb == 3
   end
 
+  def this_landlord?
```
baarkerlounger, line 271: What does `this` refer to here?

baarkerlounger (approved):

app/models/validations/financial_validations.rb
```diff
@@ -58,4 +58,27 @@ def validate_tshortfall(record)
...
+          record.errors.add :scharge, I18n.t("validations.financial.rent.scharge.this_landlord.supported_housing")
+        end
+      elsif record.other_landlord?
+        if !record.weekly_value(record.scharge).between?(0, 45) && record.is_general_needs?
+          record.errors.add :scharge, I18n.t("validations.financial.rent.scharge.other_landlord.general_needs")
+        elsif !record.weekly_value(record.scharge).between?(0, 165) && record.is_supported_housing?
+          record.errors.add :scharge, I18n.t("validations.financial.rent.scharge.other_landlord.supported_housing")
+        end
+      end
+    end
```
kosiakkatrina: There's a smaller function inside this function, struggling to get out. I'll find it in a bit

app/models/case_log.rb
kosiakkatrina: It's just one of landlord values from dictionary 🤷‍♀️ Can ask Rachel maybe

baarkerlounger: Yeah not a blocker for this PR at all, just a little vague how that's worded in the data dictionary so would be good to make it clearer here.

app/models/validations/financial_validations.rb
kosiakkatrina: Hmmm, will deffo refactor it into something that pulls these from a hash or something, cause there are a few more of these. Not sure what they are other than just ranges

app/models/validations/financial_validations.rb
```diff
@@ -58,4 +58,48 @@ def validate_tshortfall(record)
...
+                        SCHARGE_RANGES[:other_landlord_general_needs]
+                      elsif record.is_supported_housing?
+                        SCHARGE_RANGES[:other_landlord_supported_housing]
+                      end
+                    end
+
+    if scharge_range.present? && !weekly_value_in_range(record, "scharge", scharge_range[:min], scharge_range[:max])
+      record.errors.add :scharge, scharge_range[:error]
+    end
+  end
```
kosiakkatrina: @baarkerlounger this is not ideal and I feel like I'm missing a very basic refactor somewhere here, any suggestions?

baarkerlounger: If you make it a nested hash you could do a direct look up. Something like:

```
private

SCHARGE_RANGES = {
  this_landlord: {
    general_needs: {
      range: 0..55,
      error: I18n.t("validations.financial.rent.scharge.this_landlord.general_needs")
    },
    supported_housing: {
      range: 0..200,
      error: I18n.t("validations.financial.rent.scharge.this_landlord.supported_housing")
    }
  },
  other_landlord: {
    general_needs: {
      range: 0..45,
      error: I18n.t("validations.financial.rent.scharge.other_landlord.general_needs")
    },
    supported_housing: {
      range: 0..165,
      error: I18n.t("validations.financial.rent.scharge.other_landlord.supported_housing")
    }
  }
}.freeze

  def validate_scharge(record)
    scharge = SCHARGE_RANGES.dig(:this_landlord, :general_needs)
     if scharge_range.present? && !weekly_value_in_range(record, "scharge", scharge[:range])
       record.errors.add :scharge, scharge[:error]
     end
  end
```

kosiakkatrina: Hmm, that still requires writing out all the if statements for is_this_landlord and is_general_needs tho?

app/models/validations/financial_validations.rb
```diff
@@ -58,4 +58,76 @@ def validate_tshortfall(record)
       record.errors.add :tshortfall, I18n.t("validations.financial.hbrentshortfall.outstanding_no_benefits")
     end
   end
+
+  CHARGE_MAXIMUMS = {
```
baarkerlounger, line 84: This constant can probably be private too? Not expected to be called from anywhere else right?

baarkerlounger (approved):

PR #391, opened 2022-03-15
CLDC-812 rent type validation

dushan-madetech:
Screenshot 2022-03-16 at 10 47 36

### DIFF
--- a/spec/models/case_log_spec.rb
+++ b/spec/models/case_log_spec.rb
@@ -56,6 +56,10 @@ RSpec.describe CaseLog do
       expect(validator).to receive(:validate_startdate)
     end
 
+    it "validates intermediate rent product name" do
+      expect(validator).to receive(:validate_intermediate_rent_product_name)
+    end
+
     it "validates other household member details" do
       expect(validator).to receive(:validate_household_number_of_other_members)
     end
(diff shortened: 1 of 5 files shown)


baarkerlounger:
Can you also add a spec that tests that new validation method gets called on CaseLog update - check the #update describe block.

smeny (approved):

smeny (changes requested):
Just to be sure Dan's comment is visible :)

spec/models/case_log_spec.rb
```diff
@@ -56,6 +56,10 @@
       expect(validator).to receive(:validate_startdate)
     end
 
+    it "validates intermediate rent product name" do
```
dushan-madetech, line 59: @baarkerlounger Is this what you had in mind?

spec/models/case_log_spec.rb
```diff
@@ -56,6 +56,10 @@
...
+      expect(validator).to receive(:validate_intermediate_rent_product_name)
+    end
+
```
dushan-madetech, line 62: @baarkerlounger Is this what you had in mind?

baarkerlounger (approved):
Nice one

PR #392, opened 2022-03-15
Switch to JSBundling with webpack

baarkerlounger:
Follow on to https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/203

Since Webpacker is now officially considered deprecated by the Rails team and it's JS dependencies won't receive security updates any more we need to consider migrating off it at some point. 

There are several options but we need to consider that we need to support ES5/IE11 and have ES6 JS in our dependencies (Govuk frontend).

1. **Import maps** (Rails 7 default). Serve JS directly but do no transpiling so not suitable
2. **JSBundling** (Rails recommended)
      - With **ESBuild** (fast and does some transpiling but doesn't support ES5/IE11 - see previous PR) (https://esbuild.github.io/)
      - With **Rollup** (Similar to ESBuild, node rather than Go based, doesn't have the big speed benefits) (https://www.rollupjs.org/guide/en/)
      - With **Webpack** - rather than the old approach of using Webpacker as a opinionated wrapper around webpack, this approach uses webpack directly. (https://webpack.js.org/)
3. **Shakapacker** - the "official" community maintained fork of Webpacker 6 RC. Requires upgrading current install since breaking changes happened between Webpacker 5 & 6. (https://github.com/shakacode/shakapacker)
4. **Vite** - Webpack alternative (https://vite-ruby.netlify.app/).

Option 2.3 - JSBundling with Webpack is what this PR opts for.

It moves us onto the standard Rails JS bundling approach while maintaining the functionality we need (compatibility with Babel), which leaves open a potential easier migration path to faster options like ESBuild if and when IE11 is fully out of action.

- [x] Add Babel config targeting IE11
- [x] Update ReadMe

**IE 11 issues:**

**Hotwire (Turbo/Stimulus) do not support IE 11. Stimulus dropped IE 11 support after V2.0.0. We could consider downgrading to that. Turbo requires further research. This isn't a problem specific to this form of bundling and transpiling our frontend assets though, it affects our existing stack as well. As such it might be worth merging this and dealing with that separately.**

- [x] Polyfill/Transpile Stimulus such that it doesn't break _other_ JS
- [x] Polyfill/Transpile our Stimulus controllers such that they themselves work in general
- [x] Make conditional question stimulus controller work in IE11
- [ ] Make numeric calculation stimulus controller work in IE11 (controller works but IE doesn't support `<output>` tag). Covered by https://digital.dclg.gov.uk/jira/browse/CLDC-1084
- [ ] Make tasklist highlighter stimulus controller work in IE11 (doesn't work in any browser - broken on main). Covered by https://digital.dclg.gov.uk/jira/browse/CLDC-1085
- [x] Make accessible autocomplete stimulus controller work in IE11
- [ ] Polyfill/Transpile Turbo - not yet working, attempting to transpile Turbo throws t.tag undefined errors in both IE 11 and Modern browsers. Turning off Turbo altogether is potentially an option but would break our form routing logic so would need to do some work to do that. Getting Turbo to work would be far better. 

IE 11 Support:

We set transpilation targets in package.json via [browserslist](https://github.com/browserslist/browserslist). This configuration covers:

```
$ npx browserslist
and_chr 99
and_ff 96
and_qq 10.4
and_uc 12.12
android 99
baidu 7.12
bb 10
chrome 99
chrome 98
chrome 97
edge 99
edge 98
firefox 98
firefox 97
firefox 96
ie 11
ie 10
ie_mob 11
ios_saf 15.4
ios_saf 15.2-15.3
ios_saf 15.0-15.1
ios_saf 14.5-14.8
kaios 2.5
op_mini all
op_mob 64
opera 83
safari 15.4
samsung 16.0
```

We also explicitly import support for promises for IE 11 from CoreJS.


config/initializers/active_admin.rb
```diff
@@ -331,14 +331,7 @@
...
   #
-  config.use_webpacker = true
-end
-
-# Print stylesheet is no longer a separate stylesheet in upstream so we can
-# delete this when the next version bump is released but for now it prevents
-# webpacker issues.
-Rails.application.config.after_initialize do
-  ActiveAdmin.application.stylesheets.delete("active_admin/print.css")
+  # config.use_webpacker = true
```
kosiakkatrina, line 334: do we need this?

baarkerlounger: No that's why it's commented out now but it (commented version) is part of the default generated config file so I've just reverted it to that.

dushan-madetech (approved):
Screenshot 2022-03-21 at 10 25 27

95-955352_italian-chef-kiss-emoji-freetoedit-italian-chef-kiss

paulrobertlloyd:
It’s been a journey!

Hopefully this works better than Webpacker (which is a low bar, to be fair).

PR #394, opened 2022-03-16
Outlaw underscores in URLs

paulrobertlloyd:
Add a test to prevent underscores slipping into URLs

### DIFF
--- a/spec/config/routes_spec.rb
+++ b/spec/config/routes_spec.rb
@@ -1,18 +1,23 @@
 require "rails_helper"
 
 RSpec.describe "routes.rb" do
-  it "does not use underscores" do
-    paths = Rails.application.routes.routes.map { |r| r.path.spec.to_s if r.defaults[:controller] }.compact
-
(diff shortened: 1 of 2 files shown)


spec/config/routes_spec.rb
```diff
@@ -0,0 +1,18 @@
+require "rails_helper"
+
+RSpec.describe "routes.rb" do
+  it "does not use underscores" do
+    paths = Rails.application.routes.routes.map { |r| r.path.spec.to_s if r.defaults[:controller] }.compact
+
+    # Allow underscores for ActiveAdmin, Rails and Turbo routes
+    paths = paths.reject { |p| p.starts_with?("/admin") }
+    paths = paths.reject { |p| p.starts_with?("/rails") }
+    paths = paths.reject { |p| p.include?("_historical_location") }
```
paulrobertlloyd: Potentially dodgy, but an attempt to catch the URL scheme Turbo uses.

baarkerlounger (approved):

PR #396, opened 2022-03-17
CLDC-651 refused derived variable

dushan-madetech:
This PR creates the derived variable `refused` and add the logic for deriving it

### DIFF
--- a/app/models/case_log.rb
+++ b/app/models/case_log.rb
@@ -355,6 +355,7 @@ private
     self.totchild = get_totchild
     self.totelder = get_totelder
     self.totadult = get_totadult
+    self.refused = get_refused
     if %i[brent scharge pscharge supcharg].any? { |f| public_send(f).present? }
       self.brent ||= 0
       self.scharge ||= 0
@@ -446,6 +447,15 @@ private
       relat = public_send("relat#{i}")
       !age.nil? && ((age >= 16 && age < 18 && [0, 2].include?(relat)) || age >= 18 && age < 60)
     end
+  end
+
+  def get_refused
+    return 1 if [age1_known, age2_known, age3_known, age4_known, age5_known, age6_known, age7_known, age8_known].any?(1)
+    return 1 if [sex1, sex2, sex3, sex4, sex5, sex6, sex7, sex8].any?("R")
+    return 1 if [relat2, relat3, relat4, relat5, relat6, relat7, relat8].any?(3)
+    return 1 if [ecstat1, ecstat2, ecstat3, ecstat4, ecstat5, ecstat6, ecstat7, ecstat8].any?(10)
+
+    0
   end
 
   def get_inferred_la(postcode)
--- a/app/models/case_log.rb
+++ b/app/models/case_log.rb
@@ -450,10 +450,7 @@ private
   end
 
   def get_refused
-    return 1 if [age1_known, age2_known, age3_known, age4_known, age5_known, age6_known, age7_known, age8_known].any?(1)
-    return 1 if [sex1, sex2, sex3, sex4, sex5, sex6, sex7, sex8].any?("R")
-    return 1 if [relat2, relat3, relat4, relat5, relat6, relat7, relat8].any?(3)
-    return 1 if [ecstat1, ecstat2, ecstat3, ecstat4, ecstat5, ecstat6, ecstat7, ecstat8].any?(10)
+    return 1 if age_refused? || sex_refused? || relat_refused? || ecstat_refused?
 
     0
   end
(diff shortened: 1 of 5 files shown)


app/models/case_log.rb
```diff
@@ -431,6 +432,15 @@ def get_totadult
     end
   end
 
+  def get_refused
+    return 1 if [age1_known, age2_known, age3_known, age4_known, age5_known, age6_known, age7_known, age8_known].any?(1)
+    return 1 if [sex1, sex2, sex3, sex4, sex5, sex6, sex7, sex8].any?("R")
+    return 1 if [relat2, relat3, relat4, relat5, relat6, relat7, relat8].any?(3)
+    return 1 if [ecstat1, ecstat2, ecstat3, ecstat4, ecstat5, ecstat6, ecstat7, ecstat8].any?(10)
```
kosiakkatrina: Would it make sense to have methods for these and do something like:
`return 1 if age_refused? || gender_refused? ...`

app/models/case_log.rb
```diff
@@ -431,6 +432,15 @@ def get_totadult
...
+
+    0
```
kosiakkatrina, line 455: should this be nil? or just nothing which will return nil from the method

app/models/case_log.rb
kosiakkatrina: Might just end up being a one liner instead of a method 👀

app/models/case_log.rb
```diff
@@ -431,6 +432,15 @@ def get_totadult
     end
   end
 
+  def get_refused
+    return 1 if [age1_known, age2_known, age3_known, age4_known, age5_known, age6_known, age7_known, age8_known].any?(1)
```
kosiakkatrina: should it be `-9` instead of `1` or is there an error in the ticket?

app/models/case_log.rb
```diff
@@ -431,6 +432,15 @@ def get_totadult
...
+    return 1 if [sex1, sex2, sex3, sex4, sex5, sex6, sex7, sex8].any?("R")
+    return 1 if [relat2, relat3, relat4, relat5, relat6, relat7, relat8].any?(3)
```
kosiakkatrina: Same here, from the ticket seems like it's expecting `R` instead of 3. Are we saving it in the db differently?

app/models/case_log.rb
dushan-madetech: There's no `-9` since the ticket was for `age` but based on the new design we use `age_known` which just has two possible values (yes or no)

app/models/case_log.rb
dushan-madetech: Yes - `3` is `prefers not to say`

app/models/case_log.rb
baarkerlounger: Technically this is "Don't know" rather than "Tenant prefers not to say" (which == refused) right? Should we add an option for that? cc. @paulrobertlloyd

baarkerlounger (commented):
Is it correct that we just set a single field if any attribute of any tenant has been refused?

app/models/case_log.rb
baarkerlounger: Does that match what's in the data dictionary?

dushan-madetech:
> Is it correct that we just set a single field if any attribute of any tenant has been refused?

@baarkerlounger according to the ticket `REFUSED = 1 if any of the household characteristics question have been answered as 'refused' (now 'prefer not to say') for any of the household members`

app/models/case_log.rb
dushan-madetech: Updated, good shout - knew I forgot to do something 😄

app/models/case_log.rb
dushan-madetech: I think setting it to 0 is fine here, since the main concern is about when it's set to 1

db/schema.rb
```diff
@@ -297,10 +298,10 @@
     t.string "last_sign_in_ip"
     t.integer "role"
     t.string "old_user_id"
+    t.string "phone"
```
baarkerlounger: Don't think these two (phone lines) should be here. Can you revert them?

baarkerlounger (approved):
1 comment but otherwise good to go

db/schema.rb
dushan-madetech: they're in the schema on main https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/blob/main/db/schema.rb

baarkerlounger: Yep, I don't mean they should be removed altogether. Just that this PR shouldn't be changing them (since there's no migration for that here).

baarkerlounger (approved):

db/schema.rb
dushan-madetech: Fixed it 🙂

PR #397, opened 2022-03-18
Use Adminterface theme for ActiveAdmin and adjust to match main service

paulrobertlloyd:
Use [Adminterface](https://adminterface.io/)[^1] theme for ActiveAdmin, adjusting Bootstrap variables and overriding specific elements where required.

It’s a bit uncanny valley in places, and as more data is included other styles may need to be tweaked, but this UI:

* is responsive
* has a greater degree of colour contract 
* includes focus styles
* is more aligned with the main service.

(Bulk of this was done in about an hour, but then several hours were spent tweaking everything to get it closer to looking like GOV.UK 😅) 

[^1]: No longer being maintained, which is a bit annoying, but hopefully not too much of an issue.

| Before | After |
| - | - |
| Screenshot 2022-03-18 at 00 14 53 | Screenshot 2022-03-18 at 00 01 48 |
| Screenshot 2022-03-18 at 00 15 17 | Screenshot 2022-03-18 at 00 02 07 |
| Screenshot 2022-03-18 at 00 15 36 | Screenshot 2022-03-18 at 00 03 53 |

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -100,6 +100,12 @@ GEM
       tzinfo (~> 2.0)
     addressable (2.8.0)
       public_suffix (>= 2.0.2, < 5.0)
+    adminterface (0.3.0)
+      activeadmin (~> 2.9)
+      httparty (>= 0.18)
+      image_processing (~> 1.0)
+      rails (>= 6.0)
+      rainbow (~> 3.0)
     arbre (1.5.0)
       activesupport (>= 3.0.0, < 7.1)
       ruby2_keywords (>= 0.0.2, < 1.0)
@@ -192,8 +198,14 @@ GEM
       rails (>= 6.0.0)
       stimulus-rails
       turbo-rails
+    httparty (0.20.0)
+      mime-types (~> 3.0)
+      multi_xml (>= 0.5.2)
     i18n (1.10.0)
       concurrent-ruby (~> 1.0)
+    image_processing (1.12.2)
+      mini_magick (>= 4.9.5, < 5)
+      ruby-vips (>= 2.0.17, < 3)
     inherited_resources (1.13.1)
       actionpack (>= 5.2, < 7.1)
       has_scope (~> 0.6)
@@ -232,9 +244,14 @@ GEM
     marcel (1.0.2)
     matrix (0.4.2)
     method_source (1.0.0)
+    mime-types (3.4.1)
+      mime-types-data (~> 3.2015)
+    mime-types-data (3.2022.0105)
+    mini_magick (4.11.0)
     mini_mime (1.1.2)
     minitest (5.15.0)
     msgpack (1.4.5)
+    multi_xml (0.6.0)
     net-imap (0.2.3)
       digest
       net-protocol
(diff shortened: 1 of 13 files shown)


dushan-madetech:
Not Dan but figured I'd pop in my thoughts:

1) This looks much cleaner than the old version, I really like it 😄 

2) It seems there are a couple of failing specs at the moment, let me know if you need a hand with debugging those.

3) Regarding the footnote - I don't think it's much of an issue that it's not being maintained, I'd almost even be willing to say that we could potentially fork it and add any improvements/changes we want to that fork and then reference that fork in our repository (if that's something we wanted to do).

Gemfile.lock
```diff
@@ -100,6 +100,12 @@ GEM
       tzinfo (~> 2.0)
     addressable (2.8.0)
       public_suffix (>= 2.0.2, < 5.0)
+    adminterface (0.3.0)
+      activeadmin (~> 2.9)
```
baarkerlounger, line 104: This is a problem. We need ActiveAdmin 2.11+ for Rails 7 and the fact this repo is unmaintained now (not accepting pull requests either) means we'd definitely have to fork this whole project to work around this.

app/webpacker/packs/active_admin.js
```diff
@@ -1,7 +1,13 @@
-// Load Active Admin's styles into Webpacker,
-// see `active_admin.scss` for customization.
+// Load ActiveAdmin’s styles into Webpacker.
+// See `active_admin.scss` for customisation.
 import "../styles/active_admin";
 
-import "@activeadmin/activeadmin";
+import Rails from '@rails/ujs'
```
baarkerlounger, line 5: Do we need this? We're not actually using `ujs` or `activestorage` so unless this introduces them can probably remove them here.

config/webpack/environment.js
```diff
@@ -1,5 +1,7 @@
 const { environment } = require('@rails/webpacker')
-const jquery = require('./plugins/jquery')
 
-environment.plugins.prepend('jquery', jquery)
+environment.loaders.get('sass').use.splice(-1, 0, {
```
baarkerlounger, line 3: Moving away from webpacker so this will probably need to be modified for that once it's merged.

config/webpack/plugins/jquery.js
```diff
@@ -1,7 +0,0 @@
-const webpack = require('webpack')
```
baarkerlounger, line 1: Does this just remove the need for JQuery entirely? Cause that's pretty nice...

baarkerlounger (commented):
I think given the large number of dependencies this pulls in, the fact that it's totally unmaintained now is a problem.

The immediate one that would need bumping is ActiveAdmin but I suspect over time it would be harder and harder to keep this going (who would maintain it long term?). Would the overhead be worth it given the uncanny valley effect? If there's significant effort to bringing this in (forking & maintaining) - how much effort would it be to copy the approach but use the Govuk frontend directly rather than bootstrap?

That said it does look great!

paulrobertlloyd:
Useful as a proof of concept, and instigating a conversation about what we actually want to build for internal users.

PR #398, opened 2022-03-18
CLDC-656 housing needs derived variable

dushan-madetech:
Adds the `housingneeds` derived variable and rules for deriving it

### DIFF
--- a/app/models/case_log.rb
+++ b/app/models/case_log.rb
@@ -371,6 +371,7 @@ private
     self.nocharge = household_charge&.zero? ? 1 : 0
     self.underoccupation_benefitcap = 3 if renewal == 1 && year == 2021
     self.ethnic = ethnic || ethnic_group
+    self.housingneeds = get_housingneeds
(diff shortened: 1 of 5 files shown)


app/models/case_log.rb
```diff
@@ -480,6 +481,19 @@ def get_lettype
     end
   end
 
+  def get_housingneeds
```
baarkerlounger, line 484: Any way of adding a bit more context/meaning to this method?

baarkerlounger (approved):

app/models/case_log.rb
dushan-madetech: @baarkerlounger I've updated this, let me know if this is better? https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/398/commits/49a4bc0230ec2317c9afaf3801b6082740b13725

baarkerlounger: Nice :+1:

PR #400, opened 2022-03-21
URL audit

paulrobertlloyd:
* Update URL slugs in 2021/22 questions
* Use `/account` for devise routes under `:users`

### DIFF
--- a/app/controllers/auth/passwords_controller.rb
+++ b/app/controllers/auth/passwords_controller.rb
@@ -66,7 +66,7 @@ protected
   end
 
   def after_sending_reset_password_instructions_path_for(_resource)
-    confirmations_reset_path(email: params.dig(resource_class_name, "email"))
+    account_password_reset_confirmation_path(email: params.dig(resource_class_name, "email"))
   end
 
   def after_resetting_password_path_for(resource)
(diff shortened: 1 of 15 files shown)


config/routes.rb
```diff
@@ -10,21 +10,32 @@
...
+    },
+    path_names: {
+      sign_in: "sign-in",
+      sign_out: "sign-out",
+    },
+  }
 
   devise_scope :user do
-    get "confirmations/reset", to: "auth/passwords#reset_confirmation"
+    get "account/password/reset-confirmation", to: "auth/passwords#reset_confirmation"
```
baarkerlounger, line 38: I think you can use `as: confirmations/reset` to update the URL without changing the named route helper so you don't need to change the controller code above. Think that's probably a little safer in case devise expects that named route to exist elsewhere.

app/controllers/auth/passwords_controller.rb
```diff
@@ -66,7 +66,7 @@ def resource_class_name
   end
 
   def after_sending_reset_password_instructions_path_for(_resource)
-    confirmations_reset_path(email: params.dig(resource_class_name, "email"))
+    account_password_reset_confirmation_path(email: params.dig(resource_class_name, "email"))
```
baarkerlounger, line 69: See comment below. I think you can avoid the need for doing this?

config/routes.rb
baarkerlounger: Actually looks like we set this so changing it is probably cleaner :+1:

baarkerlounger (approved):
LGTM

baarkerlounger (approved):

PR #404, opened 2022-03-21
Log status on log

paulrobertlloyd:
* Fix link to next incomplete section ([CLDC-1085](https://digital.dclg.gov.uk/jira/browse/CLDC-1085))
* Update text for a log that has not been started: `This log has not been started.`
* Update text for in progress log: `X of X sections completed.` (removal of another ‘you’ – multiple users can collaborate on a log)

### DIFF
--- a/spec/features/form/tasklist_page_spec.rb
+++ b/spec/features/form/tasklist_page_spec.rb
@@ -42,4 +42,13 @@ RSpec.describe "Task List" do
     visit("/logs/#{empty_case_log.id}")
     expect(page).to have_link("Skip to next incomplete section", href: /#household-characteristics/)
   end
+
+  context "when the 'Skip to next incomplete section is clicked'" do
+    it "jumps to the next incomplete section" do
+      answer_all_questions_in_income_subsection(empty_case_log)
+      visit("/logs/#{empty_case_log.id}")
+      click_link("Skip to next incomplete section")
+      expect(page).to have_current_path("/logs/#{empty_case_log.id}#household-characteristics")
+    end
+  end
 end
--- a/spec/features/form/tasklist_page_spec.rb
+++ b/spec/features/form/tasklist_page_spec.rb
@@ -26,19 +26,20 @@ RSpec.describe "Task List" do
     sign_in user
   end
 
-  it "skips to the first section if no answers are completed" do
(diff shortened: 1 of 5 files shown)


spec/features/form/tasklist_page_spec.rb
```diff
@@ -26,19 +26,20 @@
...
-    expect(page).to have_content("You have completed 0 of 9 sections.")
+    expect(page).to have_content("1 of 9 sections completed.")
   end
 
-  it "shows the number of completed sections if one section is completed" do
+  it "show skip link for next incomplete section" do
     answer_all_questions_in_income_subsection(empty_case_log)
     visit("/logs/#{empty_case_log.id}")
-    expect(page).to have_content("You have completed 1 of 9 sections.")
+    expect(page).to have_link("Skip to next incomplete section", href: /#household-characteristics/)
```
baarkerlounger, line 43: Would be great to add one more test here for what should happen when the link is clicked (what html change do we expect?).

paulrobertlloyd: Is it possible to test that page has scrolled to the `:target`? Maybe test if href has `#household-characteristics` appended?

baarkerlounger: Is that all that's supposed to happen now? No highlighting any more? In that case yes that's probably the best check.

paulrobertlloyd: Okay, how might I write that test 😉

baarkerlounger: Something like this:

```
context "when the 'Skip to next incomplete section is clicked'" do
    it "jumps to the next incomplete section" do
      answer_all_questions_in_income_subsection(empty_case_log)
      visit("/logs/#{empty_case_log.id}")
      click_link("Skip to next incomplete section")
      expect(page).to have_current_path("/logs/#{empty_case_log.id}#household-section")
    end
  end
```

baarkerlounger: Doesn't look like the functionality actually works at the moment though

baarkerlounger:
Closed for https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/410

PR #405, opened 2022-03-21
Link to privacy policy page, not old PDF

paulrobertlloyd:
Link to the current and accessible privacy policy, not “that ghastly blue thing”.


baarkerlounger (approved):
Alright alright, no need to be rude about it :joy: .

PR #406, opened 2022-03-21
Remove hotwire turbo

baarkerlounger:
Turbo does not support IE 11/ES5 and has no plans to add support. 

This leaves 3 options:

1. Transpile Turbo with Babel and add required polyfills such that it works in IE11 and modern browsers (same approach as with Stimulus). So far unsuccessful.
2. Conditionally load Turbo based on browser feature detection with modernizr (https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/430)
3. Remove Turbo altogether

Without Turbo each form submission or link click requires a full page load. This does have a perceptible impact on UX/speed but provides the most compatibility with older browsers.  

Given the time spent on this so far it seems worth merging this for now and reverting later if we get it working at some point in the future.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -25,8 +25,8 @@ gem "govuk_design_system_formbuilder"
 gem "govuk_markdown"
 # GOV UK Notify
 gem "notifications-ruby-client"
-# Turbo and Stimulus
-gem "hotwire-rails"
+# A modest javascript framework for the html you already have
+gem "stimulus-rails"
 # Administration framework
 gem "activeadmin"
 # Admin charts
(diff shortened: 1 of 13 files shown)


paulrobertlloyd (approved):
Need to remove the progress bar styles, too. (Good while they lasted!)

kosiakkatrina (approved):

tvararu:
Thanks for writing this up! I was actually looking for examples of "gov rails projects using turbo" to see how viable it is, and whether it has any impact on browser support or accessibility (thinking SPA style page navigation could interfere with some screen readers).

PR #408, opened 2022-03-22
Feature/cldc 1050 ethnic background 2

kiddhustle:
https://digital.dclg.gov.uk/jira/browse/CLDC-1050

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,9 @@
 # or operating system, you probably want to add a global ignore instead:
 #   git config --global core.excludesfile '~/.gitignore_global'
 
+# generated files
+db.schema.rb
+
 # Ignore bundler config.
 /.bundle
 
--- a/.gitignore
+++ b/.gitignore
@@ -4,9 +4,6 @@
 # or operating system, you probably want to add a global ignore instead:
 #   git config --global core.excludesfile '~/.gitignore_global'
 
-# generated files
-db.schema.rb
-
 # Ignore bundler config.
 /.bundle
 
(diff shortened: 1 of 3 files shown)


.gitignore
```diff
@@ -4,6 +4,9 @@
 # or operating system, you probably want to add a global ignore instead:
 #   git config --global core.excludesfile '~/.gitignore_global'
 
+# generated files
+db.schema.rb
```
baarkerlounger: This shouldn't be in .gitignore because you do want to commit it sometimes, just only when you're actually making database changes.

i.e. if you've written a migration that changes the schema, you should also commit the schema. But if your local schema has just changed because your local schema has drifted out of sync with what migrations create you shouldn't.

db/schema.rb
```diff
@@ -281,10 +281,10 @@
     t.string "last_sign_in_ip"
     t.integer "role"
     t.string "old_user_id"
+    t.string "phone"
```
baarkerlounger: This looks like your DB has gotten out of sync. If you reset your database (`rake db:drop; rake db:create; rake db:migrate`), I think this goes away?

kiddhustle: aah maybe so I will give it a try

.gitignore
kiddhustle: reverted

db/schema.rb
kiddhustle: reverted

baarkerlounger (approved):

kosiakkatrina (approved):

baarkerlounger:
Covered by https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/417/files

baarkerlounger:
Covered by https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/417/files

PR #411, opened 2022-03-23
Cldc 491 rent and charges validations

kosiakkatrina:


app/models/validations/financial_validations.rb
```diff
@@ -121,13 +122,29 @@ def validate_charges(record)
...
-  def weekly_value_in_range(record, field, max)
-    record[field].present? && record.weekly_value(record[field]).present? && record.weekly_value(record[field]).between?(0, max)
+  def weekly_value_in_range(record, field, min, max)
+    record[field].present? && record.weekly_value(record[field]).present? && record.weekly_value(record[field]).between?(min, max)
+  end
+
+  def validate_rent_range(record)
+    return if record.startdate.blank?
+
+    window_end_date = Time.zone.local(record.year, 4, 1)
```
baarkerlounger: I wonder if we should make this a `case_log` method? Think we do it in more than one place now. (i.e. `case_log.window_end_date`)

lib/tasks/rent_ranges.rake
```diff
@@ -0,0 +1,26 @@
...
+
+namespace :data_import do
+  desc "Import annual rent range data"
+  task :rent_ranges, %i[start_year path] => :environment do |_task, args|
+    start_year = args[:start_year]
+    path = args[:path]
+
+    raise "Usage: rake data_import:rent_ranges[start_year,'path/to/csv_file']" if path.blank? || start_year.blank?
+
+    CSV.foreach(path, headers: true) do |row|
```
baarkerlounger, line 12: Have you tried this locally for the whole data set? How long does it take?

spec/models/validations/financial_validations_spec.rb
```diff
@@ -713,6 +713,93 @@
         expect(record.errors["household_charge"])
           .to be_empty
       end
+
+      context "when validating ranges based on LA and needstype" do
+        before do
+          LaRentRange.find_or_create_by(
```
baarkerlounger: Worth a Factory? Probably doesn't need to be a `find_or` since we're in a test?

baarkerlounger (approved):
Couple of questions but great work!

lib/tasks/rent_ranges.rake
kosiakkatrina: I did! Maybe 10 seconds 😁 Very satisfying!

app/models/validations/financial_validations.rb
kosiakkatrina: There is a `form_name` method that gets the full form name, but happy to extract a smaller method out of there!

spec/models/validations/financial_validations_spec.rb
kosiakkatrina: Hm, yes, so it does fail with just create, I tried truncating the table between the tests, but this is looks a bit neater methinks

app/models/case_log.rb
```diff
@@ -48,15 +48,16 @@ def form
     FormHandler.instance.get_form(form_name) || FormHandler.instance.forms.first.second
   end
 
+  def get_collection_start_year
```
baarkerlounger: Style thing but I usually prefer to name the method the thing it returns and use it directly. So this would become

```
def collection_start_year
    window_end_date = Time.zone.local(startdate.year, 4, 1)
    startdate < window_end_date ? startdate.year - 1 : startdate.year
end

def form_name
  "#{collection_start_year}_#{collection_start_year + 1}"
end
```

The `get`, `set` thing reminds me of java... 

Just personal preference though so feel free to ignore entirely!

spec/models/validations/financial_validations_spec.rb
baarkerlounger: That seems like a potential smell since test data should be isolated... worth checking where those records are coming from I think

baarkerlounger (approved):

app/models/case_log.rb
kosiakkatrina: Nah, that makes sense!

kosiakkatrina:
> That seems like a potential smell since test data should be isolated... worth checking where those records are coming from I think

@baarkerlounger changed it to just create and ran it a few times, works okay by the looks of it. Not sure what fixed it 🤔

spec/lib/tasks/rent_range_import_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+      task.reenable
+    end
+
+    context "when the rake task is run" do
+      let(:start_year) { 2021 }
+      let(:rent_ranges_file_path) { "./spec/fixtures/files/rent_ranges.csv" }
+      let(:wrong_file_path) { "/test/no_csv_here.csv" }
+
+      before do
+        LaRentRange.delete_all
```
baarkerlounger, line 20: Same with this I think? Shouldn't be necessary?

kosiakkatrina: 😭 I merged. There are a few more validations that will likely need to go in, so I'll chuck the fix in with those

PR #413, opened 2022-03-23
CLDC-668 wchchrg derived variable

dushan-madetech:
adds the wchchrg (weekly care home charge) derived variable and the rules for deriving it


baarkerlounger (approved):
LGTM

PR #420, opened 2022-03-24
Household member validations

baarkerlounger:
https://digital.dclg.gov.uk/jira/browse/CLDC-1066

- [x] Adds validation that if age is > 16 the economic status can't be child under 16
- [x] For compound validations in the household sections adds error messages to all fields that could trigger it
- [x] Updates the ecstat values to match the data dictionary (attached in Jira)
- [x] Updates the allowed income ranges to match the data dictionary (attached in Jira)


config/locales/en.yml
```diff
@@ -152,15 +152,24 @@ en:
       age:
```
baarkerlounger, line 152: @Dan13L3 could you check these messages please?

app/models/validations/household_validations.rb
```diff
@@ -131,11 +131,17 @@ def validate_person_age_matches_economic_status(record, person_num)
...
+      record.errors.add "age#{person_num}", I18n.t("validations.household.age.retired_over_70", person_num:)
     end
-    if age < 16 && economic_status != 8
+    if age < 16 && economic_status != 9
       record.errors.add "ecstat#{person_num}", I18n.t("validations.household.ecstat.child_under_16", person_num:)
+      record.errors.add "age#{person_num}", I18n.t("validations.household.age.child_under_16", person_num:)
+    end
+    if economic_status == 9 && age > 16
+      record.errors.add "ecstat#{person_num}", I18n.t("validations.household.ecstat.child_over_16", person_num:)
+      record.errors.add "age#{person_num}", I18n.t("validations.household.age.child_over_16", person_num:)
```
dushan-madetech: I wonder if we could have some helper methods here to clarify what the economic status integers actually mean

dushan-madetech (approved):
Single comment but otherwise looks good to me 👍

app/models/validations/household_validations.rb
baarkerlounger: Right you are

PR #421, opened 2022-03-25
Cldc 657 full postcodes

kosiakkatrina:
Change column names:
`property_postcode` -> `postcode_full`
`previous_postcode` -> `ppostcode_full`

Normalise the postcode to uppercased format with no spaces. (tickets 657 and 666 indicate that full postcode == outcode + incode so removed the space, however in ticket 897 which talks about correctly formatting the postcode we do want the space in between. Not sure what the correct format is)


baarkerlounger:
Are these values (`postcode_full`, `ppostcode_full`) in the data dictionary?

kosiakkatrina:
> Are these values (`postcode_full`, `ppostcode_full`) in the data dictionary?

No, but they're in the ticket 🤷

baarkerlounger (approved):

PR #422, opened 2022-03-25
Update en.yml

Dan13L3:
@baarkerlounger Test - has this worked?

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -152,7 +152,7 @@ en:
       age:
         retired_male: "Male tenant who is retired must be 65 or over"
         retired_female: "Female tenant who is retired must be 60 or over"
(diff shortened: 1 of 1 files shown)


config/locales/en.yml
```diff
@@ -152,7 +152,7 @@ en:
       age:
         retired_male: "Male tenant who is retired must be 65 or over"
         retired_female: "Female tenant who is retired must be 60 or over"
-        retired_over_70: 'Answer cannot be over 70 as as tenant %{person_num} has economic status that is not "Retired"'
+        retired_over_70: 'Answer cannot be over 70 as the tenant's %{person_num} working situation is not 'retired''
```
baarkerlounger, line 155: Need to change the outside quote marks to `"` if we're using `'` in the string.

baarkerlounger:
Closed by https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/424

PR #423, opened 2022-03-25
Match relat data to data dictionary

baarkerlounger:


app/models/validations/household_validations.rb
```diff
@@ -188,7 +188,7 @@ def validate_person_age_and_gender_match_economic_status(record, person_num)
   end
 
   def validate_partner_count(record)
-    partner_count = (2..8).count { |n| record.public_send("relat#{n}")&.zero? }
+    partner_count = (2..8).count { |n| tenant_is_partner?(record.public_send("relat#{n}")) }
```
kosiakkatrina: `record["relat#{n}"]` would that not work?

baarkerlounger: Yep it would. Prefer that?

kosiakkatrina: Dunno, if it works the same with nil values then maybe?

db/schema.rb
```diff
@@ -224,16 +217,36 @@
...
+    t.boolean "confirmed"
+    t.string "old_id"
+    t.string "old_org_id"
+    t.datetime "created_at", null: false
+    t.datetime "updated_at", null: false
+    t.index ["data_protection_officer_id"], name: "dpo_user_id"
+    t.index ["organisation_id", "data_protection_officer_id", "confirmed"], name: "data_protection_confirmations_unique", unique: true
+    t.index ["organisation_id"], name: "index_data_protection_confirmations_on_organisation_id"
+  end
+
```
kosiakkatrina: What is this?

baarkerlounger: :see_no_evil:

kosiakkatrina (approved):

PR #425, opened 2022-03-25
CLDC-1054: Ask total household members, derive other household members

baarkerlounger:
https://digital.dclg.gov.uk/jira/browse/CLDC-1054

We ask data providers for the total household members, and display that in check answers, and we derive other household members for ADD.


dushan-madetech (approved):
Screenshot 2022-03-25 at 14 12 56

perfectly-balanced-thanos

Looks good to me 👍

PR #426, opened 2022-03-25
CLDC-667 newprop derived variable

dushan-madetech:
Updates the keys for rsnvac (first let reasons) to match the data dictionary
Adds the newprop derived variable and rules for deriving it

### DIFF
--- a/app/models/case_log.rb
+++ b/app/models/case_log.rb
@@ -351,6 +351,13 @@ private
       self.vmonth = property_void_date.month
       self.vyear = property_void_date.year
     end
+    if rsnvac.present?
+      self.newprop = if has_first_let_vacancy_reason?
+                       1
+                     else
+                       2
+                     end
+    end
     self.incref = 1 if net_income_refused?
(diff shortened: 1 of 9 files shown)


app/models/case_log.rb
```diff
@@ -351,6 +351,13 @@ def set_derived_fields!
       self.vmonth = property_void_date.month
       self.vyear = property_void_date.year
     end
+    if rsnvac.present?
+      self.newprop = if has_first_let_vacancy_reason?
+                       1
+                     else
+                       2
+                     end
+    end
```
kosiakkatrina: `has_first_let_vacancy_reason? ? 1 : 2` ?

kosiakkatrina (approved):

app/models/case_log.rb
dushan-madetech: 95-955352_italian-chef-kiss-emoji-freetoedit-italian-chef-kiss

Much cleaner than what I put

dushan-madetech: updated in c87955044cee45d366259a17b5de9f431a9683f5

PR #427, opened 2022-03-28
Cldc 489 referral validation

kosiakkatrina:
Add referral validation:
Referral cannot be 1 (internal transfer) if response to `Who is the landlord on the tenancy agreement?` is 1 (this landlord) and `Housing situation before this letting` is either 30 (Fixed term Local Authority General Needs tenancy) or 31 (Lifetime Local Authority General Needs tenancy)

Also only validate the charges if they're present

### DIFF
--- a/app/models/validations/financial_validations.rb
+++ b/app/models/validations/financial_validations.rb
@@ -122,7 +122,7 @@ private
     %i[scharge pscharge supcharg].each do |charge|
       maximum = CHARGE_MAXIMUMS.dig(charge, LANDLORD_VALUES[record.landlord], NEEDSTYPE_VALUES[record.needstype])
 
-      if maximum.present? && record[charge].present? && !weekly_value_in_range(record, charge, 0, maximum) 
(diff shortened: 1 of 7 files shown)


app/models/validations/financial_validations.rb
```diff
@@ -122,7 +122,7 @@ def validate_charges(record)
     %i[scharge pscharge supcharg].each do |charge|
       maximum = CHARGE_MAXIMUMS.dig(charge, LANDLORD_VALUES[record.landlord], NEEDSTYPE_VALUES[record.needstype])
 
-      if maximum.present? && !weekly_value_in_range(record, charge, 0, maximum)
+      if maximum.present? && !weekly_value_in_range(record, charge, 0, maximum) && record[charge].present?
```
baarkerlounger: Shouldn't the `.present?` check go before the range check?

dushan-madetech (approved):
Looks good to me 👍

PR #428, opened 2022-03-28
CLDC-771: Allow Data Coordinator to change details for users in their organisation

baarkerlounger:
For users in their organisation Data Coordinators can:

- Update name
- Update email
- Update role

They cannot update password (only the user themselves can do that). (Though they technically could update email and use that to reset password).

Data coordinator viewing users account:
image

Data coordinator editing users account:
image

Data coordinator viewing their own account:
image

Data coordinator editing their own account:
image

### DIFF
--- a/app/views/users/edit.html.erb
+++ b/app/views/users/edit.html.erb
@@ -1,4 +1,4 @@
-<% content_for :title, current_user == @user ? "Change your personal details" : "Change #{@user.name}'s personal details" %>
+<% content_for :title, current_user == @user ? "Change your personal details" : "Change #{@user.name}’s personal details" %>
 
 <% content_for :before_content do %>
   <%= govuk_back_link(
@@ -7,7 +7,7 @@
   ) %>
 <% end %>
 
-<%= form_for(current_user, as: :user, html: { method: :patch }) do |f| %>
+<%= form_for(@user, as: :user, html: { method: :patch }) do |f| %>
   <div class="govuk-grid-row">
     <div class="govuk-grid-column-two-thirds">
       <%= f.govuk_error_summary %>
@@ -25,6 +25,10 @@
         autocomplete: "email",
         spellcheck: "false"
       %>
+
+      <%= f.govuk_check_boxes_fieldset :is_dpo, multiple: false, legend: { text: "Are they a data protection officer?", size: "m" } do %>
+        <%= f.govuk_check_box :is_dpo, 1, 0, multiple: false, link_errors: true, label: { text: "Yes" } %>
+      <% end %>
 
       <%= f.govuk_submit "Save changes" %>
     </div>
--- a/app/views/users/edit.html.erb
+++ b/app/views/users/edit.html.erb
@@ -26,7 +26,7 @@
         spellcheck: "false"
       %>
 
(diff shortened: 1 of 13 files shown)


kosiakkatrina (approved):

app/views/users/edit.html.erb
```diff
@@ -26,6 +26,10 @@
         spellcheck: "false"
       %>
 
+      <%= f.govuk_check_boxes_fieldset :is_dpo, multiple: false, legend: { text: "Are #{@user == current_user ? "you" : "they"} a data protection officer?", size: "m" } do %>
+        <%= f.govuk_check_box :is_dpo, 1, 0, multiple: false, link_errors: true, label: { text: "Yes" } %>
+      <% end %>
+
```
paulrobertlloyd: Use Yes/No radios rather than a checkbox

app/views/users/new.html.erb
```diff
@@ -31,6 +31,10 @@
         f.govuk_collection_radio_buttons :role, roles, :id, :name, legend: { text: "Role", size: "m" }
       %>
 
+      <%= f.govuk_check_boxes_fieldset :is_dpo, legend: { text: "Are they a data protection officer?", size: "m" } do %>
+        <%= f.govuk_check_box :is_dpo, :is_dpo, label: { text: "Yes" } %>
+      <% end %>
+
```
paulrobertlloyd: Ditto

app/views/users/show.html.erb
```diff
@@ -13,33 +13,43 @@
...
         row.key { 'Role' }
-        row.value { current_user.role.humanize }
+        row.value { @user.role.humanize }
         row.action()
       end %>
+
+      <%= summary_list.row do |row|
+        row.key { 'Data protection officer' }
+        row.value { @user.is_data_protection_officer? ? "Yes" : "No" }
+        row.action(visually_hidden_text: 'are they a data protection officer?', href: edit_user_path, html_attributes: { 'data-qa': 'change-are-they-a-data-protection-officer' })
```
paulrobertlloyd: Given that the visible label is ‘Change’, probably makes sense for this to say:

```suggestion
        row.action(visually_hidden_text: 'if they a data protection officer', href: edit_user_path, html_attributes: { 'data-qa': 'change-are-they-a-data-protection-officer' })
```

paulrobertlloyd (changes requested):

app/views/users/show.html.erb
baarkerlounger: Change "whether" they are a data protection officer? @Dan13L3 what's your preference?

app/views/users/edit.html.erb
baarkerlounger: Even though the answer will hardly ever be yes?

baarkerlounger: Done

app/views/users/new.html.erb
baarkerlounger: Done

app/views/users/show.html.erb
Dan13L3: @baarkerlounger Can it be dynamic, so if someone's changing someone else's details, it's 'Are they a data protection officer?', and if it's their own, 'Are you a data protection officer?' (And then the label under personal details stays as 'Data protection officer'.)

PR #430, opened 2022-03-28
CLDC-1084: output tag support

kiddhustle:
https://digital.dclg.gov.uk/jira/browse/CLDC-1084


webpack.config.js
```diff
@@ -21,16 +21,28 @@ module.exports = {
...
+          '@sect/modernizr-loader','json-loader'
+        ],
+        type: 'javascript/auto'
+      },
       {
         test: /\.(js|ts)$/,
         include: [
           path.resolve(__dirname, 'node_modules/@hotwired/stimulus'),
           path.resolve(__dirname, 'node_modules/@stimulus/polyfills'),
+          path.resolve(__dirname, 'node_modules/@hotwired/turbo'),
```
baarkerlounger: Don't we need `turbo-rails` here too?

baarkerlounger:
Still throws `Uncaught TypeError: t.tagname is undefined` for me (on Firefox 98.0.2) when navigating between form pages and changing values. Think this comes from trying to transpile Turbo without using UMD modules. I think this is also what's causing the spec failures.

It's probably worth splitting this PR out into two, one for the polyfills and HTML5 shiv stuff that make the <output> tag work if/when Turbo isn't included, and a second that attempts get Turbo working (either with Modernizer or transpilation).

image


This branch removes Turbo altogether so you could work from that for the 1st step.

https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/406

webpack.config.js
kiddhustle: it doesn't give any error it doesn't look like it

kiddhustle:
> Still throws `Uncaught TypeError: t.tagname is undefined` for me (on Firefox 98.0.2) when navigating between form pages and changing values. Think this comes from trying to transpile Turbo without using UMD modules. I think this is also what's causing the spec failures.
> 
> It's probably worth splitting this PR out into two, one for the polyfills and HTML5 shiv stuff that make the tag work if/when Turbo isn't included, and a second that attempts get Turbo working (either with Modernizer or transpilation).
> 
> image
> 
> This branch removes Turbo altogether so you could work from that for the 1st step.
> 
> #406



> 

I do remember seeing this. I just up on this and it's because Firefox doesn't support outerHTML. t looks like I removed the polyfill which refactoring the application template.

FF issue should be gone now

baarkerlounger:
@kiddhustle still getting it on a fresh checkout

kiddhustle:
@baarkerlounger regarding the failing specs, how do I get to this page?

/logs/#{id}/property-major-repairs

baarkerlounger:
@kiddhustle you can load the test env (including fixture forms) with `RAILS_ENV=test rails s`

kiddhustle:
> > Still throws `Uncaught TypeError: t.tagname is undefined` for me (on Firefox 98.0.2) when navigating between form pages and changing values. Think this comes from trying to transpile Turbo without using UMD modules. I think this is also what's causing the spec failures.
> > It's probably worth splitting this PR out into two, one for the polyfills and HTML5 shiv stuff that make the tag work if/when Turbo isn't included, and a second that attempts get Turbo working (either with Modernizer or transpilation).
> > image
> > This branch removes Turbo altogether so you could work from that for the 1st step.
> > #406
> 
> > 
> 
> I do remember seeing this. I just up on this and it's because Firefox doesn't support outerHTML. t looks like I removed the polyfill which refactoring the application template.
> 
> FF issue should be gone now

Removing the Turbo include in the webpack config resolves this

baarkerlounger:
Seems like this works for modern browsers but in IE11 still breaks JS. I get:

image
image

kiddhustle:
> Seems like this works for modern browsers but in IE11 still breaks JS. I get:
> 
> image image

Aah I just realised. This is what the Webpack config Turbo include was put in place to fix:

`path.resolve(__dirname, 'node_modules/@hotwired/turbo'),`

I will check the docs

baarkerlounger:
I thought modernizr was supposed to stop the Turbo js code being loaded and so it wouldn't matter if it was transpiled or not?

kiddhustle:
> I thought modernizr was supposed to stop the Turbo js code being loaded and so it wouldn't matter if it was transpiled or not?

That's what it should do but for some reason Turbo is loading regardless. I did if Turbo was being loaded elsewhere in the app but it doesn't appear to be.

I have no idea why it is being loaded, when the Webpack Turbo include is in place it does not load

baarkerlounger:
@kiddhustle do the places we actually use turbo tags need wrapping in modernizr as well perhaps?

kiddhustle:
@baarkerlounger Where else is Turbo used other than `applications.js`?

I see that Stimulus is used in other places but not Turbo

baarkerlounger:
@kiddhustle can see here where `turbo_frame_tags` get used https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/406

kiddhustle:
@baarkerlounger but loading those tags shouldn't cause the module to be loaded should it? I will take a closer look.

I did wonder if maybe fetch was being polyfilled and causing the test to pass. But that doesn't appear to be the case either

baarkerlounger:
No it shouldn't... only thing I can think of off the top of my head though... slightly out of ideas.

kiddhustle:
@baarkerlounger I have no idea either probably best to just remove Turbo.

Apart from the forms it is used for the controllers too isn't it?

baarkerlounger:
Closing since we didn't get this to work and we've removed Turbo for the time being now

PR #433, opened 2022-03-29
CLDC-1084: IE11 output tag support [SEPARATED]

kiddhustle:
https://digital.dclg.gov.uk/jira/browse/CLDC-1084

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -26,9 +26,10 @@ module.exports = {
         include: [
           path.resolve(__dirname, 'node_modules/@hotwired/stimulus'),
           path.resolve(__dirname, 'node_modules/@stimulus/polyfills'),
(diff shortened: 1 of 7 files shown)


webpack.config.js
```diff
@@ -26,9 +26,10 @@ module.exports = {
         include: [
           path.resolve(__dirname, 'node_modules/@hotwired/stimulus'),
           path.resolve(__dirname, 'node_modules/@stimulus/polyfills'),
+          path.resolve(__dirname, 'node_modules/@hotwired/turbo'),
```
baarkerlounger: Think this needs removing? Think that's what's breaking specs also.

kiddhustle: yea I realised that too

baarkerlounger (approved):

PR #434, opened 2022-03-29
Cldc 1065 soft validations

kosiakkatrina:
Add soft rent range validations
Change `get_question` form method to prioritise returning questions that are routed to
If no is selected on the interruption screen, route back to the previous question that is routed to vs the previous question in the form

image

image


dushan-madetech:
@kosiakkatrina I think you're missing the informative text as per the screenshot on the ticket:
Screenshot 2022-02-07 at 10 50 19

app/models/form.rb
```diff
@@ -142,4 +144,11 @@ def is_last_question?(page, subsection, case_log)
     subsection_ids = subsections.map(&:id)
     subsection.id == subsection_ids[subsection_ids.length - 1] && next_page(page, case_log) == :check_answers
   end
+
+  def previous_page(page_ids, page_index, case_log)
```
baarkerlounger, line 148: Does this work for all cases? Could we use it for back button functionality in the form pages? (Rather than the current js history call)

Worth a spec as well I think?

baarkerlounger (commented):
Looks good!

kosiakkatrina:
> @kosiakkatrina I think you're missing the informative text as per the screenshot on the ticket:
> Screenshot 2022-02-07 at 10 50 19

@dushan-madetech yeah, that was semi intentional, because I was struggling with the logic for getting those 2  ranges as separate errors, but happy to have a look this morning

app/models/form.rb
kosiakkatrina: we only call it after the interruption screen but will see if it would make sense as the back button

kosiakkatrina: After looking at it I'd be keen to leave this out for now, because there's a lot of code duplication that would need to be refactored, how important do we think it is?

dushan-madetech (approved):
95-955352_italian-chef-kiss-emoji-freetoedit-italian-chef-kiss

If Paul is happy that it matches the design then and everything is in order I'd say 🚢 it

app/models/form.rb
baarkerlounger: Important (for accessibility) but definitely doesn't need to be in this PR.

baarkerlounger: Would still be good to have some tests for this method though

spec/models/form_spec.rb
```diff
@@ -8,10 +8,20 @@
 
   describe ".next_page" do
     let(:previous_page) { form.get_page("person_1_age") }
+    let(:value_check_previous_page) { form.get_page("net_income_value_check") }
 
     it "returns the next page given the previous" do
       expect(form.next_page(previous_page, case_log)).to eq("person_1_gender")
     end
+
+    it "returns the previuos page if the current page is a value check page with a `No` answer and the page is routed to" do
```
baarkerlounger: Typo on `previous`.

Would also be good to add scenarios for:

- Previous page is a value check page with "Yes" answer
- Previous page is not a value check page but is routed to
- Previous page is not a value check page and is not routed to

config/forms/2021_2022.json
```diff
@@ -4573,12 +4573,12 @@
                 }
               }
             },
-            "check_income_amount": {
+            "net_income_value_check": {
               "depends_on": [{ "net_income_soft_validation_triggered?": true }],
               "title_text": "Net income is outside the expected range based on the main tenant’s working situation",
```
paulrobertlloyd: We’re using ‘lead tenant’ throughout the service:

```suggestion
              "title_text": "Net income is outside the expected range based on the lead tenant’s working situation",
```

config/locales/en.yml
```diff
@@ -224,6 +224,11 @@ en:
         message: "Net income is lower than expected based on the main tenant’s working situation. Are you sure this is correct?"
       in_soft_max_range:
         message: "Net income is higher than expected based on the main tenant’s working situation. Are you sure this is correct?"
```
paulrobertlloyd:
```suggestion
        message: "Net income is higher than expected based on the lead tenant’s working situation. Are you sure this is correct?"
```

spec/fixtures/forms/2021_2022.json
```diff
@@ -562,7 +562,7 @@
                 "title_text": "Net income is outside the expected range based on the main tenant’s working situation",
```
paulrobertlloyd:
```suggestion
                "title_text": "Net income is outside the expected range based on the lead tenant’s working situation",
```

spec/models/form_spec.rb
kosiakkatrina: 🤦 
Previous page is a value check page with "Yes" answer - Yup!
Previous page is not a value check page but is routed to/Previous page is not a value check page and is not routed to - we wouldn't call previous_page in these scenarios tho. It would go down the route of normal next page  (?)

config/locales/en.yml
```diff
@@ -224,6 +224,11 @@ en:
         message: "Net income is lower than expected based on the main tenant’s working situation. Are you sure this is correct?"
```
paulrobertlloyd:
```suggestion
        message: "Net income is lower than expected based on the lead tenant’s working situation. Are you sure this is correct?"
```

spec/helpers/interuption_screen_helper_spec.rb
```diff
@@ -22,7 +22,7 @@
       it "returns correct informative text" do
         informative_text = {
           "translation" => "soft_validations.net_income.hint_text",
-          "argument" => %w[ecstat1 earnings],
+          "argument" => { "ecstat1": "question", "earnings": "question" },
         }
         expect(display_informative_text(informative_text, case_log))
           .to eq("<p>You told us the main tenant’s working situation is: <strong>Full-time – 30 hours or more</strong></p><p>The household income you have entered is <strong>£750.00 every week</strong></p>")
```
paulrobertlloyd:
```suggestion
          .to eq("<p>You told us the lead tenant’s working situation is: <strong>Full-time – 30 hours or more</strong></p><p>The household income you have entered is <strong>£750.00 every week</strong></p>")
```

spec/models/form_spec.rb
kosiakkatrina: although the first scenario we wouldn't call previous_page either, but I guess it checks that new if statement

paulrobertlloyd (changes requested):
Just one smol thing: ‘lead tenant’ not ‘main tenant’. Otherwise 👌

spec/models/form_spec.rb
baarkerlounger: I guess it's still useful in a sort of "defensive programming" sense? i.e. if the next person misunderstands when that method can/should be called and tries to use it for that scenario what should happen (perhaps it should raise an error if it considers that incorrect usage?).

Possibly veering into overkill territory here though so happy to merge as is.

kosiakkatrina: Added a few more

kosiakkatrina:
> Just one smol thing: ‘lead tenant’ not ‘main tenant’. Otherwise 👌

@paulrobertlloyd Yup, updated those now!

baarkerlounger (approved):

spec/models/form_spec.rb
baarkerlounger: chefkiss

app/models/case_log.rb
```diff
@@ -124,6 +124,10 @@ def net_income_soft_validation_triggered?
     net_income_in_soft_min_range? || net_income_in_soft_max_range?
   end
 
+  def rent_in_soft_validation_range?
```
baarkerlounger, line 127: Looks like these methods don't actually get called at all. Were they left in by mistake?

PR #436, opened 2022-03-29
CLDC-649: Export provider type from owning organisation

baarkerlounger:
https://digital.dclg.gov.uk/jira/browse/CLDC-649


spec/fixtures/exports/case_logs.xml
```diff
@@ -173,5 +173,6 @@
     <relat6/>
     <relat7/>
     <relat8/>
+    <providertype>LA</providertype>
```
kosiakkatrina: I think we'd want to export either 1 (LA) or 2 (PRP) 🤔

PR #438, opened 2022-03-31
CLDC-1014: Add Pagination for Logs Index

baarkerlounger:
https://digital.dclg.gov.uk/jira/browse/CLDC-1014

Shows 20 logs per page

image
image


paulrobertlloyd:
Looks good, though I wouldn’t show any pagination (or the ‘Showing 1 to X of X logs’ text) if there is only one page.

Also… the page title should be updated to include pagination details where relevant, for example: 

`Logs (page 2 of 10) – Submit social housing lettings and sales data (CORE) – GOV.UK`

app/views/case_logs/index.html.erb
```diff
@@ -1,14 +1,14 @@
-<% content_for :title, "Logs" %>
+<% title = @pagy.pages > 1 ? "Logs (page #{@pagy.page} of #{@pagy.pages})" : "Logs" %>
```
paulrobertlloyd: Probably fine for this PR, but DRYer if the title for any page using `@pagy` shows the page number and number of pages. Also ensures this accessibility feature isn’t missed on subsequent paged views.

paulrobertlloyd:
Seeing ‘← Previous‘ on page with 4 logs: 

Screenshot 2022-03-31 at 13 22 36

paulrobertlloyd:
Also, previous (or next) link shouldn’t show unless there is a previous or next page.

app/views/case_logs/index.html.erb
baarkerlounger: Done

baarkerlounger:
@paulrobertlloyd  done

app/frontend/styles/_pagination.scss
```diff
@@ -0,0 +1,134 @@
+// https://github.com/alphagov/govuk-frontend/blob/add-pagination-prototype/src/govuk/components/pagination/_index.scss
+.govuk-pagination {
```
paulrobertlloyd: One last thing (promise!). We should be using the `app-*` namespace for this component, as it’s not part of govuk-frontend (yet). Did you use these styles: https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data-prototype/blob/main/app/components/pagination/_pagination.scss

app/views/pagy/_nav.html.erb
```diff
@@ -0,0 +1,40 @@
+<% link = pagy_link_proc(pagy) -%>
+<% if pagy.pages > 1 %>
+  <nav class="govuk-pagination" id="pagination-label" aria-label="results navigation">
```
paulrobertlloyd: Update namespace in the markup too, obvs.

paulrobertlloyd (changes requested):

app/frontend/styles/_pagination.scss
baarkerlounger: Yep, copied and pasted from there just changed `app-` to `govuk-` to match the markup but will do it the other way around then.

paulrobertlloyd: https://design-system.service.gov.uk/get-started/extending-and-modifying-components/#use-a-unique-prefix-for-component-names

app/views/pagy/_nav.html.erb
baarkerlounger: Done

paulrobertlloyd (approved):
🚢

PR #439, opened 2022-03-31
CLDC-1086: Change answer page routing

kosiakkatrina:
Route back to the check answers if answer is changed from check answers page

### DIFF
--- a/spec/features/form/check_answers_page_spec.rb
+++ b/spec/features/form/check_answers_page_spec.rb
@@ -127,7 +127,7 @@ RSpec.describe "Form Check Answers Page" do
       end
     end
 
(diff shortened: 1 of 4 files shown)


spec/features/form/check_answers_page_spec.rb
```diff
@@ -127,6 +127,18 @@
       end
     end
 
+    context "when the user changes their answer" do
```
baarkerlounger: Can we make clear that they're changing their answer "from the check answers page" not just in general (for example by clicking back button)?

baarkerlounger (approved):

baarkerlounger (approved):

PR #441, opened 2022-03-31
CSV download

baarkerlounger:
Currently this download all logs a user has permission to see (since we don't have filters yet). i.e. the Download does not consider pagination.

The CSV includes answer value labels rather than values (i.e. "Yes" rather than 1)

Link:
image


app/views/case_logs/_log_list.html.erb
```diff
@@ -1,6 +1,9 @@
 <figure class="app-figure">
   <figcaption id="<%= title.dasherize %>" class="app-figure__caption">
-    <strong><%= pagy.count %></strong> total <%= title.downcase %>
+    <span class="govuk-!-margin-right-4">
+      <strong><%= pagy.count %></strong> total <%= title.downcase %>
+    </span>
+    <a class="govuk-link" download href="/logs.csv">Download (CSV)</a>
```
paulrobertlloyd: Missed this in the prototype, but if I were to be super pedantic…

```suggestion
    <a class="govuk-link" download href="/logs.csv" type="text/csv">Download (CSV)</a>
```

paulrobertlloyd (approved):
I’m not sure the downloaded CSV should respect pagination (only search/filtering).

app/views/case_logs/_log_list.html.erb
paulrobertlloyd: We can also use the `download` attribute to give the downloaded CSV a filename… not sure what a useful name would be, though.

baarkerlounger:
@paulrobertlloyd that was my assumption as well, was just being explicit about it in the description

app/views/case_logs/_log_list.html.erb
baarkerlounger: Filename is set in the controller currently

paulrobertlloyd: That’s probably best; `download` isn’t supported by IE11

PR #442, opened 2022-04-01
CLDC-1117: Allow data providers to see but not edit users in their org

baarkerlounger:
https://digital.dclg.gov.uk/jira/browse/CLDC-1117

Data provider viewing their org details:
image

Data provider viewing their org users:
image

Data provider viewing other user in same org:
image


dushan-madetech (approved):
Looks good to me 👍

PR #443, opened 2022-04-01
Remove unused methods

baarkerlounger:
Remove methods that aren't called anywhere.


kosiakkatrina (approved):
Ah, very good! Nice spot!

PR #447, opened 2022-04-04
CLDC-1074: Derive landlord value from owning organisation provider type

baarkerlounger:


spec/models/case_log_spec.rb
```diff
@@ -332,17 +350,17 @@
           end
 
           it "correctly derives and saves weekly service charge" do
-            case_log.update!(scharge: 100, period: 2)
+            case_log.update!(scharge: 70, period: 2)
```
dushan-madetech, line 353: Why did you have to change the value of the floats in the weekly charges tests?

dushan-madetech (approved):
I had a single question but happy to leave an approval for this since all the tests are passing

spec/models/case_log_spec.rb
baarkerlounger: Because the original ones were too high for the organisation type but that wasn't being set so the validation wasn't triggering.  Now it is being set, since it's automatically derived from the organisation so needs to be valid.

Calculation being tested should be working the same though.

PR #448, opened 2022-04-05
CLDC-1091: User Account URLs

baarkerlounger:
Changes the URL for your own account paths from:

`/users/2` => `/account`
`/users/2/edit` => `/account/edit`
`/users/2/password/edit` => `/account/edit/password`

Viewing other user's account still has paths:
`/users/3`
`/users3/edit` 

Devise expects the password reset path for a not signed in user to be at `.../password/edit` so to avoid the clash we point editing password for a signed in user at `.../edit/password`.

### DIFF
--- a/app/views/users/show.html.erb
+++ b/app/views/users/show.html.erb
@@ -15,7 +15,7 @@
         row.key { 'Name' }
         row.value { @user.name }
         if current_user == @user || current_user.data_coordinator?
-          row.action(visually_hidden_text: 'name', href: edit_user_path, html_attributes: { 'data-qa': 'change-name' })
+          row.action(visually_hidden_text: 'name', href: aliased_user_edit(@user, current_user), html_attributes: { 'data-qa': 'change-name' })
         else
           row.action()
         end
@@ -25,7 +25,7 @@
         row.key { 'Email address' }
         row.value { @user.email }
         if current_user == @user || current_user.data_coordinator?
-          row.action(visually_hidden_text: 'email address', href: edit_user_path, html_attributes: { 'data-qa': 'change-email' })
+          row.action(visually_hidden_text: 'email address', href: aliased_user_edit(@user, current_user), html_attributes: { 'data-qa': 'change-email' })
         else
           row.action()
         end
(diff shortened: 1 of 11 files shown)


app/views/users/show.html.erb
```diff
@@ -71,7 +71,7 @@
         row.key { 'Key contact' }
         row.value { @user.is_key_contact? ? "Yes" : "No" }
         if current_user.data_coordinator?
-          row.action(visually_hidden_text: "are #{current_user == @user ? "you" : "they"} a key contact?", href: edit_user_path, html_attributes: { "data-qa": "change-are-#{current_user == @user ? "you" : "they"}-a-key-contact" })
+          row.action(visually_hidden_text: "are #{pronoun(@user, current_user)} a key contact?", href: aliased_user_edit(@user, current_user), html_attributes: { "data-qa": "change-are-#{pronoun(@user, current_user)}-a-key-contact" })
         else
           row.action()
         end
```
dushan-madetech: It feel like there's a lot of logic living in this view file

spec/controllers/users_controller_spec.rb
```diff
@@ -0,0 +1,18 @@
...
+  let(:params) { { id: other_user.id } }
+  let(:user) { FactoryBot.create(:user, :data_coordinator) }
+  let(:other_user) { FactoryBot.create(:user, organisation: user.organisation) }
+
+  before do
+    sign_in user
+  end
+
+  describe "GET #edit_password" do
+    it "returns not found" do
```
dushan-madetech: This text could be a little more descriptive to make it clear that it's testing you can't edit the password of another user - it's not immediately obvious without reading into the test code itself

dushan-madetech (approved):
Couple of comments but nothing worth holding this back over 👍

app/views/users/show.html.erb
dushan-madetech: I wonder if any of it could be extracted to the user helper? Just pondering here but not a strong opinion

baarkerlounger: Tried extracting out the row action creation but I think that gets messier because then you're passing the row element into the helper method which leads to lots of mocking/stubbing for testing. Or you create the whole summary list row in the helper but then you're maybe just moving all the logic from one file to another? Didn't find a great answer...

PR #449, opened 2022-04-05
Cldc 1010 filter

kosiakkatrina:
Add the side filter panel to the case logs list page:
image

image

image


app/controllers/case_logs_controller.rb
```diff
@@ -76,6 +76,11 @@ def destroy
     end
   end
 
+  def filter
+    cookies[:case_logs_filters] = { status: params[:status] }.to_json
```
baarkerlounger: I think we should use `session[:case_logs_filters]` rather than `cookies[:case_logs_filters]`. I wouldn't expect filter choices to persist beyond your current session?

app/views/case_logs/_log_filters.erb
```diff
@@ -0,0 +1,32 @@
...
+            data-action="click->filter#toggleFilter">
+            Close
+            </button>
+        </div>
+        <div class="app-filter__content">
+            <div class="govuk-form-group app-filter__group">
+                <%= form_with url: "/logs/filter", html: { method: :post }, builder: GOVUKDesignSystemFormBuilder::FormBuilder do |f| %>
+                    <%= f.govuk_check_boxes_fieldset :status, legend: { text: "Status", size: "s"} do %>
+                        <div class="govuk-checkboxes govuk-checkboxes--small" data-module="govuk-checkboxes">
+                            <% statuses = {in_progress: "In progress", submitted: "Completed", not_started: "Not started"} %>
```
baarkerlounger: Could we use a helper to derive these from `CaseLog::STATUS`? Such that we only have one source of truth for what valid statuses are?

app/views/case_logs/_log_filters.erb
```diff
@@ -0,0 +1,32 @@
...
+            class="app-filter__close" 
+            type="button" 
+            data-controller="filter"
+            data-action="click->filter#toggleFilter">
+            Close
+            </button>
+        </div>
+        <div class="app-filter__content">
+            <div class="govuk-form-group app-filter__group">
+                <%= form_with url: "/logs/filter", html: { method: :post }, builder: GOVUKDesignSystemFormBuilder::FormBuilder do |f| %>
```
baarkerlounger: Don't think you need to specify the form builder here since the govuk one is set as the default anyway.
https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/blob/main/config/initializers/default_form_builder.rb

db/schema.rb
```diff
@@ -316,9 +316,9 @@
     t.integer "failed_attempts", default: 0
     t.string "unlock_token"
     t.datetime "locked_at", precision: nil
+    t.string "phone"
```
baarkerlounger: Let's remove this since there's no migration here

config/routes.rb
```diff
@@ -83,6 +83,8 @@
     end
   end
 
+  post "/logs/filter", to: "case_logs#filter"
```
baarkerlounger: Does this need it's own route? Could it not just be a `GET index` request with extra URL params? That way you can bookmark or share a filter too?

db/schema.rb
kosiakkatrina: Hm, this happens every time I migrate, I assume that it does to other people too? Wouldn't it be easier to commit this instead of unstaging this file every time we do PRs? Or do you mean just not in this PR?

config/routes.rb
kosiakkatrina: Yeah, that makes sense. Lemme shift stuff around

db/schema.rb
baarkerlounger: I think this happens every time because migrations have been run out of order on your local (i.e. you've run a migration while working on a branch, not rolled it back, and then run one that was merged into main before merging your PR).

If you dropped your DB and ran migrations on a fresh one I don't think you'd get this (if you did we should commit it)

app/controllers/case_logs_controller.rb
```diff
@@ -7,7 +7,9 @@ class CaseLogsController < ApplicationController
   before_action :find_resource, except: %i[create index edit]
 
   def index
-    @pagy, @case_logs = pagy(current_user.case_logs)
+    set_session_filters if params[:status].present?
```
baarkerlounger, line 10: If filters are just url params do we actually need to store them as sessions at all?

kosiakkatrina: I thought if someone visits `/logs<filter_here>` and then goes to `/logs`, we'd still want to keep the filters?

db/schema.rb
kosiakkatrina: Hm, so after running `db:drop` and `db:reset` my schema looks the way it does here. I'll commit this!

app/controllers/case_logs_controller.rb
baarkerlounger: Interesting, I'm not sure. I think my expectation would have been that if I press back they're kept but not if I do a fresh page load. @roobottom do you have thoughts?

spec/requests/case_logs_controller_spec.rb
```diff
@@ -209,6 +209,45 @@
...
+                              managing_organisation: organisation)
+          end
+          let!(:completed_case_log) do
+            FactoryBot.create(:case_log, :completed,
+                              owning_organisation: organisation,
+                              managing_organisation: organisation)
+          end
+
+          it "only shows case logs for selected status" do
+            in_progress_case_row_log = "<a class=\"govuk-link\" href=\"/logs/#{in_progress_case_log.id}\">#{in_progress_case_log.id}</a>"
```
baarkerlounger: I know this is proliferating a style I started but in general it's probably better to use the Capybara methods where possible (`has_selector?`, `has_css?`, `has_link?`, `has_xpath?` etc) than matching escaped html strings like this. 

(Not a blocker for this PR)

kosiakkatrina: Hm, would we want to do that for request specs as well as feature?

baarkerlounger: Ideally probably where possible. Not super important to change existing ones but I think it is generally more readable and less brittle?

baarkerlounger: You can just add `let(:page) { Capybara::Node::Simple.new(response.body) }`, you don't need to actually run the test with the Capybara drivers (the slow bit).

baarkerlounger (approved):

app/controllers/case_logs_controller.rb
roobottom: So the use case is:

1. Set filters
2. Find the <thing> I'm looking for
3. click on it.
4. Come back to the list
5. expect my filters still to be enabled?

I'm not sure, I think it's worth a conversation with the wider group about what the actual use case is.

baarkerlounger: That's the flow yeah. I think my assumption would be that if I've done my thing and then clicked the Logs link again, I'd expect to be starting a new search (or create a new log).

In terms of taking it to the wider group are you comfortable making a decision to be this PR merged and then we revisit if needed in a future ticket?

kosiakkatrina: I can make that decision, it works and can always be adjusted upon user feedback 🤷‍♀️

baarkerlounger:
Needs a rebase first but I think the filters should also apply to the CSV download?

kosiakkatrina:
> Needs a rebase first but I think the filters should also apply to the CSV download?

@baarkerlounger Yes, different PR tho, cause this has too much in it

PR #450, opened 2022-04-05
CLDC-1118: Implement empty master manifest

smeny:


db/schema.rb
```diff
@@ -313,12 +318,12 @@
     t.string "last_sign_in_ip"
     t.integer "role"
     t.string "old_user_id"
+    t.string "phone"
```
baarkerlounger: Don't think the phone field order swapping should be here

spec/services/exports/case_log_export_service_spec.rb
```diff
@@ -2,38 +2,77 @@
 
 RSpec.describe Exports::CaseLogExportService do
   let(:storage_service) { instance_double(StorageService) }
+
   let(:export_filepath) { "spec/fixtures/exports/case_logs.xml" }
   let(:export_file) { File.open(export_filepath, "r:UTF-8") }
-  let(:expected_filename) { "core_2022_02_08/dat_core_2022_02_08_0001.xml" }
+
+  let(:expected_data_filename) { "core_2022_02_08/dat_core_2022_02_08_0001.xml" }
+  let(:expected_master_manifest_filename) { "Manifest_2022_02_08_0001.csv" }
```
baarkerlounger, line 10: The manifest is a csv and not an xml? :joy:

spec/services/exports/case_log_export_service_spec.rb
```diff
@@ -2,38 +2,77 @@
...
+        export_service.export_case_logs
+      end
+
+      it "generates an XML export file with the expected content" do
+        actual_content = nil
+        expected_content = replace_entity_ids(export_file.read)
+        allow(storage_service).to receive(:write_file).with(expected_data_filename, any_args) { |_, arg2| actual_content = arg2&.string }
+
+        export_service.export_case_logs
+        expect(actual_content).to eq(expected_content)
```
baarkerlounger, line 63: Wonder if this would be easier to test if we split the StringIO bit into a final step. So you can have a unit test just expect against the string value and then a second unit test check that string is passed into StringIO and written to StorageService correctly.

Not important for now though.

baarkerlounger (approved):
:+1: . Please just remove the unnecessary schema change before merging.

db/schema.rb
smeny: That was done automatically after running migration scripts, I don't think it would have any impact but would you suggest removing the change?

spec/services/exports/case_log_export_service_spec.rb
smeny: yup :)

spec/services/exports/case_log_export_service_spec.rb
smeny: I think that's what we are doing now, we have two tests, one for the filename and one for the stringio content.

db/schema.rb
baarkerlounger: Yep. As a general rule of thumb you only want to be committing schema changes if your PR has a migration that is expected to introduce them or if you've run migrations on a fresh database and got them (i.e. main schema is already incorrect/someone has forgotten to commit).

Change like this is likely just introduced on your local from running migrations in a slightly different order (not rolling back when working on a PR etc) so you just get back and forth on each PR.

PR #453, opened 2022-04-06
Update soft rent validations messages

kosiakkatrina:
Update soft validations messages to display calculated expected value for the period given:
image

image

### DIFF
--- a/app/models/case_log.rb
+++ b/app/models/case_log.rb
@@ -42,7 +42,7 @@ class CaseLog < ApplicationRecord
   HAS_BENEFITS_OPTIONS = [1, 6, 8, 7].freeze
   STATUS = { "not_started" => 0, "in_progress" => 1, "completed" => 2 }.freeze
   NUM_OF_WEEKS_FROM_PERIOD = { 2 => 26, 3 => 13, 4 => 12, 5 => 50, 6 => 49, 7 => 48, 8 => 47, 9 => 46, 1 => 52 }.freeze
-  SUFFIX_FROM_PERIOD = { 2 => "every 2 weeks", 3 => "every 4 weeks", 4 => "every calendar month" }.freeze
+  SUFFIX_FROM_PERIOD = { 2 => "every 2 weeks", 3 => "every 4 weeks", 4 => "every month" }.freeze
   enum status: STATUS
 
   def form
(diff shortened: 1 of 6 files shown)


app/models/case_log.rb
```diff
@@ -289,6 +290,16 @@ def self.to_csv
     end
   end
 
+  def soft_min_for_period
```
baarkerlounger, line 293: This might be worth monitoring. Potentially adds a lot of DB lookups although I think rails is smart enough to read from cache...

kosiakkatrina: Oooh, that is very true! We call routed_to so often 🤔 Ignore this! It wouldn't get called on routed to 🤦‍♀️

app/models/case_log.rb
```diff
@@ -289,6 +290,16 @@ def self.to_csv
...
+    soft_min = LaRentRange.find_by(start_year: collection_start_year, la:, beds:, lettype:).soft_min
+    "#{soft_value_for_period(soft_min)} #{SUFFIX_FROM_PERIOD[period].presence || 'every week'}"
+  end
+
+  def soft_max_for_period
+    soft_max = LaRentRange.find_by(start_year: collection_start_year, la:, beds:, lettype:).soft_max
+    "#{soft_value_for_period(soft_max)} #{SUFFIX_FROM_PERIOD[period].presence || 'every week'}"
```
baarkerlounger, line 300: Does `soft_value_for_period` return anything if no period is set? Does this need to? Or could it skip the whole thing?

kosiakkatrina: I think it returns "", maybe worth extracting the check for period before the db call to minimise the # of calls

kosiakkatrina: But we have this method `def rent_in_soft_min_range?
    return unless brent && weekly_value(brent) && startdate <...>` which is the conditional for checking the routing. And it returns unless those fields are filled in. We're only calling  `soft_max_for_period` when we load the screen I think, to display the value only

baarkerlounger: Does that mean that by the time we call this we _know_ `period` is set and don't need the `.presence` check here?

kosiakkatrina: Yes, we know it's there, but if you mean the presence check here, it's just for getting the suffix. Cause it would be `every week` in most of the cases but it doesn't make sense to write it out into the `SUFFIX_FROM_PERIOD`  constant for every period. I struggled to set a default hash value in a concise way cause we need to freeze our constants and this seemed like the shortest way to do it. So it's not checking for the presence of the period, but rather for the presence of value corresponding to it in the constant

baarkerlounger: Got it

baarkerlounger (approved):

PR #455, opened 2022-04-07
Devise fields don't need versioning

baarkerlounger:
Let's not totally explode our database storage requirements by creating a new paper trail version every time someone logs in or mistypes a password.

### DIFF
--- a/app/models/admin_user.rb
+++ b/app/models/admin_user.rb
@@ -6,7 +6,18 @@ class AdminUser < ApplicationRecord
 
   has_one_time_password(encrypted: true)
 
-  has_paper_trail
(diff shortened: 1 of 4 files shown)


kosiakkatrina (approved):
Happy for this to go in! Any reason why we'd want to collect that? 🤔 Or should it be more of a monitoring thing?

baarkerlounger:
@kosiakkatrina we do store those fields in the User and AdminUser table. This just means we don't *additionally* create a new record in the versions table every time one of these changes since there's not much value in that. I think if there are monitoring reasons we want it elsewhere there would be better ways of doing that.

PR #456, opened 2022-04-07
Add local housing authority referral validation

kosiakkatrina:
Add missing validation for referral

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -192,7 +192,7 @@ en:
         la_general_needs:
           internal_transfer: "Answer cannot be internal transfer as you already told us it's the same landlord on the tenancy agreement and the household had either a fixed-term or lifetime local authority general needs tenancy immediately before this letting"
         prp:
-          local_housing_referral: "Answer cannot be Nominated by a local housing authority as you already told us it's another landlord on the tenancy agreement"
+          local_housing_referral: "Answer cannot be 'nominated by a local housing authority' as you already told us it's another landlord on the tenancy agreement"
       homeless:
         assessed:
           internal_transfer: "Answer cannot be assessed as homeless as you already told us this tenancy is an internal transfer"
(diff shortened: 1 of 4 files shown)


config/locales/en.yml
```diff
@@ -189,6 +189,8 @@ en:
         reason_permanently_decanted: "Answer must be internal transfer as you already told us the tenant was permanently decanted from another property owned by this landlord"
         la_general_needs:
           internal_transfer: "Answer cannot be internal transfer as you already told us it's the same landlord on the tenancy agreement and the household had either a fixed-term or lifetime local authority general needs tenancy immediately before this letting"
+        prp:
+          local_housing_referral: "Answer cannot be Nominated by a local housing authority as you already told us it's another landlord on the tenancy agreement"
```
baarkerlounger: Think @Dan13L3 will probably want:
```
"Answer cannot be 'Nominated by a local housing authority' as you already told us it's another landlord on the tenancy agreement"
```

spec/models/validations/household_validations_spec.rb
```diff
@@ -215,6 +215,24 @@
           .to include(match(I18n.t("validations.household.prevten.la_general_needs.internal_transfer")))
       end
     end
+
+    context "when referral is nominated by a local housing authority" do
+      it "cannot have `other landlord`" do
```
baarkerlounger, line 220: Any reason to use backticks over `'` here?

baarkerlounger (approved):

spec/models/validations/household_validations_spec.rb
kosiakkatrina: No, just a habit 👀 Also, looks a bit more distinguishable within the sentence, cause otherwise we'd end up with ''' at the end. I'm not married to the idea, how important is it to you?

baarkerlounger: Not very. I was just curious. I'm also inclined to agree it stands out better visually.

config/locales/en.yml
Dan13L3: @baarkerlounger Yes, please - and to be extra picky, we should keep it lower case to be consistent: 'nominated by a local housing authority'

baarkerlounger (approved):

PR #459, opened 2022-04-08
Reset errors if a log gets successfully updated

kosiakkatrina:


spec/requests/form_controller_spec.rb
```diff
@@ -120,16 +120,35 @@
...
           let(:answer) { 2000 }
+          let(:valid_answer) { 20 }
 
           it "re-renders the same page with errors if validation fails" do
             expect(response).to redirect_to("/logs/#{case_log.id}/#{page_id.dasherize}")
+            follow_redirect!
+            expect(page).to have_content("There is a problem")
+          end
+
+          it "resets errors when fixed" do
```
baarkerlounger, line 148: Was this actually failing before the fix?

kosiakkatrina: Yes!

baarkerlounger (approved):

spec/requests/form_controller_spec.rb
baarkerlounger: Had to try for myself :joy: 

Great stuff!

PR #460, opened 2022-04-08
Update devise.en.yml

Dan13L3:
@baarkerlounger Reviewed these all in one go - let me know if anything looks odd

PR #461, opened 2022-04-08
Update devise translation strings

baarkerlounger:
@Dan13L3 can you please review this one? Tried to copy across your changes from https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/460 with the inserted fields accounted for. My logic for a full stop at the end was "yes" if there is already one in the sentence, "no" otherwise. Also worth noting a lot of these messages (around self sign up and unlocking for example) aren't actually used by the service currently since we don't let users do that, but good to have them updated in case we ever do.


Dan13L3:
thanks @baarkerlounger - all looks good to me. Full stops make sense (I'd somehow skipped the ones in the middle...)

PR #463, opened 2022-04-11
Allow support to edit users

baarkerlounger:
- [x] Support user can edit existing users from any organisation
- [x] Support user can assign "support" role to any user
- [x] Support user can invite user to own organisation (including with "support" role)
- [x] Data coordinator cannot assign "support" role to users
- [x] Support cannot change other user passwords 
- [x] No one can assign an invalid role

### DIFF
--- a/spec/requests/users_controller_spec.rb
+++ b/spec/requests/users_controller_spec.rb
@@ -642,9 +642,424 @@ RSpec.describe UsersController, type: :request do
         expect(response).to redirect_to("/organisations/#{user.organisation.id}/users")
       end
 
+      context "when the email is already taken" do
+        before do
+          FactoryBot.create(:user, email: "new_user@example.com")
+        end
+
+        it "shows an error" do
+          request
+          expect(response).to have_http_status(:unprocessable_entity)
+          expect(page).to have_content(I18n.t("validations.email.taken"))
+        end
+      end
+
+      context "when trying to assign support role" do
+        let(:params) do
+          {
+            "user": {
+              name: "new user",
+              email: "new_user@example.com",
+              role: "support",
+            },
+          }
+        end
+
+        it "shows an error" do
+          request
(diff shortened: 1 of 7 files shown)


spec/requests/users_controller_spec.rb
```diff
@@ -642,10 +642,425 @@
...
+          end
+
+          it "shows an error if passwords don't match" do
+            expect(response).to have_http_status(:unprocessable_entity)
+            expect(page).to have_selector("#error-summary-title")
+          end
+        end
+      end
+
+      context "when the current user does not matches the user ID" do
```
dushan-madetech:
```suggestion
      context "when the current user does not match the user ID" do
```

spec/requests/users_controller_spec.rb
```diff
@@ -642,10 +642,425 @@
...
+          end
+
+          it "does not bypass sign in for the support user" do
+            patch "/users/#{other_user.id}", headers: headers, params: params
+            follow_redirect!
+            expect(page).to have_content("#{other_user.reload.name}’s account")
+            expect(page).to have_content(other_user.reload.email.to_s)
+          end
+
+          context "when we try to update the user password" do
```
dushan-madetech: Who is the we referring to here?

spec/requests/users_controller_spec.rb
```diff
@@ -642,10 +642,425 @@
...
+            end
+
+            it "does update other values" do
+              expect { patch "/users/#{other_user.id}", headers: headers, params: params }
+                .to change { other_user.reload.name }.from("Danny Rojas").to("new name")
+            end
+          end
+        end
+
+        context "when the current user does not matches the user ID" do
```
dushan-madetech:
```suggestion
        context "when the current user does not match the user ID" do
```

dushan-madetech (approved):

PR #464, opened 2022-04-11
Add year filter

kosiakkatrina:
image


spec/requests/case_logs_controller_spec.rb
```diff
@@ -174,37 +174,67 @@
...
-            FactoryBot.create(:case_log, :in_progress,
-                              owning_organisation: organisation,
-                              managing_organisation: organisation)
-          end
-          let!(:completed_case_log) do
-            FactoryBot.create(:case_log, :completed,
-                              owning_organisation: organisation,
-                              managing_organisation: organisation)
-          end
+          context "with status fileter" do
```
baarkerlounger: Spelling on `filter`

app/views/case_logs/_log_filters.erb
```diff
@@ -13,13 +13,24 @@
         <div class="app-filter__content">
             <div class="govuk-form-group app-filter__group">
                 <%= form_with url: "/logs", html: { method: :get } do |f| %>
+                    <%= f.govuk_check_boxes_fieldset :year, legend: { text: "Collection year", size: "s"} do %>
```
baarkerlounger: Worth splitting these out into partials?

spec/requests/case_logs_controller_spec.rb
```diff
@@ -174,37 +174,67 @@
...
+            it "shows case logs for multiple selected statuses" do
+              get "/logs?year[]=2021&year[]=2022", headers: headers, params: {}
+              expect(page).to have_link(case_log_2021.id.to_s)
+              expect(page).to have_link(case_log_2022.id.to_s)
+            end
 
-            get "/logs", headers: headers, params: {}
-            expect(page).to have_link(in_progress_case_log.id.to_s)
-            expect(page).not_to have_link(completed_case_log.id.to_s)
+            it "shows case logs for one selected status" do
```
baarkerlounger: Worth a spec example for combining the two filters?

i.e. a status and a year one in the same query?

app/controllers/case_logs_controller.rb
```diff
@@ -122,13 +122,22 @@ def find_resource
...
-    status_filter = JSON.parse(session[:case_logs_filters])["status"] if session[:case_logs_filters].present?
-    return user_case_logs unless status_filter
+    if session[:case_logs_filters].present?
+      filters = JSON.parse(session[:case_logs_filters])
 
-    user_case_logs.filter_by_status(status_filter)
+      %w[status year].each do |category|
+        if filters[category]
+          filtered_by_category = filters[category].select(&:present?).map { |filter| user_case_logs.public_send("filter_by_#{category}", filter) }.flatten
+          user_case_logs = CaseLog.where(id: filtered_by_category.map(&:id))
```
baarkerlounger: Does this work for combined filters? Aren't you just setting the returned collection to whichever filter category was last run?

Also I think you want to build up the scopes (they're deferred and composable) so that you ultimately just run a single DB query to get the result rather than running them sequentially.

app/models/case_log.rb
```diff
@@ -35,6 +35,7 @@ class CaseLog < ApplicationRecord
 
   scope :for_organisation, ->(org) { where(owning_organisation: org).or(where(managing_organisation: org)) }
   scope :filter_by_status, ->(status) { where status: status }
+  scope :filter_by_year, ->(year) { where(startdate: Time.zone.local(year, 4, 1)...Time.zone.local(year.to_i + 1, 4, 1)) }
```
baarkerlounger: Think you need the `year.to_i` on the start date selection as well since I'm assuming it comes in as a string also? Should that be causing a spec to fail?

baarkerlounger: Also should the end date be `Time.zone.local(year.to_i + 1, 3, 31).end_of_day`?

app/views/case_logs/_log_filters.erb
```diff
@@ -13,17 +13,9 @@
...
-                            <% statuses = status_filters %>
-                            <% statuses.map do |key, option| %>
-                                <%= f.govuk_check_box "status", "#{key}",
-                                label: { text: option },
-                                checked: filter_selected?(key),
-                                size: "s" %>
-                            <% end %>
-                        </div>
-                    <% end %>
+                    <% years = {"2021": "2021/22", "2022": "2022/23"} %>
```
baarkerlounger, line 16: Ideally probably wouldn't hardcode these here but rather derive them, either from the (active?) forms we have or from the data but happy for this to go in for now.

app/controllers/case_logs_controller.rb
```diff
@@ -121,14 +121,21 @@ def find_resource
...
+        query = query.public_send("filter_by_#{category}", values)
+      end
+    end
+    query.all.includes(:owning_organisation, :managing_organisation)
   end
 
   def set_session_filters
-    session[:case_logs_filters] = { status: params[:status] }.to_json
+    new_filters = session[:case_logs_filters].present? ? JSON.parse(session[:case_logs_filters]) : {}
+    %i[status years].each { |filter| new_filters[filter] = params[filter] if params[filter].present? }
```
baarkerlounger, line 138: Ideally wouldn't hardcode these either but again not important for now

baarkerlounger (approved):

PR #467, opened 2022-04-12
Add user filter

kosiakkatrina:


app/controllers/case_logs_controller.rb
```diff
@@ -127,15 +127,15 @@ def filtered_case_logs
       filters.each do |category, values|
         next if values.reject(&:empty?).blank?
 
-        query = query.public_send("filter_by_#{category}", values)
+        query = query.public_send("filter_by_#{category}", values, current_user)
```
baarkerlounger, line 130: Can we make `current_user/all` the `value` here so we don't need to change the method signature?

app/models/case_log.rb
```diff
@@ -34,15 +34,21 @@ class CaseLog < ApplicationRecord
...
+  scope :filter_by_years, lambda { |years, user = nil|
     first_year = years.shift
     query = filter_by_year(first_year)
     years.each { |year| query = query.or(filter_by_year(year)) }
     query.all
   }
   scope :filter_by_year, ->(year) { where(startdate: Time.utc(year.to_i, 4, 1)...Time.utc(year.to_i + 1, 4, 1)) }
 
+  scope :filter_by_user, lambda { |selected_user, user|
+                           if !selected_user.include?("all") && user.present?
```
baarkerlounger, line 47: then this could just be:
```
where(id: PaperTrail::Version.where(item_type: "CaseLog", event: "create", whodunnit: user.to_global_id.uri.to_s).map(&:item_id)) unless user == "all"
```

app/controllers/case_logs_controller.rb
kosiakkatrina: Uhm, would that mean putting an if statement like this or smth:
```
if category == 'user'
    values == values.include?(''all') ? "all" : current_user
end 
```

Looks wordy unless I'm misunderstanding something, besides default values in scopes sort of combat any massive issues with having 2 arguments in the signature 🤷‍♀️

app/views/case_logs/_log_filters.erb
```diff
@@ -16,6 +16,7 @@
                     <% years = {"2021": "2021/22", "2022": "2022/23"} %>
                     <%= render partial: "filters/checkbox_filter", locals: { f: f, options: years, label: "Collection year", category: "years" } %>
                     <%= render partial: "filters/checkbox_filter", locals: { f: f, options: status_filters, label: "Status", category: "status" } %>
+                    <%= render partial: "filters/checkbox_filter", locals: { f: f, options: {"all": "All", "yours": "Yours"}, label: "Logs", category: "user" } %>
```
baarkerlounger, line 19: If you made this:

```
options: {"all": "", "yours": current_user.id}
```
Could you avoid the need for special case logic above?

app/controllers/case_logs_controller.rb
baarkerlounger: I was thinking you could do it based on what's in the view. Not super important though if it's fiddly.

app/views/case_logs/_log_filters.erb
kosiakkatrina: So we're using key as a param and value as label, it would need to be changed around. I feel like it'd be a bit weird to have a hash like this `{"In progress": "in_progress"}` but I guess possible

baarkerlounger (approved):

PR #469, opened 2022-04-13
Move privacy notice add final review

dushan-madetech:
### Introduced by this PR:
- Move the privacy notice question into the Household Characteristics section
- Adds a review page
---
### Screenshots of the frontend changes:
image
Screenshot 2022-04-13 at 09 24 06

image


app/controllers/form_controller.rb
```diff
@@ -1,33 +1,44 @@
 class FormController < ApplicationController
   before_action :authenticate_user!
-  before_action :find_resource, only: [:submit_form]
-  before_action :find_resource_by_named_id, except: [:submit_form]
+  before_action :find_resource, only: [:submit_form, :review]
+  before_action :find_resource_by_named_id, except: [:submit_form, :review]
 
   def submit_form
```
baarkerlounger, line 6: Might be worth splitting this into two now: "submit form" and "submit form page" or similar (with their own routes)

baarkerlounger: That would also avoid the need to check referrer I think?

baarkerlounger (approved):

app/models/form/question.rb
```diff
@@ -174,6 +174,8 @@ def has_inferred_display_value?(case_log)
 
   def checkbox_answer_label(case_log)
     answer = []
+    return "Yes" if id == "declaration" && value_is_yes?(case_log["declaration"])
```
baarkerlounger, line 177: Can this not just work the same as other checkbox questions? (i.e. 1 is automatically translated to Yes?

PR #470, opened 2022-04-13
Import changes

smeny:


app/models/case_log.rb
```diff
@@ -416,11 +408,11 @@ def set_derived_fields!
                          weekly_value(tshortfall)
                        end
     self.nocharge = household_charge&.zero? ? 1 : 0
-    self.underoccupation_benefitcap = 3 if renewal == 1 && year == 2021
+    self.underoccupation_benefitcap = 3 if renewal == 1 && collection_start_year == 2021
```
baarkerlounger: This would be better moved into the `if` below to avoid having the same condition twice. Which I think would also make it obvious that we have a bug there? Effectively we have:

```ruby
if is_renewal?
      self.underoccupation_benefitcap = 3 if collection_start_year == 2021
      self.underoccupation_benefitcap = 2 if collection_start_year == 2021
```

So it'll never be 3?

app/models/case_log.rb
```diff
@@ -447,7 +439,7 @@ def process_postcode_changes!
 
   def process_previous_postcode_changes!
     self.ppostcode_full = ppostcode_full.present? ? ppostcode_full.upcase.gsub(/\s+/, "") : ppostcode_full
-    process_postcode(ppostcode_full, "previous_postcode_known", "is_previous_la_inferred", "prevloc", "ppostc1", "ppostc2")
+    process_postcode(ppostcode_full, "previous_postcode_known", "is_previous_la_inferred", "prevloc", nil, nil)
```
baarkerlounger: Do we need to pass `nil` here at all or can we remove it from the method signature if it's being removed everywhere?

app/services/imports/case_logs_import_service.rb
```diff
@@ -0,0 +1,278 @@
...
+        attributes["ecstat#{index}"] = unsafe_string_as_integer(xml_doc, "P#{index}Eco")
+      end
+      # attributes["hhmemb"] =
+
+      # Not specific to our form but required for CDS and can't be inferred
+      # attributes["form"] = Integer(field_value(xml_doc, "xmlns", "FORM"))
+      # attributes["lettype"] = let_type(xml_doc, attributes)
+
+      case_log = CaseLog.new(attributes)
+      case_log.save!
```
baarkerlounger, line 153: Might be quicker (if we're importing thousands of cases), to create CaseLogs in memory (could be parallelized) and then `upsert` them all at once? (Can batched if all at once is too many)?

app/services/imports/case_logs_import_service.rb
```diff
@@ -0,0 +1,278 @@
...
+      when "GN"
+        GN_SH[:general_needs]
+      when "SH"
+        GN_SH[:supported_housing]
+      else
+        raise "Unknown needstype value: #{gn_sh}"
+      end
+    end
+
+    # This does not match renttype (CDS) which is derived by case logs logic
```
baarkerlounger: Is this comment still true?

app/services/imports/case_logs_import_service.rb
```diff
@@ -0,0 +1,278 @@
...
+    #   end
+    # end
+
+    def find_organisation_id(xml_doc, field)
+      old_visible_id = field_value(xml_doc, "xmlns", field).to_i
+      landlord = field_value(xml_doc, "xmlns", "Landlord").to_i
+
+      organisation = Organisation.find_by(old_visible_id:)
+      # Quick hack: should be removed when all organisations are imported
+      # Will fail in the future if the organisation is missing
```
baarkerlounger, line 237: Might be better (after the first organisation is successfully imported perhaps), to have a single "import org" task that runs the individual tasks in order for an org such that you know an org exists by the time you're here and just error here if one doesn't exist?

app/services/imports/case_logs_import_service.rb
smeny: Yup, I was planing to look at performance and robustness once we got all the fields in ;)

app/services/imports/case_logs_import_service.rb
smeny: Yes, our rent_type won't match what CDS has (hence the warning to avoid confusion)

baarkerlounger: Isn't that a problem?

app/models/case_log.rb
smeny: I might need @kosiakkatrina's help to understand what the original intent was.

app/models/case_log.rb
smeny: Part of the next batch of DB changes

app/services/imports/case_logs_import_service.rb
smeny: Nothing we can do, this is one field that needs to be resolved at export. The way we capture the rent type is different (we don't rely on let_type / lar etc...)

app/models/case_log.rb
kosiakkatrina: This should definitely be 2 and not 3, because 2 is "No" and 3 is "Don't know" which seems wrong to infer. Also checked the service map and renewals have underoccupation_benefitcap as "No" (2). Could you please remove the wrong line? not sure why it was added

db/schema.rb
```diff
@@ -318,12 +306,12 @@
     t.string "last_sign_in_ip"
     t.integer "role"
     t.string "old_user_id"
+    t.string "phone"
```
smeny: Adjust the phone back

kosiakkatrina (approved):
GO!

PR #472, opened 2022-04-13
Fix saving date

kosiakkatrina:
The startdate was being saved an hour early in the db, which meant that it was saved as 23:00 on the previous day. That became apparent when filtering by year was off by 1 day, setting timezone to utc when creating the dates fixes that

### DIFF
--- a/spec/models/case_log_spec.rb
+++ b/spec/models/case_log_spec.rb
@@ -1846,6 +1846,14 @@ RSpec.describe CaseLog do
       it "can filter by year(s) AND status" do
         expect(described_class.filter_by_years(%w[2021 2022]).filter_by_status("completed").count).to eq(1)
       end
+
+      it "filters based on date boundaries correctly" do
+        case_log_1.update!(startdate: Time.zone.local(2022, 4, 1))
+        case_log_2.update!(startdate: Time.zone.local(2022, 3, 31))
+
+        expect(described_class.filter_by_years(%w[2021]).count).to eq(1)
+        expect(described_class.filter_by_years(%w[2022]).count).to eq(2)
+      end
     end
 
     context "when filtering on status" do
--- a/app/models/case_log.rb
+++ b/app/models/case_log.rb
@@ -41,7 +41,7 @@ class CaseLog < ApplicationRecord
     years.each { |year| query = query.or(filter_by_year(year)) }
     query.all
   }
-  scope :filter_by_year, ->(year) { where(startdate: Time.utc(year.to_i, 4, 1)...Time.utc(year.to_i + 1, 4, 1)) }
+  scope :filter_by_year, ->(year) { where(startdate: Time.zone.local(year.to_i, 4, 1)...Time.zone.local(year.to_i + 1, 4, 1)) }
 
   scope :filter_by_user, lambda { |selected_user, user|
                            if !selected_user.include?("all") && user.present?


spec/models/case_log_spec.rb
```diff
@@ -1846,6 +1846,14 @@ def check_previous_postcode_fields
       it "can filter by year(s) AND status" do
         expect(described_class.filter_by_years(%w[2021 2022]).filter_by_status("completed").count).to eq(1)
       end
+
+      it "filters based on date boudaries correctly" do
```
baarkerlounger: Typo on boundaries

app/controllers/form_controller.rb
```diff
@@ -72,9 +72,9 @@ def responses_for_page(page)
         next unless [day, month, year].any?(&:present?)
 
         result[question.id] = if day.to_i.between?(1, 31) && month.to_i.between?(1, 12) && year.to_i.between?(2000, 2200)
-                                Date.new(year.to_i, month.to_i, day.to_i)
+                                Time.zone.local(year.to_i, month.to_i, day.to_i).utc
                               else
-                                Date.new(0, 1, 1)
+                                Time.zone.local(0, 1, 1).utc
```
baarkerlounger: Do we actually want to do this? I assume this is the source of the bug that people have reported where entering an invalid date prefills the fields with 0, 1, 1 which is unexpected.

smeny: I thought the UTC settings is at the DB level, the date should contain the timezone and adjust accordingly when saving.

smeny: https://docs.publishing.service.gov.uk/manual/conventions-for-rails-applications.html

kosiakkatrina: @baarkerlounger hmm, yeah, you're right that's the source of the bug, I think we're using it for validations to check if a value was entered at all tho (?) it's a bit hacky and weird, but I assume I wrote it

@smeny If you mean this, then we have that in our config, but it still sets it to an hour earlier 🤷‍♀️  https://docs.publishing.service.gov.uk/manual/conventions-for-rails-applications.html#specify-london-as-the-timezone

spec/models/case_log_spec.rb
kosiakkatrina: One day I'll learn how to type.. Thanks

app/controllers/form_controller.rb
smeny: Isn't that correct? We currently have BST an hour after UTC.

baarkerlounger: @smeny my understanding is:

ActiveRecord always converts and saves time as UTC. The rails app config locale, tells it what to convert back to when it pulls the record out of DB.

The problem here is we don't actually have a real time at all (and maybe shouldn't pretend that we do), we actually just have a date. 

So when you create a record in summer time you get:

```ruby
Time.zone.local(2021,4,1)
=> Thu, 01 Apr 2021 00:00:00.000000000 BST +01:00
```

Which ActiveRecord then converts to UTC by subtracting an hour.

This does a double conversion:
```ruby
Time.zone.local(2021,4,1).utc
=> 2021-03-31 23:00:00 UTC
```

What I think you actually want is to create it as UTC in the first place so that no conversion is needed:

```ruby
Time.utc(2021,4,1)
=> 2021-04-01 00:00:00 UTC
```

That way ActiveRecord converts nothing and you save what you expected.

baarkerlounger (approved):

PR #474, opened 2022-04-14
Fix check answers bugs

dushan-madetech:


app/models/form/subsection.rb
```diff
@@ -30,9 +30,10 @@ def status(case_log)
       return :cannot_start_yet
     end
 
-    qs = applicable_questions(case_log).reject { |q| CaseLog::OPTIONAL_FIELDS.include?(q.id) }
+    qs = applicable_questions(case_log)
+    qs_optional_removed = qs.reject { |q| CaseLog::OPTIONAL_FIELDS.include?(q.id) }
     return :not_started if qs.all? { |question| case_log[question.id].blank? || question.read_only? }
-    return :completed if qs.all? { |question| question.completed?(case_log) }
+    return :completed if qs_optional_removed.all? { |question| question.completed?(case_log) }
```
kosiakkatrina, line 36: is this just a refactor?

dushan-madetech: No, it was to fix a bug with the property information section since line 34 meant that the subsection was considered not started even if you answered the first 3 questions of it (since they're included in the optional fields and filtered out)

config/forms/2021_2022.json
```diff
@@ -1147,7 +1147,12 @@
                       0
                     ]
                   },
-                  "hidden_in_check_answers": true
+                  "hidden_in_check_answers": {
+                    "depends_on": [
+                      { "age1_known": 0 },
+                      { "age1_known": 1 }
+                    ]
+                  }
```
kosiakkatrina, line 1155: How does this change from the behaviour we have now?

dushan-madetech: previously it was only either true or false assigned directly but now it can be set to be conditional on something else

dushan-madetech: Right now it means that this question will only be hidden if the `age1_known` has been answered. This wasn't the case before. You could navigate away and the age question would be hidden on check answers

app/models/form/subsection.rb
kosiakkatrina: Ah, okay, gotcha!

config/forms/2021_2022.json
kosiakkatrina: What's the reason for not hiding it all the time?

dushan-madetech: If it is hidden all the time you get a bug where you can't answer the lead tenant age question if you only answer the number of household members question and then navigate away from the page

dushan-madetech: Since you will get taken to check answers upon revisiting the section - this just make sure it appears in that case

kosiakkatrina: But wouldn't you see the "tenant age" question? Happy with the change if not 🤷‍♀️

kosiakkatrina (approved):

config/forms/2021_2022.json
dushan-madetech: Before:
Screenshot 2022-04-14 at 15 03 34

After:
Screenshot 2022-04-14 at 15 02 03

dushan-madetech: Allowing you to go back and answer no for example which then gives you this:
Screenshot 2022-04-14 at 15 05 27

PR #476, opened 2022-04-19
Filter fixes

paulrobertlloyd:
Few issues with the filter layout and toggle button implementation, such that the filters were not appearing as expected (white checkboxes on grey background, borders between filter options, etc.), and the filter would disappear if you toggled filters on a narrower viewport then made the viewport wider.

This PR:

* Ensures the styles are applied correctly
* Uses the JS used on the prototype (which has been used in production on other services). This adds toggle buttons dynamically, listens for viewport changes, adds the required ARIA attributes, and manages focus. (Showing and hiding buttons isn’t enough for this pattern to be accessible).
* 2 spaces not 4 spaces for indentation 💅 

| Before | After |
| - | - |
| filter-before | filter-after |


app/frontend/modules/filter_toggle.js
```diff
@@ -0,0 +1,112 @@
+export class FilterToggle {
```
paulrobertlloyd, line 1: Not sure where best to save this file, as it’s not a controller, nor is it really something from a vendor. So created a new directory called `modules`.

app/frontend/modules/filter_toggle.js
```diff
@@ -0,0 +1,112 @@
...
+
+  enableBigMode () {
+    this.showMenu()
+    this.removeMenuButton()
+    this.removeCloseButton()
+  }
+
+  enableSmallMode () {
+    this.options.filter.container.setAttribute("tabindex", "-1")
+    this.hideMenu()
```
kosiakkatrina, line 29: If we open the side menu in small mode, resize it to large and immediately resize it back to small, it'll hide the side menu. Is that intended?

app/frontend/modules/filter_toggle.js
```diff
@@ -0,0 +1,112 @@
...
+    }
+    this.options.filter.container.removeAttribute("hidden")
+  }
+
+  onMenuButtonClick () {
+    this.toggle()
+  }
+
+  toggle () {
+    if (this.menuButton.getAttribute("aria-expanded") === "false") {
```
kosiakkatrina: shouldn't `aria-expanded` only be used for accessibility? Maybe using `display` or `hidden` attributes would be better?

app/frontend/modules/filter_toggle.js
paulrobertlloyd: Yeah, I think that’s fine (one of those things where there’s not a right answer). It’s kind of an edge case (though one that may happen if you’re on a tablet and rotate it), but I think the most important thing is that the filters don’t disappear on desktop widths, which is what’s happening at the moment.

app/frontend/modules/filter_toggle.js
paulrobertlloyd: This attribute is used to inform the user the state of the element this button controls (and thus track the state of that element), so it’s nothing to do with visibility (when the filters are hidden, the `hidden` attribute is applied to the `app-filter-layout__filter` container.)

app/frontend/modules/filter_toggle.js
kosiakkatrina: That works for me 👍 Nice spot on the desktop widths!

app/frontend/modules/filter_toggle.js
kosiakkatrina: If you agree that arias are used to provide informative content as opposed to the functional one why would we use them to determine whether we should hide or show the element? On lines 82 and 90 we're setting hidden attribute for the filter container which seems like the more indicative option that we could use for this functionality. Or we could just set a property in this class that keeps the track of open/closed state, which would prevent us from having to check the DOM in general (?)

paulrobertlloyd: Do you mean check the value of the `hidden` attribute, as opposed to the `aria-expanded` attribute?

The good thing about using an `aria` attribute (as opposed to a class) is it means it has to be accessible for it to work; there’s a direct link between what is being described and what is happening.

kosiakkatrina: Yup, that's what I meant! I deffo understand the importance having the arias! I'd probably just wouldn't expect them to be coupled with the functionality, because although what is being described and what is happening should be aligned, it is not necessarily enforced. So if for some reason in the future a bit of code gets rewritten and is impacting the aria attributes, it would probably be better if it didn't impact the toggle function itself

app/frontend/modules/filter_toggle.js
```diff
@@ -0,0 +1,112 @@
...
+    }
+    this.options.filter.container.removeAttribute("hidden")
+  }
+
+  onMenuButtonClick () {
+    this.toggle()
+  }
+
+  toggle () {
+    if (this.options.filter.container.hidden) {
```
paulrobertlloyd, line 98: @kosiakkatrina How’s about this, then?

kosiakkatrina: This makes sense to me! Thank you thank you!

kosiakkatrina (approved):

app/frontend/modules/filter_toggle.js
paulrobertlloyd: I agree, makes much more sense 😄  Thanks!

PR #477, opened 2022-04-20
Add navigation tab, remove the old navigation

kosiakkatrina:
Add new navigation bar for support users:
image

And other users:
image

Remove `Logs` and `Your organisation` from header navigation.
Remove tab navigation component from `Your organisation` page.
Add 2 new blank pages for support users (`Users` and `Organisations`)


paulrobertlloyd:
@kosiakkatrina  Few things:

* Tweaked the application layout so that the header uses the `product_name` param and sets the class which ensures navigation items appear at the end of the header. Also added an `app-header` modifier class which is used to adapt the layout of the header at different breakpoints.
* Fixed the ordering of components in the application layout (side benefit being the border below the phase banner is hidden as intended when primary navigation follows it).
* Made the primary navigation a component, like the tab navigation component it replaces.

Possibly not for this PR (?), but one issue I noticed is that navigation items don’t remain highlighted as current if you go into a nested page. Maybe that’s still to come!?

baarkerlounger (approved):

PR #482, opened 2022-04-20
Add ERB Lint

paulrobertlloyd:
Right now, we lint Ruby code for syntax errors, but don’t do the same for Ruby code used in ERB templates. This has lead to inconsistent code style, both across the application, but often within the same ERB templates.

This PR:
* Adds [ERB Lint](https://github.com/Shopify/erb-lint) as a development dependency
* Enables the [default ERB linters](https://github.com/Shopify/erb-lint#Linters) as well as [RuboCop](https://github.com/Shopify/erb-lint#Rubocop) using our config (disabling rules where relevant)
* Lints code against this config

Hopefully none of the changes are controversial, given they follow the rules we are using elsewhere.

It might be worth creating a lint task that runs Rubocop, ERB Lint and SCSS Lint together, but that can come in a later PR.

### DIFF
--- a/.erb-lint.yml
+++ b/.erb-lint.yml
@@ -1,0 +1,24 @@
+---
+EnableDefaultLinters: true
+linters:
+  Rubocop:
+    enabled: true
+    rubocop_config:
+      inherit_from:
+        - .rubocop.yml
+      Layout/ArgumentAlignment:
+        Enabled: false
+      Layout/CommentIndentation:
+        Enabled: false
+      Layout/FirstArgumentIndentation:
+        Enabled: false
+      Layout/FirstArrayElementIndentation:
+        Enabled: false
+      Layout/FirstHashElementIndentation:
+        Enabled: false
+      Layout/InitialIndentation:
+        Enabled: false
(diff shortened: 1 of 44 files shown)


.erb-lint.yml
```diff
@@ -0,0 +1,24 @@
+---
+EnableDefaultLinters: true
+linters:
+  Rubocop:
+    enabled: true
+    rubocop_config:
+      inherit_from:
+        - .rubocop.yml
+      Layout/ArgumentAlignment:
```
baarkerlounger, line 9: Any particular reason for disabling these specifically?

paulrobertlloyd: Yes, they don’t make sense in the context of ERB templates. Rubocop reads everything between the `<%` and the `%>` which are typically isolated from other code in the template.

For example, this code will throw errors:

```erb
<% subsection_status = subsection.status(@case_log) %>
```

* It is indented by 1 space
* It doesn’t have a new line after it
* It’s a useless assignment

None of which is helpful in this context.

baarkerlounger (approved):

PR #484, opened 2022-04-21
CLDC-506: Some organisations only operate in specific LAs

baarkerlounger:
- [x] Import Organisation LA data from exports
- [x] Display Organisation LA data in Org Details page
- [x] Add validation that selected LA must be one of the Organisation's LAs if they have any

image

image


app/services/imports/organisation_la_import_service.rb
```diff
@@ -0,0 +1,23 @@
...
+      organisation = Organisation.find_by(old_org_id: record_field_value(xml_document, "InstitutionId"))
+
+      OrganisationLa.create!(
+        organisation:,
+        ons_code: record_field_value(xml_document, "ONSCode"),
+      )
+    end
+
+    def record_field_value(xml_document, field)
+      xml_document.at_xpath("//#{field}")&.text
```
smeny, line 20: Maybe a method we could add at the top level? (ImportService)
Currently, the other method we have needs a namespace but there is value in having both methods.

baarkerlounger: Hmm I'm sort of hackily making this work by removing all namespaces from the doc in this task, so would need to work out another way of doing that - any ideas?

PR #486, opened 2022-04-21
Reduce differences between imported values and saved values

smeny:


.github/workflows/production_pipeline.yml
```diff
@@ -180,4 +180,5 @@ jobs:
         cf set-env $APP_NAME IMPORT_PAAS_INSTANCE $IMPORT_PAAS_INSTANCE
         cf set-env $APP_NAME EXPORT_PAAS_INSTANCE $EXPORT_PAAS_INSTANCE
         cf set-env $APP_NAME SENTRY_DSN $SENTRY_DSN
+        cf set-env $APP_NAME RAILS_LOG_TO_STDOUT true
```
baarkerlounger: Does make any practical difference?

app/services/imports/import_service.rb
```diff
@@ -13,6 +13,8 @@ def import_from(folder, create_method)
         file_io = @storage_service.get_file_io(filename)
         xml_document = Nokogiri::XML(file_io)
         send(create_method, xml_document)
+      rescue StandardError => e
```
baarkerlounger, line 16: Not a blocker for this PR but I wonder if we'd be better off wrapping the import task in a transaction and just raising an error rather than logging (so the whole task either completes successfully or completely rolls back).

baarkerlounger (approved):

.github/workflows/production_pipeline.yml
baarkerlounger: I'm wondering if you're just adding it for explicitness whether it might be unexpected that removing it would just set it back to `enabled` rather than `false` and whether it might be confusing that the default is `enabled` and this is `true` (do they do something different?). Maybe I'm overthinking this...

app/services/imports/case_logs_import_service.rb
```diff
@@ -140,17 +141,31 @@ def create_log(xml_doc)
       attributes["postcode_full"] = compose_postcode(xml_doc, "POSTCODE", "POSTCOD2")
       attributes["postcode_known"] = attributes["postcode_full"].nil? ? 0 : 1
 
-      # Not specific to our form but required for CDS and can't be inferred
+      # Not specific to our form but required for consistency (present in import)
       attributes["old_form_id"] = Integer(field_value(xml_doc, "xmlns", "FORM"))
+      attributes["created_at"] = Time.zone.parse(field_value(xml_doc, "meta", "created-date"))
+      attributes["updated_at"] = Time.zone.parse(field_value(xml_doc, "meta", "modified-date"))
```
smeny, line 147: User ID link is missing

app/services/imports/case_logs_import_service.rb
```diff
@@ -162,7 +177,11 @@ def safe_string_as_integer(xml_doc, attribute)
     # Safe: A string that represents only a decimal (or empty/nil)
     def safe_string_as_decimal(xml_doc, attribute)
       str = field_value(xml_doc, "xmlns", attribute)
-      BigDecimal(str, exception: false)
+      if str.blank?
+        BigDecimal("0.0")
+      else
+        BigDecimal(str, exception: false)
+      end
```
smeny, line 194: Charges are only 0 if one of them is set

`  if %i[brent scharge pscharge supcharg].any? { |f| public_send(f).present? }
      self.brent ||= 0
      self.scharge ||= 0
      self.pscharge ||= 0
      self.supcharg ||= 0
      self.tcharge = brent.to_f + scharge.to_f + pscharge.to_f + supcharg.to_f
    end`

app/services/imports/import_service.rb
smeny: Add the filename is the error log

smeny:
Move away processed XML files in a folder

PR #489, opened 2022-04-22
Add error to postcode for LA validation

baarkerlounger:
image


baarkerlounger:
@Dan13L3 could you check this error message?

Dan13L3:
@baarkerlounger This would be better as: 'Postcode must be in an area covered by this service'

baarkerlounger:
@Dan13L3 `Postcode must be in an area covered by %{organisation_name}`? It is specific to the org, not the service as a whole.

Dan13L3:
@baarkerlounger Ahh, okay. Your suggestion works then!

PR #493, opened 2022-04-25
Compare partial URLs to highlight navigation items

kosiakkatrina:


app/components/primary_navigation_component.html.erb
```diff
@@ -2,7 +2,7 @@
   <div class="govuk-width-container">
     <ul class="app-primary-navigation__list">
       <% items.each do |item| %>
-        <% if item.fetch(:current, false) || current_page?(item.fetch(:url)) || current_page?("#{item.fetch(:url)}/details") %>
+        <% if item.fetch(:current, false) || item.fetch(:comparable_urls).any? { |url| request.fullpath.include?(url) } %>
```
dushan-madetech: I think this should be moved into a helper of some form so you can wrap some unit tests around this line. It would also help to clarify the purpose of this.

dushan-madetech (commented):
Left a comment - I don't have context on this one so leaving the approval of this one to Dan or Stèphane since they'll probably have more

dushan-madetech (approved):
Nice updates 😄

PR #497, opened 2022-04-25
Loop the subsections on the last subsection instead of declaration

kosiakkatrina:


app/models/form.rb
```diff
@@ -87,7 +87,7 @@ def next_subsection(subsection, case_log, subsection_ids)
     next_subsection_id_index = subsection_ids.index(subsection.id) + 1
     next_subsection = get_subsection(subsection_ids[next_subsection_id_index])
 
-    if subsection.id == "declaration" && case_log.status != "completed"
+    if subsection_ids[subsection_ids.length - 1] == subsection.id && case_log.status != "completed"
```
baarkerlounger, line 90: Should there be a test for this?

kosiakkatrina: There's a test for that already! It's cause in that test we're using declaration as the last subsection, so it didn't break (I broke it in the process to find it tho)

baarkerlounger (approved):

PR #499, opened 2022-04-25
Use summary list component

paulrobertlloyd:
In the DAC report, a number of cases were reported on the change links not being correctly rendered (i.e. without a space between the visible and hidden text). Using the component means we avoid this happening in the future. (It’s also nicer and helps prevent me from continuously flinching…)


app/views/form/check_answers.html.erb
```diff
@@ -19,11 +19,15 @@
     <% end %>
     <%= display_answered_questions_summary(subsection, @case_log) %>
 
-    <dl class="govuk-summary-list">
+    <%= govuk_summary_list do |summary_list| %>
```
baarkerlounger: This shouldn't have an `=` should it?

app/views/form/review.html.erb
```diff
@@ -21,14 +21,18 @@
             <h3 class="x-govuk-summary-card__title"><%= subsection.label %></h3>
           </div>
           <div class="x-govuk-summary-card__body">
-            <dl class="govuk-summary-list">
+            <%= govuk_summary_list do |summary_list| %>
```
baarkerlounger: Same comment here. This also looks similar enough to above to potentially be extracted into a partial?

app/views/form/check_answers.html.erb
paulrobertlloyd: The initial wrapping function should, according to the docs: https://govuk-components.netlify.app/components/summary-list/#input-erb-a-summary-list-with-various-entries

baarkerlounger (approved):

app/views/form/review.html.erb
paulrobertlloyd: > This also looks similar enough to above to potentially be extracted into a partial?

Ah, good catch. Am now a using a partial for the entire list, which means passing around fewer params. 👌

PR #505, opened 2022-04-26
remove la_known

kosiakkatrina:


baarkerlounger (approved):

smeny:
I think you might have to do the same for `prevloc` (previous_la_known)

kosiakkatrina:
> I think you might have to do the same for `prevloc` (previous_la_known)

@smeny prevloc is still optional!

PR #507, opened 2022-04-26
Fix errors during import process

smeny:


app/services/imports/case_logs_import_service.rb
```diff
@@ -148,7 +148,7 @@ def create_log(xml_doc)
       attributes["old_id"] = field_value(xml_doc, "meta", "document-id")
 
       # Required for our form invalidated questions (not present in import)
-      attributes["previous_la_known"] = 1 # Defaulting to Yes (Required)
+      attributes["previous_la_known"] = attributes["prevloc"].nil? ? nil : 1 # 0 (No) is not selectable in OLD Core
```
baarkerlounger: Shouldn't this be:

```ruby
if attributes["prevloc"].nil? ? 0 : 1
```

At least assuming it's a completed log we should assume that if the haven't entered an LA they don't know it? Nil might make more sense for in progress logs where they may have just not answered it yet.

smeny: That's the problem, there is no assumption a log is completed. I guess both are valid, but with yours, we get closer to a completed log so I will make the change.

baarkerlounger: Is there no status field of some kind in the data we can use?

PR #511, opened 2022-04-26
Update en.yml

Dan13L3:
@baarkerlounger Hopefully captured all the person vs. tenant distinctions here

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -154,30 +154,30 @@ en:
         question_required: "You must answer whether the person is still serving in the UK armed forces as you told us they're a current or former regular"
         question_not_required: "You cannot answer whether the person is still serving in the UK armed forces as you told us they're not a current or former regular"
       preg_occ:
-        no_female: "You must answer 'no' as there are no female tenants aged 16-50 in the property"
+        no_female: "You must answer 'no' as there are no females aged 16-50 in the household"
       age:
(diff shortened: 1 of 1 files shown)


config/locales/en.yml
```diff
@@ -154,30 +154,30 @@ en:
         question_required: "You must answer whether the person is still serving in the UK armed forces as you told us they're a current or former regular"
         question_not_required: "You cannot answer whether the person is still serving in the UK armed forces as you told us they're not a current or former regular"
       preg_occ:
-        no_female: "You must answer 'no' as there are no female tenants aged 16-50 in the property"
+        no_female: "You must answer 'no' as there are no females aged 16-50 in the household"
```
baarkerlounger, line 157: `female persons`? `women`? `females` reads a bit odd...

Dan13L3: @baarkerlounger We've tried to go with male/female to keep it consistent with the way we phrase the gender identity question, but I think 'women' works fine too, so happy to update

baarkerlounger: @Dan13L3 your call, want to change it (if so just push another commit) or want me to merge this as is?

Dan13L3: @baarkerlounger Perhaps merge as is, as I can do a general review of the gender questions, as they might need making consistent throughout

PR #512, opened 2022-04-26
Move link to privacy notice to declaration question

paulrobertlloyd:
I went on a circuitous route on how to get the page set up right, and realised I could have used the `description` field instead of a `guidance_partial`. But given we require HTML, and it’s better to use the `privacy_policy_path` helpers to avoid link rot, using a partial is probably the better approach anyway.

| Before | After |
| - | - |
| Screenshot 2022-04-26 at 15 05 24 | Screenshot 2022-04-26 at 15 05 56 |
| Screenshot 2022-04-26 at 15 05 15 | Screenshot 2022-04-26 at 15 05 34 |


baarkerlounger (approved):
Looks much nicer!

paulrobertlloyd:
(Oh, forgot to add: we can’t include a link to policy in the hint text either, for accessibility reasons – describing a link within a relationship between hint and input is troublesome)

PR #516, opened 2022-04-26
Improved semantics for document list

paulrobertlloyd:
Using a `dl` means we can better describe the relationship between the link and its metadata.

### DIFF
--- a/app/components/document_list_component.html.erb
+++ b/app/components/document_list_component.html.erb
@@ -1,15 +1,15 @@
-<ul class="app-document-list">
+<dl class="app-document-list">
   <% items.each do |item| %>
-    <li class="app-document-list__item">
-      <h3 class="app-document-list__item-title">
+    <div class="app-document-list__item">
+      <dt class="app-document-list__item-title">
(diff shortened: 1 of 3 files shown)


app/components/document_list_component.html.erb
```diff
@@ -1,15 +1,15 @@
-<ul class="app-document-list">
+<dl class="app-document-list">
   <% items.each do |item| %>
-    <li class="app-document-list__item">
-      <h3 class="app-document-list__item-title">
+    <div class="app-document-list__item">
```
kiddhustle, line 3: Does this markup pass an HTML validator check?

I don't think a `<dl>` should contain a`<div>`

kiddhustle (changes requested):

app/components/document_list_component.html.erb
paulrobertlloyd: Valid since HTML 5.2 🎉 https://www.stefanjudis.com/today-i-learned/divs-are-valid-elements-inside-of-a-definition-list/

kiddhustle (approved):

PR #517, opened 2022-04-26
Don’t use bullet list if only one value returned

paulrobertlloyd:
Don’t use a bulleted list if only one item is returned.

| Before | After |
| - | - |
| Screenshot 2022-04-26 at 17 18 05 | Screenshot 2022-04-26 at 17 18 15 |

### DIFF
--- a/app/helpers/details_table_helper.rb
+++ b/app/helpers/details_table_helper.rb
@@ -1,10 +1,12 @@
 module DetailsTableHelper
   def details_html(attribute)
-    if attribute[:format] == :bullet
-      list = attribute[:value].map { |la| "<li>#{la}</li>" }.join
+    if attribute[:format] == :bullet && attribute[:value].length > 1
+      list = attribute[:value].map { |value| "<li>#{value}</li>" }.join
       simple_format(list, { class: "govuk-list govuk-list--bullet" }, wrapper_tag: "ul")
     else
-      simple_format(attribute[:value].to_s, {}, wrapper_tag: "div")
+      value = attribute[:value].is_a?(Array) ? attribute[:value].first : attribute[:value]
+
+      simple_format(value.to_s, { class: "govuk-body" }, wrapper_tag: "p")
     end
   end
 end
--- a/spec/helpers/details_table_helper_spec.rb
+++ b/spec/helpers/details_table_helper_spec.rb
@@ -7,8 +7,8 @@ RSpec.describe DetailsTableHelper do
     context "when given a simple attribute" do
       let(:attribute) { { name: "name", value: "Dummy org", editable: true } }
 
-      it "displays the string wrapped in a div" do
-        expect(details).to eq("<div>Dummy org</div>")
+      it "displays the string wrapped in a p" do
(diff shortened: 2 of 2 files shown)


app/helpers/details_table_helper.rb
```diff
@@ -1,10 +1,12 @@
...
-    if attribute[:format] == :bullet
-      list = attribute[:value].map { |la| "<li>#{la}</li>" }.join
+    if attribute[:format] == :bullet && attribute[:value].length > 1
+      list = attribute[:value].map { |value| "<li>#{value}</li>" }.join
       simple_format(list, { class: "govuk-list govuk-list--bullet" }, wrapper_tag: "ul")
     else
-      simple_format(attribute[:value].to_s, {}, wrapper_tag: "div")
+      value = attribute[:value].is_a?(Array) ? attribute[:value].first : attribute[:value]
+
+      simple_format(value.to_s, { class: "govuk-body" }, wrapper_tag: "p")
```
paulrobertlloyd, line 9: Is this good codez?

app/helpers/details_table_helper.rb
```diff
@@ -1,10 +1,12 @@
 module DetailsTableHelper
   def details_html(attribute)
-    if attribute[:format] == :bullet
-      list = attribute[:value].map { |la| "<li>#{la}</li>" }.join
+    if attribute[:format] == :bullet && attribute[:value].length > 1
+      list = attribute[:value].map { |value| "<li>#{value}</li>" }.join
       simple_format(list, { class: "govuk-list govuk-list--bullet" }, wrapper_tag: "ul")
     else
-      simple_format(attribute[:value].to_s, {}, wrapper_tag: "div")
+      value = attribute[:value].is_a?(Array) ? attribute[:value].first : attribute[:value]
```
baarkerlounger, line 7: Is there a case where we expect this to be an array?

paulrobertlloyd: Yup, if an array contains 1 item. This was my attempt at being DRYer. Other option is to check for length of array, and if has one item, use `simple_format` again.

baarkerlounger: As in, I think we only ever pass an array of things when we expect bullet points, everything else is passed as a string already?

This is fine, was just curious if there was a use case we're supporting here or if it's just in case

paulrobertlloyd: Well, currently we have items on organisation page that say:

> * All

The row is set to output bullets, but in this case there is only one list item.

paulrobertlloyd: Also, in this case, local authorities and rent periods may have 1 or more items, we don’t know.

baarkerlounger: Right right I see! Sorry :+1:

baarkerlounger (approved):

app/helpers/details_table_helper.rb
paulrobertlloyd: (And these always return an array, it seems). Could change that, but still good to have this behaviour as a fallback regardless I suspect.

PR #518, opened 2022-04-26
CLDC-1166: Include guidance partial on both income questions

paulrobertlloyd:
We want to show the guidance about what counts as income on both questions that ask for the household’s total combined income.

* Adding the partial to the first questions means we need to add a page title separate from the fieldset’s legend to keep the complexity low when form is announced by screen readers.
* Page titles should be unique

This is the result:

***

1. ### Household’s combined income after tax
    [▶︎ What count’s as income?](#)
  
   #### Do you know the household’s combined income after tax?

2. ### Total household income
   [▶︎ What count’s as income?](#)
  
   #### How much income does the household have in total?
   #### How often does the household receive this amount?

***

| Before | After |
| - | - |
| Screenshot 2022-04-26 at 17 57 17 | Screenshot 2022-04-26 at 17 56 47 | 
Screenshot 2022-04-26 at 17 57 42 | Screenshot 2022-04-26 at 17 56 59 |

### DIFF
--- a/config/forms/2021_2022.json
+++ b/config/forms/2021_2022.json
@@ -4539,12 +4539,13 @@
           ],
           "pages": {
             "income_known": {
-              "header": "",
+              "header": "Household’s combined income after tax",
               "description": "",
               "questions": {
                 "net_income_known": {
                   "check_answer_label": "Do you know the household’s combined income?",
-                  "header": "Do you know the household’s combined income?",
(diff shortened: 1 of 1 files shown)


paulrobertlloyd:
@Dan13L3 Thoughts?

Dan13L3:
@paulrobertlloyd Looks clear to me!

baarkerlounger (approved):

PR #522, opened 2022-04-27
CLDC-1101: Case log filter by organisation

baarkerlounger:
image
image


paulrobertlloyd:
Will review on Monday, but while you’re here, I noticed that no message is shown currently if there are no logs to show. We should fix that!

kosiakkatrina (approved):

paulrobertlloyd (approved):
Tweaked the autocomplete styling and filter layout to prevent it from breaking (not quite sure why input isn’t respecting a `100%` value…)

PR #520, opened 2022-04-27
Ensure organisation link in primary navigation is highlighted when relevant

paulrobertlloyd:

PR #526, opened 2022-04-27
Add navigation items helper

paulrobertlloyd:
(Follows on from #520, where I thought I’d fixed this… 😓)

With the users being a concern of an organisation, it’s hard to correctly highlight the current item in the primary navigation. Options with the current approach is that either ‘Users’ is not highlighted when visiting `/users`, or both ‘Users’ and ‘Your organisation’ are highlighted if we go by the URL (or current controller).

Screenshot 2022-04-27 at 11 06 27

This is an attempt to abstract out the logic required to ensure the correct item in the (primary) navigation is highlighted, based both on the current controller AND the current action.

Over engineered? My suspicion is that managing this stuff will get more complicated once we start to add support views which have a second row of navigation, so this might be the start of a way of handling that complexity and moving it out of the views.

### To do

- [ ] Don’t highlight ‘Users’ when viewing pages in ‘Your account’

### DIFF
--- a/app/components/primary_navigation_component.html.erb
+++ b/app/components/primary_navigation_component.html.erb
@@ -2,13 +2,13 @@
   <div class="govuk-width-container">
     <ul class="app-primary-navigation__list">
       <% items.each do |item| %>
-        <% if highlighted_tab?(item, request.fullpath) %>
+        <% if item.current %>
           <li class="app-primary-navigation__item app-primary-navigation__item--current">
-            <%= govuk_link_to item[:name], item[:url], class: "app-primary-navigation__link", aria: { current: "page" } %>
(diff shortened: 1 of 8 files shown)


app/components/primary_navigation_component.html.erb
```diff
@@ -2,13 +2,13 @@
   <div class="govuk-width-container">
     <ul class="app-primary-navigation__list">
       <% items.each do |item| %>
-        <% if highlighted_tab?(item, request.fullpath) %>
+        <% if item.current %>
```
kosiakkatrina, line 5: is this meant to be `item.current?` ?

paulrobertlloyd: `item.current` works, `item.current?` doesn’t.

baarkerlounger: The failing tests suggest `.current` doesn't work?

PR #527, opened 2022-04-28
make previous_la_known dynamically optional

kosiakkatrina:


app/models/case_log.rb
```diff
@@ -419,7 +419,8 @@ def reset_invalidated_dependent_fields!
   end
 
   def dynamically_not_required
-    (form.invalidated_questions(self) + form.readonly_questions).map(&:id).uniq
+    previous_la_known_field = postcode_known? ? %w[previous_la_known] : []
+    ((form.invalidated_questions(self) + form.readonly_questions).map(&:id) + previous_la_known_field).uniq
   end
```
dushan-madetech, line 424: Do we have any tests to cover this change in behaviour?

dushan-madetech (commented):
Left a comment - I feel like this should have some kind of test to verify it works as intended and to verify we don't unintentionally break it in the future (so we don't end up having to fix the same issue twice)

app/models/case_log.rb
kosiakkatrina: So one of the tests failed, because it was no longer considering the log as completed, because the previous_la_required is no longer always optional. I've fixed that test by changing the case_log factory to have previous_la_required set as true. Other than that, all of the methods (dynamically_not_required, mandatory_fields etc) are private, so they aren't unit tested 👀

dushan-madetech: That makes sense 👍

dushan-madetech: Will approve now 🙂

dushan-madetech (approved):

PR #530, opened 2022-04-28
CLDC-1181: fix tasklist skip links

kiddhustle:
https://digital.dclg.gov.uk/jira/browse/CLDC-1181


paulrobertlloyd:
Was this broken? The [example in the prototype kit](https://govuk-prototype-kit.herokuapp.com/docs/templates/task-list) adds the `id` attribute to the status tag… though not quite sure why, as it makes more sense where you have it now.

kiddhustle:
this was the recommendation from the DAC report so I guess it is an improvement

paulrobertlloyd:
Yeah, happy to go with this 🙂

PR #531, opened 2022-04-28
CLDC-1185: now wrap label in div

kiddhustle:
https://digital.dclg.gov.uk/jira/browse/CLDC-1185

### DIFF
--- a/app/helpers/question_view_helper.rb
+++ b/app/helpers/question_view_helper.rb
@@ -22,6 +22,6 @@ private
   def label_tag(page_header, conditional)
     return "" if conditional
 
-    page_header.blank? ? "h1" : "h2"
+    page_header.blank? ? "h1" : "div"
   end
 end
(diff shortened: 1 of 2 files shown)


app/helpers/question_view_helper.rb
```diff
@@ -22,6 +22,6 @@ def label_size(page_header, conditional)
   def label_tag(page_header, conditional)
     return "" if conditional
 
-    page_header.blank? ? "h1" : "h2"
+    page_header.blank? ? "h1" : "div"
```
paulrobertlloyd, line 25: Given this is passed to the form builder, I wonder if we can provide `nil`?

kiddhustle: yep that appears to render `<div>` also

paulrobertlloyd: In which case, probably best to leave is as `div`, as that’s more explicit. 👍

paulrobertlloyd: (I thought `nil` would remove the wrapping div, but I guess that’s still required because `legend` be `legend`)

paulrobertlloyd (approved):

PR #533, opened 2022-04-28
Correct hint text for times offered (not social let)

paulrobertlloyd:
Hint text for this question has additional hint if not a social let.

### DIFF
--- a/config/forms/2022_2023.json
+++ b/config/forms/2022_2023.json
@@ -675,7 +675,7 @@
                 "offered": {
                   "check_answer_label": "Times previously offered since becoming available",
                   "header": "Since becoming available for re-let, how many times has the property been previously offered?",
-                  "hint_text": "If the property is being offered for let for the first time, enter 0.",
+                  "hint_text": "This is after the last tenancy ended. If the property is being offered for let for the first time, enter 0.",
                   "type": "numeric",
                   "min": 0,
                   "max": 150,


config/forms/2022_2023.json
```diff
@@ -675,7 +675,7 @@
                 "offered": {
                   "check_answer_label": "Times previously offered since becoming available",
                   "header": "Since becoming available for re-let, how many times has the property been previously offered?",
-                  "hint_text": "If the property is being offered for let for the first time, enter 0.",
+                  "hint_text": "This is after the last tenancy ended. If the property is being offered for let for the first time, enter 0.",
```
roobottom, line 678: Can we get Dan to OK this pls.

paulrobertlloyd: @Dan13L3 Is this correct?

baarkerlounger: It matches 2021_2022 so I'd expect so:

https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/blob/main/config/forms/2021_2022.json#L678

baarkerlounger (approved):

config/forms/2022_2023.json
Dan13L3: @baarkerlounger @paulrobertlloyd This hint text is correct if user has selected 'no - re-let of existing social housing'

PR #540, opened 2022-05-03
Set hhmemb to totadult and tchild sum if hhmemb is not given

kosiakkatrina:
Co-authored-by: Ted <tedbooton@gmail.com>
Co-authored-by: Dushan <dushan@madetech.com>


dushan-madetech (approved):
Approving since we paired on most of this 👍

app/services/imports/case_logs_import_service.rb
```diff
@@ -489,5 +489,18 @@ def postcode_known(attributes)
         1
       end
     end
+
+    def household_members(xml_doc, previous_status)
+      hhmemb = safe_string_as_integer(xml_doc, "HHMEMB")
+      if previous_status.include?("submitted") && hhmemb.nil?
+        hhmemb = people_with_details(xml_doc).count
+        return nil if hhmemb.zero?
```
baarkerlounger, line 497: Should this be a valid submitted log?

spec/fixtures/softwire_imports/case_logs/0ead17cb-1668-442d-898c-0d52879ff592.xml
```diff
@@ -393,7 +393,7 @@
     <Q18aValid>1</Q18aValid>
     <Q18Valid>1</Q18Valid>
     <VACDAYS>0</VACDAYS>
-    <HHMEMB>2</HHMEMB>
```
baarkerlounger, line 396: This looks like we're removing data that was there?

app/services/imports/case_logs_import_service.rb
baarkerlounger: i.e. why are we returning `nil` and not `1`?

baarkerlounger (approved):

PR #545, opened 2022-05-04
Age known believe import fixes

dushan-madetech:
Fixes issue with `age_known` being set to nil

### DIFF
--- a/app/services/imports/case_logs_import_service.rb
+++ b/app/services/imports/case_logs_import_service.rb
@@ -342,13 +342,15 @@ module Imports
 
     def age_known(xml_doc, index, hhmemb)
       return nil if hhmemb.present? && index > hhmemb
-
       age_refused = string_or_nil(xml_doc, "P#{index}AR")
-      if age_refused == "AGE_REFUSED"
-        1 # No
-      else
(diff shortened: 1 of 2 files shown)


kosiakkatrina (approved):

app/services/imports/case_logs_import_service.rb
```diff
@@ -500,7 +502,7 @@ def household_members(xml_doc, previous_status)
     end
 
     def people_with_details(xml_doc)
-      ((2..8).map { |x| string_or_nil(xml_doc, "P#{x}Rel") } + [string_or_nil(xml_doc, "P1Age")]).compact
+      ((2..8).map { |x| string_or_nil(xml_doc, "P#{x}Rel") } + [string_or_nil(xml_doc, "P1Sex")]).compact
```
baarkerlounger, line 506: Can you refuse this too? Should we be checking for any of them?

baarkerlounger (approved):

app/services/imports/case_logs_import_service.rb
dushan-madetech: Only age can be refused 
image

PR #546, opened 2022-05-04
Enable dynamically dependent answer options

baarkerlounger:
Previously economic status was being derived for persons under 16, but we weren't covering the case where data providers don't know the exact age of the child (so put `age2_known: no`) but **do** know that they are a child under 16.

This means that we can't always derive that economic status, and do need to dynamically calculate whether that answer option should be shown or not. This PR adds `depends_on` conditions for question answer options.

It also moves the method for evaluating `depends_on` conditions to the form so we can reuse it for pages, questions, subsections, answer options etc.

### DIFF
--- a/app/models/case_log.rb
+++ b/app/models/case_log.rb
@@ -477,10 +477,14 @@ private
     (2..8).each do |idx|
       if public_send("age#{idx}") && public_send("age#{idx}") < 16
         self["ecstat#{idx}"] = 9
-      elsif public_send("ecstat#{idx}") == 9 && (public_send("age#{idx}").nil? || public_send("age#{idx}") >= 16)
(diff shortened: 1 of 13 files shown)


app/models/case_log.rb
```diff
@@ -477,12 +477,16 @@ def set_derived_fields!
     (2..8).each do |idx|
       if public_send("age#{idx}") && public_send("age#{idx}") < 16
         self["ecstat#{idx}"] = 9
-      elsif public_send("ecstat#{idx}") == 9 && (public_send("age#{idx}").nil? || public_send("age#{idx}") >= 16)
+      elsif public_send("ecstat#{idx}") == 9 && (public_send("age#{idx}").nil? || public_send("age#{idx}") >= 16) && age_known?(idx)
```
dushan-madetech, line 480: My only comment is that we could do some minor refactoring of this to make the full set of conditions easier to understand - like you've done with the `age_known?` method

PR #548, opened 2022-05-05
Cldc 1149 radio buttons filter

Ted-U:
Changed checkboxes in logs filter for "user" and "all" to radio buttons to prevent both options being selected at the same time.


Gemfile.lock
```diff
@@ -379,33 +379,33 @@ GEM
...
       unicode-display_width (>= 1.4.0, < 3.0)
-    rubocop-ast (1.15.1)
-      parser (>= 3.0.1.1)
-    rubocop-govuk (4.3.0)
-      rubocop (= 1.25.0)
-      rubocop-ast (= 1.15.1)
-      rubocop-rails (= 2.13.2)
+    rubocop-ast (1.17.0)
+      parser (>= 3.1.1.0)
+    rubocop-govuk (4.4.0)
```
baarkerlounger: This version of rubocop unfortunately has a bug in it that causes it to error on our linter. I'd suggest reverting the bump/locking govuk-rubocop to 4.3.0 until that's fixed and then pipeline should pass for this.

baarkerlounger (approved):

PR #549, opened 2022-05-05
CLDC-744-joint-tenancy-validation

dushan-madetech:
Adds validation for the joint tenancy field
Screenshot 2022-05-06 at 12 07 36
Screenshot 2022-05-06 at 12 08 26

### DIFF
--- a/app/models/validations/tenancy_validations.rb
+++ b/app/models/validations/tenancy_validations.rb
@@ -35,10 +35,10 @@ module Validations::TenancyValidations
 
   def validate_joint_tenancy(record)
     return unless record.collection_start_year?
-      if record.hhmemb == 1 && record.joint != 2 && record.collection_start_year >= 2022
-        record.errors.add :joint, I18n.t("validations.tenancy.not_joint")
-        record.errors.add :hhmemb, I18n.t("validations.tenancy.not_joint")
-      end
+
+    if record.hhmemb == 1 && record.joint != 2 && record.collection_start_year >= 2022
+      record.errors.add :joint, I18n.t("validations.tenancy.not_joint")
+      record.errors.add :hhmemb, I18n.t("validations.tenancy.not_joint")
     end
   end
 end
--- a/app/models/validations/tenancy_validations.rb
+++ b/app/models/validations/tenancy_validations.rb
@@ -34,7 +34,7 @@ module Validations::TenancyValidations
   end
 
   def validate_joint_tenancy(record)
-    if !record.collection_start_year.blank?
+    return unless record.collection_start_year?
       if record.hhmemb == 1 && record.joint != 2 && record.collection_start_year >= 2022
         record.errors.add :joint, I18n.t("validations.tenancy.not_joint")
         record.errors.add :hhmemb, I18n.t("validations.tenancy.not_joint")
--- a/app/models/validations/tenancy_validations.rb
+++ b/app/models/validations/tenancy_validations.rb
@@ -34,7 +34,7 @@ module Validations::TenancyValidations
   end
 
   def validate_joint_tenancy(record)
-    return unless record.collection_start_year?
+    return unless record.collection_start_year
 
     if record.hhmemb == 1 && record.joint != 2 && record.collection_start_year >= 2022
       record.errors.add :joint, I18n.t("validations.tenancy.not_joint")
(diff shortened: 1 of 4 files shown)


app/models/validations/tenancy_validations.rb
```diff
@@ -32,4 +32,10 @@ def validate_fixed_term_tenancy(record)
   def validate_other_tenancy_type(record)
     validate_other_field(record, 3, :tenancy, :tenancyother)
   end
+
+  def validate_joint_tenancy(record)
+    if record.hhmemb == 1 && record.joint != 2 && record.collection_start_year == 2022
```
baarkerlounger: Should this be `record.collection_start_year >= 2022`?

app/models/validations/tenancy_validations.rb
```diff
@@ -32,4 +32,10 @@ def validate_fixed_term_tenancy(record)
...
+      record.errors.add :joint, I18n.t("validations.tenancy.not_joint")
```
baarkerlounger, line 40: Need to add this to `hhmemb` too since you could trigger it from either page

app/models/validations/tenancy_validations.rb
dushan-madetech: Good catch

dushan-madetech: I hadn't considered future forms

app/models/validations/tenancy_validations.rb
dushan-madetech: Done

app/models/validations/tenancy_validations.rb
dushan-madetech: I had to add a check to make sure the start date wasn't blank otherwise it would error out on >= for a nil value

app/models/validations/tenancy_validations.rb
baarkerlounger: Might need to update the message based on where it's triggered from? @Dan13L3

spec/models/validations/tenancy_validations_spec.rb
```diff
@@ -146,4 +146,26 @@
...
+        tenancy_validator.validate_joint_tenancy(record)
+        expect(record.errors["joint"]).to include(match(expected_error))
+        expect(record.errors["hhmemb"]).to include(match(expected_error))
+      end
+
+      it "does not display an error if the data inputter says the letting is not a joint tenancy" do
+        record.hhmemb = 1
+        record.joint = 2
+        tenancy_validator.validate_joint_tenancy(record)
+        expect(record.errors["joint"]).to be_empty
```
baarkerlounger, line 168: Maybe should expect `(record.errors["hhmemb"]).to be_empty` here as well?

app/models/validations/tenancy_validations.rb
```diff
@@ -32,4 +32,13 @@ def validate_fixed_term_tenancy(record)
   def validate_other_tenancy_type(record)
     validate_other_field(record, 3, :tenancy, :tenancyother)
   end
+
+  def validate_joint_tenancy(record)
+    if !record.collection_start_year.blank?
```
baarkerlounger: `return unless record.collection_start_year`?

app/models/validations/tenancy_validations.rb
Dan13L3: @baarkerlounger @dushan-madetech The error message looks fine to me, as we reference 'household', so should make sense if triggered by either question

baarkerlounger: @Dan13L3 would you not want it reversed so that if it's the household members you're filling out it reads:

"There must be more than 1 person in the household as you told us this is a joint tenancy"

Dan13L3: @baarkerlounger Ahh, I misunderstood - I thought we were saying we could only have a single error for both questions. But yes, agree with your suggestion. Although we'd need to tweak slightly to: "There must be more than one person in the household as you told us this is a joint tenancy"

dushan-madetech: @Dan13L3 @baarkerlounger Updated the message - check the screenshots I've attached to the PR

config/forms/2022_2023.json
```diff
@@ -946,7 +946,7 @@
               "header": "",
               "description": "",
               "questions": {
-                "joint_tenancy": {
```
dushan-madetech, line 949: This was incorrect so I updated it to match the schema

app/models/validations/tenancy_validations.rb
Dan13L3: Thanks @dushan-madetech - looks good to me

baarkerlounger (approved):

PR #550, opened 2022-05-05
Set tshortfall to 0 if it should exists but is not provided

kosiakkatrina:


app/services/imports/case_logs_import_service.rb
```diff
@@ -513,5 +513,12 @@ def household_members(xml_doc, previous_status)
     def people_with_details(xml_doc)
       ((2..8).map { |x| string_or_nil(xml_doc, "P#{x}Rel") } + [string_or_nil(xml_doc, "P1Sex")]).compact
     end
+
+    def tshortfall(xml_doc, attributes)
+      shortfall = safe_string_as_decimal(xml_doc, "Q18dyes")
+      return 0 if shortfall.blank? && attributes["hbrentshortfall"] == 1
```
baarkerlounger: Should this also check for the manual override? I assume we don't want to be setting it for in progress logs where this just hasn't been filled in yet?

kosiakkatrina: ah, good point!

baarkerlounger (approved):

PR #552, opened 2022-05-06
Radio button on log filter is now preset to "All"

Ted-U:
When user loads logs page the "all" filter is pre-selected

Screenshot 2022-05-06 at 11 35 21

### DIFF
--- a/app/views/case_logs/_log_filters.erb
+++ b/app/views/case_logs/_log_filters.erb
@@ -6,9 +6,10 @@
     <div class="app-filter__content">
       <%= form_with url: "/logs", html: { method: :get } do |f| %>
         <% years = {"2021": "2021/22", "2022": "2022/23"} %>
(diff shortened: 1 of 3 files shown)


app/views/case_logs/_log_filters.erb
```diff
@@ -6,9 +6,10 @@
     <div class="app-filter__content">
       <%= form_with url: "/logs", html: { method: :get } do |f| %>
         <% years = {"2021": "2021/22", "2022": "2022/23"} %>
+        <% @all_or_yours = {"all": "All", "yours": "Yours"} %>
```
baarkerlounger: Don't think this needs to be an instance (`@`) variable?

baarkerlounger (approved):
Ship it!

app/views/case_logs/_log_filters.erb
Ted-U: I have removed the `@`

PR #556, opened 2022-05-09
Improve DB seeding

baarkerlounger:
Improve the seed file such that it's actually useful to run in a new production environment as well as a new dev environment.

- [x] Created DLUHC Organisation
- [x] Create dummy users if dev
- [x] Seed LA Rent Ranges

### DIFF
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -6,7 +6,8 @@
 #   movies = Movie.create([{ name: 'Star Wars' }, { name: 'Lord of the Rings' }])
 #   Character.create(name: 'Luke', movie: movies.first)
 
-org = Organisation.create!(
+# rubocop:disable Rails/Output
+org = Organisation.find_or_create_by!(
   name: "DLUHC",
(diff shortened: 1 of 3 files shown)


db/seeds.rb
```diff
@@ -6,7 +6,8 @@
 #   movies = Movie.create([{ name: 'Star Wars' }, { name: 'Lord of the Rings' }])
 #   Character.create(name: 'Luke', movie: movies.first)
 
-org = Organisation.create!(
+# rubocop:disable Rails/Output
```
dushan-madetech, line 9: Out of curiosity, what was this cop complaining about?

dushan-madetech (approved):
I like this a lot! 🚢 it

db/seeds.rb
baarkerlounger: It wants you to log to rails logger rather than print to stdout. Which mostly makes sense but here I'm specifically assuming a local dev environment with this being run from terminal. Could redirect the logger to stdout, but this seems simpler...

PR #559, opened 2022-05-09
Add offset to organisation id

baarkerlounger:
To ensure we don't have clashes with existing organisation data in CDS

### DIFF
--- a/db/migrate/20220509110837_add_id_offset_organisations.rb
+++ b/db/migrate/20220509110837_add_id_offset_organisations.rb
@@ -1,7 +1,0 @@
-class AddIdOffsetOrganisations < ActiveRecord::Migration[7.0]
-  def up
-    execute "SELECT setval('organisations_id_seq', 100000000)"
-  end
-
-  def down; end
-end
--- a/db/migrate/20220509110837_add_id_offset_organisations.rb
+++ b/db/migrate/20220509110837_add_id_offset_organisations.rb
@@ -1,5 +1,7 @@
 class AddIdOffsetOrganisations < ActiveRecord::Migration[7.0]
-  def change
+  def up
     execute "SELECT setval('organisations_id_seq', 100000000)"
   end
+
+  def down; end
 end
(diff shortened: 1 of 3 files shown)


db/migrate/20220509110837_add_id_offset_organisations.rb
```diff
@@ -0,0 +1,7 @@
+class AddIdOffsetOrganisations < ActiveRecord::Migration[7.0]
+  def up
+    execute "SELECT setval('organisations_id_seq', 100000000)"
+  end
+
+  def down; end
```
dushan-madetech: Is there a reason you do this with up/down instead of change?

dushan-madetech (approved):

db/migrate/20220509110837_add_id_offset_organisations.rb
baarkerlounger: It's not reversible so rubocop complains if you put it in a change (which it expects to be reversible).

I'm not sure there's really a reversible way of doing this since it's a bit unusual and effectively "data". There's maybe an argument for not doing this at all and instead just adding the offset to our export in code... @smeny thoughts?

smeny: I think making it not reversible is sensible, I don't see what value we could revert to.
Doing it at export is feasible but then we would always have to translate the IDs for export which I find a bit tedious.
We modelled everything so far with the assumption we would match CDS, so let's stay consistent.

db/migrate/20220509110837_add_id_offset_organisations_and_logs.rb
```diff
@@ -0,0 +1,8 @@
+class AddIdOffsetOrganisationsAndLogs < ActiveRecord::Migration[7.0]
+  def up
+    execute "SELECT setval('organisations_id_seq', 100000000)"
+    execute "SELECT setval('case_logs_id_seq', 300000000000)"
```
baarkerlounger, line 4: 300,000,000,000

The max size of a bigint is 9,223,372,036,854,775,807, so starting at this offset is not a problem.

baarkerlounger:
Alternative approach: https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/562

PR #566, opened 2022-05-10
Use rake task to send onboarding emails

baarkerlounger:
Allows us to generate onboarding emails via:

```bash
cf run-task dluhc-core-production --command "bundle exec rake onboarding_emails:send[2]" --name believe-onboarding-emails
```

Also removes random erb file that somehow seems to have gotten copied into spec folder


lib/tasks/onboarding_emails.rake
```diff
@@ -0,0 +1,22 @@
...
+    raise "Organisation #{organisation_id} does not exist" unless organisation
+
+    organisation.users.each do |user|
+      next unless URI::MailTo::EMAIL_REGEXP.match?(user.email)
+
+      onboarding_template_id = "b48bc2cd-5887-4611-8296-d0ab3ed0e7fd".freeze
+      token = user.send(:set_reset_password_token)
+      url = "#{host}/account/password/edit?reset_password_token=#{token}"
+      personalisation = { name: user.name || user.email, link: url }
+      DeviseNotifyMailer.new.send_email(user.email, onboarding_template_id, personalisation)
```
dushan-madetech, line 19: Does this return anything we can log if the email fails to deliver or this times out for some reason?

dushan-madetech (approved):
Leaving a comment but I'm quite happy for this to be merged in

lib/tasks/onboarding_emails.rake
```diff
@@ -0,0 +1,22 @@
+namespace :onboarding_emails do
+  desc "Send onboarding emails to private beta users"
+  task :send, %i[organisation_id host] => :environment do |_task, args|
+    organisation_id = args[:organisation_id]
+    host = args[:host]
```
smeny: Is there a reason why we make the host configurable?
You still need to have rights to run a task but you could potentially target anything which doesn't sound right to me.

lib/tasks/onboarding_emails.rake
baarkerlounger: I think Notify itself logs failed email sending so I don't think we need to in Rails but worth checking on the first attempt.

lib/tasks/onboarding_emails.rake
baarkerlounger: I suppose it's always configurable but you're right this could use the ENV var we already set. Have changed it.

PR #569, opened 2022-05-11
Remove needs type question until we support supported housing logs

baarkerlounger:


dushan-madetech (approved):
Could you undo this in the future with a `git revert` or would you undo it manually?

baarkerlounger:
@dushan-madetech would probably do it manually but as long as there's no conflicts with future work can't see why you couldn't do a git revert.

PR #573, opened 2022-05-12
CLDC-1196: Fix repeated use of Password in error summary (and use smart quotes)

paulrobertlloyd:
Not quite sure what `ErrorSummaryFullMessagesPresenter` is providing, beyond bugs. Seems like using no helper works better! 🗑️ 

Also… use smart quotes in `blank` and `confirmation` messages. And for `confirmation` message, refer to ‘new password’ instead of ‘Password’. Could probably be improved further, but these changes certain improve upon ‘Password Password’.   

| Before | After |
| - | - |
| Screenshot 2022-05-12 at 18 12 18 | Screenshot 2022-05-12 at 18 11 45 |
| Screenshot 2022-05-12 at 18 12 27 | Screenshot 2022-05-12 at 18 11 26 |


baarkerlounger:
@paulrobertlloyd this PR reintroduces:

image

Which that helper was meant to solve. If you're trying to test this locally set https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/blob/main/config/initializers/devise.rb#L228 to something lower like `2.minutes`, restart your rails server and send yourself a password reset link.

paulrobertlloyd:
What if we update the locale string for this message to include… I guess ‘Token’? I guess we’re fighting with Devise here a little, which is frustrating.

config/locales/devise.en.yml
```diff
@@ -57,6 +57,8 @@ en:
   errors:
     messages:
       already_confirmed: "has already been confirmed. Sign in."
+      blank: "can’t be blank"
+      confirmation: "doesn’t match new password"
       confirmation_period_expired: "needs to be confirmed within %{period}. Request a new account."
       expired: "has expired. Request a new token."
```
paulrobertlloyd: Would changing this string be enough to resolve the original bug?

```suggestion
      expired: "Token has expired. Request a new token."
```

config/locales/devise.en.yml
```diff
@@ -57,6 +57,8 @@ en:
   errors:
     messages:
       already_confirmed: "has already been confirmed. Sign in."
+      blank: "can’t be blank"
+      confirmation: "doesn’t match new password"
```
paulrobertlloyd, line 61: This string assumes that `confirmation` only applies to passwords. Is this the correct assumption?

baarkerlounger: I think so yes, we're not currently using the confirmable module though so I don't think this message actually ever shows anyway

paulrobertlloyd: It shows when you change your password, and use 2 different values (that was how I came across the rogue straight quote!)

config/locales/devise.en.yml
baarkerlounger: Yep this does solve this case (I suppose there may be others but since this fixes definitely existing ones that seems fine).

config/locales/devise.en.yml
baarkerlounger: Ah right, you've added this. Not sure if it's a strictly safe assumption for ever but it is the only thing in Devise using a`validates_confirmation_of` validation so should be fine.

https://github.com/heartcombo/devise/blob/main/lib/devise/models/validatable.rb#L41

baarkerlounger (approved):

PR #574, opened 2022-05-12
CLDC-1200: Use `inputmode="email"` instead of `type="email"`

paulrobertlloyd:
[CLDC-1200](https://digital.dclg.gov.uk/jira/browse/CLDC-1200)

> Voice activation users using Dragon were unable to directly dictate their email address into the email address input field. This is because the input type ‘email’ interferes with Dragon software.

Replaces all occurrences of `govuk_email_field` with `govuk_text_field` and adds the `inputmode` attribute.

### DIFF
--- a/app/views/devise/confirmations/new.html.erb
+++ b/app/views/devise/confirmations/new.html.erb
@@ -18,9 +18,10 @@
 
       <p class="govuk-body">Enter your email address to get a new invitation link.</p>
 
-      <%= f.govuk_email_field :email,
+      <%= f.govuk_text_field :email,
         label: { text: "Email address" },
         autocomplete: "email",
+        inputmode: "email",
         spellcheck: "false",
         value: (resource.pending_reconfirmation? ? resource.unconfirmed_email : resource.email) %>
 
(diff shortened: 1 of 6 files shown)


baarkerlounger:
Should this not rather be filed as a bug upstream?

tvararu:
Hi :wave: 

> Should this not rather be filed as a bug upstream?

Yeah, this could be an improvement baked into the [design system guidance for entering email addresses](https://design-system.service.gov.uk/patterns/email-addresses/), assuming it doesn't break something else.

I can't find any mentions on `inputmode` or `dragon` in the relevant backlog discussion: https://github.com/alphagov/govuk-design-system-backlog/issues/45

I think it's worth a deeper investigation and perhaps more testing, maybe it's only recent versions of Dragon that break this functionality?

The core `govuk-frontend` components and patterns should be pretty robust by this point :thinking:

paulrobertlloyd:
Hello stranger!

Started a discussion about this on x-gov Slack and opened an issue to update the guidance: https://github.com/alphagov/govuk-design-system/issues/2187

Assuming the guidance gets updated, can then make change upstream in the GOV.UK Form Builder gem (this has no impact on `govuk-frontend`).

PR #576, opened 2022-05-13
Hide inactive users and allow support users to view all users

baarkerlounger:
- [x] Inactive users are hidden from the Organisation users view (Data Coordinators)
- [x] User index view exists for support users
- [x] Inactive users are hidden from the User index view (Support users)
- [x] User index view redirects to organisation users view for data coordinators
- [x] Support users can invite user to any organisation

Users tab from Data Coordinator account
image

Users tab from Support account
image

Invite new user from Support account
image


app/views/organisations/index.html.erb
```diff
@@ -0,0 +1 @@
+
```
dushan-madetech, line 1: This is just empty?

dushan-madetech (approved):

app/views/organisations/index.html.erb
baarkerlounger: Yep it's just a placeholder for https://digital.dclg.gov.uk/jira/browse/CLDC-1102 for now

dushan-madetech: Ah, I'll rebase the 1102 branch in that case

PR #578, opened 2022-05-16
CLDC-1224: Tenancy type and tenancy length

baarkerlounger:
https://digital.dclg.gov.uk/jira/browse/CLDC-1224

- [x] From Rachel (1 Secure no longer exists. It has been split into 6 & 7). Incorrectly using 1 rather than 4 for AST.
- [x] Tenancy length is not mandatory for 21/22
- [x] Tenancy length is mandatory for 22/23 (Assured Shorthold and Secure Fixed Term)
- [x] Tenancy length is optional for "Other"
- [x] Tenancy length is hidden for other 22/23 options

### DIFF
--- a/app/models/case_log.rb
+++ b/app/models/case_log.rb
@@ -468,9 +468,19 @@ private
   end
 
   def dynamically_not_required
-    previous_la_known_field = postcode_known? ? %w[previous_la_known] : []
-    tshortfall_field = tshortfall_unknown? ? %w[tshortfall] : []
-    previous_la_known_field + tshortfall_field
+    not_required = []
+    not_required << "previous_la_known" if postcode_known?
+    not_required << "tshortfall" if tshortfall_unknown?
+    not_required << "tenancylength" if tenancylength_optional?
+
+    not_required
+  end
+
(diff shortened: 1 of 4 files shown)


app/models/case_log.rb
```diff
@@ -461,9 +468,19 @@ def reset_invalidated_dependent_fields!
...
+    not_required << "tenancylength" if tenancylength_optional?
+
+    not_required
+  end
+
+  def tenancylength_optional?
+    return false unless collection_start_year
+    return true if collection_start_year < 2022
+
+    collection_start_year >= 2022 && ![4, 6].include?(tenancy)
```
dushan-madetech: Could you not call `!is_secure_tenancy?` here instead of `![4, 6].include?(tenancy)`

dushan-madetech (approved):
Left a comment but nothing to block merging this 🙂

app/models/case_log.rb
baarkerlounger: Secure tenancy is 6 and 7. This is fixed term, I'll make a helper.

PR #580, opened 2022-05-17
Confirmable

baarkerlounger:
Devise has the concepts of:

- Registerable (Sign yourself up for a service - we don't use this)
- Confirmable (Confirm your email address is valid)
- Password (Create/reset a password)

These map to controllers. We did not originally use confirmable, and then bolted that functionality onto the password reset controller method. That worked but becomes an increasingly poor abstraction as their use cases diverge (for example different content, different link token timeouts etc).

This PR moves us onto using `confirmable` for onboarding new users which is closer to intended upstream behaviour.

- A confirmation email is automatically sent any time a user is created, including from console. This means the rake task is no longer needed as the users created by the import tasks will automatically have an email sent. Users migrated from the old system get the onboarding email template, will users invited from the new system directly get the invite confirmation template.

- When an account is confirmed, if the user has not logged in before they are redirected to set a password, and then logged in after setting one. If they have previously logged in, the assumption is they need to reconfirm because they've changed their email address but do not need to set a password again.

- Confirmation links expire after 3 days

- Password reset links continue to expire after 3 hours

Regular invite email template:
image

Beta onboarding email template:
image

Clicking onboarding link takes you to set your password:
image

If your link has expired you are shown this page:
image

Note. when this is deployed all existing users will be set to uncomfirmed. They will need updating when released.

User.all.each { |u| u.confirmed_at = Time.zone.now; u.save! }


paulrobertlloyd:
Not sure you need the confirmation banner on that last image. We might want to add a brief paragraph below the title saying something along the lines of ‘You need to create a password to complete the creation of your account’ (but something better than that). @Dan13L3

Dan13L3:
@baarkerlounger @paulrobertlloyd How about 'You'll need to create a password to finish setting up your account'?

baarkerlounger:
@Dan13L3 Maybe just "Create a password to finish setting up your account"?

Dan13L3:
@baarkerlounger Sure, works for me

paulrobertlloyd (changes requested):
Got quite a few questions on this, but general feedback before getting into specifics:

* We’re using ‘confirm email’ (which I guess is the out of box localisation for Devise) but in our case are we not using this more like ‘complete the set up of your account’?
* We’re using error summaries and validation messages where we should instead be using notification banners
* As such, if you try and sign in and need to ‘confirm’ your email, we should redirect you to the resend invitation link page with a notification banner, rather than show an error summary.
* Same for if you click on an invitation link that’s expired
* We probably don’t (and shouldn’t) have the link to resend an invitation link on the sign in page… I’m not sure it’s necessary, given links are only ever sent by data coordinators, so if you clicked that link and tried to get an invite link with an email that wasn’t invited, you’d get… nothing?

baarkerlounger:
@paulrobertlloyd what do you think about merging this (since it doesn't make anything worse, and at least ups the link expiry from 3 hours to 3 days), and then adding a design and separate PR for coordinators to be able to retrigger the invite email for users that haven't clicked the link?

paulrobertlloyd:
> @paulrobertlloyd what do you think about merging this (since it doesn't make anything worse, and at least ups the link expiry from 3 hours to 3 days), and then adding a design and separate PR for coordinators to be able to retrigger the invite email for users that haven't clicked the link?

Sounds good to me!

paulrobertlloyd (approved):

spec/requests/users_controller_spec.rb
```diff
@@ -1026,7 +1026,7 @@
               it "allows changing email, dpo, key_contact" do
                 patch "/users/#{other_user.id}", headers: headers, params: params
                 other_user.reload
-                expect(other_user.email).to eq(new_email)
+                expect(other_user.unconfirmed_email).to eq(new_email)
```
kosiakkatrina, line 1029: Do we set email to == unconfirmed_email after it's been confirmed? Is that tested? Also, do we set unconfirmed email to nil after it's been confirmed?

baarkerlounger: Devise does yes. This only applies when you're "reconfirming", i.e. if you've changed your email address, it doesn't get used on first creation.

baarkerlounger: https://github.com/heartcombo/devise/blob/f8d1ea90bc328012f178b8a6616a89b73f2546a4/lib/devise/models/confirmable.rb#L91

kosiakkatrina (approved):

PR #581, opened 2022-05-17
Export improvements

Ted-U:
Added exception handling, a message will be seen if there is an error and the data will not be saved.

### DIFF
--- a/app/services/exports/case_log_export_service.rb
+++ b/app/services/exports/case_log_export_service.rb
@@ -6,8 +6,9 @@ module Exports
     end
 
     def export_case_logs
-      case_logs = retrieve_case_logs
-      export = save_export_run
+      current_time = Time.zone.now
+      case_logs = retrieve_case_logs(current_time)
+      export = save_export_run(current_time)
       write_master_manifest(export)
       write_export_data(case_logs)
     end
@@ -22,13 +23,14 @@ module Exports
 
   private
 
-    def save_export_run
+    def save_export_run(current_time)
       today = Time.zone.today
       last_daily_run_number = LogsExport.where(created_at: today.beginning_of_day..today.end_of_day).maximum(:daily_run_number)
       last_daily_run_number = 0 if last_daily_run_number.nil?
 
       export = LogsExport.new
       export.daily_run_number = last_daily_run_number + 1
+      export.started_at = current_time
       export.save!
       export
     end
@@ -49,9 +51,15 @@ module Exports
       @storage_service.write_file(file_path, string_io)
     end
 
-    def retrieve_case_logs
-      params = { from: Time.current.beginning_of_day, to: Time.current, status: CaseLog.statuses[:completed] }
-      CaseLog.where("updated_at >= :from and updated_at <= :to and status = :status", params)
+    def retrieve_case_logs(current_time)
+      recent_export = LogsExport.order("started_at").last
+      if recent_export
+        params = { from: recent_export.started_at, to: current_time, status: CaseLog.statuses[:completed] }
+        CaseLog.where("updated_at >= :from and updated_at <= :to and status = :status", params)
+      else
+        params = { to: current_time, status: CaseLog.statuses[:completed] }
+        CaseLog.where("updated_at <= :to and status = :status", params)
+      end
     end
 
     def build_manifest_csv_io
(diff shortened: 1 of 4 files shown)


app/services/exports/case_log_export_service.rb
```diff
@@ -6,10 +6,14 @@ def initialize(storage_service, logger = Rails.logger)
...
     def export_case_logs
-      case_logs = retrieve_case_logs
-      export = save_export_run
+      current_time = Time.zone.now
+      case_logs = retrieve_case_logs(current_time)
+      export = save_export_run(current_time)
       write_master_manifest(export)
       write_export_data(case_logs)
+      export.save!
+    rescue StandardError => e
```
baarkerlounger: Wonder whether letting this go straight to Sentry might be more useful than just logging since it's likely to mean a run has totally failed and we'd want that to be noisy?

kosiakkatrina: Would we not have a rescue block at all then?

spec/services/exports/case_log_export_service_spec.rb
```diff
@@ -46,10 +46,17 @@ def replace_entity_ids(export_template)
     end
 
     context "and case logs are available for export" do
+      let!(:time_now) { Time.zone.now }
```
baarkerlounger: This can just be `let`? I think, no need for the bang since you're freezing it in a before do?

spec/services/exports/case_log_export_service_spec.rb
```diff
@@ -63,6 +70,30 @@ def replace_entity_ids(export_template)
...
+        expect(records_from_db.count).to eq(1)
+      end
+
+      it "gets the logs for correct timeframe" do
+        start_time = Time.zone.local(2022, 4, 13, 2, 2, 2)
+        export = LogsExport.new(started_at: start_time, daily_run_number: 1)
+        export.save!
+        params = { from: start_time, to: time_now }
+        allow(CaseLog).to receive(:where).with("updated_at >= :from and updated_at <= :to", params).and_return([])
+        export_service.export_case_logs
```
baarkerlounger, line 87: Is this expecting anything? Does it ever fail?

spec/services/exports/case_log_export_service_spec.rb
```diff
@@ -63,6 +70,30 @@ def replace_entity_ids(export_template)
...
+      end
+
+      context "when this is the first export" do
+        it "gets the logs for the timeframe up until the current time" do
+          params = { to: time_now }
+          allow(CaseLog).to receive(:where).with("updated_at <= :to", params).and_return([])
```
baarkerlounger: Shouldn't there be something like:

```ruby 
allow(CaseLog).to receive(:where).with("updated_at <= :to", params).and_return([])
expect(CaseLog).to receive(:where).with("updated_at <= :to", params).exactly(1).times
```

spec/services/exports/case_log_export_service_spec.rb
kosiakkatrina: Ah, good point!

spec/services/exports/case_log_export_service_spec.rb
kosiakkatrina: Right, so the `allow` is an expect in this case, rubicon (rubicon :)))))) thanks autocorrect rubocop*) prefers that to `expect`. It does fail if we change the parameters in `with`

spec/services/exports/case_log_export_service_spec.rb
kosiakkatrina: Hm, so it does fail already if it receives the request 0 times, not sure if it'd fail if it received it more times tho

app/services/exports/case_log_export_service.rb
baarkerlounger: I don't think so

baarkerlounger: You could trigger Sentry manually within the rescue block if you wanted to also log it, but not sure there's much benefit?

baarkerlounger (approved):

spec/services/exports/case_log_export_service_spec.rb
dushan-madetech: Did you run rubocop without `bundle exec` before it?

PR #582, opened 2022-05-17
Fix organisation filter

baarkerlounger:
image

Scenario:

1. Select "Specific Organisation"
2. Type and select an organisation
3. Click "Apply Filters"
4. Select "All"
5. Click "Apply Filters"

Conditional (nested) questions don't automatically get their values cleared when they're hidden so this was still showing the specific selection.

- [x] Fix the filters on the backend so that if both values are present we filter correctly
- [x] Clear the conditional value on hide in JS

### DIFF
--- a/app/frontend/controllers/conditional_filter_controller.js
+++ b/app/frontend/controllers/conditional_filter_controller.js
@@ -1,0 +1,13 @@
+import { Controller } from "@hotwired/stimulus"
+
+export default class extends Controller {
+  initialize() {
+    this.clearIfHidden()
+  }
+
+  clearIfHidden() {
+    if(this.element.style["display"] == "none") {
+      this.element.value = ""
+    }
+  }
+}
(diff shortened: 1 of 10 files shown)


app/frontend/controllers/conditional_filter_controller.js
```diff
@@ -0,0 +1,13 @@
...
+export default class extends Controller {
+  initialize() {
+    this.clearIfHidden()
+  }
+
+  clearIfHidden() {
+    if(this.element.style["display"] == "none") {
+      this.element.value = ""
+    }
+  }
```
dushan-madetech, line 12: should this have a unit test for the js?

baarkerlounger: Yep, that's what this test is for https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/582/files#diff-d319a8fedb48067418a804668bdd086508689cb5ec51d54525788f418c32c705R529

dushan-madetech (approved):

PR #583, opened 2022-05-18
User management tweaks

paulrobertlloyd:
Low hanging and easy to tweak improvements to user management screens:

* Show if a user is a key contact/dpo in list of users:

  Screenshot 2022-05-18 at 13 04 30

* Add `.app-tag--small` modifier
* Use consistent label size for adding/editing a user
* Add missing hint describing what a key contact is
* Fix heading when editing a user who does not have name

I’d like to have added hint text describing each of the different roles in this PR, but not sure I can do that.


baarkerlounger:
Needs linting but like the design :+1:

PR #585, opened 2022-05-18
CLDC-1210: Add status tag helper

paulrobertlloyd:
We use statuses in a number of places, and we should be consistent with their naming and colouring (right now we are not).

This PR consolidates status text and colours, and provides a helper for generating tags with the correct attributes. Additionally, a custom class can be supplied.


smeny (approved):
Love the change 🤩

PR #586, opened 2022-05-18
CLDC-1217: Retirement soft validation

baarkerlounger:
https://digital.dclg.gov.uk/jira/browse/CLDC-1217

- [x] Don't trigger soft validation screen if tenant prefers not to say for economic status
- [x] Update content for gender

image
image

### DIFF
--- a/app/models/case_log.rb
+++ b/app/models/case_log.rb
@@ -409,9 +409,9 @@ class CaseLog < ApplicationRecord
     return unless gender
 
     if %w[M X].include?(gender)
-      "men and non-binary people"
+      "male and non-binary people"
     elsif gender == "F"
-      "women"
+      "females"
     end
   end
 
--- a/spec/models/validations/soft_validations_spec.rb
+++ b/spec/models/validations/soft_validations_spec.rb
@@ -70,4 +70,98 @@ RSpec.describe Validations::SoftValidations do
       end
     end
   end
+
+  describe "retirement soft validations" do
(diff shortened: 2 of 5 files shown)


spec/models/validations/soft_validations_spec.rb
```diff
@@ -70,4 +70,98 @@
...
+
+      context "when the tenant is non-binary" do
+        it "shows the interruption screen" do
+          record.update!(age1: 85, sex1: "X", ecstat1: 10)
+          expect(record.person_1_not_retired_over_soft_max_age?).to be false
+        end
+      end
+    end
+  end
 end
```
dushan-madetech: Should these all say `it "does not show the interruption screen" do` based on the PR description?

app/models/case_log.rb
```diff
@@ -409,9 +409,9 @@ def plural_gender_for_person(person_num)
     return unless gender
 
     if %w[M X].include?(gender)
-      "men and non-binary people"
+      "male and non-binary people"
```
dushan-madetech, line 412: Would it not be `males` instead?

dushan-madetech (commented):
Left some comments

app/models/case_log.rb
baarkerlounger: Not based on the text in the screen (since people is plural as opposed to the genders themselves)

dushan-madetech (approved):

PR #587, opened 2022-05-18
CLDC-1118: Implement data export structure (CDS)

smeny:


app/services/exports/case_log_export_service.rb
```diff
@@ -20,57 +31,123 @@ def is_omitted_field?(field_name)
...
       @storage_service.write_file(file_path, string_io)
     end
 
-    def write_export_data(case_logs)
-      string_io = build_export_xml_io(case_logs)
-      file_path = "#{get_folder_name}/#{get_file_name}.xml"
-      @storage_service.write_file(file_path, string_io)
+    def get_archive_name(case_log, base_number, increment)
+      collection_start = case_log.collection_start_year
+      month = case_log.startdate.month
```
smeny, line 73: We might have to skip any log where we do not have a startdate (collection year will be nil too)

baarkerlounger (approved):

app/services/exports/case_log_export_service.rb
baarkerlounger: That's probably fine, if we don't have that we really don't have any meaningful data yet anyway.

PR #594, opened 2022-05-20
No two factor authentication for development environment

moarpheus:
Currently the support user will need to confirm their login using 2 factor authentication. While it is a required step in production this is an unnecessary hurdle for developer. 

This small amendment will allow **any type of user** in development environment to login without 2fa.

Changes - test and code for the user model.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -90,6 +90,7 @@ class User < ApplicationRecord
   end
 
   def need_two_factor_authentication?(_request)
+    return false if Rails.env.development?
     support?
   end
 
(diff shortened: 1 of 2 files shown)


kosiakkatrina (approved):

kosiakkatrina (commented):
Perhaps we want to remove all the instances of        
 `allow(support_user).to receive(:need_two_factor_authentication?).and_return(false)`
 in the tests and also not request it for the test env? 🤔 Unless we're testing it, in which case it's fine as it is

moarpheus:
> Perhaps we want to remove all the instances of `allow(support_user).to receive(:need_two_factor_authentication?).and_return(false)` in the tests and also not request it for the test env? 🤔 Unless we're testing it, in which case it's fine as it is

This opens up possibility to add new specs/code that would pass the tests, but fail or introduce unexpected behaviour when deployed to prod ( unless you are aware of this ). I would keep single explicit test for devs environment and leave the rest as is.

PR #595, opened 2022-05-20
Try different Sentry custom instrumentation

smeny:


smeny:
Will test this locally

PR #599, opened 2022-05-23
Cucumber rails demo

moarpheus:
Implemented cucumber rails and screenshots - moved couple of features to demonstrate.


features/support/env.rb
```diff
@@ -0,0 +1,22 @@
+require 'cucumber/rails'
+require "capybara-screenshot/cucumber"
+
+ActionController::Base.allow_rescue = false
```
baarkerlounger, line 4: What is this for?

features/support/env.rb
```diff
@@ -0,0 +1,22 @@
...
+
+begin
```
baarkerlounger, line 6: Do we need database cleaner? Doesn't look like it's required for any existing tests?

features/sign_in.feature
```diff
@@ -0,0 +1,8 @@
+Feature: Sign in
```
baarkerlounger, line 1: Should these not be in `/spec/features`?

config/environments/development.rb
```diff
@@ -1,6 +1,10 @@
 require "active_support/core_ext/integer/time"
 
 Rails.application.configure do
+  # Configure 'rails notes' to inspect Cucumber files
+  config.annotations.register_directories('features')
```
baarkerlounger: Why do we modify the dev environment?

spec/support/factory_bot.rb
```diff
@@ -0,0 +1,3 @@
+RSpec.configure do |config|
+  config.include FactoryBot::Syntax::Methods
```
baarkerlounger, line 2: We already have this in `/spec/spec_helper.rb`.

spec/rails_helper.rb
```diff
@@ -9,6 +9,8 @@
 require "selenium-webdriver"
 require "view_component/test_helpers"
 
+Dir[Rails.root.join('spec/support/**/*.rb')].each { |f| require f }
```
baarkerlounger: If we're adding this here we should probably remove the explicit requires from individual specs that use this?

Gemfile
```diff
@@ -87,7 +87,10 @@ end
 
 group :test do
   gem "capybara", require: false
+  gem "capybara-screenshot"
```
baarkerlounger: Am I right in thinking this provides plain HTML only? No CSS?

moarpheus:
I think that much should illustrate cucumber, feel free to try it out.

baarkerlounger:
@moarpheus given the number of conflicts and that we're probably not prioritising this right now - any objection to closing this for now and coming back to it another time if wanted?

moarpheus:
> @moarpheus given the number of conflicts and that we're probably not prioritising this right now - any objection to closing this for now and coming back to it another time if wanted?

Of course, I wanted to present this, but too busy with doing dev work. This can wait until less hactic.

PR #600, opened 2022-05-23
CLDC-1124: User search

baarkerlounger:
image

image

### DIFF
--- a/app/views/users/index.html.erb
+++ b/app/views/users/index.html.erb
@@ -8,23 +8,7 @@
   <%= govuk_button_link_to "Invite user", new_user_path, html: { method: :get } %>
 <% end %>
 
-<% path = current_user.support? ? users_path : users_organisation_path(current_user.organisation) %>
-<%= form_with model: @user, url: path, method: "get", local: true do |f| %>
-<div class="app-search govuk-!-margin-bottom-4">
-  <div class="govuk-form-group app-search__form-group">
-    <label class="govuk-label govuk-!-margin-bottom-2" for="user-search-field">
-      Search by name or email address
-    </label>
-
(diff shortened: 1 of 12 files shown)


app/views/users/index.html.erb
```diff
@@ -7,6 +7,25 @@
...
+    </label>
+
+    <input class="govuk-input app-search__input" id="user-search-field" name="user-search-field" type="search">
+  </div>
+
+  <button class="govuk-button app-search__button undefined" data-module="govuk-button">
+    Search
+  </button>
+
+</div>
```
paulrobertlloyd: Might be nice to make this a component (perhaps in a later PR).

app/views/users/index.html.erb
```diff
@@ -7,11 +7,30 @@
...
+<% end %>
+
 <%= govuk_table do |table| %>
   <%= table.caption(size: "s", classes: %w[govuk-!-text-align-left govuk-!-margin-top-4 govuk-!-margin-bottom-4]) do |caption| %>
     <span class="govuk-!-margin-right-4">
-      <strong><%= @pagy.count %></strong><span style="font-weight: normal"> total users</span>
-    </span>
+      <strong><span style="font-weight: normal"> Matches </span><%= @pagy.count %><span style="font-weight: normal"> of </span><%= User.count%> <span style="font-weight: normal"> total users</strong> </span>
```
baarkerlounger, line 32: I think we probably only want to do this if we have a search param? If not we probably want to just leave the original @pagy.count.

moarpheus (approved):
Looks good, except maybe module without a unit test, but is covered in request specs.

PR #602, opened 2022-05-24
Cldc 1223 pregnancy soft validations

kosiakkatrina:
Update hard validation pregnancy age from 16-50 to 11-65
Remove hard validation if there are no women in the household
Add soft validations for when there are no women in a pregnant household or there are only women in the age ranges of 11-16/50-65

image

image

### DIFF
--- a/app/models/validations/household_validations.rb
+++ b/app/models/validations/household_validations.rb
@@ -36,7 +36,7 @@ module Validations::HouseholdValidations
   end
 
   def validate_pregnancy(record)
-    if (record.has_pregnancy? || record.pregnancy_refused?) && !women_of_child_bearing_age_in_household(record)
+    if (record.has_pregnancy? || record.pregnancy_refused?) && women_in_household(record) && !women_of_child_bearing_age_in_household(record)
       record.errors.add :preg_occ, I18n.t("validations.household.preg_occ.no_female")
     end
   end
@@ -136,7 +136,13 @@ private
     (1..8).any? do |n|
       next if record["sex#{n}"].nil? || record["age#{n}"].nil?
 
-      (record["sex#{n}"]) == "F" && record["age#{n}"] >= 16 && record["age#{n}"] <= 50
+      (record["sex#{n}"]) == "F" && record["age#{n}"] >= 11 && record["age#{n}"] <= 65
+    end
+  end
+
+  def women_in_household(record)
+    (1..8).any? do |n|
+      record["sex#{n}"] == "F"
     end
   end
 
--- a/app/models/validations/soft_validations.rb
+++ b/app/models/validations/soft_validations.rb
@@ -48,7 +48,23 @@ module Validations::SoftValidations
     end
   end
 
+  def no_females_in_the_household?
+    (1..8).none? do |n|
+      public_send("sex#{n}") == "F"
+    end && preg_occ == 1
+  end
+
+  def female_in_pregnant_household_in_soft_validation_range?
+    (females_in_age_range(11, 15) || females_in_age_range(51, 65)) && !females_in_age_range(16, 50)
+  end
+
 private
+
+  def females_in_age_range(min, max)
+    (1..8).any? do |n|
+      public_send("sex#{n}") == "F" && public_send("age#{n}").present? && public_send("age#{n}").between?(min, max)
+    end
+  end
 
   def tenant_is_retired?(economic_status)
     economic_status == 5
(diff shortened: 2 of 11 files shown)


Gemfile.lock
```diff
@@ -264,11 +264,11 @@ GEM
       net-protocol
       timeout
     nio4r (2.5.8)
-    nokogiri (1.13.5-arm64-darwin)
```
baarkerlounger: Think this is already in main if you rebase?

app/models/validations/household_validations.rb
```diff
@@ -36,7 +36,7 @@ def validate_armed_forces(record)
   end
 
   def validate_pregnancy(record)
-    if (record.has_pregnancy? || record.pregnancy_refused?) && !women_of_child_bearing_age_in_household(record)
+    if (record.has_pregnancy? || record.pregnancy_refused?) && women_in_household(record) && !women_of_child_bearing_age_in_household(record)
```
baarkerlounger, line 39: Why do we need to add this condition? Isn't it already covered by the next method?

config/locales/en.yml
```diff
@@ -157,7 +157,7 @@ en:
         question_required: "You must answer whether the person is still serving in the UK armed forces as you told us they’re a current or former regular"
         question_not_required: "You cannot answer whether the person is still serving in the UK armed forces as you told us they’re not a current or former regular"
       preg_occ:
-        no_female: "You must answer ‘no’ as there are no females aged 16-50 in the household"
+        no_female: "You must answer ‘no’ as there are no women aged 11-65 in the household"
```
baarkerlounger: @Dan13L3 should this be females? Think we updated everywhere else but I suppose here precision about gender vs sex is important?

baarkerlounger:
I think we also need to add the validation screen if you answer preg_occ first and then change a tenant gender or age question such that preg_occ needs confirming again?

config/locales/en.yml
Dan13L3: @baarkerlounger I think we should go with 'females' as that's how we're phrasing the gender question currently - though if we end up revisiting the gender issue overall, then it might need to change

kosiakkatrina: it's women in the soft validations, should I change that too?
 ```
'You told us somebody in the household is pregnant [h1]
You also told us there are no women living at the property.
[or]
You also told us that any women living at the property are in the following age ranges:
11 to 16
50 to 65
Are you sure this is correct?
```

File: app/models/validations/household_validations.rb (hunk 2 above)
Comment (kosiakkatrina) [reply app/models/validations/household_validations.rb:39]: Because if there are no women in the household we don't want to throw an error. If there are no women `!women_of_child_bearing_age_in_household` will return true and the error will get thrown

File: Gemfile.lock (hunk 1 above)
Comment (kosiakkatrina) [reply Gemfile.lock:267]: > I think we also need to add the validation screen if you answer preg_occ first and then change a tenant gender or age question such that preg_occ needs confirming again?

@baarkerlounger from what I gather from the discussion in the ticket, might be worth deferring that part until a further discussion

**Comment by kosiakkatrina** [comment]
> I think we also need to add the validation screen if you answer preg_occ first and then change a tenant gender or age question such that preg_occ needs confirming again?

@baarkerlounger from what I gather from the discussion in the ticket, might be worth deferring that part until a further discussion

**Comment by baarkerlounger** [comment]
@kosiakkatrina I've commented on the ticket as well but I think that was already agreed previously. Kind of has to happen on both sides otherwise you can just bypass the confirmation by answering the questions in reverse, which can't be right.

File: app/models/validations/soft_validations.rb [hunk 4]
```diff
@@ -48,8 +48,24 @@ def rent_in_soft_max_range?
     end
   end
 
+  def no_females_in_the_household?
```
smeny: following the previous discussion should this be women as well? (might not be worth renaming everything)

baarkerlounger: I think we're going with `female` everywhere for now

spec/models/validations/soft_validations_spec.rb
```diff
@@ -164,4 +164,43 @@
...
+    end
+
+    context "when female tenants are in 50-65 age range" do
+      it "shows the interruption screen" do
+        record.update!(age1: 54, sex1: "F", preg_occ: 1, hhmemb: 1, age1_known: 0)
+        expect(record.female_in_pregnant_household_in_soft_validation_range?).to be true
+      end
+    end
+
+    context "when female tenants are outside or soft validation ranges" do
```
baarkerlounger: Typo on `of`?

baarkerlounger (approved):

PR #603, opened 2022-05-24
Email allowlist

baarkerlounger:
This would ideally be configurable outside of a code change but this is a first pass implementation to enable us to continue doing dummy migration runs.


kosiakkatrina (approved):

app/mailers/devise_notify_mailer.rb
```diff
@@ -1,11 +1,15 @@
 class DeviseNotifyMailer < Devise::Mailer
   require "notifications/client"
 
+  ALLOWLIST = %w[madetech.com levellingup.gov.uk communities.gov.uk].freeze
```
dushan-madetech: Does defining this list in the code introduce any potential security risks?

smeny: It does give away which domains we allow, but I don't think that sharing this information increases the attack surface or facilitates a potential existing exploit (we already rate limit).

smeny (approved):

app/mailers/devise_notify_mailer.rb
moarpheus: It exposes domains, probably not desirable. Simple thing would be to store those values in secrets e.g. credentials.yml which is encoded.

app/mailers/devise_notify_mailer.rb
```diff
@@ -41,6 +45,11 @@ def confirmation_instructions(record, token, _opts = {})
     )
   end
 
+  def intercept_send?(email)
+    email_domain = email.split("@").last.downcase
```
moarpheus, line 49: I inverted logic to `Rails.env.staging? && ALLOWLIST.exclude?(email_domain)` and got tests passing and probably easier to read as well

app/mailers/devise_notify_mailer.rb
```diff
@@ -41,6 +45,11 @@ def confirmation_instructions(record, token, _opts = {})
...
+    !(Rails.env.production? || Rails.env.test?) && !ALLOWLIST.include?(email_domain)
```
moarpheus: I inverted logic to Rails.env.staging? && ALLOWLIST.exclude?(email_domain) and got tests passing and probably easier to read as well

baarkerlounger: My reasoning for this way round was that if another environment was added (sandbox for example) you probably want that to default to the block behaviour rather than the allow behaviour?

Will change the include to exclude though :+1: .

moarpheus: That I did not see and it makes sense!

PR #607, opened 2022-05-24
User search fixes

baarkerlounger:
- [x] Update query result caption
- [x] Add clear search link
- [x] Search input should show search term
- [x] Update page title

image


paulrobertlloyd:
Lovely stuff. Added a missing horizontal rule, and used the `form_group` params on the search input. (Think we should keep the horizontal break outside the confines of the component, as its usage is context dependent… or at least could be).

paulrobertlloyd (approved):

PR #608, opened 2022-05-25
Cldc 1123 search logs

moarpheus:


app/components/log_search_component.html.erb
```diff
@@ -0,0 +1,16 @@
+<%= form_with model: @case_logs, url: "/logs", method: "get", local: true do |f| %>
```
baarkerlounger: This can probably just use the generic search component?

app/components/log_search_component.rb
```diff
@@ -0,0 +1,9 @@
+class LogSearchComponent < ViewComponent::Base
```
baarkerlounger: Same here

app/controllers/case_logs_controller.rb
```diff
@@ -12,6 +12,12 @@ def index
 
     @pagy, @case_logs = pagy(filtered_case_logs(current_user.case_logs))
 
+    search_param = params["search-field"]
```
baarkerlounger: If you use the generic component this is just `search` not `search-field` I think

app/controllers/case_logs_controller.rb
```diff
@@ -12,6 +12,12 @@ def index
...
+
+    if search_param
+      @pagy, @case_logs = pagy(@case_logs.search_by(search_param))
```
baarkerlounger: Think this probably needs to include `filtered_case_logs()` and `filtered_collection()`. Something like:

```
@pagy, @case_logs = pagy(filtered_case_logs(filtered_collection(current_user.case_logs, search_param)))`
```

baarkerlounger: Where `filtered_collection` is doing the `search_by` or returning all if no search term, and then `filtered_case_logs` is applying the filters.

app/models/case_log.rb
```diff
@@ -51,6 +51,10 @@ class CaseLog < ApplicationRecord
                            end
                          }
 
+  scope :search_by_id, ->(id) { where(id: id) }
+  scope :search_by_tenancy_code, ->(code) { where(tenancy_code: code) }
```
baarkerlounger: I was thinking a good naming convention would be `filter_by_x` for something that expects to match exactly like these, and `search_by` for something matching less exactly i.e. `LIKE` or some other fuzzy match.

app/components/log_search_component.html.erb
moarpheus: WIP still will be refactored to use generic one.

app/components/log_search_component.rb
moarpheus: WIP still will be refactored to use generic one.

app/controllers/case_logs_controller.rb
moarpheus: WIP still will be refactored to use generic one.

app/controllers/case_logs_controller.rb
moarpheus: WIP this will be refactored and expanded and will live in a separate module

app/models/case_log.rb
moarpheus: Good spot, I went for the opposite - will amend!

app/controllers/case_logs_controller.rb
moarpheus: Done

app/views/case_logs/index.html.erb
```diff
@@ -1,8 +1,21 @@
+<% item_label = @pagy.count > 1 ? "logs" : "log" %>
+
+<% if @searched.present? %>
```
baarkerlounger, line 3: I don't think this will do anything unless you pass it to `content_for :title`?

moarpheus: Indeed and we use it on line 9 I think

moarpheus: We use item_label in log_list

baarkerlounger: Not the `item_label` the `title`. We're setting it here but not actually passing it to `content_for :title` on line 9, which still just has a hardcoded string.

I'm also not sure the title is correct for all scenarios here.

It should probably be:

1. No search term = "Logs"
2. Search term = Logs ((#{@pagy.count} #{item_label} matching ‘#{@searched}’ of #{@total_count} total logs)

moarpheus: Gotcha

spec/factories/case_log.rb
```diff
@@ -52,11 +52,11 @@
       reservist { 0 }
       illness { 1 }
       preg_occ { 2 }
-      tenancy_code { "BZ757" }
+      tenancy_code { Faker::Alphanumeric.alphanumeric(number: 10) }
```
smeny: Tenancy code is not used in our current form, I think it should be tenant code (this is already used and populated in production).

Note: Having a random value will break all the export logic which expects pre-determined values. We will need to think of a better strategy to make this work.

moarpheus: _I can find a log by entering a tenancy code_ - this is from requirement on the card. I made change to the test and it passes it.

app/models/case_log.rb
```diff
@@ -51,6 +51,17 @@ class CaseLog < ApplicationRecord
                            end
                          }
 
+  scope :filter_by_id, ->(id) { where(id:) }
+  scope :filter_by_tenancy_code, ->(code) { where(tenancy_code: code) }
```
baarkerlounger: I think we want these to be case insensitive (they are for searching user by name/email, and org by name)

baarkerlounger:
Search bar also needs implementing for `organisation/:id/logs`

baarkerlounger (approved):
Couple of things left here:

1. CSV download doesn't respect search results (still downloads all rather than just matching values)
2. Title doesn't include the matching count which is a little inconsistent with Org and User page
3. I wonder whether search should accept partial matches (LIKE rather than EQ) similar to user and org
4. Also 1 test still failing

Aside from the failing test happy to merge this and fix those separately though (please squash and merge)

PR #610, opened 2022-05-25
Organisation search

baarkerlounger:
image
image


spec/requests/users_controller_spec.rb
```diff
@@ -381,7 +381,7 @@
           end
 
           it "updates the table caption" do
-            expect(page).to have_content("1 user found matching ‘filter’ of 5 total users.")
+            expect(page).to have_content("1 user found matching ‘filter’ of 4 total users.")
```
baarkerlounger, line 384: This is changing because previously we were incorrectly looking at the size of the entire User collection instead of the Organisation's user collection.

dushan-madetech (approved):

PR #611, opened 2022-05-25
Remove active admin

baarkerlounger:
Remove ActiveAdmin and AdminUser class since we're building support journeys into the app itself and are no longer expecting to use this.

Simplifies our dependency graph, and Devise user model.


dushan-madetech (approved):
A lot of deletions so giving this an approval. As long as the tests/linting passes 🚢  it

PR #612, opened 2022-05-25
CLDC-1272: Handles unicode characters in postcode

smeny:


app/models/case_log.rb
```diff
@@ -718,4 +723,8 @@ def soft_value_for_period(value)
 
     (value * 52 / num_of_weeks).round(2)
   end
+
+  def upcase_and_remove_whitespace(string)
+    string.present? ? string.upcase.gsub(/\s+/, "") : string
```
moarpheus, line 728: `string.upcase.gsub(/\s+/, "") || string` should do the trick too

moarpheus (approved):

app/models/case_log.rb
smeny: I like it but this wouldn't protect a call made with a nil value.

baarkerlounger:
Should we not rather swap the order so that Postcodes are validated before that method is called rather than validating twice?

spec/models/case_log_spec.rb
```diff
@@ -1996,5 +1996,13 @@ def previous_postcode_fields
         expect(case_log.plural_gender_for_person_3).to be_nil
       end
     end
+
+    context "when a postcode contains unicode characters" do
+      let(:case_log) { FactoryBot.build(:case_log, postcode_full: "SR81LS\u00A0") }
+
+      it "triggers a validation error" do
+        expect { case_log.save! }.to raise_error(ActiveRecord::RecordInvalid, /Enter a postcode in the correct format/)
```
baarkerlounger, line 2004: Probably want to be referencing a translation string here?

baarkerlounger:
Order is because we need the value set before we can be sure we can validate it

PR #613, opened 2022-05-25
Updates exported fields based on May 25th feedback

smeny:


app/services/exports/case_log_export_constants.rb
```diff
@@ -0,0 +1,138 @@
...
+    "benefits",
+    "brent",
+    "builtype",
+    "cap",
+    "cbl",
+    "chcharge",
+    "chr",
+    "cligrp1",
+    "cligrp2",
+    "createddate", # New metadata coming from our system
```
kosiakkatrina, line 24: If that's something that we define, maybe `created_date` would be better?

smeny: I would agree with you but the issue is that the counterpart is `uploadddate` which is already defined. CDS seems to avoid using underscores in general.

spec/services/exports/case_log_export_service_spec.rb
```diff
@@ -16,9 +16,8 @@
 
   def replace_entity_ids(case_log, export_template)
     export_template.sub!(/\{id\}/, (case_log["id"] + Exports::CaseLogExportService::LOG_ID_OFFSET).to_s)
-    export_template.sub!(/\{owning_org_id\}/, case_log["owning_organisation_id"].to_s)
-    export_template.sub!(/\{managing_org_id\}/, case_log["managing_organisation_id"].to_s)
-    export_template.sub!(/\{created_by_id\}/, case_log["created_by_id"].to_s)
+    export_template.sub!(/\{owning_org_id\}/, (case_log["owning_organisation_id"] + Exports::CaseLogExportService::LOG_ID_OFFSET).to_s)
```
kosiakkatrina, line 19: are `owning_org_id` and `case_log["owning_organisation_id"]` different values?

app/services/exports/case_log_export_constants.rb
kosiakkatrina: Hm, I might be misreading the comment, so this metadata is already expected in the export?

spec/services/exports/case_log_export_service_spec.rb
smeny: `{is owning_org_id}` is a placeholder inside our CSV and XML examples that we replace with the owning organisation ID we get when Factory Bot initialises the case log (and related organisations). This is because we cannot hardcode these IDs in advance as they are increasing.

app/services/exports/case_log_export_constants.rb
smeny: No no, it is not. I am following the existing convention. If we don't, we will end up with new fields using underscores and old ones concatenated.

I am not saying I find the convention great, but I think staying consistent is more important.

smeny: But you might argue there are already fields with underscores, so happy to go with what the team prefers tbh.

spec/services/exports/case_log_export_service_spec.rb
kosiakkatrina: Oh, of course 🤦‍♀️  tyty

app/services/exports/case_log_export_constants.rb
moarpheus: Tricky one, I would probably also follow the trend here, which looks like predominately there are composite names without underscores.

app/services/exports/case_log_export_service.rb
```diff
@@ -158,32 +150,62 @@ def build_manifest_xml(record_number)
...
+        attribute_hash["maningorgid"] = case_log.managing_organisation.old_visible_id || (case_log.managing_organisation.id + LOG_ID_OFFSET)
+        attribute_hash["maningorgname"] = case_log.managing_organisation.name
+        attribute_hash["manhcnum"] = case_log.managing_organisation.housing_registration_no
+      end
+
+      # Mapping which would require a change in our data model
+      attribute_hash["createddate"] = attribute_hash["created_at"]
+      attribute_hash["uploaddate"] = attribute_hash["updated_at"]
+      attribute_hash["tenancycode"] = attribute_hash["tenancy_code"]
+      attribute_hash["ppcodenk"] = attribute_hash["previous_postcode_known"]
```
moarpheus, line 182: Maybe this needs be `ppcodekn` oppose to `ppcodenk` ? Unless already agreed upon with the client?

baarkerlounger (approved):

app/services/exports/case_log_export_service.rb
smeny: `ppcodenk` is already existing.
I think it stands for "not known" (which is confusing with ours).

PR #615, opened 2022-05-26
CLDC-1277: no route matches bug

kosiakkatrina:
don't display the save and go to the next incomplete section button if it returns an error

### DIFF
--- a/app/views/form/check_answers.html.erb
+++ b/app/views/form/check_answers.html.erb
@@ -24,8 +24,9 @@
 
     <%= form_with model: @case_log, method: "get" do |f| %>
       <%= f.govuk_submit "Save and return to log" do %>
-        <% if @case_log.status == "in_progress" %>
-          <%= govuk_button_link_to "Save and go to next incomplete section", "/logs/#{@case_log.id}/#{@case_log.form.next_incomplete_section_redirect_path(subsection, @case_log)}", secondary: true %>
+        <% next_incomplete_section_redirect_path = @case_log.form.next_incomplete_section_redirect_path(subsection, @case_log) %>
+        <% if @case_log.status == "in_progress" && next_incomplete_section_redirect_path != "error" %>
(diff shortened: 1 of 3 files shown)


app/views/form/check_answers.html.erb
```diff
@@ -24,8 +24,9 @@
 
     <%= form_with model: @case_log, method: "get" do |f| %>
       <%= f.govuk_submit "Save and return to log" do %>
-        <% if @case_log.status == "in_progress" %>
-          <%= govuk_button_link_to "Save and go to next incomplete section", "/logs/#{@case_log.id}/#{@case_log.form.next_incomplete_section_redirect_path(subsection, @case_log)}", secondary: true %>
+        <% next_incomplete_section_redirect_path = @case_log.form.next_incomplete_section_redirect_path(subsection, @case_log) %>
+        <% if @case_log.status == "in_progress" && next_incomplete_section_redirect_path != "error" %>
```
baarkerlounger, line 28: Does the method actually return a string "error"?

spec/requests/form_controller_spec.rb
```diff
@@ -100,6 +109,16 @@
             expect(response).to have_http_status(:not_found)
           end
         end
+
+        context "when no other sections are enabled" do
+          before do
+            allow(in_progress_case_log.form).to receive(:next_incomplete_section_redirect_path).and_return("error")
```
baarkerlounger: Does this need mocking? Shouldn't just using a case log with only 1 field answered using the 22/23 fixture work?

app/views/form/check_answers.html.erb
kosiakkatrina: yup. Happy to change, if that's a good use of time

baarkerlounger: Ah I see it literally does, missed that. I think I'd expect `nil` returned but this is fine too.

spec/requests/form_controller_spec.rb
kosiakkatrina: done! Nice one

baarkerlounger (approved):

PR #616, opened 2022-05-26
Organisation Page Consistency

baarkerlounger:
Design before (in built product) is inconsistent. This PR attempts to make it consistent. It's possible that page title and page header actually need to be two separate things in the case of being on a nested resource page?

**Customer Support View:**

Before:
image
image

After:
image 
image

Coordinator view is unchanged


moarpheus (approved):
LGTM

kosiakkatrina (approved):
Makes sense from the consistency side

PR #619, opened 2022-05-27
Sort users by organisation and role

baarkerlounger:
Currently our user and organisation views don't sort by anything so the order is somewhat random.

- [x] Sort users by organisation name and then by role (when support is viewing all users)
- [x] Sort users by role (when viewed for a single organisation)
- [x] Sort organisations by name


moarpheus (approved):
LGTM

PR #620, opened 2022-05-27
Cldc 1220 better logging status discrepancies

dushan-madetech:
Logs the old id of a case log which was imported with a status discrepancy and returns the old ids of all case logs which have status discrepancies at the end of the run

### DIFF
--- a/app/services/imports/case_logs_import_service.rb
+++ b/app/services/imports/case_logs_import_service.rb
@@ -2,6 +2,9 @@ module Imports
   class CaseLogsImportService < ImportService
     def create_logs(folder)
       import_from(folder, :create_log)
+      if @logs_with_discrepancies.count > 0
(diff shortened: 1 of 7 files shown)


app/services/imports/case_logs_import_service.rb
```diff
@@ -226,6 +229,8 @@ def fields_not_present_in_softwire_data
     def check_status_completed(case_log, previous_status)
       if previous_status.include?("submitted") && case_log.status != "completed"
         @logger.warn "Case log #{case_log.id} is not completed"
+        @logger.warn "Case log with old id:#{case_log.old_id} is incomplete but status should be complete"
```
baarkerlounger, line 232: Do we need the line above this? Could log old id and new id in the same warning?

dushan-madetech: I just kept them separate but I could merge them into one warning if we want that 🙂

baarkerlounger (approved):

PR #624, opened 2022-05-30
Cldc 1226 deactivate users

kosiakkatrina:
- [x] An active user can be deactivated
- [x] Deactivation link shown is below page title on user detail page:
image

- [x] A confirmation screen is shown before a user can be deactivated:
image

- [x] A notification banner is shown on the user’s detail page if successfully deactivated
- [x] Message shown below page title on user detail page: This user has been deactivated
- [x] Details for a deactivated user cannot be changed
image

- [x] On the list of users, ‘Last logged in’ shows ‘Deactivated’ for a deactivated user 
image

- [ ] A deactivated user can be reactivated
- [x] A confirmation screen is shown before a user can be reactivated
image

- [x] A notification banner is shown on the user’s detail page if successfully reactivated
image

- [ ] The invitation email is resent allowing the user to be informed of their account being reactivated and gloving them an opportunity to reset their password


app/controllers/users_controller.rb
```diff
@@ -30,13 +30,25 @@ def index
...
   def update
     if @user.update(user_params)
       if @user == current_user
         bypass_sign_in @user
         flash[:notice] = I18n.t("devise.passwords.updated") if user_params.key?("password")
         redirect_to account_path
       else
+        case user_params[:active]
+        when "false"
+          @user.update!(confirmed_at: nil, sign_in_count: 0, encrypted_password: "")
```
baarkerlounger: Why do we set the password to empty string?

app/models/user.rb
```diff
@@ -73,13 +73,16 @@ def is_data_protection_officer!
...
   BETA_ONBOARDING_TEMPLATE_ID = "b48bc2cd-5887-4611-8296-d0ab3ed0e7fd".freeze
+  USER_REACTIVATED_TEMPLATE_ID = "ac45a899-490e-4f59-ae8d-1256fc0001f9".freeze
 
   def reset_password_notify_template
     RESET_PASSWORD_TEMPLATE_ID
   end
 
   def confirmable_template
-    if was_migrated_from_softwire?
+    if last_sign_in_at.present?
```
baarkerlounger: Which email are you now sent if you change your email address?

baarkerlounger (approved):
Couple of questions but looks good to me

baarkerlounger:
Should we be sorting and/or filtering deactivated users in any way? (maybe topic for a separate PR)

app/controllers/users_controller.rb
kosiakkatrina: So that after getting reactivated the user wouldn't be able to login with the old creds

app/models/user.rb
kosiakkatrina: A wrong one ><, got reactivation email to my current (not new) address. Lemme fix that. Tyty!

kosiakkatrina:
> Should we be sorting and/or filtering deactivated users in any way? (maybe topic for a separate PR)

@baarkerlounger Yup! Was thinking the same thing

app/models/user.rb
kosiakkatrina: Hmm, it seems like it's working wrong on staging. I get a `Confirm your account` email, but I get it to an old email, surely we'd want it to go to the updated email?

app/controllers/users_controller.rb
baarkerlounger: Wouldn't that already be achieved by the overriding of `valid_for_authentication?`. Also if that returned true wouldn't this let you log in without a password?

app/models/user.rb
baarkerlounger: Not 100% sure but probably to both...? i.e. if it's unexpected you'd want your old email to be contacted?

app/views/users/toggle_active.html.erb
```diff
@@ -0,0 +1,26 @@
...
+            </h1>
+            <% if action == "deactivate" %>
+                <p>Deactivating this user will mean they can no longer access this service to submit CORE data.</p>
+                <p>Any logs this user has already submitted will not be affected.</p>
+            <% end %>
+            <% active_value = action != "deactivate" %>
+            <div hidden>
+                <%= f.govuk_text_field :active,
+                value: active_value %>
+            </div>
```
paulrobertlloyd: We can use a `hidden` input type here. Hopefully this works:

```suggestion
            <%= f.govuk_text_field :active,
                value: active_value,
                type: "hidden" %>
```

app/models/user.rb
kosiakkatrina: Hm, perhaps. But making that mistake is dangerous to begin with. I'll change it to both tho.

paulrobertlloyd:
In the reactivation email, we say:
> An account has been created for you to submit CORE data on behalf of DLUHC
> …
> Click the link below to set up your password:

Does this copy need updating for reactivations? Something more along the lines of:

> Your account has been reactivated for you to submit CORE data on behalf of DLUHC
> …
> Click the link below to reset your password:

@Dan13L3

config/locales/devise.en.yml
```diff
@@ -55,8 +55,8 @@ en:
       send_paranoid_instructions: "If your account exists, you will receive an email in a few minutes with instructions for how to unlock it"
       unlocked: "Your account has been successfully unlocked. Sign in to continue."
     activation:
-      deactivated: "%{user_name} has been deactivated."
-      reactivated: "%{user_name} has been reactivated."
+      deactivated: "%{user_name}’s account has been deactivated."
+      reactivated: "%{user_name}’s account has been reactivated."
```
kosiakkatrina: Do we need to account for the name ending in `s` ?

baarkerlounger: Probably... `user_name.pluralize`?

Dan13L3:
@paulrobertlloyd Yep, I discussed this with @kosiakkatrina yesterday. We were going to go with:

'Your account for submitting social housing lettings and sales data on behalf of [organisation name] has been reactivated.

Your username is [email address].

Click the link below to create a new password:'

app/models/user.rb
kosiakkatrina: Maybe we should have something along the lines of: "If you think this is incorrect please ignore this email"

app/controllers/users_controller.rb
kosiakkatrina: Frontend (?) validation doesn't allow an empty password. The user would be valid for authentication at that point because they'd be reactivated (?)

app/models/user.rb
```diff
@@ -73,13 +73,16 @@ def is_data_protection_officer!
...
   def reset_password_notify_template
     RESET_PASSWORD_TEMPLATE_ID
   end
 
   def confirmable_template
-    if was_migrated_from_softwire?
+    if last_sign_in_at.present? && (unconfirmed_email.blank? || unconfirmed_email == email)
+      USER_REACTIVATED_TEMPLATE_ID
+    elsif was_migrated_from_softwire?
       BETA_ONBOARDING_TEMPLATE_ID
```
kosiakkatrina: This looks odd, would we not want to send CONFIRMABLE_TEMPLATE_ID if the email address changes instead of BETA_ONBOARDING_TEMPLATE_ID after the user have used the system and decided to change the email?

app/models/user.rb
baarkerlounger: I suppose the idea of sending to your old email is to prevent someone else (that has gained access to your account), changing the email without your knowledge and/or preventing you from locking yourself out if you've made a mistake in the email address you've changed it to.

I'm actually not really sure what we want to happen in each of those cases though so that maybe needs more thought... @paulrobertlloyd ?

app/views/users/toggle_active.html.erb
kosiakkatrina: Hm, I can't seem to get it to work like this for some reason. Getting the label displayed 
image

app/models/user.rb
baarkerlounger: I think the reason for this was because Softwire doesn't validate emails so we can't be sure they're correct. The anticipated flow was:

1. We import users
2. Some emails are incorrect so they don't get an email and contact their coordinator
3. Coordinator fixes their email

It probably needs an additional switch for if they've signed in before

app/controllers/users_controller.rb
baarkerlounger: Is that a problem? I thought resetting a users password was mostly for their convenience (if they've been deactivated for a while they may not remember it) rather than a need... (if so do we need to do this at all or can we just leave them to reset it if required?).

Gemfile
```diff
@@ -59,6 +59,7 @@ gem "sentry-rails"
 gem "sentry-ruby"
 # Pagination
 gem "pagy"
+gem "possessive"
```
baarkerlounger: Is this for a case that ActiveSupport `.pluralize` doesn't cover?

app/views/users/toggle_active.html.erb
paulrobertlloyd: Ah, of course the GOV.UK Form Builder includes the label. Updated to use the standard `hidden_field` helper from Rails instead:

```
<%= f.hidden_field :active, value: active_value %>
```

Gemfile
kosiakkatrina: pluralize pluralizes the word, so instead of `Baark's` we'd get `Baarks`

paulrobertlloyd: Guessing we don’t want to pluralize a user name, we just want to make sure we say either: Paul’s or James’ (not James’s)

app/views/users/toggle_active.html.erb
```diff
@@ -12,11 +12,11 @@
                 <p>Any logs this user has already submitted will not be affected.</p>
             <% end %>
             <% active_value = action != "deactivate" %>
-            <div hidden>
-                <%= f.govuk_text_field :active,
-                value: active_value %>
-            </div>
+
+            <%= f.hidden_field :active, value: active_value %>
```
kosiakkatrina, line 16: Ah sweet!

Gemfile
baarkerlounger: Makes sense

app/models/user.rb
kosiakkatrina: What if you don't have access to that email anymore? Which seems like the main use case to change it to begin with (I think)

app/controllers/users_controller.rb
kosiakkatrina: I mean if it's not required to reset the password, I'd happily remove that bit. Idk why I assumed it was

app/models/user.rb
kosiakkatrina: Separate ticket?

kosiakkatrina: Ah nvm, it's one if statement, I can do it

paulrobertlloyd (approved):

app/mailers/devise_notify_mailer.rb
```diff
@@ -36,10 +36,20 @@ def reset_password_instructions(record, token, _opts = {})
 
   def confirmation_instructions(record, token, _opts = {})
     url = "#{user_confirmation_url}?confirmation_token="
+
+    username = record.email
+    if record.confirmable_template == User::CONFIRMABLE_TEMPLATE_ID && (record.unconfirmed_email.present? && record.unconfirmed_email != record.email)
```
baarkerlounger: Is this intended to handle reconfirmations? Might be worth extracting out into it's own method?

kosiakkatrina: This is for email changes. Good point

kosiakkatrina: better now?

PR #625, opened 2022-05-30
Major repairs question should be valid if void date question is

baarkerlounger:
https://digital.dclg.gov.uk/jira/browse/CLDC-1282

Major repairs should be asked if void date is asked so this PR makes the depends on the same.

Once merged we will need to:

1. Update all log statuses to account for logs that didn't previously answer Major Repairs but now need to
2. Re-import the major repairs data for all imported logs
```ruby
CaseLog.all.each { |l| l.send(:update_status!); l.save! }
```
```bash
cf run-task dluhc-core-production --command "bundle exec rake core:data_import_field['major_repairs','2022-05-10/logs/']" --name believe-major-repairs-reimport
cf run-task dluhc-core-production --command "bundle exec rake core:data_import_field['major_repairs','2022-05-27-barrow/logs/']" --name barrow-major-repairs-reimport
```

Affected imported logs missing Major Repairs: Yes, Major Repairs Date:
```bash
$ grep -rn '<MRCDAY>' | wc -l
135
```

Affected imported logs missing Major Repairs: No
```bash
$ grep -rn '<MRCDAY/>' | wc -l
1799
```

Affected total logs missing Major Repairs
```ruby
CaseLog.where(renewal: 0, rsnvac: [5, 6, 8, 9, 10, 11, 12, 13, 18, 19], majorrepairs: nil).count
=> 1922
```

### DIFF
--- a/app/services/imports/case_logs_field_import_service.rb
+++ b/app/services/imports/case_logs_field_import_service.rb
@@ -4,6 +4,8 @@ module Imports
       case field
       when "tenant_code"
         import_from(folder, :update_tenant_code)
+      when "major_repairs"
+        import_from(folder, :update_major_repairs)
       else
         raise "Updating #{field} is not supported by the field import service"
       end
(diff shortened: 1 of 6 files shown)


app/services/imports/case_logs_field_import_service.rb
```diff
@@ -4,16 +4,38 @@ def update_field(field, folder)
...
+
+      if record.present?
+        previous_status = field_value(xml_doc, "meta", "status")
+        major_repairs_date = compose_date(xml_doc, "MRCDAY", "MRCMONTH", "MRCYEAR")
+        major_repairs = if major_repairs_date.present? && previous_status.include?("submitted")
+                          1
+                        elsif previous_status.include?("submitted")
+                          0
+                        end
+        if major_repairs.present? && record.majorrepairs.blank? && record.status != "completed"
```
smeny: There are a few cases where this block would skip with no logs.
* Partially submitted logs (saved status)
* Completed logs on our side with no `majorrepairs` value

PR #632, opened 2022-05-31
Fix user view for no role

baarkerlounger:
https://sentry.io/organizations/dluhc-core/issues/3314962691/?project=6196323

-  Data protection officers may have no other role - fix error with user details view for that case
-  Remove data accessor as a role as organisations don't use it in that way
-  Order independent array comparison for user log specs


moarpheus (approved):
does the job

PR #633, opened 2022-06-01
check if title text exists in display_title_text

kosiakkatrina:


app/helpers/interruption_screen_helper.rb
```diff
@@ -22,6 +22,8 @@ def display_informative_text(informative_text, case_log)
   end
 
   def display_title_text(title_text, case_log)
+    return "" if title_text.blank?
```
smeny, line 25: If `title_text` is an array or hash, maybe `empty?` is more explicit.

baarkerlounger: Is there an interruption screen that is not intended to have a title?

kosiakkatrina: Not anymore, but leaving this check just in case. This took a completely different turn. We previously allowed title_text as both strings and hashes, but now it's just hashes

baarkerlounger (approved):

PR #636, opened 2022-06-01
CLDC-1281: Enable support users to create organisations

baarkerlounger:
https://digital.dclg.gov.uk/jira/browse/CLDC-1281

image
image

@paulrobertlloyd this is intended as a first effort inline with "Inviting a new user". Both will likely need reworking (both to enable more fields to be added/edited and to match design guide) but I think that can be outside the scope of this PR which is more about plumbing the functionality.

### DIFF
--- a/spec/requests/organisations_controller_spec.rb
+++ b/spec/requests/organisations_controller_spec.rb
@@ -282,7 +282,7 @@ RSpec.describe OrganisationsController, type: :request do
               phone: "011101101",
               provider_type: "LA",
               holds_own_stock: "true",
-              housing_registration_no: "7917937"
+              housing_registration_no: "7917937",
             },
           }
         end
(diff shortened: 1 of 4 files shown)


paulrobertlloyd:
Do we need a field for registration number, given we show this in the table? If so, probably want to label this ‘Regulator of Social Housing registration number’ and make it have a width of `10`.

spec/requests/organisations_controller_spec.rb
```diff
@@ -644,132 +689,184 @@
...
+              provider_type: "LA",
+              holds_own_stock: "true",
+              housing_registration_no: "7917937",
+            },
+          }
+        end
+        let(:request) { post "/organisations", headers:, params: }
+
+        it "creates a new organisation" do
+          expect { request }.to change(Organisation, :count).by(1)
```
moarpheus, line 872: Could we also test that params correctly set for the new organisation record?

moarpheus (approved):

PR #638, opened 2022-06-06
CLDC-1285: Setting pregnancy to yes is allowed if a tenant's age is unknown

smeny:


app/models/case_log.rb
```diff
@@ -433,6 +433,18 @@ def plural_gender_for_person(person_num)
     end
   end
 
+  def age_known?(person_num)
```
smeny, line 436: `age_#{person_num}known` can be `nil` and as such would cause both methods to return `false`

baarkerlounger (approved):

PR #643, opened 2022-06-07
Dry searched logic

moarpheus:


app/helpers/title_helper.rb
```diff
@@ -0,0 +1,20 @@
...
+      actual_title = support_user_sab_nav?(current_user, organisation_name) ? organisation_name : page_title
+      "#{actual_title} (#{count} #{item_label} matching ‘#{searched}’)"
+    else
+      support_user_sab_nav?(current_user, organisation_name) ? "#{organisation_name} (#{page_title})" : page_title
+    end
+  end
+
+private
+
+  def support_user_sab_nav?(current_user, organisation_name)
```
baarkerlounger: Should this be `sub_nav`?

baarkerlounger (approved):

app/views/organisations/show.html.erb
```diff
@@ -1,12 +1,13 @@
 <% title = current_user.support? ? "#{@organisation.name} (Organisation details)" : "Organisation details" %>
+<% title = format_title(nil, current_user.support? ? "Organisation details" : "About your organisation", current_user, nil, nil, @organisation.name) %>
```
paulrobertlloyd:
```suggestion
<% title = format_title(nil, current_user.support? ? "About this organisation" : "About your organisation", current_user, nil, nil, @organisation.name) %>
```

app/helpers/title_helper.rb
paulrobertlloyd: Or maybe `support_sub_nav`?

Ultimately, we’d likely need this title behaviour – i.e. showing the selected sub navigation label in brackets – whenever there is sub navigation on a page (some upcoming views for schemes will need this), but for now this works perfectly.

paulrobertlloyd (approved):
Love it. Much cleaner and drier 🙂

Just one content tweak.

PR #645, opened 2022-06-08
headings partial

moarpheus:


app/views/organisations/index.html.erb
```diff
@@ -3,15 +3,11 @@
...
-  <% if current_user.support? && request.path != "/organisations" %>
-    <span class="govuk-caption-l"><%= @organisation.name %></span>
-    <h1 class="govuk-heading-l">Organisations</h1>
-  <% end %>
-</h1>
+<% if request.path == "/organisations" %>
+  <%= render partial: "headings", locals: { main: "Organisations", sub: nil } %>
+<% else %>
+  <%= render partial: "headings", locals: { main: @organisation.name, sub: "Organisations" } %>
+<% end %>
```
baarkerlounger: You could just set locals in the if/else since you're always rendering the same partial in both cases?

moarpheus: True, will look into this. Thought about moving logic into partial, but it has quite a few conditions and it won't pretty or easy to digest or scale.

paulrobertlloyd:
This partial could be used across the service, not just within organisation views, as there are many places we show a caption above a heading.

On the prototype we use a macro for this very purpose: https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data-prototype/blob/f134016044d611c43ab2056995822af12085ccd7/app/views/macros.njk#L2-L7

It takes the heading name and a number of optional parameters: heading level, caption and heading size.

I wonder if a helper or component might be best suited to this?

baarkerlounger (approved):

baarkerlounger:
@paulrobertlloyd I'm not sure this is complex enough (currently) to really justify a component. The partial can already be included anywhere you want it (though should probably be moved to a shared directory if it's going to be).

baarkerlounger:
The logic for determining the locals could go into a testable helper though. Probably micro-optimising at this point...

moarpheus:
> The logic for determining the locals could go into a testable helper though. Probably micro-optimising at this point...

Could do indeed, maybe we can refactor it again after supported housing is done, as this probably will introduce more conditions into play?

PR #647, opened 2022-06-08
CLDC-1253: Use email link in request to create a new organisation on the start page

paulrobertlloyd:
* There’s a link on the start page to contact the help desk if you need to create an account, but users would need to have a log in for Jira Service Desk (JSD), which is unlikely if they are not yet a user of CORE.

  Change this to a link to email <dluhc.digital-services@levellingup.gov.uk> with the subject line: `CORE: Request a new account`. Including ‘CORE’ in the subject line means the email will get routed through to JSD, where it can get actioned more quickly.

* Also update the email address in the footer to use the same email (this time with the subject line simply `CORE:`), as again, this means queries can be picked up and tracked within JSD.

### DIFF
--- a/app/views/layouts/_footer.html.erb
+++ b/app/views/layouts/_footer.html.erb
@@ -22,7 +22,7 @@
           <h3 class="govuk-heading-s govuk-!-margin-bottom-1">Email</h3>
           <ul class="govuk-list govuk-!-font-size-16 govuk-!-margin-bottom-0">
             <li>
-              <%= govuk_mail_to("submitcoredata@levellingup.gov.uk", class: "govuk-footer__link") %>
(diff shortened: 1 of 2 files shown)


baarkerlounger:
@paulrobertlloyd isn't the jira log in separate from having an account for this service? I think we _want_ them to log in and raise a helpdesk query?

paulrobertlloyd:
I believe the thinking is that you might not have a JIRA log in in addition to not having an account on this service, because you are a new organisation outside the world of CORE. I’ll double check.

baarkerlounger (approved):

baarkerlounger:
@paulrobertlloyd what do we need to get this moving?

baarkerlounger (approved):

paulrobertlloyd:
Got there eventually ☺️

PR #648, opened 2022-06-08
Cldc 1227 supported housing index

moarpheus:
Screenshot 2022-06-10 at 12 17 06
Screenshot 2022-06-10 at 12 17 17
Screenshot 2022-06-10 at 12 17 28


app/views/schemes/_scheme_list.html.erb
```diff
@@ -0,0 +1,39 @@
+<section class="app-table-group" tabindex="0" aria-labelledby="<%= title.dasherize %>">
+  <%= govuk_table do |table| %>
+    <%= table.caption(classes: %w[govuk-!-font-size-19 govuk-!-font-weight-regular]) do |caption| %>
+      <span class="govuk-!-margin-right-4">
```
baarkerlounger, line 4: Not a blocker for this PR but I think this caption logic gets repeated a lot and could perhaps become a component.

moarpheus: 100 agreed, just wanted to push the behemoth over the line at this point, can make a bit prettier next round

db/migrate/20220608144156_create_schemes.rb
```diff
@@ -0,0 +1,11 @@
+class CreateSchemes < ActiveRecord::Migration[7.0]
+  def change
+    create_table :schemes do |t|
+      t.string :code
+      t.string :service_name
+      t.references :organisation, null: false, foreign_key: true
```
baarkerlounger, line 6: I think elsewhere we've relied on application level `belongs_to` rather than database level foreign key constraint. This probably makes sense though... does a scheme have any meaning without an organisation?

moarpheus: I am under impression that there isn't one without org, happy to amend if I am wrong.

app/models/scheme.rb
```diff
@@ -0,0 +1,7 @@
+class Scheme < ApplicationRecord
+  belongs_to :organisation
```
baarkerlounger, line 2: Also worth adding

`has_many: :schemes` to `Organisation I think? (So you can do `organisation.schemes`)

moarpheus: I had it earlier but removed, since didnt need it excplicitely in this case - yagni. But will likely need it at some point

baarkerlounger (approved):

app/models/scheme.rb
baarkerlounger: I'd add it, think we'll definitely need it (as a console convenience if nothing else)

db/migrate/20220608144156_create_schemes.rb
baarkerlounger: Should probably actually add this for users now too, and set delete to cascade

app/models/scheme.rb
moarpheus: Will do

db/migrate/20220608144156_create_schemes.rb
moarpheus: With current spec - definitely cascade the whole lot, no point leaving ghosts behind

baarkerlounger: https://digital.dclg.gov.uk/jira/browse/CLDC-1315

app/views/schemes/_scheme_list.html.erb
baarkerlounger: https://digital.dclg.gov.uk/jira/browse/CLDC-1316

PR #649, opened 2022-06-08
Refactor depends on

baarkerlounger:
This refactors the existing depends on definition in the Form definition JSONs. Currently they all specify a method and a value to check for, this also adds the object class the method should be applied to, rather than assuming it's always the "case log".

The intention is that this allows us to also specify depends on conditions on the user - such as that a page should only show for a particular user role etc.

### DIFF
--- a/app/models/form.rb
+++ b/app/models/form.rb
@@ -157,22 +157,26 @@ class Form
     depends_on.any? do |conditions_set|
       return false unless conditions_set
 
-      conditions_set.all? do |x|
+      conditions_set.all? do |condition|
+        object, method, value = condition.values_at("object", "method", "value")
+        object_instance = case object
+                          when "case_log"
+                            case_log
+                          when "user"
+                            nil
+                          end
 
-        object = x["object"]
-        method = x["method"]
-        value = x["value"]
         if value.is_a?(Hash) && value.key?("operator")
           operator = value["operator"]
           operand = value["operand"]
-          case_log[method]&.send(operator, operand)
+          object_instance[method]&.send(operator, operand)
         else
           parts = method.split(".")
 
-          case_log_value = send_chain(parts, case_log)
+          object_instance_value = send_chain(parts, object_instance)
 
-          value.nil? ? case_log_value == value : !case_log_value.nil? && case_log_value == value
-        end 
+          value.nil? ? object_instance_value == value : !object_instance_value.nil? && object_instance_value == value
+        end
       end
     end
   end
--- a/app/models/form.rb
+++ b/app/models/form.rb
@@ -157,17 +157,22 @@ class Form
     depends_on.any? do |conditions_set|
       return false unless conditions_set
 
-      conditions_set.all? do |question, value|
+      conditions_set.all? do |x|
+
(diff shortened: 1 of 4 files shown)


config/forms/2021_2022.json
```diff
@@ -9,6 +9,20 @@
...
+                "owning_organisation_id": {
+                  "check_answer_label": "Which owning org?",
+                  "header": "Which owning org?",
+                  "hint_text": "",
+                  "type": "select",
+                  "answer_options": "get_all_orgs_answer_options"
+                }
+              },
+              "derived": true
+            },
```
dushan-madetech: Do you want this in the refactor pr?

app/models/form.rb
```diff
@@ -157,16 +157,25 @@ def depends_on_met(depends_on, case_log)
...
+                            case_log
+                          when "user"
+                            nil
+                          end
+
         if value.is_a?(Hash) && value.key?("operator")
           operator = value["operator"]
           operand = value["operand"]
-          case_log[question]&.send(operator, operand)
+          object_instance[method]&.send(operator, operand)
```
dushan-madetech, line 172: Could this blow up if object_instance is nil?

app/models/form.rb
```diff
@@ -157,16 +157,25 @@ def depends_on_met(depends_on, case_log)
...
 
-      conditions_set.all? do |question, value|
+      conditions_set.all? do |condition|
+        object, method, value = condition.values_at("object", "method", "value")
+        object_instance = case object
+                          when "case_log"
+                            case_log
+                          when "user"
+                            nil
+                          end
```
dushan-madetech, line 167: Could maybe shorten it to use a ternary operator I think? 
`object_instance = object == "case_log" ? case_log : nil`

baarkerlounger: Sort of deliberate doing this to indicate where the next PR is going. i.e. "user" is an expected value that doesn't do anything yet but will.

app/models/form.rb
baarkerlounger: Yes but that should never currently be the case (otherwise our existing code could also blow up)

config/forms/2021_2022.json
baarkerlounger: Nope good catch

app/models/form.rb
dushan-madetech: Fair enough, we could maybe add some logging here in the unlikely event it does blow up? Log the various values so we can easily debug if needed

dushan-madetech: But it's not a strong opinion so happy to go with your call here

baarkerlounger: Being changed in the next PR anyway but for the moment I think you'd absolutely want it to hard crash since it would indicate there was an error in our JSON and the form is likely to be wrong.

dushan-madetech:
Just giving it a manual test before I give the approval 🙂

dushan-madetech (approved):
Seems to be working faultlessly, 🚢 it when you think it's ready

moarpheus (approved):

PR #650, opened 2022-06-08
CLDC-1304: Export fixes

smeny:
This PR contains all changes related to recent feedback on fields we export:
- Refactored all derived variables calculation into a dedicated `DerivedVariables::CaseLogVariables` module
- Created new derived variables: `vacdays`, `new_old`, `hh_type`
- Renamed `tenant_code` to `tenancycode`
- Renamed `previous_postcode_known` to `ppcodenk`
- Renamed `shelteredaccom` to `sheltered`
- Removed duplicate field `tenancy_code` only present in our test forms
- Removed section `tenancy_information` in our test form (containing `tenancy_code`)
- Corrected renewal derived variable logic, using values removed from 22/23


app/services/exports/case_log_export_service.rb
```diff
@@ -181,15 +181,82 @@ def apply_cds_transformation(case_log, export_mode)
...
+      return if case_log.startdate.nil?
+
+      if case_log.mrcdate.present?
+        (case_log.startdate - case_log.mrcdate).to_i / 1.day
+      elsif case_log.voiddate.present?
+        (case_log.startdate - case_log.voiddate).to_i / 1.day
+      end
+    end
+
+    def get_new_old(case_log)
```
dushan-madetech: I'm not quite sure what this method is doing?

dushan-madetech (approved):
Left a comment but spotted nothing worth blocking this over

app/services/exports/case_log_export_service.rb
dushan-madetech: The name is unclear to me - what is new old

baarkerlounger:
This looks like we're calculating derived variables in the export task?

I think the disadvantages of that are:

- Splits where derived variables are calculated rather than having a single place to look (some in the model, some in the export service)
- Means these need to be calculated on every export task (every day) for a (potentially) large amount of logs which might make the export task slower than it needs to be
- Means they need to be moved if future validations are created that depend on them


The benefits are:

- We don't need to calculate them (or check if we need to calculate them) on every form page submission

app/services/exports/case_log_export_service.rb
```diff
@@ -181,15 +181,82 @@ def apply_cds_transformation(case_log, export_mode)
...
       end
 
+      attribute_hash["hhtype"] = get_hhtype(case_log)
+      attribute_hash["new_old"] = get_new_old(case_log)
+      attribute_hash["vacdays"] = get_vacdays(case_log)
+
       attribute_hash
     end
 
+    def get_vacdays(case_log)
```
baarkerlounger: Personal preference rather than anything that needs changing but I think nouns (what it returns) over verbs for method names is more idiomatic ruby (i.e. this would be `vacdays` rather than `get_vacdays`.

app/models/derived_variables/case_log_variables.rb
```diff
@@ -0,0 +1,168 @@
...
+    self.wtshortfall = if tshortfall && receives_housing_related_benefits?
+                         weekly_value(tshortfall)
+                       end
+    self.nocharge = household_charge&.zero? ? 1 : 0
+    self.housingneeds = get_housingneeds
+    if is_renewal?
+      self.underoccupation_benefitcap = 2 if collection_start_year == 2021
+      self.homeless = 1
+      self.referral = 0
+      self.waityear = 2
```
smeny, line 50: For `homeless` and `waityear` we will have to update values in production.
The former cannot have value 2, the latter cannot have value 1 for 22/23 forms.

baarkerlounger:
Be worth adding in the description I think how you've determined which (and how many) production logs will be affected by the derivation (and validation?) changes and how we'll fix them.

Do any go from being valid to invalid?

app/models/derived_variables/case_log_variables.rb
baarkerlounger: Do we know how many logs this affects?

app/models/case_log.rb
```diff
@@ -210,7 +210,7 @@ def tshortfall_unknown?
   end
 
   def is_fixed_term_tenancy?
-    [4, 6].include?(tenancy)
+    [4, 6, 7].include?(tenancy)
```
baarkerlounger: Does this affect any production logs?

app/models/case_log.rb
```diff
@@ -359,13 +367,12 @@ def ethnic_refused?
...
   def receives_housing_related_benefits?
-    receives_housing_benefit_only? || receives_uc_with_housing_element_excl_housing_benefit? ||
-      receives_housing_benefit_and_universal_credit?
-  end
-
-  def benefits_unknown?
-    # 3: Don’t know
-    hb == 3
+    if collection_start_year <= 2021
+      receives_housing_benefit_only? || receives_uc_with_housing_element_excl_housing_benefit? ||
```
baarkerlounger, line 370: Same question here

app/models/validations/financial_validations.rb
```diff
@@ -51,10 +51,21 @@ def validate_negative_currency(record)
   end
 
   def validate_tshortfall(record)
-    if record.has_hbrentshortfall? &&
+    return unless record.startdate
+
+    if record.collection_start_year <= 2021
```
baarkerlounger, line 62: Same question here

db/migrate/20220613123730_add_export_fields.rb
```diff
@@ -0,0 +1,13 @@
+class AddExportFields < ActiveRecord::Migration[7.0]
+  def change
+    change_table :case_logs, bulk: true do |t|
+      t.column :hhtype, :integer
+      t.column :new_old, :integer
+      t.column :vacdays, :integer
+      t.rename :tenant_code, :tenancycode
+      t.rename :previous_postcode_known, :ppcodenk
```
baarkerlounger, line 8: Possible this triggers a couple of seconds of downtime/Sentry crash if one of the servers still has old application code after the first server has finished running migrations but should be fine.

spec/features/form/check_answers_page_spec.rb
```diff
@@ -220,13 +220,13 @@
       it "they can click a button to skip sections until the next incomplete section" do
         visit("/logs/#{skip_section_case_log.id}/household-characteristics/check-answers")
         click_link("Save and go to next incomplete section")
-        expect(page).to have_current_path("/logs/#{skip_section_case_log.id}/tenancy-code")
+        expect(page).to have_current_path("/logs/#{skip_section_case_log.id}/property-information/check-answers")
```
baarkerlounger, line 224: Is this one still testing what it was supposed to? Should be taking you to an incomplete question rather than a check answers page I think?

app/models/derived_variables/case_log_variables.rb
smeny: `homeless` value 2 was present on a single record last time I checked.
I haven't checked `waityear` yet.

spec/features/form/check_answers_page_spec.rb
smeny: Previously we had both tenancy code and tenant code in the test form.
I deleted the tenancy code section and kept the tenant code (now renamed tenancy code).

baarkerlounger: Yep, but doesn't that now mean this test doesn't have the data it expected? (And incomplete question left to navigate to)?

app/models/case_log.rb
smeny: This is mostly used in validations. As 7 is a new value, the only risk is that some logs could be now invalid.

app/models/case_log.rb
smeny: This has no effect since all we do is remove a check on a removed value (for documentation purposes more than anything).

app/models/validations/financial_validations.rb
smeny: Since this is mostly used for validation, and checking a new value (refused), we could potentially have invalid logs flagged as completed.

spec/features/form/check_answers_page_spec.rb
smeny: It is hard for me to assess tbh. These tests were written with a specific configuration in mind, the problem is that our configuration drives database fields. I am not sure we are getting much value from these tests, either they should check a general logic (based on hardcoded states) or test an existing configuration used in production.

app/models/validations/financial_validations.rb
baarkerlounger: Could we check if that affects any production logs before hand?

app/models/case_log.rb
baarkerlounger: Can we check before hand if any production logs are affected?

spec/features/form/check_answers_page_spec.rb
baarkerlounger: I think they _are_ intended to check a general logic (the test form fixtures are just arbitrary after all and not meant to reflect a real form (although you're right that it's somewhat constrained by the database coupling)). So we could have just left the old question/section in there (and just changed names to something else if confusing).

 Think the purpose of *this* particular test was to check that the "Save and go to next incomplete section" button works correctly for the scenario where there is an incomplete section still to go (which with the form changes is no longer the case here).
 
 Let me know if that does/doesn't make sense, might be easier to pair or chat through.

dushan-madetech: the purpose of this test is to check that the save and go to next incomplete section button would actually skip a completed subsection and take you to the next incomplete one. E.g. for the real form if you were on the household characteristics subsection and had already completed the household needs subsection (the next one in the form) then clicking the button would skip that one and take you to the household situation subsection

dushan-madetech: Hope the above clarifies the intended behaviour but if not I'm also happy to pair on this

spec/fixtures/exports/case_logs.csv
```diff
@@ -1,2 +1,2 @@
...
-2,BZ737,35,F,2,4,6,0,2,32,M,6,,,,,,,,,,,,,,,,,,,1,0,1,0,1,2,0,5,1,SE26RT,6,7,3,2,1,68,1,1,2,2,1,NW15TY,1,1,1,0,,1,0,0,0,0,0,0,1,0,0,0,0,0,0,0,0,0,1,0,0,0,,,4,123,E09000003,E07000105,6,1,2020-05-05 10:36:49 UTC,0,2022-02-02 10:36:49 UTC,1,2,1,2019-11-03 00:00:00 UTC,2,1,7,0,0,2,0,,200.0,50.0,40.0,35.0,325.0,12.0,,1,1,0,100.0,25.0,20.0,17.5,162.5,6.0,0,1,,2,P,,,,,,,,,,0,4,2,638,{id},{owning_org_id},DLUHC,1234,{managing_org_id},DLUHC,1234,2022-02-08 16:52:15 UTC,2022-02-08 16:52:15 UTC
```
baarkerlounger: Is this a checkbox? Should it be 1/0?

smeny: that's the change you pushed yesterday, 2 instead of 0.

smeny: (conflicts etc...)

app/models/case_log.rb
```diff
@@ -338,6 +331,12 @@ def receives_housing_benefit_only?
     hb == 1
   end
 
+  def benefits_unknown?
+    # 3: Don’t know
```
kosiakkatrina: The method name is quite clear, is this comment providing some other value?

app/models/validations/financial_validations.rb
```diff
@@ -51,10 +51,21 @@ def validate_negative_currency(record)
...
-        (record.benefits_unknown? ||
-          record.receives_no_benefits? ||
-            record.receives_universal_credit_but_no_housing_benefit?)
+    return unless record.startdate
+
+    cannot_have_outstanding_amount = if record.collection_start_year <= 2021
+                                       record.has_hbrentshortfall? &&
+                                         (record.benefits_unknown? ||
+                                           record.receives_no_benefits? ||
+                                           record.receives_universal_credit_but_no_housing_benefit?)
```
kosiakkatrina: Would it make sense to extract these conditions?

app/models/case_log.rb
smeny: The idea was to explicitly identify what option is behind values, but you are right, we could remove the comment since it is not adding much.

spec/features/form/check_answers_page_spec.rb
smeny: So pairing on this, the test is still correct.

The confusion comes from the fact that the next section is already prefilled to some extent ( in-progress case log which prefills the postcode). When that happens, we go to the check answers page of that section rather than the first question. So property information is indeed the next incomplete section (which is what is under test).

app/models/derived_variables/case_log_variables.rb
smeny: 1 record (renewal = 1 and homeless = 2)
1 record (renewal = 1 and waityear = 1)

Same record ID

app/models/case_log.rb
smeny: I think I am going to remove the value, we are not checking for fixed terms for option 2 (which was lifetime). We have 49 records with option 7 (assured lifetime) and no tenancylength.

app/models/derived_variables/case_log_variables.rb
baarkerlounger: Meaning that 1 record will become invalid? I guess we should just clear the value and have it set back to in progress and inform them?

smeny: we just need to change the values for this single record

app/models/validations/financial_validations.rb
smeny: No results for hbrentshortfall = 1 and hb = 10

PR #653, opened 2022-06-10
Support only questions

baarkerlounger:
- This PR enables Support Users to create a log for any organisation from the top level `/logs` page. They can set the owning organisation for the log in the "Set up your log" section.

- It does **not** yet enable creating a log for a specific organisation from the `/organisation/:id/logs` page, as that will follow in a separate PR.

<br/>

**Support User:**

image
image

**Non-support User**
image
image


app/models/case_log.rb
```diff
@@ -31,8 +31,8 @@ class CaseLog < ApplicationRecord
   before_validation :set_derived_fields!
   before_save :update_status!
 
-  belongs_to :owning_organisation, class_name: "Organisation"
-  belongs_to :managing_organisation, class_name: "Organisation"
+  belongs_to :owning_organisation, class_name: "Organisation", optional: true
```
baarkerlounger, line 34: Slightly unfortunate side effect of allowing Support users to set the `owning_organisation` for a log through the form is that it requires it to be possible to create a log that doesn't yet have an organisation.

This could perhaps be mitigated by not "creating" the log on clicking the create button, but rather just calling `.new` and keeping an object in session until a first answer is provided. That's likely better anyway as it also eliminates the `not_started` class of logs but would be a larger change than I want to make in this PR.

moarpheus: That's a very good point and I am happy with the reason provided!

moarpheus (approved):
Looks good lets ship this!

PR #654, opened 2022-06-10
Log whenever a user encounters a validation error message to Logit

baarkerlounger:


moarpheus (approved):
Should do the trick!

sjmoore4:
Thank you Dan for doing this.

would I be able to get access to Logit to extract this data?

PR #660, opened 2022-06-14
Differentiate between internal and external interfaces

paulrobertlloyd:
As an extra safeguard, using a different border and tag colour in the header can help distinguish the internal support interface (which has more destructive functions) and external customer facing interface, especially if a user may have access to both. 

It also helps us identify the different views/permissions when reviewing the design.

## External customer

Screenshot 2022-06-14 at 13 19 29

## Internal support user

Screenshot 2022-06-14 at 13 19 56

### DIFF
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -43,7 +43,7 @@
     <%= govuk_skip_link %>
 
     <%= govuk_header(
-      classes: govuk_header_classes,
+      classes: govuk_header_classes(current_user),
       service_url: current_user.nil? ? "/" : "/logs",
       navigation_classes: "govuk-header__navigation--end",
     ) do |component|
@@ -60,7 +60,7 @@
 
     <%= govuk_phase_banner(
       classes: "govuk-width-container",
-      tag: govuk_phase_banner_tag,
+      tag: govuk_phase_banner_tag(current_user),
       text: "This is a new service – help us improve it by #{feedback_link}".html_safe,
     ) %>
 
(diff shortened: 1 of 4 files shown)


app/views/layouts/application.html.erb
```diff
@@ -43,7 +43,7 @@
     <%= govuk_skip_link %>
 
     <%= govuk_header(
-      classes: "app-header",
+      classes: current_user && current_user.support? ? "app-header app-header--orange" : "app-header",
```
paulrobertlloyd: I wonder if this values (or maybe if user is logged in and a support user) should be derided somewhere in the application logic, rather than in the view?

app/views/layouts/application.html.erb
```diff
@@ -60,7 +60,10 @@
 
     <%= govuk_phase_banner(
       classes: "govuk-width-container",
-      tag: { text: "Beta" },
+      tag: {
+        colour: current_user && current_user.support? ? "orange" : nil,
+        text: current_user && current_user.support? ? "Support beta" : "Beta"
```
paulrobertlloyd: Ditto.

app/views/layouts/application.html.erb
baarkerlounger: Perhaps a method in `helpers/application_helper.rb` called `govuk_header_classes`?

baarkerlounger: Also easier to test that way

baarkerlounger (approved):

PR #661, opened 2022-06-14
CLDC-1248-select-needs-type

dushan-madetech:
Screenshot 2022-06-14 at 15 15 40

### DIFF
--- a/app/models/case_log.rb
+++ b/app/models/case_log.rb
@@ -114,7 +114,7 @@ class CaseLog < ApplicationRecord
 
   def needs_question_enabled?
     FeatureToggle.needs_question_enabled?
-  end 
+  end
 
   def weekly_net_income
     return unless earnings && incfreq
--- a/app/models/case_log.rb
+++ b/app/models/case_log.rb
@@ -112,6 +112,10 @@ class CaseLog < ApplicationRecord
     status == "in_progress"
   end
 
+  def needs_question_enabled?
+    FeatureToggle.needs_question_enabled?
+  end 
+
   def weekly_net_income
     return unless earnings && incfreq
 
@@ -520,7 +524,7 @@ private
 
   def set_derived_fields!
     # TODO: Remove once we support supported housing logs
-    self.needstype = 1 unless needstype
+    self.needstype = 1 unless needs_question_enabled?
     if rsnvac.present?
       self.newprop = has_first_let_vacancy_reason? ? 1 : 2
     end
--- a/app/models/case_log.rb
+++ b/app/models/case_log.rb
@@ -112,8 +112,8 @@ class CaseLog < ApplicationRecord
     status == "in_progress"
   end
 
-  def needs_question_enabled?
-    FeatureToggle.needs_question_enabled?
+  def supported_housing_schemes_enabled?
+    FeatureToggle.supported_housing_schemes_enabled?
   end
 
   def weekly_net_income
@@ -524,7 +524,7 @@ private
 
   def set_derived_fields!
     # TODO: Remove once we support supported housing logs
-    self.needstype = 1 unless needs_question_enabled?
+    self.needstype = 1 unless FeatureToggle.supported_housing_schemes_enabled?
     if rsnvac.present?
       self.newprop = has_first_let_vacancy_reason? ? 1 : 2
     end
--- a/config/initializers/feature_toggle.rb
+++ b/config/initializers/feature_toggle.rb
@@ -1,9 +1,9 @@
 class FeatureToggle
   def self.needs_question_enabled?
     !Rails.env.production?
-  end 
+  end
 
   def self.show_schemes_button?
     Rails.env.production?
   end
-end 
+end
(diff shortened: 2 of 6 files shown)


config/initializers/feature_toggle.rb
```diff
@@ -0,0 +1,9 @@
+class FeatureToggle
+  def self.needs_question_enabled?
```
baarkerlounger: I think `needs_question_enabled` and `show_schemes_button` are a single feature flag not two - in that they are part of schemes which all needs to be on or off together

app/models/case_log.rb
```diff
@@ -520,7 +524,7 @@ def tenancylength_optional?
 
   def set_derived_fields!
     # TODO: Remove once we support supported housing logs
-    self.needstype = 1 unless needstype
+    self.needstype = 1 unless needs_question_enabled?
```
baarkerlounger: I think I'd use the feature toggle class directly here rather than a wrapper method so it's more obvious this *is* a feature flag and not application logic.

config/initializers/feature_toggle.rb
```diff
@@ -0,0 +1,9 @@
...
+    !Rails.env.production?
+  end
+
+  def self.show_schemes_button?
```
baarkerlounger: I think this should probably always follow a set pattern of: `true` if you want the feature on, `false` if you don't.

This seems like for `needs_question_enabled?` true is on but for `show_schemes_button` true is actually off?

app/models/case_log.rb
dushan-madetech: I did it this way since the `depends_on` currently relies on the case log model

dushan-madetech: I wasn't quite sure how to do this without modifying the `depends_on` method

moarpheus:
Maybe a bit complicated at this point - but I thought about having a dedicated config file e.g. config/application.yml where you can define your vars per env and access in the app ( you would need to load it on the start in initializer ). There is also a gem to manage those kind of things ( Figaro ) but probably not enough case to use it just now. @baarkerlounger let me know what you think

development:
  some_flag: false

production:
  some_flag: true

test:
  some_flag: true

config/initializers/feature_toggle.rb
dushan-madetech: This is probably down to poor naming on my part - I wanted the needs question to be enabled on non prod environments

config/initializers/feature_toggle.rb
dushan-madetech: My thinking was I wanted to give us flexibility to enable/disable specific parts of the feature if desired - happy to change if this isn't desired 😄

app/models/case_log.rb
baarkerlounger: As in, I would probably just do:

```ruby
self.needstype = 1 unless FeatureToggle.needs_question_enabled?
```

config/initializers/feature_toggle.rb
baarkerlounger: Yeah I don't think it is, it's pretty much all or nothing for schemes.

config/initializers/feature_toggle.rb
baarkerlounger: Yep, that part makes sense to me, but if I'm reading this right I think:

```ruby
> FeatureToggle.show_schemes_button?
=> true
```

Would actually *disable* the button, which seems counter-intuitive?

dushan-madetech: Will amend this to follow the suggested pattern 😄

app/models/case_log.rb
dushan-madetech: Ah whoops I misread where this comment was made - ignore me 🤦

config/initializers/feature_toggle.rb
dushan-madetech: @baarkerlounger I've amended the code to fix this - it was structured in a way that returning true here would hide the button so I've update the condition to display the button. Will link to that commit in the next comment

dushan-madetech: @baarkerlounger commit is here: https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/661/commits/120b7dde9aa035e07fa959ca1b7bf69d49de3142 

check the changes in `app/components/primary_navigation_component.rb`, this amendment should now make it line up with your expectations

app/models/case_log.rb
```diff
@@ -112,8 +112,8 @@ def in_progress?
     status == "in_progress"
   end
 
-  def needs_question_enabled?
-    FeatureToggle.needs_question_enabled?
+  def supported_housing_schemes_enabled?
```
baarkerlounger, line 115: Think we can remove this method? Not used anywhere now right?

app/models/case_log.rb
```diff
@@ -112,6 +112,10 @@ def in_progress?
     status == "in_progress"
   end
 
+  def supported_housing_schemes_enabled?
```
baarkerlounger, line 115: I wonder if using `method_missing` and delegating might make this more generically useful... (or more complicated)

baarkerlounger (approved):

app/models/case_log.rb
dushan-madetech: I might need some more explanation around this as I'm not 100% sure what you mean

PR #663, opened 2022-06-14
Refactor log setup into it's own form

baarkerlounger:
Given that log setup is a little different from other form section in that it:

- Must be completed first (in order)
- Must be consistent across years and not part of the annual data collection change cycle

It makes sense to extract it out. This would also help make it easier to make the setup section not JSON defined at all but rather a coded part of the application in a subsequent PR (which may be useful as it's likely to require a lot more "application context" like having user specific questions/answer options etc).


spec/features/form/tasklist_page_spec.rb
```diff
@@ -27,6 +27,16 @@
...
+  let(:setup_completed_log) do
+    FactoryBot.create(
+      :case_log,
+      owning_organisation: user.organisation,
+      managing_organisation: user.organisation,
+      renewal: 1,
+      rent_type: 1,
+      startdate: Time.zone.local(2021, 5, 1),
+    )
+  end
```
dushan-madetech: This is what the `:about_completed` trait in the factory aimed to do . You could use that and possibly rename it to setup completed instead

dushan-madetech (approved):
Left one comment - suggestion but not required. This looks good, I like this refactor a lot 😄

moarpheus (approved):
LGTM

PR #665, opened 2022-06-15
Fix lettings allocation (cbl, chr, cap) import

baarkerlounger:
For lettings allocations we have a checkbox questions (1 = Yes, 0 = No) so we want to import:

```
1 Yes => 1
2 No  => 0
null  => nil
```

Once released we will need to run:

```
cf run-task dluhc-core-production --command "bundle exec rake core:data_import_field['lettings_allocation','2022-05-10/logs/']" --name believe-lettings_allocation-reimport

cf run-task dluhc-core-production --command "bundle exec rake core:data_import_field['lettings_allocation','2022-05-27-barrow/logs/']" --name barrow-lettings_allocation-reimport
```


moarpheus (approved):

smeny (approved):
LGTM

I might want to check on the export PR, how all checkboxes are handled.

PR #667, opened 2022-06-15
Lint JavaScript

paulrobertlloyd:
We currently lint Ruby, ERB and SCSS, that just leaves JavaScript. GDS use and recommend [StandardJS](https://standardjs.com), so that’s what I’ve added here.

Also pins `govuk-frontend` to v4.0.1 given issue highlighted in #644, and adding a new package will upgrade this to the next minor version.

Should appear as part of CI, as added to the `lint` Rake task. Also available by running `rake standard`.

### DIFF
--- a/app/frontend/controllers/conditional_question_controller.js
+++ b/app/frontend/controllers/conditional_question_controller.js
@@ -1,36 +1,36 @@
-import { Controller } from "@hotwired/stimulus"
+import { Controller } from '@hotwired/stimulus'
 
 export default class extends Controller {
-  initialize() {
+  initialize () {
     this.displayConditional()
   }
 
-  displayConditional() {
-    if(this.element.checked) {
-      let selectedValue = this.element.value
-      let conditional_for = JSON.parse(this.element.dataset.info)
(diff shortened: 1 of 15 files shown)


app/frontend/controllers/conditional_question_controller.js
```diff
@@ -1,36 +1,36 @@
...
-      let conditional_for = JSON.parse(this.element.dataset.info)
-      Object.entries(conditional_for).map(([targetQuestion, conditions]) => {
-          if(conditions.map(String).includes(String(selectedValue))) {
+  displayConditional () {
+    if (this.element.checked) {
+      const selectedValue = this.element.value
+      const conditionalFor = JSON.parse(this.element.dataset.info)
+
+      Object.entries(conditionalFor).forEach(([targetQuestion, conditions]) => {
+        if (!conditions.map(String).includes(String(selectedValue))) {
```
paulrobertlloyd, line 14: Worth taking a note of this change. Changed the `if`/`else` logic to prevent using an empty block, and now using `forEach` instead of `map` because we’re not modifying the entries.

baarkerlounger (approved):
Looks fine to me

PR #669, opened 2022-06-15
Cldc 1307 supported housing scheme locations

moarpheus:
Screenshot 2022-06-15 at 14 10 06

### DIFF
--- a/app/models/location.rb
+++ b/app/models/location.rb
@@ -1,3 +1,22 @@
 class Location < ApplicationRecord
   belongs_to :scheme
+
+  WHEELCHAIR_ADAPTATION = {
+    0 => "No",
+    1 => "Yes",
+  }.freeze
+
+  def display_attributes
+    [
+      { name: "Location code ", value: location_code, suffix: false },
+      { name: "Postcode", value: postcode, suffix: county },
+      { name: "Type of unit", value: type_of_unit, suffix: false },
+      { name: "Type of building", value: type_of_building, suffix: false },
+      { name: "Wheelchair adaptation", value: wheelchair_adaptation_display, suffix: false },
+    ]
+  end
+
+  def wheelchair_adaptation_display
+    WHEELCHAIR_ADAPTATION[wheelchair_adaptation]
+  end
 end
--- a/app/models/location.rb
+++ b/app/models/location.rb
@@ -1,0 +1,3 @@
+class Location < ApplicationRecord
+  belongs_to :scheme
+end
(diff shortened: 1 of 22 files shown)


app/models/location.rb
```diff
@@ -0,0 +1,22 @@
+class Location < ApplicationRecord
+  belongs_to :scheme
+
+  WHEELCHAIR_ADAPTATION = {
```
baarkerlounger: Should this be an enum? (The benefits are it converts between DB and display/console value for you, and it enforces that saved value must be one of those)

app/views/schemes/locations.html.erb
```diff
@@ -0,0 +1,41 @@
...
+<%= govuk_back_link(href: request.referer.to_s) %>
+
+<%= render partial: "organisations/headings", locals: { main: @scheme.service_name, sub: nil } %>
+
+<%= render SubNavigationComponent.new(items: scheme_items(request.path, @scheme.id, @scheme.locations.count.eql?(1) ? "1 location" : "#{@scheme.locations.count} locations")) %>
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-three-quarters">
+    <% @locations.each do |location| %>
+      <section class="x-govuk-summary-card govuk-!-margin-bottom-6">
```
baarkerlounger, line 13: Should this be using a govuk component? @paulrobertlloyd ?

baarkerlounger (approved):

app/models/location.rb
moarpheus: Will definitely have a look

app/views/schemes/locations.html.erb
paulrobertlloyd: It’s not a GOV.UK component (yet!) but in the interim, we should add our own component, as we’re using this pattern in 2 places now (and there’s a ticket to use it in a third).

moarpheus: Not sure where is the other place, but will be happy to extract into separate component in the *next iteration? @baarkerlounger @paulrobertlloyd

paulrobertlloyd: Currently used on the review log page (i.e. https://dluhc-core-staging.london.cloudapps.digital/logs/13828/review, https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/blob/06adfbaf6d692a282facee6726a76d3d91d245d3/app/views/form/review.html.erb), and [CLDC-1310](https://digital.dclg.gov.uk/jira/browse/CLDC-1310) is for using summary cards for listing tenants.

Component should take all the properties that can be passed to the summary list component in addition to a header title and header actions.

PR #671, opened 2022-06-16
CLDC-1290 Permitted user can create a new scheme

moarpheus:
Screenshot 2022-06-29 at 16 00 48
Screenshot 2022-06-29 at 16 00 03
Screenshot 2022-06-29 at 15 59 56
Screenshot 2022-06-29 at 15 59 49
Screenshot 2022-06-29 at 15 59 43
Screenshot 2022-06-29 at 15 59 38


app/controllers/schemes_controller.rb
```diff
@@ -25,12 +26,77 @@ def locations
...
+  def support_patch?
+    params[:scheme][:intended_stay]
+  end
+
+  def details_patch?
+    params[:scheme][:service_name]
+  end
+
+  def clean_params
+    code = "S#{SecureRandom.alphanumeric(5)}".upcase
```
baarkerlounger: Any reason not to just use the auto incrementing ID here rather than introducing a new gem?

moarpheus: This didnt occur to me at the time. And I wanted something that will generate me some IDs quickly so I can focus on other bits, will be more than happy to change this.

moarpheus: Since we have to prefix our code with 'S' - will need to write logic for this

app/controllers/schemes_controller.rb
```diff
@@ -98,7 +85,22 @@ def check_answers
     render "schemes/check_answers"
   end
 
-private
+  private
+
+  def derive_next_path page
```
baarkerlounger: Personal style thing/just a nit but I think where possible naming methods with nouns after the thing they return is more idiomatic than naming them with a verb after the thing they do (i.e. `next_page_path` vs `derive_next_page`). (Unless you're deliberately doing something procedural I suppose).

app/models/scheme.rb
```diff
@@ -7,103 +9,139 @@ class Scheme < ApplicationRecord
...
 
-  def support_type_display
-    SUPPORT_TYPE[support_type]
-  end
+private
 
-  def intended_stay_display
-    INTENDED_STAY[intended_stay]
+  def create_code
+    self.code = Scheme.last.nil? ? "S1" : "S#{Scheme.last.code[1..].to_i + 1}"
```
baarkerlounger: Is this field (and the `before_create`) actually needed at all or could we just prepend "S" to "scheme.id" where ever we display it? It doesn't seem worth storing?

baarkerlounger: Also "Scheme.last.code" seems unlikely to be race condition safe

moarpheus: Then we need to write logic not to prefix S for the imported schemes and also I am not sure that import not going to bring repeated codes in, this makes me think that potentially there is a better solution to address 'code' issue. Noted in the non race condition safe, will look into that.

baarkerlounger: Might be worth syncing up on but my expectation is that old schemes are imported with an auto-incrementing "id" (same as users or case logs), and we also save an additional "old_id" or "old_code" field that has the old value (and will be unique since it uniquely identified the scheme in the old system - new schemes won't have a value for this field).

On display we "ideally" always display the auto-incrementing "id" with "S" prepended. If we need to display the old code, we can do that if it's set _or_ display the id if it isn't. i.e. something like 

```ruby
def display_code
    scheme.old_code || "S" + scheme.id.to_s
end
```

app/controllers/schemes_controller.rb
```diff
@@ -25,20 +26,120 @@ def locations
...
+  def new
+    @scheme = Scheme.new
+  end
+
+  def create
+    @scheme = Scheme.new(scheme_params)
+    if @scheme.save
+      render "schemes/primary_client_group"
+    else
+      @scheme.errors.add(:organisation_id, message: @scheme.errors[:organisation])
```
baarkerlounger, line 38: Shouldn't this happen in the model?

baarkerlounger (approved):
Looks good to me when all is green.

PR #672, opened 2022-06-16
Refactor log setup into code

baarkerlounger:
The idea behind this refactor is that while most of our form can be modelled as "config" that can change over time, the "Setup your log" section is fundamentally different in a few key ways:

- It can't be completed in any order, it *must* be completed first
- Changing answers here is likely to invalidate large parts of the rest of the form
- Questions and answers here depend heavily on application context like who the user is, what organisations exist etc

Effectively the "Setup your log" section is really closer to a form to select your form rather than a part of the form under the hood. This PR moves it from being JSON "config" to being a coded part of our application, this makes it easier to add logic like answer options depending on specific model data or methods etc.

The "Setup your log" section created by this code is still part of the Form object, so other than how it's constructed nothing else really changes downstream.


moarpheus (approved):
Looks very good

PR #673, opened 2022-06-16
Replaced repeated code with view component for results table 

Ted-U:
in orgs, schemes, users and logs


app/components/view_result_component.rb
```diff
@@ -0,0 +1,13 @@
+class ViewResultComponent < ViewComponent::Base
```
baarkerlounger: Maybe `SearchResultCaptionComponent`?

baarkerlounger:
Be good to see some unit tests for the component and looks like it needs linting (`bundle exec rake lint`) but otherwise :+1:

app/components/view_result_component.rb
Ted-U: I have updated the component name as suggested

Ted-U:
> Be good to see some unit tests for the component and looks like it needs linting (`bundle exec rake lint`) but otherwise 👍

Linting is done

app/components/search_result_caption_component.html.erb
```diff
@@ -0,0 +1,7 @@
+  <span class="govuk-!-margin-right-4">
+    <% if searched.present? %>
+      <strong><%= count %></strong> <%= item_label %> found matching ‘<%= searched %>’ of <strong><%= total_count %></strong> total <%= item %>. <%= govuk_link_to("Clear search", request.path) %>
+    <% else %>
+      <strong><%= count %></strong> total <%= item %>.
+    <% end %>
```
baarkerlounger, line 6: Looks like the `title` is missing here

baarkerlounger, line 6: Looks like the `title` is missing here

Ted-U:
> Be good to see some unit tests for the component and looks like it needs linting (`bundle exec rake lint`) but otherwise 👍

Unit tests now added

app/components/search_result_caption_component.html.erb
Ted-U: Title has been replaced with item but I can change that for clarity if needed?

baarkerlounger (approved):

moarpheus (approved):

PR #674, opened 2022-06-17
Add an organisation question for the support user

baarkerlounger:
- [x] Support users see an extra question for owning organisation
- [x] Support users see an extra question for user (created by)
- [x] The two extra questions are hidden and auto set for non-support users
- [x] The two extra questions are hidden from check answers for non-support users
- [x] The values for the two extra questions are not cleared even if the page isn't routed to
- [x] The answer options for organisation is based off organisations that exist in the database. If a user has already been selected, only that users organisation is available to select.
- [x] The answer options for user is based off users that exist in the database. If an owning organisation has already been selected, only that organisations users are available to select.
- [x] If owning organisation is set by support user, managing organisation is derived (same org)
- [ ] Check performance impact (particularly additional database queries)

### DIFF
--- a/app/models/derived_variables/case_log_variables.rb
+++ b/app/models/derived_variables/case_log_variables.rb
@@ -7,6 +7,7 @@ module DerivedVariables::CaseLogVariables
 
   def set_derived_fields!
     # TODO: Remove once we support supported housing logs
+    self.managing_organisation_id ||= owning_organisation_id
     self.needstype = 1 unless supported_housing_schemes_enabled?
     if rsnvac.present?
       self.newprop = has_first_let_vacancy_reason? ? 1 : 2
--- a/app/models/derived_variables/case_log_variables.rb
+++ b/app/models/derived_variables/case_log_variables.rb
@@ -6,8 +6,9 @@ module DerivedVariables::CaseLogVariables
   end
 
   def set_derived_fields!
-    # TODO: Remove once we support supported housing logs
+    # TODO: Remove once we support parent/child relationships
     self.managing_organisation_id ||= owning_organisation_id
+    # TODO: Remove once we support supported housing logs
     self.needstype = 1 unless supported_housing_schemes_enabled?
     if rsnvac.present?
       self.newprop = has_first_let_vacancy_reason? ? 1 : 2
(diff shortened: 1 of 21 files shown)


app/models/form/setup/questions/owning_organisation_id.rb
```diff
@@ -0,0 +1,39 @@
...
+
+  def label_from_value(value)
+    return unless value
+
+    answer_options[value]
+  end
+
+  def hidden_in_check_answers
+    !form.current_user.support?
+  end
```
dushan-madetech, line 38: I wonder if you could DRY this out somehow since it appears elsewhere? 🤔 probably not worth it though

spec/requests/form_controller_spec.rb
```diff
@@ -141,6 +142,24 @@
...
+          let(:user) { FactoryBot.create(:user, :support) }
+
+          it "routes to the page" do
+            get "/logs/#{case_log.id}/organisation"
+            expect(response).to have_http_status(:ok)
+          end
+        end
+
+        context "when the dependency is not met" do
+          it "routes to the page" do
```
dushan-madetech: I think this description is incorrect - they get redirected instead of routed to the page, at least based on the test code

spec/models/form/setup/questions/owning_organisation_id_spec.rb
```diff
@@ -0,0 +1,91 @@
+require "rails_helper"
+
+RSpec.describe Form::Setup::Questions::OwningOrganisationId, type: :model do
```
dushan-madetech, line 3: The tests here feel very similar to the others in `spec/models/form/setup/questions` so maybe a possibility of some future refactoring there? 🤔 Not a big hang up for this PR though

dushan-madetech (approved):
Left some comments - this was a pretty big PR so I might have missed some small bits but nothing really stood out to me as anything to block this over

app/models/form/setup/questions/owning_organisation_id.rb
baarkerlounger: I think it's just incidentally the same (happens to be the same condition), as opposed to the same for some rule that we could abstract.

app/models/derived_variables/case_log_variables.rb
```diff
@@ -7,6 +7,7 @@ def supported_housing_schemes_enabled?
 
   def set_derived_fields!
     # TODO: Remove once we support supported housing logs
```
smeny: I think this comment belongs to the line bellow

PR #675, opened 2022-06-20
Fix index schemes data coordinator

moarpheus:
Nothing more to add, this is leftover from switching supported-housing to schemes


baarkerlounger:
Should there be a spec for the failing case?

PR #676, opened 2022-06-20
Content tweaks for user and organisation management

paulrobertlloyd:
From the mob session:

* Use **Owns housing stock** (Yes/No) instead of **Holds own stock** (True/False)
* Use **Invite somebody to submit CORE data** instead of **Invite a user to submit CORE data**
* Use **Is this person…**/**Are you…** instead of **Are they**/**Are you**. Update helper name from `pronoun` to `perspective` to reflect this change

### DIFF
--- a/app/views/users/new.html.erb
+++ b/app/views/users/new.html.erb
@@ -55,13 +55,13 @@
         [OpenStruct.new(id: true, name: "Yes"), OpenStruct.new(id: false, name: "No")],
         :id,
         :name,
-        legend: { text: "Are #{pronoun(@user, current_user)} the organisation’s data protection officer?", size: "m" } %>
+        legend: { text: "#{perspective(@user, current_user)} the organisation’s data protection officer?", size: "m" } %>
 
       <%= f.govuk_collection_radio_buttons :is_key_contact,
         [OpenStruct.new(id: true, name: "Yes"), OpenStruct.new(id: false, name: "No")],
         :id,
         :name,
-        legend: { text: "Is this user a key contact for this service?", size: "m" },
+        legend: { text: "#{perspective(@user, current_user)} a key contact for this service?", size: "m" },
         hint: { text: "This is a person responsible for sharing information about social housing lettings and sales data within the organisation." } %>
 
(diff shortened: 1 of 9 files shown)


app/views/users/new.html.erb
```diff
@@ -55,13 +55,13 @@
         [OpenStruct.new(id: true, name: "Yes"), OpenStruct.new(id: false, name: "No")],
         :id,
         :name,
-        legend: { text: "Are #{pronoun(@user, current_user)} the organisation’s data protection officer?", size: "m" } %>
+        legend: { text: "Is this person the organisation’s data protection officer?", size: "m" } %>
```
baarkerlounger: Any reason not to use `#{perspective(@user, current_user}` here? That way the wording is in a single place?

app/views/users/new.html.erb
```diff
@@ -55,13 +55,13 @@
...
 
       <%= f.govuk_collection_radio_buttons :is_key_contact,
         [OpenStruct.new(id: true, name: "Yes"), OpenStruct.new(id: false, name: "No")],
         :id,
         :name,
-        legend: { text: "Is this user a key contact for this service?", size: "m" },
+        legend: { text: "Is this person a key contact for this service?", size: "m" },
```
baarkerlounger: Same here

app/views/users/new.html.erb
paulrobertlloyd: Don’t need `perspective` here as only a support user/data coordinator can create a new user (you can’t create yourself). Page originally used the helper for only one of the strings. Not fussed either way, however.

baarkerlounger: Alternatively "Is this person" could become a translation string that's referenced in both places? Either way would be useful to have it in a single place is it's intended to be consistent I think.

paulrobertlloyd: Good point. Will go with the simpler option and use the `perspective` helper on both new and edit views.

paulrobertlloyd:
Not sure why tests failing here and local, seem to relate to schemes somehow?

baarkerlounger (approved):

PR #678, opened 2022-06-20
Update error message

Dan13L3:
@baarkerlounger I spotted this in the mob session earlier, so have updated the error to give the user more of a clear, GDS-style direction. Let me know if it looks okay


baarkerlounger:
Closed in favour of https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/681

PR #680, opened 2022-06-20
Default owning organisation if created from their page

baarkerlounger:
image
image
image

### DIFF
--- a/app/views/organisations/logs.html.erb
+++ b/app/views/organisations/logs.html.erb
@@ -12,6 +12,10 @@
 <h2 class="govuk-visually-hidden">Logs</h2>
 
 <div class="app-filter-layout" data-controller="filter-layout">
+  <div class="govuk-button-group app-filter-toggle">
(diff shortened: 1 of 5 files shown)


app/views/organisations/logs.html.erb
```diff
@@ -12,6 +12,10 @@
 <h2 class="govuk-visually-hidden">Logs</h2>
 
 <div class="app-filter-layout" data-controller="filter-layout">
+  <div class="govuk-button-group app-filter-toggle">
+    <%= govuk_button_to "Create a new lettings log for #{@organisation.name}", case_logs_path(case_log: { owning_organisation_id: @organisation.id }, method: :post) %>
```
paulrobertlloyd: I wonder if this should be ‘Create a new lettings log for this organisation’. Just that some organisation names are looooooooong.

moarpheus: Paul raised a good point. And tbh - there is org name already just above the button, so no ambiguity about what org this log is for.

kosiakkatrina (approved):

PR #683, opened 2022-06-21
CLDC-456-start-date-validations

dushan-madetech:
Adds tenancy start date validations:
- Start date is no later than 14 days from the current date
- The start date is before the scheme end date (if one is present) for a supported housing tenancy

Screenshot 2022-06-23 at 16 29 30

Screenshot 2022-06-23 at 16 29 42


app/models/case_log.rb
```diff
@@ -35,7 +35,8 @@ class CaseLog < ApplicationRecord
   belongs_to :owning_organisation, class_name: "Organisation", optional: true
   belongs_to :managing_organisation, class_name: "Organisation", optional: true
   belongs_to :created_by, class_name: "User", optional: true
-
+  belongs_to :scheme, required: false
```
baarkerlounger: `required` option is deprecated. `optional: true` as above is recommended now I think (https://apidock.com/rails/ActiveRecord/Associations/ClassMethods/belongs_to)

app/models/validations/setup_validations.rb
```diff
@@ -5,6 +5,33 @@ def validate_irproduct_other(record)
     end
   end
 
+  def validate_startdate(record)
+    if record.startdate > Time.zone.today
```
baarkerlounger: Not sure if we want this but commented on the ticket

app/models/validations/setup_validations.rb
```diff
@@ -5,6 +5,33 @@ def validate_irproduct_other(record)
...
+      record.errors.add :startdate, I18n.t("validations.setup.startdate.today_or_earlier")
+    end 
+    
+    if record.voiddate.present?
+      if (record.startdate.to_date - record.voiddate.to_date).to_i.abs > 730
+        record.errors.add :startdate, I18n.t("validations.setup.startdate.voiddate_difference")
+      end 
+    end 
+
+    if record.mrcdate.present?
```
baarkerlounger: Is this not already covered by https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/blob/main/app/models/validations/date_validations.rb#L12?

app/models/validations/setup_validations.rb
```diff
@@ -5,6 +5,33 @@ def validate_irproduct_other(record)
     end
   end
 
+  def validate_startdate(record)
+    if record.startdate > Time.zone.today
+      record.errors.add :startdate, I18n.t("validations.setup.startdate.today_or_earlier")
+    end 
+    
+    if record.voiddate.present?
```
baarkerlounger: Is this not already covered by https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/blob/main/app/models/validations/date_validations.rb#L17?

app/models/validations/setup_validations.rb
```diff
@@ -5,6 +5,33 @@ def validate_irproduct_other(record)
...
+    end 
+
+    if record.mrcdate.present?
+      if (record.startdate.to_date - record.mrcdate.to_date).to_i.abs > 730
+        record.errors.add :startdate, I18n.t("validations.setup.startdate.mrcdate_difference")
+      end 
+    end 
+
+    if record.scheme_id.present?
+      scheme_end_date = Scheme.find(record.scheme_id).end_date
```
baarkerlounger: Given we have an association do we need to do a lookup? Should be able to just do `record.scheme.end_date`?

config/locales/en.yml
```diff
@@ -61,6 +61,9 @@ en:
     setup:
       intermediate_rent_product_name:
         blank: "Enter name of other intermediate rent product"
+      startdate:
+        later_than_14_days_after: "The tenancy start date must be not be later than 14 days from today's date"
+        before_scheme_end_date: "The start date of a supported housing tenancy must be before the supported housing scheme end date"
```
dushan-madetech: @Dan13L3 I can't trigger these on the system locally to provide screenshots so you'll have to review them here

config/initializers/feature_toggle.rb
```diff
@@ -4,4 +4,10 @@ def self.supported_housing_schemes_enabled?
 
     false
   end
+
+  def self.startdate_two_week_validation_enabled?
+    return true if Rails.env.production?
```
baarkerlounger: Can probably just make this true everywhere for now?

config/locales/en.yml
baarkerlounger: Why can't these be triggered locally?

dushan-madetech: The later than 14 days one - I've set this to only trigger on production but will update now

The other one - I think I might be able to actually but I thought I couldn't

config/initializers/feature_toggle.rb
dushan-madetech: Ah, I understood that we only wanted it to be true for production but could toggle it off when needed based on the Slack thread

dushan-madetech: Will update this

config/locales/en.yml
baarkerlounger: Should be able to trigger both now since we'll turn it on everywhere initially?

config/initializers/feature_toggle.rb
dushan-madetech: @baarkerlounger set it to true all the time

config/locales/en.yml
dushan-madetech: @Dan13L3 I've updated the description to show the screenshots I was able to grab when I triggered the errors locally

spec/requests/case_logs_controller_spec.rb
```diff
@@ -275,21 +277,25 @@
...
-                                startdate: Time.zone.local(2022, 12, 1),
-                                tenancy: 6,
-                                managing_organisation: organisation)
+              case_log = FactoryBot.build(:case_log, :completed,
+                                          owning_organisation: organisation,
+                                          mrcdate: Time.zone.local(2022, 2, 1),
+                                          startdate: Time.zone.local(2022, 12, 1),
+                                          tenancy: 6,
+                                          managing_organisation: organisation)
+              case_log.save!(validate: false)
```
baarkerlounger: Any reason not to use the time cop pattern here?

app/models/validations/date_validations.rb
```diff
@@ -34,6 +34,17 @@ def validate_startdate(record)
     if record.startdate < first_collection_start_date || record.startdate > second_collection_end_date
       record.errors.add :startdate, I18n.t("validations.date.outside_collection_window")
     end
+
+    if FeatureToggle.startdate_two_week_validation_enabled? && (record.startdate > Time.zone.today + 14)
+      record.errors.add :startdate, I18n.t("validations.setup.startdate.later_than_14_days_after")
+    end
+
+    if record.scheme_id.present?
+      scheme_end_date = Scheme.find(record.scheme_id).end_date
```
baarkerlounger:
```suggestion
      scheme_end_date = record.scheme.end_date
```

spec/requests/case_logs_controller_spec.rb
dushan-madetech: It's creating both a 21 and 22 case log and I wasn't sure I could freeze timecop to 2 seperate times

baarkerlounger: Shouldn't any '21 log pass that validation anyway? Little worried this might cover other validation failures.

On a related note we may have a bit of a timebomb in our specs in general with the collection window validation? Once the '21 window closes presumably all logs we create in specs become invalid?

baarkerlounger (approved):

spec/requests/case_logs_controller_spec.rb
dushan-madetech: I hadn't considered freezing it to the later date yesterday 🤦

dushan-madetech: Will update to use the timecop pattern

config/initializers/feature_toggle.rb
```diff
@@ -4,4 +4,8 @@ def self.supported_housing_schemes_enabled?
 
     false
   end
+
+  def self.startdate_two_week_validation_enabled?
+    true
```
smeny, line 9: This is enabled for all environments right?

smeny: (Comment was saying production only)

baarkerlounger: Yep, for now this is enabled everywhere. The purpose of the feature flag is to enable it to eventually be turned off in staging when testing new forms changes for next year.

dushan-madetech: @smeny I forgot to update the PR description yesterday when I updated the feature flag

spec/requests/case_logs_controller_spec.rb
dushan-madetech: @baarkerlounger - I've updated it to use the timecop pattern in https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/683/commits/3db034d9b10141bb9514282962a4cfcd9ba6e26a 😄

PR #685, opened 2022-06-21
Fix placement of back links on scheme pages

paulrobertlloyd:
* Place back links on scheme pages before content
* Use default text value for back links


moarpheus (approved):
There will be further changes to the back button from scheme locations page, as you can get caught in the loop. Basically it needs to point to the show page

PR #686, opened 2022-06-21
Make user dependent form content request safe

baarkerlounger:
- We set the current user for the form in the case_log/form controller when creating or updating the case log.
- The form is held by the FormHandler singleton which is created initialised at boot
- As a result a single form object is shared by each case log, which just holds a reference to whichever user last made a request that affects a case log pointing to that form. 
- This is obviously not at all thread safe (where thread means request)

This PR does a couple of things to fix that:

1. Make all form evaluation functional - i.e. if we need to check the user in a method, the user needs to be passed down from the request, no relying on side effects/instance variables
2. Move `derived` to question rather than page level, since it is really the answer to a question that is derived not the page (semantics)
3. Fix the content for the Organisation (header) and Needs Type (hint text) questions

There are a few places where we pass nil rather than an actual user for convenience because we know we don't need to evaluate it but this makes for an inconsistent API. We should attempt to eliminate that but I wanted to keep this PR as small as possible for now and refactor in small steps.

Support User:

- [x] When a new log is created from top level nav bar, they see Organisation and Created By Questions
- [x] Organisation and User appear in check answers
- [x] Organisation and User need to be completed for Setup to be completed

Non-Support user

- [x] When a new log is created they do not see Organisation and Created By Questions
- [x] Organisation and User do not appear in check answers
- [x] Organisation and User are automatically set

### DIFF
--- a/app/helpers/check_answers_helper.rb
+++ b/app/helpers/check_answers_helper.rb
@@ -1,14 +1,32 @@
 module CheckAnswersHelper
   include GovukLinkHelper
 
-  def display_answered_questions_summary(subsection, case_log)
-    total = subsection.applicable_questions_count(case_log)
-    answered = subsection.answered_questions_count(case_log)
+  def display_answered_questions_summary(subsection, case_log, current_user)
+    total = total_count(subsection, case_log, current_user)
+    answered = answered_questions_count(subsection, case_log, current_user)
     if total == answered
       '<p class="govuk-body">You answered all the questions.</p>'.html_safe
     else
       "<p class=\"govuk-body\">You have answered #{answered} of #{total} questions.</p>".html_safe
     end
+  end
+
+private
+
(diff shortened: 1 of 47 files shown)


moarpheus (approved):

app/helpers/check_answers_helper.rb
```diff
@@ -1,16 +1,34 @@
...
     if total == answered
       '<p class="govuk-body">You answered all the questions.</p>'.html_safe
     else
       "<p class=\"govuk-body\">You have answered #{answered} of #{total} questions.</p>".html_safe
     end
   end
 
+private
+
+  def answered_questions_count(subsection, case_log, current_user)
```
baarkerlounger, line 16: These move out of the class and into the helper because they're currently only used for view logic, no where else. (i.e. they're actually specifically for the check answers page).

dushan-madetech (approved):

PR #691, opened 2022-06-22
Search schemes by postcode

kosiakkatrina:
Add a search by postcode scope. The join with locations table means that search_by will only return the results that have associated location with it whenever searching, _but we are not displaying the schemes without location anyway_ - yes we are, but perhaps we shouldn't !

### DIFF
--- a/app/models/scheme.rb
+++ b/app/models/scheme.rb
@@ -5,7 +5,7 @@ class Scheme < ApplicationRecord
 
   scope :search_by_code, ->(code) { where("code ILIKE ?", "%#{code}%") }
   scope :search_by_service_name, ->(name) { where("service_name ILIKE ?", "%#{name}%") }
-  scope :search_by_postcode, ->(postcode) { joins(:locations).where("REPLACE(locations.postcode, ' ', '') ILIKE ?", "%#{postcode.delete(' ')}%") }
+  scope :search_by_postcode, ->(postcode) { joins(:locations).where("locations.postcode ILIKE ?", "%#{postcode.delete(' ')}%") }
   scope :search_by, ->(param) { search_by_postcode(param).or(search_by_service_name(param)).or(search_by_code(param)).distinct }
 
   SCHEME_TYPE = {
--- a/app/models/scheme.rb
+++ b/app/models/scheme.rb
@@ -5,7 +5,8 @@ class Scheme < ApplicationRecord
 
   scope :search_by_code, ->(code) { where("code ILIKE ?", "%#{code}%") }
   scope :search_by_service_name, ->(name) { where("service_name ILIKE ?", "%#{name}%") }
-  scope :search_by, ->(param) { search_by_code(param).or(search_by_service_name(param)) }
+  scope :search_by_postcode, ->(postcode) { joins(:locations).where("REPLACE(locations.postcode, ' ', '') ILIKE ?", "%#{postcode.delete(' ')}%") }
+  scope :search_by, ->(param) { search_by_postcode(param).or(search_by_service_name(param)).or(search_by_code(param)).distinct }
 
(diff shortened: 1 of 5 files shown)


app/models/scheme.rb
```diff
@@ -4,7 +4,8 @@ class Scheme < ApplicationRecord
 
   scope :search_by_code, ->(code) { where("code ILIKE ?", "%#{code}%") }
   scope :search_by_service_name, ->(name) { where("service_name ILIKE ?", "%#{name}%") }
-  scope :search_by, ->(param) { search_by_code(param).or(search_by_service_name(param)) }
+  scope :search_by_postcode, ->(postcode) { joins(:locations).where("REPLACE(locations.postcode, ' ', '') ILIKE ?", "%#{postcode.delete(' ')}%") }
```
baarkerlounger: I wonder if we might be better off removing the space in the data?

spec/requests/organisations_controller_spec.rb
```diff
@@ -76,6 +76,7 @@
           let(:search_param) { "CODE321" }
 
           before do
+            FactoryBot.create(:location, scheme: searched_scheme)
```
moarpheus, line 79: This is probably redundant, unless you explicitly set the the location postcode to something _not_ matching the search param, to signify that matching result is being returned not because it matched the postcode for location. ( same applies for 2 other cases below ).

app/models/scheme.rb
moarpheus: We could store without any spaces and all capitals for example when creating/importing data, part of creating locations for schemes ticket ( it doesnt exist yet, but I will make it before end of the day ).

spec/requests/organisations_controller_spec.rb
kosiakkatrina: search_by will only return the results that have associated location with it whenever searching, because of the join on the tables

moarpheus: I don't think there is a constraint that scheme must a have location, so presumably search have to return results, even if no locations are present

kosiakkatrina: Seems like a scheme should have at least one location associated with it, so filtering out the ones that don't have any within search_by doesn't seem like a massive issue

app/models/scheme.rb
kosiakkatrina: Yup. Sounds good, might look into implementing it as part of this PR

baarkerlounger: If we do (which would be good), we probably also need a rake task we can run to fix production data we already have to that standard (could just run it through console but probably good to start getting into the habit of doing it properly).

paulrobertlloyd:
Have we updated the label on the search input?

app/models/scheme.rb
baarkerlounger: Looks like all production data for logs already normalises postcode values so think we're good here.

app/models/scheme.rb
```diff
@@ -5,7 +5,8 @@ class Scheme < ApplicationRecord
 
   scope :search_by_code, ->(code) { where("code ILIKE ?", "%#{code}%") }
   scope :search_by_service_name, ->(name) { where("service_name ILIKE ?", "%#{name}%") }
-  scope :search_by, ->(param) { search_by_code(param).or(search_by_service_name(param)) }
+  scope :search_by_postcode, ->(postcode) { joins(:locations).where("locations.postcode ILIKE ?", "%#{postcode.delete(' ')}%") }
```
baarkerlounger, line 8: Is the `.delete(' ')` still needed here?

kosiakkatrina: I think so, the user can still search with a space

PR #694, opened 2022-06-23
Remove 2fa flash message

baarkerlounger:


paulrobertlloyd:
Need to remove the localisation string too, for completeness.

baarkerlounger:
@paulrobertlloyd not worth removing the strings I think (unless we do it more thoroughly) - we currently have all localisation strings for devise regardless of whether we currently use them or not in there (means it's easy to add them back if we ever want to).

paulrobertlloyd (approved):
👍 Thanks!

PR #697, opened 2022-06-27
Update to GOV.UK Frontend 4.2.0; use `govukPagination` component

paulrobertlloyd:
* Update `govuk-frontend` to v4.2.0
* Update `govuk-prototype-components` to v0.4.0
* Update module imports for the above packages (issues from #644 appear to have been fixed)
* Update `govuk-components` to v3.1.1, which includes `govuk_pagination`
* Replace bespoke pagination component with that provided by the above and `govuk-frontend` v4.2.0
* Update tests – [design system recommendation](https://design-system.service.gov.uk/components/pagination/#first-and-last-pages) is not to show previous/next page links on first/last pages:

  > Do not show the previous page link on the first page — and do not show the next page link on the last page.

Screenshot of new pagination

(Preview attained by temporarily changing item limit in Pagy’s config)


paulrobertlloyd:
(Am rethinking showing total number of results per page below pagination; seems slightly repetitive given we provide a summary of results above the table, too. For another day…)

app/views/pagy/_nav.html.erb
```diff
@@ -1,41 +1,6 @@
...
-          </span></a>
-      </li>
-    </ul>
-    <p class="app_pagination__results">
-      Showing <b><%= pagy.from %></b> to <b><%= pagy.to %></b> of <b><%= pagy.count %></b> <%= item_name %>
-    </p>
-  </nav>
+  <p class="govuk-body">
+    Showing <b><%= pagy.from %></b> to <b><%= pagy.to %></b> of <b><%= pagy.count %></b> <%= item_name %>
+  </p>
```
peteryates, line 5: Think the component should cover an optional description like this @paulrobertlloyd?

paulrobertlloyd: In the gem component? In the early versions of the component in the design system, I know there was the ability to add this extra info (which is why I included it in our bespoke component), but seems like that got stripped out. Can’t see any docs/code for this in the gem component, either.

peteryates: No the gem one works just like the original. I can make it take an extra argument/block for an accompanying description though.

baarkerlounger (approved):
Well this is just stellar!

app/views/pagy/_nav.html.erb
paulrobertlloyd: I’d keep it in line with whatever the design system component does.

PR #700, opened 2022-06-28
Add care home charges minimum validation

kosiakkatrina:
If the answer to Does the household pay rent or other charges for the accommodation? is Yes, then Care home charge and Total Charge in Household rent and charges should be at least £10 per week.

We already have a validation for total charge, it applies to general needs as well


dushan-madetech (approved):
🚢 it

PR #701, opened 2022-06-28
Cldc 955 skip for now button

Ted-U:


baarkerlounger:
Think the declaration page needs to specifically not have a skip for now link since you have to answer that before you can answer the rest of the subsection.

baarkerlounger (approved):

app/controllers/form_controller.rb
```diff
@@ -120,4 +126,31 @@ def successful_redirect_path
     redirect_path = @case_log.form.next_page_redirect_path(@page, @case_log, current_user)
     send(redirect_path, @case_log)
   end
+
+  def mandatory_questions_with_no_response(responses_for_page)
+    session["fields"] = {}
+    calc_questions = @page.questions.map(&:result_field)
```
dushan-madetech, line 132: What is `calc_questions`?

config/forms/2021_2022.json
```diff
@@ -1228,11 +1226,7 @@
               }
             },
             "females_in_soft_age_range_in_pregnant_household_lead_age_value_check": {
-              "depends_on": [
-                {
-                  "female_in_pregnant_household_in_soft_validation_range?": true
-                }
-              ],
+              "depends_on": [{ "female_in_pregnant_household_in_soft_validation_range?": true }],
```
dushan-madetech, line 1250: Your editor seems to have changed the formatting/indentation. Or did you deliberately change it?

spec/features/form/form_navigation_spec.rb
```diff
@@ -47,11 +56,23 @@
...
+      click_link(text: "Skip for now")
+      expect(page).to have_current_path("/logs/#{empty_case_log.id}/person-1-age")
+    end
+
+    it "routes to check answers when skipping on the last page in the form" do
+      visit("logs/#{empty_case_log.id}/propcode")
+      click_link(text: "Skip for now")
+      expect(page).to have_current_path("/logs/#{empty_case_log.id}/household-characteristics/check-answers")
+    end
+
```
dushan-madetech, line 75: You could wrap this in a context block with a description about skip links

spec/features/form/form_navigation_spec.rb
```diff
@@ -47,11 +56,23 @@
       pages = question_answers.map { |_key, val| val[:path] }
       pages[0..-2].each_with_index do |val, index|
         visit("/logs/#{id}/#{val}")
-        click_button("Save and continue")
+        click_link("Skip for now")
```
dushan-madetech, line 59: Why did this change?

spec/features/form/form_navigation_spec.rb
```diff
@@ -89,4 +110,57 @@
       end
     end
   end
+
+  describe "Editing a log" do
```
dushan-madetech, line 114: This should maybe a context instead of a describe, something like `context "When a user is editing a log" do`

spec/features/form/form_navigation_spec.rb
```diff
@@ -89,4 +110,57 @@
...
+
+    context "when clicking save and continue on a mandatory question with no input" do
+      let(:id) { empty_case_log.id }
+
+      it "shows a validation error on radio questions" do
+        visit("/logs/#{id}/renewal")
+        click_button("Save and continue")
+        expect(page).to have_selector("#error-summary-title")
+        expect(page).to have_selector("#case-log-renewal-error")
+        expect(page).to have_title("Error")
```
dushan-madetech, line 129: Given that these lines are the same as those on 135-137 and 146-148 (minus a few minor changes) you could probably extract a method for this

dushan-madetech (approved):
I left a couple of comments but I don't think it's anything worth blocking this PR over. 🚢 it when you're ready

paulrobertlloyd (changes requested):
Number of issues encountered on first pass:

1. ~~I can’t skip the ‘What is the tenancy start date?’ question, instead I am shown validation. If I enter a value to bypass this question and click ‘Skip for now’, then return to the question from the check your answers page, the validation message is shown. Wondering if this might be an issue with all date based questions.~~
2. Changing an answer should show ‘Save changes’ button alongside the ‘Cancel’ link. The button currently says ‘Save and continue’.
3. ~~Clicking ‘Save and continue’ reloads the same page, still showing a ‘Save and continue’ but now alongside a link that says ‘Skip for now’. This is incorrect, it should submit the form with the (changed) value.~~

I can’t test any further as I can’t select a user to create a log. But feels broken, and more testing needed.

app/controllers/form_controller.rb
baarkerlounger: It's basically the total charges question (i.e. the one that isn't really asked of users but is the the calculated result of the  sum of the other charges)

config/forms/2021_2022.json
baarkerlounger: Was deliberate, made it easier to use code folding to see where to make the changes

spec/features/form/form_navigation_spec.rb
baarkerlounger: Because "Save and continue" now requires you to provide an answer where previously it didn't and we're not providing one here.

baarkerlounger:
@paulrobertlloyd 

1. Can't reproduce this. Do you have more info?
2. :+1: @Ted-U could you take a look at this?
3. Not sure where you're seeing that? How do you reproduce?

paulrobertlloyd:
Was an issue with an old log with incorrect validation on letting start date. Just issue 2. to resolve.

app/models/organisation.rb
```diff
@@ -1,10 +1,10 @@
 class Organisation < ApplicationRecord
-  has_many :users
-  has_many :owned_case_logs, class_name: "CaseLog", foreign_key: "owning_organisation_id"
+  has_many :users, dependent: :destroy
```
baarkerlounger: Why are we adding `dependent: :destroy` here and below? That seems unrelated to the changes in this PR and I assume has been brought in accidentally from your other branch. I'd suggest removing it from here otherwise gets hard to keep the two changes separate in terms of deploying and reverting.

db/schema.rb
```diff
@@ -378,8 +382,10 @@
   end
 
   add_foreign_key "case_logs", "locations"
+  add_foreign_key "case_logs", "organisations", column: "organisations_id"
```
baarkerlounger: Same, would also remove these here as there is no migration in this PR introducing these

app/views/form/page.html.erb
```diff
@@ -40,12 +40,12 @@
       <%= f.hidden_field :page, value: @page.id %>
 
       <div class="govuk-button-group">
-        <% if !@page.id.include?("value_check") %>
+        <% if !@page.id.include?("value_check") && if request.query_parameters["referrer"] != "check_answers" %>
```
baarkerlounger, line 43: Possibly getting complex enough to be worth extracting to a helper that has a spec?

paulrobertlloyd:
All looks good from my perspective.

paulrobertlloyd (approved):

baarkerlounger (changes requested):
DB changes needed before this can be merged

db/schema.rb
```diff
@@ -252,6 +252,7 @@
     t.integer "type_of_unit"
     t.string "old_id"
     t.integer "old_visible_id"
+    t.string "mobility_type"
```
baarkerlounger: As a general rule of thumb you only want to be committing schema changes if you're also committing a migration that introduces them.

baarkerlounger (approved):

PR #702, opened 2022-06-28
Update documentation (content, presentation and images)

paulrobertlloyd:
I kinda went a bit crazy. Good job I’ve got 5 days of holiday to recover…


baarkerlounger:
@paulrobertlloyd what's the reason for swapping the architecture image? How was the new one created/is it editable?

The old one was created with diagrams.net/draw.io and had the draw.io file embedded so it could be opened and edited easily.

paulrobertlloyd:
> What's the reason for swapping the architecture image

Because it looks better? 😉 

> The old one was created with diagrams.net/draw.io and had the draw.io file embedded so it could be opened and edited easily.

Oooh, thus the file size! Happy to try and recreate it there (edibility is important); want to merge in the meantime, or merge without the image change?

baarkerlounger:
@paulrobertlloyd preference would be to merge this without that image change and then have that as a separate PR?

> Because it looks better? wink

That seems highly subjective :joy:

PR #703, opened 2022-06-29
Document personas

baarkerlounger:


paulrobertlloyd:
A few questions:

* Is this the right place for this sort of documentation? As opposed to other DLUHC locations such as Confluence, SharePoint etc.
* Would design related resources be better hosted with, and shown in the context of, the design history?
* And if so (or regardless of such), should the design history site live in this same repo? (I know there are some issues with hosting the design history on a repo in this organisation as it prevents access to the Forestry CMS, but not sure if anyone is actually using that). If not, it should probably return to this organisation at some point, rather than @roobottom's personal account.

baarkerlounger:
@paulrobertlloyd I think this documentation can/should include anything necessary for/helpful to developing the app - it seems like personas could reasonably fall under that?

Also anything that could generally be publicly useful since confluence isn't public.

Whether design history should live here too I'm not sure, it could perhaps just be linked to from these docs?

baarkerlounger:
@paulrobertlloyd any objections to merging?

paulrobertlloyd:
@baarkerlounger Might start a discussion on Slack, just to survey opinion. Does make me feel a bit icky, this one.

baarkerlounger:
Closing this in favour of putting personas in the design blog.

PR #704, opened 2022-06-30
CLDC-1337-permitted-user-can-add-location-when-creating-scheme

moarpheus:
Screenshot 2022-07-05 at 11 26 54
Screenshot 2022-07-05 at 11 26 44
Screenshot 2022-07-05 at 11 26 36


app/controllers/locations_controller.rb
```diff
@@ -0,0 +1,57 @@
...
+      location_params[:add_another_location] == "Yes" ? redirect_to(location_new_scheme_path) : redirect_to(scheme_check_answers_path(@scheme, anchor: "locations"))
+    else
+      render :details, status: :unprocessable_entity
+    end
+  end
+
+private
+
+  def find_scheme
+    @scheme = params[:scheme_id] && params[:id] ? Scheme.find(params[:scheme_id]) : Scheme.find(params[:id])
```
baarkerlounger: Should `params[:id]` ever refer to scheme on a request to the `locations` controller? That seems like it's wrong?

app/controllers/locations_controller.rb
```diff
@@ -0,0 +1,57 @@
...
+
+  def authenticate_action!
+    if %w[new create details update].include?(action_name) && !((current_user.organisation == @scheme.organisation) || current_user.support?)
+      render_not_found and return
+    end
+  end
+
+  def location_params
+    required_params = params.require(:location).permit(:postcode, :name, :total_units, :type_of_unit, :wheelchair_adaptation, :add_another_location).merge(scheme_id: @scheme.id)
+    required_params[:postcode] = required_params[:postcode].delete(" ").upcase.encode("ASCII", "UTF-8", invalid: :replace, undef: :replace, replace: "") if required_params[:postcode]
```
baarkerlounger, line 64: I think we also have this in the CaseLog class where we save postcodes. Might be worth extracting and including?

app/controllers/locations_controller.rb
moarpheus: Good point, will review my routing.

app/models/location.rb
```diff
@@ -1,13 +1,28 @@
 class Location < ApplicationRecord
+  include Validations::PropertyValidations
```
baarkerlounger, line 2: Might also be worth extracting the postcode regexp to somewhere (same place as the postcode normalisation maybe?) and including that instead since it's a bit misleading to include the PropertyValidations module here when we're not using any of the actual validations.

app/models/location.rb
```diff
@@ -1,13 +1,28 @@
...
+  validate :validate_postcode
   belongs_to :scheme
 
+  attr_accessor :add_another_location
```
baarkerlounger, line 6: Should this be a field? Seems more like an action?

moarpheus: I need to access this param part of my form, but it doesn't scream to be a field, more of toggle

app/views/locations/new.html.erb
```diff
@@ -0,0 +1,63 @@
...
+
+      <% type_of_units_selection = Location.type_of_units.keys.map { |key, _| OpenStruct.new(id: key, name: key.to_s.humanize) } %>
+
+      <%= f.govuk_collection_radio_buttons :type_of_unit,
+                                           type_of_units_selection,
+                                           :id,
+                                           :name,
+                                           legend: { text: "What is this type of scheme?", size: "m" } %>
+
+      <% wheelchair_user_selection = Location.wheelchair_adaptations.keys.reverse.map { |key, _| OpenStruct.new(id: key, name: key.to_s.humanize) } %>
```
baarkerlounger: Worth just defining these the right way around in the model since we call `.reverse` everywhere we use it?

moarpheus: Done

config/routes.rb
```diff
@@ -45,6 +45,10 @@
 
     member do
       get "locations", to: "schemes#locations"
```
baarkerlounger: Might be able to just nest `locations` as a resource under `schemes`?

kosiakkatrina (commented):
Is there any reason why the delete text is in the view mode for the locations?

moarpheus:
> Is there any reason why the delete text is in the view mode for the locations?

It is per design, I am not implementing 'delete' just yet though - hence it is not clickable.

kosiakkatrina:
> > Is there any reason why the delete text is in the view mode for the locations?
> 
> It is per design, I am not implementing 'delete' just yet though - hence it is not clickable.

@moarpheus I'd say maybe leave it out for now? As the supported housing flag will probably be removed before this functionality is available and could confuse the user

baarkerlounger (approved):

PR #705, opened 2022-07-01
CLDC-1249 supported housing scheme

smeny:


spec/features/log_spec.rb
```diff
@@ -90,7 +90,7 @@
         visit("logs/#{log_id}/setup/check-answers")
         expect(page).to have_content("Owning organisation #{support_user.organisation.name}")
         expect(page).to have_content("User #{support_user.name}")
-        expect(page).to have_content("You have answered 2 of 7 questions")
```
smeny, line 93: This is all impacted by needs type being now displayed

app/models/form/setup/pages/location.rb
```diff
@@ -0,0 +1,20 @@
...
+    super("location", hsh, subsection)
+    @header = ""
+    @description = ""
+    @questions = questions
+    @depends_on = [{
+      "supported_housing_schemes_enabled?" => true,
+      "needstype" => 2,
+      "scheme_has_multiple_locations?" => true,
+    }]
+    @derived = true
```
smeny: Not sure if it makes sense for the "page" to be derived

baarkerlounger: It doesn't this should be on the `question`

app/models/form/setup/questions/location_id.rb
```diff
@@ -0,0 +1,38 @@
...
+    @type = "radio"
+    @derived = true unless FeatureToggle.supported_housing_schemes_enabled?
+    @answer_options = answer_options
+  end
+
+  def answer_options
+    answer_opts = {}
+    return answer_opts unless ActiveRecord::Base.connected?
+
+    Location.select(:id, :postcode, :address_line1).each_with_object(answer_opts) do |location, hsh|
```
baarkerlounger: Think this should be `name` rather than `address_line1`?

app/models/form/setup/questions/scheme_id.rb
```diff
@@ -0,0 +1,38 @@
...
+    Scheme.select(:id, :service_name).each_with_object(answer_opts) do |scheme, hsh|
+      hsh[scheme.id.to_s] = scheme.service_name
+      hsh
+    end
+  end
+
+  def displayed_answer_options(case_log)
+    return {} unless case_log.created_by
+
+    user_org_scheme_ids = Scheme.where(organisation_id: case_log.created_by.organisation_id).map(&:id).map(&:to_s)
```
baarkerlounger: I think using `select(:id)` as above avoids needing to loop over the results just to extract that?

app/models/form/setup/questions/location_id.rb
smeny: Locations currently do not have names (I will need to remove the address lines).

kosiakkatrina (commented):
I think the way it is right now, it resets location to nil unless there are 2 locations associated with the scheme that's selected

smeny:
Pairing with Kat to finish the remaining issues:
* Resetting location when a scheme is changed
* Derive the location when a single one exists for a scheme

PR #709, opened 2022-07-04
Update shortfall validation

baarkerlounger:
From: https://digital.dclg.gov.uk/jira/browse/MSD-37600

```
Universal Credit works differently from Housing Benefit meaning what the enquirer says does make sense. We could adapt the validation to be based on the type of benefit but I think that's needlessly complicating it.

Let's change the validation to check that the outstanding amount is between £0.01 and the calculated weekly rent (I say calculated because they can enter rent as either a weekly, monthly or annual amount).

If the outstanding amount is below £0.01 then either they should have answered 'no outstanding amount' or the amount is wrong.

If the outstanding amount is above the weekly then either the rent entered, frequency of rent, or outstanding amount is wrong.
```

image

### DIFF
--- a/app/models/form/question.rb
+++ b/app/models/form/question.rb
@@ -173,6 +173,24 @@ class Form::Question
     type == "radio" && RADIO_REFUSED_VALUE[id.to_sym]&.include?(value)
   end
 
+  def suffix_label(case_log)
+    return "" unless suffix
+    return suffix if suffix.is_a?(String)
+
+    label = ""
+
+    suffix.each do |s|
+      condition = s["depends_on"]
+      next unless condition
+
+      answer = case_log.send(condition.keys.first)
+      if answer == condition.values.first
+        label = s["label"]
+      end
+    end
+    label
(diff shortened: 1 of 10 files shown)


app/models/form/question.rb
```diff
@@ -237,12 +237,6 @@ def enabled_inferred_answers(inferred_answers, case_log)
     inferred_answers.filter { |_key, value| value.all? { |condition_key, condition_value| case_log[condition_key] == condition_value } }
   end
 
-  ANSWER_SUFFIX_LABELS = {
```
baarkerlounger, line 240: This was used to translate when we were saving answer strings rather than values, but isn't needed anymore now that we've switched (https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/273/files#diff-2166414bbc981092824071e3e02784bc25f12a2888f8931cdd463e7693ab407bR2329)

smeny (approved):

PR #718, opened 2022-07-08
CLDC-1329: Add plausible analytics script

baarkerlounger:


app/views/layouts/application.html.erb
```diff
@@ -33,6 +33,10 @@
         console.log(<%= session.to_json.html_safe %>)
       </script>
     <% end %>
+
+    <% if Rails.env.production? %>
+      <script defer data-domain=<% ENV["APP_HOST"].split("https://")[1] %> src="https://plausible.io/js/plausible.js"></script>
```
baarkerlounger: Could vendor this to avoid needing to make an extra request but I think the risk of Plausible changing the script might be greater...

dushan-madetech (approved):

PR #721, opened 2022-07-08
Small refactorings

moarpheus:


app/controllers/locations_controller.rb
```diff
@@ -70,7 +70,7 @@ def authenticate_action!
 
   def location_params
     required_params = params.require(:location).permit(:postcode, :name, :total_units, :type_of_unit, :wheelchair_adaptation, :add_another_location).merge(scheme_id: @scheme.id)
-    required_params[:postcode] = required_params[:postcode].delete(" ").upcase.encode("ASCII", "UTF-8", invalid: :replace, undef: :replace, replace: "") if required_params[:postcode]
+    required_params[:postcode] = PostcodeService.clean(required_params[:postcode].delete(" ").upcase) if required_params[:postcode]
```
baarkerlounger: Looks like the `.delete(" ")` already happens in PostcodeService so probably isn't needed here? Maybe the `.upcase` should go there too? I assume we want that everywhere.

config/initializers/postcode_regex.rb
```diff
@@ -0,0 +1 @@
+POSTCODE_REGEXP = /^(([A-Z]{1,2}[0-9][A-Z0-9]?|ASCN|STHL|TDCU|BBND|[BFS]IQQ|PCRN|TKCA) ?[0-9][A-Z]{2}|BFPO ?[0-9]{1,4}|(KY[0-9]|MSR|VG|AI)[ -]?[0-9]{4}|[A-Z]{2} ?[0-9]{2}|GE ?CX|GIR ?0A{2}|SAN ?TA1)$/i
```
baarkerlounger, line 1: Would have added this to the PostcodeService I think just in terms of bundling together related functionality but I suppose this matches the pattern we use for dates.

baarkerlounger (approved):

PR #722, opened 2022-07-11
Form documentation

dushan-madetech:
Adding documenation for the:

- form
- sections
- subsections
- pages
- questions 

To explain how these work and interact with each other as well as any keys/values they may take


docs/form_builder.md
```diff
@@ -181,6 +181,10 @@ rake form_definition:validate_all
 
 This will validate all forms in directories `["config/forms", "spec/fixtures/forms"]`
 
+## Form models and definition
```
baarkerlounger, line 184: Worth adding this to the `Form Runner` doc as well I think since your documentation kind of touches on parts of both

baarkerlounger (approved):
Very nice!

docs/form_builder.md
dushan-madetech: Done 😄

docs/form_builder.md
```diff
@@ -181,6 +181,10 @@ rake form_definition:validate_all
...
+
+For information about the form model and related models (section, subsection, page, question) and how these relate to each other follow [this link](docs/form/form.md)
```
baarkerlounger, line 186: Actually this link doesn't work because you're using relative path. Think it needs to be either `form/form.md` or /docs/form/form.md`)

PR #724, opened 2022-07-11
Update documentation for schemes

paulrobertlloyd:
Update documentation for schemes to describe how they work in the new world.


baarkerlounger (approved):
Generally :+1: but I think it's worth leaving in the part about these acting as pre-defined answer templates (or something to that effect)

baarkerlounger (approved):

PR #726, opened 2022-07-11
Cldc 1315 set foreign key constraint on users

Ted-U:


app/models/organisation.rb
```diff
@@ -1,11 +1,11 @@
 class Organisation < ApplicationRecord
-  has_many :users
-  has_many :owned_case_logs, class_name: "CaseLog", foreign_key: "owning_organisation_id"
-  has_many :managed_case_logs, class_name: "CaseLog", foreign_key: "managing_organisation_id"
+  has_many :users, dependent: :destroy
+  has_many :owned_case_logs, class_name: "CaseLog", foreign_key: "owning_organisation_id", dependent: :destroy
+  has_many :managed_case_logs, class_name: "CaseLog", foreign_key: "managing_organisation_id", dependent: :destroy
```
baarkerlounger, line 14: Don't think we want it here. I'm not sure deleting a managing organisation implies the owning organisation's logs should be deleted.

app/models/organisation.rb
```diff
@@ -1,11 +1,11 @@
...
   has_many :data_protection_confirmations
   has_many :organisation_rent_periods
-  has_many :owned_schemes, class_name: "Scheme", foreign_key: "stock_owning_organisation_id"
-  has_many :managed_schemes, class_name: "Scheme"
+  has_many :owned_schemes, class_name: "Scheme", foreign_key: "stock_owning_organisation_id", dependent: :destroy
+  has_many :managed_schemes, class_name: "Scheme", dependent: :destroy
```
baarkerlounger, line 18: Same here

app/models/organisation.rb
Ted-U: I have deleted the "dependent: :destroy" and pushed again.

baarkerlounger:
@Ted-U tests and linter failing here

db/schema.rb
```diff
@@ -10,7 +10,11 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
+<<<<<<< HEAD
```
baarkerlounger: Merge conflict markers

baarkerlounger: More below

baarkerlounger:
Doesn't seem to work as expected.

**Test case:** 

Create an organisation with a user, a case log and a scheme.
Run SQL query: `DELETE FROM organisations WHERE id = <org_id>;`

**Expected result:**

The user, case log and scheme are also deleted

**Actual result:**

You can't run the query because the scheme foreign key prevents it being deleted. If you first delete the scheme the user and case log are left with dangling references to an Organisation that no longer exists.

baarkerlounger:
Superseded by https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/765

PR #727, opened 2022-07-11
CLDC-1228 Scheme and Location imports

smeny:
* Rename the import folder
* Rename total_units to units
* Added import logic for schemes and locations


app/services/imports/scheme_import_service.rb
```diff
@@ -0,0 +1,61 @@
...
+  class SchemeImportService < ImportService
+    def create_schemes(folder)
+      import_from(folder, :create_scheme)
+    end
+
+    def create_scheme(xml_document)
+      old_id = string_or_nil(xml_document, "id")
+      status = string_or_nil(xml_document, "status")
+
+      return unless status == "Approved"
```
baarkerlounger: Worth logging scheme files that were processed but not imported due to status?

app/services/imports/scheme_location_import_service.rb
```diff
@@ -0,0 +1,166 @@
...
+    def create_scheme_locations(folder)
+      import_from(folder, :create_scheme_location)
+    end
+
+    def create_scheme_location(xml_document)
+      management_group = location_field_value(xml_document, "mgmtgroup")
+      schemes = Scheme.where(old_id: management_group)
+      raise "Scheme not found with legacy ID #{management_group}" if schemes.empty?
+
+      if schemes.size == 1 && schemes.first&.locations&.empty?
```
baarkerlounger: Do you need the first `&.` here? Surely if `schemes.size == 1` we know we have a scheme value?

app/services/imports/scheme_location_import_service.rb
```diff
@@ -0,0 +1,166 @@
...
+      else
+        scheme = find_scheme_to_merge(schemes, xml_document)
+        scheme ||= duplicate_scheme(schemes, xml_document)
+      end
+      add_location(scheme, xml_document)
+    end
+
+  private
+
+    REGISTERED_UNDER_CARE_ACT = {
```
baarkerlounger, line 23: Worth referencing the constant from the model here so that the values can't get out of sync?

baarkerlounger:
Generally LGTM but think I'd appreciate a walk through to check I've understood everything properly

app/services/imports/scheme_location_import_service.rb
smeny: They are not the same values.
This is the label we use as a suffix on the scheme name.

app/services/imports/scheme_location_import_service.rb
smeny: Good point

app/services/imports/scheme_import_service.rb
smeny: Will do now

PR #728, opened 2022-07-11
Add accessible autocomplete improvements

kosiakkatrina:
This is one part of autocomplete improvements, it includes:
- Autocomplete searches do not perform exact match, i.e ‘kenilworth care’ would return ‘Kenilworth elderly care home’
- Autocomplete searches ignore punctuation, i.e. ‘big reds care’ would return ‘Big Red’s Care Housing’
- Autocomplete searches ignore stop words, i.e. ‘the housing trust’ would return ‘Housing Trust Limited’

It does not include:
- Autocomplete options can provide synonyms to return results for, i.e. the postcodes in a scheme, i.e. ‘BN1 1AA’ would return ‘Brighton Housing Trust’
- Autocomplete options can display appended text (with original option shown in bold), i.e. number of properties in a scheme
- Autocomplete options can display a hint, i.e. the client group(s) used by a scheme

### DIFF
--- a/spec/fixtures/forms/2021_2022.json
+++ b/spec/fixtures/forms/2021_2022.json
@@ -326,7 +326,7 @@
                       "E07000178": "Oxford",
                       "E07000114": "Thanet",
                       "E09000033": "Westminster",
-                      "E06000014": "York"
+                      "E06000014": "The one and only york town"
                     }
                   }
                 },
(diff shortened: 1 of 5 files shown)


spec/fixtures/forms/2021_2022.json
```diff
@@ -326,7 +326,7 @@
                       "E07000178": "Oxford",
                       "E07000114": "Thanet",
                       "E09000033": "Westminster",
-                      "E06000014": "York"
+                      "E06000014": "The one and only york town"
```
baarkerlounger, line 329: :joy:

baarkerlounger:
Looks like this currently breaks converting the select dropdown to type ahead for the organisation log filter that support users get

kosiakkatrina:
> Looks like this currently breaks converting the select dropdown to type ahead for the organisation log filter that support users get

@baarkerlounger, yup, I think it was because there was nothing selected/entered by default, so there were no matches to begin with. Added a check for that now

baarkerlounger (approved):

PR #729, opened 2022-07-11
Generate website for technical docs

paulrobertlloyd:
Using the [Just the docs](https://github.com/just-the-docs/just-the-docs) remote Jekyll theme (used by the ViewComponent project, among others) to generate a website from the documentation in the `/docs` folder.

This makes it easier to see the relationship between different pages, provides linkable URLs, and makes the docs a bit more presentable.

Also:

* Moves API browser to `/api`
* Consolidates service overview on home page
* Reorganises some files, including those around generating the form. This change needs review as still not clear to me how some of these pieces relate to each other, but have tried to move the general overview into its own page, below which sit pages on the runner, builder and then the definitions @dushan-madetech)

Can view the site locally using `cd docs && bundle exec jekyll serve`

Screenshot 2022-07-11 at 17 24 22


README.md
```diff
@@ -5,28 +5,13 @@
 
 Ruby on Rails app that handles the submission of lettings and sales of social housing data in England. Currently in private beta.
 
-## Domain documentation
+## Useful resources
```
baarkerlounger, line 8: Think the original is more descriptive here?

baarkerlounger:
> Can view the site locally using cd docs && bundle exec jekyll serve
Is this documented anywhere in the er... documentation? 

When not local this still runs off Github pages I assume?

paulrobertlloyd:
> When not local this still runs off Github pages I assume?

Yep. Strictly speaking don’t need a `Gemfile` or `Gemfile.lock`, but do if want to view site locally.

baarkerlounger:
Also worth noting those ADRs aren't super up to date, we definitely missed a few so maybe an action there to go and backfill them. Off the top off my head at least:

- Dropping `Webpacker` for `jsbundling-rails`, `webpack` and `propshaft`
- Supporting IE11 (polyfills, babel config etc)
- Dropping Turbo
- Dropping Admin Users and the Active Admin Panel
- Moving the Setup section out of form definitions and into code
- S3 import/export mechanism

paulrobertlloyd:
> Also worth noting those ADRs aren't super up to date

Shouldn’t ADRs also have additional metadata, such as date accepted, status etc.?

baarkerlounger:
Probably also want to include everything in docs in `.cfignore` to not deploy it. I thought that was already the case but looks like it may not be...

baarkerlounger:
> Shouldn’t ADRs also have additional metadata, such as date accepted, status etc.?

We've not really had a formal enough process around these to warrant that I think. They we're more a sort of history of decisions as made (and as a result kind of need to be read from start to finish since any individual ADR could be superseded by a subsequent one). Might also be worth dropping entirely in favour of just having all the "as-is" documented well and not worrying as much about how we got there but I suppose that loses some of the knowledge of what we've already tried.

baarkerlounger (approved):

.gitignore
```diff
@@ -28,6 +28,12 @@
 # Ignore master key for decrypting credentials and more.
 /config/master.key
 
+# Ignore generated documentation files
+_site
```
baarkerlounger, line 32: Probably also want these in `.dockerignore` and `.cfignore`?

README.md
```diff
@@ -7,26 +7,11 @@ Ruby on Rails app that handles the submission of lettings and sales of social ho
 
 ## Domain documentation
```
dushan-madetech, line 8: I'd maybe just add a little bit to the main readme about how to spin up the documentation website locally

docs/form/builder.md
```diff
@@ -181,10 +160,6 @@ rake form_definition:validate_all
 
 This will validate all forms in directories `["config/forms", "spec/fixtures/forms"]`
 
-## Form models and definition
-
-For information about the form model and related models (section, subsection, page, question) and how these relate to each other follow [this link](/docs/form/form.md)
-
```
dushan-madetech: I would keep this but just link it to the definition.md instead

docs/form/runner.md
```diff
@@ -19,7 +24,3 @@ ERB templates:
 Routes for each form page are generated by looping over each Page instance in each Form instance held by the form handler and defining a `GET` path. The corresponding controller method is also auto-generated with meta-programming via the same looping in `app/controllers/form_controller.rb`
 
 All form pages submit to the same controller method (`app/controllers/form_controller.rb#submit_form`) which validates and persists the data, and then redirects to the next form page that identifies as `routed_to` given the current case log state.
-
-## Form models and definition
-
-For information about the form model and related models (section, subsection, page, question) and how these relate to each other follow [this link](/docs/form/form.md)
```
dushan-madetech: Same comment as on the builder.md

dushan-madetech (approved):
I like the look of the this a lot! 😄

PR #730, opened 2022-07-12
CLDC-1348 basic validation for scheme and location + front update

moarpheus:
Screenshot 2022-07-14 at 09 02 36
Screenshot 2022-07-14 at 09 02 25
Screenshot 2022-07-14 at 09 02 18
Screenshot 2022-07-14 at 09 02 10
Screenshot 2022-07-14 at 09 01 58
Screenshot 2022-07-14 at 09 01 48
Screenshot 2022-07-14 at 09 01 29

### DIFF
--- a/app/controllers/schemes_controller.rb
+++ b/app/controllers/schemes_controller.rb
@@ -25,9 +25,9 @@ class SchemesController < ApplicationController
   end
 
   def create
-    @scheme = Scheme.new(scheme_params.except(:support_services_provider_before_type_cast))
+    @scheme = Scheme.new(scheme_params)
 
-    validation_errors scheme_params.except(:support_services_provider)
+    validation_errors scheme_params
 
     if @scheme.errors.empty? && @scheme.save
       if scheme_params[:support_services_provider].zero?
@@ -46,9 +46,9 @@ class SchemesController < ApplicationController
     check_answers = params[:scheme][:check_answers]
     page = params[:scheme][:page]
 
-    validation_errors scheme_params.except(:support_services_provider)
+    validation_errors scheme_params
 
-    if @scheme.errors.empty? && @scheme.update(scheme_params.except(:support_services_provider_before_type_cast))
+    if @scheme.errors.empty? && @scheme.update(scheme_params)
       if check_answers
         if confirm_secondary_page? page
           redirect_to scheme_secondary_client_group_path(@scheme, check_answers: "true")
@@ -175,7 +175,7 @@ private
     if current_user.data_coordinator?
       full_params[:owning_organisation_id] = current_user.organisation_id
     end
-    full_params
+    full_params.except(:support_services_provider_before_type_cast)
   end
 
   def search_term
--- a/app/controllers/schemes_controller.rb
+++ b/app/controllers/schemes_controller.rb
@@ -26,8 +26,9 @@ class SchemesController < ApplicationController
 
   def create
     @scheme = Scheme.new(scheme_params)
+
     if @scheme.save
-      if @scheme.support_services_provider.eql?"The same organisation that owns the housing stock"
+      if scheme_params[:support_services_provider] == "The same organisation that owns the housing stock"
         redirect_to scheme_primary_client_group_path(@scheme)
       else
         redirect_to scheme_support_services_provider_path(@scheme)
@@ -135,7 +136,11 @@ private
     when "support"
       new_location_path
     when "details"
-      scheme_support_services_provider_path(@scheme)
+      if @scheme.support_services_provider.eql?"The same organisation that owns the housing stock"
+        scheme_primary_client_group_path(@scheme)
+      else
+        scheme_support_services_provider_path(@scheme)
+      end
     when "edit-name"
       scheme_path(@scheme)
     end
@@ -156,11 +161,15 @@ private
                                                      :support_services_provider,
                                                      :intended_stay)
 
-    required_params[:sensitive] = required_params[:sensitive].to_i if required_params[:sensitive]
+    same_org_providing_support = required_params[:support_services_provider] == "The same organisation that owns the housing stock"
+
+    full_params = same_org_providing_support ? required_params.merge(managing_organisation_id: required_params[:owning_organisation_id]) : required_params
+
+    full_params[:sensitive] = full_params[:sensitive].to_i if full_params[:sensitive]
     if current_user.data_coordinator?
-      required_params[:owning_organisation_id] = current_user.organisation_id
+      full_params[:owning_organisation_id] = current_user.organisation_id
     end
-    required_params
+    full_params
   end
 
   def search_term
--- a/app/controllers/schemes_controller.rb
+++ b/app/controllers/schemes_controller.rb
@@ -60,6 +60,14 @@ class SchemesController < ApplicationController
   end
 
   def primary_client_group
+    case
+    when request.referrer.include?("new") || request.referrer.include?("details")
+      @back_button_path = scheme_details_path(@scheme)
+    when request.referrer.include?("provider")
+      @back_button_path = scheme_support_services_provider_path(@scheme)
+    when request.query_parameters["check_answers"]
+      @back_button_path = scheme_check_asnwers_path(@scheme)
+    end
     render "schemes/primary_client_group"
   end
 
@@ -87,6 +95,10 @@ class SchemesController < ApplicationController
     render "schemes/edit_name"
   end
 
+  def support_services_provider
+    render "schemes/support_services_provider"
+  end
+
 private
 
   def confirm_secondary_page?(page)
@@ -97,21 +109,23 @@ private
     case
     when page.include?("primary")
       "schemes/primary_client_group"
-    when page.include?("primary")
+    when page.include?("confirm")
       "schemes/confirm_secondary"
-    when page.include?("primary")
+    when page.include?("secondary-client")
       "schemes/secondary_client_group"
-    when page.include?("primary")
+    when page.include?("support")
       "schemes/support"
-    when page.include?("primary")
+    when page.include?("details")
       "schemes/details"
-    when page.include?("primary")
+    when page.include?("edit")
       "schemes/edit_name"
     end
   end
 
   def next_page_path(page)
     case page
+    when "support-services-provider"
+      scheme_primary_client_group_path(@scheme)
     when "primary-client-group"
       scheme_confirm_secondary_client_group_path(@scheme)
     when "confirm-secondary"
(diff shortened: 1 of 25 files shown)


app/views/locations/new.html.erb
```diff
@@ -20,7 +20,7 @@
                              width: 5 %>
 
       <%= f.govuk_text_field :name,
-                             label: { text: "Location name (optional)", size: "m" },
+                             label: { text: "Name (optional)", size: "m" },
```
baarkerlounger: Why is this reverting? Didn't we just change this?

app/views/schemes/details.html.erb
```diff
@@ -61,13 +40,32 @@
...
-        :name,
-        legend: { text: "Is this scheme registered under the Care Standards Act 2000?", size: "m" } %>
+                                           @scheme.care_acts_options_with_hints,
+                                           :id,
+                                           :name,
+                                           :description,
+                                           legend: { text: "Is this scheme registered under the Care Standards Act 2000?", size: "m" },
+                                           bold_labels: false %>
+
+      <% organisations = Organisation.all.map { |org| OpenStruct.new(id: org.id, name: org.name) } %>
```
baarkerlounger, line 55: Have we dropped the null option here? Should we have?

app/views/schemes/details.html.erb
```diff
@@ -61,13 +40,32 @@
...
+      <% if current_user.support? %>
+        <%= f.govuk_collection_select :owning_organisation_id,
+                                      organisations,
+                                      :id,
+                                      :name,
+                                      label: { text: "Which organisation owns the housing stock for this scheme?", size: "m" },
+                                      "data-controller": %w[accessible-autocomplete conditional-filter] %>
+      <% end %>
+
+      <% support_services_provider_selection = Scheme.support_services_providers.keys.map { |key, _| OpenStruct.new(id: key, name: key.to_s.humanize) } %>
```
baarkerlounger: Same question here

app/views/schemes/new.html.erb
```diff
@@ -59,12 +44,30 @@
...
+      <% if current_user.support? %>
+        <%= f.govuk_collection_select :owning_organisation_id,
+                                      answer_options,
+                                      :id,
+                                      :name,
+                                      label: { text: "Which organisation owns the housing stock for this scheme?", size: "m" },
+                                      "data-controller": %w[accessible-autocomplete conditional-filter] %>
+      <% end %>
+
+      <% support_services_provider_selection = Scheme.support_services_providers.keys.map { |key, _| OpenStruct.new(id: key, name: key.to_s.humanize) } %>
```
baarkerlounger: Same question here

app/views/schemes/support.html.erb
```diff
@@ -14,19 +14,22 @@
...
-        :name,
-        legend: { text: "Level of support given", size: "m" } %>
+                                           @scheme.support_level_options_with_hints,
+                                           :id,
+                                           :name,
+                                           :description,
+                                           legend: { text: "Level of support given", size: "m" },
+                                           bold_labels: false %>
 
       <% intended_stay_selection = Scheme.intended_stays.keys.excluding("Missing").map { |key, _| OpenStruct.new(id: key, name: key.to_s.humanize) } %>
```
baarkerlounger: Same for all the select questions basically

app/views/schemes/support_services_provider.html.erb
```diff
@@ -0,0 +1,36 @@
+<% content_for :title, "Which organisation provides the support services used by this scheme?" %>
+
+<% content_for :before_content do %>
+  <%= govuk_back_link(
+    text: "Back",
+    href: request.query_parameters["check_answers"] ? "/schemes/#{@scheme.id}/check-answers" : "/schemes/#{@scheme.id}/details",
```
baarkerlounger, line 6: We have this logic in the view here but in the controller on the other one. Don't mind either approach but should probably be consistent about it?

app/views/schemes/details.html.erb
moarpheus: This is not an optional field and this page is never accessed without existing form ( e.g. you can't get there without populating this field in the first place ).

app/views/locations/new.html.erb
moarpheus: Its per prototype, can you confirm this please @paulrobertlloyd

app/views/schemes/support_services_provider.html.erb
moarpheus: Moved back logic to front, controller is already fat enough

app/views/schemes/new.html.erb
moarpheus: Moved to the view

app/views/schemes/support.html.erb
moarpheus: Moved to the view

app/views/schemes/details.html.erb
moarpheus: Moved to the view

paulrobertlloyd:
Couple of remaining issues spotted:

1. If the user is a data coordinator, the first option should say ‘Your organisation’ instead of ‘The same organisation that owns the housing stock’. Support users should only see the label shown currently.

    Screenshot 2022-07-14 at 10 10 55

2. Label for location name should say: ‘Location name (optional)’

    Screenshot 2022-07-14 at 10 11 26

3. When checking the answers for a scheme, all fields should be editable, and the change links should take the user to the view that allows all fields to be edited. Editing only becomes restricted once a scheme has been saved.

    Screenshot 2022-07-14 at 10 11 53

4. On the page for selecting the support organisation: 

    1. The page should use two-thirds layout (like other forms)
    2. The page title should also be the form’s label (Which organisation manages this scheme should not be shown)

    Screenshot 2022-07-14 at 10 12 52

app/controllers/schemes_controller.rb
```diff
@@ -162,17 +162,24 @@ def scheme_params
                                                      :secondary_client_group,
                                                      :support_type,
                                                      :support_services_provider,
+                                                     :support_services_provider_before_type_cast,
                                                      :intended_stay)
 
-    same_org_providing_support = required_params[:support_services_provider] == "The same organisation that owns the housing stock"
+    same_org_providing_support = required_params[:support_services_provider_before_type_cast] == "0"
```
baarkerlounger: What about:

```ruby
 required_params = params.require(:scheme).permit(:service_name,
                                                     :sensitive,
                                                     :owning_organisation_id,
                                                     :managing_organisation_id,
                                                     :scheme_type,
                                                     :registered_under_care_act,
                                                     :id,
                                                     :has_other_client_group,
                                                     :primary_client_group,
                                                     :secondary_client_group,
                                                     :support_type,
                                                     :support_services_provider_before_type_cast,
                                                     :intended_stay)
 params = required_params.except("support_services_before_type_cast")
 if required_params["support_services_provider_before_type_cast"]
    params.merge(Scheme::SUPPORT_SERVICES_PROVIDER.invert[required_params["support_services_provider_before_type_cast"]]) 
end
params
```

Then you can use `scheme_params` as before everywhere else in the controller without needing to change anything else?

moarpheus: I will most certainly try this today

app/controllers/schemes_controller.rb
```diff
@@ -118,13 +161,19 @@ def scheme_params
...
                                                      :support_type,
-                                                     :intended_stay)
+                                                     :support_services_provider,
+                                                     :support_services_provider_before_type_cast,
+                                                     :intended_stay).merge(support_services_provider: params[:scheme][:support_services_provider_before_type_cast])
+
+    full_params = required_params[:support_services_provider] == "0" && required_params[:owning_organisation_id].present? ? required_params.merge(managing_organisation_id: required_params[:owning_organisation_id]) : required_params
+
+    full_params[:sensitive] = full_params[:sensitive].to_i if full_params[:sensitive]
+    full_params[:support_services_provider] = full_params[:support_services_provider].to_i if !full_params[:support_services_provider]&.empty?
```
baarkerlounger: Can make this `.present?` I think and then you don't need the nil safe operator or the double negation?

app/controllers/schemes_controller.rb
```diff
@@ -118,13 +161,19 @@ def scheme_params
                                                      :primary_client_group,
                                                      :secondary_client_group,
                                                      :support_type,
-                                                     :intended_stay)
+                                                     :support_services_provider,
+                                                     :support_services_provider_before_type_cast,
+                                                     :intended_stay).merge(support_services_provider: params[:scheme][:support_services_provider_before_type_cast])
+
+    full_params = required_params[:support_services_provider] == "0" && required_params[:owning_organisation_id].present? ? required_params.merge(managing_organisation_id: required_params[:owning_organisation_id]) : required_params
```
baarkerlounger, line 174: What is 0 here? Can this be a lookup or a named constant or something?

app/controllers/schemes_controller.rb
```diff
@@ -25,9 +25,16 @@ def new
   end
 
   def create
-    @scheme = Scheme.new(scheme_params)
-    if @scheme.save
-      render "schemes/primary_client_group"
+    @scheme = Scheme.new(scheme_params.except(:support_services_provider_before_type_cast))
```
baarkerlounger: If you move the `.except` down into the params method you can dry this up a bit still?

app/models/scheme.rb
```diff
@@ -17,39 +17,37 @@ class Scheme < ApplicationRecord
...
 
   SCHEME_TYPE = {
-    "Missing": 0,
-    "Foyer": 4,
     "Direct Access Hostel": 5,
-    "Other Supported Housing": 6,
+    "Foyer": 4,
     "Housing for older people": 7,
+    "Other Supported Housing": 6,
+    "Missing": 0,
```
baarkerlounger, line 33: Why do we explicitly need a missing option? Is this equivalent to something other than null?

app/views/schemes/primary_client_group.html.erb
```diff
@@ -1,9 +1,17 @@
 <% content_for :title, "What client group is this scheme intended for?" %>
 
+<% if request.referer&.include?("new") || request.referer&.include?("details") %>
```
baarkerlounger, line 3: This should probably go in a helper?

app/models/scheme.rb
moarpheus: It was in the list of types in the doc, I presumed this would required on import. Will be happy to remove if deemed redundant.

baarkerlounger: @smeny what do you think?

baarkerlounger (approved):

PR #733, opened 2022-07-12
Move doc website docs to do website

baarkerlounger:


docs/documentation_website.md
```diff
@@ -0,0 +1,21 @@
+---
+nav_order: 10
+---
+
+# This documentation website
```
paulrobertlloyd, line 5:
```suggestion
# Documentation
```

docs/documentation_website.md
```diff
@@ -0,0 +1,21 @@
...
+
+### Running documentation locally
```
paulrobertlloyd:
```suggestion
## Running this technical documentation site locally
```

docs/documentation_website.md
```diff
@@ -0,0 +1,21 @@
+---
+nav_order: 10
+---
+
+# This documentation website
+
```
paulrobertlloyd, line 6: Maybe some preamble about where different documentation can be found (perhaps linking to other sources, Confluence, Sharepoint, design history etc.)

docs/documentation_website.md
baarkerlounger: This page is specifically intended to document this jekyll website documentation setup rather than being a guide to all documentation in general so think the original captures that more accurately?

docs/documentation_website.md
baarkerlounger: Same as above - think that's out of scope here but we could add a new page for it if you think it would be valuable?

paulrobertlloyd: In which case, maybe just have a single title for this page.

paulrobertlloyd (approved):

PR #734, opened 2022-07-12
Location name label

baarkerlounger:
image

image


sona-mhclg:
Is that 2 options? If so the one with the postcode is great.

baarkerlounger:
@sona-mhclg no it's the change applied to both views you see it on. The first is the form you see when you're creating a new location as part of a new scheme, the second is a screenshot from when you click into an existing location.

sona-mhclg:
Ah got it. All looks good.

PR #735, opened 2022-07-12
update unittype mapping when infering the case log value

kosiakkatrina:


app/models/derived_variables/case_log_variables.rb
```diff
@@ -5,8 +5,8 @@ module DerivedVariables::CaseLogVariables
     "Self-contained flat or bedsit with common facilities" => 2,
     "Shared flat" => 3,
     "Shared house or hostel" => 4,
-    "Bungalow" => 5,
-    "Self-contained house" => 6,
+    "Bungalow" => 6,
```
baarkerlounger: Could we just reference the location model constant here rather than duplicating it so they can't get out of sync again?

kosiakkatrina: I mean, we could do, but this actually represents how we should be saving the data in the case log (coming from data dictionary). So when we get a text value, we know how to map it to an integer. If the mapping changes (for whatever reason) in the locations, we still want to use this mapping that we have here. The reason it failed before is because this mapping was just wrong. If this question was in the JSON I'd probably derive the values from there, but we never ask this question, it's always inferred, that's why the mapping is here

baarkerlounger: I guess the point is they have to be the same right? (They're not just coincidentally the same values, they represent the same thing), so we should have a single source of truth for them.

If the values in Locations changes and this doesn't, this breaks?

kosiakkatrina: No. I wouldn't think so, what do you mean by breaks? It would take the string value that we get from the `location.type_of_unit` and map it to the integer that corresponds to it in the case log. It doesn't really matter what numbers it's mapping to in the locations table

moarpheus: It wont technically break, but it will be inconsistent with Location types. e.g. If in Locations we will remove option "Shared flat" you will still be able to create log with this location type which is incorrect.

kosiakkatrina: tweaked it a bit now, it's never a quickfix...

app/models/case_log.rb
```diff
@@ -460,6 +460,10 @@ def age_unknown?(person_num)
     public_send("age#{person_num}_known") == 1
   end
 
+  def unittype_sh
+    return location.type_of_unit_before_type_cast if location
```
baarkerlounger: `return` isn't really needed here

app/models/derived_variables/case_log_variables.rb
```diff
@@ -83,7 +75,6 @@ def set_derived_fields!
       if location
         self.la = location.county
         self.postcode_full = location.postcode
-        self.unittype_sh = TYPE_OF_UNIT_MAP[location.type_of_unit]
         self.builtype = form.questions.find { |x| x.id == "builtype" }.answer_options.find { |_key, value| value["value"] == location.type_of_building }.first
```
baarkerlounger: Did we want to remove the `builtype` stuff also?

kosiakkatrina: Yes, different commit, sorry!

baarkerlounger (approved):

PR #739, opened 2022-07-13
add synonyms, appended text and hint text to the autocomplete

kosiakkatrina:
Add improvements to the accessible autocomplete:
- Add matching on the postcodes associated with the scheme locations
- Add appended text that displays the number of locations
- Add the hint text that displays primary and secondary client groups

image

### DIFF
--- a/app/models/form/question.rb
+++ b/app/models/form/question.rb
@@ -136,7 +136,11 @@ class Form::Question
               labels = answer_options[value.to_s]
               labels["value"] if labels
             when "select"
-              answer_options[value.to_s]
+              if answer_options[value.to_s].respond_to?(:service_name)
+                answer_options[value.to_s].service_name
+              else
+                answer_options[value.to_s]
+              end
             else
               value.to_s
             end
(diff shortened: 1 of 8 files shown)


app/models/form/question.rb
```diff
@@ -191,6 +191,25 @@ def suffix_label(case_log)
     label
   end
 
+  def answer_option_synonyms(answer_id)
```
baarkerlounger: Not sure how performant this'll be. We're doing (3 * number of active schemes in the organisation) + 1 database calls here just to render the type ahead question. Might be better to have these methods take a model as it's argument and expect the model to provide the hint, and synonym values.

That way you can have a single database call to fetch all the required records, and it's easier to extend these methods to other things than schemes without building up big if/case statements?

spec/features/form/accessible_autocomplete_spec.rb
```diff
@@ -79,6 +59,32 @@
     end
   end
 
+  context "when searching schemes" do
+    let(:scheme) { FactoryBot.create(:scheme, owning_organisation_id: case_log.created_by.organisation_id, primary_client_group: "Q", secondary_client_group: "P") }
+
+    before do
+      FactoryBot.create(:location, scheme:, postcode: "W6 0ST")
+      FactoryBot.create(:location, scheme:, postcode: "SE6 1LB")
+      case_log.update!(needstype: 2)
```
baarkerlounger, line 68: Any reason not to just have this in the FactoryBot create?

baarkerlounger (approved):
Lovely!

spec/features/form/accessible_autocomplete_spec.rb
```diff
@@ -58,8 +59,34 @@
...
+
+    it "displays hint text under the options", js: true do
+      find("#case-log-scheme-id-field").click.native.send_keys("w", "6", :down, :enter)
+      expect(find(".autocomplete__option__hint", visible: :hidden, text: /Young people at risk, Young people leaving care/)).to be_present
+    end
+  end
+
   it "has the correct option selected if one has been saved" do
-    case_log.update!(postcode_known: 0, previous_la_known: 1, prevloc: "E07000178")
+    case_log.update!(postcode_known: 0, previous_la_known: 1, prevloc: "Oxford")
```
kosiakkatrina: This is wrong. It's saving the la/prevloc as a string instead to an E code now. Needs fixing. Edit: actually works fine

spec/features/form/accessible_autocomplete_spec.rb
kosiakkatrina: In case it affects other tests 🤷

PR #741, opened 2022-07-13
DLUHC-1228: Scheme import improvements

smeny:
* Set owning organisation as agent when directly managed
* Map the mobility type

### DIFF
--- a/app/services/imports/scheme_import_service.rb
+++ b/app/services/imports/scheme_import_service.rb
@@ -5,20 +5,18 @@ module Imports
     end
 
     def create_scheme(xml_document)
-      old_id = string_or_nil(xml_document, "id")
-      status = string_or_nil(xml_document, "status")
-
-      if status == "Approved"
+      attributes = scheme_attributes(xml_document)
+      if attributes["status"] == "Approved"
         Scheme.create!(
-          owning_organisation_id: find_owning_organisation_id(xml_document),
-          managing_organisation_id: find_managing_organisation_id(xml_document),
-          service_name: string_or_nil(xml_document, "name"),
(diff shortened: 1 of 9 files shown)


app/services/imports/scheme_import_service.rb
```diff
@@ -39,16 +37,33 @@ def safe_string_as_integer(xml_doc, attribute)
...
+      attributes["old_visible_id"] = string_or_nil(xml_doc, "visible-id")
+      attributes["status"] = string_or_nil(xml_doc, "status")
+      attributes["service_name"] = string_or_nil(xml_doc, "name")
+      attributes["arrangement_type"] = string_or_nil(xml_doc, "arrangement_type")
+      attributes["owning_org_old_id"] = string_or_nil(xml_doc, "institution")
+      attributes["owning_organisation_id"] = find_owning_organisation_id(attributes["owning_org_old_id"])
+      attributes["management_org_old_visible_id"] = safe_string_as_integer(xml_doc, "agent")
+      attributes["managing_organisation_id"] = find_managing_organisation_id(attributes["management_org_old_visible_id"])
+
+      if attributes["arrangement_type"] == "D" && attributes["managing_organisation_id"].nil?
```
baarkerlounger, line 52: What is "D"? Could this be a named constant?

baarkerlounger (approved):

app/services/imports/scheme_import_service.rb
smeny: Directly managed.
We got a test that makes this explicit, so I think this is OK?

baarkerlounger: Not a blocker for the PR by any means. Just a nitpick :slightly_smiling_face:

PR #744, opened 2022-07-14
Cldc 1366 add location startdate

kosiakkatrina:
Add a location start date question to the form:
image

In a log only show schemes where at least 1 location exists and is active
In a log only show locations that are active (i.e has no start date or has a start date today or in the past)

### DIFF
--- a/app/models/form/setup/questions/scheme_id.rb
+++ b/app/models/form/setup/questions/scheme_id.rb
@@ -22,7 +22,7 @@ class Form::Setup::Questions::SchemeId < ::Form::Question
   def displayed_answer_options(case_log)
     return {} unless case_log.created_by
 
-    user_org_scheme_ids = Scheme.select(:id).where(owning_organisation_id: case_log.created_by.organisation_id).joins(:locations).merge(Location.where("startdate <= ? or startdate IS NULL", Date.today)).map(&:id)
+    user_org_scheme_ids = Scheme.select(:id).where(owning_organisation_id: case_log.created_by.organisation_id).joins(:locations).merge(Location.where("startdate <= ? or startdate IS NULL", Time.zone.today)).map(&:id)
     answer_options.select do |k, _v|
       user_org_scheme_ids.include?(k.to_i)
     end
--- a/app/models/form/setup/questions/scheme_id.rb
+++ b/app/models/form/setup/questions/scheme_id.rb
@@ -22,7 +22,7 @@ class Form::Setup::Questions::SchemeId < ::Form::Question
   def displayed_answer_options(case_log)
     return {} unless case_log.created_by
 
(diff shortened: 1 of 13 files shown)


app/views/locations/new.html.erb
```diff
@@ -45,6 +45,10 @@
                                            :name,
                                            hint: { text: "This includes stairlifts, ramps, level-access showers or grab rails" },
                                            legend: { text: "Are the majority of units in this location built or adapted to wheelchair-user standards?", size: "m" } %>
+      
+      <%= f.govuk_date_field :startdate,
+                              legend: { text: "When did the first property in this location become available?", size: "m" },
```
baarkerlounger: @Dan13L3 should this text be:

`When did the first property in this location become available under this scheme?`

I'm wondering if there might be cases where the property has been available for social let but not with the supported services previously that might be ambiguous/confusing without additional clarification?

app/views/locations/edit.html.erb
```diff
@@ -44,6 +44,10 @@
                                            hint: { text: "This includes stairlifts, ramps, level-access showers or grab rails" },
                                            legend: { text: "Are the majority of units in this location built or adapted to wheelchair-user standards?", size: "m" } %>
 
+      <%= f.govuk_date_field :startdate,
+                              legend: { text: "When did the first property in this location become available?", size: "m" },
```
baarkerlounger: @kosiakkatrina worth using the translation file for this since we have the string in multiple places?

app/models/form/setup/questions/scheme_id.rb
```diff
@@ -22,7 +22,7 @@ def answer_options
   def displayed_answer_options(case_log)
     return {} unless case_log.created_by
 
-    user_org_scheme_ids = Scheme.select(:id).where(owning_organisation_id: case_log.created_by.organisation_id).map(&:id)
+    user_org_scheme_ids = Scheme.select(:id).where(owning_organisation_id: case_log.created_by.organisation_id).joins(:locations).merge(Location.where("startdate <= ? or startdate IS NULL", Time.zone.now)).map(&:id)
```
baarkerlounger: Should we be comparing to a date here rather than a time?

kosiakkatrina: Time.zone.today is the closest I could get to with the rubicon

kosiakkatrina: Why is it that my Mac thinks of rubocop as a soft drink or an existential concept?

baarkerlounger (approved):
:+1: when @Dan13L3 is happy with the content

Dan13L3:
@baarkerlounger @kosiakkatrina Happy with the content change - definitely makes sense to clarify. Should be good to go now

smeny (changes requested):
Putting a blocker here.
Before we move on, we need to clarify if both start and end date apply at the scheme or location level.

baarkerlounger:
@smeny we've confirmed it's location level. Schemes should be available if they have any location with a valid start date (or none), and for each scheme only their available locations should be available to select.

For import all schemes and location will be considered active/valid.

PR #745, opened 2022-07-15
CLDC-1228: Handles household charges

smeny:


baarkerlounger:
As discussed:

- County field will be removed
- Location#location_code field will store the LA
- Creating a location will do the postcode API call to look up LA

Ideally we want to have all fields defined on scheme or location to be pointed to by the case log rather than duplicated by the case log.

app/services/imports/case_logs_import_service.rb
```diff
@@ -590,5 +601,49 @@ def apply_household_consistency!(attributes)
...
+          attributes["pscharge"].nil? &&
+          attributes["supcharg"].nil?
+        attributes["brent"] ||= BigDecimal("0.0")
+        attributes["scharge"] ||= BigDecimal("0.0")
+        attributes["pscharge"] ||= BigDecimal("0.0")
+        attributes["supcharg"] ||= BigDecimal("0.0")
+      end
+    end
+
+    def clear_household_charges(attributes)
```
smeny: Delete this method

spec/services/imports/case_logs_import_service_spec.rb
```diff
@@ -203,5 +215,17 @@ def open_file(directory, filename)
         end
       end
     end
+
+    context "and this is a supported housing log" do
```
smeny, line 219: expect no warning

app/services/postcode_service.rb
```diff
@@ -1,4 +1,25 @@
 class PostcodeService
+  def initialize
+    @pio = Postcodes::IO.new
+  end
+
+  def infer_la(postcode)
```
baarkerlounger, line 6: Love this! (Refactoring it out into the service)

app/models/case_log.rb
```diff
@@ -102,6 +102,22 @@ def self.editable_fields
     attribute_names - AUTOGENERATED_FIELDS
   end
 
+  def la
```
baarkerlounger, line 105: Nice!

baarkerlounger (approved):

PR #747, opened 2022-07-18
add unittype_sh column to csv export

kosiakkatrina:


app/models/case_log.rb
```diff
@@ -395,10 +395,10 @@ def is_prevten_la_general_needs?
 
   def self.to_csv
     CSV.generate(headers: true) do |csv|
-      csv << attribute_names
+      csv << attribute_names + %w[unittype_sh]
 
       all.find_each do |record|
-        csv << record.attributes.map do |att, val|
+        csv << record.attributes.merge({ "unittype_sh" => record.unittype_sh }).map do |att, val|
```
baarkerlounger, line 417: Should maybe be a test for this?

baarkerlounger: Given "We’ll also want all the other fields from the scheme and from the location" it might be easier to just loop over scheme and location fields entirely as well.

kosiakkatrina: Will look at adding a test, probably worth looping, but perhaps when we know exactly how the export is gonna look like?

spec/models/case_log_spec.rb
```diff
@@ -2235,4 +2235,24 @@ def previous_postcode_fields
...
     end
   end
+
+  describe "csv download" do
+    let(:csv_export_file) { File.open("spec/fixtures/files/case_logs_download.csv", "r:UTF-8") }
+    let(:scheme) { FactoryBot.create(:scheme) }
+    let(:location) { FactoryBot.create(:location, scheme:, type_of_unit: 6, postcode: "SE11TE") }
+
+    before do
+      described_class.destroy_all
```
baarkerlounger: Why is this necessary? Feels like a hack for a problem elsewhere?

kosiakkatrina: Hmm, just to make sure that we only have 1 record when we make the download. Doesn't seem to fail without it, so happy to remove

baarkerlounger (approved):

spec/models/case_log_spec.rb
baarkerlounger: If it's needed I'd guess it's order dependent, some other test is leaking data - should probably fix it at site if that's the case though.

PR #749, opened 2022-07-19
CLDC-1228: Handles soft validation confirmations

smeny:


baarkerlounger (approved):

app/services/imports/case_logs_import_service.rb
```diff
@@ -207,6 +207,13 @@ def create_log(xml_doc)
         attributes["postcode_full"] = nil
       end
 
+      # Sets all soft validations to yes by default
```
baarkerlounger: Might be worth having the comment say a bit more about *why* we're doing that rather than just that we are but not super important

PR #755, opened 2022-07-19
Extract and fix answer selected method

kosiakkatrina:
Extract answer_selected method that determines if the given answer for the question is selected. 
Handle schema_id question differently than other select questions because of it's structure.
Add an empty ("Select an option") option to the schema_id question.

### DIFF
--- a/app/models/form/setup/questions/scheme_id.rb
+++ b/app/models/form/setup/questions/scheme_id.rb
@@ -10,7 +10,7 @@ class Form::Setup::Questions::SchemeId < ::Form::Question
   end
 
   def answer_options
-    answer_opts = {}
+    answer_opts = { "" => "Select an option" }
     return answer_opts unless ActiveRecord::Base.connected?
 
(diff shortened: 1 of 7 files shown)


app/models/form/setup/questions/scheme_id.rb
```diff
@@ -24,14 +24,18 @@ def displayed_answer_options(case_log)
...
-      user_org_scheme_ids.include?(k.to_i)
+      user_org_scheme_ids.include?(k.to_i) || k.blank?
     end
   end
 
   def hidden_in_check_answers?(case_log, _current_user = nil)
     !supported_housing_selected?(case_log)
   end
 
+  def answer_selected?(case_log, answer)
```
baarkerlounger, line 35: Should there be a guard clause here that makes clear this method only applies to select questions?

baarkerlounger (approved):

app/models/form/setup/questions/scheme_id.rb
kosiakkatrina: Probably

baarkerlounger (approved):

PR #757, opened 2022-07-20
Add fail fast flag to pipelines

dushan-madetech:
Adds the `--fail-fast` flag to the pipelines to free up the runners if a test fails. This will be particularly useful if the commit is already known to have failing tests or for stopping right away on any flaky tests


baarkerlounger (approved):
Got to keep that server load down!

PR #760, opened 2022-07-20
CLDC-1378 scheme start date validation

moarpheus:
Screenshot 2022-07-20 at 10 32 24

If the start date is supplied it goes through validation.

### DIFF
--- a/app/controllers/locations_controller.rb
+++ b/app/controllers/locations_controller.rb
@@ -16,12 +16,11 @@ class LocationsController < ApplicationController
   end
 
   def create
-    if valid_date_params?(location_params)
+    if date_params_missing?(location_params) || valid_date_params?(location_params)
       @location = Location.new(location_params)
       if @location.save
         location_params[:add_another_location] == "Yes" ? redirect_to(new_location_path(id: @scheme.id)) : redirect_to(scheme_check_answers_path(scheme_id: @scheme.id))
       else
-        @location.errors.add(:startdate)
         render :new, status: :unprocessable_entity
       end
     else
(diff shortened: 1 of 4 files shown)


app/controllers/locations_controller.rb
```diff
@@ -46,6 +52,19 @@ def update
 
 private
 
+  def valid_date_params?(location_params)
+    is_integer?(location_params["startdate(1i)"]) && is_integer?(location_params["startdate(2i)"]) && is_integer?(location_params["startdate(3i)"]) &&
+      Date.valid_date?(location_params["startdate(1i)"].to_i, location_params["startdate(2i)"].to_i, location_params["startdate(3i)"].to_i)
+  end
+
+  def date_params_missing?(location_params)
+    location_params["startdate(1i)"].blank? && location_params["startdate(2i)"].blank? && location_params["startdate(3i)"].blank?
```
baarkerlounger: Isn't it enough for any one of these to be blank? (i.e. || instead of &&)

app/controllers/locations_controller.rb
```diff
@@ -46,6 +52,19 @@ def update
...
+  end
+
+  def is_integer?(string)
+    string.to_i.to_s == string
```
baarkerlounger, line 65: Seems a bit odd to check if something is an integer by converting it to string? I'd expect something like:

```
begin
  Integer(string)
rescue ArgumentError
  false
end
```

baarkerlounger: Oh I see, we're checking it doesn't contain anything else besides integers...

app/controllers/locations_controller.rb
moarpheus: Good spot! One is enough yup

baarkerlounger (approved):

PR #761, opened 2022-07-20
Fix data mapping for mantype

dushan-madetech:
change to use the `arrangement_type` column

### DIFF
--- a/app/controllers/schemes_controller.rb
+++ b/app/controllers/schemes_controller.rb
@@ -30,7 +30,7 @@ class SchemesController < ApplicationController
     validation_errors scheme_params
 
     if @scheme.errors.empty? && @scheme.save
-      if scheme_params[:arrangement_type] == "The same organisation that owns the housing stock"
+      if @scheme.arrangement_type_before_type_cast == "D"
         redirect_to scheme_primary_client_group_path(@scheme)
       else
         redirect_to scheme_support_services_provider_path(@scheme)
@@ -100,9 +100,7 @@ private
 
   def validation_errors(scheme_params)
     scheme_params.each_key do |key|
-      if key == "support_services_provider"
-        @scheme.errors.add("support_services_provider_before_type_cast".to_sym) if scheme_params[key].to_s.empty?
-      elsif scheme_params[key].to_s.empty?
+      if scheme_params[key].to_s.empty?
         @scheme.errors.add(key.to_sym)
       end
     end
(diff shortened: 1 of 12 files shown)


app/controllers/schemes_controller.rb
```diff
@@ -143,9 +143,9 @@ def next_page_path(page)
     when "support"
       new_location_path
     when "details"
-      if @scheme.support_services_provider_before_type_cast&.zero?
+      if @scheme.arrangement_type == "The same organisation that owns the housing stock"
         scheme_primary_client_group_path(@scheme)
-      elsif @scheme.support_services_provider_before_type_cast.positive?
```
smeny, line 148: For conditionals I wouldn't rely on the text value (it can change based on design), but used the letter instead.

baarkerlounger:
Think we should be comparing to key not string everywhere but otherwise :+1: .

app/views/schemes/new.html.erb
```diff
@@ -65,15 +65,13 @@
...
-                                           support_services_provider_selection,
+      <%= f.govuk_collection_radio_buttons :arrangement_type,
+                                           mantype_selection,
                                            :id,
                                            lambda { |option|
-                                             if option.id.zero? && !current_user.support?
+                                             if option.id == "The same organisation that owns the housing stock" && !current_user.support?
                                                "Your organisation"
                                              else
                                                option.name
```
dushan-madetech: Changing this is more involved than first thought - potentially worth reworking in a future PR?

app/views/schemes/details.html.erb
```diff
@@ -63,14 +63,13 @@
...
+
+      <%= f.govuk_collection_radio_buttons :arrangement_type,
+                                           mantype_selection,
                                            :id,
                                            lambda { |option|
-                                             if option.id.zero? && !current_user.support?
+                                             if option.id == "The same organisation that owns the housing stock" && !current_user.support?
                                                "Your organisation"
                                              else
                                                option.name
```
dushan-madetech: Changing this is more involved than first thought - potentially worth reworking in a future PR?

app/views/schemes/new.html.erb
baarkerlounger: What about leaving the keys and value strings here but making the checks something like:

```ruby
if Scheme::ARRANGEMENTTYPE[option.id].zero?
```

dushan-madetech: The problem I was having with keys and values was that the option selected wouldn't remain selected if you hit save and continue to move on and then hit the back button to return to the page. I'm not sure if arrangement type can be compared to zero since the enum maps to strings 🤔

baarkerlounger: Right well `== "0"`  or `to_i` then. The example was just the idea that we look up the thing we're comparing to from a single source of truth rather than duplicating the comparison string across the code base.

dushan-madetech: I think I've got your suggestion implemented @baarkerlounger 🙂

PR #762, opened 2022-07-20
Add redis to infra diagram

baarkerlounger:


paulrobertlloyd:
Shouldn’t Redis also be a square grey box with a database icon inside it?

baarkerlounger:
@paulrobertlloyd not sure - that was the redis icon in Diagram.io. Happy for you to redraw though if you have something else in mind?

paulrobertlloyd (approved):
Updated the image to use Redis icon more like the other items.

PR #763, opened 2022-07-20
Add mobility type question

kosiakkatrina:
- [x] Add mobility type question to the locations form
image

- [x] remove wheelchair_adaptation
- [x] infer WCHAIR using mobility type
- [x] Add validation
- [x] Display mobility type in locations table
image


app/models/location.rb
```diff
@@ -15,10 +15,10 @@ class Location < ApplicationRecord
...
 
   MOBILITY_TYPE = {
-    "Property fitted with equipment and adaptations (if not designed to above standards)": "A",
+    "Wheelchair-user standard": "W",
+    "Fitted with equipment and adaptations": "A",
     "Property designed to accessible general standard": "M",
     "None": "N",
-    "Property designed to wheelchair user standard": "W",
     "Missing": "X",
   }.freeze
```
smeny, line 16: Looks correct to me :)

app/views/locations/edit.html.erb
```diff
@@ -36,13 +36,16 @@
...
-      <% wheelchair_user_selection = Location.wheelchair_adaptations.keys.map { |key, _| OpenStruct.new(id: key, name: key.to_s.humanize) } %>
-      <%= f.govuk_collection_radio_buttons :wheelchair_adaptation,
-                                           wheelchair_user_selection,
-                                           :id,
-                                           :name,
-                                           hint: { text: I18n.t("hints.location.wheelchair_adaptation") },
-                                           legend: { text: I18n.t("questions.location.wheelchair_adaptation"), size: "m" } %>
+      <% mobility_types_to_display = Location.mobility_types.excluding("Property designed to accessible general standard", "Missing") %>
+
+      <% mobility_type_selection = mobility_types_to_display.map { |key, value| OpenStruct.new(id: key, name: key.to_s.humanize, description: I18n.t("questions.descriptions.location.mobility_type.#{value}")) } %>
```
baarkerlounger: Think this should be `id: value`? That we in the controller we're dealing with "M" rather than the string etc

app/views/locations/new.html.erb
```diff
@@ -37,14 +37,16 @@
...
-      <% wheelchair_user_selection = Location.wheelchair_adaptations.keys.map { |key, _| OpenStruct.new(id: key, name: key.to_s.humanize) } %>
+      <% mobility_types_to_display = Location.mobility_types.excluding("Property designed to accessible general standard", "Missing") %>
 
-      <%= f.govuk_collection_radio_buttons :wheelchair_adaptation,
-                                           wheelchair_user_selection,
-                                           :id,
-                                           :name,
-                                           hint: { text: I18n.t("hints.location.wheelchair_adaptation") },
-                                           legend: { text: I18n.t("questions.location.wheelchair_adaptation"), size: "m" } %>
+      <% mobility_type_selection = mobility_types_to_display.map { |key, value| OpenStruct.new(id: key, name: key.to_s.humanize, description: I18n.t("questions.descriptions.location.mobility_type.#{value}")) } %>
```
baarkerlounger: Same here. Could maybe also dry this up?

app/views/locations/edit.html.erb
kosiakkatrina: Hmm, we're using key in all the other questions. The id looks nicer for it as well `location-mobility-type-none-field` vs `location-mobility-type-n-field`. We're also not doing anything with it in the controller and we use enums to map before saving so I don't see it too valuable to change 🤔 Can be convinced otherwise

baarkerlounger: So I think it's more semantically correct in general. If are values have an id and a label (as we do in the enum), we should use those for the html we produce?

Might not matter straight away since rails will convert for us, but it becomes more important if at any point in the future you want to include it in conditional logic of some kind (since you'd rather have `params["mobility_type"] == "W"` which shouldn't ever need changing than `params["mobility_type"] == "Wheelchair-user standard"` which more feasibly might?

smeny: Dushan has a similar problem with enums. I don't think we can load and save the value without using the "text" reference (or at least I haven't seen how). If you use the letter as the ID, then the option selected won't load.

baarkerlounger: You can do it by using the `before_typecast` field as the field you pass into the form builder. i.e. something like:

```ruby
<%= f.govuk_collection_radio_buttons :mobility_type_before_typecast,
                                            mobility_type_selection,
                                            :id,
                                            :name,
                                            :description,
                                            legend: { text: I18n.t("questions.location.mobility_type"), size: "m" } %>
```

baarkerlounger: Will admit it's less than ideal though

baarkerlounger: (Happy to merge this as is if the consensus it's the suggestion is worse than current)

moarpheus: _type_before_typecast are awkward to handle especially if you have some validations, I did this for scheme managing org - not very pretty.

kosiakkatrina: Calm, will merge cause it looks like a larger refactor 🦐

baarkerlounger: @moarpheus agreed. The amount of wrangling it actually takes I think is probably an indication we're going against rails convention here and probably just shouldn't use enums if this is what we want. Happy to merge as is.

PR #765, opened 2022-07-21
CLDC-1315: Model deletion cascade and foreign key constraints

Ted-U:


db/migrate/20220720214646_add_delete_cascade_from_orgs_down.rb
```diff
@@ -0,0 +1,11 @@
+class AddDeleteCascadeFromOrgsDown < ActiveRecord::Migration[7.0]
+  def up
+    remove_foreign_key :schemes, :organisations, column: "owning_organisation_id"
+    remove_foreign_key :schemes, :organisations, column: "managing_organisation_id"
```
baarkerlounger, line 4: Why are we removing this foreign key?

db/migrate/20220720214646_add_delete_cascade_from_orgs_down.rb
```diff
@@ -0,0 +1,11 @@
...
+    add_foreign_key :schemes, :organisations, column: "owning_organisation_id", on_delete: :cascade
+  end
+
+  def down
+    remove_foreign_key :schemes, :organisations, column: "owning_organisation_id"
```
baarkerlounger: This down isn't quite the opposite of up - it won't get you back to your original state because originally you did have a foreign key, just without the delete cascade option

db/schema.rb
```diff
@@ -379,8 +379,9 @@
...
 
   add_foreign_key "case_logs", "locations"
-  add_foreign_key "case_logs", "schemes"
+  add_foreign_key "case_logs", "organisations", column: "owning_organisation_id", on_delete: :cascade
+  add_foreign_key "case_logs", "schemes", on_delete: :cascade
   add_foreign_key "locations", "schemes"
-  add_foreign_key "schemes", "organisations", column: "managing_organisation_id"
-  add_foreign_key "schemes", "organisations", column: "owning_organisation_id"
+  add_foreign_key "schemes", "organisations", column: "owning_organisation_id", on_delete: :cascade
+  add_foreign_key "users", "organisations", on_delete: :cascade
```
baarkerlounger, line 387: Where is this coming from? Looks like there's a migration not committed here?

db/migrate/20220720214646_add_delete_cascade_from_orgs_down.rb
Ted-U: My migration file was unbalanced, I have added back the foreign key for managing organisations.

db/migrate/20220720214646_add_delete_cascade_from_orgs_down.rb
Ted-U: Should be fixed now

db/migrate/20220720214646_add_delete_cascade_from_orgs_down.rb
```diff
@@ -0,0 +1,11 @@
+class AddDeleteCascadeFromOrgsDown < ActiveRecord::Migration[7.0]
+  def up
+    add_foreign_key :schemes, :organisations, column: "owning_organisation_id", on_delete: :cascade
```
baarkerlounger, line 3: You're adding the same key twice here, and removing it twice below which isn't right.

Best way is to run the migration up and down locally and check it gives the expected schema in both cases.

db/migrate/20220720214646_add_delete_cascade_from_orgs_down.rb
```diff
@@ -0,0 +1,9 @@
...
```
baarkerlounger, line 3: Does this work if a foreign key already exists or do you need to remove the existing one first?

(i.e. if you do `rake db:drop; rake db:create; rake db:migrate` is it successful?)

db/schema.rb
Ted-U: This is fixed

db/migrate/20220720214646_add_delete_cascade_from_orgs_down.rb
Ted-U: It does now!

db/migrate/20220720214646_add_delete_cascade_from_orgs_down.rb
```diff
@@ -0,0 +1,7 @@
+class AddDeleteCascadeFromOrgsDown < ActiveRecord::Migration[7.0]
+  def change
+    remove_foreign_key :schemes, :organisations, column: "owning_organisation_id"
+    remove_foreign_key :schemes, :organisations, column: "managing_organisation_id"
```
baarkerlounger: I don't think we want to remove the managing org foreign key

db/migrate/20220722133738_add_delete_caselogs_users.rb
```diff
@@ -0,0 +1,6 @@
+class AddDeleteCaselogsUsers < ActiveRecord::Migration[7.0]
+  def change
+    add_foreign_key :case_logs, :organisations, column: "owning_organisation_id", on_delete: :cascade
```
baarkerlounger, line 3: Do we not need to remove these first? Did they not exist in the first place?

Ted-U: This relationship doesn't appear in the schema in main so I don't think it existed previously.

db/migrate/20220720214646_add_delete_cascade_from_orgs_down.rb
Ted-U: Fixed in latest commit

baarkerlounger:
Code looks good. I think what's missing here now is checking whether this migration will fail on Staging (I suspect it will since we likely have records there that won't satisfy the foreign key constraints) and preparing a script or something for how to deal with those records.

baarkerlounger:
Looks like data is all fine on staging and prod now. Let's merge this after Monday release.

kosiakkatrina (approved):

PR #767, opened 2022-07-21
shuffle schemes feature tests

kosiakkatrina:


spec/features/form/helpers.rb
```diff
@@ -22,4 +22,63 @@ def sign_in(user)
     fill_in("user[password]", with: user.password)
     click_button("Sign in")
   end
+
+  def fill_in_and_save_scheme_details
```
baarkerlounger: Think it would be worth making this a separate helper since it doesn't really have anything to do with the "form" (form being the case log form, rather than _any_ form)

kosiakkatrina: Yeah, I was thinking about it, but there were signing in methods so I just chucked it in here. I don't mind moving it

moarpheus (approved):
Looks VERY good to me!

baarkerlounger (approved):

PR #769, opened 2022-07-22
Cldc 1388 Show local authority for location

moarpheus:
Screenshot 2022-07-22 at 09 40 46

### DIFF
--- a/app/services/postcode_service.rb
+++ b/app/services/postcode_service.rb
@@ -3,7 +3,7 @@ class PostcodeService
     @pio = Postcodes::IO.new
   end
 
-  def infer_la(postcode)
+  def lookup(postcode)
     # Avoid network calls when postcode is invalid
     return unless postcode.match(POSTCODE_REGEXP)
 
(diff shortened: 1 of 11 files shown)


spec/features/schemes_spec.rb
```diff
@@ -383,10 +384,12 @@
           end
 
           it "displays information about the first created location" do
-            expect(page).to have_content "SW1P4DF"
+            expect(page).to have_content "AA11AA"
             expect(page).to have_content "Some name"
             expect(page).to have_content "Self-contained house"
             expect(page).to have_content "None"
+            expect(page).to have_content "Local authority"
+            expect(page).to have_content "E09000033"
```
baarkerlounger: Think we should be showing the label rather than the code in UI. i.e Westminster rather than E09000033

app/services/postcode_service.rb
```diff
@@ -20,6 +20,23 @@ def infer_la(postcode)
     end
   end
 
+  def infer_admin_district(postcode)
```
baarkerlounger: Isn't this the same as the method above?

moarpheus: Returns different value - I will refactor those two into one, to either return code and value or return postcode object

baarkerlounger (approved):

PR #774, opened 2022-07-22
Fix scheme question when case log doesn't have an organisation yet

baarkerlounger:
Fixes https://sentry.io/organizations/dluhc-core/issues/3438180511

Triggered when you try to view the schemes question as a support user without having answered the organisation question or the created by question yet.

### DIFF
--- a/app/models/case_log.rb
+++ b/app/models/case_log.rb
@@ -539,10 +539,17 @@ private
     self.created_by = nil if created_by.organisation != owning_organisation
   end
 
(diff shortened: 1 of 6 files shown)


dushan-madetech (approved):

moarpheus (dismissed):

smeny (approved):

kosiakkatrina (commented):
It allows you to select an organisation and scheme that don't match 😂

PR #775, opened 2022-07-22
Cldc 1377 scheme status

kosiakkatrina:
Reset managing org for the sheme when a different organisation is selected as providing support
Only display schemes that have managing org id selected in the case log form
Only display relevant answers on the show scheme page
Validate all fields completed before marking scheme as confirmed
image

### DIFF
--- a/app/controllers/schemes_controller.rb
+++ b/app/controllers/schemes_controller.rb
@@ -104,7 +104,7 @@ private
       end
     end
 
-    if @scheme.arrangement_type_same? && scheme_params[:arrangement_type] != "D"
+    if @scheme.arrangement_type_same? && scheme_params[:arrangement_type] != "The same organisation that owns the housing stock"
       @scheme.errors.delete(:managing_organisation_id)
     end
   end
@@ -191,11 +191,11 @@ private
   end
 
   def arrangement_type_set_to_same_org?(required_params)
-    required_params[:arrangement_type] == "D" || (required_params[:arrangement_type].blank? && @scheme.present? && @scheme.arrangement_type_same?)
+    required_params[:arrangement_type] == "The same organisation that owns the housing stock" || (required_params[:arrangement_type].blank? && @scheme.present? && @scheme.arrangement_type_same?)
   end
 
   def arrangement_type_changed_to_different_org?(required_params)
-    @scheme.present? && @scheme.arrangement_type_same? && required_params[:arrangement_type] != "D" && required_params[:managing_organisation_id].blank?
+    @scheme.present? && @scheme.arrangement_type_same? && required_params[:arrangement_type] != "The same organisation that owns the housing stock" && required_params[:managing_organisation_id].blank?
   end
 
   def search_term
--- a/app/controllers/schemes_controller.rb
+++ b/app/controllers/schemes_controller.rb
@@ -104,7 +104,7 @@ private
       end
     end
 
-    if @scheme.arrangement_type_same? && scheme_params[:arrangement_type] != "The same organisation that owns the housing stock"
+    if @scheme.arrangement_type_same? && arrangement_type_value(scheme_params[:arrangement_type]) != "D"
       @scheme.errors.delete(:managing_organisation_id)
     end
   end
@@ -191,11 +191,15 @@ private
   end
 
   def arrangement_type_set_to_same_org?(required_params)
-    required_params[:arrangement_type] == "The same organisation that owns the housing stock" || (required_params[:arrangement_type].blank? && @scheme.present? && @scheme.arrangement_type_same?)
+    arrangement_type_value(required_params[:arrangement_type]) == "D" || (required_params[:arrangement_type].blank? && @scheme.present? && @scheme.arrangement_type_same?)
   end
 
   def arrangement_type_changed_to_different_org?(required_params)
-    @scheme.present? && @scheme.arrangement_type_same? && required_params[:arrangement_type] != "The same organisation that owns the housing stock" && required_params[:managing_organisation_id].blank?
+    @scheme.present? && @scheme.arrangement_type_same? && arrangement_type_value(required_params[:arrangement_type]) != "D" && required_params[:managing_organisation_id].blank?
+  end
+
+  def arrangement_type_value(key)
+    key.present? ? Scheme::ARRANGEMENT_TYPE[key.to_sym] : nil
   end
 
   def search_term
--- a/app/controllers/schemes_controller.rb
+++ b/app/controllers/schemes_controller.rb
@@ -103,6 +103,10 @@ private
         @scheme.errors.add(key.to_sym)
       end
     end
+
+    if @scheme.arrangement_type_same? && scheme_params[:arrangement_type] != "The same organisation that owns the housing stock"
+      @scheme.errors.delete(:managing_organisation_id)
+    end
   end
 
   def confirm_secondary_page?(page)
@@ -170,14 +174,20 @@ private
                                                      :intended_stay,
                                                      :confirmed)
 
-    full_params = required_params[:arrangement_type] == "D" && required_params[:owning_organisation_id].present? ? required_params.merge(managing_organisation_id: required_params[:owning_organisation_id]) : required_params
+    if @scheme.present? && @scheme.arrangement_type_same? && required_params[:arrangement_type] != "The same organisation that owns the housing stock" && required_params[:managing_organisation_id].blank?
+      required_params[:managing_organisation_id] = nil
+    end
 
-    full_params[:sensitive] = full_params[:sensitive].to_i if full_params[:sensitive]
+    if required_params[:arrangement_type] == "The same organisation that owns the housing stock"
+      required_params[:managing_organisation_id] = required_params[:owning_organisation_id] || @scheme.owning_organisation_id
+    end
+
+    required_params[:sensitive] = required_params[:sensitive].to_i if required_params[:sensitive]
 
     if current_user.data_coordinator?
-      full_params[:owning_organisation_id] = current_user.organisation_id
+      required_params[:owning_organisation_id] = current_user.organisation_id
     end
-    full_params
+    required_params
   end
 
   def search_term
--- a/app/controllers/schemes_controller.rb
+++ b/app/controllers/schemes_controller.rb
@@ -178,7 +178,7 @@ private
       required_params[:managing_organisation_id] = nil
     end
 
-    if required_params[:arrangement_type] == "The same organisation that owns the housing stock"
+    if required_params[:arrangement_type] == "The same organisation that owns the housing stock" || (required_params[:arrangement_type].blank? && @scheme.arrangement_type_same?)
       required_params[:managing_organisation_id] = required_params[:owning_organisation_id] || @scheme.owning_organisation_id
     end
 
--- a/app/controllers/schemes_controller.rb
+++ b/app/controllers/schemes_controller.rb
@@ -128,6 +128,8 @@ private
       "schemes/details"
     elsif page.include?("edit")
       "schemes/edit_name"
+    elsif page.include?("check-answers")
+      "schemes/check_answers"
     end
   end
 
--- a/app/controllers/schemes_controller.rb
+++ b/app/controllers/schemes_controller.rb
@@ -104,7 +104,7 @@ private
       end
     end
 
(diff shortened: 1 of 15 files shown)


app/controllers/schemes_controller.rb
```diff
@@ -103,6 +103,10 @@ def validation_errors(scheme_params)
         @scheme.errors.add(key.to_sym)
       end
     end
+
+    if @scheme.arrangement_type_same? && scheme_params[:arrangement_type] != "The same organisation that owns the housing stock"
```
dushan-madetech: Note to self: 

If possible change to not refer to the text value (as this is likely to change)

baarkerlounger: This seems like a good idea before merging still?

app/controllers/schemes_controller.rb
```diff
@@ -170,14 +174,28 @@ def scheme_params
                                                      :intended_stay,
                                                      :confirmed)
 
-    full_params = required_params[:arrangement_type] == "D" && required_params[:owning_organisation_id].present? ? required_params.merge(managing_organisation_id: required_params[:owning_organisation_id]) : required_params
+    if arrangement_type_changed_to_different_org?(required_params)
+      required_params[:managing_organisation_id] = nil
```
baarkerlounger, line 178: Is there any scenario where this could be set but to a different id than the owning org that needs checking for?

app/controllers/schemes_controller.rb
```diff
@@ -170,14 +174,28 @@ def scheme_params
...
     if current_user.data_coordinator?
-      full_params[:owning_organisation_id] = current_user.organisation_id
+      required_params[:owning_organisation_id] = current_user.organisation_id
     end
-    full_params
+    required_params
+  end
+
+  def arrangement_type_set_to_same_org?(required_params)
+    required_params[:arrangement_type] == "The same organisation that owns the housing stock" || (required_params[:arrangement_type].blank? && @scheme.present? && @scheme.arrangement_type_same?)
```
baarkerlounger: Be good to remove the string comparison here too

app/controllers/schemes_controller.rb
kosiakkatrina: So, the params come thru as strings and not the values we store. I think it might be a part of a bigger refactor, cause all of our values come thru like that atm

smeny: For context, we discussed this with Dushan over the past few days.
While the label for an option can change reasonably often (design/year changes), the letter code/ID won't. So as much as possible, we want to rely on the code.

app/controllers/schemes_controller.rb
kosiakkatrina: Hm, not sure I understand, can you rephrase? or give an example?

app/models/form/setup/questions/scheme_id.rb
```diff
@@ -21,7 +21,7 @@ def answer_options
 
   def displayed_answer_options(case_log)
     organisation = case_log.owning_organisation || case_log.created_by&.organisation
-    schemes = organisation ? Scheme.select(:id).where(owning_organisation_id: organisation.id) : Scheme.select(:id)
+    schemes = organisation ? Scheme.select(:id).where(owning_organisation_id: organisation.id).where.not(managing_organisation_id: nil) : Scheme.select(:id).where.not(managing_organisation_id: nil)
```
baarkerlounger: We have the `confirmed` field on `schemes` now. The idea was that that field is only set on the final "Create scheme" press, at which point all fields should have been validated. So we could check for that here rather than that each individual field has been set?

Doesn't make a huge difference now since it's currently just `managing_org_id` but when editing functionality is introduced and any field could potentially be missing it should make state management easier?

app/controllers/schemes_controller.rb
kosiakkatrina: The only thing I can think of doing is using the map to get the db value, but if we're thinking of refactoring this to pass the value in the params at some point, would that still be useful?

app/controllers/schemes_controller.rb
baarkerlounger: @kosiakkatrina I think the best solution to that for now is something like:

```ruby
if Scheme::CONSTANT[scheme_params[:arrangement_type]] == "D"
```

etc. So still a string comparison really but at least you have a single source of truth for what that string is in the code base.

app/models/form/setup/questions/scheme_id.rb
kosiakkatrina: Yeah, that's fair. I didn't realise we introduced that! Is that already setting the state?

app/controllers/schemes_controller.rb
baarkerlounger: I think so yes.

A) Because I'm not sure we'll ever actually do that refactor, and B) because even if we do, I don't think it makes that refactor any harder and makes this easier to change in the meantime?

app/controllers/schemes_controller.rb
kosiakkatrina: Okay, had the same thought. Happy to update

app/models/form/setup/questions/scheme_id.rb
baarkerlounger: Yep, it should be being set when you confirm on the check answers page

https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/blob/main/app/views/schemes/check_answers.html.erb#L134

app/controllers/schemes_controller.rb
baarkerlounger: We're checking if the `arrangement_type` is that the organisation providing support services is separate from the organisation owning the stock?

And our check is just for nil, should it also check for the scenario where a managing_org_id param is present but doesn't match the owning_org_id or should that never happen?

spec/features/schemes_spec.rb
```diff
@@ -623,6 +630,20 @@
...
+              choose "Another organisation"
+              click_button "Save and continue"
+              expect(page).to have_content(another_organisation.name)
+            end
+
+            it "does not display the answer if it's changed to the same support provider" do
+              click_link("Change", href: "/schemes/#{scheme.id}/details?check_answers=true", match: :first)
+              choose "The same organisation that owns the housing stock"
+              click_button "Save and continue"
+              expect(page).not_to have_content("Organisation providing support")
```
moarpheus, line 645: It would be good to have controller request test, that will investigate the state of the scheme after selecting the same org ( basically making sure that values are updated in DB ).

app/models/form/setup/questions/scheme_id.rb
kosiakkatrina: What about the data imports?

kosiakkatrina: Also, this doesn't seem to validate? I can create a record with values missing

kosiakkatrina:
@Dan13L3 Is "Please answer all the required questions" okay for the schemes validation? We have individual validations on specific pages, this one is for the overall scheme completeness. Can make it more specific if that makes sense

Dan13L3:
@kosiakkatrina We could have an overall message of: 'You must answer all the required questions for this scheme'. But I wonder if it's more helpful for the user to just list all the individual answers that are missing and have a specific error for each? I can create a list of them that follows the 'Check answers' pattern above if helpful?

kosiakkatrina:
> @kosiakkatrina We could have an overall message of: 'You must answer all the required questions for this scheme'. But I wonder if it's more helpful for the user to just list all the individual answers that are missing and have a specific error for each? I can create a list of them that follows the 'Check answers' pattern above if helpful?

@paulrobertlloyd Should we have individual answer validations or an overall one for the CYA schemes page?

paulrobertlloyd:
We should ideally have both; an error summary at the top of the page which lists all the errors preventing the scheme from being saved, and then highlight the individual summary rows that are invalid.

We experimented with doing something like this when we were going to allow users to review a log before submitting:

Screenshot 2022-07-25 at 12 57 23

Here is the HTML for the first row shown in the example above:

```html
<div class="govuk-summary-list__row">
    <dt class="govuk-summary-list__key">
        Tenant has seen the privacy notice
    </dt>
    <dd class="govuk-summary-list__value app-summary-list__value--error">
        <p id="privacy-notice" class="govuk-error-message">
            <span class="govuk-visually-hidden">Error:</span> Select if the tenant has seen the DLUHC privacy notice
        </p>
        <span class="app-!-colour-muted">You didn’t answer this question</span>
    </dd>
    <dd class="govuk-summary-list__actions"> 
        <a class="govuk-link" href="/logs/P0NJL/household-characteristics/privacy-notice">Answer<span class="govuk-visually-hidden"> if tenant has seen the DLUHC privacy notice</span></a>
    </dd>
</div>
```

CSS can be found here: https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data-prototype/blob/62810962aa01076f9ba1c8e3a73bf198d30814a7/app/assets/stylesheets/application.scss#L68-L71

app/models/form/setup/questions/scheme_id.rb
```diff
@@ -21,7 +21,7 @@ def answer_options
 
   def displayed_answer_options(case_log)
     organisation = case_log.owning_organisation || case_log.created_by&.organisation
-    schemes = organisation ? Scheme.select(:id).where(owning_organisation_id: organisation.id) : Scheme.select(:id)
+    schemes = organisation ? Scheme.select(:id).where(owning_organisation_id: organisation.id).where(confirmed: true) : Scheme.select(:id).where(confirmed: true)
```
baarkerlounger: Not important but any reason not to just make this 1 where query? (Don't think we need `where().where()` right?

spec/requests/schemes_controller_spec.rb
```diff
@@ -557,13 +557,22 @@
...
 
       before do
         sign_in user
         patch "/schemes/#{scheme_to_update.id}", params:
       end
 
+      context "when confirming the scheme" do
+        let(:params) { { scheme: { owning_organisation_id: user.organisation.id, arrangement_type: "V", confirmed: true, page: "check-answers" } } }
+
+        it "does not allow the scheme to be confirmed" do
```
baarkerlounger, line 570: Is there a context description missing here? This should only be true when data is missing or invalid right?

baarkerlounger (approved):

kosiakkatrina:
> We should ideally have both; an error summary at the top of the page which lists all the errors preventing the scheme from being saved, and then highlight the individual summary rows that are invalid.
> 
> We experimented with doing something like this when we were going to allow users to review a log before submitting:
> 
> Screenshot 2022-07-25 at 12 57 23
> 
> Here is the HTML for the first row shown in the example above:
> 
> ```
> <div class="govuk-summary-list__row">
>     <dt class="govuk-summary-list__key">
>         Tenant has seen the privacy notice
>     </dt>
>     <dd class="govuk-summary-list__value app-summary-list__value--error">
>         <p id="privacy-notice" class="govuk-error-message">
>             <span class="govuk-visually-hidden">Error:</span> Select if the tenant has seen the DLUHC privacy notice
>         </p>
>         <span class="app-!-colour-muted">You didn’t answer this question</span>
>     </dd>
>     <dd class="govuk-summary-list__actions"> 
>         <a class="govuk-link" href="/logs/P0NJL/household-characteristics/privacy-notice">Answer<span class="govuk-visually-hidden"> if tenant has seen the DLUHC privacy notice</span></a>
>     </dd>
> </div>
> ```
> 
> CSS can be found here: https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data-prototype/blob/62810962aa01076f9ba1c8e3a73bf198d30814a7/app/assets/stylesheets/application.scss#L68-L71

@paulrobertlloyd Updated this now

app/views/schemes/_scheme_summary_list_row.html.erb
```diff
@@ -0,0 +1,22 @@
...
+            <%= details_html(attribute) %>
+        </dd>
+    <% else %>
+        <dd class="govuk-summary-list__value">
+            <%= details_html(attribute) %>
+        </dd>
+    <% end %>
+    <% if !scheme.confirmed? || attribute[:name] == "Name" %>
+        <dd class="govuk-summary-list__actions">
+            <a class="govuk-link" href="<%= change_link %>">Change</a>
```
baarkerlounger, line 19: Think this should probably say "Answer" rather than "Change" if it's missing but happy for that to be a separate PR.

baarkerlounger (approved):

PR #778, opened 2022-07-22
CLDC-1389: Set default option on accessible autocomplete

baarkerlounger:
image

### DIFF
--- a/app/frontend/controllers/accessible_autocomplete_controller.js
+++ b/app/frontend/controllers/accessible_autocomplete_controller.js
@@ -6,14 +6,14 @@ import { enhanceOption, suggestion, sort } from '../modules/search'
 export default class extends Controller {
   connect () {
     const selectEl = this.element
-    const selectOptions = Array.from(selectEl.options)
+    const selectOptions = Array.from(selectEl.options).filter(function (option, index, arr) { return option.value !== '' })
     const options = selectOptions.map((o) => enhanceOption(o))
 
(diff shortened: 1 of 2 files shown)


dushan-madetech (approved):

paulrobertlloyd:
Erm… I think the default, when enhanced, should be an empty value.

baarkerlounger:
@paulrobertlloyd cool, could you comment that on the ticket then please?

https://digital.dclg.gov.uk/jira/browse/CLDC-1379

app/frontend/controllers/accessible_autocomplete_controller.js
```diff
@@ -13,7 +13,7 @@ export default class extends Controller {
     const rawFieldName = matches ? `${matches[1]}[${matches[2]}_raw]` : ''
 
     accessibleAutocomplete.enhanceSelectElement({
-      defaultValue: '',
+      defaultValue: options[0].name,
```
smeny: Is it guaranteed that options[0] is always present?

paulrobertlloyd:
> @paulrobertlloyd cool, could you comment that on the ticket then please?

Updated the ACs.

baarkerlounger:
@smeny @dushan-madetech @paulrobertlloyd 

Changed this to filter out the "Select an option" value (so it doesn't appear if you start typing "Se..." either, and left the box blank by default when JS is enabled.

PR #779, opened 2022-07-25
Use new formbuilder config option to disable bold radio labels

paulrobertlloyd:
The GOV.UK Formbuilder [now provides an option to disable the default behaviour for emboldening radio labels with hint text](https://github.com/DFE-Digital/govuk-formbuilder/releases/tag/v3.1.1). This PR uses that option to disable this behaviour, and removes the individual overrides on certain radio collections.


moarpheus (approved):
Fantastic, thanks Paul!

PR #780, opened 2022-07-25
CLDC-1372 Show inferred location when checking log answers

moarpheus:
For a supported housing log, when showing the location on the check your answers page, show:

- the locations’s inferred LA to the right of the postcode in grey (using same markup as used for reviewing property information on a general needs log)

Screenshot 2022-07-25 at 14 28 08

### DIFF
--- a/app/models/case_log.rb
+++ b/app/models/case_log.rb
@@ -110,6 +110,14 @@ class CaseLog < ApplicationRecord
     end
   end
 
+  def location_admin_district
+    if location
+      location.location_admin_district
+    else
+      super
+    end
+  end
+
   def postcode_full
     if location
       location.postcode
--- a/app/models/case_log.rb
+++ b/app/models/case_log.rb
@@ -110,14 +110,6 @@ class CaseLog < ApplicationRecord
     end
   end
 
-  def location_admin_district
-    if location
-      location.location_admin_district
-    else
-      super
-    end
-  end
-
   def postcode_full
     if location
       location.postcode
--- a/app/models/form/question.rb
+++ b/app/models/form/question.rb
@@ -53,12 +53,12 @@ class Form::Question
   def get_inferred_answers(case_log)
     return [] unless inferred_answers
 
-    enabled_inferred_answers(inferred_answers, case_log).keys.map do |answer|
-      question = form.get_question(answer, case_log)
+    enabled_inferred_answers(inferred_answers, case_log).keys.map do |question_id|
+      question = form.get_question(question_id, case_log)
       if question.present?
-        question.label_from_value(case_log[answer])
+        question.label_from_value(case_log[question_id])
       else
-        Array(answer.to_s.split(".")).inject(case_log) { |log, method| log.present? ? log.public_send(*method) : "" }
+        Array(question_id.to_s.split(".")).inject(case_log) { |log, method| log.present? ? log.public_send(*method) : "" }
       end
     end
   end
--- a/app/models/form/question.rb
+++ b/app/models/form/question.rb
@@ -66,7 +66,7 @@ class Form::Question
   def get_extra_check_answer_value(case_log)
     return nil unless extra_check_answer_value
 
-    case_log.public_send(extra_check_answer_value)
+    case_log.form.get_question("la", nil).label_from_value(case_log.la)
   end
 
   def read_only?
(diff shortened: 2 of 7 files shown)


app/views/form/_check_answers_summary_list.html.erb
```diff
@@ -3,7 +3,12 @@
     <% summary_list.row do |row| %>
       <% row.key { question.check_answer_label.to_s.presence || question.header.to_s } %>
       <% row.value do %>
-        <%= get_answer_label(question, @case_log) %><br>
+        <%= get_answer_label(question, @case_log) %>
```
paulrobertlloyd: This should match what we do in property information, with a larger space between the value and the inferred value – should be something like this:

```suggestion
        <span class="govuk-!-margin-right-4"><%= get_answer_label(question, @case_log) %></span>
```

app/models/case_log.rb
```diff
@@ -110,6 +110,14 @@ def la
     end
   end
 
+  def location_admin_district
+    if location
+      location.location_admin_district
+    else
+      super
```
baarkerlounger: Do we have an attribute on `CaseLog` for this? Isn't this equivalent to LA? (LA has the code but the code can be translated?)

baarkerlounger: i.e. What is `la` and what is `location_admin_district`?

app/models/form/question.rb
```diff
@@ -51,16 +53,22 @@ def answer_label(case_log)
   def get_inferred_answers(case_log)
     return [] unless inferred_answers
 
-    enabled_inferred_answers(inferred_answers, case_log).keys.map do |x|
-      question = form.get_question(x, case_log)
+    enabled_inferred_answers(inferred_answers, case_log).keys.map do |answer|
```
baarkerlounger: Is this actually an answer or is it a question id?

app/models/case_log.rb
moarpheus: Now this is where I also struggled in the beginning - or maybe I still do? In our conversation people seem to refer to the LA by its name ( e.g. "Westmintser" ), however in codebase, this appears to be the actual CODE ( e.g. "E08000003" ). Not aware of 'translation' between those two. In my previous card I coded to add **location_admin_district** column to the location, to populate on create/update alongside with the **location_code**. 
location_admin_district - this name comes from API output, same named field holds this value.

So to sum up for **location**:

location_code is this "E08000003" ( presumably the same what la is )
location_admin_district is this -  "Westminster"

baarkerlounger: Right so currently we have all of these values in the form definition JSON (since we need to be able to ask people for the location if they don't know the postcode). So we should only need to store LA value *E08000003* and then should be able to get/display LA label by doing:

```ruby
case_log.form.get_question("la").label_from_value("case_log["la"])
```

app/models/form/question.rb
```diff
@@ -51,16 +53,22 @@ def answer_label(case_log)
...
-        Array(x.to_s.split(".")).inject(case_log) { |o, a| o.present? ? o.public_send(*a) : "" }
+        Array(question_id.to_s.split(".")).inject(case_log) { |log, method| log.present? ? log.public_send(*method) : "" }
       end
     end
   end
 
+  def get_extra_check_answer_value(case_log)
+    return nil unless extra_check_answer_value
+
+    case_log.form.get_question("la", nil).label_from_value(case_log.la)
```
baarkerlounger: Shouldn't this be in `app/models/form/setup/questions/location_id.rb`?

app/models/form/setup/questions/location_id.rb
```diff
@@ -5,6 +5,7 @@ def initialize(_id, hsh, page)
     @header = "Which location is this log for?"
     @hint_text = ""
     @type = "radio"
+    @extra_check_answer_value = "location_admin_district"
```
baarkerlounger: Given that the value of this isn't actually used for anything and all the logic is in the method now anyway do we need this at all? 

i.e. could we just have

```ruby
# app/models/form/question.rb

def get_extra_check_answer_value(case_log)
  return nil
end


# app/models/form/setup/questions/location_id.rb

def get_extra_check_answer_value(case_log)
  case_log.form.get_question("la", nil).label_from_value(case_log.la)
end
```

moarpheus: With the way it is implemented now, I guess not.

baarkerlounger (approved):

PR #781, opened 2022-07-25
CLDC-1381-Postcode-display-formatting

Ted-U:


app/models/case_log.rb
```diff
@@ -567,13 +566,19 @@ def age_under_16?(person_num)
     public_send("age#{person_num}") && public_send("age#{person_num}") < 16
   end
 
+  def format_postcode(value)
```
baarkerlounger: I think this probably doesn't really belong in the case log model/class since we display postcodes in multiple places (e.g. scheme locations).

There's a postcode service class that would be a better fit, but take a look at how we do date formatting for display https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/blob/main/config/initializers/date_formats.rb

app/helpers/tab_nav_helper.rb
```diff
@@ -1,13 +1,14 @@
 module TabNavHelper
   include GovukLinkHelper
+  include Helpers
```
moarpheus: I don't think you need this inclusion

app/models/case_log.rb
```diff
@@ -568,12 +567,12 @@ def age_under_16?(person_num)
   end
 
   def process_postcode_changes!
-    self.postcode_full = upcase_and_remove_whitespace(postcode_full)
+    self.postcode_full = format_postcode(postcode_full)
```
moarpheus: I think this is where data is being processed before saved in DB, so no changes needed here.

app/models/case_log.rb
```diff
@@ -568,12 +567,12 @@ def age_under_16?(person_num)
...
     process_postcode(postcode_full, "postcode_known", "is_la_inferred", "la")
   end
 
   def process_previous_postcode_changes!
-    self.ppostcode_full = upcase_and_remove_whitespace(ppostcode_full)
+    self.ppostcode_full = format_postcode(ppostcode_full)
```
moarpheus: same here

lib/ext/string.rb
```diff
@@ -0,0 +1,15 @@
+class String
```
baarkerlounger: I think just having a formatter method that takes a string is maybe better than monkey patching a core class. If you do want to go this way though maybe look into refinements?

spec/features/form/postcode_format_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+      managing_organisation: user.organisation,
+    )
+  end
+  let(:id) { case_log.id }
+
+  before do
+    sign_in user
+  end
+
+  context "when a postcode is input" do
```
moarpheus: I'm not sure we need feature test for this, we can write a unit test for String class looking into all possible cases and request tests where the postcode formatting will appear on the front.

app/models/case_log.rb
```diff
@@ -57,7 +56,7 @@ class CaseLog < ApplicationRecord
   scope :filter_by_id, ->(id) { where(id:) }
   scope :filter_by_tenant_code, ->(tenant_code) { where("tenancycode ILIKE ?", "%#{tenant_code}%") }
   scope :filter_by_propcode, ->(propcode) { where("propcode ILIKE ?", "%#{propcode}%") }
-  scope :filter_by_postcode, ->(postcode_full) { where("postcode_full ILIKE ?", "%#{postcode_full.gsub(/\s+/, '')}%") }
+  scope :filter_by_postcode, ->(postcode_full) { where("postcode_full ILIKE ?", "%#{UKPostcode.parse(postcode_full)}%") }
```
smeny: `to_s` in not needed in string interpolation (rubocop)

spec/requests/case_logs_controller_spec.rb
```diff
@@ -792,7 +792,7 @@
         it "downloads logs matching both csv and filter logs" do
           get "/logs?status[]=completed&search=#{postcode}", headers:, params: {}
           csv = CSV.parse(response.body)
-          expect(csv.count).to eq(2)
+          expect(csv.count).to eq(1)
```
smeny: I don't understand why this was previously `2`.
Only the log matching both status and postcode should be returned.

app/models/case_log.rb
```diff
@@ -118,6 +117,22 @@ def postcode_full
     end
   end
 
+  def postcode_full=(postcode)
+    if postcode
+      super UKPostcode.parse(postcode).to_s
+    else
+      super nil
```
smeny, line 124: passing nil to be library crashes the parser

smeny:
Maybe an extra refactor for tomorrow, we could abstract our UKPostcode library within the postcode service we currently uses.

app/models/case_log.rb
kosiakkatrina: Can this be extracted as a method? We've got the same code in 3 places

spec/features/schemes_helpers.rb
```diff
@@ -68,7 +68,7 @@ def fill_in_and_save_location
   end
 
   def fill_in_and_save_second_location
-    fill_in "Postcode", with: "XX1 1XX"
+    fill_in "Postcode", with: "AA1 2AA"
```
kosiakkatrina: Maybe worth flipping this so that we enter a postcode with no spaces and it saves and formats it with spaces?

kosiakkatrina:
One case that doesn't seem to work for me locally. When searching for a case log with this postcode "AA1 2AA", if I enter "AA12" that case log won't show up. Not sure if it's intended tho.

spec/requests/case_logs_controller_spec.rb
baarkerlounger: Header row. i.e. 1 log would return 2 rows. 1 row suggests no logs match at all.

app/models/case_log.rb
baarkerlounger: Does super need to run for the nil case? Can it not just be:

```ruby
  super UKPostcode.parse(postcode) if postcode
```

config/application.rb
```diff
@@ -12,7 +12,7 @@
 # require "action_text/engine"
 require "action_view/railtie"
 # require "action_cable/engine"
-# require "sprockets/railtie"
+# require "sprockets/railtie"
```
baarkerlounger: This looks accidental?

app/models/case_log.rb
smeny: We can assign a nil value to some fields, I think that's what the form does when it invalidates some fields.

spec/features/schemes_helpers.rb
smeny: Not a bad idea, the issue here was that invalid postcodes are not parsed properly (returns the same input string).

smeny:
> One case that doesn't seem to work for me locally. When searching for a case log with this postcode "AA1 2AA", if I enter "AA12" that case log won't show up. Not sure if it's intended tho.

That's because "AA12" won't match the string with the space (we enter into pattern-matching territory). As there wasn't a test case for this, I assumed we didn't need to support such as case.

kosiakkatrina (approved):

PR #782, opened 2022-07-26
Add location "available from" column

smeny:
Screenshot 2022-07-25 at 15 45 27
Screenshot 2022-07-25 at 15 48 10

### DIFF
--- a/app/views/schemes/check_answers.html.erb
+++ b/app/views/schemes/check_answers.html.erb
@@ -3,7 +3,6 @@
 
 <%= form_for(@scheme, as: :scheme, method: :patch) do |f| %>
   <div class="govuk-grid-row">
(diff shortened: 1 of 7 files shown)


app/views/schemes/check_answers.html.erb
```diff
@@ -3,7 +3,7 @@
 
 <%= form_for(@scheme, as: :scheme, method: :patch) do |f| %>
   <div class="govuk-grid-row">
-    <div class="govuk-grid-column-three-quarters-from-desktop">
+    <div class="govuk-grid-column-full">
```
moarpheus: We can remove this completely as we are not overriding anything that comes from layouts!

moarpheus (approved):

PR #785, opened 2022-07-26
Fix rent value check

baarkerlounger:
Rent value check field gets cleared because showing/hiding its page depends_on rent soft validation being triggered. The soft validation check depends on lettype which is only derived after we check which questions should get cleared.

This removes the `lettype` derived variable from the soft validation calculation such that both work.

### DIFF
--- a/app/models/case_log.rb
+++ b/app/models/case_log.rb
@@ -619,9 +619,9 @@ private
   end
 
   def get_lettype
-    return unless renttype.present? && needstype.present? && owning_organisation.present? && owning_organisation[:provider_type].present?
+    return unless rent_type.present? && needstype.present? && owning_organisation.present? && owning_organisation[:provider_type].present?
 
-    case RENT_TYPE_MAPPING_LABELS[renttype]
+    case RENT_TYPE_MAPPING_LABELS[RENT_TYPE_MAPPING[rent_type]]
     when "Social Rent"
       if is_supported_housing?
         owning_organisation[:provider_type] == "PRP" ? 2 : 4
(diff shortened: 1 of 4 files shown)


app/models/case_log.rb
```diff
@@ -619,9 +619,9 @@ def get_has_benefits
   end
 
   def get_lettype
-    return unless renttype.present? && needstype.present? && owning_organisation.present? && owning_organisation[:provider_type].present?
+    return unless rent_type.present? && needstype.present? && owning_organisation.present? && owning_organisation[:provider_type].present?
 
-    case RENT_TYPE_MAPPING_LABELS[renttype]
+    case RENT_TYPE_MAPPING_LABELS[RENT_TYPE_MAPPING[rent_type]]
```
smeny, line 624: Since we are refactoring this, any reason why we use the labels for comparison?

spec/services/imports/case_logs_import_service_spec.rb
```diff
@@ -229,6 +229,31 @@ def open_file(directory, filename)
...
+          lettype: 1,
+          soft_max: 900,
+          hard_max: 1500,
+          soft_min: 500,
+          hard_min: 100,
+        )
+      end
+
+      it "completes the log" do
+        allow(logger).to receive(:warn)
```
smeny: We need to remove this line

app/models/case_log.rb
baarkerlounger: Just for readability, so you know what each case actually represents

smeny (approved):

PR #786, opened 2022-07-27
Fix next incomplete section button when first page of next subsection…

baarkerlounger:
… is not routed to

### DIFF
--- a/app/models/form.rb
+++ b/app/models/form.rb
@@ -79,7 +79,7 @@ class Form
     when :in_progress
       "#{next_subsection.id}/check_answers".dasherize
     when :not_started
-      first_question_in_subsection = next_subsection.pages.first.id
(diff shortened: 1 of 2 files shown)


app/models/form.rb
```diff
@@ -79,7 +79,7 @@ def next_incomplete_section_redirect_path(subsection, case_log)
     when :in_progress
       "#{next_subsection.id}/check_answers".dasherize
     when :not_started
-      first_question_in_subsection = next_subsection.pages.first.id
+      first_question_in_subsection = next_subsection.pages.find { |page| page.routed_to?(case_log, nil) }.id
```
smeny, line 82: is `routed_to?` guaranteed to return a result?

baarkerlounger: Not strictly but if you had a section where no pages were routed to you've probably got an issue with your form design already (and a hard crash might be good to alert us to that fact)

smeny (approved):

PR #787, opened 2022-07-27
CLDC-1171 update design for inviting and editing users

moarpheus:
Screenshot 2022-07-29 at 12 34 39
Screenshot 2022-07-29 at 12 34 33
Screenshot 2022-07-29 at 12 34 20

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -74,7 +74,7 @@ class UsersController < ApplicationController
       redirect_to created_user_redirect_path
     else
       unless @resource.errors[:organisation].empty?
-        @resource.errors.add(:organisation_id, message: @user.errors[:organisation])
+        @resource.errors.add(:organisation_id, message: @resource.errors[:organisation])
         @resource.errors.delete(:organisation)
       end
       render :new, status: :unprocessable_entity
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -65,7 +65,6 @@ class UsersController < ApplicationController
   def new
     @organisation_id = params["organisation_id"]
     @resource = User.new
-    debugger
   end
 
   def create
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -69,7 +69,8 @@ class UsersController < ApplicationController
 
   def create
     @user = User.new(user_params.merge(org_params).merge(password_params))
-    if @user.save
+
+    if valid_attributes? && @user.save
       redirect_to created_user_redirect_path
     else
       unless @user.errors[:organisation].empty?
@@ -102,6 +103,14 @@ class UsersController < ApplicationController
   end
 
 private
+
+  def valid_attributes?
+    @user.valid?
+    if user_params[:role] && !current_user.assignable_roles.key?(user_params[:role].to_sym)
+      @user.errors.add :role, I18n.t("validations.role.invalid")
+    end
+    @user.errors.empty?
+  end
 
   def format_error_messages
     errors = @user.errors.to_hash
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -109,6 +109,9 @@ private
     if user_params[:role].present? && !current_user.assignable_roles.key?(user_params[:role].to_sym)
       @resource.errors.add :role, I18n.t("validations.role.invalid")
     end
+    if user_params[:role].empty?
+      @resource.errors.add :role, I18n.t("activerecord.errors.models.user.attributes.role.blank")
+    end
     @resource.errors.empty?
   end
 
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -63,29 +63,21 @@ class UsersController < ApplicationController
   end
 
   def new
+    debugger
     @organisation_id = params["organisation_id"]
-    @resource = User.new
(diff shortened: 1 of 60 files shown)


app/models/user.rb
```diff
@@ -149,4 +153,16 @@ def can_toggle_active?(user)
   def valid_for_authentication?
     super && active?
   end
+
+private
+
+  def validate_email
+    unless email_valid?(email)
+      if User.exists?(["email LIKE ?", "%#{email}%"])
+        errors.add :email
```
baarkerlounger: Do we not need to add a message here?

spec/requests/users_controller_spec.rb
```diff
@@ -423,7 +423,7 @@
       context "when a search parameter is passed" do
         let!(:other_user_2) { FactoryBot.create(:user, organisation: user.organisation, name: "joe", email: "other@example.com") }
         let!(:other_user_3) { FactoryBot.create(:user, name: "User 5", organisation: user.organisation, email: "joe@example.com") }
-        let!(:other_org_user) { FactoryBot.create(:user, name: "User 4", email: "joe@other_example.com") }
```
baarkerlounger, line 426: Why removing the `_` here?

spec/requests/users_controller_spec.rb
```diff
@@ -935,6 +922,24 @@
...
+              email: "",
+              role: "support",
+            },
+          }
+        end
+
+        it "shows an error" do
+          request
+          expect(response).to have_http_status(:unprocessable_entity)
+          expect(page).to have_content(I18n.t("validations.role.invalid"))
```
baarkerlounger: It should also have an error on `email` here right?

spec/requests/users_controller_spec.rb
baarkerlounger: Because it's not a valid character for a domain, right...

spec/models/form/setup/questions/created_by_id_spec.rb
```diff
@@ -20,6 +20,10 @@
     }
   end
 
+  before do
```
baarkerlounger: Is this just to avoid hitting the database?

moarpheus: I have validations for name presence, can't create one without it - so have to avoid using DB, but presumably this test is still valid as we potentially have users on staging/prod maybe that don't have names

app/models/user.rb
moarpheus: nope

spec/requests/users_controller_spec.rb
moarpheus: Yup, previously validations were't working properly, hence this was accepted.

app/controllers/users_controller.rb
```diff
@@ -112,6 +108,17 @@ def reactivate
 
 private
 
+  def valid_attributes?
```
kosiakkatrina: Can we rename this? From the name I'd expect this to return a bool and do nothing else, but it's adding errors  to the resource, which isn't clear straightaway

app/controllers/users_controller.rb
```diff
@@ -161,7 +168,7 @@ def created_user_redirect_path
   end
 
   def find_resource
-    @user = params[:id] ? User.find_by(id: params[:id]) : current_user
+    @user = User.find_by(id: params[:id]) || User.find_by(id: params[:user_id]) || current_user
```
kosiakkatrina: Should we search by :user_id before we do :id? Is there ever a scenario where :id would be passed as organisation id?

baarkerlounger: Alternatively maybe specify by action explicitly if we know when we're expecting each

app/controllers/users_controller.rb
baarkerlounger: Perhaps `validate_attributes`?

app/models/user.rb
baarkerlounger: Does this not need a message?

config/initializers/filter_parameter_logging.rb
```diff
@@ -2,5 +2,5 @@
 
 # Configure sensitive parameters which will be filtered from the log file.
 Rails.application.config.filter_parameters += %i[
-  passw secret token _key crypt salt certificate otp ssn
```
baarkerlounger, line 5: :joy:  Don't want to ask how much debugging frustration finding this config took

spec/models/form/setup/questions/created_by_id_spec.rb
baarkerlounger: That might be worth a comment in the code since it's not obvious

smeny: I don't think we should test a case where the user is saved in an invalid state (should never happen or would be fixed on our side).

baarkerlounger: Is that validation correct? I thought the reason we did that originally was that we can't guarantee the value is populated in the legacy system? If so what do we do for import?

app/controllers/users_controller.rb
smeny: Changed so it does one thing (validate), we check if we have errors separately

app/controllers/users_controller.rb
smeny: Changed the order

PR #788, opened 2022-07-27
Add the bottom guidance partial

smeny:


app/views/form/guidance/_scheme_selection.html.erb
```diff
@@ -0,0 +1,5 @@
+<% if current_user.data_provider? %>
+  <p class="govuk-body">If you can’t find the supported housing service you’re looking for or not sure which to choose, contact a data coordinator at <%= current_user.organisation.name %>.</p>
+<% elsif current_user.data_coordinator? %>
+  <p class="govuk-body">or <%= govuk_link_to "create a new supported housing service", new_scheme_path %></p>
```
baarkerlounger, line 4: Should "Create" be capitalised here? @Dan13L3

baarkerlounger (approved):

app/views/form/guidance/_scheme_selection.html.erb
smeny: According to the content design in https://digital.dclg.gov.uk/jira/browse/CLDC-1325 I don't think so.

Dan13L3: @baarkerlounger Should be fine lower case, as it's supposed to run on from the main content on the page

PR #791, opened 2022-07-28
Search schemes by location name

baarkerlounger:
https://communities-govuk.slack.com/archives/C01QGR78H9S/p1659007533170599

Note. it may not be immediately obvious *why* a scheme has matched a search result since we don't show location names in the list

image
image


moarpheus (approved):
Probably need to amended the search bar title to reflect that, otherwise can get back unexpected results without obvious reason why.

PR #797, opened 2022-08-01
Cldc 1394 schemes cya

kosiakkatrina:
This is a part of CLDC 1394 ticket that addresses some of the navigation around scheme check answers:

- [x] When you create a scheme and add locations it should route to Check your answers page (Scheme details tab)
- [x] When you edit a confirmed scheme name it should route to Check your answers page (Scheme details tab) This would only show change links for the editable fields
- [x] When you add a location to an already confirmed scheme it should route to Check your answers page (location tab). Only the newly added location can be editable, can only edit location names for the others.
- [x] "Create scheme" button is replaced with "Save” button on a confirmed scheme, saving the locations marks them as confirmed, the confirmation banner should say ‘updated’ not ‘created’
- [x] Clicking Save takes you back to the locations list for the scheme - this is not implemented, it takes the user back to the view scheme with only the editable fields having a "change" link
- [x] Clicking that link takes you to the change name form, and clicking save takes you back to CYA (Scheme details tab)
- [x] Navigating to form pages for confirmed scheme directly redirects you to view scheme page

### DIFF
--- a/app/controllers/schemes_controller.rb
+++ b/app/controllers/schemes_controller.rb
@@ -5,6 +5,7 @@ class SchemesController < ApplicationController
   before_action :authenticate_user!
   before_action :find_resource, except: %i[index]
   before_action :authenticate_scope!
+  before_action :redirect_if_scheme_confirmed, only: %i[primary_client_group confirm_secondary_client_group secondary_client_group support details]
 
   def index
     redirect_to schemes_organisation_path(current_user.organisation) unless current_user.support?
@@ -253,5 +254,9 @@ private
     if %w[show locations primary_client_group confirm_secondary_client_group secondary_client_group support details check_answers edit_name].include?(action_name) && !((current_user.organisation == @scheme&.owning_organisation) || current_user.support?)
       render_not_found and return
     end
+  end
+
+  def redirect_if_scheme_confirmed
+    redirect_to @scheme if @scheme.confirmed?
   end
 end
--- a/app/controllers/schemes_controller.rb
+++ b/app/controllers/schemes_controller.rb
@@ -7,7 +7,6 @@ class SchemesController < ApplicationController
   before_action :authenticate_scope!
 
   def index
-    flash[:notice] = "#{Scheme.find(params[:scheme_id].to_i).service_name} has been created." if params[:scheme_id]
     redirect_to schemes_organisation_path(current_user.organisation) unless current_user.support?
     all_schemes = Scheme.all.order("service_name ASC")
 
@@ -52,11 +51,17 @@ class SchemesController < ApplicationController
 
     check_answers = params[:scheme][:check_answers]
     page = params[:scheme][:page]
+    scheme_previously_confirmed = @scheme.confirmed?
 
     validation_errors scheme_params
     if @scheme.errors.empty? && @scheme.update(scheme_params)
       if scheme_params[:confirmed] == "true"
-        @scheme.locations.each {|location| location.update!(confirmed:true)}
+        @scheme.locations.each { |location| location.update!(confirmed: true) }
+        flash[:notice] = if scheme_previously_confirmed
+                           "#{@scheme.service_name} has been updated."
+                         else
+                           "#{@scheme.service_name} has been created."
+                         end
         redirect_to scheme_path(@scheme)
       elsif check_answers
         if confirm_secondary_page? page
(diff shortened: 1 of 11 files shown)


app/helpers/check_answers_helper.rb
```diff
@@ -11,6 +11,18 @@ def display_answered_questions_summary(subsection, case_log, current_user)
...
   end
 
+  def can_change_scheme_answer?(attribute_name, scheme)
+    editable_attributes = current_user.support? ? ["Name", "Confidential information", "Housing stock owned by"] : ["Name", "Confidential information"]
+    !scheme.confirmed? || editable_attributes.include?(attribute_name) 
+  end
+
+  def get_location_change_link_href(scheme, location)
+    if location.confirmed?
+      "/schemes/#{scheme.id}/locations/#{location.id}/edit-name"
```
jamesrose:
```suggestion
      location_edit_name_path(scheme, location)
```

Can we use route helper methods for both conditions?

app/controllers/schemes_controller.rb
```diff
@@ -48,7 +48,10 @@ def update
 
     validation_errors scheme_params
     if @scheme.errors.empty? && @scheme.update(scheme_params)
-      if check_answers
+      if scheme_params[:confirmed] == "true"
+        @scheme.locations.each {|location| location.update!(confirmed:true)}
```
jamesrose: It's possible to shorten this slightly to:

```suggestion
        @scheme.locations.update!(confirmed: true)
```

app/views/schemes/_scheme_summary_list_row.html.erb
```diff
@@ -1,4 +1,4 @@
-<div class="<%= "govuk-summary-list__row #{scheme.confirmed? && attribute[:name] != 'Name' ? 'govuk-summary-list__row--no-actions' : ''}" %>">
+<div class="<%= "govuk-summary-list__row #{scheme.confirmed? && !["Name", "Confidential information", "Housing stock owned by"].include?(attribute[:name]) ? 'govuk-summary-list__row--no-actions' : ''}" %>">
```
jamesrose: Is it possible to re-use `can_change_scheme_answer?` here?

spec/features/schemes_spec.rb
```diff
@@ -285,16 +285,20 @@
...
                 it "allows you to edit the newly added location" do
-                  click_link 'Locations'
-                  expect(page).to have_link(nil , href: /edit/)
+                  click_link "Locations"
+                  expect(page).to have_link(nil, href: /edit/)
                 end
 
                 context "when you click save" do
-                  xit "takes you to view location tab and displays a banner" do
+                  xit "takes you to view location tab" do
```
jamesrose: Should this still be disabled?

kosiakkatrina: Yup! I just extracted it into a separate test

kosiakkatrina:
Might need to run migrations to mark the locations for completed schemes as completed?

app/controllers/schemes_controller.rb
```diff
@@ -64,23 +72,43 @@ def update
   end
 
   def primary_client_group
-    render "schemes/primary_client_group"
+    if @scheme.confirmed?
```
jamesrose: It looks like you're doing the same thing for each of these group actions. Do you think this would help the repetition?

```ruby
before_filter :redirect_if_scheme_confirmed, only: [:primary_client_group, ...]

# ... 

private

def redirect_if_scheme_confirmed
  redirect_to @scheme if @scheme.confirmed?
end
```

app/views/locations/edit.html.erb
```diff
@@ -9,7 +9,7 @@
 
 <%= render partial: "organisations/headings", locals: { main: "Add a location to this scheme", sub: @scheme.service_name } %>
 
-<%= form_for(@location, method: :patch, url: location_path) do |f| %>
+<%= form_for(@location, method: :patch, url: location_path(location_id: @location.id)) do |f| %>
```
baarkerlounger: Think you can just pass `@location` here?

baarkerlounger:
@kosiakkatrina yep, is it just marking all current prod locations as confirmed?

Also need to resolve some conflicts.

app/controllers/schemes_controller.rb
kosiakkatrina: Awesome! Ty!

kosiakkatrina:
> @kosiakkatrina yep, is it just marking all current prod locations as confirmed?
> 
> Also need to resolve some conflicts.

@baarkerlounger Well, all of the locations that belong to confirmed schemes (Which I think currently are all locations, yes)

baarkerlounger (approved):

PR #804, opened 2022-08-02
Return 404 when scheme not found

baarkerlounger:


smeny (approved):

app/controllers/locations_controller.rb
```diff
@@ -77,20 +82,20 @@ def find_scheme
...
                 Scheme.find(params[:id])
               else
-                @location.scheme
+                @location&.scheme
               end
   end
 
   def find_location
-    @location = params[:location_id].present? ? Location.find(params[:location_id]) : Location.find(params[:id])
+    @location = params[:location_id].present? ? Location.find_by(id: params[:location_id]) : Location.find_by(id: params[:id])
```
smeny, line 90: what is the difference here? I thought find was guaranteed to return a single result based on the ID.

smeny (approved):
Just left a question, but good to go regardless.

app/controllers/locations_controller.rb
baarkerlounger: `find_by` returns nil when no results. `find` throws an error.

PR #812, opened 2022-08-04
Back button on edit location page routes to check answers for locations

kosiakkatrina:
It was previously routing to support, was likely accidentally copied from /locations/new page

### DIFF
--- a/app/views/locations/edit.html.erb
+++ b/app/views/locations/edit.html.erb
@@ -3,7 +3,7 @@
 <% content_for :before_content do %>
   <%= govuk_back_link(
     text: "Back",
-    href: scheme_check_answers_path(@scheme.id, anchor: "locations"),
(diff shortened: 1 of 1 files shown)


jamesrose (approved):

app/views/locations/edit.html.erb
```diff
@@ -3,7 +3,7 @@
 <% content_for :before_content do %>
   <%= govuk_back_link(
     text: "Back",
-    href: "/schemes/#{@scheme.id}/support",
+    href: scheme_check_answers_path(@scheme.id, anchor: "locations"),
```
jamesrose: You don't need to be explicit about the ID here:

```suggestion
    href: scheme_check_answers_path(@scheme, anchor: "locations"),
```

kosiakkatrina: of course 🤦‍♀️  Thank you!

PR #813, opened 2022-08-04
Cldc 1424 display postcode question

kosiakkatrina:
Reset wchair when needstype changes from supported housing to not supported housing - this was causing the property information into `in progress` state, which meant that it went straight to check your answers page.
Check your answers page for general needs now displays the `postcode_known` question in case it was not answered
image

We also reset location for the case log to nil if case log needstype changes to general needs. It was not being reset as part of the `reset_not_routed_questions` because it is marked as a special case - if there is only 1 location available we don't route to the locations question, instead we infer that location as the answer and we don't want it to be cleared

### DIFF
--- a/app/models/case_log.rb
+++ b/app/models/case_log.rb
@@ -536,7 +536,8 @@ private
     dependent_questions = { waityear: [{ key: :renewal, value: 0 }],
                             homeless: [{ key: :renewal, value: 0 }],
                             referral: [{ key: :renewal, value: 0 }],
-                            underoccupation_benefitcap: [{ key: :renewal, value: 0 }] }
+                            underoccupation_benefitcap: [{ key: :renewal, value: 0 }],
+                            wchair: [{ key: :needstype, value: 1 }] }
 
     dependent_questions.each do |dependent, conditions|
       condition_key = conditions.first[:key]
--- a/app/models/case_log.rb
+++ b/app/models/case_log.rb
@@ -537,7 +537,8 @@ private
                             homeless: [{ key: :renewal, value: 0 }],
                             referral: [{ key: :renewal, value: 0 }],
                             underoccupation_benefitcap: [{ key: :renewal, value: 0 }],
-                            wchair: [{ key: :needstype, value: 1 }] }
+                            wchair: [{ key: :needstype, value: 1 }],
+                            location_id: [{ key: :needstype, value: 1 }] }
 
     dependent_questions.each do |dependent, conditions|
       condition_key = conditions.first[:key]
--- a/config/forms/2021_2022.json
+++ b/config/forms/2021_2022.json
@@ -34,7 +34,14 @@
                   "conditional_for": {
                     "postcode_full": [1]
                   },
-                  "hidden_in_check_answers": true
+                  "hidden_in_check_answers": {
+                    "depends_on": [{
+                      "postcode_known": 0
+                    },
+                    {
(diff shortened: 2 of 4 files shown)


config/forms/2021_2022.json
```diff
@@ -1122,7 +1129,11 @@
               }
             },
             "females_in_soft_age_range_in_pregnant_household_lead_hhmemb_value_check": {
-              "depends_on": [{ "female_in_pregnant_household_in_soft_validation_range?": true }],
+              "depends_on": [
+                {
+                  "female_in_pregnant_household_in_soft_validation_range?": true
+                }
+              ],
```
kosiakkatrina, line 1136: formatting only

config/forms/2021_2022.json
```diff
@@ -1248,7 +1259,11 @@
               }
             },
             "females_in_soft_age_range_in_pregnant_household_lead_age_value_check": {
-              "depends_on": [{ "female_in_pregnant_household_in_soft_validation_range?": true }],
+              "depends_on": [
+                {
+                  "female_in_pregnant_household_in_soft_validation_range?": true
+                }
+              ],
```
kosiakkatrina, line 1266: formatting only

config/forms/2021_2022.json
```diff
@@ -1356,7 +1371,11 @@
               }
             },
             "females_in_soft_age_range_in_pregnant_household_lead_value_check": {
-              "depends_on": [{ "female_in_pregnant_household_in_soft_validation_range?": true }],
+              "depends_on": [
+                {
+                  "female_in_pregnant_household_in_soft_validation_range?": true
+                }
+              ],
```
kosiakkatrina, line 1378: formatting only

app/models/case_log.rb
```diff
@@ -536,7 +536,9 @@ def reset_derived_questions
     dependent_questions = { waityear: [{ key: :renewal, value: 0 }],
```
baarkerlounger, line 536: Do you remember why we needed to hardcode these dependencies here rather than being able to derive them from the question definition?

baarkerlounger (approved):

app/models/case_log.rb
kosiakkatrina: not exactly, the first thing that comes to mind is that maybe validations were failing for some reason if we didn't reset these. Do you?

kosiakkatrina: https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/322 bad PR descriptions come back to haunt me

baarkerlounger: Looks like it's because it's a slightly special case. Usually in the definition when we says a question `depends_on` something, when it's not met it's not routed to *and* any answer is cleared.

In these cases, it *is* routed to but we still want an inferred answer cleared.

Source. https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/322


This probably warrants becoming it's own more explicit thing but for the time being maybe a comment in the code here is better than nothing?

kosiakkatrina: `So we're already getting the depends on to determine if something is not routed to and if it's not we clear it.
In this case we want to clear something that is routed to, but only if it's one of the fields we're inferring. We can get the questions that are routed to from this JSON, but we would still need to filter them down to the ones that get inferred, no?`
good thing we got comments

PR #814, opened 2022-08-05
Only consider active locations for routing

baarkerlounger:


kosiakkatrina (approved):
Nice! Was wondering if we need a validation on the CYA locations page as well, the way we have for schemes, but I think currently we can't create a location that is invalid cause it's only 1 form page, so should be okay for now

PR #816, opened 2022-08-05
Strip whitespaces from names before saving organisation, scheme or a user

kosiakkatrina:
Saving it with a space was erroring out when trying to select them from an accessible select. 
It also makes sense not to store the leading/trailing spaces in the names

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -55,6 +55,7 @@ gem "redis"
 gem "sentry-rails"
 gem "sentry-ruby"
 # Possessives in strings
+gem "auto_strip_attributes"
 gem "possessive"
 
 group :development, :test do
--- a/Gemfile
+++ b/Gemfile
@@ -5,6 +5,7 @@ git_source(:github) { |repo| "https://github.com/#{repo}.git" }
 
 ruby "3.1.2"
 
+gem "auto_strip_attributes"
 # Bundle edge Rails instead: gem 'rails', github: 'rails/rails', branch: 'main'
 gem "rails", "~> 7.0.2"
 # Use postgresql as the database for Active Record
@@ -55,7 +56,6 @@ gem "redis"
 gem "sentry-rails"
 gem "sentry-ruby"
 # Possessives in strings
-gem "auto_strip_attributes"
 gem "possessive"
 
 group :development, :test do
--- a/app/models/application_record.rb
+++ b/app/models/application_record.rb
@@ -2,7 +2,6 @@ class ApplicationRecord < ActiveRecord::Base
   self.abstract_class = true
 
   def strip_whitespaces
-    self.service_name = service_name.strip unless !respond_to?("service_name") || service_name.nil?
-    self.name = name.strip unless !respond_to?("name") || name.nil?
+    fields_to_strip.each { |field| self[field] = self[field].strip if field.present? }
   end
 end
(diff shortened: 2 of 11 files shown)


baarkerlounger:
We probably want this on location name as well?

app/models/organisation.rb
```diff
@@ -74,4 +76,8 @@ def display_attributes
       { name: "data_protection_agreement", value: data_protection_agreement_string, editable: false },
     ]
   end
+
+  def strip_whitespace
+    self.name = name.strip unless name.nil?
```
baarkerlounger: https://api.rubyonrails.org/classes/String.html#method-i-squish might be better than strip?

app/models/scheme.rb
```diff
@@ -228,4 +230,8 @@ def validate_confirmed
       end
     end
   end
+
+  def strip_whitespace
+    self.service_name = service_name.strip unless service_name.nil?
```
baarkerlounger: I wonder if this is worth genericising to all fields that respond to strip/squish and extracting to `ApplicationRecord`?

app/models/scheme.rb
```diff
@@ -19,6 +19,8 @@ class Scheme < ApplicationRecord
 
   validate :validate_confirmed
 
+  before_validation :strip_whitespace
```
baarkerlounger: Could just override `field=` methods... (has the advantage of setting it in memory the same as you'll save it already before actually doing the save, but might be unnecessary indirection...)

app/models/organisation.rb
kosiakkatrina: Hmm, could there ever be a reason to have a double space in the name?

app/models/scheme.rb
kosiakkatrina: Yeah, possibly, will have a look

app/models/organisation.rb
baarkerlounger: Not that I know of, but I could be wrong.

kosiakkatrina: If we're generalising this, I think I'd prefer just to strip it for now, cause it feels less destructive

app/models/scheme.rb
kosiakkatrina: How strongly do you feel about it?

app/models/application_record.rb
```diff
@@ -1,3 +1,8 @@
 class ApplicationRecord < ActiveRecord::Base
   self.abstract_class = true
+
+  def strip_whitespaces
+    self.service_name = service_name.strip unless !respond_to?("service_name") || service_name.nil?
```
baarkerlounger: Do you need the nil checks here? Wouldn't that be caught by the respond_to already?

baarkerlounger: You could maybe also have this take (*args) that you iterate over and the models could pass the fields that need stripping?

kosiakkatrina: can't pass in as a param from the before_validation, but added a `fields_to_strip` method that returns those fields for each class

Gemfile
```diff
@@ -55,6 +55,7 @@ gem "redis"
 gem "sentry-rails"
 gem "sentry-ruby"
 # Possessives in strings
+gem "auto_strip_attributes"
```
baarkerlounger: Need to fix the comments here

baarkerlounger (approved):

PR #817, opened 2022-08-05
CLDC-1171: User validation order

baarkerlounger:
Fixes the last two comments on: https://digital.dclg.gov.uk/jira/browse/CLDC-1171

This ordering comes from including `Devise::Validatable` which includes a few default validations (email and password) which run before our name validations. It's an optional module that's expected to be removed if you need to customise it (https://github.com/heartcombo/devise/blob/main/lib/devise/models/validatable.rb).

So we remove it here and add our own instead in the correct order.

We also make role optional as we have some real world users in production already who do not currently have a role, as they are data protection officers only (which can be combined with another role for some users), and do not interact with the system in any other way.

Long term it might make sense to refactor roles into their own model rather than making them an attribute of users, such that a user can have multiple roles but that seems beyond the scope here.

Support user view:
image

Coordinator view:
image

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -58,7 +58,7 @@ class UsersController < ApplicationController
       end
     elsif user_params.key?("password")
       format_error_messages
-      @minimum_password_length = User.password_length.min
+      @minimum_password_length = Devise.password_length.min
       render "devise/passwords/edit", locals: { resource: @user, resource_name: "user" }, status: :unprocessable_entity
     else
       format_error_messages
@@ -79,7 +79,6 @@ class UsersController < ApplicationController
       redirect_to created_user_redirect_path
     else
       unless @resource.errors[:organisation].empty?
-        @resource.errors.add(:organisation_id, message: @resource.errors[:organisation])
         @resource.errors.delete(:organisation)
       end
       render :new, status: :unprocessable_entity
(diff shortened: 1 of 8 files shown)


app/controllers/users_controller.rb
```diff
@@ -79,15 +79,14 @@ def create
       redirect_to created_user_redirect_path
     else
       unless @resource.errors[:organisation].empty?
-        @resource.errors.add(:organisation_id, message: @resource.errors[:organisation])
         @resource.errors.delete(:organisation)
```
smeny, line 82: Still not sure why we need to do this, I know it has to do with `organisation` VS `organisation_id`

app/views/users/new.html.erb
```diff
@@ -52,7 +52,8 @@
                                            lambda { |option|
                                              option.description&.map { |hint| content_tag(:li, hint) }&.reduce(:+)
                                            },
-                                           legend: { text: "Role", size: "m" } %>
+                                           legend: { text: "Role (optional)", size: "m" },
+                                           hint: { text: "This does not need to be selected if the user is a data protection officer only" } %>
```
smeny: Can't we infer the DPO role here and hide the question?

app/controllers/users_controller.rb
baarkerlounger: It's so that the form builder gem links the error to the correct form field. i.e. because the form field references `organisation_id` the error needs to be on `organisation_id` rather than on `organisation`.

This PR makes it so that the model now does that anyway, but it also produces a duplicate(ish) error on `organisation` that we don't want/need to display on the view so we remove it here.

smeny (approved):
Just a few questions but seems OK to go otherwise.

app/views/users/new.html.erb
baarkerlounger: How would we do that given they're inviting a new user?

PR #820, opened 2022-08-08
Cldc 1338 dates validation

kosiakkatrina:
Change the hard validations to be under 2 years between the tenancy start date and major repairs/void date
Add soft validations when the major repairs/void date is within 2 to 10 years of the tenancy start date
image

Update hidden in check answers for interruption questions to show the question only if it's not been answered:
image


app/models/validations/soft_validations.rb
```diff
@@ -62,6 +62,14 @@ def all_tenants_age_and_gender_information_completed?
     end
   end
 
+  def major_repairs_date_in_soft_range?
+    mrcdate.present? && startdate.present? && mrcdate.between?(startdate.to_date - 3650, startdate.to_date - 730)
```
baarkerlounger: The numbers could probably be named constants?

baarkerlounger (approved):

PR #821, opened 2022-08-08
CLDC-1219: Single import task

smeny:


baarkerlounger (approved):
LGTM. I wonder whether any further checking is required that the individual import tasks all do something sensible if they're missing expected data?

PR #824, opened 2022-08-09
Cldc 1235 conditional shortfall

natdeanlewissoftwire:
Tested locally, all functionality present
Minimal risk
No documentation required

ticket: https://digital.dclg.gov.uk/jira/browse/CLDC-1235

check-answers page: 
image

question page:
image

### DIFF
--- a/config/forms/2021_2022.json
+++ b/config/forms/2021_2022.json
@@ -7876,28 +7876,18 @@
                 {
                   "hb": 6,
                   "household_charge": null
-                },
-                {
-                  "hb": 8,
-                  "household_charge": 0
-                },
-                {
-                  "hb": 8,
-                  "household_charge": null
                 }
               ]
             },
-            "outstanding_amount_known": {
+            "outstanding_amount": {
               "header": "",
               "description": "",
               "questions": {
                 "tshortfall_known": {
-                  "check_answer_label": "",
-                  "header": "",
-                  "hint_text": "",
-                  "hidden_in_check_answers": true,
+                  "check_answer_label": "Do you know the outstanding amount?",
+                  "header": "Can you estimate the outstanding amount?",
+                  "hint_text": "You only need to give an approximate figure.",
                   "type": "radio",
-                  "derived": true,
                   "answer_options": {
                     "0": {
                       "value": "Yes"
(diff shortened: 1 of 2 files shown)


kosiakkatrina (commented):
Do we also need to update 21/22 form? Where we do use the real form in our tests it's 21/22, although the collection window is closed for that one, maybe worth updating it just to see if there is any impact on the tests

baarkerlounger:
@kosiakkatrina I don't think we do since the form UI itself shouldn't be used anymore for that window and the fields are staying the same here so no changes for import or export.

(But there'd be no harm in changing there also).

baarkerlounger (approved):

baarkerlounger:
Actually sorry the window isn't technically closed yet (see comment at the bottom of https://digital.dclg.gov.uk/jira/browse/CLDC-1428) so we probably should do this for 21/22 form also.

kosiakkatrina:
> @kosiakkatrina I don't think we do since the form UI itself shouldn't be used anymore for that window and the fields are staying the same here so no changes for import or export.
> 
> (But there'd be no harm in changing there also).

@baarkerlounger I think we do use the actual form for some tests, for example to check if a case log is completed, specific fields are reset based on routing etc. But perhaps no impact from this change, if that's what you mean

natdeanlewissoftwire:
Thanks @kosiakkatrina and @baarkerlounger, have copied the changes over to 2021_2022.json as well

config/forms/2021_2022.json
```diff
@@ -7971,14 +7956,6 @@
                 {
                   "hb": 6,
                   "hbrentshortfall": 1
-                },
-                {
```
baarkerlounger: Why are these options being removed?

config/forms/2021_2022.json
```diff
@@ -7876,48 +7876,33 @@
                 {
                   "hb": 6,
                   "household_charge": null
-                },
-                {
```
baarkerlounger: Same question here

config/forms/2021_2022.json
natdeanlewissoftwire: Just fixing this now, please ignore

baarkerlounger (approved):

PR #826, opened 2022-08-10
Change accessibility requirements question

kosiakkatrina:
Change accessibility requirements question into 3 separate questions that are easier to follow.
Before:
image

After:
image
image

Check your answers page:
image
image

Remove validations as conditional routing and resetting the values prevents invalid inputs.
Will likely need to run db query or migration to backfill the new housingneeds_type and housingneeds_other fields


app/models/case_log.rb
```diff
@@ -25,6 +25,7 @@ class CaseLog < ApplicationRecord
   before_validation :reset_scheme_location!, if: :scheme_changed?, unless: :location_changed?
   before_validation :process_postcode_changes!, if: :postcode_full_changed?
   before_validation :process_previous_postcode_changes!, if: :ppostcode_full_changed?
+  before_validation :set_housingneeds_fields, if: :housingneeds_present?
```
baarkerlounger: Is this really a special case thing or are these just derived values that could be in the derived variables?

app/models/case_log.rb
```diff
@@ -723,4 +700,57 @@ def soft_value_for_period(value)
...
+  def no_housingneeds?
+    housingneeds == 2
+  end
+
+  def unknown_housingneeds?
+    housingneeds == 3
+  end
+
+  def set_housingneeds_fields
+    set_housingneeds_values_to_nil
```
baarkerlounger: Name seems a bit misleading if we're setting to 0 rather than nil?

app/models/case_log.rb
```diff
@@ -723,4 +700,57 @@ def soft_value_for_period(value)
...
+
+    self.housingneeds_a = 1 if fully_wheelchair_accessible?
+    self.housingneeds_b = 1 if essential_wheelchair_access?
+    self.housingneeds_c = 1 if level_access_housing?
+    self.housingneeds_f = 1 if other_housingneeds?
+    set_housingneeds_values_to_nil unless has_housingneeds?
```
baarkerlounger: Seems a bit odd to call this twice

app/models/case_log.rb
```diff
@@ -723,4 +700,57 @@ def soft_value_for_period(value)
   def upcase_and_remove_whitespace(string)
     string.present? ? string.upcase.gsub(/\s+/, "") : string
   end
+
+  def housingneeds_present?
+    housingneeds.present?
+  end
```
jamesrose: Do you need a method at all here? Can you get away with just using `if :housingneeds?`?

kosiakkatrina: Ah, I think I added it because the initial values for housingneeds contained 0 as an option, but it doesn't anymore. Good spot!

app/models/case_log.rb
kosiakkatrina: It was failing a test but it seems it was due to a set up of it

app/models/case_log.rb
kosiakkatrina: Changed that a bit

app/models/case_log.rb
kosiakkatrina: Changed that a bit

config/forms/2021_2022.json
```diff
@@ -5396,40 +5396,79 @@
...
+                  }
+                }
+              }
+            },
+            "type_of_access_needs": {
+              "header": "Disabled access needs",
+              "description": "",
+              "questions": {
+                "housingneeds_type": {
+                  "header": "What type of access need do they have?",
```
jamesrose, line 5430:
```suggestion
                  "header": "What types of access need do they have?",
```

I think this should be plural.

baarkerlounger:
Code looks good, presumably this will require a data update for existing logs?

kosiakkatrina:
> Code looks good, presumably this will require a data update for existing logs?

@baarkerlounger yup, this is what I've got for that:

```
CaseLog.where(housingneeds_a: 1).update(housingneeds_type: 0)
CaseLog.where(housingneeds_b: 1).update(housingneeds_type: 1)
CaseLog.where(housingneeds_c: 1).update(housingneeds_type: 2)
CaseLog.where(housingneeds_f: 1).update(housingneeds_other: 1)
```

app/models/derived_variables/case_log_variables.rb
```diff
@@ -186,4 +187,23 @@ def reset_scheme_location!
       self.location = scheme.locations.first
     end
   end
+
+  def set_housingneeds_fields
+    self.housingneeds_a = fully_wheelchair_accessible? ? 1 : 0
+    self.housingneeds_b = essential_wheelchair_access? ? 1 : 0
+    self.housingneeds_c = level_access_housing? ? 1 : 0
+    self.housingneeds_f = other_housingneeds? ? 1 : 0
+    set_housingneeds_values_to_zero unless has_housingneeds?
```
baarkerlounger, line 196: It feels brittle that this is order dependent but I suppose that's already the case for a whole bunch of derivation

baarkerlounger: Think it might work better inline, as a single method that sets housing needs fields but just personal preference so don't feel the need to change

baarkerlounger (approved):

paulrobertlloyd:
🎉

PR #828, opened 2022-08-11
CLDC-1433 Add parent child relationship model

natdeanlewissoftwire:
Tested locally
Risk of another migration from same base at the same time - should be checked before merge
No docs required
Ticket: https://digital.dclg.gov.uk/jira/browse/CLDC-1443

### DIFF
--- a/app/models/organisation.rb
+++ b/app/models/organisation.rb
@@ -6,6 +6,10 @@ class Organisation < ApplicationRecord
   has_many :organisation_rent_periods
   has_many :owned_schemes, class_name: "Scheme", foreign_key: "owning_organisation_id", dependent: :delete_all
   has_many :managed_schemes, class_name: "Scheme", foreign_key: "managing_organisation_id"
+  has_many :parent_organisation_relationships, foreign_key: :child_organisation_id, class_name: "OrganisationRelationship"
+  has_many :parent_organisations, through: :parent_organisation_relationships
+  has_many :child_organisation_relationships, foreign_key: :parent_organisation_id, class_name: "OrganisationRelationship"
+  has_many :child_organisations, through: :child_organisation_relationships
 
   scope :search_by_name, ->(name) { where("name ILIKE ?", "%#{name}%") }
   scope :search_by, ->(param) { search_by_name(param) }
--- a/db/migrate/20220810152340_add_parent_child_relationships.rb
+++ b/db/migrate/20220810152340_add_parent_child_relationships.rb
@@ -1,0 +1,9 @@
+class AddParentChildRelationships < ActiveRecord::Migration[7.0]
+  def change
+    create_table :organisation_relationships do |t|
+      t.integer :child_organisation_id, foreign_key: true
+      t.integer :parent_organisation_id, foreign_key: true
+      t.timestamps
+    end
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema[7.0].define(version: 2022_08_02_125711) do
+ActiveRecord::Schema[7.0].define(version: 2022_08_10_152340) do
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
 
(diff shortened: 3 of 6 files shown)


db/migrate/20220810152340_add_parent_child_relationships.rb
```diff
@@ -0,0 +1,9 @@
+class AddParentChildRelationships < ActiveRecord::Migration[7.0]
+  def change
+    create_table :organisation_relationships do |t|
+      t.integer :child_organisation_id, foreign_key: true
+      t.integer :parent_organisation_id, foreign_key: true
+      t.timestamps
```
natdeanlewissoftwire, line 6: Apparently this is good practice, but current tables in the database seem inconsistent as to whether they have timestamp data - do we want this?

db/schema.rb
```diff
@@ -383,6 +390,7 @@
   add_foreign_key "case_logs", "organisations", column: "owning_organisation_id", on_delete: :cascade
   add_foreign_key "case_logs", "schemes"
   add_foreign_key "locations", "schemes"
+  add_foreign_key "schemes", "organisations", column: "managing_organisation_id"
```
natdeanlewissoftwire: Already merged in, not sure why it's showing up here as this is now conistent with main

app/models/organisation.rb
```diff
@@ -6,6 +6,10 @@ class Organisation < ApplicationRecord
   has_many :organisation_rent_periods
   has_many :owned_schemes, class_name: "Scheme", foreign_key: "owning_organisation_id", dependent: :delete_all
   has_many :managed_schemes, class_name: "Scheme", foreign_key: "managing_organisation_id"
+  has_many :parent_organisation_relationships, foreign_key: :child_organisation_id, class_name: "OrganisationRelationship"
```
jamesrose, line 9: @natdeanlewissoftwire Could you add some model tests for this? I think at a minimum we would like to test whether an organisation can have parents and children.

db/migrate/20220810152340_add_parent_child_relationships.rb
jamesrose: I generally keep timestamps on these kinds of join tables. It adds very little overhead, but it's useful to debug when a relationship was created.

db/schema.rb
jamesrose: Do you just need to rebase your branch?

db/migrate/20220810152340_add_parent_child_relationships.rb
baarkerlounger: I think all existing tables have timestamps? Which ones don't?

db/schema.rb
natdeanlewissoftwire: Of course, thanks :)

db/migrate/20220810152340_add_parent_child_relationships.rb
natdeanlewissoftwire: @baarkerlounger Ah you're right, they all have created_at at least - it's the couple that don't have updated_at columns I was thinking of (logs_exports and versions), never mind!

baarkerlounger: `versions` isn't really ours, it comes from https://github.com/paper-trail-gem/paper_trail

spec/factories/organisation.rb
```diff
@@ -11,17 +11,15 @@
     holds_own_stock { true }
   end
 
-  factory :organisation_la do
-    organisation
-    ons_code { "E07000041" }
-    created_at { Time.zone.now }
-    updated_at { Time.zone.now }
-  end
-
```
natdeanlewissoftwire, line 20: this is the (I _think_) unused factory I was referring to @jamesrose

spec/factories/organisation.rb
```diff
@@ -11,17 +11,15 @@
...
     organisation
     rent_period { 2 }
     created_at { Time.zone.now }
     updated_at { Time.zone.now }
   end
+
+  factory :organisation_relationship do
+    child_organisation { FactoryBot.create(:organisation) }
+    parent_organisation { FactoryBot.create(:organisation) }
+  end
```
natdeanlewissoftwire, line 24: I put this back in the organisation file, I think that's simpler than having a separate organisation_relationship file especially as organisation_rent_period is here too

spec/models/organisation_spec.rb
```diff
@@ -27,6 +27,22 @@
         .to raise_error(ActiveRecord::RecordInvalid, "Validation failed: Provider type #{I18n.t('validations.organisation.provider_type_missing')}")
     end
 
+    context "with parent/child association" do
+      let(:child_organisation) { FactoryBot.create(:organisation, name: "DLUHC_Child") }
```
jamesrose:
```suggestion
      let(:child_organisation) { FactoryBot.create(:organisation, name: "DLUHC Child") }
```

nit: underscored name

jamesrose (approved):

PR #829, opened 2022-08-11
CLDC-1221: Support reading import from zip archive

smeny:
- Changed the fixtures name to match import folders
- Reused the storage service interface to allow for an easy swap between s3 and archive
- Rename StorageService into S3StorageService


baarkerlounger (approved):
Think the pattern works quite nicely here

PR #833, opened 2022-08-12
CLDC-1310-seperate-tenants-fix

Ted-U:
Before:
Screenshot 2022-08-17 at 10 01 15

After:
Screenshot 2022-08-17 at 10 00 13

### DIFF
--- a/app/components/check_answers_summary_list_card_component.html.erb
+++ b/app/components/check_answers_summary_list_card_component.html.erb
@@ -1,0 +1,25 @@
+<div class="check-answers-summary-list">
+<%= govuk_summary_list do |summary_list| %>
+  <% applicable_questions.each do |question| %>
+    <% summary_list.row do |row| %>
+      <% row.key { question.check_answer_label.to_s.presence || question.header.to_s } %>
+      <% row.value do %>
+        <span class="govuk-!-margin-right-4"><%= get_answer_label(question) %></span>
+        <% extra_value = question.get_extra_check_answer_value(case_log) %>
+        <% if extra_value %>
+          <span class="govuk-!-font-weight-regular app-!-colour-muted"><%= extra_value %></span>
+        <% end %>
+        <br>
+        <% question.get_inferred_answers(case_log).each do |inferred_answer| %>
+          <span class="govuk-!-font-weight-regular app-!-colour-muted"><%= inferred_answer %></span>
+        <% end %>
+      <% end %>
+      <% row.action(
+        text: question.action_text(case_log),
+        href: question.action_href(case_log, question.page.id),
+        visually_hidden_text: question.check_answer_label.to_s.downcase,
+      ) %>
+    <% end %>
+  <% end %>
+<% end %>
+</div>
--- a/app/components/check_answers_summary_list_card_component.html.erb
+++ b/app/components/check_answers_summary_list_card_component.html.erb
@@ -1,25 +1,37 @@
-<div class="check-answers-summary-list">
-<%= govuk_summary_list do |summary_list| %>
-  <% applicable_questions.each do |question| %>
-    <% summary_list.row do |row| %>
-      <% row.key { question.check_answer_label.to_s.presence || question.header.to_s } %>
-      <% row.value do %>
-        <span class="govuk-!-margin-right-4"><%= get_answer_label(question) %></span>
-        <% extra_value = question.get_extra_check_answer_value(case_log) %>
-        <% if extra_value %>
-          <span class="govuk-!-font-weight-regular app-!-colour-muted"><%= extra_value %></span>
-        <% end %>
-        <br>
-        <% question.get_inferred_answers(case_log).each do |inferred_answer| %>
-          <span class="govuk-!-font-weight-regular app-!-colour-muted"><%= inferred_answer %></span>
+<div class="x-govuk-summary-card govuk-!-margin-bottom-6">
+  <% if applicable_questions.first.check_answers_card_number != 0 %>
+    <div class="x-govuk-summary-card__header">
+      <% if applicable_questions.first.check_answers_card_number == 1 %>
+        <h3 class="x-govuk-summary-card__title">Lead tenant</h3>
+      <% end %>
+      <% if applicable_questions.first.check_answers_card_number > 1 %>
+        <h3 class="x-govuk-summary-card__title">Person <%= applicable_questions.first.check_answers_card_number %></h3>
+      <% end %>
+    </div>
+  <% end %>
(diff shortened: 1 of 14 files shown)


app/views/form/check_answers.html.erb
```diff
@@ -17,10 +17,16 @@
     <% end %>
     <%= display_answered_questions_summary(subsection, @case_log, current_user) %>
 
-    <%= render partial: "form/check_answers_summary_list", locals: {
-      subsection:,
-      case_log: @case_log,
-    } %>
+    <% if subsection.id == "household_characteristics" %>
```
baarkerlounger: We should probably avoid hardcoding specific section names here, will make the code brittle. Could we instead add the key everywhere or infer based on it's absence such that we can do this generically?

spec/components/check_answers_summary_list_card_component_spec.rb
```diff
@@ -0,0 +1,16 @@
...
+
+RSpec.describe CheckAnswersSummaryListCardComponent, type: :component do
+  context "when given a set of questions" do
+    let(:user) { FactoryBot.build(:user) }
+    let(:case_log) { FactoryBot.build(:case_log, :completed) }
+    let(:subsection_id) { "household_characteristics" }
+    let(:subsection) { case_log.form.get_subsection(subsection_id) }
+    let(:questions) { subsection.applicable_questions(case_log) }
+
+    it "renders a summary list card for the answers to those questions" do
```
baarkerlounger, line 11: There should probably be a few more tests here for scenarios like:

- The questions passed in include some that are hidden in check answers
- Some questions are answered and some aren't (do we show the right labels and links?)

spec/features/form/check_answers_page_spec.rb
```diff
@@ -133,6 +133,19 @@
       end
     end
 
+    it "has no summary cards if the subsection is not household characteristics" do
```
baarkerlounger: This should probably be phrased in terms of attributes rather than a specific section name

app/views/form/check_answers.html.erb
dushan-madetech: I'd considered that but this is the only subsection that uses the summary list design for the check answers page (so I scrapped trying to generalise it since we probably wouldn't need it for any others) and we're doing the same kind of check for the setup section just a few lines above. Happy to try and generalise if you think it would be needed 😄

spec/features/form/check_answers_page_spec.rb
dushan-madetech: Do you have a suggestion for how it might look phrased in terms of attributes? I did it this way since this design only applies to the household characteristics section 🙂

spec/components/check_answers_summary_list_card_component_spec.rb
dushan-madetech: I feel like I might be able write both of those assertions in one test using the details known question 🤔

app/views/form/check_answers.html.erb
baarkerlounger: My thinking is that "Setup" is a legit special case that behaves differently from everything else whereas this just incidently is. i.e. I think you could reasonably expect to add this functionality to any arbitrary set of questions in any section, so I probably would genericise it, but I agree we don't strictly need to.

spec/features/form/check_answers_page_spec.rb
baarkerlounger: I guess the number of cards shown should match the number the definition for the section expects

dushan-madetech: The section definition doesn't define the number of cards shown, it's defined at the question level using `check_answers_card_number`. When you get the applicable questions for a subsection you group them into cards using group by. I wasn't quite sure how to express this test in any other way, maybe something like `"does not group questions into summary cards if the questions in the subsection don't have a check_answers_card_number attribute"`?

app/views/form/check_answers.html.erb
dushan-madetech: Would it be following the principle of "You ain't gonna need it" (YAGNI) if we were to generalise this only if a need for it actually arises? Asking because I'm not 100% sure 😄

baarkerlounger: 🤷‍♂️ maybe but I think there's value in doing it because of the intent it signals. Those questions don't display as multiple cards _because_ they're in the household section (after all, we could move them and would still expect them to display as multiple cards), but rather because they have a check_answers_card_number property.

I.e. the reason to avoid hardcoding the section name here is so that the comparison is made to the right thing, the thing that really determines this rather than something that is just incidentally true at the moment.

app/views/form/check_answers.html.erb
```diff
@@ -17,10 +17,16 @@
     <% end %>
     <%= display_answered_questions_summary(subsection, @case_log, current_user) %>
 
-    <%= render partial: "form/check_answers_summary_list", locals: {
-      subsection:,
-      case_log: @case_log,
-    } %>
+    <% if subsection.applicable_questions(@case_log).map(&:check_answers_card_number).compact.length > 0 %>
```
baarkerlounger: Might be worth extracting a helper method here?

baarkerlounger (approved):

app/views/form/check_answers.html.erb
dushan-madetech: Agreed, will do that now 🙂

spec/components/check_answers_summary_list_card_component_spec.rb
dushan-madetech: I've tried to write these suggested tests as unit tests for the component 🙂

app/components/check_answers_summary_list_card_component.html.erb
```diff
@@ -0,0 +1,37 @@
+<div class="x-govuk-summary-card govuk-!-margin-bottom-6">
+  <% if applicable_questions.first.check_answers_card_number != 0 %>
+    <div class="x-govuk-summary-card__header">
+      <% if applicable_questions.first.check_answers_card_number == 1 %>
+        <h3 class="x-govuk-summary-card__title">Lead tenant</h3>
```
baarkerlounger, line 5: This should be coming from the page/question definition shouldn't it? At present this isn't really a re-usable component.

spec/components/check_answers_summary_list_card_component_spec.rb
baarkerlounger: I think here that maybe leads to testing the implementation rather than the outcome given that those are effectively "private" methods that aren't expected to be used anywhere outside this component I think that might be making the idea of the "unit" too small.

PR #836, opened 2022-08-18
Cldc 1432 csv labels and codes

kosiakkatrina:
Remove some derived fields and fields that are specific to our system from csv download for non support users.
Not sure if the list for the fields not to display is full as it wasn't very clear which ones to remove. 
The list of words to omit lives in a constant as there is no obvious way to get them automatically, so they should be easy to update

Get boolean values to display as Yes/No in the csv download

Display names form owning organisation, managing organisation and created by instead of IDs


app/models/case_log.rb
```diff
@@ -74,6 +74,7 @@ class CaseLog < ApplicationRecord
   NUM_OF_WEEKS_FROM_PERIOD = { 2 => 26, 3 => 13, 4 => 12, 5 => 50, 6 => 49, 7 => 48, 8 => 47, 9 => 46, 1 => 52 }.freeze
   SUFFIX_FROM_PERIOD = { 2 => "every 2 weeks", 3 => "every 4 weeks", 4 => "every month" }.freeze
   RETIREMENT_AGES = { "M" => 67, "F" => 60, "X" => 67 }.freeze
+  CSV_FIELDS_TO_OMIT = %w[hhmemb net_income_value_check sale_or_letting first_time_property_let_as_social_housing renttype needstype postcode_known is_la_inferred totchild totelder totadult net_income_known is_carehome previous_la_known is_previous_la_inferred age1_known age2_known age3_known age4_known age5_known age6_known age7_known age8_known letting_allocation_unknown details_known_2 details_known_3 details_known_4 details_known_5 details_known_6 details_known_7 details_known_8 rent_type wrent wscharge wpschrge wsupchrg wtcharge wtshortfall rent_value_check old_form_id old_id retirement_value_check tshortfall_known pregnancy_value_check hhtype new_old vacdays].freeze
```
baarkerlounger, line 77: This might be worth multi-lining

kosiakkatrina: It's either this or one per line (sorry rubocop says)

dushan-madetech (approved):
Looks good to me 🙂

PR #837, opened 2022-08-22
Cldc 1433 csv output symbols correctly

natdeanlewissoftwire:
Tested locally
Minimal risk
No docs required - variable name makes this clear.


app/controllers/case_logs_controller.rb
```diff
@@ -13,6 +13,7 @@ def index
 
     all_logs = current_user.case_logs
     unpaginated_filtered_logs = filtered_case_logs(filtered_collection(all_logs, search_term))
+    byte_order_mark = "\uFEFF"
```
baarkerlounger: Wonder if this is worth DRY-ing? I assume we'd want it anywhere we download a csv?

spec/requests/case_logs_controller_spec.rb
```diff
@@ -755,7 +755,7 @@
 
       it "downloads a CSV file with headers" do
         csv = CSV.parse(response.body)
-        expect(csv.first.first).to eq("id")
+        expect(csv.first.first).to eq("\uFEFFid")
```
baarkerlounger, line 758: Might be worth still leaving in:

```ruby
        expect(csv.first.second).to eq("id")
```

baarkerlounger (approved):

spec/requests/case_logs_controller_spec.rb
natdeanlewissoftwire: The second header is "status", the first header is "\uFEFF" + "id" - I don't think it's worth testing the second is it?

baarkerlounger: Ah gotcha, missed that

app/controllers/case_logs_controller.rb
natdeanlewissoftwire: Good point, I've put byte_order_mark into application_controller instead :)

PR #838, opened 2022-08-22
Cldc 1361 only stock owning orgs see schemes

Ted-U:


spec/factories/organisation.rb
```diff
@@ -11,6 +11,13 @@
     holds_own_stock { true }
   end
 
+  factory :organisation_la do
```
baarkerlounger: Why are we adding this?

Ted-U: This came out of a merge conflict, I will remove it

baarkerlounger (approved):

PR #839, opened 2022-08-22
Updated dependencies and added yaml column fix

dushan-madetech:
Was getting an error when runninng the tests after the upgrades (see https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/runs/7953019336?check_suite_focus=true)
so I updated the environments as per the discussion here https://discuss.rubyonrails.org/t/cve-2022-32224-possible-rce-escalation-bug-with-serialized-columns-in-active-record/81017


baarkerlounger (approved):
Given that all the timestamps in the versions table are autogenerated this seems fine.

PR #840, opened 2022-08-22
Rename Case Log to Lettings Log

dushan-madetech:
This PR renames case log to lettings log as a precursor to work we'll be doing to add sales logs to the system


app/models/lettings_log.rb
```diff
@@ -14,13 +14,15 @@ def validate(record)
   end
 end
 
-class CaseLog < ApplicationRecord
+class LettingsLog < ApplicationRecord
+  self.table_name = "lettings_logs"
```
baarkerlounger: Why is this needed? Isn't this the default?

app/models/scheme.rb
```diff
@@ -2,7 +2,7 @@ class Scheme < ApplicationRecord
   belongs_to :owning_organisation, class_name: "Organisation"
   belongs_to :managing_organisation, optional: true, class_name: "Organisation"
   has_many :locations, dependent: :delete_all
-  has_many :case_logs, dependent: :delete_all
+  has_many :case_logs, class_name: "LettingsLog", dependent: :delete_all
```
baarkerlounger: `has_many :lettings_logs`?

dushan-madetech: I haven't gotten around to changing this file yet but will be sure I make this change 😄

app/models/lettings_log.rb
dushan-madetech: If you're referring to the `self.table_name` line that was in the branch when I picked it up so I just updated it. This is still very much a WIP branch so I'll remove this line by the time it gets to review 😄

db/migrate/20220823083657_change_case_log_to_lettings_log.rb
```diff
@@ -0,0 +1,5 @@
+class ChangeCaseLogToLettingsLog < ActiveRecord::Migration[7.0]
+  def change
+    rename_table :case_logs, :lettings_logs
```
jamesrose, line 3: I didn't do this because it wouldn't be a zero downtime deploy. The old application instance would error as it would still be reading from the previous table.

baarkerlounger: True but the rename should be pretty close to instantaneous so given the usage volume I wouldn't expect it to be much of an issue.

baarkerlounger (approved):

db/migrate/20220823083657_change_case_log_to_lettings_log.rb
jamesrose: The rename will be quick, but the upper bound is the time it takes to switch the instances. Your call!

jamesrose (approved):

PR #841, opened 2022-08-22
Remove schemes feature flag

baarkerlounger:


app/models/form/setup/pages/rent_type.rb
```diff
@@ -3,7 +3,6 @@ def initialize(_id, hsh, subsection)
     super("rent_type", hsh, subsection)
     @header = ""
     @description = ""
-    @depends_on = [{ "supported_housing_schemes_enabled?" => true }]
```
baarkerlounger, line 6: This was incorrect and shouldn't have been here since rent_type is not supported housing specific.

spec/models/form/setup/pages/rent_type_spec.rb
```diff
@@ -26,8 +26,4 @@
   it "has the correct description" do
     expect(page.description).to eq("")
   end
-
-  it "has the correct depends_on" do
```
baarkerlounger, line 30: But at least it was also tested...

dushan-madetech (approved):
Nice 👍

PR #842, opened 2022-08-22
Cldc 1434 csv column ordering

kosiakkatrina:
Keep metadata at the beginning of the csv
Order csv attributes based on question order in the form including checkbox answer options
Move is inferred csv attributes next to the inferred fields
Add is dpo attribute to csv
Move scheme attributes to the  back
Order of fields:

**Metadata**
```
id, status, created_at, updated_at, created_by_name, is_dpo, owning_organisation_name, managing_organisation_name, 
```
**Setup section**
```
needstype, renewal, startdate, rent_type, irproduct_other, tenancycode, propcode,
```
**Property information section**
```
postcode_known, postcode_full, is_la_inferred, la, first_time_property_let_as_social_housing, unitletas, rsnvac, offered, unittype_gn, builtype, 
wchair, beds, voiddate, void_date_value_check, majorrepairs, mrcdate, major_repairs_date_value_check,

```
**Tenancy information section**
```
startertenancy, tenancy, tenancyother, tenancylength, sheltered,
```
**Household characteristics section**
```
declaration, hhmemb, pregnancy_value_check, age1_known, age1, sex1, ethnic_group, ethnic, national, ecstat1, retirement_value_check, 
details_known_2, relat2, age2_known, age2, sex2, ecstat2, details_known_3, relat3, age3_known, age3, sex3, ecstat3, 
details_known_4, relat4, age4_known, age4, sex4, ecstat4, details_known_5, relat5, age5_known, age5, sex5, ecstat5, 
details_known_6, relat6, age6_known, age6, sex6, ecstat6, details_known_7, relat7, age7_known, age7, sex7, ecstat7, 
details_known_8, relat8, age8_known, age8, sex8, ecstat8,
```

**Household needs section**
```
armedforces, leftreg, reservist, preg_occ, housingneeds_a, housingneeds_b, housingneeds_c, housingneeds_f, housingneeds_g, housingneeds_h,
illness, illness_type_4, illness_type_5, illness_type_2, illness_type_6, illness_type_7, illness_type_3, illness_type_9, illness_type_8, illness_type_1, illness_type_10, illness_type_0,
```
**Household situation section**
```
layear, waityear, reason, reasonother, prevten, underoccupation_benefitcap, homeless, ppcodenk, ppostcode_full, previous_la_known, is_previous_la_inferred, prevloc, 
reasonpref, rp_homeless, rp_insan_unsat, rp_medwel, rp_hardship, rp_dontknow, cbl, cap, chr, letting_allocation_unknown, referral, 
```

**Income,  benefits and outgoings section**
```
net_income_known, earnings, incfreq, net_income_value_check, hb, benefits, household_charge, 
period, is_carehome, chcharge, brent, scharge, pscharge, supcharg, tcharge, rent_value_check, hbrentshortfall, 
tshortfall_known, tshortfall,
```

**Other fields**
```
property_owner_organisation, property_manager_organisation, sale_or_letting, purchaser_code, property_relet, incref, 
sale_completion_date, renttype, lettype, totchild, totelder, totadult, nocharge, ethnic_other, 
has_benefits, wrent, wscharge, wpschrge, wsupchrg, wtcharge, wtshortfall, refused, housingneeds, wchchrg, newprop, 
old_form_id, lar, irproduct, old_id, joint, hhtype, new_old, vacdays, housingneeds_type, housingneeds_other, 
unittype_sh,
```

**Scheme fields**
```
scheme_id, location_id
```


app/models/case_log.rb
```diff
@@ -439,6 +439,10 @@ def created_by_name
     created_by&.name
   end
 
+  def is_dpo
```
baarkerlounger, line 441: Delegate to `created_by` maybe?

app/models/case_log.rb
```diff
@@ -460,10 +464,69 @@ def self.label_from_value(value)
   end
 
   def self.csv_attributes(user)
```
baarkerlounger: Wonder if it's worth splitting out the CSV specific stuff into it's own module or service class or something... this isn't really a suggestion just a pondering...

kosiakkatrina: Was thinking the same thing. I would've liked to chain all of the new commands as well, but unsure how. Very keen to refactor this to something nicer

app/models/case_log.rb
```diff
@@ -460,10 +464,69 @@ def self.label_from_value(value)
...
+    attributes = attribute_names + %w[unittype_sh]
+    attributes = replace_csv_id_fields_with_names(attributes)
+
+    attributes = order_csv_attributes(attributes)
+
     user.present? && !user.support? ? attributes - CSV_FIELDS_TO_OMIT : attributes
   end
 
+  def self.replace_csv_id_fields_with_names(initial_attributes)
+    attributes = initial_attributes.clone
```
baarkerlounger: Do we need the `.clone` here?

app/models/case_log.rb
```diff
@@ -460,10 +464,69 @@ def self.label_from_value(value)
...
+    { "managing_organisation_id": "managing_organisation_name", "owning_organisation_id": "owning_organisation_name", "created_by_id": "created_by_name" }.each { |current, new_attribute| attributes[attributes.index(current.to_s)] = new_attribute }
+    attributes
+  end
+
+  def self.order_csv_attributes(initial_attributes)
+    attributes = initial_attributes.clone
+    ordered_default_form_questions = move_checkbox_answer_options(FormHandler.instance.forms.first.second.questions)
```
baarkerlounger: Should this always take the first form? What if the logs being downloaded are all for the second collection window?

app/models/case_log.rb
kosiakkatrina: Well it changes the original if we don't do that. I don't know how I feel about that 😅

app/models/case_log.rb
kosiakkatrina: Yeah, I just went with the default. I'm not sure what it should do, this is the easiest if we assume that the 2 years questions are relatively similar. I'll spike some other options

app/models/case_log.rb
kosiakkatrina: Oh, I think the issue with that is in case created_by is nil it throws an error with a delegate

app/models/case_log.rb
kosiakkatrina: Could do something like this? https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/842/commits/3fc1005e2124a5ed3257bfeca2ee94f64de0ad14 (There's a small typo there)
The fields that come out are the same other than 2 of them are in different places

app/models/case_log.rb
baarkerlounger: Even with `allow_nil: true`?

app/services/csv/case_log_csv_service.rb
```diff
@@ -0,0 +1,90 @@
+module Csv
+  class CaseLogCsvService
+    CSV_FIELDS_TO_OMIT = %w[hhmemb net_income_value_check sale_or_letting first_time_property_let_as_social_housing renttype needstype postcode_known is_la_inferred totchild totelder totadult net_income_known is_carehome previous_la_known is_previous_la_inferred age1_known age2_known age3_known age4_known age5_known age6_known age7_known age8_known letting_allocation_unknown details_known_2 details_known_3 details_known_4 details_known_5 details_known_6 details_known_7 details_known_8 rent_type wrent wscharge wpschrge wsupchrg wtcharge wtshortfall rent_value_check old_form_id old_id retirement_value_check tshortfall_known pregnancy_value_check hhtype new_old vacdays].freeze
+
+    def initialize(user)
```
baarkerlounger, line 5: Currently this would export all logs right? Shouldn't it rather take a collection as an input?

kosiakkatrina: _.all_ returns already filtered logs

baarkerlounger: It does if you call it on a collection, but if you call it on the class (as on line 14) what is it filtering by?

kosiakkatrina: I mean from what I can see, it does filter it out, I'm happy to pass it in, but I'd be passing in the same thing, because that's what was being used in the to_csv method to begin with (I think ??)

baarkerlounger: Yep you're right it does.

baarkerlounger (approved):

PR #843, opened 2022-08-23
Cldc 1436 csv schemes info

natdeanlewissoftwire:
Tested locally
Minimal risk
One comment documenting unclear variable naming - happy to rename it if that's better?

ticket: https://digital.dclg.gov.uk/jira/browse/CLDC-1436


app/models/case_log.rb
```diff
@@ -439,6 +439,91 @@ def created_by_name
     created_by&.name
   end
 
+  def scheme_code
+    scheme&.id ? "S#{scheme.id}" : nil
+  end
+
+  def scheme_name
```
baarkerlounger: Could use delegate for these instead perhaps? https://apidock.com/rails/Module/delegate

app/models/case_log.rb
```diff
@@ -442,6 +442,29 @@ def is_dpo
...
   end
 
+  delegate :service_name, :sensitive, :registered_under_care_act, :primary_client_group, :has_other_client_group, :secondary_client_group, :support_type, :intended_stay, :created_at, prefix: "scheme", to: :scheme, allow_nil: true
+  delegate :scheme_type, to: :scheme, allow_nil: true
+
+  def scheme_code
+    scheme&.id ? "S#{scheme.id}" : nil
+  end
+
+  def scheme_owning_organisation_name
```
baarkerlounger, line 452: These could be delegated as well couldn't they?

app/models/case_log.rb
```diff
@@ -442,6 +442,29 @@ def is_dpo
...
+
+  def scheme_managing_organisation_name
+    scheme&.managing_organisation&.name
+  end
+
+  delegate :postcode, :name, :units, :type_of_unit, :mobility_type, :startdate, prefix: "location", to: :location, allow_nil: true
+  delegate :location_admin_district, to: :location, allow_nil: true
+
+  # This is not the location_code in the db, location.id is just called code in the UI
+  def location_code
```
baarkerlounger, line 464: I guess these can't because we're using `code` rather than `id`? Although actually that makes me wonder whether we shouldn't rather be consistent and refer to `id` when we mean `id`?

app/models/case_log.rb
natdeanlewissoftwire: Yes, though the name still has to be defined separately I think? Unless I delegated the org names to schemes and **then** to case_logs but that seems like overkill. I've just delegated the orgs themselves for now

app/models/case_log.rb
natdeanlewissoftwire: So this is weird because the location's postcode is being recognised as its id, rather than the column actually named id - if I delegate the id from location then it actually returns the postcode for some reason.

baarkerlounger (approved):

PR #847, opened 2022-08-25
Cldc 1429 output la code

natdeanlewissoftwire:
Tested locally
Slight risk from future confusion arising from la and prevloc not being displayed in the same way as other columns in the database - it is la_label and prevloc_label that have human-readable locations, whereas previously la and prevloc were human readable not codes.
No documentation required

Ticket: https://digital.dclg.gov.uk/jira/browse/CLDC-1429


app/models/case_log.rb
```diff
@@ -465,6 +465,14 @@ def location_code
     location&.id
   end
 
+  def la_label
```
baarkerlounger: Are these methods needed at all if they're just aliases and we special case them in the csv service anyway?

natdeanlewissoftwire: Yep they're still necessary, as we still want the _label columns to appear in the csv alongside la and prevloc. Deleting these methods means this error surfaces on csv download: 
NoMethodError in CaseLogsController#index
undefined method `la_label' for #<CaseLog...

baarkerlounger (approved):

app/models/case_log.rb
baarkerlounger: Was thinking we could just call these bits directly there but not super important

PR #851, opened 2022-08-30
CLDC-1362 Improve CSV download performance

CheshireSwift:
Adds async CSV download functionality to logs. This PR:

- Adds two interstitial pages to the download button to show that it will be emailed vs. previously downloaded inline
- Adds Sidekiq to the project
- Adds worker instance to the project
- Injects name of CSV bucket into environment

We've already added the S3 bucket to `staging` and `production` services.


app/services/filter_service.rb
```diff
@@ -0,0 +1,22 @@
+class FilterService
+  def self.filter_by_search(base_collection, search_term = nil)
+    if search_term.present?
+      base_collection.search_by(search_term)
+    else
+      base_collection
+    end
+  end
+
+  def self.filter_case_logs(logs, search_term, filters, all_orgs, user)
```
CheshireSwift: This was copied from `case_logs_filter`, which isn't specced. Not sure if we want to remedy that, or if we're happy that the page tests (which will make use of it) are sufficient.

app/controllers/organisations_controller.rb
```diff
@@ -103,7 +103,8 @@ def logs
         end
 
         format.csv do
-          send_data unpaginated_filtered_logs.to_csv, filename: "logs-#{@organisation.name}-#{Time.zone.now}.csv"
+          EmailCsvJob.perform_later(current_user, search_term, filters, "logs-#{Time.zone.now}.csv")
```
CheshireSwift: This needs to be moved out to a POST + confirm flow for the organisation logs the same way it is for the main logs page. I could use some direction on how best to do that without just duplicating the ERB + routes (Rails helpers etc. in `_log_list.html.erb`?)

app/jobs/email_csv_job.rb
```diff
@@ -0,0 +1,16 @@
+class EmailCsvJob < ApplicationJob
+  queue_as :default
+
+  CSV_DOWNLOAD_TEMPLATE_ID = "QQ".freeze # Need an E-mail template + ID
```
CheshireSwift: E-mail template required.

app/jobs/email_csv_job.rb
```diff
@@ -0,0 +1,16 @@
...
+
+  def perform(user, search_term = nil, filters = {}, all_orgs = false)
+    logs = FilterService.filter_case_logs(user.case_logs, search_term, filters, all_orgs, user)
+
+    filename = "logs-#{Time.zone.now}.csv"
+
+    storage_service = S3StorageService.new(PaasConfigurationService.new, ENV["CSV_DOWNLOAD_PAAS_INSTANCE"])
```
CheshireSwift: S3 setup also required.

config/routes.rb
```diff
@@ -77,6 +77,8 @@
     collection do
       post "bulk-upload", to: "bulk_upload#bulk_upload"
       get "bulk-upload", to: "bulk_upload#show"
+      post "email-csv", to: "case_logs#email_csv"
```
CheshireSwift: Is it worth splitting these out into their own controller (a la `bulk_upload`), rather than bolting onto `case_logs`?

baarkerlounger: Does it need a new route at all? Is there a reason not to use the existing route/controller action based on format requested and then extract the email specific logic elsewhere?

app/controllers/case_logs_controller.rb
```diff
@@ -9,21 +9,18 @@ class CaseLogsController < ApplicationController
...
+
+        unpaginated_filtered_logs = filtered_case_logs(all_logs, search_term, filters)
+
         @pagy, @case_logs = pagy(unpaginated_filtered_logs)
         @searched = search_term.presence
         @total_count = all_logs.size
       end
-
-      format.csv do
-        send_data unpaginated_filtered_logs.to_csv, filename: "logs-#{Time.zone.now}.csv"
```
baarkerlounger: i.e. we could just call `email_csv` from in here as before?

app/jobs/email_csv_job.rb
```diff
@@ -0,0 +1,16 @@
...
+    storage_service.write_file(filename, logs.to_csv)
+    url = storage_service.get_presigned_url(filename)
+    DeviseNotifyMailer.new.send_email(user.email, CSV_DOWNLOAD_TEMPLATE_ID, { url: })
```
jamesrose: I think `DeviseNotifyMailer` is intended to be inherited as part of another mailer class. So, you could bring in the template ID into that:

```ruby
class UserMailer < DeviseNotifyMailer
  TEMPLATE_ID = "..."

  def csv_export_mail(...)
     send_email(...)
   end
end

File: app/controllers/case_logs_controller.rb [hunk 8]
```diff
@@ -9,21 +9,18 @@ class CaseLogsController < ApplicationController
   before_action :find_resource, except: %i[create index edit]
 
   def index
-    set_session_filters
-
-    all_logs = current_user.case_logs
-    unpaginated_filtered_logs = filtered_case_logs(filtered_collection(all_logs, search_term))
+    filters = set_updated_session_filters
```
baarkerlounger: Personal preference but I think nouns over verbs for methods seems more idiomatic (it doesn't seem like this one has side effects).

app/controllers/modules/case_logs_filter.rb
```diff
@@ -1,23 +1,16 @@
...
-      filters.each do |category, values|
-        next if Array(values).reject(&:empty?).blank?
-        next if category == "organisation" && params["organisation_select"] == "all"
-
-        logs = logs.public_send("filter_by_#{category}", values, current_user)
-      end
-    end
-    logs = logs.order(created_at: :desc)
-    current_user.support? ? logs.all.includes(:owning_organisation, :managing_organisation) : logs
+  def filtered_case_logs(logs, search_term, filters)
```
baarkerlounger: Why is the filtering logic changing? Shouldn't we be returning the same records in the CSV as before?

app/jobs/email_csv_job.rb
jamesrose: @CheshireSwift Do you have access to Notify?

app/jobs/email_csv_job.rb
baarkerlounger: `DeviseNotifyMailer` is meant to be specific to authentication related stuff really (i.e. https://github.com/heartcombo/devise) so might be better to have another mailer but extract common functionality if there is?

jamesrose: @baarkerlounger Whoops, yes. I misread the name of that: I thought it was `NotifyMailer`.

config/routes.rb
CheshireSwift: I'm not sure I follow. Are you suggesting that the route should still be a request for `text/csv`? That seems like an odd interface if the outcome is redirecting them to an HTML page that tells them they'll receive an E-mail.

Or am I misunderstanding what requested formats are for/doing?

app/controllers/modules/case_logs_filter.rb
CheshireSwift: The logic isn't (shouldn't be!) changing, but needed commonising between the job and this page. The goal for extracting it is ensuring we do get the same logs back, without having to pass them all to the job (and thus retrieve them only to serialise them into redis).

app/jobs/email_csv_job.rb
CheshireSwift: @jamesrose Not that I'm aware of, assuming it's an independent login from other services?

app/controllers/case_logs_controller.rb
CheshireSwift: I struggled with naming this; it has side effects (it sets the filters on the session for reading back next time), but we also need access to those filters for the E-mail job.

My immediate inclination was splitting it into two parts (the function to get the filters, the function to set them on the session), but I think in practice any time we're getting them we'll also want to update the session. I was avoiding the need to always call both together.

Welcome to suggestions on naming or organisation on this, I'm certainly not entirely happy with it as-is 😄

app/jobs/email_csv_job.rb
jamesrose: @CheshireSwift Realistically we'll want the first three methods of `DeviseNotifyMailer` extracted into `NotifyMailer`. Or even, just `ApplicationMailer`.

config/routes.rb
baarkerlounger: Hmm no that is what I was suggesting, and when you put it that way it does seem a bit odd... I guess I was thinking of it like, what you're really asking the server for is still something in a text/csv format, the fact that it now happens to arrive via an email just being an implementation detail, but it's not _really_ correct I suppose...

app/jobs/email_csv_job.rb
baarkerlounger: Yes it is. https://www.notifications.service.gov.uk/. You'll need to ask Gary Meyler to provide you a login (teams or email is best)

app/controllers/case_logs_controller.rb
jamesrose: I don't think that's quite right as we're not returning that data anymore. The action is to send an email vs return logs.

app/controllers/case_logs_controller.rb
jamesrose: Yes, I think it would be better to split out into two parts. The first sets the filters to the session and the second is a getter for either the raw filters or the filtered case logs.

```ruby
  before_action :set_session_filters # Or, keep inline in actions.
  before_action :find_session_filters

  private

  # Inline just for example
  def set_session_filters
    ...
  end

  def find_session_filters
    @session_filters ||= load_session_filters(session[:case_logs_filters]) # extract this method from set_session_filters
  end
```

spec/requests/case_logs_controller_spec.rb
```diff
@@ -961,4 +876,78 @@
...
+
+      it "passes filter parameters" do
+        expect {
+          post "/logs/email-csv?status[]=completed", headers:, params: {}
+        }.to enqueue_job(EmailCsvJob).with(user, nil, { "status" => %w[completed] }, false)
+      end
+
+      it "passes a combination of search term and filter parameters" do
+        postcode = "XX1 1TG"
+        a = %w[test]
```
jamesrose:
```suggestion
```

This looks like it doesn't belong here.

app/jobs/email_csv_job.rb
CheshireSwift: Given that `DeviseNotifyMailer` is extending `Devise::Mailer` and `ApplicationMailer` is extending `ActionMailer::Base`, what's the desired inheritance hierarchy here? I'm keen to share common parts (the `intercept_send?` method jumps out), but not sure of the preferred way of handling the potential multiple inheritance. Some sort of mixin?

Thus far my WIP `CsvDownloadMailer` doesn't inherit from anywhere, and I've duplicated `intercept_send?`, but obviously that's not an ideal solution 😄

app/controllers/case_logs_controller.rb
CheshireSwift: ~~Don't we need an unconditional assignment to `@session_filters` (since we're looking to update it each time?)~~

I'm guessing Rails controllers don't persist instance variables between requests, and thus `@session_filters` will be empty unless we've already called it for this request? 🙂

app/controllers/organisations_controller.rb
CheshireSwift: I've got an implementation of this. Not 100% happy, but welcome either suggestions on how to improve, or agreement that it's a reasonable approach.

app/jobs/email_csv_job.rb
jamesrose: I think we should just copy over the `send_email` email methods to `ApplicationMailer` for now.

app/controllers/case_logs_controller.rb
jamesrose: @CheshireSwift The `before_action` will set it on every action unless it's already been set.

CheshireSwift:
I've manually tested this (locally) for both the mainline `/logs` page and the `/organisation/{id}/logs` page that support users have access to, up to seeing the email job queued.

I haven't manually tested the job/mail behaviour (only the unit tests), since that'll involve actually talking to S3 + Notify.

app/controllers/modules/lettings_logs_filter.rb
```diff
@@ -1,23 +1,21 @@
...
-        next if Array(values).reject(&:empty?).blank?
-        next if category == "organisation" && params["organisation_select"] == "all"
-
-        logs = logs.public_send("filter_by_#{category}", values, current_user)
-      end
-    end
-    logs = logs.order(created_at: :desc)
-    current_user.support? ? logs.all.includes(:owning_organisation, :managing_organisation) : logs
+  def filtered_lettings_logs(logs, search_term, filters)
+    all_orgs = params["organisation_select"] == "all"
```
baarkerlounger, line 3: Could pass this in like you do for the job (saves doing the comparison twice)

app/jobs/email_csv_job.rb
```diff
@@ -0,0 +1,21 @@
+class EmailCsvJob < ApplicationJob
+  queue_as :default
+
+  BYTE_ORDER_MARK = "\uFEFF".freeze # Required to ensure Excel always reads CSV as UTF-8
+
+  EXPIRATION_TIME = 3.hours.to_i
+
+  def perform(user, search_term = nil, filters = {}, all_orgs = false, organisation = nil) # rubocop:disable Style/OptionalBooleanParameter
```
baarkerlounger: Any reason not to just make these keyword args rather than turning off the cop?

app/jobs/email_csv_job.rb
```diff
@@ -0,0 +1,21 @@
...
+    unfiltered_logs = organisation.present? && user.support? ? LettingsLog.all.where(owning_organisation_id: organisation.id) : user.lettings_logs
```
baarkerlounger: Think the `.all` might be redundant here?

baarkerlounger: Also not sure if the precedence might be a bit ambiguous, might be easier to parse with parenthesis?

spec/jobs/email_csv_job_spec.rb
```diff
@@ -0,0 +1,159 @@
...
+    end
+
+    context "when writing to S3" do
+      before do
+        FactoryBot.create_list(:lettings_log, 4, owning_organisation: other_organisation)
+      end
+
+      def expect_csv
+        expect(storage_service).to receive(:write_file) do |_filename, data|
+          # Ignore BOM
```
baarkerlounger: BOM?

app/mailers/csv_download_mailer.rb
```diff
@@ -0,0 +1,11 @@
+class CsvDownloadMailer < NotifyMailer
+  CSV_DOWNLOAD_TEMPLATE_ID = "7890e3b9-8c0d-4d08-bafe-427fd7cd95bf".freeze
```
baarkerlounger, line 2: Does this template exist already?

baarkerlounger (approved):
Comments are mostly questions - LGTM.

app/jobs/email_csv_job.rb
CheshireSwift: Sidekiq doesn't support serialising keyword args. But I should probably add a comment to that effect 🙂

app/mailers/csv_download_mailer.rb
CheshireSwift: Yes, I created it in Notify (per Sarah's content spec).

spec/jobs/email_csv_job_spec.rb
CheshireSwift: Byte Order Marker (from a fix Nat added to make Excel import the CSV as UTF-8 automatically). Will de-abbreviate in comment.

jamesrose (approved):
Nice work!

PR #852, opened 2022-09-01
Cldc 1436 csv schemes info

natdeanlewissoftwire:
This update removes scheme_id and locaiton_id from support users' exports as well - this was only removed for non-support users before

Ticket: https://digital.dclg.gov.uk/jira/browse/CLDC-1436


dushan-madetech (approved):
Looks good to me! 👍

PR #853, opened 2022-09-01
CLDC-1428 Prevent creation of new 2021/22 logs

davidm-m:
Since we don't make any distinction between a newly created log and a log that's being edited I decided to use the `created_at` field in the database to determine whether a log should be allowed to have a tenancy start date in the 2021/22 financial year. A few points:
* This implementation means we can set the end of a collection window in the form definition and have it be closed for new submissions without needing a release
* I'm unsure if `record.startdate < second_collection_start_date` is the correct way to see if a log falls into the first collection window - I think it will always work, but there might be a more idiomatic way to check?
* We need the nullcheck on `created_at` because logs created via the API are validated before being inserted into the database, so the `created_at` field isn't populated yet
* The `lettings_log_controller` tests use a 2021 log as an example and so needs the time travel to not get an error about submitting a log after the collection window - is it worth raising a ticket to update that test to a 2022 log?

### DIFF
--- a/app/models/validations/date_validations.rb
+++ b/app/models/validations/date_validations.rb
@@ -31,7 +31,9 @@ module Validations::DateValidations
   def validate_startdate(record)
     return unless record.startdate && date_valid?("startdate", record)
 
-    if record.created_at > first_collection_end_date && record.startdate < second_collection_start_date
+    created_at = record.created_at || Time.zone.now
+
+    if created_at > first_collection_end_date && record.startdate < second_collection_start_date
       record.errors.add :startdate, I18n.t("validations.date.outside_collection_window")
     end
 
--- a/app/models/validations/date_validations.rb
+++ b/app/models/validations/date_validations.rb
@@ -31,6 +31,10 @@ module Validations::DateValidations
   def validate_startdate(record)
     return unless record.startdate && date_valid?("startdate", record)
 
+    if record.created_at > first_collection_end_date && record.startdate < second_collection_start_date
+      record.errors.add :startdate, I18n.t("validations.date.outside_collection_window")
+    end
+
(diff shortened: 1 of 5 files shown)


jamesrose:
As an aside, it makes me sad that `created_at` is a nullable field.

app/models/validations/date_validations.rb
```diff
@@ -57,6 +63,14 @@ def first_collection_start_date
     @first_collection_start_date ||= FormHandler.instance.forms.map { |form| form.second.start_date }.compact.min
```
jamesrose, line 63: Do you need this line now you've added the memoization method below?

davidm-m: I'm pretty sure we still need this one, otherwise after we remove the 2021 form someone could enter any start date in the past and it would pass validation.

spec/requests/lettings_logs_controller_spec.rb
```diff
@@ -49,6 +51,7 @@
       end
 
       before do
+        travel_to Time.utc(2022, 2, 8, 16, 52, 15)
         post "/logs", headers:, params: params.to_json
       end
```
dushan-madetech: I think that you might be able to change this to use Timecop like we do in other time related tests. 

It would look something like this:

```ruby
before do
  Timecop.freeze(Time.utc(2022, 2, 8, 16, 52, 15))
  post "/logs", headers:, params: params.to_json
end

after do
  Timecop.unfreeze
end
```

davidm-m: I thought I'd already searched to see if we used Timecop in the codebase but clearly hadn't 🤦‍♂️
Thanks for catching!

PR #854, opened 2022-09-02
Store object changes in PaperTrail versions

jamesrose:
If you add an object_changes column to your versions table, PaperTrail will store the changes diff in each version.

You can test this by making a change on an audited model and inspecting the `changeset`:

```ruby
irb(main):008:0> Organisation.last.versions.last.changeset
  PaperTrail::Version Load (0.4ms)  SELECT "versions".* FROM "versions" WHERE "versions"."item_id" = $1 AND "versions"."item_type" = $2 ORDER BY "versions"."created_at" DESC, "versions"."id" DESC LIMIT $3  [["item_id", 2], ["item_type", "Organisation"], ["LIMIT", 1]]
  Organisation Load (0.2ms)  SELECT "organisations".* FROM "organisations" WHERE "organisations"."id" = $1 LIMIT $2  [["id", 2], ["LIMIT", 1]]
=> {"name"=>["FooBar LTD", "BarFoo"], "updated_at"=>[Thu, 18 Aug 2022 10:59:18.067767000 BST +01:00, Fri, 02 Sep 2022 09:28:15.401814000 BST +01:00]}
```

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema[7.0].define(version: 2022_08_23_083657) do
+ActiveRecord::Schema[7.0].define(version: 2022_09_02_082245) do
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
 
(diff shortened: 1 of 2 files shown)


baarkerlounger:
Think there should be a schema file change committed here?

db/schema.rb
```diff
@@ -232,40 +232,12 @@
     t.integer "void_date_value_check"
     t.integer "housingneeds_type"
     t.integer "housingneeds_other"
-    t.index ["created_by_id"], name: "index_case_logs_on_created_by_id"
```
baarkerlounger: This looks like there's more going on here than what the migration does

jamesrose: Yep, just fixing that, but something doesn't seem right with our schema.

baarkerlounger (approved):

PR #855, opened 2022-09-02
Record PaperTrail whodunnit for console users on production

jamesrose:
This will allow us to audit who has made changes in the console to audited records.

This was taken from the PaperTrail documentation [here](https://github.com/paper-trail-gem/paper_trail/wiki/Setting-whodunnit-in-the-rails-console#ask-for-a-name-only-ifwhen-you-start-changing-records-in-the-console), but adjusted so that it is applicable only to our production environment.

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -131,4 +131,20 @@ Rails.application.configure do
 
   # see https://discuss.rubyonrails.org/t/cve-2022-32224-possible-rce-escalation-bug-with-serialized-columns-in-active-record/81017
   config.active_record.yaml_column_permitted_classes = [Time]
+
+  # From https://github.com/paper-trail-gem/paper_trail/wiki/Setting-whodunnit-in-the-rails-console
+  console do
+    PaperTrail.request.whodunnit = lambda {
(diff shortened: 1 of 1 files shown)


config/environments/production.rb
```diff
@@ -131,4 +131,20 @@
...
   config.active_record.yaml_column_permitted_classes = [Time]
+
+  # From https://github.com/paper-trail-gem/paper_trail/wiki/Setting-whodunnit-in-the-rails-console
+  console do
+    PaperTrail.request.whodunnit = lambda {
+      @paper_trail_whodunnit ||= begin
+        email = nil
+        until email.present?
+          # rubocop:disable Rails/Output
+          puts "Enter your email address for PaperTrail"
```
baarkerlounger, line 142: Wonder if there's anyway to automatically get this based on the user used to initiate the ssh session

baarkerlounger (approved):

config/environments/production.rb
jamesrose: Good idea. I'll have a look.

PR #856, opened 2022-09-02
rename indexes

kosiakkatrina:
Generated this by deleting the schema file, recreating tables and applying the migrations

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -232,12 +232,12 @@ ActiveRecord::Schema[7.0].define(version: 2022_09_02_082245) do
     t.integer "void_date_value_check"
     t.integer "housingneeds_type"
     t.integer "housingneeds_other"
-    t.index ["created_by_id"], name: "index_case_logs_on_created_by_id"
-    t.index ["location_id"], name: "index_case_logs_on_location_id"
-    t.index ["managing_organisation_id"], name: "index_case_logs_on_managing_organisation_id"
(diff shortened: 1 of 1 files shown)


baarkerlounger:
Could be wrong but without a migration I don't _think_ this will actually change the index names of prod database, so would lead to the schema potentially not matching what's actually there.

Might be better as @jamesrose suggested to write a migration that removes the old index and adds this one to achieve this result.

dushan-madetech:
> Could be wrong but without a migration I don't _think_ this will actually change the index names of prod database, so would lead to the schema potentially not matching what's actually there.
> 
> Might be better as @jamesrose suggested to write a migration that removes the old index and adds this one to achieve this result.

@baarkerlounger This is actually done by [this migration](https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/blob/main/db/migrate/20220823083657_change_case_log_to_lettings_log.rb) but in the process of rebasing the large rename branch multiple times the case logs bit slipped through. You could see this by dropping the databases, creating them and then running the migrations (on main) and you'll end up with the same diff as this PR

baarkerlounger:
@dushan-madetech that migration renames the table but unless I'm missing something it doesn't rename the indexes on it?

kosiakkatrina:
> @dushan-madetech that migration renames the table but unless I'm missing something it doesn't rename the indexes on it?

@baarkerlounger it did for me, try rolling back and migrating

baarkerlounger (approved):

PR #857, opened 2022-09-02
set housingneeds_type to none of the above if no housingneeds selected

kosiakkatrina:


baarkerlounger:
Should this have a test? Does it require updating prod data?

kosiakkatrina:
> Should this have a test? Does it require updating prod data?

@baarkerlounger, no changes needed to the prod data, not sure about the test. I guess it should be covered by the import tests, but I think all that tests is that importing it twice imports it the same way (?)

baarkerlounger:
@kosiakkatrina how come no production data changes? Aren't we introducing a new value here? Or are we saying there's currently no logs in production with this scenario?

For tests I meant a failing test for this change.

kosiakkatrina:
> 

@baarkerlounger I've already updated it when I did the update for the previous PR. I just realised that the import needs to be updated too

baarkerlounger (approved):
Think a test would still be useful

PR #858, opened 2022-09-02
Extract base log and create sales log

dushan-madetech:
- [x] Creates a Sales log class
- [x] Creates a parent Log class
- [x] Defines a sales log form in code instead of json
- [x] Add a feature flag for the new "Create a new sales log" button

TODO:
- [x] Doesn't delete logs if a user is deleted
- [x] Add a separate sales logs page behind a feature flag
- [x] Route `/logs` to `/lettings-logs`

Later:
- [ ] Refactor FormHandler to use date and type of log instead of form name - [CLDC-1600](https://digital.dclg.gov.uk/jira/browse/CLDC-1600)
- [ ] Double check classes that refer to lettings_log but should refer to log - [CLDC-1601](https://digital.dclg.gov.uk/jira/browse/CLDC-1601)
- [ ] Check if any functionality living in the log models should be in the form model (for example reset_not_routed_questions) - [CLDC-1602](https://digital.dclg.gov.uk/jira/browse/CLDC-1602)
- [ ] Add `Which organisation owns this log?` and `Which user are you creating this log for?` when a support user is creating a sales log - [CLDC-1603](https://digital.dclg.gov.uk/jira/browse/CLDC-1603)


app/controllers/form_controller.rb
```diff
@@ -109,11 +106,19 @@ def responses_for_page(page)
   end
 
   def find_resource
-    @lettings_log = current_user.lettings_logs.find_by(id: params[:id])
+    @log = if URI(request.referer).path.split("/")[1] == "sales-logs"
```
jamesrose: Ouch! I don't think this is going to work reliably. Remember that referer is entirely optional. Can you think of a better way to do this?

app/models/sales_log.rb
```diff
@@ -0,0 +1,70 @@
...
+  def collection_start_year
+    return @sale_year if @sale_year
+    return unless saledate
+
+    window_end_date = Time.zone.local(saledate.year, 4, 1)
+    @sale_year = saledate < window_end_date ? saledate.year - 1 : saledate.year
+  end
+
+  def form
+    FormHandler.instance.get_form(form_name) || FormHandler.instance.get_form("2022_2023_sales")
```
jamesrose, line 30: Do we have to hard code this default value?

app/controllers/form_controller.rb
dushan-madetech: We'll definitely try to! I think this was the first attempt we came up with so definitely scope for improving this 😄

app/models/sales_log.rb
dushan-madetech: I'd imagine not in the long term, for the sake of ease and progressing onwards with the implementation of the Sales log we've used this approach. Will see if I can replace this with a better solution by the time this PR gets to a proper review ready state 🙂

app/models/form_handler.rb
```diff
@@ -14,9 +14,12 @@ def current_form
     forms[forms.keys.max_by(&:to_i)]
   end
 
-private
+  def sales_forms
+    sales_sections = [Form::Sales::Setup::Sections::Setup]
+    { "2022_2023_sales" => Form.new(nil, "2022_2023_sales", sales_sections, "sales") }
```
jamesrose, line 19: Do we need to name these? Could we get away with something we don't need to change every year, such as `current` and `next`?

app/controllers/form_controller.rb
```diff
@@ -4,62 +4,59 @@ class FormController < ApplicationController
   before_action :find_resource_by_named_id, except: %i[submit_form review]
 
   def submit_form
-    if @lettings_log
-      @page = @lettings_log.form.get_page(params[:lettings_log][:page])
+    if @log
+      @page = @log.form.get_page(params[@log.class.name.underscore.to_sym][:page])
```
jamesrose: How about https://apidock.com/rails/v5.2.3/ActiveModel/Naming/param_key/class or `@log.model_name.param_key`?

app/controllers/form_controller.rb
```diff
@@ -68,25 +65,25 @@ def review
...
+          @log[field] = value
         end
       end
     end
   end
 
   def responses_for_page(page)
     page.questions.each_with_object({}) do |question, result|
-      question_params = params["lettings_log"][question.id]
+      question_params = params[@log.class.name.underscore][question.id]
```
baarkerlounger: `@log.model_name.param_key` here as well?

app/controllers/form_controller.rb
```diff
@@ -68,25 +65,25 @@ def review
...
       if question.type == "date"
-        day = params["lettings_log"]["#{question.id}(3i)"]
-        month = params["lettings_log"]["#{question.id}(2i)"]
-        year = params["lettings_log"]["#{question.id}(1i)"]
+        day = params[@log.class.name.underscore]["#{question.id}(3i)"]
```
baarkerlounger: Also here?

app/controllers/form_controller.rb
```diff
@@ -109,11 +106,19 @@ def responses_for_page(page)
   end
 
   def find_resource
-    @lettings_log = current_user.lettings_logs.find_by(id: params[:id])
+    @log = if params.key?("sales_log")
```
baarkerlounger, line 109: Would prefer to be more explicit about the two expected cases and error handling here but just a nit

app/controllers/form_controller.rb
```diff
@@ -68,25 +65,25 @@ def review
...
+          @log[field] = value
         end
       end
     end
   end
 
   def responses_for_page(page)
     page.questions.each_with_object({}) do |question, result|
-      question_params = params["lettings_log"][question.id]
+      question_params = params[@model_name.param_key][question.id]
```
baarkerlounger: Aren't you missing @log here?

app/models/form.rb
```diff
@@ -5,20 +5,38 @@ class Form
...
-    @questions = pages.flat_map(&:questions)
-    @start_date = Time.iso8601(form_definition["start_date"])
-    @end_date = Time.iso8601(form_definition["end_date"])
+  def initialize(form_path, name, sections_in_form = [], type = "lettings")
+    if type == "sales"
+      # @form_definition = sections.map do |s|
+      #   s.new(nil, nil, self)
+      # end
+      @name = name
+      # @setup_sections = [Form::Setup::Sections::Setup.new(nil, nil, self)]
```
baarkerlounger: Are these comments still needed?

app/models/form.rb
```diff
@@ -5,20 +5,38 @@ class Form
...
+      # @form_definition = JSON.parse(File.open(form_path).read)
+      @form_sections = sections_in_form.map { |sec| sec.new(nil, nil, self) }
+      @type = "sales"
+      @sections = form_sections
+      @subsections = sections.flat_map(&:subsections)
+      @pages = subsections.flat_map(&:pages)
+      @questions = pages.flat_map(&:questions)
+      @start_date = Time.zone.local(name[0..3], 4, 1)
+      @end_date = Time.zone.local(start_date.year + 1, 4, 1)
```
baarkerlounger: This doesn't actually seem like it's necessarily true since they get extra time for late submissions and it seems that date moves more ad-hoc than we thought.

app/models/form.rb
```diff
@@ -5,20 +5,38 @@ class Form
...
-    @pages = subsections.flat_map(&:pages)
-    @questions = pages.flat_map(&:questions)
-    @start_date = Time.iso8601(form_definition["start_date"])
-    @end_date = Time.iso8601(form_definition["end_date"])
+  def initialize(form_path, name, sections_in_form = [], type = "lettings")
+    if type == "sales"
+      # @form_definition = sections.map do |s|
+      #   s.new(nil, nil, self)
+      # end
+      @name = name
```
baarkerlounger, line 10: Does a form actually need a name?

app/controllers/form_controller.rb
dushan-madetech: Yep 🤦

app/models/form_handler.rb
```diff
@@ -27,7 +30,15 @@ def get_all_forms
...
+
+  def get_all_forms
+    lettings_forms.merge(sales_forms)
+  end
+
   def directories
     Rails.env.test? ? ["spec/fixtures/forms"] : ["config/forms"]
   end
+
+  # SALES_FORM = [Form::Sales::Sections::Setup, Form::Sales::Sections::Property]
```
baarkerlounger: Is this still needed?

app/models/sales_log.rb
```diff
@@ -0,0 +1,72 @@
...
+
+class SalesLog < Log
+  has_paper_trail
+
+  validates_with SalesLogValidator
+  before_save :update_status!
+
+  scope :filter_by_organisation, ->(org, _user = nil) { where(owning_organisation: org).or(where(managing_organisation: org)) }
+
+  STATUS = { "not_started" => 0, "in_progress" => 1, "completed" => 2 }.freeze
```
baarkerlounger: Should this stuff be in the parent model? Sales and Lettings logs will have the same statuses right?

app/models/sales_log.rb
```diff
@@ -0,0 +1,72 @@
...
+    attribute_names
+  end
+
+  def form_name
+    return unless saledate
+
+    "#{collection_start_year}_#{collection_start_year + 1}_sales"
+  end
+
+  def collection_start_year
```
baarkerlounger: This should probably also be generisized since it's essentially the same logic for both (and you'd expect to want to change them together I think)

app/models/sales_log.rb
```diff
@@ -0,0 +1,72 @@
...
+    self.status = if all_fields_completed? && errors.empty?
+                    "completed"
+                  elsif all_fields_nil?
+                    "not_started"
+                  else
+                    "in_progress"
+                  end
+  end
+
+  def all_fields_completed?
```
baarkerlounger: Same with these

app/models/user.rb
```diff
@@ -9,6 +9,8 @@ class User < ApplicationRecord
   belongs_to :organisation, optional: true
   has_many :owned_lettings_logs, through: :organisation, dependent: :delete_all
   has_many :managed_lettings_logs, through: :organisation
+  has_many :owned_sales_logs, through: :organisation, dependent: :delete_all
```
baarkerlounger: Does this imply that if we delete the user we delete their logs?

app/models/form.rb
dushan-madetech: I think they'll be needed in the near future - currently sales form only has one section and one question but will have more as we work on the remaining tickets. Happy to remove this in the meantime though 🙂

config/routes.rb
```diff
@@ -84,9 +84,21 @@
...
+        get page.id.to_s.dasherize, to: "form#show_page"
+      end
+
+      form.subsections.map do |subsection|
+        get "#{subsection.id.to_s.dasherize}/check-answers", to: "form#check_answers"
+      end
+    end
+  end
+
+  resources :sales_logs, path: "/sales-logs" do
```
baarkerlounger, line 106: Should we change the path above to "lettings-logs" to match this?

spec/features/form/validations_spec.rb
```diff
@@ -37,7 +37,7 @@
 
   describe "Question validation" do
     context "when the tenant age is invalid" do
-      it "shows validation for under 0" do
+      it "shows validation for under 0", js: true do
```
baarkerlounger: Does this need js or was this just for debugging?

spec/models/sales_log_spec.rb
```diff
@@ -0,0 +1,85 @@
...
+    # let!(:in_progress_sales_log) { FactoryBot.create(:sales_log, :in_progress) }
+    let!(:completed_sales_log) { FactoryBot.create(:sales_log, :completed) }
+
+    it "is set to not started for an empty sales log" do
+      expect(empty_sales_log.not_started?).to be(true)
+      # expect(empty_sales_log.in_progress?).to be(false)
+      expect(empty_sales_log.completed?).to be(false)
+    end
+
+    # it "is set to in progress for a started sales log" do
```
baarkerlounger, line 53: Why are these tests commented out?

app/models/form.rb
dushan-madetech: I think we just put this as a starting point - happy to amend this now or once we know more about if sales logs have the same approach for late submissions

app/models/form.rb
dushan-madetech: Probably not, but it doesn't hurt for distinguishing between the 4 forms we'll need to run

baarkerlounger (commented):
Should probably also rename instances of `lettings_log` in `models/form/*.rb` to `log`

app/models/form_handler.rb
dushan-madetech: Removed 🙂

app/models/user.rb
dushan-madetech: I think so - we just mimicked what was already in place for lettings logs

spec/models/sales_log_spec.rb
dushan-madetech: We only have one question for a sales log so currently it can only have a status of completed or not started (I think anyway)

spec/features/form/validations_spec.rb
dushan-madetech: I suspect this was left over from debugging, nice catch 🙂

app/models/user.rb
baarkerlounger: That absolutely should  not happen for either type of log since the organisation not the user owns the log. Can we double check that please?

app/models/sales_log.rb
dushan-madetech: I would assume so! Will move it across and can always reintroduce it here if that doesn't turn out to be true 😄

app/models/user.rb
dushan-madetech: I can't delete a user or an organisation in order to tests this - it's throwing an error when I attempt to do so through the database directly. Do I need to do it through the rails console?

app/models/form.rb
baarkerlounger: If we keep it in we should probably at least match the lettings form end date? (1st July)

app/models/user.rb
baarkerlounger: What error do you get? (But yes doing so through the rails console is also a valid test)

baarkerlounger: Could also write a spec for it

app/models/form.rb
baarkerlounger: Don't they all get passed into as params though? We don't need to separate this from line 16 for Sales do we?

app/controllers/form_controller.rb
```diff
@@ -123,18 +128,18 @@ def is_referrer_check_answers?
...
-      page_ids = @lettings_log.form.subsection_for_page(@page).pages.map(&:id)
+      page_ids = @log.form.subsection_for_page(@page).pages.map(&:id)
       page_index = page_ids.index(@page.id)
-      next_page = @lettings_log.form.next_page(@page, @lettings_log, current_user)
-      previous_page = @lettings_log.form.previous_page(page_ids, page_index, @lettings_log, current_user)
+      next_page = @log.form.next_page(@page, @log, current_user)
+      previous_page = @log.form.previous_page(page_ids, page_index, @log, current_user)
       if next_page.to_s.include?("value_check") || next_page == previous_page
-        return "/logs/#{@lettings_log.id}/#{next_page.dasherize}?referrer=check_answers"
+        return "/logs/#{@log.id}/#{next_page.dasherize}?referrer=check_answers"
```
jamesrose: Can we use a route here?

app/views/form/check_answers.html.erb
```diff
@@ -1,7 +1,7 @@
 <% content_for :title, "#{subsection.id.humanize} - Check your answers" %>
 <% content_for :breadcrumbs, govuk_breadcrumbs(breadcrumbs: {
   "Logs" => "/logs",
-  "Log #{@lettings_log.id}" => "/logs/#{@lettings_log.id}",
+  "Log #{@log.id}" => "/logs/#{@log.id}",
```
jamesrose: Can we use routes here?

app/views/form/check_answers.html.erb
```diff
@@ -12,27 +12,27 @@
...
 
-    <%= form_with model: @lettings_log, method: "get" do |f| %>
+    <%= form_with model: @log, method: "get" do |f| %>
       <%= f.govuk_submit "Save and return to log" do %>
-        <% next_incomplete_section_redirect_path = @lettings_log.form.next_incomplete_section_redirect_path(subsection, @lettings_log) %>
-        <% if @lettings_log.status == "in_progress" && next_incomplete_section_redirect_path != "error" %>
-          <%= govuk_button_link_to "Save and go to next incomplete section", "/logs/#{@lettings_log.id}/#{next_incomplete_section_redirect_path}", secondary: true %>
+        <% next_incomplete_section_redirect_path = @log.form.next_incomplete_section_redirect_path(subsection, @log) %>
+        <% if @log.status == "in_progress" && next_incomplete_section_redirect_path != "error" %>
+          <%= govuk_button_link_to "Save and go to next incomplete section", "/logs/#{@log.id}/#{next_incomplete_section_redirect_path}", secondary: true %>
```
jamesrose: Can we use a route here?

app/views/form/page.html.erb
```diff
@@ -42,10 +42,10 @@
       <div class="govuk-button-group">
         <% if !@page.id.include?("value_check") && if request.query_parameters["referrer"] != "check_answers" %>
           <%= f.govuk_submit "Save and continue" %>
-          <%= govuk_link_to "Skip for now", send(@lettings_log.form.next_page_redirect_path(@page, @lettings_log, current_user), @lettings_log) %>
+          <%= govuk_link_to "Skip for now", send(@log.form.next_page_redirect_path(@page, @log, current_user), @log) %>
           <% else %>
           <%= f.govuk_submit "Save changes" %>
-            <%= govuk_link_to "Cancel", "/logs/#{@lettings_log.id}/setup/check-answers" %>
+            <%= govuk_link_to "Cancel", "/logs/#{@log.id}/setup/check-answers" %>
```
jamesrose: Can we use a route here?

app/models/form.rb
kosiakkatrina: Have a look now? Changed how we're creating the form to be a bit more like the lettings form - separated the setup section https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/858/commits/0ef59ce0bf35143113471d732133975d57c5e7ab

app/models/form.rb
kosiakkatrina: https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/858/commits/04075e99fea8183ae556143c8821774134cdee4c

app/models/sales_log.rb
kosiakkatrina: https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/858/commits/e27901f32d0416fc7dd20db0c58502ea0672c75b

app/models/sales_log.rb
kosiakkatrina: https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/858/commits/221e4593bc39f96c7660d83050d9a4baf72dd2dc

app/models/sales_log.rb
kosiakkatrina: https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/858/commits/e27901f32d0416fc7dd20db0c58502ea0672c75b

app/models/form.rb
```diff
@@ -7,21 +7,22 @@ class Form
...
-      # @form_definition = JSON.parse(File.open(form_path).read)
+      @setup_sections = [Form::Sales::Setup::Sections::Setup.new(nil, nil, self)]
       @form_sections = sections_in_form.map { |sec| sec.new(nil, nil, self) }
       @type = "sales"
-      @sections = form_sections
+      @sections = setup_sections + form_sections
       @subsections = sections.flat_map(&:subsections)
       @pages = subsections.flat_map(&:pages)
       @questions = pages.flat_map(&:questions)
       @start_date = Time.zone.local(name[0..3], 4, 1)
```
jamesrose, line 18: IMO we should replace `name` with the start year if that's how we would like to use it.

app/models/form_handler.rb
```diff
@@ -15,7 +15,7 @@ def current_form
   end
 
   def sales_forms
-    sales_sections = [Form::Sales::Setup::Sections::Setup]
+    sales_sections = [] # Add section classes here e.g. Form::Sales::Property::Sections::PropertyInformation
     { "2022_2023_sales" => Form.new(nil, "2022_2023_sales", sales_sections, "sales") }
```
jamesrose, line 19: How about?

```suggestion
    { :current => Form.new(nil, 2022, sales_sections, "sales") }
```

app/models/form.rb
kosiakkatrina: So that's a good point, I think the intention was to make it work the easiest way possible to unblock the rest of the sales tickets.
There was a conversation about structuring our the list of forms in a little bit different way within the form handler, something like this: [{type: "sales", start_year: "2022" }, {type: "lettings", start_year: "2021" }...] etc.
I think maybe not as part of this PR, but immediately after as part of this piece of work

baarkerlounger: Current, next and previous might be other good labels that don't need updating every year

app/models/form_handler.rb
baarkerlounger: `{ current: Form.new(nil, 2022, sales_sections, "sales") }`?

app/controllers/form_controller.rb
```diff
@@ -123,18 +128,21 @@ def is_referrer_check_answers?
...
-      page_ids = @lettings_log.form.subsection_for_page(@page).pages.map(&:id)
+      page_ids = @log.form.subsection_for_page(@page).pages.map(&:id)
       page_index = page_ids.index(@page.id)
-      next_page = @lettings_log.form.next_page(@page, @lettings_log, current_user)
-      previous_page = @lettings_log.form.previous_page(page_ids, page_index, @lettings_log, current_user)
+      next_page = @log.form.next_page(@page, @log, current_user)
+      previous_page = @log.form.previous_page(page_ids, page_index, @log, current_user)
       if next_page.to_s.include?("value_check") || next_page == previous_page
-        return "/logs/#{@lettings_log.id}/#{next_page.dasherize}?referrer=check_answers"
+        binding.pry
```
baarkerlounger: :rotating_light:

app/models/lettings_log.rb
```diff
@@ -531,16 +520,6 @@ def unittype_sh
...
-    self.status = if all_fields_completed? && errors.empty?
-                    "completed"
-                  elsif all_fields_nil?
-                    "not_started"
-                  else
-                    "in_progress"
-                  end
-  end
-
   def reset_not_routed_questions
```
baarkerlounger, line 508: Shouldn't this be common to both as well? (It's more related to how the form works than specific log data, so perhaps should actually live in the form model anyway...)

app/views/form/page.html.erb
Ted-U: Sorted now, we have added a route as suggested.

spec/models/user_spec.rb
```diff
@@ -269,4 +269,37 @@
...
+        :lettings_log,
+        :completed,
+        owning_organisation: user.organisation,
+        managing_organisation: user.organisation,
+        created_by: user,
+      )
+    end
+
+    context "when the user is deleted" do
+      it "owned lettings logs are not deleted as a result" do
```
baarkerlounger, line 293: Worth adding a sales log version of this too?

spec/models/user_spec.rb
```diff
@@ -269,4 +269,37 @@
...
+        expect(described_class.count).to eq(1)
+        expect(LettingsLog.count).to eq(1)
+        user.destroy!
+        expect(described_class.count).to eq(0)
+        expect(LettingsLog.count).to eq(1)
+      end
+
+      it "owned lettings logs are not deleted as a result" do
```
baarkerlounger: This looks like it's testing that they are correctly deleted when the organisation is deleted? In that case might this test be better in the Organisation spec? (Although I think one maybe already exists)

app/models/log.rb
```diff
@@ -4,11 +4,25 @@ class Log < ApplicationRecord
...
+  scope :filter_by_status, ->(status, _user = nil) { where status: }
+  scope :filter_by_years, lambda { |years, _user = nil|
+    first_year = years.shift
+    query = filter_by_year(first_year)
+    years.each { |year| query = query.or(filter_by_year(year)) }
+    query.all
+  }
+  scope :filter_by_id, ->(id) { where(id:) }
+  scope :filter_by_user, lambda { |selected_user, user|
+    if !selected_user.include?("all") && user.present?
```
jamesrose, line 22: What is `all` and why do we need to filter it out?

app/views/logs/edit.html.erb
```diff
@@ -1,6 +1,6 @@
-<% content_for :title, "Log #{@lettings_log.id}" %>
+<% content_for :title, "Log #{@log.id}" %>
 <% content_for :breadcrumbs, govuk_breadcrumbs(breadcrumbs: {
-  "Logs" => "/logs",
+  "Logs" => (@log.is_a?(SalesLog) ? "/sales-logs" : "/lettings-logs").to_s,
```
jamesrose: Could you put this in a helper and use routes?

app/components/log_summary_component.html.erb
```diff
@@ -3,11 +3,11 @@
     <div class="govuk-grid-column-one-half">
       <header class="app-log-summary__header">
         <h2 class="app-log-summary__title">
-          <%= govuk_link_to lettings_log_path(log) do %>
+          <%= govuk_link_to log.is_a?(LettingsLog) ? lettings_log_path(log) : sales_log_path(log) do %>
```
jamesrose: I'd look at adding a `lettings?` method to the base class that defaults to false. Then you can add one to `LettingsLog` that is true:

```suggestion
          <%= govuk_link_to log.lettings? ? lettings_log_path(log) : sales_log_path(log) do %>
```

app/controllers/form_controller.rb
```diff
@@ -109,11 +106,19 @@ def responses_for_page(page)
...
+    @log = if params.key?("sales_log")
+             current_user.sales_logs.find_by(id: params[:id])
+           else
+             current_user.lettings_logs.find_by(id: params[:id])
+           end
   end
 
   def find_resource_by_named_id
-    @lettings_log = current_user.lettings_logs.find_by(id: params[:lettings_log_id])
+    @log = if params["sales_log_id"].present?
```
jamesrose:
```suggestion
    @log = if params[:sales_log_id].present?
```

Does this need to be a string?

app/controllers/organisations_controller.rb
```diff
@@ -87,30 +87,54 @@ def update
...
-          send_data byte_order_mark + unpaginated_filtered_logs.to_csv, filename: "logs-#{@organisation.name}-#{Time.zone.now}.csv"
+          send_data byte_order_mark + unpaginated_filtered_logs.to_csv, filename: "lettings-logs-#{@organisation.name}-#{Time.zone.now}.csv"
         end
       end
     else
       redirect_to(lettings_logs_path)
     end
   end
 
+  def sales_logs
```
jamesrose, line 119: https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/851 is about to be merged that changes this. Do you want to revise?

app/views/form/check_answers.html.erb
```diff
@@ -12,27 +12,27 @@
...
 
-    <%= form_with model: @lettings_log, method: "get" do |f| %>
+    <%= form_with model: @log, method: "get" do |f| %>
       <%= f.govuk_submit "Save and return to log" do %>
-        <% next_incomplete_section_redirect_path = @lettings_log.form.next_incomplete_section_redirect_path(subsection, @lettings_log) %>
-        <% if @lettings_log.status == "in_progress" && next_incomplete_section_redirect_path != "error" %>
-          <%= govuk_button_link_to "Save and go to next incomplete section", "/logs/#{@lettings_log.id}/#{next_incomplete_section_redirect_path}", secondary: true %>
+        <% next_incomplete_section_redirect_path = @log.form.next_incomplete_section_redirect_path(subsection, @log) %>
+        <% if @log.status == "in_progress" && next_incomplete_section_redirect_path != "error" %>
+          <%= govuk_button_link_to "Save and go to next incomplete section", send("#{@log.class.name.underscore}_#{next_incomplete_section_redirect_path.underscore.tr('/', '_')}_path", @log), secondary: true %>
```
jamesrose: 😬 

Can you just add a helper here?

app/views/form/page.html.erb
```diff
@@ -42,10 +42,10 @@
       <div class="govuk-button-group">
         <% if !@page.id.include?("value_check") && if request.query_parameters["referrer"] != "check_answers" %>
           <%= f.govuk_submit "Save and continue" %>
-          <%= govuk_link_to "Skip for now", send(@lettings_log.form.next_page_redirect_path(@page, @lettings_log, current_user), @lettings_log) %>
+          <%= govuk_link_to "Skip for now", send(@log.form.next_page_redirect_path(@page, @log, current_user), @log) %>
           <% else %>
           <%= f.govuk_submit "Save changes" %>
-            <%= govuk_link_to "Cancel", "/logs/#{@lettings_log.id}/setup/check-answers" %>
+            <%= govuk_link_to "Cancel", send("#{@log.class.name.underscore}_#{@page.subsection.id}_check_answers_path", @log) %>
```
jamesrose: This path should be in a helper.

app/views/form/review.html.erb
```diff
@@ -1,7 +1,7 @@
 <% content_for :title, "Review lettings log" %>
 <% content_for :breadcrumbs, govuk_breadcrumbs(breadcrumbs: {
   "Logs" => "/logs",
-  "Log #{@lettings_log.id}" => "/logs/#{@lettings_log.id}",
+  "Log #{@log.id}" => "/logs/#{@log.id}",
```
jamesrose, line 4: Please can you change these both to routes?

app/views/layouts/application.html.erb
```diff
@@ -34,7 +34,7 @@
       </script>
     <% end %>
 
-    <% if Rails.env.production? %>
+    <% if Rails.env.production? && ENV["APP_HOST"].present? %>
```
jamesrose, line 37: Why was this added?

app/models/log.rb
kosiakkatrina: It's a filter. We can select either "All" or "Yours" when we're filtering the logs by user. This is not a new implementation, just moved it to parent class

app/views/logs/edit.html.erb
```diff
@@ -10,10 +10,10 @@
       <%= content_for(:title) %>
     </h1>
 
-    <% if @lettings_log.status == "in_progress" %>
-      <p class="govuk-body govuk-!-margin-bottom-7"><%= get_subsections_count(@lettings_log, :completed) %> of <%= get_subsections_count(@lettings_log, :all) %> sections completed.</p>
+    <% if @log.status == "in_progress" %>
```
jamesrose, line 13:
```suggestion
    <% if @log.in_progress? %>
```

I believe this is an enum.

app/views/logs/edit.html.erb
```diff
@@ -22,14 +22,14 @@
           </a>
         <% end %>
       </p>
-    <% elsif @lettings_log.status == "not_started" %>
+    <% elsif @log.status == "not_started" %>
```
jamesrose, line 25:
```suggestion
    <% elsif @log.not_started? %>
```

app/views/logs/edit.html.erb
```diff
@@ -22,14 +22,14 @@
...
       <p class="govuk-body">This log has not been started.</p>
-    <% elsif @lettings_log.status == "completed" %>
+    <% elsif @log.status == "completed" %>
```
jamesrose, line 27:
```suggestion
    <% elsif @log.completed? %>
```

app/views/logs/edit.html.erb
```diff
@@ -22,14 +22,14 @@
...
       <p class="govuk-body">
-        <%= status_tag(@lettings_log.status) %>
+        <%= status_tag(@log.status) %>
       </p>
       <p class="govuk-body">
-        You can <%= govuk_link_to "review and make changes to this log", "/logs/#{@lettings_log.id}/review" %> until 2nd June <%= @lettings_log.collection_start_year.present? ? @lettings_log.collection_start_year + 1 : "" %>.
+        You can <%= govuk_link_to "review and make changes to this log", "/lettings-logs/#{@log.id}/review" %> until 2nd June <%= @log.collection_start_year.present? ? @log.collection_start_year + 1 : "" %>.
```
jamesrose, line 32: This should be a route

app/views/organisations/_organisation_list.html.erb
```diff
@@ -22,7 +22,7 @@
           <% row.cell(header: true, html_attributes: {
             scope: "row",
           }) do %>
-            <%= govuk_link_to(organisation.name, "organisations/#{organisation.id}/logs") %>
+            <%= govuk_link_to(organisation.name, "organisations/#{organisation.id}/lettings-logs") %>
```
jamesrose, line 25: This should be a route.

app/components/log_summary_component.html.erb
```diff
@@ -3,11 +3,11 @@
...
             <span class="govuk-visually-hidden">Log </span><%= log.id %>
           <% end %>
         </h2>
-        <% if log.tenancycode? or log.propcode? %>
+        <% if log.is_a?(LettingsLog) && (log.tenancycode? or log.propcode?) %>
```
jamesrose: You can reuse the method above here as well

app/components/log_summary_component.html.erb
```diff
@@ -25,7 +25,7 @@
         <% end %>
       </header>
 
-      <% if log.needstype? or log.startdate? %>
+      <% if log.is_a?(LettingsLog) && (log.needstype? or log.startdate?) %>
```
jamesrose, line 28: You can reuse the method above here as well

app/views/layouts/application.html.erb
moarpheus: Stops you from mocking prod env in tests in Feature flags otherwise

baarkerlounger: Can't we just provide a mock value in tests? Seems cleaner than changing the actual code here for something completely unrelated.

jamesrose (approved):
Nice work!

PR #859, opened 2022-09-02
Check if check_answers_card_number is present

kosiakkatrina:


baarkerlounger (approved):
Happy for this to go in for now since I think it'll solve the immediate issue but I think if we want to support questions not specifying a check answers card number we should be explicit about that with a test.

We also should remove the hard coding of the card titles from this component and move that to the definition as well so that this component is actually reusable.

PR #860, opened 2022-09-06
CLDC-1324 remove not started status

moarpheus:


app/controllers/form_controller.rb
```diff
@@ -113,7 +113,11 @@ def find_resource
   end
 
   def find_resource_by_named_id
-    @lettings_log = current_user.lettings_logs.find_by(id: params[:lettings_log_id])
+    if request.path.include? "new"
```
jamesrose: It's quite unusual to do this in Rails. Is there anything that stops us from returning a new object if we can't find a letting log?

app/helpers/tasklist_helper.rb
```diff
@@ -17,7 +17,11 @@ def next_page_or_check_answers(subsection, lettings_log, current_user)
            else
              "lettings_log_#{next_question_page(subsection, lettings_log, current_user)}_path"
            end
-    send(path, lettings_log)
+    if lettings_log.id
+      send(path, lettings_log)
+    else
+      "/logs/new/#{next_question_page(subsection, lettings_log, current_user)}"
```
jamesrose, line 23: We should use the route here.

config/routes.rb
```diff
@@ -75,8 +75,11 @@
 
   resources :lettings_logs, path: "/logs" do
     collection do
+      get "new-log", to: "lettings_logs#create"
+      get "new", to: "lettings_logs#show"
       post "bulk-upload", to: "bulk_upload#bulk_upload"
       get "bulk-upload", to: "bulk_upload#show"
+      get "new/organisation", to: "form#organisation"
```
baarkerlounger: Is this needed/related to the logs stuff?

app/helpers/tasklist_helper.rb
moarpheus: I'll refactor and prettify at the end, mainly focusing on getting this spike far enough to see if this would work.

app/controllers/form_controller.rb
moarpheus: There actually is. We want to differentiate between:
1. found lettings log
2. not found log ( render 404 ) 
3. not found log - but arriving here from the new page

config/routes.rb
moarpheus: I haven't implemented code to dynamically generate all possible routes yet - just getting just enough for a spike

app/controllers/form_controller.rb
```diff
@@ -60,6 +60,20 @@ def review
           render_not_found
         end
       end
+      define_method("new_log_#{page.id}") do |_errors = {}|
```
baarkerlounger, line 63: Doesn't need to be done as part of this PR but worth considering whether this needs to be a dynamically defined set of methods at all (think @jamesrose pointed this out previously) - currently it pretty much does the same thing for every page it's defined for anyway so could just be a single method they all route to with page id as a param.

PR #861, opened 2022-09-06
Cldc 1390 la lookup fallback

natdeanlewissoftwire:
I've made a few decisions on this as some designs were missing, approved by Niki but will mention here as well for clarity:

The local authority fallback page is a separate page to the main edit/create location page, as the postcodes api call only runs when the save and continue button is clicked on the main page. The fallback page looks like this: 
image
Both skip and back return the user to the locations selection page.
If a user has selected to add another location on the create location page, this is passed through via a url query string on the fallback page (e.g. ...schemes/6/locations/67/edit-local-authority?add_another_location=Yes)
Once a user has selected a local authority there is no option to remove it, only to change it. I think this is fine as it follows other patterns on the site I think but happy to add this function (or alternatively add a validator that forces users to add a local authority, at the moment they can skip and still save locations without this being required).
On the check your answers page, not yet saved locations are links back to the edit-local-authority page (with the text "Select local authority" if none has been added). Saved locations' local authorities are just text and no longer editable by the user through links on the site.: 
image

validation for local authorities is that they must be from the list provided and not "Select an option" or nil, the error message is here: 
image


app/views/locations/edit_local_authority.html.erb
```diff
@@ -0,0 +1,35 @@
...
...land", "South Kesteven", "South Lakeland", "South Norfolk", "South Northamptonshire", "South Oxfordshire", "South Ribble", "South Somerset", "South Staffordshire", "South Tyneside", "Southampton", "Southend-on-Sea", "Southwark", "Spelthorne", "St Albans", "St. Helens", "Stafford", "Staffordshire Moorlands", "Stevenage", "Stockport", "Stockton-on-Tees", "Stoke-on-Trent", "Stratford-on-Avon", "Stroud", "Sunderland", "Surrey Heath", "Sutton", "Swale", "Swindon", "Tameside", "Tamworth", "Tandridge", "Teignbridge", "Telford and Wrekin", "Tendring", "Test Valley","Tewkesbury", "Thanet", "Three Rivers", "Thurrock", "Tonbridge and Malling", "Torbay", "Torridge", "Tower Hamlets", "Trafford", "Tunbridge Wells", "Uttlesford", "Vale of White Horse", "Wakefield", "Walsall", "Waltham Forest", "Wandsworth", "Warrington", "Warwick", "Watford", "Waverley", "Wealden", "Wellingborough", "Welwyn Hatfield", "West Berkshire", "West Devon", "West Lancashire", "West Lindsey", "West Oxfordshire", "West Suffolk", "Westminster", "Wigan", "Wiltshire", "Winchester", "Windsor and Maidenhead", "Wirral", "Woking", "Wokingham", "Wolverhampton", "Worcester", "Worthing", "Wychavon", "Wyre", "Wyre Forest", "York"] %>
```
natdeanlewissoftwire: This is the same list used in the json forms -  would it be worth centralising this somewhere?

baarkerlounger: Yep, I think you could currently get it with something like `FormHandler.instance.current_form.get_question("la").answer_options.values` which would avoid duplicating the values but admittedly isn't very elegant either. I suppose if the JSON forms are refactored away it'll be easier to just store this somewhere more dedicated.

natdeanlewissoftwire: @baarkerlounger Okay, I've used that reference, thanks - I put in some validation to ensure the user doesn't submit "Select an option" or nil as the local authority. They can still skip though, is it a requirement that they enter a local authority or can we allow them to skip (or click back) and submit without one?

spec/request_helper.rb
```diff
@@ -8,6 +8,12 @@ def self.stub_http_requests
 
     WebMock.stub_request(:get, "https://api.postcodes.io/postcodes/AA11AA")
            .to_return(status: 200, body: "{\"status\":200,\"result\":{\"postcode\":\"AA1 1AA\",\"admin_district\":\"Westminster\",\"codes\":{\"admin_district\":\"E09000033\"}}}", headers: {})
+    WebMock.stub_request(:get, "https://api.postcodes.io/postcodes/AA12AA")
+           .to_return(status: 200, body: "{\"status\":200,\"result\":{\"postcode\":\"AA1 2AA\",\"admin_district\":\"Westminster\",\"codes\":{\"admin_district\":\"E09000033\"}}}", headers: {})
+    WebMock.stub_request(:get, "https://api.postcodes.io/postcodes/NW1L5DP")
+           .to_return(status: 200, body: "{\"status\":200,\"result\":{\"postcode\":\"NW1L 5DP\",\"admin_district\":\"Westminster\",\"codes\":{\"admin_district\":\"E09000033\"}}}", headers: {})
+    WebMock.stub_request(:get, "https://api.postcodes.io/postcodes/ZZ11ZZ")
+           .to_return(status: 200, body: "{\"status\":200,\"result\":{\"postcode\":\"ZZ1 1ZZ\",\"admin_district\":\"Westminster\",\"codes\":{\"admin_district\":\"E09000033\"}}}", headers: {})
```
natdeanlewissoftwire, line 16: These are just to make existing tests pass - should I add new tests for the edit local authority page or should that be a separate ticket?

natdeanlewissoftwire: Have added tests to this ticket

baarkerlounger: Wonder if it would be better to have these with the specific tests they're used in, since I guess they're not generic?

spec/requests/locations_controller_spec.rb
```diff
@@ -264,6 +269,29 @@
           expect(page).to have_content("Check your answers before creating this scheme")
         end
       end
+
+      context "when postcodes.io doesn't return a la_code" do
+        let!(:params) do
```
baarkerlounger: Don't think you need the bang here since you use params in the request?

spec/requests/locations_controller_spec.rb
```diff
@@ -414,6 +442,29 @@
...
```
baarkerlounger: Same

baarkerlounger (approved):
LGTM

spec/requests/locations_controller_spec.rb
natdeanlewissoftwire: Ah, thanks!

PR #864, opened 2022-09-09
CLDC-1553 Added migration to change referral values of 0 to 1 for lettings logs

davidm-m:
Tested this locally, the migration ran without issues and all my lettings logs with 0 for referral now have 1.

### DIFF
--- a/db/migrate/20220909120838_correct_referral_value.rb
+++ b/db/migrate/20220909120838_correct_referral_value.rb
@@ -1,5 +1,0 @@
-class CorrectReferralValue < ActiveRecord::Migration[7.0]
-  def change
-    LettingsLog.where(referral: 0).update_all(referral: 1)
-  end
-end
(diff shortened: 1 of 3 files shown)


db/migrate/20220909120838_correct_referral_value.rb
```diff
@@ -0,0 +1,5 @@
+class CorrectReferralValue < ActiveRecord::Migration[7.0]
+  def change
+    LettingsLog.where(referral: 0).update_all(referral: 1)
```
baarkerlounger: Is this reversible? 

I think it's generally best practice to only use migrations for schema changes, not for data changes - not sure it really matters hugely in this case but I think using rake tasks for data changes is a good idea.

davidm-m: Good point! I don't know much about Rails so didn't realise this was the sort of thing you could do with a rake task. I've changed it to one since it's not much effort.

db/schema.rb
```diff
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema[7.0].define(version: 2022_09_02_082245) do
+ActiveRecord::Schema[7.0].define(version: 2022_09_09_120838) do
```
baarkerlounger: This PR probably shouldn't change the schema since there's no migration in it now

baarkerlounger (commented):
Aside from the schema change :+1:

PR #866, opened 2022-09-13
CLDC-1183-assistive-tech

Ted-U:


app/views/organisations/users.html.erb
```diff
@@ -9,7 +9,7 @@
   <%= render SubNavigationComponent.new(
     items: secondary_items(request.path, @organisation.id),
   ) %>
-  <h2 class="govuk-visually-hidden">Users</h2>
+  <h2 class="govuk-visually-hidden"> </h2>
```
baarkerlounger, line 12: Why remove the string but leave the empty header in?

baarkerlounger:
I'm not really sure this fixes https://digital.dclg.gov.uk/jira/browse/CLDC-1183 (though I may have misunderstood). 

I think the empty link being referred to was when we didn't have a name/email for a user so there was nothing to properly read out. The more general issue with dragon looks like it's an upstream one https://github.com/alphagov/govuk-design-system/issues/2187.

Ted-U:
> I'm not really sure this fixes https://digital.dclg.gov.uk/jira/browse/CLDC-1183 (though I may have misunderstood).
> 
> I think the empty link being referred to was when we didn't have a name/email for a user so there was nothing to properly read out. The more general issue with dragon looks like it's an upstream one [alphagov/govuk-design-system#2187](https://github.com/alphagov/govuk-design-system/issues/2187).

Agreed

PR #867, opened 2022-09-13
Configure new Redis instances for Sidekiq and Rack::Attack

jamesrose:
Previously we had one instance called `dluhc-core-{environment}-redis-rate-limit` that was used exclusively for `Rack::Attack`.

This PR configures Sidekiq and Rack::Attack to use a new generically named instance called `dluhc-core-{environment}-redis` on `staging` and `production`. 

These new Redis instances have already been provisioned on each environment. Staging is a `micro-6.x` and production is `micro-ha-6.x`.

Later, once this PR has been deployed, I will remove `dluhc-core-{environment}-redis-rate-limit` instances.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -20,3 +20,10 @@ AllCops:
 
 Style/Documentation:
   Enabled: false
+
+Rails/UnknownEnv:
+  Environments:
+    - production
+    - staging
+    - development
+    - test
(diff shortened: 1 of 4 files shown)


baarkerlounger (approved):

.rubocop.yml
```diff
@@ -20,3 +20,10 @@ AllCops:
 
 Style/Documentation:
   Enabled: false
+
+Rails/UnknownEnv:
```
baarkerlounger, line 24: Why is this needed?

jamesrose: @baarkerlounger Rubocop complains that `Rails.env.staging?` isn't a valid environment if you aren't explicit about them here.

PR #873, opened 2022-09-15
Refactor form namespacing

baarkerlounger:
Refactor form namespacing so that's easier to see what is specific to lettings/sales and what is common as well as making it easier to compose new forms.

image


jamesrose (approved):
Nice! Is it worth thinking about how we might namespace this in the context of a future form running concurrently?

baarkerlounger:
@jamesrose I'm not 100% sure if it makes more sense to add additional (current, next, previous or year type) namespaces for that or if the delta is small enough that it makes more sense to have code switches in just the classes that change to have different year on year behaviour.

PR #875, opened 2022-09-15
Cldc 1600 refactor form handler

kosiakkatrina:
Add previous, current and next forms to form handler and use them across the codebase
Structure forms in FormHandler to contain type, start year and the Form
Remove name from Form initializer and add an optional start year

### DIFF
--- a/app/models/form.rb
+++ b/app/models/form.rb
@@ -3,9 +3,11 @@ class Form
               :start_date, :end_date, :type, :name, :setup_definition,
               :setup_sections, :form_sections
 
+  include Form::Setup
+
   def initialize(form_path, name, sections_in_form = [], type = "lettings")
     if type == "sales"
-      @name = name
+      @name = "#{start_year}_#{start_year + 1}_sales"
       @setup_sections = [Form::Sales::Sections::Setup.new(nil, nil, self)]
       @form_sections = sections_in_form.map { |sec| sec.new(nil, nil, self) }
       @type = "sales"
@@ -13,8 +15,8 @@ class Form
       @subsections = sections.flat_map(&:subsections)
       @pages = subsections.flat_map(&:pages)
       @questions = pages.flat_map(&:questions)
-      @start_date = Time.zone.local(name[0..3], 4, 1)
-      @end_date = Time.zone.local(start_date.year + 1, 7, 1)
+      @start_date = Time.zone.local(start_year, 4, 1)
+      @end_date = Time.zone.local(start_year + 1, 7, 1)
       @form_definition = {
         "form_type" => type,
         "start_date" => start_date,
@@ -24,7 +26,6 @@ class Form
     else
       raise "No form definition file exists for given year".freeze unless File.exist?(form_path)
 
-      @name = name
       @setup_sections = [Form::Lettings::Sections::Setup.new(nil, nil, self)]
       @form_definition = JSON.parse(File.open(form_path).read)
       @form_sections = form_definition["sections"].map { |id, s| Form::Section.new(id, s, self) }
@@ -35,6 +36,7 @@ class Form
       @questions = pages.flat_map(&:questions)
       @start_date = Time.iso8601(form_definition["start_date"])
       @end_date = Time.iso8601(form_definition["end_date"])
+      @name = "#{start_date.year}_#{end_date.year}"
     end
   end
 
--- a/app/models/form.rb
+++ b/app/models/form.rb
@@ -5,7 +5,7 @@ class Form
 
   include Form::Setup
 
-  def initialize(form_path, name, sections_in_form = [], type = "lettings")
+  def initialize(form_path, start_year = "", sections_in_form = [], type = "lettings")
     if type == "sales"
       @name = "#{start_year}_#{start_year + 1}_sales"
       @setup_sections = [Form::Sales::Sections::Setup.new(nil, nil, self)]
--- a/app/models/form.rb
+++ b/app/models/form.rb
@@ -5,7 +5,6 @@ class Form
 
   def initialize(form_path, start_year = "", sections_in_form = [], type = "lettings")
     if type == "sales"
-      @name = "#{start_year}_#{start_year + 1}_sales"
       @setup_sections = [Form::Sales::Sections::Setup.new(nil, nil, self)]
       @form_sections = sections_in_form.map { |sec| sec.new(nil, nil, self) }
       @type = "sales"
@@ -34,8 +33,8 @@ class Form
       @questions = pages.flat_map(&:questions)
       @start_date = Time.iso8601(form_definition["start_date"])
       @end_date = Time.iso8601(form_definition["end_date"])
-      @name = "#{start_date.year}_#{end_date.year}"
     end
+    @name = "#{start_date.year}_#{end_date.year}_#{type}"
   end
 
   def get_subsection(id)
--- a/app/models/form.rb
+++ b/app/models/form.rb
@@ -3,8 +3,6 @@ class Form
               :start_date, :end_date, :type, :name, :setup_definition,
               :setup_sections, :form_sections
 
-  include Form::Setup
-
   def initialize(form_path, start_year = "", sections_in_form = [], type = "lettings")
     if type == "sales"
       @name = "#{start_year}_#{start_year + 1}_sales"
--- a/app/models/form_handler.rb
+++ b/app/models/form_handler.rb
@@ -7,7 +7,7 @@ class FormHandler
   end
 
   def get_form(form)
-    @forms[form].present? ? @forms[form]["form"] : nil
+    @forms[form]["form"] if @forms[form].present?
   end
 
   def current_lettings_form
--- a/app/models/form_handler.rb
+++ b/app/models/form_handler.rb
@@ -24,13 +24,27 @@ class FormHandler
     directories.each do |directory|
       Dir.glob("#{directory}/*.json").each do |form_path|
         form_name = File.basename(form_path, ".json")
-        forms[form_name] = Form.new(form_path, form_name)
+        form = Form.new(form_path, form_name)
+        forms[form_name] = form
+        if form.start_date.year + 1 == current_collection_start_year && forms["previous_lettings"].blank?
+          forms["previous_lettings"] = form
+        elsif form.start_date.year == current_collection_start_year && forms["current_lettings"].blank?
+          forms["current_lettings"] = form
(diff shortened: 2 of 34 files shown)


app/models/sales_log.rb
```diff
@@ -21,13 +21,13 @@ def self.editable_fields
   end
 
   def form_name
-    return unless saledate
+    return unless startdate
```
baarkerlounger: Isn't the field called `saledate` on sales logs still?

config/routes.rb
```diff
@@ -93,23 +93,23 @@
     end
 
     FormHandler.instance.lettings_forms.each do |_key, form|
-      form.pages.map do |page|
+      form["form"].pages.map do |page|
```
baarkerlounger: `form["form"]` seems a little strange...

app/services/csv/lettings_log_csv_service.rb
```diff
@@ -51,7 +51,13 @@ def set_csv_attributes
 
     def ordered_form_questions
       downloaded_form_years = LettingsLog.all.map(&:collection_start_year).uniq.compact
-      downloaded_form_fields = downloaded_form_years.count == 1 && downloaded_form_years[0].present? ? FormHandler.instance.get_form("#{downloaded_form_years[0]}_#{downloaded_form_years[0] + 1}").questions : FormHandler.instance.forms.first.second.questions
+
+      if downloaded_form_years.count == 1 && downloaded_form_years[0].present?
+        form_name = FormHandler.instance.form_name_from_start_year(downloaded_form_years[0], "lettings")
+        downloaded_form_fields = FormHandler.instance.get_form(form_name).questions
+      else
+        downloaded_form_fields = FormHandler.instance.current_lettings_form.questions
```
baarkerlounger, line 59: I wonder if this needs to consider the case where a question has been dropped from the current form but is still there on the previous. You'd expect it to still be there in the download for those logs I think? In which case we might need a combined question set? cc. @CheshireSwift

app/models/sales_log.rb
kosiakkatrina: Tyvm, it didn't fail cause there was a function that returns saledate for startdate! Will change it here tho

spec/models/form_handler_spec.rb
```diff
@@ -1,43 +1,85 @@
 require "rails_helper"
 
 RSpec.describe FormHandler do
-  let(:test_form_name) { "2021_2022" }
+  let(:form_handler) { described_class.instance }
+
+  before { Singleton.__init__(described_class) }
```
baarkerlounger: What does this do/why is it needed?

app/models/form_handler.rb
```diff
@@ -7,29 +7,50 @@ def initialize
...
+        lettings_form_definition = { "form" => form, "type" => "lettings", "start_year" => form.start_date.year }
+
+        form_to_set = form_name_from_start_year(form.start_date.year, "lettings")
+        forms[form_to_set] = lettings_form_definition if forms[form_to_set].blank?
       end
     end
     forms
   end
 
+  def current_collection_start_year
```
baarkerlounger, line 42: This maybe justifies it's own unit test? Unless there is one and I've missed it...

config/routes.rb
kosiakkatrina: Yeah I was thinking about this...
For me it felt like if we were to access a specific form by `FormHandler.instance.get_form(_)` that would return just the form object and if we call `FormHandler.instance.forms[0]` that would return the hash with form, type and start year.
We could potentially have something similar where `lettings_forms` returns what it currently does and perhaps `get_lettings_forms` could only return an array of form objects (?) not sure how useful all those methods would be tho

app/services/csv/lettings_log_csv_service.rb
kosiakkatrina: So I think this is only used for ordering, the field should still get downloaded as long as it is saved as a database field. It might just be at the end

app/models/form.rb
```diff
@@ -35,6 +34,7 @@ def initialize(form_path, name, sections_in_form = [], type = "lettings")
       @questions = pages.flat_map(&:questions)
       @start_date = Time.iso8601(form_definition["start_date"])
       @end_date = Time.iso8601(form_definition["end_date"])
+      @name = "#{start_date.year}_#{end_date.year}"
```
CheshireSwift: Minor point, but I think it'd be good to have fields in the same order in both branches if possible (i.e. name first, etc.)

spec/models/form_handler_spec.rb
kosiakkatrina: When we change time using timecop we need to reload the FormHandler instance because current_form would have to change to a different one in a different year and we're only initialising the forms at the very start I think.
This resets the the FormHandler

config/routes.rb
baarkerlounger: I think what feels off is that we're sort of referring to two different object types/data structures (one a super set of the other) and calling them both "form", which might be confusing.

app/models/form.rb
kosiakkatrina: Hm, I thought that, the reason this is at the bottom is because we're using the fields that get set on lines 35/36. I could move the sales name to the bottom tho

spec/models/form_handler_spec.rb
baarkerlounger: Cool. Worth a code comment maybe? Or think it's obvious enough?

app/models/form.rb
CheshireSwift: Ah, yeah, missed that (was conflating `start_year` and `start_date.year`). I think it's okay as-is, given that 👍

spec/models/form_handler_spec.rb
kosiakkatrina: I wouldn't have known what this does if I hadn't written it, so happy to add a comment

app/models/form.rb
```diff
@@ -3,18 +3,18 @@ class Form
               :start_date, :end_date, :type, :name, :setup_definition,
               :setup_sections, :form_sections
 
-  def initialize(form_path, name, sections_in_form = [], type = "lettings")
+  def initialize(form_path, start_year = "", sections_in_form = [], type = "lettings")
```
jamesrose, line 6:
```suggestion
  def initialize(form_path, start_year, sections_in_form = [], type = "lettings")
```

Is `start_year` optional?

app/models/form.rb
```diff
@@ -3,18 +3,18 @@ class Form
...
     if type == "sales"
-      @name = name
+      @name = "#{start_year}_#{start_year + 1}_sales"
```
```suggestion
      @name = "#{start_year}_#{start_year + 1}_#{type}"
```

You can probably shift this out if the `if` condition here because it should be shared between types.

app/models/form_handler.rb
```diff
@@ -7,29 +7,50 @@ def initialize
...
+  def current_sales_form
+    forms["current_sales"]["form"]
   end
 
   def sales_forms
     sales_sections = [] # Add section classes here e.g. Form::Sales::Property::Sections::PropertyInformation
-    { "2022_2023_sales" => Form.new(nil, "2022_2023_sales", sales_sections, "sales") }
+    current_form = Form.new(nil, current_collection_start_year, sales_sections, "sales")
+    previous_form = Form.new(nil, current_collection_start_year - 1, sales_sections, "sales")
+    { "current_sales" => { "form" => current_form, "type" => "sales", "start_year" => current_form.start_date.year },
```
jamesrose: I wonder if we need `type` and `start_year` here as keys if both can be inferred from the form itself. It might simplify this to:

```suggestion
    {
      "current_sales" => Form.new(nil, current_collection_start_year, sales_sections, "sales"), 
      "previous_sales" => Form.new(nil, current_collection_start_year - 1, sales_sections, "sales"), 
    }
```

What do you think?

app/models/form.rb
kosiakkatrina: I think so. We get the start year for lettings form from the JSON file at the moment, so we wouldn't have it before we create the lettings form (line 35, same file)

app/models/form_handler.rb
kosiakkatrina: I mean, it does make loads of sense, cause we can get the year and the type from the form itself. 
Let me give this a go

kosiakkatrina: Oh lol, sorry, didn't see your actual comment, but I agree 😅

app/models/form_handler.rb
```diff
@@ -10,26 +10,46 @@ def get_form(form)
...
+  def current_sales_form
+    forms["current_sales"]
   end
 
   def sales_forms
     sales_sections = [] # Add section classes here e.g. Form::Sales::Property::Sections::PropertyInformation
-    { "2022_2023_sales" => Form.new(nil, "2022_2023_sales", sales_sections, "sales") }
+    current_form = Form.new(nil, current_collection_start_year, sales_sections, "sales")
+    previous_form = Form.new(nil, current_collection_start_year - 1, sales_sections, "sales")
+    { "current_sales" => current_form,
```
jamesrose, line 25: 👌

app/models/form.rb
jamesrose: Ah, that makes sense, because we need to support a manual initialization _and_ taking the year from the form. Thanks!

app/services/csv/lettings_log_csv_service.rb
```diff
@@ -51,7 +51,13 @@ def set_csv_attributes
 
     def ordered_form_questions
       downloaded_form_years = LettingsLog.all.map(&:collection_start_year).uniq.compact
-      downloaded_form_fields = downloaded_form_years.count == 1 && downloaded_form_years[0].present? ? FormHandler.instance.get_form("#{downloaded_form_years[0]}_#{downloaded_form_years[0] + 1}").questions : FormHandler.instance.forms.first.second.questions
+
+      if downloaded_form_years.count == 1 && downloaded_form_years[0].present?
```
jamesrose, line 55: It's not super clear what's going on here, but I can see you haven't changed it materially in this PR, so let's come back to it!

jamesrose (approved):

app/services/csv/lettings_log_csv_service.rb
kosiakkatrina: I think the purpose of this part is to order the columns in csv file in the same order as the form questions in the UI.
This if statement is choosing which form to use to get the questions order. If the csv is for records from only one year (for example, the list was filtered down to one year) it will choose the form for that year. Otherwise it will choose a default form. It's not the clearest, so refactoring this might be a good idea

PR #877, opened 2022-09-15
Sale completion date

baarkerlounger:
https://digital.dclg.gov.uk/jira/browse/CLDC-1472
https://digital.dclg.gov.uk/jira/browse/CLDC-1464

image


jamesrose (approved):

kosiakkatrina (commented):
There is a saledate column in the db and a few methods in the model that use it (for example to filter by year) could we remove/change those?

baarkerlounger:
@kosiakkatrina yes that column hasn't changed and is correct, this changes the page id (for the url) not the question id

PR #878, opened 2022-09-16
Lower case logs

baarkerlounger:


jamesrose (approved):
I've approved this, but you're blocked on a test failure.

PR #881, opened 2022-09-16
Cldc 1475 shared ownership scheme type

kosiakkatrina:
Add shared ownership scheme type question and page


app/models/form/sales/questions/shared_ownership_type.rb
```diff
@@ -0,0 +1,22 @@
+class Form::Sales::Questions::SharedOwnershipType < ::Form::Question
+  def initialize(id, hsh, page)
+    super
+    @id = "type"
+    @check_answer_label = "Type of shared ownership sale"
+    @header = "What is the type of shared ownership sale?"
+    @hint_text = "A shared ownership sale is when the purchaser buys up to 75% of the property value and pays rent ro the Private Registered Provider (PRP) on the remaining portion"
```
baarkerlounger: Should this say "to the Private..."?

app/models/sales_log.rb
```diff
@@ -3,6 +3,7 @@ def validate(record); end
 end
 
 class SalesLog < Log
+  self.inheritance_column = :_type_disabled
```
baarkerlounger, line 6: For anyone wondering without this ActiveRecord assumes a column named type implies Single Table Inheritance applies and removes the column when serialising to JSON.

app/models/form/sales/questions/shared_ownership_type.rb
```diff
@@ -0,0 +1,22 @@
...
+    @type = "radio"
+    @answer_options = ANSWER_OPTIONS
+    @page = page
+  end
+
+  ANSWER_OPTIONS = {
+    "2" => { "value" => "Shared Ownership" },
```
baarkerlounger, line 14: Do the keys here actually need to be strings?

baarkerlounger (approved):

app/models/form/sales/questions/shared_ownership_type.rb
kosiakkatrina: That's what we have for the rest of the questions with answer options, so I'd think so

app/models/form/sales/questions/shared_ownership_type.rb
kosiakkatrina: Yup, I just copied it over from the ticket 😅 Good spot!

PR #882, opened 2022-09-20
Add discounted ownership type

kosiakkatrina:
Add discounted ownership type question and page to the sales setup section


spec/models/form/sales/questions/discounted_ownership_type_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+      "8" => { "value" => "Right to Aquire (RTA)" },
+      "14" => { "value" => "Preserved Right to Buy (PRTB)" },
+      "27" => { "value" => "Voluntary Right to Buy (VRTB)" },
+      "9" => { "value" => "Right to Buy (RTB)" },
+      "29" => { "value" => "Rent to Buy - Full Ownership" },
+      "21" => { "value" => "Social HomeBuy for outright purchase" },
+      "22" => { "value" => "Any other equity loan scheme" },
+    })
+  end
+end
```
CheshireSwift, line 45: I have mixed feelings on the usefulness of tests for the questions (at least when they're this simple). It feels very repetitious, but I worry extending that thought just ends up at them being data rather than code again 😅

kosiakkatrina: I see what you mean... I think especially with the answer option mappings, as they shouldn't really change, if they get changed accidentally this test would be a good indication that it might impact the data we're collecting.

kosiakkatrina: It is testing a lot of content, but I suppose that's because the content impacts the behaviour. I think these tests give me a little bit more confidence of the codebase, but deffo open to alternative ways of testing this 😅

CheshireSwift: Definitely a good point on the mappings. And yeah, for more straightforward content I don't have a better option off the top of my head, it might just be that on balance this is the best bet.

dushan-madetech (approved):
Nicely done

jamesrose (approved):

PR #884, opened 2022-09-20
Outright sale type

kosiakkatrina:
Add outright sale type and other sale type questions


app/helpers/conditional_questions_helper.rb
```diff
@@ -1,6 +1,6 @@
 module ConditionalQuestionsHelper
   def conditional_questions_for_page(page)
-    page.questions.map(&:conditional_for).compact.map(&:keys).flatten
+    page.questions.map(&:conditional_for).compact.map(&:keys).flatten.map(&:to_s)
```
jamesrose: What was the motivation for the changes in this file?

jamesrose (approved):

app/helpers/conditional_questions_helper.rb
kosiakkatrina: The conditional_for key was coming through as a symbol from the question. Not entirely sure why

kosiakkatrina: Ah, turns out it's just cause I'm silly like that 💁‍♀️🤦‍♀️
https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/884/commits/0dcc106d9b00179650c5d3b68c277e1d6ad94def

PR #886, opened 2022-09-20
CLDC-1474 sales setup ownership scheme

dushan-madetech:
Adds the ownership scheme question to the setup section of a Sales log


jamesrose (approved):

app/models/form/sales/questions/ownership_scheme.rb
```diff
@@ -0,0 +1,17 @@
+class Form::Sales::Questions::OwnershipScheme < ::Form::Question
+  def initialize(id, hsh, page)
+    super
+    @id = "ownershipsch"
+    @check_answer_label = "Purchase made under ownership scheme"
+    @header = "Was this purchase made through an ownership scheme?"
```
mseedat-moj, line 6: Are these strings meant to be in the locale file or is it ok to keep here?

mseedat-moj (commented):
comment about where strings live

app/models/form/sales/questions/ownership_scheme.rb
kosiakkatrina: Hm, these files mainly contain the definitions and very rarely some specific functionality for a question so they're unlikely to grow large. There are a lot of questions so if we keep all the strings in the locale file it might end up growing very large (which is what we are trying to avoid by putting these questions in separate files). I'd say it makes sense to keep them here

dushan-madetech: I think it's okay to live here as there aren't any plans for this service to be available in any other language besides English. The questions in a form can change ever year so leaving these in one file also makes it easier to remove any questions if it's decided that they're no longer needed in future forms as it means less places to delete things

PR #888, opened 2022-09-21
Cldc 1430 csv export

kosiakkatrina:
Remove unused `ethnic_other`, `sale_completion_date` and `sale_or_letting` columns from lettings_log table.
Update the csv export:
Add `collection_start_year`
Rename `rent_type` to `rent_type_detail`
Format `startdate`, `mrcdate` and `voiddate`

Refactor getting csv export labels into a method


jamesrose (approved):
All looks good to me @kosiakkatrina 👍 

The only thing that I noticed here is that you use the field name `collection_start_year` instead of `collection_year` as on the ticket. Was that intentional?

kosiakkatrina:
> All looks good to me @kosiakkatrina 👍
> 
> The only thing that I noticed here is that you use the field name `collection_start_year` instead of `collection_year` as on the ticket. Was that intentional?

@jamesrose A few things I guess, I went for it because we already have collection_start_year in the model and collection window can span over 2 years, so for me it made more sense to have start year. Happy to change if that makes more sense

jamesrose:
@kosiakkatrina Could you confirm with Sam, please?

PR #892, opened 2022-09-22
Cldc 1483 sales setup buyers living

dushan-madetech:
Add changes for the sales form question about buyers living in the property


kosiakkatrina (approved):
Looks good!

PR #893, opened 2022-09-22
Cldc 1482 sales log buyer company

natdeanlewissoftwire:
Tested locally
Minimal risk
No docs required

image

To be linked up with following questions as part of their respective tickets

Ticket: https://digital.dclg.gov.uk/jira/browse/CLDC-1482


app/models/form/sales/pages/buyer_company.rb
```diff
@@ -0,0 +1,15 @@
+class Form::Sales::Pages::BuyerCompany < ::Form::Page
```
jamesrose, line 1: Please could you add specs for the page and question objects here? You can use https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/885/files#diff-1732d36de4a2d8798f802b2beb2d85fc7547dc6a3640caa318d2304a16ee1c1d as an example.

natdeanlewissoftwire: added - I left out the :in-progress sales log in the sales_log.rb factory as that's shared behaviour so can just be picked up from Kat's PR

baarkerlounger:
Could also add it here: https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/blob/main/spec/features/sales_log_spec.rb#L36 if we want that to become an overall spec of the form composition (i.e. questions are asked in the correct order, routing is correct etc) but that spec could explode in size quite easily so not sure if we want that.

Otherwise think this just needs linting and then good to go.

kosiakkatrina (commented):
This looks good! Just needs a rebase

dushan-madetech (approved):
I think this might need a rebase but other than that it seems good to me 👍

Ted-U:
Looks like it needs a migration to add a "COMPANYBUY" column

natdeanlewissoftwire:
> Looks like it needs a migration to add a "COMPANYBUY" column

This is already in main, Dushan cherry picked the migration and updated schema as it affected a ticket he was working on that I believe is already merged now hence why it's not appearing in the diff.

PR #894, opened 2022-09-23
Allow a user to have a link to many legacy users

jamesrose:
While migrating users from the previous service to the new one we discovered that email addresses are not unique in the previous service. This means that one user in the new service might relate to multiple users in the previous service.

This PR:

- Adds a new `LegacyUser` model that can belong to a `User`. Each `LegacyUser` model has one `old_user_id` that corresponds to the user ID in the legacy service.
- Updates the user import service so that we create this association for new users
- Creates a Rake script to backfill the association for existing users


app/models/user.rb
```diff
@@ -113,7 +114,11 @@ def confirmable_template
   end
 
   def was_migrated_from_softwire?
-    old_user_id.present?
+    legacy_users.any?
+  end
+
+  def old_user_id
```
baarkerlounger: I think this method should probably just be removed altogether since it's original purpose isn't really valid anymore any remaining usages are likely a mistake.

app/services/imports/lettings_logs_import_service.rb
```diff
@@ -226,7 +226,7 @@ def create_log(xml_doc)
       # Sets the log creator
       owner_id = field_value(xml_doc, "meta", "owner-user-id").strip
       if owner_id.present?
-        attributes["created_by"] = User.find_by(old_user_id: owner_id)
+        attributes["created_by"] = LegacyUser.find_by(old_user_id: owner_id)&.user
```
baarkerlounger: Given that a missing `created_by` attribute will cause issues I wonder whether logging something here already will help speed up investigating.

db/schema.rb
```diff
@@ -232,12 +240,12 @@
     t.integer "void_date_value_check"
     t.integer "housingneeds_type"
     t.integer "housingneeds_other"
-    t.index ["created_by_id"], name: "index_lettings_logs_on_created_by_id"
-    t.index ["location_id"], name: "index_lettings_logs_on_location_id"
-    t.index ["managing_organisation_id"], name: "index_lettings_logs_on_managing_organisation_id"
-    t.index ["old_id"], name: "index_lettings_logs_on_old_id", unique: true
-    t.index ["owning_organisation_id"], name: "index_lettings_logs_on_owning_organisation_id"
-    t.index ["scheme_id"], name: "index_lettings_logs_on_scheme_id"
+    t.index ["created_by_id"], name: "index_case_logs_on_created_by_id"
```
baarkerlounger: This looks wrong to me? Why are these index names changing?

lib/tasks/sync_legacy_users.rake
```diff
@@ -0,0 +1,11 @@
+namespace :core do
+  # TODO: Remove once ran on all environments.
+  desc "Creates a LegacyUser object for any existing Users"
+  task sync_legacy_users: :environment do
+    User.where.not(old_user_id: nil).includes(:legacy_users).find_each do |user|
+      next if user.legacy_users.where(old_user_id: user.old_user_id).any?
```
baarkerlounger, line 6: I guess this could just be `next if user.legacy_users` since we know any existing users in the system would only have one anyway?

spec/factories/user.rb
```diff
@@ -19,5 +18,15 @@
     confirmed_at { Time.zone.now }
     created_at { Time.zone.now }
     updated_at { Time.zone.now }
+
+    transient do
```
baarkerlounger, line 22: Would probably remove this same as the method above, not sure it's still a useful attribute now?

spec/factories/user.rb
```diff
@@ -19,5 +18,15 @@
...
+      old_user_id { SecureRandom.uuid }
+    end
+
+    after(:create) do |user, evaulator|
```
baarkerlounger: Should this be `evaluator`?

app/models/user.rb
jamesrose: That's a great point. I've removed this method and updated `was_migrated_from_softwire?` above.

spec/factories/user.rb
baarkerlounger: Oh I see, you're just using it as the input for the after_create, :+1:

lib/tasks/sync_legacy_users.rake
jamesrose: I was hoping to make this script idempotent by default so that it could be ran multiple times if needed. The validations later should catch it, but it would raise.

spec/factories/user.rb
jamesrose: Yes, this is still useful. We still have a `old_user_id` column on users for now because we need to make sure all the associations are created before we can remove it.

This transient value means that we don't have to update tests yet :)

spec/factories/user.rb
jamesrose: Good spot!!

jamesrose: Fixed in [0547e55](https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/894/commits/0547e55f57cdb6f8426f07bf814755dc0957e86d)

app/models/user.rb
jamesrose: Fixed in [e54b9d0](https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/894/commits/e54b9d0d59e2a080e82aa7dcddf0a1b5171ae0c3)

db/schema.rb
jamesrose: Agreed! Resolved in [3c4d864](https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/894/commits/3c4d864902e7430555c0a52095edee9c3fb0e8af).

baarkerlounger (approved):

app/services/imports/lettings_logs_import_service.rb
jamesrose: Sure, see: [7aacbbc](https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/894/commits/7aacbbc624a3b4634110d2662b0a56466ad4c69b)

PR #898, opened 2022-09-26
[CLDC-1484] Add Buyer 1 age question

ghost:
ticket: https://digital.dclg.gov.uk/jira/browse/CLDC-1484

## Section and subsection
Screenshot 2022-09-26 at 11 20 16

## Select yes and shows answer box
Screenshot 2022-09-26 at 08 28 40

## Select no does not show answer box
Screenshot 2022-09-26 at 08 28 47

## List of answers
Screenshot 2022-09-26 at 11 21 13


baarkerlounger (approved):
Feels like this is maybe close enough to the lettings version to be the same question but also maybe not... :rocket:

PR #900, opened 2022-09-26
Set vacdays after setting the voiddate

kosiakkatrina:
Set vacdays correctly for supported housing, renewal letting logs.
We use voiddate to calculate vacdays, so we should set voiddate before setting vacdays.

Would need to run this on prod to update the existing logs:
LettingsLog.where(needstype: 2, renewal: 1).update_all(vacdays: 0)

### DIFF
--- a/app/models/derived_variables/lettings_log_variables.rb
+++ b/app/models/derived_variables/lettings_log_variables.rb
@@ -58,7 +58,6 @@ module DerivedVariables::LettingsLogVariables
 
     self.hhtype = household_type
     self.new_old = new_or_existing_tenant
-    self.vacdays = property_vacant_days
 
     if is_supported_housing?
       if location
@@ -68,6 +67,7 @@ module DerivedVariables::LettingsLogVariables
         self.voiddate = startdate
       end
     end
+    self.vacdays = property_vacant_days
 
     set_housingneeds_fields if housingneeds?
   end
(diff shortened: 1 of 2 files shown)


app/models/derived_variables/lettings_log_variables.rb
```diff
@@ -68,6 +67,7 @@ def set_derived_fields!
         self.voiddate = startdate
       end
     end
+    self.vacdays = property_vacant_days
```
baarkerlounger, line 70: Is the issue here that `property_vacant_days` uses `self.voiddate` so getting `vacdays` right was order of operations dependent?

kosiakkatrina: Yup

baarkerlounger: I guess the spec helps protect against that in future, might still be less brittle to make it more functional though. I.e. if `property_vacant_days` requires `voiddate` then make that explicit and pass it in as a param?

baarkerlounger (approved):

app/models/derived_variables/lettings_log_variables.rb
kosiakkatrina: true, but I think we've got a few variables like that. Especially the location ones

PR #906, opened 2022-09-27
CLDC 858 add validations to number of beds question in sales

Ted-U:
1 of 3 validations added. The remaining two validations are compound and related to questions which are not currently on main

- [x] Validation 1 - Number of bedrooms - Must be between 1 and 9
- [ ] Compound validation 1: number of bedrooms and property type - Cannot be 2 if BEDS is greater than 1
- [ ] Compound SOFT validation 2: Full purchase price, local authority, and number of bedrooms

https://digital.dclg.gov.uk/jira/browse/CLDC-858


asmega:
closing stale PR

PR #909, opened 2022-09-27
Cldc 1503 lead nationality

kosiakkatrina:
Add buyer's 1 nationality questions and update conditional question controller to work with both log types
image


app/frontend/controllers/conditional_question_controller.js
```diff
@@ -7,15 +7,20 @@ export default class extends Controller {
 
   displayConditional () {
     if (this.element.checked) {
+      console.log(this.element)
```
baarkerlounger: Was this just for debugging or did you mean to leave it in?

kosiakkatrina: Yikes 😬 Just debugging. Tyty!

spec/features/form/conditional_questions_spec.rb
```diff
@@ -44,5 +52,15 @@
       visit("/lettings-logs/#{id}/property-postcode")
       expect(page).to have_field("lettings-log-postcode-full-field", with: "NW1 6RT")
     end
+
+    it "gets cleared if the conditinoal question is hidden after editing the answer" do
```
natdeanlewissoftwire: conditional* 😄

dushan-madetech (approved):
Just needs a little rebase and then you're good 👍

baarkerlounger (approved):

db/schema.rb
```diff
@@ -329,22 +329,24 @@
...
     t.integer "buylivein"
     t.integer "builtype"
     t.integer "proptype"
-    t.string "relat2"
-    t.string "otherrelat2"
     t.integer "age2"
     t.integer "age2_known"
+    t.string "relat2"
+    t.string "otherrelat2"
     t.integer "ethnic"
```
natdeanlewissoftwire, line 350: I'm not sure how to avoid these reordering - I think it's because of the migrations added since this branch was made. I've done a rebase but that didn't seem to help. I know it's not an issue really but if schema order is something we should keep consistent how do I do that?

jamesrose: If it's just re-ordering, I wouldn't worry too much about it.

jamesrose (approved):

PR #911, opened 2022-09-27
[CLDC-1467] add type of building question to sales

Ted-U:
https://digital.dclg.gov.uk/jira/browse/CLDC-1467

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -348,6 +348,7 @@ ActiveRecord::Schema[7.0].define(version: 2022_10_05_150022) do
     t.integer "age2"
     t.integer "age2_known"
     t.string "relat2"
+    t.string "otherrelat2"
     t.string "sex2"
     t.integer "noint"
     t.integer "buy2livein"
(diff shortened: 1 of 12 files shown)


db/schema.rb
```diff
@@ -338,7 +339,7 @@
     t.integer "age1_known"
     t.string "sex1"
     t.integer "buy1livein"
-    t.integer "buylivein"
```
baarkerlounger: Why is this field being removed? I don't think this is correct, you might need to rebase and migrate.

Ted-U: Thanks Dan, fixed now.

baarkerlounger:
@Ted-U is this a duplicate of https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/918/files now and if so can it be closed?

ghost:
@baarkerlounger picking up on the question you asked last to Ted. It seems like it was a duplicate of https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/918 but the page ID was not correct so I fixed it in this PR

jamesrose (approved):

PR #912, opened 2022-09-28
[CLDC-1222] Convert lettings log import to background job

mseedat-moj:
### Summary
Introduces background job processing for letting case log importing.

### Ticket
https://digital.dclg.gov.uk/jira/browse/CLDC-1222

### Notes
This PR will allow large batches of logs to be processed by taking advantage of multiple servers to perform the required validations and database updates. 

Case log imports typically take approximately 50ms to process (localhost), however the need to validate Postcodes using an external web service can add up to 5 seconds (circuit-breaker in place to max-out at 5 seconds). 

When processing 100s/1000s of imports in serial, this processing round-trip will increase overall batch processing time and therefore possibly cause server time outs, etc.

**Out of scope in this PR:**
There is no reporting/collation of processed records in this PR. An experimental [branch](https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/tree/CLDC-122-experimental-background-job-lettings-log-import-with-report) demonstrates the use of PubSub events via Wisper to allow basic metrics to be collated during background job processing


baarkerlounger:
@mseedat-moj we already use redis and sidekiq for queueing and background job processing so probably better to use those here also rather than introducing wisper and resque as well

Procfile.dev
```diff
@@ -1,2 +1,3 @@
+QUEUE=* rake resque:work
```
jamesrose: I don't think you need this anymore.

app/events/import.rb
```diff
@@ -0,0 +1,6 @@
+
+  module Import
```
jamesrose: nit: file spacing in this file

app/helpers/lettings_test_helper.rb
```diff
@@ -0,0 +1,2 @@
+module LettingsTestHelper
```
baarkerlounger: Is this needed?

app/jobs/cleanup_job.rb
```diff
@@ -0,0 +1,7 @@
+class CleanupJob < ApplicationJob
```
baarkerlounger: Also this?

jamesrose (approved):
LGTM. Nice work!

spec/jobs/lettings_log_import_job_spec.rb
```diff
@@ -0,0 +1,29 @@
+require "rails_helper"
+
+RSpec.describe LettingsLogImportJob do
```
ghost, line 3: RSpec provides some nice test helpers for jobs:
https://relishapp.com/rspec/rspec-rails/docs/job-specs/job-spec

PR #915, opened 2022-09-28
CLDC-1466 property unit type

dushan-madetech:
Adds the property unit type question to sales log and associated tests


jamesrose (approved):
LGTM. I think you need to adjust your commit messages when you squash :)

PR #916, opened 2022-09-29
CLDC-1523 buyer 2 age

dushan-madetech:
Adds changes needed for buyer 2 age along with accompanying tests

### DIFF
--- a/app/models/form/sales/questions/age2.rb
+++ b/app/models/form/sales/questions/age2.rb
@@ -1,0 +1,11 @@
+class Form::Sales::Questions::Age2 < ::Form::Question
+  def initialize(id, hsh, page)
+    super
+    @id = "age2"
+    @check_answer_label = "Buyer 2’s age"
+    @header = "Age"
+    @type = "numeric"
+    @page = page
+    @width = 2
+  end
+end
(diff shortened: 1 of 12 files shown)


app/models/form/sales/questions/age2.rb
```diff
@@ -0,0 +1,11 @@
+class Form::Sales::Questions::Age2 < ::Form::Question
+  def initialize(id, hsh, page)
+    super
+    @id = "age2"
+    @check_answer_label = "Buyer 2’s age"
+    @header = "Age"
+    @type = "numeric"
+    @page = page
+    @width = 2
```
baarkerlounger, line 9: Should this be 3?

baarkerlounger (approved):

app/models/form/sales/questions/age2.rb
dushan-madetech: The width?

baarkerlounger: Yep? You could feasibly have a 100+ year old buyer?

dushan-madetech: I just mimicked what was set up for the lead tenant age too. I think that one should also be changed too if this one is

PR #919, opened 2022-09-30
Cldc 1522 buyer 2 relationship

natdeanlewissoftwire:
ticket: https://digital.dclg.gov.uk/jira/browse/CLDC-1522

Not sure if this PR should include pages for person 2/3/4/5 as is suggested in the ticket?

image

Tested locally
Minimal risk
No docs required

### DIFF
--- a/app/models/form/sales/questions/buyer2_relationship_to_buyer1.rb
+++ b/app/models/form/sales/questions/buyer2_relationship_to_buyer1.rb
@@ -9,7 +9,7 @@ class Form::Sales::Questions::Buyer2RelationshipToBuyer1 < ::Form::Question
     @page = page
     @answer_options = ANSWER_OPTIONS
     @conditional_for = {
-      "otherrelat2" => ["X"],
+      "otherrelat2" => %w[X],
     }
     @hidden_in_check_answers = {
       "depends_on" => [
@@ -22,7 +22,7 @@ class Form::Sales::Questions::Buyer2RelationshipToBuyer1 < ::Form::Question
 
   ANSWER_OPTIONS = {
     "P" => { "value" => "Parent" },
-    "C" => { "value" => "Child" },
+    "C" => { "value" => "Child", "hint" => "Must be eligible for child benefit, aged under 16 or under 20 if still in full-time education." },
     "X" => { "value" => "Other" },
     "R" => { "value" => "Buyer prefers not to say" },
   }.freeze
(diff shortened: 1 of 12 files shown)


app/models/form/sales/questions/buyer2_relationship_to_buyer1.rb
```diff
@@ -0,0 +1,29 @@
...
+    @hint_text = ""
+    @page = page
+    @answer_options = ANSWER_OPTIONS
+    @conditional_for = {
+      "otherrelat2" => %w[X],
+    }
+    @hidden_in_check_answers = {
+      "depends_on" => [
+        {
+          "relat2" => "X",
```
baarkerlounger, line 17: i.e. This is hidden in check answers when the answer is other? Why?

app/models/form/sales/questions/other_buyer2_relationship_to_buyer1.rb
```diff
@@ -0,0 +1,10 @@
+class Form::Sales::Questions::OtherBuyer2RelationshipToBuyer1 < ::Form::Question
+  def initialize(id, hsh, page)
+    super
+    @id = "otherrelat2"
```
baarkerlounger, line 4: Is this the question that should be hidden in check answers? (i.e. if you haven't said the relationship is other, we don't need to show "what other"?

app/models/form/sales/questions/buyer2_relationship_to_buyer1.rb
natdeanlewissoftwire: When it's Other we actually just show the text the user input rather than the word Other as well - I think that's what we want? Same format as here: https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/909/files#diff-8ea21e0b734464835ea4577b4152b914cbabde55e6a35c36f02edfb69f6ce242

baarkerlounger: Right that makes sense, sorry early morning :sweat_smile:

baarkerlounger (approved):

app/models/form/sales/questions/other_buyer2_relationship_to_buyer1.rb
natdeanlewissoftwire: Yep, I think I'm doing that already - here's what the CYA page looks like in the two cases:
Non-other:
image
Other with user input:
image

PR #920, opened 2022-09-30
CLDC-1625 Added create new sales log for this organisation button on organisation view

davidm-m:
I just copied the existing lettings logs tests and changed them to fit sales logs. I removed the ones around searching because that isn't fully implemented yet - you can't search for e.g. purchaser code. I'm happy to add in a TODO but couldn't see a ticket to reference in the backlog.

### DIFF
--- a/app/views/organisations/logs.html.erb
+++ b/app/views/organisations/logs.html.erb
@@ -14,10 +14,10 @@
 
 <div class="app-filter-layout" data-controller="filter-layout">
   <div class="govuk-button-group app-filter-toggle">
-    <% if current_page?(:controller => 'organisations', :action => 'lettings_logs') %>
+    <% if current_page?(controller: 'organisations', action: 'lettings_logs') %>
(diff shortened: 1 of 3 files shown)


app/views/organisations/logs.html.erb
```diff
@@ -14,7 +14,12 @@
 
 <div class="app-filter-layout" data-controller="filter-layout">
   <div class="govuk-button-group app-filter-toggle">
-    <%= govuk_button_to "Create a new lettings log for this organisation", lettings_logs_path(lettings_log: { owning_organisation_id: @organisation.id }, method: :post) %>
+    <% if current_page?(:controller => 'organisations', :action => 'lettings_logs') %>
```
baarkerlounger: Think the `controller: 'organisations` has syntax is preferred

baarkerlounger:
There definitely should be a ticket for searching sales logs by purchaser code so might be worth creating and implementing here as well

baarkerlounger (approved):

PR #932, opened 2022-10-05
CLDC-1505 Buyer 1 working situation

CheshireSwift:


natdeanlewissoftwire (approved):
There's a failing test to fix but otherwise looks good to me 😄 
(I think you just need to add a value for ecstat1 to the completed sales log in spec/factories/sales_log.rb)

CheshireSwift:
Yep, spotted the test when I got back from lunch, will push fix after meeting 😄

PR #939, opened 2022-10-07
Cldc 1529 person 1 age

natdeanlewissoftwire:
tested locally
minimal risk
no docs required

ticket: https://digital.dclg.gov.uk/jira/browse/CLDC-1529

image

### DIFF
--- a/app/models/form/sales/pages/person1_age.rb
+++ b/app/models/form/sales/pages/person1_age.rb
@@ -1,0 +1,22 @@
+class Form::Sales::Pages::Person1Age < ::Form::Page
+  def initialize(id, hsh, subsection)
+    super
+    @id = "person_1_age"
+    @header = ""
+    @description = ""
+    @subsection = subsection
+    @depends_on = [
+      { "hholdcount" => 1 },
+      { "hholdcount" => 2 },
+      { "hholdcount" => 3 },
+      { "hholdcount" => 4 },
+    ]
+  end
+
+  def questions
+    @questions ||= [
+      Form::Sales::Questions::Person1AgeKnown.new(nil, nil, self),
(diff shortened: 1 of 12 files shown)


app/models/form/sales/pages/person1_age.rb
```diff
@@ -0,0 +1,22 @@
+class Form::Sales::Pages::Person1Age < ::Form::Page
+  def initialize(id, hsh, subsection)
+    super
+    @id = "person_1_age"
+    @header = ""
+    @description = ""
+    @subsection = subsection
+    @depends_on = [
```
CheshireSwift, line 8: Not a comment on this so much as a general thought, it might be good to refactor our deps to support conditions, now that we're in a fully-fledged programming language rather than JSON.

spec/factories/sales_log.rb
```diff
@@ -39,6 +39,8 @@
       ecstat2 { "1" }
       hholdcount { "1" }
       wheel { 1 }
+      age3_known { 0 }
```
CheshireSwift, line 42: This looks like our factory output is invalid by default? (age is not known, but also 40)

CheshireSwift (approved):
Only one real query, otherwise lgtm 👍

app/models/form/sales/pages/person1_age.rb
natdeanlewissoftwire: Completely agree, I hate this too :D

spec/factories/sales_log.rb
natdeanlewissoftwire: I don't think so  - 0 is "Yes" for age3_known (following a pattern used elsewhere) which is confusing tbf

CheshireSwift: Ah, okay. Odd, but sure.

PR #940, opened 2022-10-11
[CLDC-1512] Add buyer1 income

ghost:
https://digital.dclg.gov.uk/jira/browse/CLDC-1512

Blocked: needs a decision on whether we need to implement a details component


app/models/form/sales/questions/buyer1_income_known.rb
```diff
@@ -0,0 +1,32 @@
+class Form::Sales::Questions::Buyer1IncomeKnown < ::Form::Question
+  def initialize(id, hsh, page)
+    super
+    @id = "income1nk"
+    @check_answer_label = "Buyer 1’s gross annual income"
+    @header = "Do you know buyer 1’s annual income?"
+    @type = "radio"
+    @answer_options = ANSWER_OPTIONS
+    @page = page
+    @hint_text = "What counts as income?
```
jamesrose: I'm not sure a multiline string is going to translate onto the page. Could you double check?

app/models/form/sales/questions/buyer1_income_known.rb
```diff
@@ -0,0 +1,32 @@
...
+    housing benefit
+    child benefit
+    council tax support"
+    @conditional_for = {
+      "income1" => [0],
+    }
+  end
+
+  ANSWER_OPTIONS = {
+    "0" => { "value" => "Yes" },
```
jamesrose, line 18: 😢

app/models/form/sales/questions/buyer1_income_known.rb
ghost: @jamesrose this is apparently supposed to be a details component, I left a comment for you here:

https://digital.dclg.gov.uk/jira/browse/CLDC-1512

What would you recommend going for?

app/models/form/sales/questions/buyer1_income_known.rb
ghost: does this need correction or we just go with what was requested?

jamesrose: Likely what is requested, as it will match what exists for sales logs, but it's unfortunate that the zero value is yes.

jamesrose (approved):
Nice!

PR #944, opened 2022-10-14
Cldc 1537 details for person x

natdeanlewissoftwire:
tested locally
slight risk in changing the header to be a header OR header_partial - this shouldn't affect previous code however.
no docs required

Similar to previous ticket, could be simplified but I've worked on getting this merged instead as time is a constraint.

Ticket: https://digital.dclg.gov.uk/jira/browse/CLDC-1537

Page: 
image

CYA: 
image


app/models/form.rb
```diff
@@ -60,6 +60,8 @@ def next_page(page, log, current_user)
     page_index = page_ids.index(page.id)
     page_id = if page.id.include?("value_check") && log[page.questions[0].id] == 1 && page.routed_to?(log, current_user)
                 previous_page(page_ids, page_index, log, current_user)
+              elsif page.id == ("person_2_known")
+                "person_2_age"
```
natdeanlewissoftwire, line 64: fixing this now...

app/views/form/page.html.erb
```diff
@@ -11,11 +11,14 @@
...
-      <% if @page.header.present? %>
+      <% if @page.header.present? || @page.header_partial.present? %>
         <h1 class="govuk-heading-l">
           <% if !@page.hide_subsection_label %>
             <span class="govuk-caption-l"><%= @subsection.label %></span>
           <% end %>
+          <% if @page.header_partial.present? %>
+            <% @page.header = render partial: "form/headers/#{@page.header_partial}", locals: { hholdcount: @log.hholdcount } %>
+          <% end %>
           <%= @page.header %>
```
jamesrose:
```suggestion
          <% if @page.header_partial.present? %>
            <%= render partial: "form/headers/#{@page.header_partial}", locals: { log: @log} %>
          <% else %>
           <%= @page.header %>
          <% end %>
```

I think I'd make this generic so that it can other types of partials can use anything in the log. In my example above I made the header rendering conditional so it's a little clearer what's happening.

jamesrose: @natdeanlewissoftwire Just checking that you saw this comment!

natdeanlewissoftwire: will come back to this if we finish this sprint's work early/when sales logs are back in scope

app/views/form/headers/_person_2_known_page.erb
```diff
@@ -1 +1 @@
-You have given us the details for 1 of the <%= hholdcount %> other people in the household
+You have given us the details for 1 of the <%= log.hholdcount %> other people in the household
```
kosiakkatrina, line 1: Would it make sense at all to put these headers into the `config/locales/en.yml` instead of their own files?

natdeanlewissoftwire: Hi Kat :) ah, I hadn't thought of that, will look into it this morning, thanks.

natdeanlewissoftwire: So at the moment I can't see a way of passing the `hholdcount` into `en.yml` from a sales page without adding extra functionality to the page class - I don't think this is something that I can say we'll definitely need later so I don't think it makes sense to add that feature just for this ticket.

app/views/form/page.html.erb
```diff
@@ -32,7 +36,7 @@
           <% if question.type == "interruption_screen" %>
             <%= render partial: "form/#{question.type}_question", locals: { question:, caption_text: @subsection.label, page_header: @page.header, lettings_log: @log, title_text: @page.title_text, informative_text: @page.informative_text, form: @form, f:, conditional: false } %>
           <% else %>
-            <%= render partial: "form/#{question.type}_question", locals: { question:, caption_text: @subsection.label, page_header: @page.header, lettings_log: @log, f:, conditional: false } %>
+            <%= render partial: "form/#{question.type}_question", locals: { question:, caption_text: @page.header_partial.present? ? nil : @subsection.label, page_header: @page.header, lettings_log: @log, f:, conditional: false } %>
```
jamesrose, line 39: Why do we not have caption text when there's a header partial?

jamesrose (approved):

app/views/form/page.html.erb
natdeanlewissoftwire: Otherwise the caption text shows twice, once above the header partial, once above the header

PR #946, opened 2022-10-17
Cldc 1675 rent validations

natdeanlewissoftwire:
tested locally
minimal risk
no docs required

soft validation:
image

hard validation: 
image

ticket: https://digital.dclg.gov.uk/jira/browse/CLDC-1675


app/helpers/interruption_screen_helper.rb
```diff
@@ -5,7 +5,12 @@ def display_informative_text(informative_text, lettings_log)
     translation_params = {}
     informative_text["arguments"].each do |argument|
       value = if argument["label"]
-                lettings_log.form.get_question(argument["key"], lettings_log).answer_label(lettings_log).downcase
+                pre_casing_value = lettings_log.form.get_question(argument["key"], lettings_log).answer_label(lettings_log)
```
natdeanlewissoftwire, line 8: Couldn't be completely sure of the reason why this was being `.downcase`ed before, but I've just replaced it for local authorities as presumably there must have been a reason (there are some cases where gender is filled in dynamically so it might be that those needed to be lower case).

ghost (approved):

PR #949, opened 2022-10-19
WIP: Cldc 1661 new page for housing providers

natdeanlewissoftwire:


config/routes.rb
```diff
@@ -79,6 +79,8 @@
       post "logs/email-csv", to: "organisations#email_csv"
       get "logs/csv-confirmation", to: "lettings_logs#csv_confirmation"
       get "schemes", to: "organisations#schemes"
+      get "managing-agents", to: "organisation_relationships#managing_agents"
+      get "housing-providers", to: "organisations#housing_providers"
```
ghost, line 82: think we should use the same `organisation_relationships` controller

PR #950, opened 2022-10-20
Cldc 1661 housing providers branch

natdeanlewissoftwire:
tested locally
minimal risk - mostly new content
no docs required

ticket: https://digital.dclg.gov.uk/jira/browse/CLDC-1661

data coordinator view: 
image

data provider view:
image

support view:
image

search functional, add/remove are not yet as these are in following tickets


natdeanlewissoftwire:
fixing the checks...

spec/requests/organisations_controller_spec.rb
```diff
@@ -1025,11 +1121,11 @@
             end
 
             it "updates the table caption" do
-              expect(page).to have_content("1 organisation found matching ‘#{search_param}’")
+              expect(page).to have_content("1 organisations found matching ‘#{search_param}’")
```
ghost, line 1124: i think this change from single to plural does not look correct, might we worth to confirm with a content designer

ghost:
@natdeanlewissoftwire looks great! apart from the comment on singular/plural, the only thing I can see is the `remove` button not being on the right end of the row like in the designs:
Screenshot 2022-10-20 at 11 49 05

spec/requests/organisations_controller_spec.rb
natdeanlewissoftwire: So it's more of a quick change that was easy to do - at the moment it's showing singular for all numbers (we just happen to be testing for 1) - I think it's better to show plural for all numbers until we add proper pluralisation

natdeanlewissoftwire: the users page already shows plural for all numbers I think, so it's not even consistent as it stands

natdeanlewissoftwire:
> @natdeanlewissoftwire looks great! apart from the comment on singular/plural, the only thing I can see is the `remove` button not being on the right end of the row like in the designs: Screenshot 2022-10-20 at 11 49 05

Ah nice, that reminds me I meant to make the search box look like the designs as well though I'm not sure if that matters - I'll check both these with the design team

ghost (approved):
🚀

PR #955, opened 2022-10-26
feat: update logs link

natdeanlewissoftwire:
tested locally
minimal risk
no docs required


jamesrose (approved):

jamesrose:
This is why we should use path helpers :)

PR #956, opened 2022-10-27
[CLDC-1659] Create managing agents

ghost:
https://digital.dclg.gov.uk/jira/browse/CLDC-1659

This PR implements the flow to add new managing agents


app/views/organisation_relationships/add_managing_agent.html.erb
```diff
@@ -0,0 +1,36 @@
+<%= form_with model: @organisation, url: managing_agents_organisation_path, method: "post", local: true do |f| %>
+  <% if current_user.support? %>
+    <%= render partial: "organisations/headings", locals: { main: @organisation.name, sub: nil } %>
+    <%= render SubNavigationComponent.new(items: secondary_items(request.path, @organisation.id)) %>
+    <h2 class="govuk-visually-hidden">Add Managing Agent</h2>
+  <% end %>
```
natdeanlewissoftwire: I think lines 6 and 7 can be removed, just noticed on the equivalent housing providers page too.

app/views/organisation_relationships/managing_agents.html.erb
```diff
@@ -1,23 +1,47 @@
...
   ) %>
-  <h2 class="govuk-visually-hidden">managing agents</h2>
+  <h2 class="govuk-visually-hidden">Managing Agents</h2>
+    <% if params["related_organisation_id"] %>
+      <%= govuk_notification_banner(
+            title_text: "Success",
+            success: true, title_heading_level: 3,
+            title_id: "swanky-notifications"
+          ) do |notification_banner|
+        notification_banner.heading(text: "#{Organisation.find(params["related_organisation_id"]).name} is now one of your managing agents")
```
natdeanlewissoftwire: small thing, I think this should say "this organisation's" rather than that "your" as the support user might be editing for a different organisation than their own. There's a few other places where the designs say "your" assuming the user is a data coordinator but I think we need some conditional text that shows "this organisation" instead for support users

app/views/organisation_relationships/managing_agents.html.erb
```diff
@@ -1,23 +1,47 @@
...
+      end %>
+  <% end %>
   <p class="govuk-body">A managing agent can submit logs for your organisation.</p>
+  <% if @total_count == 0 %>
+    <p class="govuk-body">You do not currently have any managing agents.</p>
```
natdeanlewissoftwire: e.g. I think this should be "This organisation does" not "You do"

spec/requests/organisation_relationships_controller_spec.rb
```diff
@@ -0,0 +1,473 @@
+require "rails_helper"
```
natdeanlewissoftwire, line 1: thanks for making this, definitely a better approach than cramming it into the org controller spec

app/views/organisation_relationships/add_managing_agent.html.erb
ghost: I'll update both 👍

app/views/organisation_relationships/managing_agents.html.erb
ghost: yeah you're totally right! updating

natdeanlewissoftwire (approved):

PR #957, opened 2022-10-28
Cldc 1663 remove housing provider

natdeanlewissoftwire:
tested locally
risk of users being able to delete other org relationships by changing id in url - presumably this should be put into a header instead? However they can't use this to delete other org's relationships, only other relationships with their own org, so not a huge risk.
no docs required

ticket: https://digital.dclg.gov.uk/jira/browse/CLDC-1663

image
image


ghost:
there must be something off with you git, are you branching off from `main`?

app/controllers/organisation_relationships_controller.rb
```diff
@@ -66,6 +66,20 @@ def create!(child_organisation_id:, parent_organisation_id:, relationship_type:)
...
+  end
+
+  def delete_housing_provider
+    organisation_relationship_to_remove_id = OrganisationRelationship.find_by(child_organisation_id: @organisation.id, parent_organisation_id: organisation_to_remove_id, relationship_type: OrganisationRelationship::OWNING).id
+    delete(organisation_relationship_to_remove_id)
+    redirect_to housing_providers_organisation_path(removed_organisation_id: organisation_to_remove_id)
+  end
+
+  def delete(organisation_relationship_to_remove_id)
+    OrganisationRelationship.destroy(organisation_relationship_to_remove_id)
```
ghost: prefer `destroy!` with a bang so that it'll raise an error when it's not successful

app/controllers/organisation_relationships_controller.rb
```diff
@@ -66,6 +66,20 @@ def create!(child_organisation_id:, parent_organisation_id:, relationship_type:)
     @resource.save!
   end
 
+  def remove_housing_provider
+    render "organisation_relationships/remove_housing_provider", layout: "application"
+  end
+
+  def delete_housing_provider
+    organisation_relationship_to_remove_id = OrganisationRelationship.find_by(child_organisation_id: @organisation.id, parent_organisation_id: organisation_to_remove_id, relationship_type: OrganisationRelationship::OWNING).id
```
ghost: `find_by!` same as below

app/helpers/navigation_items_helper.rb
```diff
@@ -40,17 +40,17 @@ def secondary_items(path, current_organisation_id)
         NavigationItem.new("Schemes", "/organisations/#{current_organisation_id}/schemes", subnav_supported_housing_schemes_path?(path)),
         NavigationItem.new("Users", "/organisations/#{current_organisation_id}/users", subnav_users_path?(path)),
         NavigationItem.new("About this organisation", "/organisations/#{current_organisation_id}", subnav_details_path?(path)),
-        (NavigationItem.new("Housing providers", housing_providers_organisation_path(current_user.organisation), housing_providers_path?(path)) if FeatureToggle.managing_owning_enabled?),
-        (NavigationItem.new("Managing agents", managing_agents_organisation_path(current_user.organisation), managing_agents_path?(path)) if FeatureToggle.managing_owning_enabled?),
+        (NavigationItem.new("Housing providers", "/organisations/#{current_organisation_id}/housing-providers", housing_providers_path?(path)) if FeatureToggle.managing_owning_enabled?),
```
ghost: why not using the path helpers?

app/controllers/organisation_relationships_controller.rb
```diff
@@ -76,6 +90,10 @@ def related_organisation_id
     params["organisation"]["related_organisation_id"]
   end
 
+  def organisation_to_remove_id
+    params["organisation_to_remove_id"]
```
ghost: can't this be `params["id"]`?

natdeanlewissoftwire:
> there must be something off with you git, are you branching off from main?

I think I branched this one off the add housing provider branch because it was from before that was merged, might do a rebase to tidy it up.

app/helpers/navigation_items_helper.rb
natdeanlewissoftwire: For support users the path should include the selected org id not the id of the org the user belongs to. Otherwise they can't access other org's pages. Thinking about it again `housing_providers_organisation_path(current_organisation_id)` should have the same effect

app/controllers/organisation_relationships_controller.rb
natdeanlewissoftwire: i think params["id"] is the current user's org, which would be wrong for support users who can remove relationships for other organisations as well as for their own.

app/controllers/organisation_relationships_controller.rb
```diff
@@ -93,24 +93,37 @@ def create_managing_agent
...
+  end
+
+  def delete_housing_provider
+    organisation_relationship_to_remove = OrganisationRelationship.find_by!(child_organisation_id: @organisation.id, parent_organisation_id: organisation_to_remove_id, relationship_type: OrganisationRelationship::OWNING)
+    delete(organisation_relationship_to_remove)
+    redirect_to housing_providers_organisation_path(removed_organisation_id: organisation_to_remove_id)
+  end
+
+  def delete(organisation_relationship_to_remove)
+    organisation_relationship_to_remove.destroy!
```
jamesrose, line 90: I don't think you need to extract this :)

app/controllers/organisation_relationships_controller.rb
```diff
@@ -93,24 +93,37 @@ def create_managing_agent
...
   end
 
   def related_organisation_id
     params["organisation"]["related_organisation_id"]
   end
 
-  def organisation
-    @organisation ||= Organisation.find(params[:id])
+  def organisation_to_remove_id
+    params["organisation_to_remove_id"]
```
jamesrose: How about `target_organisation` or similar?

app/controllers/organisation_relationships_controller.rb
natdeanlewissoftwire: True, I was copying the pattern we used in create/add methods for sharing content between housing provider/managing agent but I'll see if there's anything more substantial we can share in this method (otherwise I'll remove it)

app/controllers/organisation_relationships_controller.rb
```diff
@@ -93,24 +93,37 @@ def create_managing_agent
...
   end
 
+  def remove_housing_provider
+    render "organisation_relationships/remove_housing_provider", layout: "application"
+  end
+
+  def delete_housing_provider
+    organisation_relationship_to_remove = OrganisationRelationship.find_by!(child_organisation_id: @organisation.id, parent_organisation_id: organisation_to_remove_id, relationship_type: OrganisationRelationship::OWNING)
+    delete(organisation_relationship_to_remove)
+    redirect_to housing_providers_organisation_path(removed_organisation_id: organisation_to_remove_id)
```
jamesrose: We can use flash notices instead of passing the removed organisation in: https://api.rubyonrails.org/classes/ActionDispatch/Flash.html

Let me know if you'd like to chat about how this works.

app/controllers/organisation_relationships_controller.rb
```diff
@@ -93,24 +93,37 @@ def create_managing_agent
     redirect_to managing_agents_organisation_path(related_organisation_id:)
   end
 
+  def remove_housing_provider
+    render "organisation_relationships/remove_housing_provider", layout: "application"
```
jamesrose:
```suggestion
```

You shouldn't need to be explicit about the render here. All of those are implicit as the action name matches the template name.

app/views/organisation_relationships/remove_housing_provider.html.erb
```diff
@@ -0,0 +1,23 @@
+<% @organisation_to_remove_id = params["organisation_to_remove_id"] %>
+<%= form_with url: housing_providers_organisation_path(organisation_to_remove_id: @organisation_to_remove_id), method: "delete", local: true do |f| %>
+  <% if current_user.support? %>
+    <%= render partial: "organisations/headings", locals: { main: @organisation.name, sub: nil } %>
+    <%= render SubNavigationComponent.new(items: secondary_items(request.path, @organisation.id)) %>
+    <h2 class="govuk-visually-hidden">Remove Housing Provider</h2>
+  <% end %>
+  <% @housing_provider = Organisation.find(@organisation_to_remove_id) %>
```
jamesrose: I'd move this into the controller. You won't need to do this at the top of the file, then:

```
<% @organisation_to_remove_id = params["organisation_to_remove_id"] %>
```

ghost (approved):

app/controllers/organisation_relationships_controller.rb
```diff
@@ -93,24 +78,34 @@ def create_managing_agent
...
   end
 
+  def remove_housing_provider
+    @target_organisation_id = target_organisation_id
+  end
+
+  def delete_housing_provider
+    organisation_relationship_to_remove = OrganisationRelationship.find_by!(child_organisation_id: @organisation.id, parent_organisation_id: target_organisation_id, relationship_type: OrganisationRelationship::OWNING)
+    organisation_relationship_to_remove.destroy!
+    flash[:notice] = "#{Organisation.find(target_organisation_id).name} is no longer one of #{current_user.data_coordinator? ? 'your' : "this organisation's"} housing providers"
```
jamesrose:
```suggestion
    flash[:notice] = "#{target_organisation.name} is no longer one of #{current_user.data_coordinator? ? 'your' : "this organisation's"} housing providers"
```

As discussed earlier, we can extract this like we've done with `organisation` below — then we won't need finders for each method.

app/controllers/organisation_relationships_controller.rb
```diff
@@ -93,24 +78,34 @@ def create_managing_agent
     redirect_to managing_agents_organisation_path(related_organisation_id:)
   end
 
+  def remove_housing_provider
+    @target_organisation_id = target_organisation_id
+  end
+
+  def delete_housing_provider
+    organisation_relationship_to_remove = OrganisationRelationship.find_by!(child_organisation_id: @organisation.id, parent_organisation_id: target_organisation_id, relationship_type: OrganisationRelationship::OWNING)
```
jamesrose:
```suggestion
    organisation_relationship_to_remove = OrganisationRelationship.find_by!(child_organisation: @organisation, parent_organisation: target_organisation, relationship_type: OrganisationRelationship::OWNING)
```

PR #959, opened 2022-10-31
Cldc 1675 rent validations

natdeanlewissoftwire:
tested locally
minimal risk - tightens validations
no docs required

https://digital.dclg.gov.uk/jira/browse/CLDC-1675

This PR makes rent range validations work for supported housing logs (previously working for general needs only) - the number of logs affected by these validations not being in place needs to be checked

Tests added for hard validation would have caught the issue for soft validation too - should separate tests be added for soft validation or is it okay to leave the pre-existing soft validation tests as they are?


app/models/lettings_log.rb
```diff
@@ -435,13 +435,29 @@ def self.to_csv(user = nil)
     Csv::LettingsLogCsvService.new(user).to_csv
   end
 
+  def validation_beds
```
jamesrose: Is `beds` only set when the log is a general needs letting?

app/models/lettings_log.rb
```diff
@@ -435,13 +435,29 @@ def self.to_csv(user = nil)
...
+    if needstype == 2
+      0
+    else
+      beds.nil? ? nil : [beds, 4].min
+    end
+  end
+
+  def validation_la
```
jamesrose: Does only supported housing have a location set?

natdeanlewissoftwire: Yes - for general needs we ask the user the postcode (and infer the la or ask if we can't) directly and add that to the log. For supported housing we ask the user the scheme (and location if there are more than 1 for the scheme) and then need to use the local authority code from that instead.

app/models/lettings_log.rb
```diff
@@ -435,13 +435,29 @@ def self.to_csv(user = nil)
     Csv::LettingsLogCsvService.new(user).to_csv
   end
 
+  def validation_beds
+    if needstype == 2
+      0
+    else
+      beds.nil? ? nil : [beds, 4].min
```
natdeanlewissoftwire: Beds is at most 4 in the rent range data - Jemma is currently checking that 4 beds in the data means 4 or greater in the property but this is the assumption we have agreed makes sense at the moment

app/models/lettings_log.rb
natdeanlewissoftwire: Yes - we don't ask a question about beds for supported housing logs, and this corresponds to the rows in the rent range data with beds = 0

app/models/lettings_log.rb
natdeanlewissoftwire: Jemma has given me some more behaviour to add in so this is not ready to merge

app/models/lettings_log.rb
jamesrose: Do we need to do this at all, then? Look at the method `la` that delegates to `location.location_code` if one is set, otherwise uses `la`.

app/models/lettings_log.rb
jamesrose: Would `beds` not be zero already, in that case?

app/models/lettings_log.rb
natdeanlewissoftwire: This is now in, ready to merge again

app/models/lettings_log.rb
natdeanlewissoftwire: Nope - just checked, `beds` is nil for supported housing

app/models/lettings_log.rb
natdeanlewissoftwire: Ah you're right, thanks 👍

app/models/lettings_log.rb
```diff
@@ -435,13 +435,22 @@ def self.to_csv(user = nil)
...
   end
 
+  def validation_beds
+    max_beds_in_rent_ranges = 4
+    if needstype == 2
+      0
+    else
+      beds.nil? ? nil : [beds, max_beds_in_rent_ranges].min
+    end
+  end
```
jamesrose:
```suggestion
  def beds_for_la_rent_range
    return 0 if is_supported_housing?
        
    beds.nil? ? nil : [beds, 4].min
  end
```

I think I'd re-write this slightly to improve readability. What do you think?

app/models/validations/financial_validations.rb
```diff
@@ -181,20 +181,34 @@ def validate_rent_range(record)
...
-    rent_range = LaRentRange.find_by(start_year: collection_year, la: record.la, beds: record.beds, lettype: record.lettype)
+
+    max_beds_in_rent_ranges = 4
+    validation_beds = if record.needstype == 2
+                        0
+                      else
+                        record.beds.nil? ? nil : [record.beds, max_beds_in_rent_ranges].min
+                      end
+
+    rent_range = LaRentRange.find_by(start_year: collection_year, la: record.la, beds: validation_beds, lettype: record.lettype)
```
jamesrose:
```suggestion
    rent_range = LaRentRange.find_by(start_year: collection_year, la: record.la, beds: record.beds_for_la_rent_range, lettype: record.lettype)
```

Do you need to repeat yourself here?

app/models/validations/financial_validations.rb
```diff
@@ -181,20 +181,34 @@ def validate_rent_range(record)
...
         record.errors.add :beds, I18n.t("validations.financial.brent.beds.below_hard_min")
         record.errors.add :la, I18n.t("validations.financial.brent.la.below_hard_min")
+        record.errors.add :postcode_known, I18n.t("validations.financial.brent.postcode_known.below_hard_min")
+        record.errors.add :scheme_id, I18n.t("validations.financial.brent.scheme_id.below_hard_min")
+        record.errors.add :location_id, I18n.t("validations.financial.brent.location_id.below_hard_min")
         record.errors.add :rent_type, I18n.t("validations.financial.brent.rent_type.below_hard_min")
         record.errors.add :needstype, I18n.t("validations.financial.brent.needstype.below_hard_min")
         record.errors.add :period, I18n.t("validations.financial.brent.period.below_hard_min")
-      else
+      elsif !(record.beds.present? && record.beds > max_beds_in_rent_ranges)
```
jamesrose:
```suggestion
      elsif record.beds.blank? || record.beds < LaRentRange::MAX_BEDS
```

I think this might be better on the `LaRentRange` model itself, otherwise you have a magic `4` repeated twice in the codebase. We can also flip the logic to make it more readable. What do you think?

app/models/validations/soft_validations.rb
```diff
@@ -28,15 +28,20 @@ def net_income_in_soft_min_range?
...
 
-    rent_range = LaRentRange.find_by(start_year: collection_start_year, la:, beds:, lettype: get_lettype)
-    rent_range.present? && weekly_value(brent).between?(rent_range.soft_max, rent_range.hard_max)
+    rent_range = LaRentRange.find_by(start_year: collection_start_year, la:, beds: validation_beds, lettype: get_lettype)
+    max_beds_in_rent_ranges = 4
+    if beds.present? && beds > max_beds_in_rent_ranges
+      rent_range.present? && weekly_value(brent) > rent_range.soft_max
+    else
+      rent_range.present? && weekly_value(brent).between?(rent_range.soft_max, rent_range.hard_max)
+    end
```
jamesrose:
```suggestion
    if beds.present? && rent_range.present? && beds > LaRentRange::MAX_BEDS
      weekly_value(brent) > rent_range.soft_max
    else
      weekly_value(brent).between?(rent_range.soft_max, rent_range.hard_max)
    end
```

You can move `rent_range.present?` to the first condition as it is shared by both branches.

natdeanlewissoftwire: Won't this give an error in the else section if `rent_range` is not present when we try to access `.soft_max` or `.hard_max`?

app/models/validations/financial_validations.rb
natdeanlewissoftwire: Yep much nicer, wasn't sure where to put this so ended up duplicating.

app/models/validations/financial_validations.rb
natdeanlewissoftwire: Nope, thanks :D

app/models/lettings_log.rb
natdeanlewissoftwire: That's better, I've used the MAX_BEDS value here too ofc :)

app/models/validations/soft_validations.rb
jamesrose: Ah yes! You're right. I think just the `LaRentRange::MAX_BEDS` change, then.

jamesrose (approved):
Nice work!

PR #966, opened 2022-11-03
[CLDC-1468] Add postcode to sales property info journey Q14

mseedat-moj:
Ticket https://digital.dclg.gov.uk/jira/browse/CLDC-1468

## Summary

Introduces a new Property Postcode question screen which uses Postcode service to retrieve Postcode Outcode/Incode/Local Authority

## Note

- PostcodeService now returns an OpenStruct object which begins the refactoring journey to allow behaviours on the returned postcode data e.g. `.result?`
- `la_known` is also set
- `la` is also automatically set - however this may be unnecessary as the next question explicitly asks the user to enter the local authority involved in the sale
- Follows lettings log pattern for postcode setting
- Assumes the postcode is correct if the Postcode Service is not reachable/times-out but the postcode outcode/incode is not set.

## Screenshots
Add postcode
Screenshot 2022-11-04 at 12 45 11

Local authority is automatically set via postcode entry
Screenshot 2022-11-04 at 12 45 17


app/services/postcode_service.rb
```diff
@@ -8,18 +8,24 @@ def lookup(postcode)
...
     if postcode_lookup && postcode_lookup.info.present?
-      {
+      OpenStruct.new({
         location_code: postcode_lookup.codes["admin_district"],
         location_admin_district: postcode_lookup&.admin_district,
-      }
+        incode: postcode_lookup.incode,
+        outcode: postcode_lookup.outcode,
+        result?: postcode_lookup.outcode.present?,
+      })
```
mseedat-moj, line 27: Introduced OpenStruct to begin process of refactoring the response into an Object which can have behaviours such as `.result?`

app/models/sales_log.rb
```diff
@@ -47,4 +55,66 @@ def not_started?
...
+  def reset_postcode_fields
+    self.pcodenk = true
+    self.postcode_known = 0
+    self.pcode1 = nil
+    self.pcode2 = nil
+    self.la = nil
+    self.la_known = 0
+  end
+
+  def postcode_lookup
```
jamesrose, line 94: How much of this can be put on `Log` and shared between each log type?

kosiakkatrina (commented):
There are a few design differences from the lettings log, I'm not sure if we want to keep the design consistent between the two types of logs, but if that was a conscious decision feel free to ignore the following:
1. If the LA is inferred from the postcode we would show it underneath the postcode as such. I think the reasoning for that being not to confuse the inputter cause they've never actually entered that value. Also, do we want the LA to be editable if it was inferred?
image

2. The postcode input box is wider in sales logs, the width is 5 in lettings logs and 10 in sales logs. Maybe worth changing one of them to make it consistent

PR #970, opened 2022-11-07
[CLDC-1627] Experimental alteration of text for answers/questions

mseedat-moj:
Introduces an alternative mechanism to have a different display name from the question name


asmega:
closing stale PR

PR #971, opened 2022-11-07
[CLDC-122] Experimental background job lettings log import with basic reporting

mseedat-moj:
Spawns background jobs to generate lettings log imports due to the amount of time the PostcodeService can take to retrieve/validate postcode information.

This meant introducing a basic summary reporting system that doesn't rely on logs only.


mseedat-moj:
This is an experimental PR because It introduces Wisper etc to allow summary generation. A more basic version was already completed: #912

PR #972, opened 2022-11-08
Blank out all invalid fields for in-progress logs

jamesrose:
Only import fields from the old CORE service that are logically valid. If a log is in a `saved` or `submitted-invalid` then it might have fields that are logically invalid at import. The problem here is that while the old CORE service will accept logically invalid data, the new service will reject it on creation.

This PR blanks out those fields so that they can be saved. The expectation is that the organisation will re-enter the information.

### DIFF
--- a/app/services/imports/lettings_logs_import_service.rb
+++ b/app/services/imports/lettings_logs_import_service.rb
@@ -247,12 +247,12 @@ module Imports
       apply_date_consistency!(attributes)
       apply_household_consistency!(attributes)
 
-      lettings_log = save_lettings_log(attributes)
+      lettings_log = save_lettings_log(attributes, previous_status)
       compute_differences(lettings_log, attributes)
       check_status_completed(lettings_log, previous_status) unless @logs_overridden.include?(lettings_log.old_id)
     end
 
-    def save_lettings_log(attributes)
+    def save_lettings_log(attributes, previous_status)
       lettings_log = LettingsLog.new(attributes)
       begin
         lettings_log.save!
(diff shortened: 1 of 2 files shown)


app/services/imports/lettings_logs_import_service.rb
```diff
@@ -264,21 +264,29 @@ def save_lettings_log(attributes)
...
         @logs_overridden << lettings_log.old_id
         attributes.delete("referral")
-        save_lettings_log(attributes)
+        save_lettings_log(attributes, previous_status)
       elsif lettings_log.errors.of_kind?(:referral, :internal_transfer_fixed_or_lifetime)
         @logger.warn("Log #{lettings_log.old_id}: Removing internal transfer referral since previous tenancy is fixed terms or lifetime")
         @logs_overridden << lettings_log.old_id
         attributes.delete("referral")
-        save_lettings_log(attributes)
+        save_lettings_log(attributes, previous_status)
```
kosiakkatrina, line 289: Unless I'm misunderstanding, these 2 conditions are doing the same thing with the exception of a different error getting logged? Can we have these 2 in one condition and pass the type of error to the error message? (Is it a good place to store the error message in locales?)

kosiakkatrina: Is there a situation where the log would trigger the first if statement on the first run, then rerun, and trigger the elsif on the second run?

jamesrose: @kosiakkatrina In the end you still need a condition for to check which field is erroring, so it didn't seem worth the refactor.

jamesrose: @kosiakkatrina To your second question: no, it won't be possible. We clear all erroring fields on the first condition so there will be no errors a second time. On the middle and final conditions we clear `referral` so it wouldn't trigger again.

kosiakkatrina (approved):

PR #973, opened 2022-11-08
CLDC-1618 Add cookie banner

CheshireSwift:
Was previously making use of `govuk_cookie_banner` helper, but it didn't support wrapping the button group in a form, so have just added a partial with explicit HTML.

Have verified that the respective buttons on the banner behave, and that no (GA) cookies are set when the user refuses them.

### DIFF
--- a/app/views/cookies/_banner.html.erb
+++ b/app/views/cookies/_banner.html.erb
@@ -1,0 +1,48 @@
+<div class="govuk-cookie-banner js-cookie-banner" role="region" aria-label="Cookies on CORE">
+  <div class="govuk-cookie-banner__message govuk-width-container js-cookie-banner__form">
+    <div class="govuk-grid-row">
+      <div class="govuk-grid-column-two-thirds">
+        <h2 class="govuk-cookie-banner__heading govuk-heading-m">Cookies on CORE</h2>
+
+        <div class="govuk-cookie-banner__content">
+          <p class="govuk-body">We use some essential cookies to make this service work.</p>
+          <p class="govuk-body">We’d like to use analytics cookies so we can understand how you use the service and make improvements.</p>
+        </div>
+      </div>
+    </div>
+
+    <%= form_with url: cookies_path, class: "js-cookie-form", method: :put do |f| %>
+      <div class="govuk-button-group">
+        <button type="submit" class="govuk-button" name="cookies_form[analytics_consent]" value="on">
+          Accept analytics cookies
+        </button>
+        <button type="submit" class="govuk-button" name="cookies_form[analytics_consent]" value="off">
+          Reject analytics cookies
+        </button>
+        <%= govuk_link_to "View cookies", cookies_path %>
+      </div>
+    <% end %>
+  </div>
+
+  <div class="govuk-cookie-banner__message govuk-width-container js-cookie-banner__success" role="alert" hidden>
+    <div class="govuk-grid-row">
+      <div class="govuk-grid-column-two-thirds">
+
+        <div class="govuk-cookie-banner__content">
+          <p class="govuk-body">
+            <span class="js-cookie-message">
+              You’ve set your cookie preferences.
+            </span>
+            You can <%= govuk_link_to "change your cookie settings", cookies_path %> at any time.
+          </p>
+        </div>
+      </div>
+    </div>
+
+    <div class="govuk-button-group">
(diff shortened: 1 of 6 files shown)


app/views/layouts/application.html.erb
```diff
@@ -72,13 +72,17 @@
       document.body.className = ((document.body.className) ? document.body.className + " js-enabled" : "js-enabled");
     </script>
 
-    <!-- Google Tag Manager (noscript) -->
-    <% if cookies[:accept_analytics_cookies] %>
+    <% if cookies[:accept_analytics_cookies] == "on" %>
```
CheshireSwift, line 75: For this to be relevant the user would have to be following the very specific behaviour of having JS on for the banner and then turning it off, or having it off the whole time and explicitly going to the cookies page to accept cookies. But just in case 🙂

app/views/layouts/application.html.erb
```diff
@@ -25,15 +25,15 @@
...
-    <!-- Google Tag Manager (doesn't store personal info until permission given) -->
-    <script>(function(w,d,s,l,i){w[l]=w[l]||[];w[l].push({'gtm.start':
-    new Date().getTime(),event:'gtm.js'});var f=d.getElementsByTagName(s)[0],
-    j=d.createElement(s),dl=l!='dataLayer'?'&l='+l:'';j.async=true;j.src=
-    'https://www.googletagmanager.com/gtm.js?id='+i+dl;f.parentNode.insertBefore(j,f);
-    })(window,document,'script','dataLayer','<%= gtm_container %>');</script>
-    <!-- End Google Tag Manager -->
-
     <% if cookies[:accept_analytics_cookies] == "on" %>
+      <!-- Google Tag Manager (doesn't store personal info until permission given) -->
```
CheshireSwift, line 29: Moved entire GTM inside cookie pref, otherwise `_ga` cookie gets set.

app/views/layouts/application.html.erb
```diff
@@ -72,13 +72,17 @@
       document.body.className = ((document.body.className) ? document.body.className + " js-enabled" : "js-enabled");
     </script>
 
-    <!-- Google Tag Manager (noscript) -->
-    <% if cookies[:accept_analytics_cookies] %>
+    <% if cookies[:accept_analytics_cookies] == "on" %>
+      <!-- Google Tag Manager (noscript) -->
       <noscript>
```
CheshireSwift, line 77: For this to be relevant the user would have to be following the very specific behaviour of having JS on for the banner and then turning it off, or having it off the whole time and explicitly going to the cookies page to accept cookies. But just in case 🙂

natdeanlewissoftwire:
I wonder if it's worth putting a test that GA cookies do/don't appear - is this something that could be broken in the future or do you think it's safe to assume it won't?

CheshireSwift:
Yep, good shout on a test. I think given the branching cases it might be challenging to hit every possible failure mode, but I can definitely add something in.

ghost (approved):

jamesrose (approved):

app/views/cookies/_banner.html.erb
```diff
@@ -0,0 +1,48 @@
...
+        <h2 class="govuk-cookie-banner__heading govuk-heading-m">Cookies on CORE</h2>
+
+        <div class="govuk-cookie-banner__content">
+          <p class="govuk-body">We use some essential cookies to make this service work.</p>
+          <p class="govuk-body">We’d like to use analytics cookies so we can understand how you use the service and make improvements.</p>
+        </div>
+      </div>
+    </div>
+
+    <%= form_with url: cookies_path, class: "js-cookie-form", method: :put do |f| %>
```
jamesrose, line 14: Ideally this shouldn't take you off the current page when you submit. Do you know why it seems to be doing that?

ghost (approved):

app/views/cookies/_banner.html.erb
CheshireSwift: ~~The entire cookie banner script appears to have gone missing in a rebase 😱~~
*It was deliberately omitted due to an outstanding question, and I forgot to re-add it 😅*

Have it locally, will add back in.

(The rest of the behaviour should still be correct without it, just the page reload and a lack of analytics until they navigate to a new page.)

jamesrose (approved):

PR #975, opened 2022-11-08
CLDC-1698 Update tenancy length validations

davidm-m:
I've got some concerns about the error message now, it's pretty long with a lot of information (see screenshot below). It wouldn't have been accurate to leave it as it was, but maybe there's a better way to word it?

image

### DIFF
--- a/app/models/validations/tenancy_validations.rb
+++ b/app/models/validations/tenancy_validations.rb
@@ -49,6 +49,6 @@ module Validations::TenancyValidations
   end
 
   def min_tenancy_length(record)
-    record.needstype == 2 || record.renttype == 3 ? 1 : 2
+    record.is_supported_housing? || record.renttype == 3 ? 1 : 2
   end
 end
--- a/app/models/validations/tenancy_validations.rb
+++ b/app/models/validations/tenancy_validations.rb
@@ -17,7 +17,7 @@ module Validations::TenancyValidations
           "validations.tenancy.length.shorthold",
           min_tenancy_length: min_tenancy_length(record),
           rent_type: rent_type_mapping(record)[record.rent_type],
-          needs_type: needs_type_mapping[record.needstype]
+          needs_type: needs_type_mapping[record.needstype],
         ),
       },
       {
@@ -26,7 +26,7 @@ module Validations::TenancyValidations
           "validations.tenancy.length.secure",
           min_tenancy_length: min_tenancy_length(record),
           rent_type: rent_type_mapping(record)[record.rent_type],
-          needs_type: needs_type_mapping[record.needstype]
+          needs_type: needs_type_mapping[record.needstype],
         ),
       },
     ]
--- a/app/models/validations/tenancy_validations.rb
+++ b/app/models/validations/tenancy_validations.rb
@@ -16,8 +16,6 @@ module Validations::TenancyValidations
         error: I18n.t(
           "validations.tenancy.length.shorthold",
           min_tenancy_length: min_tenancy_length(record),
-          rent_type: rent_type_mapping(record)[record.rent_type],
-          needs_type: needs_type_mapping[record.needstype],
         ),
       },
       {
@@ -25,8 +23,6 @@ module Validations::TenancyValidations
         error: I18n.t(
           "validations.tenancy.length.secure",
           min_tenancy_length: min_tenancy_length(record),
-          rent_type: rent_type_mapping(record)[record.rent_type],
-          needs_type: needs_type_mapping[record.needstype],
         ),
       },
     ]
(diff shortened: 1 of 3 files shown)


spec/models/validations/tenancy_validations_spec.rb
```diff
@@ -64,11 +69,16 @@
...
+            "validations.tenancy.length.secure",
+            min_tenancy_length: 2,
+            needs_type: "General needs",
+            rent_type: "Affordable Rent"
+          ) }
 
           before { record.tenancy = 1 }
 
-          context "when tenancy length is greater than 1" do
+          context "when tenancy length is less than 2" do
```
davidm-m, line 81: The old test names were the wrong way round, these should be accurate now

app/models/validations/tenancy_validations.rb
```diff
@@ -41,4 +51,26 @@ def validate_joint_tenancy(record)
...
+  def rent_type_mapping(record)
+    {
+      1 => "Affordable Rent",
+      2 => "London Affordable Rent",
+      4 => "London Living Rent",
+      3 => "Rent to Buy",
+      0 => "Social Rent",
+      5 => record.irproduct,
+    }
+  end
```
davidm-m: I'm a bit unhappy with this duplicating the same info that's in the question class, but I don't know if there's a sensible way to have a single source of truth for this.

config/locales/en.yml
```diff
@@ -301,8 +301,8 @@ en:
     tenancy:
       length:
         fixed_term_not_required: "You must only answer the length of the tenancy if it's fixed-term"
-        shorthold: "Enter a tenancy length between 2 and 99 years for a Fixed Term – Assured Shorthold Tenancy (AST)"
-        secure: "Enter a tenancy length between 2 and 99 years (or don't specify the length) for a Secure (including flexible) tenancy"
+        shorthold: "Enter a tenancy length between %{min_tenancy_length} and 99 years for a %{needs_type} %{rent_type} Fixed Term – Assured Shorthold Tenancy (AST)"
+        secure: "Enter a tenancy length between %{min_tenancy_length} and 99 years (or don't specify the length) for a %{needs_type} %{rent_type} Secure (including flexible) tenancy"
```
jamesrose: Do we actually need to include the `%{needs_type} %{rent_type}` part of this, here?

app/models/validations/tenancy_validations.rb
jamesrose: Do the values map directly to what is in `Form::Lettings::Questions::RentType`? Just in a different hash format?

davidm-m: Yes, except when they've chosen "Other intermediate rent product" we display the other name they've entered

config/locales/en.yml
jamesrose:
```suggestion
        secure: "Enter a tenancy length between %{min_tenancy_length} and 99 years (or don't specify the length) for a Secure (including flexible) tenancy of this type"
```

... perhaps we could make it generic like this?

davidm-m: If it doesn't include it it's not technically accurate, because you can have an AST with a tenancy length of 1 year, but only when it is General Needs and Affordable or Social Rent

davidm-m: I think my worry is that a user will complete one log with a tenancy length of 1 year that is accepted, then another that hits this error, and they will be confused why the error message looks like it should apply to the first log too.
The tenancy type, rent type, and needs type all affect whether this error appears so I think the user should be told about them if possible.

```suggestion
        secure: "Enter a tenancy length between %{min_tenancy_length} and 99 years (or don't specify the length) for a tenancy of this type"
```

How about..?

jamesrose: Yes, that's a fair challenge.

app/models/validations/tenancy_validations.rb
jamesrose: Do you know why the values of `LettingsLog::RENT_TYPE_MAPPING_LABELS` are slightly different? For example, `Social Rent` is 1 there but zero here.

Ah, it looks like we map `rent_type` to different variables internally.

```ruby
module DerivedVariables::LettingsLogVariables
  RENT_TYPE_MAPPING = { 0 => 1, 1 => 2, 2 => 2, 3 => 3, 4 => 3, 5 => 3 }.freeze
```

config/locales/en.yml
davidm-m: I do actually quite like this. Also means I can get rid of the rent type and needs type hashes. I'll give this a go 🙂

app/models/validations/tenancy_validations.rb
jamesrose: Oh wait, I'm getting confused between `rent_type` which is our question, and `renttype` which is derived.

davidm-m: Looks like that one's just to discern between Social Rent, Affordable Rent, and Intermediate Rent products?
"Affordable Rent" and "London Affordable Rent" both fall under the umbrella of Affordable Rent products for instance.

app/models/validations/tenancy_validations.rb
```diff
@@ -41,4 +47,8 @@ def validate_joint_tenancy(record)
       record.errors.add :hhmemb, I18n.t("validations.tenancy.joint_more_than_one_member")
     end
   end
+
+  def min_tenancy_length(record)
+    record.needstype == 2 || record.renttype == 3 ? 1 : 2
```
jamesrose:
```suggestion
    record.is_supported_housing? == 2 || record.renttype == 3 ? 1 : 2
```

I try to use our methods if we can. Apart from being easier to read, it helps us if we ever migrate values across forms.

jamesrose (approved):
Just one comment to make the min length method a little clearer, but otherwise, nice work!

PR #979, opened 2022-11-09
Add validation to dependent ccharge field

jamesrose:
Although both fields are on the same page, we need to add a validation here so that we can clear both values on import.

If we don't do this then the validation will continue to trigger and the importer will get stuck in an endless loop.


kosiakkatrina (approved):
Looks good

PR #981, opened 2022-11-09
Cldc 1668 deactivate location

kosiakkatrina:
- [x] When the user clicks on the `Deactivate this location` button they are taken to a page where they can deactivate from the the beginning of the current collection period or from a specific date that's after the beginning of the current collection period
image

- [x] It displays relevant errors
image
image

- [x] After confirming deactivation the user is taken to the location page 
- [x] After confirming deactivation the user can see a confirmation banner
- [x] `Reactivate this location` button is shown on deactivated or soon to deactivate locations

image

- [x] The status tag in location page displays the correct status
image
image
image

- [x] Update the conditional question controller to work with all conditional questions and not just the log form ones
- [x] Add current_collection_start_date to the form handler

This PR does not update the logs or send the notifications to the users

### DIFF
--- a/app/models/form_handler.rb
+++ b/app/models/form_handler.rb
@@ -50,7 +50,7 @@ class FormHandler
   end
 
   def current_collection_start_date
-    Time.utc(current_collection_start_year, 4, 5)
+    Time.utc(current_collection_start_year, 4, 1)
   end
 
   def form_name_from_start_year(year, type)
--- a/app/models/form_handler.rb
+++ b/app/models/form_handler.rb
@@ -49,6 +49,10 @@ class FormHandler
     today < window_end_date ? today.year - 1 : today.year
   end
 
+  def current_collection_start_date
+    Time.utc(current_collection_start_year, 4, 5)
+  end
+
   def form_name_from_start_year(year, type)
     form_mappings = { 0 => "current_#{type}", 1 => "previous_#{type}", -1 => "next_#{type}" }
     form_mappings[current_collection_start_year - year]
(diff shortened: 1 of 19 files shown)


app/models/form_handler.rb
```diff
@@ -49,6 +49,10 @@ def current_collection_start_year
     today < window_end_date ? today.year - 1 : today.year
   end
 
+  def current_collection_start_date
+    Time.utc(current_collection_start_year, 4, 5)
```
natdeanlewissoftwire: I checked this with the design team, it should be 1st April not 5th

kosiakkatrina: Ohh, nice one! Thank you!

app/models/lettings_log.rb
```diff
@@ -38,6 +38,7 @@ class LettingsLog < Log
   scope :filter_by_propcode, ->(propcode) { where("propcode ILIKE ?", "%#{propcode}%") }
   scope :filter_by_postcode, ->(postcode_full) { where("REPLACE(postcode_full, ' ', '') ILIKE ?", "%#{postcode_full.delete(' ')}%") }
   scope :filter_by_location_postcode, ->(postcode_full) { left_joins(:location).where("REPLACE(locations.postcode, ' ', '') ILIKE ?", "%#{postcode_full.delete(' ')}%") }
+  scope :filter_by_before_startdate, ->(date) { left_joins(:location).where("lettings_logs.startdate >= ?", date) }
```
natdeanlewissoftwire: I don't think we're using this anywhere, is that right?

kosiakkatrina: Yeah. Not right now. I was gonna use it to update the logs but probably best to add it separately. Will update

natdeanlewissoftwire (approved):
Looks good to me! I'll copy some more of this across to my branch when it's merged :)

PR #982, opened 2022-11-09
CLDC-1694 Fix issue with non-renewal lettings allowing 'renewal' for vacancy reason

SamSeed-Softwire:
Would it make sense to also write a test to check that no errors are seen when the letting is not a renewal but a different vacancy reason is chosen? Presumably that would be might be overkill?

As well as reviewing my code, I wouldn't mind feedback on my commit messages if they could be improved!

### DIFF
--- a/app/models/validations/property_validations.rb
+++ b/app/models/validations/property_validations.rb
@@ -39,7 +39,7 @@ module Validations::PropertyValidations
       record.errors.add :referral, I18n.t("validations.household.referral.rsnvac_non_temp")
     end
 
-    if record.renewal == 0 && record.rsnvac == 14
+    if record.renewal && record.renewal.zero? && record.rsnvac == 14
       record.errors.add :rsnvac, I18n.t("validations.property.rsnvac.not_a_renewal")
     end
   end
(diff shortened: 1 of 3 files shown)


app/models/validations/property_validations.rb
```diff
@@ -38,6 +38,10 @@ def validate_rsnvac(record)
       record.errors.add :rsnvac, I18n.t("validations.property.rsnvac.referral_invalid")
       record.errors.add :referral, I18n.t("validations.household.referral.rsnvac_non_temp")
     end
+
+    if record.renewal && record.renewal.zero? && record.rsnvac == 14
```
jamesrose:
```suggestion
    if record.renewal.present? && record.renewal.zero? && record.rsnvac == 14
```

`present?` if a good habit, but not a show stopper here.

jamesrose (approved):

jamesrose:
> Would it make sense to also write a test to check that no errors are seen when the letting is not a renewal but a different vacancy reason is chosen?

You could do this, but I suspect it it is already tested implicitly elsewhere.

jamesrose (approved):

PR #983, opened 2022-11-10
Parse meta fields where namespace declarations aren't included

jamesrose:
We import XML files that define logs from the previous CORE service. For an unknown reason, some of these files don't include the required namespace declarations to parse `meta` values. Instead of changing the source, this change introduces a new method that forces the namespace declaration for meta fields.

This allows us to import documents regardless of whether they specify namespace bindings for `meta` or not.

### DIFF
--- a/app/services/imports/import_service.rb
+++ b/app/services/imports/import_service.rb
@@ -19,8 +19,12 @@ module Imports
       end
     end
 
-    def field_value(xml_document, namespace, field)
-      xml_document.at_xpath("//#{namespace}:#{field}")&.text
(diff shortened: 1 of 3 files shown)


ghost (approved):
LGTM but don't we have any test covering these changes?

jamesrose:
@bibblobcode The tests downstream that are in place are sufficient. The test is "we haven't broken the importer" and "we can still get meta values out" which _are_ tested.

davidm-m (approved):
LGTM!

PR #984, opened 2022-11-10
Import locations that belong to now inactive schemes

jamesrose:
We were previously blocking the import of locations that belonged to schemes where the end date was before the current date. This meant that we were unable to import logs that are associated with schemes that _were_ valid in relation to when they were created, but have since expired.

This PR relaxes that constraint so that we can import logs, schemes and locations that have an `end_date` in the past but a `startdate` before that.


davidm-m (approved):
Looks good!

PR #986, opened 2022-11-11
Change type of legacy old_visible_id fields to strings

jamesrose:
These fields are set using IDs from the previous CORE service. There was an incorrect assumption that these fields were integers so we were casting them as such. We have discovered that these fields are strings (e.g., `027`) not integers (e.g., `27`) so this change adjusts our types. We discovered this because on migrating some organisations resources weren't being properly linked.

This PR changes the types of the following columns from integers to strings:

- `organisations.old_visible_id`
- `schemes.old_visible_id`
- `locations.old_visible_id`

Changing a type is an unsafe operation in Postgres so we’ll block each table for read/writes while the operation happens. Fortunately the tables are _tiny_, so it will take only a few ms, which is acceptable. We would [usually add another column, migrate the data and depreciate the old one](https://github.com/ankane/strong_migrations#changing-the-type-of-a-column).

There's a bit more context on this [Slack thread](https://communities-govuk.slack.com/archives/C03QXAE3Y3W/p1668161848022789).

### DIFF
--- a/spec/fixtures/imports/logs/0b4a68df-30cc-474a-93c0-a56ce8fdad3b.xml
+++ b/spec/fixtures/imports/logs/0b4a68df-30cc-474a-93c0-a56ce8fdad3b.xml
@@ -19,7 +19,7 @@
     <FORM>123456</FORM>
     <Landlord source-value="2">2 Local Authority</Landlord>
     <Group>
-      <_1cmangroupcode>123</_1cmangroupcode>
+      <_1cmangroupcode>0123</_1cmangroupcode>
       <_1cschemecode>10</_1cschemecode>
       <schPC/>
       <Q1e>3 No</Q1e>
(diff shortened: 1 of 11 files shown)


spec/fixtures/imports/logs/0b4a68df-30cc-474a-93c0-a56ce8fdad3b.xml
```diff
@@ -19,7 +19,7 @@
     <FORM>123456</FORM>
     <Landlord source-value="2">2 Local Authority</Landlord>
     <Group>
-      <_1cmangroupcode>123</_1cmangroupcode>
+      <_1cmangroupcode>0123</_1cmangroupcode>
```
davidm-m, line 22: Ids in the old system are only ever padded to three digits so I wouldn't expect to see anything like this, but this should still work as a test

jamesrose: I think that's fine. The test here is that we accept _any_ value as a string without any modifications. We're not going to make any assumptions about what that field can look like.

kosiakkatrina (approved):

davidm-m (approved):

PR #980, opened 2022-11-09
Cldc 1671 deactivate scheme

natdeanlewissoftwire:
ticket: https://digital.dclg.gov.uk/jira/browse/CLDC-1671
tested locally
risk of conflict with simultaneous location deactivation work by @kosiakkatrina but we've been communicating to ameliorate this
no docs required

This is quite a large ticket, so this PR doesn't include all the functionality yet - updating logs will be done separately

deactivation:
image

confirmation:
image

success: 
image

PR #987, opened 2022-11-11
CLDC-1666 Update locations table

SamSeed-Softwire:
Remaining work:
* Get confirmation from Nicky/Jemma on whether to use the location id or the location_code (and if so what location_code should be - should it be the local authority code as it currently is, or the actual name of the corresponding local authority, or something else?). Have currently left as location id. Also once confirmed, need to confirm the desired order of columns in the table.
* Potentially add tests to the `spec/features/schemes_spec.rb` to capture more location statuses than just 'active'. Currently waiting for the feature tests to be expanded to cover the flows which would lead these statuses: 'deactivated', 'deactivating_soon' (will be done as part of  https://digital.dclg.gov.uk/jira/browse/CLDC-1668?), 'reactivated', 'reactivating_soon' (will be done as part of https://digital.dclg.gov.uk/jira/browse/CLDC-1670?) and 'incomplete' (will be done as part of  https://digital.dclg.gov.uk/jira/browse/CLDC-1669?)

### DIFF
--- a/app/models/location.rb
+++ b/app/models/location.rb
@@ -384,11 +384,7 @@ class Location < ApplicationRecord
   end
 
   def status(date = Time.zone.now)
-    return :incomplete if
-      postcode.blank? ||
-        units.blank? ||
-        type_of_unit.blank? ||
-        mobility_type.blank?
+    return :incomplete if [postcode, units, type_of_unit, mobility_type].any?(&:blank?)
     return :deactivated if open_deactivation&.deactivation_date.present? && date >= open_deactivation.deactivation_date
     return :deactivating_soon if open_deactivation&.deactivation_date.present? && date < open_deactivation.deactivation_date
     return :reactivating_soon if recent_deactivation&.reactivation_date.present? && date < recent_deactivation.reactivation_date
--- a/app/models/location.rb
+++ b/app/models/location.rb
@@ -384,6 +384,11 @@ class Location < ApplicationRecord
   end
 
   def status(date = Time.zone.now)
+    return :incomplete if
+      postcode.blank? ||
+        units.blank? ||
+        type_of_unit.blank? ||
+        mobility_type.blank?
     return :deactivated if open_deactivation&.deactivation_date.present? && date >= open_deactivation.deactivation_date
     return :deactivating_soon if open_deactivation&.deactivation_date.present? && date < open_deactivation.deactivation_date
     return :reactivating_soon if recent_deactivation&.reactivation_date.present? && date < recent_deactivation.reactivation_date
--- a/app/models/location.rb
+++ b/app/models/location.rb
@@ -384,7 +384,7 @@ class Location < ApplicationRecord
   end
 
   def status(date = Time.zone.now)
-    return :incomplete if [postcode, units, type_of_unit, mobility_type].any?(&:blank?)
+    return :incomplete unless confirmed
     return :deactivated if open_deactivation&.deactivation_date.present? && date >= open_deactivation.deactivation_date
     return :deactivating_soon if open_deactivation&.deactivation_date.present? && date < open_deactivation.deactivation_date
     return :reactivating_soon if recent_deactivation&.reactivation_date.present? && date < recent_deactivation.reactivation_date
--- a/app/views/locations/index.html.erb
+++ b/app/views/locations/index.html.erb
@@ -43,7 +43,7 @@
           <%= table.body do |body| %>
             <%= body.row do |row| %>
               <% row.cell(text: location.id) %>
-              <% row.cell(text: simple_format(location_cell_postcode(location, "/schemes/#{@scheme.id}/locations/#{location.id}"), { class: "govuk-!-font-weight-bold" }, wrapper_tag: "div")) %>
+              <% row.cell(text: simple_format(location_cell_postcode(location, scheme_location_path(@scheme, @location)), { class: "govuk-!-font-weight-bold" }, wrapper_tag: "div")) %>
               <% row.cell(text: status_tag(location.status)) %>
               <% end %>
           <% end %>
@@ -95,11 +95,11 @@
       <%= table.body do |body| %>
         <%= body.row do |row| %>
           <% row.cell(text: location.id) %>
-          <% row.cell(text: simple_format(location_cell_postcode(location, "/schemes/#{@scheme.id}/locations/#{location.id}"), { class: "govuk-!-font-weight-bold" }, wrapper_tag: "div")) %>
+          <% row.cell(text: simple_format(location_cell_postcode(location, scheme_location_path(@scheme, @location)), { class: "govuk-!-font-weight-bold" }, wrapper_tag: "div")) %>
           <% row.cell(text: location.units) %>
           <% row.cell(text: simple_format("<span>#{location.type_of_unit}</span>")) %>
           <% row.cell(text: location.mobility_type) %>
-          <% row.cell(text: simple_format(location_cell_location_admin_district(location, "/schemes/#{@scheme.id}/locations/#{location.id}/edit-local-authority"), wrapper_tag: "div")) %>
+          <% row.cell(text: simple_format(location_cell_location_admin_district(location, "#{scheme_location_path(@scheme, @location)}/edit-local-authority"), wrapper_tag: "div")) %>
           <% row.cell(text: location.startdate&.to_formatted_s(:govuk_date)) %>
           <% end %>
       <% end %>
--- a/app/views/locations/index.html.erb
+++ b/app/views/locations/index.html.erb
@@ -49,7 +49,7 @@
           <% end %>
         <% end %>
       <% end %>
-      <%= govuk_button_link_to "Add a location", new_scheme_location_path(scheme_id: @scheme.id), secondary: true %>
+      <%= govuk_button_link_to "Add a location", new_scheme_location_path(@scheme), secondary: true %>
     </div>
   </div>
 
@@ -105,7 +105,7 @@
       <% end %>
     <% end %>
   <% end %>
-  <%= govuk_button_link_to "Add a location", new_scheme_location_path(scheme_id: @scheme.id), secondary: true %>
+  <%= govuk_button_link_to "Add a location", new_scheme_location_path(@scheme), secondary: true %>
 
 <% end %>
 
--- a/app/views/locations/index.html.erb
+++ b/app/views/locations/index.html.erb
@@ -43,7 +43,7 @@
           <%= table.body do |body| %>
             <%= body.row do |row| %>
               <% row.cell(text: location.id) %>
-              <% row.cell(text: simple_format(location_cell_postcode(location, scheme_location_path(@scheme, @location)), { class: "govuk-!-font-weight-bold" }, wrapper_tag: "div")) %>
+              <% row.cell(text: simple_format(location_cell_postcode(location, scheme_location_path(@scheme, location)), { class: "govuk-!-font-weight-bold" }, wrapper_tag: "div")) %>
               <% row.cell(text: status_tag(location.status)) %>
               <% end %>
           <% end %>
@@ -95,11 +95,11 @@
       <%= table.body do |body| %>
         <%= body.row do |row| %>
           <% row.cell(text: location.id) %>
-          <% row.cell(text: simple_format(location_cell_postcode(location, scheme_location_path(@scheme, @location)), { class: "govuk-!-font-weight-bold" }, wrapper_tag: "div")) %>
+          <% row.cell(text: simple_format(location_cell_postcode(location, scheme_location_path(@scheme, location)), { class: "govuk-!-font-weight-bold" }, wrapper_tag: "div")) %>
           <% row.cell(text: location.units) %>
           <% row.cell(text: simple_format("<span>#{location.type_of_unit}</span>")) %>
           <% row.cell(text: location.mobility_type) %>
-          <% row.cell(text: simple_format(location_cell_location_admin_district(location, "#{scheme_location_path(@scheme, @location)}/edit-local-authority"), wrapper_tag: "div")) %>
+          <% row.cell(text: simple_format(location_cell_location_admin_district(location, "#{scheme_location_path(@scheme, location)}/edit-local-authority"), wrapper_tag: "div")) %>
           <% row.cell(text: location.startdate&.to_formatted_s(:govuk_date)) %>
           <% end %>
       <% end %>
--- a/app/views/locations/index.html.erb
+++ b/app/views/locations/index.html.erb
@@ -28,10 +28,10 @@
         <% end %>
         <%= table.head do |head| %>
           <%= head.row do |row| %>
-            <% row.cell(header: true, text: "Code", html_attributes: {
+            <% row.cell(header: true, text: "Postcode", html_attributes: {
               scope: "col",
             }) %>
-            <% row.cell(header: true, text: "Postcode", html_attributes: {
+            <% row.cell(header: true, text: "Location code", html_attributes: {
               scope: "col",
             }) %>
             <% row.cell(header: true, text: "Status", html_attributes: {
@@ -42,8 +42,8 @@
         <% @locations.each do |location| %>
           <%= table.body do |body| %>
             <%= body.row do |row| %>
-              <% row.cell(text: location.id) %>
               <% row.cell(text: simple_format(location_cell_postcode(location, scheme_location_path(@scheme, location)), { class: "govuk-!-font-weight-bold" }, wrapper_tag: "div")) %>
+              <% row.cell(text: location.id) %>
               <% row.cell(text: status_tag(location.status)) %>
               <% end %>
           <% end %>
--- a/app/views/locations/index.html.erb
+++ b/app/views/locations/index.html.erb
@@ -11,45 +11,102 @@
 
 <%= render partial: "organisations/headings", locals: { main: @scheme.service_name, sub: nil } %>
 
-<div class="govuk-grid-row">
-  <div class="govuk-grid-column-two-thirds-from-desktop">
-    <%= render SubNavigationComponent.new(items: scheme_items(request.path, @scheme.id, "Locations")) %>
+<% if FeatureToggle.new_locations_table_layout_enabled? %>
+  <div class="govuk-grid-row">
+    <div class="govuk-grid-column-two-thirds-from-desktop">
+      <%= render SubNavigationComponent.new(items: scheme_items(request.path, @scheme.id, "Locations")) %>
 
-    <h2 class="govuk-visually-hidden">Locations</h2>
+      <h2 class="govuk-visually-hidden">Locations</h2>
 
-    <%= render SearchComponent.new(current_user:, search_label: "Search by location name or postcode", value: @searched) %>
+      <%= render SearchComponent.new(current_user:, search_label: "Search by location name or postcode", value: @searched) %>
 
-    <%= govuk_section_break(visible: true, size: "m") %>
+      <%= govuk_section_break(visible: true, size: "m") %>
 
-    <%= govuk_table do |table| %>
-      <%= table.caption(classes: %w[govuk-!-font-size-19 govuk-!-font-weight-regular]) do |caption| %>
-        <%= render(SearchResultCaptionComponent.new(searched: @searched, count: @pagy.count, item_label:, total_count: @total_count, item: "locations", path: request.path)) %>
-      <% end %>
-      <%= table.head do |head| %>
-        <%= head.row do |row| %>
-          <% row.cell(header: true, text: "Code", html_attributes: {
-            scope: "col",
-          }) %>
-          <% row.cell(header: true, text: "Postcode", html_attributes: {
-            scope: "col",
-          }) %>
-          <% row.cell(header: true, text: "Status", html_attributes: {
-            scope: "col",
-          }) %>
+      <%= govuk_table do |table| %>
+        <%= table.caption(classes: %w[govuk-!-font-size-19 govuk-!-font-weight-regular]) do |caption| %>
+          <%= render(SearchResultCaptionComponent.new(searched: @searched, count: @pagy.count, item_label:, total_count: @total_count, item: "locations", path: request.path)) %>
+        <% end %>
+        <%= table.head do |head| %>
+          <%= head.row do |row| %>
+            <% row.cell(header: true, text: "Code", html_attributes: {
+              scope: "col",
+            }) %>
+            <% row.cell(header: true, text: "Postcode", html_attributes: {
+              scope: "col",
+            }) %>
+            <% row.cell(header: true, text: "Status", html_attributes: {
+              scope: "col",
+            }) %>
+          <% end %>
+        <% end %>
+        <% @locations.each do |location| %>
+          <%= table.body do |body| %>
+            <%= body.row do |row| %>
+              <% row.cell(text: location.id) %>
+              <% row.cell(text: simple_format(location_cell_postcode(location, "/schemes/#{@scheme.id}/locations/#{location.id}"), { class: "govuk-!-font-weight-bold" }, wrapper_tag: "div")) %>
+              <% row.cell(text: status_tag(location.status)) %>
+              <% end %>
+          <% end %>
         <% end %>
       <% end %>
-      <% @locations.each do |location| %>
-        <%= table.body do |body| %>
-          <%= body.row do |row| %>
-            <% row.cell(text: location.id) %>
-            <% row.cell(text: simple_format(location_cell_postcode(location, "/schemes/#{@scheme.id}/locations/#{location.id}"), { class: "govuk-!-font-weight-bold" }, wrapper_tag: "div")) %>
-            <% row.cell(text: status_tag(location.status)) %>
-            <% end %>
-        <% end %>
+      <%= govuk_button_link_to "Add a location", new_scheme_location_path(scheme_id: @scheme.id), secondary: true %>
+    </div>
+  </div>
+
+<% else %>
+  <%= render SubNavigationComponent.new(items: scheme_items(request.path, @scheme.id, "Locations")) %>
+
+  <h2 class="govuk-visually-hidden">Locations</h2>
+
+  <%= render SearchComponent.new(current_user:, search_label: "Search by location name or postcode", value: @searched) %>
+
+  <%= govuk_section_break(visible: true, size: "m") %>
+
+  <%= govuk_table do |table| %>
+    <%= table.caption(classes: %w[govuk-!-font-size-19 govuk-!-font-weight-regular]) do |caption| %>
+      <%= render(SearchResultCaptionComponent.new(searched: @searched, count: @pagy.count, item_label:, total_count: @total_count, item: "locations", path: request.path)) %>
+    <% end %>
+    <%= table.head do |head| %>
+      <%= head.row do |row| %>
+        <% row.cell(header: true, text: "Code", html_attributes: {
+          scope: "col",
+        }) %>
+        <% row.cell(header: true, text: "Postcode", html_attributes: {
+          scope: "col",
+        }) %>
+        <% row.cell(header: true, text: "Units", html_attributes: {
+          scope: "col",
+        }) %>
+        <% row.cell(header: true, text: "Common unit type", html_attributes: {
+          scope: "col",
+        }) %>
+        <% row.cell(header: true, text: "Mobility type", html_attributes: {
+          scope: "col",
+        }) %>
+        <% row.cell(header: true, text: "Local authority", html_attributes: {
+          scope: "col",
+        }) %>
+        <% row.cell(header: true, text: "Available from", html_attributes: {
+          scope: "col",
+        }) %>
       <% end %>
     <% end %>
-    <%= govuk_button_link_to "Add a location", new_scheme_location_path(scheme_id: @scheme.id), secondary: true %>
-  </div>
-</div>
+    <% @locations.each do |location| %>
+      <%= table.body do |body| %>
+        <%= body.row do |row| %>
+          <% row.cell(text: location.id) %>
+          <% row.cell(text: simple_format(location_cell_postcode(location, "/schemes/#{@scheme.id}/locations/#{location.id}"), { class: "govuk-!-font-weight-bold" }, wrapper_tag: "div")) %>
+          <% row.cell(text: location.units) %>
+          <% row.cell(text: simple_format("<span>#{location.type_of_unit}</span>")) %>
+          <% row.cell(text: location.mobility_type) %>
+          <% row.cell(text: simple_format(location_cell_location_admin_district(location, "/schemes/#{@scheme.id}/locations/#{location.id}/edit-local-authority"), wrapper_tag: "div")) %>
+          <% row.cell(text: location.startdate&.to_formatted_s(:govuk_date)) %>
+          <% end %>
+      <% end %>
+    <% end %>
+  <% end %>
+  <%= govuk_button_link_to "Add a location", new_scheme_location_path(scheme_id: @scheme.id), secondary: true %>
+
+<% end %>
 
 <%== render partial: "pagy/nav", locals: { pagy: @pagy, item_name: "locations" } %>
--- a/app/views/locations/index.html.erb
+++ b/app/views/locations/index.html.erb
@@ -11,17 +11,25 @@
 
 <%= render partial: "organisations/headings", locals: { main: @scheme.service_name, sub: nil } %>
 
-<div class="govuk-grid-row">
-  <div class="govuk-grid-column-two-thirds-from-desktop">
-    <% if FeatureToggle.new_locations_table_layout_enabled? %>
-      <%= render SubNavigationComponent.new(items: scheme_items(request.path, @scheme.id, "Locations")) %>
+<% if FeatureToggle.new_locations_table_layout_enabled? %>
+  <div class="govuk-grid-row">
+    <div class="govuk-grid-column-two-thirds-from-desktop">
+<% end %>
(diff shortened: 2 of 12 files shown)


app/views/locations/index.html.erb
```diff
@@ -11,57 +11,45 @@
...
-      <%= body.row do |row| %>
-        <% row.cell(text: location.id) %>
-        <% row.cell(text: simple_format(location_cell_postcode(location, "/schemes/#{@scheme.id}/locations/#{location.id}"), { class: "govuk-!-font-weight-bold" }, wrapper_tag: "div")) %>
-        <% row.cell(text: location.units) %>
-        <% row.cell(text: simple_format("<span>#{location.type_of_unit}</span>")) %>
-        <% row.cell(text: location.mobility_type) %>
-        <% row.cell(text: simple_format(location_cell_location_admin_district(location, "/schemes/#{@scheme.id}/locations/#{location.id}/edit-local-authority"), wrapper_tag: "div")) %>
-        <% row.cell(text: location.startdate&.to_formatted_s(:govuk_date)) %>
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds-from-desktop">
```
SamSeed-Softwire: Given there aren't many columns in the table now, Nick and Nicky suggested making it smaller, so I've applied a 2/3 scaling when the user is using a desktop device.

app/views/locations/index.html.erb
```diff
@@ -11,57 +11,45 @@
...
+          <% row.cell(header: true, text: "Status", html_attributes: {
+            scope: "col",
+          }) %>
+        <% end %>
+      <% end %>
+      <% @locations.each do |location| %>
+        <%= table.body do |body| %>
+          <%= body.row do |row| %>
+            <% row.cell(text: location.id) %>
+            <% row.cell(text: simple_format(location_cell_postcode(location, "/schemes/#{@scheme.id}/locations/#{location.id}"), { class: "govuk-!-font-weight-bold" }, wrapper_tag: "div")) %>
```
SamSeed-Softwire: Location name is displayed underneath the postcode, instead of in a separate column, due to a steer from Nick.

spec/models/location_spec.rb
```diff
@@ -119,6 +119,34 @@
...
+      end
+
+      it "returns incomplete when the mobility type is missing" do
+        location.mobility_type = nil
+        expect(location.status).to eq(:incomplete)
+        location.mobility_type = ' '
+        expect(location.status).to eq(:incomplete)
+      end
+    end
+
```
SamSeed-Softwire, line 148: The fields which are (currently) mandatory when setting up a new location are postcode, units, type_of_unit and mobility_type. Do you think adding tests for the fields which will be made mandatory by https://digital.dclg.gov.uk/jira/browse/CLDC-1669 should be done as part of that ticket, or this one?

spec/features/schemes_spec.rb
```diff
@@ -204,9 +204,8 @@
                 locations.each do |location|
                   expect(page).to have_content(location.id)
                   expect(page).to have_content(location.postcode)
-                  expect(page).to have_content(location.units)
-                  expect(page).to have_content(location.type_of_unit)
-                  expect(page).to have_content(location.startdate&.to_formatted_s(:govuk_date))
+                  expect(page).to have_content(location.name)
+                  expect(page).to have_content("Active")
```
SamSeed-Softwire, line 208: I wasn't sure if there was a better way to do this!

app/views/locations/index.html.erb
jamesrose:
```suggestion
            <% row.cell(text: simple_format(location_cell_postcode(location, scheme_location_path(@scheme, @location)), { class: "govuk-!-font-weight-bold" }, wrapper_tag: "div")) %>
```

We can use Rails route helpers here.

spec/models/location_spec.rb
```diff
@@ -119,6 +119,34 @@
       Timecop.freeze(2022, 6, 7)
     end
 
+    context "when location is missing some mandatory information" do
+      it "returns incomplete when the postcode is missing" do
+        location.postcode = nil
+        expect(location.status).to eq(:incomplete)
+        location.postcode = ' '
```
jamesrose:
```suggestion
        location.postcode = ""
```

spec/models/location_spec.rb
```diff
@@ -119,6 +119,34 @@
...
+
+      it "returns incomplete when the number of units is missing" do
+        location.units = nil
+        expect(location.status).to eq(:incomplete)
+      end
+
+      it "returns incomplete when the unit type is missing" do
+        location.type_of_unit = nil
+        expect(location.status).to eq(:incomplete)
+        location.type_of_unit = ' '
```
jamesrose:
```suggestion
        location.type_of_unit = ""
```

spec/models/location_spec.rb
```diff
@@ -119,6 +119,34 @@
...
+        expect(location.status).to eq(:incomplete)
+      end
+
+      it "returns incomplete when the mobility type is missing" do
+        location.mobility_type = nil
+        expect(location.status).to eq(:incomplete)
+        location.mobility_type = ' '
```
jamesrose:
```suggestion
        location.mobility_type = ""
```

app/views/locations/index.html.erb
```diff
@@ -11,57 +11,45 @@
...
+            <% row.cell(text: status_tag(location.status)) %>
+            <% end %>
         <% end %>
+      <% end %>
     <% end %>
-  <% end %>
-<% end %>
-<%= govuk_button_link_to "Add a location", new_scheme_location_path(scheme_id: @scheme.id), secondary: true %>
+    <%= govuk_button_link_to "Add a location", new_scheme_location_path(scheme_id: @scheme.id), secondary: true %>
```
jamesrose:
```suggestion
    <%= govuk_button_link_to "Add a location", new_scheme_location_path(@scheme), secondary: true %>
```

You can usually inject the object directly into the helper and let Rails figure out the ID.

app/models/location.rb
```diff
@@ -373,6 +373,11 @@ def available_from
   end
 
   def status
+    return :incomplete if
+      postcode.blank? ||
+        units.blank? ||
+        type_of_unit.blank? ||
+        mobility_type.blank?
```
jamesrose: Have you thought about how you might query for active locations? We have decided to calculate the status dynamically. Did you consider having a status column on locations that you update whenever a location is updated?

```suggestion
    return :incomplete if [postcode, units, type_of_unit, mobility_type].any?(&:blank?)
```

We can make this a little shorter with `any?` — what do you think?

app/views/locations/index.html.erb
```diff
@@ -11,57 +11,45 @@
...
-    <%= table.body do |body| %>
-      <%= body.row do |row| %>
-        <% row.cell(text: location.id) %>
-        <% row.cell(text: simple_format(location_cell_postcode(location, "/schemes/#{@scheme.id}/locations/#{location.id}"), { class: "govuk-!-font-weight-bold" }, wrapper_tag: "div")) %>
-        <% row.cell(text: location.units) %>
-        <% row.cell(text: simple_format("<span>#{location.type_of_unit}</span>")) %>
-        <% row.cell(text: location.mobility_type) %>
-        <% row.cell(text: simple_format(location_cell_location_admin_district(location, "/schemes/#{@scheme.id}/locations/#{location.id}/edit-local-authority"), wrapper_tag: "div")) %>
-        <% row.cell(text: location.startdate&.to_formatted_s(:govuk_date)) %>
+<div class="govuk-grid-row">
```
jamesrose: These changes should probably be behind a feature flag.

app/models/location.rb
SamSeed-Softwire: @jamesrose I hadn't considered that - I had seen that filtering for active locations was done using a scope, and accepted that as the way things were done. I will make sure to consider this as a potential option in future work.

I've had a read through the [thread](https://softwire.slack.com/archives/C03QXAE3Y3W/p1668425372916019) in Slack and think I understand it. I assume no further changes are needed on the above bit of code right now? I could add a scope to filter for incomplete locations (for example) but assuming that it would be better to do that if/when the need arises, rather than now?

Thanks!

spec/models/location_spec.rb
SamSeed-Softwire: @jamesrose I wrote ' ' to capture the case where the postcode is actual whitespace rather than an empty string. Is this just overkill do you think?

app/helpers/tag_helper.rb
```diff
@@ -7,6 +7,9 @@ module TagHelper
     in_progress: "In progress",
     completed: "Completed",
     active: "Active",
+    incomplete: "Incomplete",
+    activating_soon: "Activating soon",
+    reactivating_soon: "Reactivating soon",
```
natdeanlewissoftwire: Do we need both `activating_soon` and `reactivating_soon `? I think activating/reactivating might be talking about the same thing, but perhaps have been mistakenly called different things in different areas of the designs. AFAIK both just mean a scheme/location is currently inactive but has a reactivation date in the future.

config/initializers/feature_toggle.rb
```diff
@@ -20,4 +20,10 @@ def self.location_toggle_enabled?
 
     false
   end
+
+  def self.new_locations_table_layout_enabled?
+    return true unless Rails.env.production?
+
+    false
```
natdeanlewissoftwire, line 27: i know you're just copying old code (I did the same) but it occurs to me every time i see these they could just be 
```suggestion
    def self.[method_name]?
        !Rails.env.production?
    end
```

config/initializers/feature_toggle.rb
```diff
@@ -20,4 +20,10 @@ def self.location_toggle_enabled?
 
     false
   end
+
+  def self.new_locations_table_layout_enabled?
```
natdeanlewissoftwire, line 24: i know you're just copying old code (I did the same) but it occurs to me every time i see these they could just be 
```suggestion
  def self.new_locations_table_layout_enabled?
       !Rails.env.production?
  end
```

config/initializers/feature_toggle.rb
```diff
@@ -20,4 +20,10 @@ def self.location_toggle_enabled?
...
+    return true unless Rails.env.production?
+
+    false
+  end
```
natdeanlewissoftwire, line 28: I know this is just duplicating previous methods (I did the same for a scheme_toggle one) but these could probably be 
```suggestion
def self.location_toggle_enabled?
    !Rails.env.production?
  end
```
etc.

natdeanlewissoftwire: I know this is just duplicating previous methods (I did the same for a scheme_toggle one) but these could probably be 
```suggestion
def self.location_toggle_enabled?
    !Rails.env.production?
  end
```
etc.

natdeanlewissoftwire: (I've changed the other ones here but if you could update yours too that would be good I think https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/980/files#diff-a32908531df21028fc3e01feaacaf51a0a6f36f976d0a1e13676e8d1d5caf20aR6)

app/helpers/tag_helper.rb
natdeanlewissoftwire: (there's a few other places where these both pop up in this PR that would be affected if I'm right that we only need one)

SamSeed-Softwire: Have asked Nicky about this.

natdeanlewissoftwire: turns out activating soon is required when a scheme is set up to activate in the future but has never been deactivated before

spec/requests/locations_controller_spec.rb
```diff
@@ -972,12 +977,19 @@
...
           expect(page).to have_content(location.id)
           expect(page).to have_content(location.postcode)
-          expect(page).to have_content(location.type_of_unit)
-          expect(page).to have_content(location.startdate&.to_formatted_s(:govuk_date))
+          if FeatureToggle.new_locations_table_layout_enabled?
+            expect(page).to have_content(location.name)
+            expect(page).to have_content(location.status)
+          else
+            expect(page).to have_content(location.type_of_unit)
+            expect(page).to have_content(location.mobility_type)
```
SamSeed-Softwire: Testing for mobility_type and location_admin_district was previously not done here, not sure why, so I've added it in. Shout if you disagree!

app/views/locations/index.html.erb
```diff
@@ -11,57 +11,102 @@
 
 <%= render partial: "organisations/headings", locals: { main: @scheme.service_name, sub: nil } %>
 
-<%= render SubNavigationComponent.new(items: scheme_items(request.path, @scheme.id, "Locations")) %>
+<% if FeatureToggle.new_locations_table_layout_enabled? %>
+  <div class="govuk-grid-row">
+    <div class="govuk-grid-column-two-thirds-from-desktop">
```
kosiakkatrina, line 16: I would personally remove this div because it squishes the table (unless that's in the designs)

app/views/locations/index.html.erb
```diff
@@ -11,57 +11,102 @@
 
 <%= render partial: "organisations/headings", locals: { main: @scheme.service_name, sub: nil } %>
 
-<%= render SubNavigationComponent.new(items: scheme_items(request.path, @scheme.id, "Locations")) %>
+<% if FeatureToggle.new_locations_table_layout_enabled? %>
```
kosiakkatrina: I would move the feature toggle just above the table ( `<%= govuk_table do |table| %>` ) and only conditionally display the table content instead of the whole page

kosiakkatrina: A bit less repetition methinks

spec/features/schemes_spec.rb
```diff
@@ -204,9 +204,14 @@
...
-                  expect(page).to have_content(location.units)
-                  expect(page).to have_content(location.type_of_unit)
-                  expect(page).to have_content(location.startdate&.to_formatted_s(:govuk_date))
+                  if FeatureToggle.new_locations_table_layout_enabled?
+                    expect(page).to have_content(location.name)
+                    expect(page).to have_content("Active")
+                  else
+                    expect(page).to have_content(location.units)
+                    expect(page).to have_content(location.type_of_unit)
+                    expect(page).to have_content(location.startdate&.to_formatted_s(:govuk_date))
```
kosiakkatrina: Does this ever get executed?

kosiakkatrina: Cause I'd expect out feature toggle to always be on in test environments, unless we mock `new_locations_table_layout_enabled?`

spec/requests/locations_controller_spec.rb
```diff
@@ -867,14 +867,19 @@
...
-          expect(page).to have_content(location.location_admin_district)
-          expect(page).to have_content(location.startdate&.to_formatted_s(:govuk_date))
+          if FeatureToggle.new_locations_table_layout_enabled?
+            expect(page).to have_content(location.name)
+            expect(page).to have_content(location.status)
+          else
+            expect(page).to have_content(location.type_of_unit)
+            expect(page).to have_content(location.mobility_type)
+            expect(page).to have_content(location.location_admin_district)
+            expect(page).to have_content(location.startdate&.to_formatted_s(:govuk_date))
```
kosiakkatrina: Same as previous comment

spec/requests/locations_controller_spec.rb
```diff
@@ -972,12 +977,19 @@
...
-          expect(page).to have_content(location.type_of_unit)
-          expect(page).to have_content(location.startdate&.to_formatted_s(:govuk_date))
+          if FeatureToggle.new_locations_table_layout_enabled?
+            expect(page).to have_content(location.name)
+            expect(page).to have_content(location.status)
+          else
+            expect(page).to have_content(location.type_of_unit)
+            expect(page).to have_content(location.mobility_type)
+            expect(page).to have_content(location.location_admin_district)
+            expect(page).to have_content(location.startdate&.to_formatted_s(:govuk_date))
```
kosiakkatrina: Same as above

spec/features/schemes_spec.rb
SamSeed-Softwire: My understanding is that the tests would get run when deployed to production, where the feature toggle is turned off. I posted in Slack about this earlier today, shout if you disagree with the response there! https://softwire.slack.com/archives/C03QXAE3Y3W/p1668599965402619

app/views/locations/index.html.erb
SamSeed-Softwire: The reason I did it above `<%= render SubNavigationComponent.new(items: scheme_items(request.path, @scheme.id, "Locations")) %>` rather than just above the table was because I wanted the 2/3 scaling to also only be shown when the feature toggle is enabled, and the 2/3 scaling should include the stuff just above the table as well I think. There's also the question of whether we want to apply the 2/3 scaling at all, but ignoring that for now do you agree with my reasoning?

app/views/locations/index.html.erb
SamSeed-Softwire: This wasn't in the Figma designs but I did it based on conversations with Nick and Nicky who thought it might be sensible to reduce the amount of empty space on screen for desktop devices. Will show it to them and see what they think, to confirm they're happy with what I've done.

SamSeed-Softwire: I've spoken to Nicky and she confirmed she wanted the 2/3 scaling for the new table layout because otherwise the table has too much empty space. She also said that she wanted the scheme details page (i.e. the page you get to when clicking 'Schemes' just to the left of locations) to have the same scaling.

However when using the old table layout for the locations table there are a lot more columns so it looks too squeezed if using the 2/3 scaling. So if the feature toggle is off then I don't apply the 2/3 scaling to either screen.

image

app/views/locations/index.html.erb
SamSeed-Softwire: @kosiakkatrina I've updated this now to remove repetition, lmk what you think. Thanks!

db/schema.rb
```diff
@@ -360,19 +360,19 @@
...
     t.integer "details_known_2"
     t.integer "details_known_3"
     t.integer "details_known_4"
-    t.integer "age4"
-    t.integer "age4_known"
-    t.integer "age5"
-    t.integer "age5_known"
-    t.integer "age6"
-    t.integer "age6_known"
     t.index ["created_by_id"], name: "index_sales_logs_on_created_by_id"
```
SamSeed-Softwire: Not sure why these got rearranged...

app/models/location.rb
```diff
@@ -374,6 +374,7 @@ def available_from
   end
 
   def status
+    return :incomplete if [postcode, units, type_of_unit, mobility_type].any?(&:blank?)
     return :active if deactivation_date.blank?
     return :deactivating_soon if Time.zone.now < deactivation_date
```
natdeanlewissoftwire, line 380: do we need returns here for the other statuses?

app/views/locations/index.html.erb
kosiakkatrina: Oh, sorry, I think this is a different page I had in mind. Isn't this page rendering this? 
image

kosiakkatrina: Wait maybe I misread your comment

spec/features/schemes_spec.rb
kosiakkatrina: So I think the `RAILS_ENV` for the tests would always be `test`, I think we set that in `/spec/rails_helper.rb`.

We could do something like this to mock `RAILS_ENV`
```
allow(ENV).to receive(:[])
allow(ENV).to receive(:[]).with("RAILS_ENV").and_return('production')
```

Or we could mock the `new_locations_table_layout_enabled?` behaviour I think by doing something like this:
```
allow(FeatureToggle).to receive(: new_locations_table_layout_enabled?).and_return(false)
```

This way we can test the behaviour on production locally 

Similar to what we're doing here:
https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/blob/main/spec/requests/sales_logs_controller_spec.rb#L31

app/models/location.rb
```diff
@@ -374,6 +374,7 @@ def available_from
   end
 
   def status
+    return :incomplete if [postcode, units, type_of_unit, mobility_type].any?(&:blank?)
     return :active if deactivation_date.blank?
```
SamSeed-Softwire: @natdeanlewissoftwire I can't find your comment now ("do we need returns here for the other statuses?") - not sure if you deleted it or if it was because I force pushed? Although I'd be surprised if force pushing caused the comment to be deleted.

Anyways, I was trying to keep the scope of this ticket to be just the display of the statuses. Under that logic I arguably didn't need to add this line returning the incomplete status... I think it was because I knew the de/reactivating logic was being done in other tickets but I didn't think the incomplete status was covered elsewhere currently.

Happy to remove. I've added a comment on https://digital.dclg.gov.uk/jira/browse/CLDC-1669 now anyway indicating that the incomplete status logic will need to be updated as part of that ticket.

app/views/locations/index.html.erb
kosiakkatrina: Yup, that makes sense!

app/views/locations/index.html.erb
kosiakkatrina: It's fine if that's in designs

SamSeed-Softwire: Sorry if my comment was confusing. The picture above was just to demonstrate the other page (scheme details) which I've _also_ made 2/3 scaling now (in order to match the locations list table) following a decision from Nicky. Happy with the above?

SamSeed-Softwire: Ah sorry I hadn't seen your comment "It's fine if that's in designs" because you made it since I last refreshed the page. Thanks.

app/models/location.rb
natdeanlewissoftwire: Yep - I deleted as I re-read your description of the PR and saw that you left out `activating soon` and `reactivating soon` intentionally, I agree that's the right thing to do :D

app/views/schemes/show.html.erb
```diff
@@ -10,19 +10,28 @@
 
 <%= render partial: "organisations/headings", locals: { main: @scheme.service_name, sub: nil } %>
 
-<%= render SubNavigationComponent.new(items: scheme_items(request.path, @scheme.id, "Locations")) %>
+<% if FeatureToggle.new_locations_table_layout_enabled? %>
```
natdeanlewissoftwire: I think you could probably just use the `location_toggle_enabled` right? Or is there a reason this behaviour has to be displayed separately to the deactivating/reactivating locations behaviour?

natdeanlewissoftwire: There's maybe also an argument for `location_toggle_enabled` and `scheme_toggle_enabled` being the same as well to be honest but maybe those are a bit more independent

spec/features/schemes_spec.rb
natdeanlewissoftwire: I think this is fine, but if you wanted to test all these at once I think you could do something like  
```
page.should satisfy {|page| page.has_content?('foo') and page.has_content?('bar')}
```

https://www.rubydoc.info/github/rspec/rspec-expectations/RSpec%2FMatchers:satisfy

app/views/schemes/show.html.erb
```diff
@@ -10,19 +10,28 @@
...
+  <div class="govuk-grid-row">
+    <div class="govuk-grid-column-two-thirds-from-desktop">
```
natdeanlewissoftwire, line 15: Also, why are you changing the schemes/show page, I can't see any changes to this page in the designs for the ticket?

SamSeed-Softwire: I had an in-person chat with Nicky last night where she gave a steer to make the the schemes/show page match the locations list page (just in terms of the 2/3 scaling) :)

db/schema.rb
natdeanlewissoftwire: I think it's just down to migration orders, happens to me sometimes but it's not a problem as long as we're not losing/adding fields we don't expect to

app/views/schemes/show.html.erb
SamSeed-Softwire: I hadn't looked at `location_toggle_enabled`, tbh I was assuming each separate ticket/change in functionality would have a separate feature toggle. But are you saying that all of the current changes to the locations pages should all be under the same feature toggle? If so happy to just use `location_toggle_enabled` :).

db/schema.rb
SamSeed-Softwire: Thanks! In general would you commit changes like this?

spec/features/schemes_spec.rb
SamSeed-Softwire: Thanks - is there an advantage to doing it this way? It seems pretty similar to me.

natdeanlewissoftwire: I don't think so, leave it as it is imo

db/schema.rb
natdeanlewissoftwire: Yep, schema should be committed as a rule (though I don't think it actually matters in this case but simpler to just always commit it)

spec/models/location_spec.rb
SamSeed-Softwire: Made the decision to do this as part of 1669 - have a left a comment on there.

db/schema.rb
SamSeed-Softwire: Cool, thanks!

app/views/schemes/show.html.erb
SamSeed-Softwire: @natdeanlewissoftwire updated this now to use `location_toggle_enabled` instead. I'm using this to toggle the 2/3 scaling on the `app/views/schemes/show.html.erb` page instead of `scheme_toggle_enabled` because I only want to toggle the 2/3 width on the scheme show page when the locations page gets toggled - i.e. it's directly linked to the locations toggle, and not related to the other work on schemes which falls under the schemes toggle.

spec/requests/locations_controller_spec.rb
```diff
@@ -867,7 +867,18 @@
...
+          expect(page).to have_content(location.id)
+          expect(page).to have_content(location.postcode)
+          expect(page).to have_content(location.name)
+          expect(page).to have_content(location.status)
+        end
+      end
+
+      it "shows locations with correct data wben the new locations layout feature toggle is disabled" do
+        allow(FeatureToggle).to receive(:location_toggle_enabled?).and_return(false)
+        get "/schemes/#{scheme.id}/locations"
```
SamSeed-Softwire, line 881: @kosiakkatrina happy with how I've done this? To avoid duplicating lots of code I just repeated the `get "/schemes/#{scheme.id}/locations"` call.

app/views/locations/index.html.erb
```diff
@@ -11,57 +11,102 @@
...
+    <% end %>
+    <% @locations.each do |location| %>
+      <%= table.body do |body| %>
+        <%= body.row do |row| %>
+          <% row.cell(text: location.id) %>
+          <% row.cell(text: simple_format(location_cell_postcode(location, scheme_location_path(@scheme, location)), { class: "govuk-!-font-weight-bold" }, wrapper_tag: "div")) %>
+          <% row.cell(text: location.units) %>
+          <% row.cell(text: simple_format("<span>#{location.type_of_unit}</span>")) %>
+          <% row.cell(text: location.mobility_type) %>
+          <% row.cell(text: simple_format(location_cell_location_admin_district(location, "#{scheme_location_path(@scheme, location)}/edit-local-authority"), wrapper_tag: "div")) %>
```
jamesrose:
```suggestion
          <% row.cell(text: simple_format(location_cell_location_admin_district(location, scheme_location_edit_local_authority_path(@scheme, location), wrapper_tag: "div")) %>
```

This should be a route. If you ever have to construct routes manually, you are likely going astray.

https://guides.rubyonrails.org/routing.html

app/views/locations/index.html.erb
```diff
@@ -11,57 +11,102 @@
...
+        }) %>
+      <% end %>
+    <% end %>
+    <% @locations.each do |location| %>
+      <%= table.body do |body| %>
+        <%= body.row do |row| %>
+          <% row.cell(text: location.id) %>
+          <% row.cell(text: simple_format(location_cell_postcode(location, scheme_location_path(@scheme, location)), { class: "govuk-!-font-weight-bold" }, wrapper_tag: "div")) %>
+          <% row.cell(text: location.units) %>
+          <% row.cell(text: simple_format("<span>#{location.type_of_unit}</span>")) %>
```
jamesrose:
```suggestion
          <% row.cell do %>
            <span><%= simple_format(location.type_of_unit %></span>
          <% end %>
```

I _think_ you can do this. When you are rendering HTML you should try and write it outside of Ruby, if you can.

app/views/schemes/show.html.erb
```diff
@@ -10,19 +10,28 @@
...
-      <% row.value { attr[:name].eql?("Status") ? status_tag(attr[:value]) : details_html(attr) } %>
-      <% row.action(text: "Change", href: scheme_edit_name_path(scheme_id: @scheme.id)) if attr[:edit] %>
-    <% end %>
-  <% end %>
+      <%= govuk_summary_list do |summary_list| %>
+        <% display_scheme_attributes(@scheme).each do |attr| %>
+          <% next if current_user.data_coordinator? && attr[:name] == ("Housing stock owned by") %>
+          <%= summary_list.row do |row| %>
+            <% row.key { attr[:name].eql?("Registered under Care Standards Act 2000") ? "Registered under Care Standards Act 2000" : attr[:name].to_s.humanize } %>
+            <% row.value { attr[:name].eql?("Status") ? status_tag(attr[:value]) : details_html(attr) } %>
```
jamesrose:
```suggestion
            <% row.value { attr[:name].eql?("Status") ? status_tag(attr[:value]) : details_html(attr) } %>
```

This logic belongs in the helper, that's what it's for! Can you adjust `SchemesHelper#display_scheme_attributes` directly to wrap `status_tag` and `details_html` where appropriate. We can also adjust the `Registered under Care Standards Act 2000` field and pass in `current_user` so that we can filter out the `Housing stock owned by` column.

app/views/schemes/show.html.erb
```diff
@@ -10,19 +10,28 @@
 
 <%= render partial: "organisations/headings", locals: { main: @scheme.service_name, sub: nil } %>
 
-<%= render SubNavigationComponent.new(items: scheme_items(request.path, @scheme.id, "Locations")) %>
+<% if FeatureToggle.location_toggle_enabled? %>
```
jamesrose, line 13: Are there any material changes to making this a grid? It might not be worth putting it behind a feature flag if not.

spec/models/location_spec.rb
jamesrose: @SamSeed-Softwire You are testing a case that isn't possible. Try saving a postcode with a whitespace postcode:

```ruby
irb(main):006:0> location = Location.new(postcode: "   "); location.valid?
=> false
irb(main):007:0> location.errors[:postcode]
=> ["Enter a postcode in the correct format, for example AA1 1AA"]
```

spec/requests/locations_controller_spec.rb
kosiakkatrina: That's fine! As long as it tested and readable I'm happy

spec/models/location_spec.rb
SamSeed-Softwire: @jamesrose thanks. I was trying to write a test under the assumption that the postcode validation might not be working correctly. But I suppose there's already an explicit test to make sure it is:

```
    it "does add an error when the postcode is invalid" do
      location.postcode = "invalid"
      expect { location.save! }
        .to raise_error(ActiveRecord::RecordInvalid, "Validation failed: Postcode #{I18n.t('validations.postcode')}")
    end
```

app/views/schemes/show.html.erb
SamSeed-Softwire: @jamesrose this is because we only want to apply the 2/3 scaling on this page if it's also applied on the location list page - basically Nicky wanted the 2 pages to match each other's scaling. Hence why it's behind the locations feature flag. Shout if you disagree.

app/models/location.rb
```diff
@@ -374,6 +374,8 @@ def available_from
   end
 
   def status
+    return :incomplete if [postcode, units, type_of_unit, mobility_type].any?(&:blank?)
```
kosiakkatrina: Could do something similar to this: https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/1003#discussion_r1025176900

SamSeed-Softwire: Thanks @kosiakkatrina. From what I can see, schemes have a method to validate whether they're confirmed (see app/models/scheme.rb#199). However there doesn't seem to be an equivalent validation function for locations - it just seems like if a scheme is confirmed, then its locations are marked as confirmed too (see app/controllers/schemes_controller.rb#107). Assuming that's correct, do you think `confirmed` is actually that useful for locations currently? I think perhaps not, and that we should raise a new ticket to update how locations get marked as confirmed. Let me know your thoughts! Thanks 🙂

SamSeed-Softwire: @kosiakkatrina any thoughts on this?

jamesrose (approved):

app/models/location.rb
kosiakkatrina: Hiya, sorry for getting to this late. 
So the scheme gets marked as confirmed from the check your answers page for that scheme and at that point all of the locations associated with it would be completed (because we'd throw validation errors when adding a location otherwise).
If a new location gets added to that scheme afterwards it wouldn't be marked as confirmed until the scheme is saved from check your answers page again. Which again, would not work unless the location is completed.
So I think using confirmed should work fine

SamSeed-Softwire: @kosiakkatrina thanks, I understand that.

However I've implemented this change and my tests are now failing - instead of returning `incomplete` the status method is now returning `active`, so it doesn't seem to be working as expected. Can't work out why this might be. Any thoughts?

kosiakkatrina (approved):

PR #988, opened 2022-11-11
Update routing to go back to location page

kosiakkatrina:
Route back to view location page after changing the name:
image

Unless you have originally come from check your answers, in which case route back to check your answers, so that unconfirmed locations could be saved/confirmed:
image


natdeanlewissoftwire (approved):
lgtm 🚀

PR #989, opened 2022-11-14
Parse user fields where namespace declarations aren't included

jamesrose:
This is the same as https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/983 but for users instead of logs.

We import XML files that define logs from the previous CORE service. For an unknown reason, some of these files don't include the required namespace declarations to parse `user` values. Instead of changing the source, this change introduces a new method that forces the namespace declaration for user fields.

This allows us to import users regardless of whether they specify namespace bindings or not.

### DIFF
--- a/app/services/imports/user_import_service.rb
+++ b/app/services/imports/user_import_service.rb
@@ -45,7 +45,7 @@ module Imports
     end
 
     def user_field_value(xml_document, field)
-      field_value(xml_document, "user", field)
+      field_value(xml_document, "user", field, { "user" => "dclg:user" })
     end
 
     def role(field_value)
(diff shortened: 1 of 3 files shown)


app/services/imports/user_import_service.rb
```diff
@@ -45,7 +45,7 @@ def create_user(xml_document)
     end
 
     def user_field_value(xml_document, field)
-      field_value(xml_document, "user", field)
+      field_value(xml_document, "user", field, { "user" => "dclg:user" })
```
natdeanlewissoftwire, line 48: I wonder if dclg should be changed to dluhc to be up to date, or if that's something we should be aware might need to happen in the future? Not a major issue I appreciate :)

natdeanlewissoftwire (approved):
LGTM :)

app/services/imports/user_import_service.rb
jamesrose: Given the inflexibility of the source I doubt it will change!

PR #990, opened 2022-11-14
Update enabled method

kosiakkatrina:
Update enabled method to work the same way routed_to? works. 
Instead of checking the status of other subsections to determine whether the subsection is enabled, we can now check individual questions and call methods.

For lettings logs instead of checking that the setup section is completed we check that all the answers except for scheme and location are answered. That is to prevent the rest of the fields from clearing if scheme and location is missing.

For sales logs we check that all the questions in the setup section are answered

### DIFF
--- a/app/models/sales_log.rb
+++ b/app/models/sales_log.rb
@@ -47,4 +47,8 @@ class SalesLog < Log
   def completed?
     status == "completed"
   end
+
+  def setup_completed?
+    form.setup_sections.first.subsections.first.status(self) == :completed
+  end
 end
--- a/app/models/sales_log.rb
+++ b/app/models/sales_log.rb
@@ -49,6 +49,6 @@ class SalesLog < Log
   end
 
   def setup_completed?
-    form.setup_sections.first.subsections.first.status(self) == :completed
+    form.setup_sections.all? { |sections| sections.subsections.all? { |subsection| subsection.status(self) == :completed } }
   end
 end
--- a/spec/fixtures/forms/2021_2022.json
+++ b/spec/fixtures/forms/2021_2022.json
@@ -1099,9 +1099,10 @@
           "label": "Declaration",
           "depends_on": [
             {
-              "household_characteristics": "completed",
-              "household_needs": "completed",
-              "property_information": "completed"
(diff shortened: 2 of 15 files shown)


spec/fixtures/forms/2021_2022.json
```diff
@@ -1099,9 +1099,10 @@
           "label": "Declaration",
           "depends_on": [
             {
-              "household_characteristics": "completed",
-              "household_needs": "completed",
-              "property_information": "completed"
+              "armedforces": 1
```
natdeanlewissoftwire, line 1102: Why does this depend on armedforces being 1 or 3?

kosiakkatrina: This is just to pass the tests. I wanted to keep it dependent on something so that the behaviour still gets tested, and one of the things in the test set up was armed forces, so I made it dependent on that. No significance otherwise

jamesrose:
> Instead of checking the status of other subsections to determine whether the subsection is enabled, we can now check individual questions and call methods.

Can we do both?

app/models/sales_log.rb
```diff
@@ -47,4 +47,8 @@ def not_started?
   def completed?
     status == "completed"
   end
+
+  def setup_completed?
+    form.setup_sections.first.subsections.first.status(self) == :completed
```
jamesrose: Is it enough to check the first section? Do we not need to check that all sections and subsections are complete? 

```suggestion
    form.setup_sections.all? { |sections| sections.subsections.all? { |subsection| subsection.status(self) == :completed }}
```

kosiakkatrina:
> 

@jamesrose potentially, but the way `enabled?` method was written was specifically to check the status of a specific subsection(s). Now we can just call a method, which can be anything, including a check for a subsection status as I'm doing for the sales logs in `setup_completed?`

app/models/sales_log.rb
kosiakkatrina: Hm, you're right. I didn't think about it cause we only have 1 atm 🤦

jamesrose (approved):

PR #996, opened 2022-11-15
Cldc 1668 multiple location deactivations

kosiakkatrina:
Add a `location_deactivation_periods` table to store multiple deactivation periods per location.
Update `status` location method to look at the most recent deactivation to determine the location status. 
The most recent deactivation is the record that has `deactivation_date` but doesn't have a `reactivation_date`. 
There should ever be 1 or 0 of this kind of a record, because deactivated locations need to be reactivated before they can be deactivated again.
Display deactivation timeline:
image

When a location gets deactivated:
- all logs that have startdate after deactivation date have location and scheme set to nil
- a deactivation record with `deactivation_date` is added to `location_deactivations` table


natdeanlewissoftwire:
Do we think this pr should also include the "availability" text on the scheme/location page that gives the list of previous active/inactive time periods?

kosiakkatrina:
> Do we think this pr should also include the "availability" text on the scheme/location page that gives the list of previous active/inactive time periods?

@natdeanlewissoftwire Yup, will do that. Thanks!

db/migrate/20221115101732_add_deactivations_table.rb
```diff
@@ -0,0 +1,10 @@
+class AddDeactivationsTable < ActiveRecord::Migration[7.0]
+  def change
+    create_table :location_deactivations do |t|
```
ghost: is this table going to be used to store both activations and deactivations?

kosiakkatrina: Yes, I think that's the idea.
The way I'm thinking about it is that `reactivation_date` can only be added to an existing `location_deactivations` record that has `deactivation_date`. 

My reasoning behind it is that we deactivate a location first and then reactivate it. 
We can't deactivate a location twice before it gets reactivated either, so it should be easy to track which record we add the reactivation to (which would be the only record with a missing reactivation date).

Let me know if that makes sense to you and if you've got different ideas on how we can approach this!

ghost: If this table is going to be used to store both activations and deactivations then I believe it is not the most optimal way to represent the data we plan to store.

Similar to what we did with OrganisationRelationships for Housing Provider and Managing agent; we could have a shared table with a generic name (LocationEvent? ActivationEvent? LocationStatusEvent) with a single `event_time` field and `event_type` field (activation/deactivation). 

Additionally, if the logic related to activating/deactivating schemes is similar to the one of Locations, we could make this a polymorphic table and share it across both Locations and Events
```
    create_table :status_events do |t|
      t.datetime :event_time
      t.string :event_type
      t.belongs_to :eventable (Location/Scheme)
      t.timestamps
    end
```

kosiakkatrina: I guess this could work, the only thing I'm worried about atm is whether we might need to associate which reactivation belongs to which deactivation for any reason.

I was thinking about a single table with a type column for location/scheme but ended up going down the separate tables route cause it was somewhat more simple to understand.

While I can see the benefit of generalising things, could it be a bit premature to do it at the moment? Are there any other events we expect to save?
I'd like to hear how it's more optimal to save it that way, cause  I'm struggling to see it atm

ghost: A generic table allows deduplication but it is not necessary. 

I feel that this table currently does not make things clearer because the table name does not correctly describe what it contains (location_deactivations).

Also two column called `deactivation_date`, `reactivation_date` could imply that both values can be set which would never be the case. For this reason I suggested replacing them with `type` and `date` columns

app/helpers/locations_helper.rb
```diff
@@ -32,8 +32,17 @@ def display_attributes(location)
...
-      { name: "Availability", value: "Available from #{location.available_from.to_formatted_s(:govuk_date)}" },
+      { name: "Availability", value: location_availability(location) },
       { name: "Status", value: location.status },
     ]
   end
+
+  def location_availability(location)
+    availability = "Active from #{location.available_from.to_formatted_s(:govuk_date)}"
+    location.location_deactivations.each do |deactivation|
+      availability << " to #{(deactivation.deactivation_date - 1.day).to_formatted_s(:govuk_date)}\nDeactivated on #{deactivation.deactivation_date.to_formatted_s(:govuk_date)}"
```
natdeanlewissoftwire, line 43: just to flag you’ve seen that I’ve converted Dates to Times so we actually save UTC midnight rather than local midnight in the database for our date - I'm not sure if this will cause an issue with `deactivation_date - 1.day`, do you think it will? https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/980/files#diff-bee6109a705a78cdfa12162d723fe642c7b7c8e53045513503b6d8e574e41b15R206

db/migrate/20221115101732_add_deactivations_table.rb
kosiakkatrina: So, I thought that both values would be set and that would constitute a deactivation period.
So a record like this:

```
id | deactivation_date | reactivation_date | location_id
1 |  2022-8-9 | 2022-8-20 | 3
```

would mean that the location with id 3 was not active between the 9th of August and 20th of August 2022.

If there's a record like this:

```
id | deactivation_date | reactivation_date | location_id
1 |  2022-8-9 | 2022-8-20 | 3
2 |  2022-8-23 | nil | 3
```

would mean that the location with id 3 was not active between the 9th of August and 20th of August 2022. And then got deactivated again on the 23rd of August and is currently deactivated

app/helpers/locations_helper.rb
kosiakkatrina: Hmm, I think it should still be fine, but I'll rebase and check when you get your PR merged in! Thanks!

db/migrate/20221115101732_add_deactivations_table.rb
kosiakkatrina: I can see what you mean with the table name. I feel like it's clear to me cause I'm viewing deactivations as periods, but maybe that's not how everyone sees them. Is `location_deactivation_periods` clearer? Or would naming it something else make sense?

ghost: Okay I see it... mmh would `location_deactivation_periods` table with `start_date` and `end_date` columns make sense?

kosiakkatrina: Happy to rename the table, the start and end dates might be a bit confusing tho. `End date of deactivation` feels like a double negative so I'd find it confusing. Maybe `deactivated_on` and `reactivated_on` ?

natdeanlewissoftwire: Just to chip in I like `location_deactivation_periods` but keeping `deactivation_date` and `reactivation_date` as that's more similar to the mental model implied by the UI (as you separately add deactivation and reactivation dates rather than submitting a whole period at a time)

kosiakkatrina: Renamed it to that for now, thanks both

ghost (approved):

PR #1003, opened 2022-11-17
feat: add new design to scheme page

natdeanlewissoftwire:
Update the scheme page designs to the latest one (including status behind a feature flag)

tested locally
minimal risk
no docs required

ticket: https://digital.dclg.gov.uk/jira/browse/CLDC-1666

`reactivating_soon` status will be added with the scheme reactivation ticket
`activating_soon` does not apply to schemes

image


app/models/scheme.rb
```diff
@@ -211,11 +209,16 @@ def validate_confirmed
...
+  def required_attributes
+    attribute_names - %w[id created_at updated_at old_id old_visible_id confirmed end_date sensitive secondary_client_group total_units has_other_client_group deactivation_date deactivation_date_type]
+  end
+
   def available_from
     created_at
   end
 
   def status
+    return :incomplete if required_attributes.any? { |attribute| self[attribute].blank? }
```
jamesrose:
```suggestion
    return :incomplete unless confirmed
```

Is this not the same thing?

natdeanlewissoftwire: Ah, nice I hadn't realise that was what confirmed was actually testing

kosiakkatrina (approved):
Just the merge conflicts, looks good otherwise

PR #1004, opened 2022-11-17
Multiple scheme deactivations

kosiakkatrina:
Add a `scheme_deactivation_periods` table to store multiple deactivation periods per scheme.
Update status scheme method to look at the most recent deactivation to determine the scheme status.
The most recent deactivation is the record that has `deactivation_date` but doesn't have a `reactivation_date`.
There should ever be 1 or 0 of this kind of a record, because deactivated schemes need to be reactivated before they can be deactivated again.
Display deactivation timeline in the individual scheme view.

When a scheme gets deactivated:
- all logs associated with that scheme that have `startdate` after deactivation date have location and scheme set to nil
- a deactivation record with `deactivation_date` is added to `scheme_deactivation_periods` table


app/helpers/schemes_helper.rb
```diff
@@ -26,4 +26,13 @@ def display_scheme_attributes(scheme)
     end
     base_attributes
   end
+
+  def scheme_availability(scheme)
+    availability = "Active from #{scheme.available_from.to_formatted_s(:govuk_date)}"
+    scheme.scheme_deactivation_periods.each do |deactivation|
+      availability << " to #{(deactivation.deactivation_date - 1.day).to_formatted_s(:govuk_date)}\nDeactivated on #{deactivation.deactivation_date.to_formatted_s(:govuk_date)}"
```
natdeanlewissoftwire, line 33: I think we also need "Activating on..." in the case that the scheme is `:reactivating_soon` - happy to put this in later though for locations and schemes, or it could go in this PR, what do you think?

natdeanlewissoftwire: Or is "Active from {future_activation_date}" enough do you think?

app/models/scheme.rb
```diff
@@ -216,8 +217,9 @@ def available_from
   end
 
   def status
-    return :active if deactivation_date.blank?
-    return :deactivating_soon if Time.zone.now < deactivation_date
+    recent_deactivation = scheme_deactivation_periods.deactivations_without_reactivation.first
+    return :active if recent_deactivation.blank?
+    return :deactivating_soon if Time.zone.now < recent_deactivation.deactivation_date
```
natdeanlewissoftwire, line 222: again `:activating_soon` should be here too, but maybe not part of this PR

app/helpers/schemes_helper.rb
kosiakkatrina: Would that be a part of reactivation work?

natdeanlewissoftwire (approved):
LGTM! Thanks for copying this across :)

app/helpers/schemes_helper.rb
natdeanlewissoftwire: Sure, sounds good

PR #1006, opened 2022-11-17
Unfreeze time

kosiakkatrina:
Some non related tests started to fail because we didn't reset time after freezing it.
This unfreezes all the frozen time that were missed after tests to prevent that


natdeanlewissoftwire:
I've tested this with the seed that was previously failing for me (15251) and they now all pass with this fix :)

natdeanlewissoftwire (approved):
LGTM :)

PR #1016, opened 2022-11-21
[1018] Make cookies secure for production like envs

asmega:
# Context

- https://digital.dclg.gov.uk/jira/browse/CLDC-1018
- Cookies were not set with the `secure` flag. This means they would continue to work if downgraded from `https` to `http`

# The changes

- Set the cookie to have `secure` flag for production like envs
- I tested this with a few deploys and appears to not log you out and just upgrade the cookie

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -35,5 +35,11 @@ module DataCollector
     config.exceptions_app = routes
 
     config.active_job.queue_adapter = :sidekiq
+
+    config.session_store(
+      :cookie_store,
+      key: "_data_collector_session",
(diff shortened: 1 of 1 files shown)


jamesrose:
Good spot. Is this going to log everyone out?

asmega:
> Good spot. Is this going to log everyone out?

Good question. not actually sure. i think i'd have to do some testing first

asmega:
> Good spot. Is this going to log everyone out?

Just did a few deploys to test and seems to not log you out and just upgrades the cookie to `secure`

jamesrose (approved):

PR #1018, opened 2022-11-22
Deactivation does not affect logs

kosiakkatrina:
Only show confirm deactivation page when a location or scheme gets deactivated if it affects at least 1 log (base on designs).
Otherwise deactivate the scheme/location and route to the following page with a success banner

### DIFF
--- a/app/views/locations/deactivate_confirm.html.erb
+++ b/app/views/locations/deactivate_confirm.html.erb
@@ -4,7 +4,7 @@
      <% end %>
    <h1 class="govuk-heading-l">
      <span class="govuk-caption-l"><%= @location.postcode %></span>
-     This change will affect <%= @location.lettings_logs.filter_by_before_startdate(@deactivation_date).count %> logs
+     This change will affect <%= @affected_logs.count %> logs
    </h1>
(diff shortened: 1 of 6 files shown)


app/views/locations/deactivate_confirm.html.erb
```diff
@@ -4,7 +4,7 @@
      <% end %>
    <h1 class="govuk-heading-l">
      <span class="govuk-caption-l"><%= @location.postcode %></span>
-     This change will affect <%= @location.lettings_logs.count %> logs
+     This change will affect <%= @location.lettings_logs.filter_by_before_startdate(@deactivation_date).count %> logs
```
jamesrose: Why don't we just hide this line if there is no impact? I think we should still give people an opportunity to back out.

kosiakkatrina: There was a discussion in the ticket and this was the decision

jamesrose: Okay! Can we memoize this query? We're executing it twice.

jamesrose (approved):

PR #1019, opened 2022-11-23
[1711] Orgs parent-child relationship now atomic

asmega:
# Context

- https://digital.dclg.gov.uk/jira/browse/CLDC-1711
- Prior to this change a relation between 2 organisations was represented by 2 independent `OrganisationRelationship` records. One for each side of the relationship
- We were not creating both sides of the relationship therefore only one party would see the relationship
- This has now been simplified so only one `OrganisationRelationship` record is needed as we do not need each way of the relationship to be independent of the other

# The changes

- The column `OrganisationRelationship#relationship_type` has been removed
- A single `OrganisationRelationship` record now represents the 2 way relationship between 2 orgs
- Therefore there is no need to create 2 records to represent the relationship
- Therefore when deleting the relationship only one `OrganisationRelationship` record will need to be deleted
- Added missing database indexes to `OrganisationRelationship` which should reduce the need for full table scans
- Added a unique index to prevent duplicate relationships from entering the system
- Added foreign key constraints to prevent bad data integrity between relationships and orgs
- Seeding of relationships is now idempotent so review apps can run the seeding process over and over without issue

### DIFF
--- a/app/controllers/organisation_relationships_controller.rb
+++ b/app/controllers/organisation_relationships_controller.rb
@@ -39,7 +39,6 @@ class OrganisationRelationshipsController < ApplicationController
 
   def create_housing_provider
     child_organisation = @organisation
-    relationship_type = OrganisationRelationship::OWNING
     if params[:organisation][:related_organisation_id].empty?
       @organisation.errors.add :related_organisation_id, "You must choose a housing provider"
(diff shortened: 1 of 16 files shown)


ghost (approved):

jamesrose (approved):
LGTM!

PR #1022, opened 2022-11-23
Cldc 1609 schemes and location validations

kosiakkatrina:
Add validations to lettings log startdate, location_id and scheme_id to prevent a startdate during an inactive location/scheme period from being selected.
When a location is deactivated on selected date but has a reactivation date
image

When a location is deactivated on selected date and has no reactivation date
image

When a location is not yet active
image

When a scheme is deactivated on selected date and has no reactivation date
image

When a scheme is deactivated on selected date and has a reactivation
image

When a scheme is not yet active
image


app/models/validations/date_validations.rb
```diff
@@ -91,4 +101,15 @@ def date_valid?(question, record)
   def is_rsnvac_first_let?(record)
     [15, 16, 17].include?(record["rsnvac"])
   end
+
+  def inactive_status(date, deactivation_periods, available_from)
+    return if date.blank?
+
+    closest_reactivation = deactivation_periods.reverse.find { |period| period.reactivation_date.present? && date.between?(period.deactivation_date, period.reactivation_date - 1.day) } if deactivation_periods.present?
```
SamSeed-Softwire: Do we not need to sort the reactivation periods before reversing them? I couldn't find sorting being applied anywhere.

app/models/validations/date_validations.rb
```diff
@@ -59,6 +59,16 @@ def validate_startdate(record)
     if record["mrcdate"].present? && record.startdate < record["mrcdate"]
       record.errors.add :startdate, I18n.t("validations.setup.startdate.after_major_repair_date")
     end
+
+    location_inactive_status = inactive_status(record.startdate, record.location&.location_deactivation_periods, record.location&.available_from)
```
SamSeed-Softwire: This is for my own learning benefit, sorry! Is `record.location&.location_deactivation_periods` a shorthand for getting all rows (if any exist) from the `location_deactivation_periods` table for the location in question?

SamSeed-Softwire: Where does the location get chosen when creating a lettings log? I can't seem to actually pick one anywhere. I'm assuming it's possible, because in your comments at the beginning of this PR it seems like picking a tenancy start date (for a lettings log) triggers a validation error if either the scheme _or_ location is deactivated on the chosen date.

app/models/validations/setup_validations.rb
```diff
@@ -5,9 +5,40 @@ def validate_irproduct_other(record)
...
   end
 
+  def validate_location(record)
+    validate_location_during_startdate(record, :location_id)
+  end
+
+  def validate_scheme(record)
+    validate_location_during_startdate(record, :scheme_id)
+
+    scheme_inactive_status = inactive_status(record.startdate, record.scheme&.scheme_deactivation_periods, record.scheme&.available_from)
```
SamSeed-Softwire: Not essential but might it be cleaner to write a `validate_scheme_during_startdate` method as well, to make it clearer that this block of code is doing something equivalent to the `validate_location_during_startdate` method call just above?

app/models/validations/date_validations.rb
kosiakkatrina: Hm, I think this might have been linter. I suspect I've had something like:
`deactivation_periods.select { |period| period.reactivation_date.present? && date.between?(period.deactivation_date, period.reactivation_date - 1.day) }.last `

My assumption was that the records are returned ordered by `created_at` date, but even if it's so I haven't actually found any proof for that, so I updated the statement to explicitly order it. Thanks!

app/models/validations/date_validations.rb
kosiakkatrina: `record.location&.location_deactivation_periods`
is the same as writing
`record.location.location_deactivation_periods if record.location.present?` I think

kosiakkatrina: You'd need to choose a scheme first and the next question would be location. There has to be more than 1 location available for that scheme tho for the location question to show up

app/models/validations/setup_validations.rb
kosiakkatrina: Yup, can extract it. Didn't do it because we're only running it once, but happy to if it makes it clearer!

app/models/validations/date_validations.rb
```diff
@@ -91,4 +101,15 @@ def date_valid?(question, record)
   def is_rsnvac_first_let?(record)
     [15, 16, 17].include?(record["rsnvac"])
   end
+
+  def inactive_status(date, deactivation_periods, available_from)
+    return if date.blank?
+
+    closest_reactivation = deactivation_periods.order(created_at: :desc).find { |period| period.reactivation_date.present? && date.between?(period.deactivation_date, period.reactivation_date - 1.day) } if deactivation_periods.present?
```
jamesrose:
```suggestion
    closest_reactivation = deactivation_periods.order(created_at: :desc).find { |period| period.reactivation_date.present? && date.between?(period.deactivation_date, period.reactivation_date - 1.day) } if deactivation_periods.present?
```

Can you make this an ActiveRecord query?

app/models/validations/date_validations.rb
```diff
@@ -91,4 +101,15 @@ def date_valid?(question, record)
   def is_rsnvac_first_let?(record)
     [15, 16, 17].include?(record["rsnvac"])
   end
+
+  def inactive_status(date, deactivation_periods, available_from)
```
jamesrose: Does this method re-implement `Location#status` and `Scheme#status`?

app/models/validations/setup_validations.rb
SamSeed-Softwire: Fbm - will leave it up to you!

jamesrose (approved):

PR #1023, opened 2022-11-23
Cldc 1672 scheme reactivation

natdeanlewissoftwire:
tested locally
small risk despite significant changes as this is mostly (duplication of location reactivation that has already been reviewed)
no docs required

ticket: https://digital.dclg.gov.uk/jira/browse/CLDC-1672

### DIFF
--- a/app/controllers/locations_controller.rb
+++ b/app/controllers/locations_controller.rb
@@ -50,12 +50,12 @@ class LocationsController < ApplicationController
   end
 
   def new_reactivation
-    @location_deactivation_period = LocationDeactivationPeriod.deactivations_without_reactivation.first
+    @location_deactivation_period = @location.location_deactivation_periods.deactivations_without_reactivation.first
     render "toggle_active", locals: { action: "reactivate" }
   end
 
   def reactivate
-    @location_deactivation_period = LocationDeactivationPeriod.deactivations_without_reactivation.first
+    @location_deactivation_period = @location.location_deactivation_periods.deactivations_without_reactivation.first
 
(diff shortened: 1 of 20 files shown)


app/controllers/locations_controller.rb
```diff
@@ -50,12 +50,12 @@ def deactivate
   end
 
   def new_reactivation
-    @location_deactivation_period = LocationDeactivationPeriod.deactivations_without_reactivation.first
+    @location_deactivation_period = @location.location_deactivation_periods.deactivations_without_reactivation.first
```
natdeanlewissoftwire, line 53: This fixes a bug where the wrong deactivation period could be returned if there are multiple without reactivations in the db due to there being multiple locations/schemes.

config/locales/en.yml
```diff
@@ -313,19 +313,25 @@ en:
...
 
     location:
       toggle_date:
         not_selected: "Select one of the options"
         invalid: "Enter a valid day, month and year"
         out_of_range: "The date must be on or after the %{date}"
       reactivation:
-        before_deactivation: "This location was deactivated on %{date}\nThe reactivation date must be on or after deactivation date"
-      deactivation: 
+        before_deactivation: "This location was deactivated on %{date}. The reactivation date must be on or after deactivation date"
```
natdeanlewissoftwire, line 333: The new line wasn't dislaying correctly so this is an alternative to avoid making a special case/adding multiple errors which seems a bit hacky.

kosiakkatrina (approved):

PR #1026, opened 2022-11-24
Update available_from methods

kosiakkatrina:
Sets the `available_from` date to the beginning of the collection period in which the location or scheme have been created.

Before (if no `startdate` exists) it was setting it to the beginning of the current collection window if it was created later than the beginning of the collection window.

However if it was created before the current collection window start date, it was being set to creation date.

This would become an issue when we reach the next collection year, because all of the locations that were created after the current collection year start date will have their `available_from` dates changed to a year later.

### DIFF
--- a/app/models/location.rb
+++ b/app/models/location.rb
@@ -372,8 +372,7 @@ class Location < ApplicationRecord
   def available_from
     return startdate if startdate.present?
 
-    window_end_date = Time.zone.local(created_at.year, 4, 1)
-    created_at < window_end_date ? Time.zone.local(created_at.year - 1, 4, 1) : Time.zone.local(created_at.year, 4, 1)
+    FormHandler.instance.collection_start_date(created_at)
   end
 
   def status
--- a/app/models/location.rb
+++ b/app/models/location.rb
@@ -370,7 +370,10 @@ class Location < ApplicationRecord
   end
 
   def available_from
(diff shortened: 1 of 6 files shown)


natdeanlewissoftwire (approved):
LGTM!

app/models/location.rb
```diff
@@ -370,7 +370,10 @@ def postcode=(postcode)
   end
 
   def available_from
-    startdate || [created_at, FormHandler.instance.current_collection_start_date].min
+    return startdate if startdate.present?
+
+    window_end_date = Time.zone.local(created_at.year, 4, 1)
+    created_at < window_end_date ? Time.zone.local(created_at.year - 1, 4, 1) : Time.zone.local(created_at.year, 4, 1)
```
jamesrose: Too many magic dates! We should take this from the form. Imagine we need to change it later...

kosiakkatrina: Hm, yeah, okay. Not looking at a specific form, because we wouldn't have one at this point, but moved it to the form handler. How's that?

jamesrose: How do you know that April 1st is always going to be a date that a collection window ends?

kosiakkatrina: Hm, has it changed in the past?
It's used across the codebase, both in form handler and in form model. I don't think there's any other date I could get at the moment. Did you have a different approach in mind?

jamesrose (approved):

PR #1027, opened 2022-11-24
1628: a scheme's owning organisation must hold stock

SamSeed-Softwire:
## Context

* https://digital.dclg.gov.uk/jira/browse/CLDC-1628
* Steer from Jemma: this ticket should **only cover** adding validation such that users should be presented with an error if they select a non-stock-owning org for the "Which organisation owns the housing stock for this scheme?" question.

## The changes
* Added a `validate_owning_organisation` method onto the `Scheme` model (which uses new validation text added to `config/locales/en.yml`).
* Tests (all in `spec/requests/schemes_controller_spec.rb`)
  * Test to check that entering an invalid owning org should trigger the validation error mentioned above.
  * Test to check that when the user is a data coordinator, the owning org of a scheme is set as the user's org even if when creating the scheme a different owning org was specified.
  * Removed a test for `owning_organisation_id` being missing for support users, because the test above it already covered that case.
  * Removed `owning_organisation_id` from a 'missing required params' test for data coordinators, since it is not a required param for data coordinators.
  * Minor improvements to some test names/descriptions.

## Screenshots
image
image

## Questions / things I wasn't sure about
* Does it make sense to also add a test in `spec/models/scheme_spec.rb`? I tried to write one which created a scheme with an invalid owning org and then checked the contents of `errors`, but this didn't work because the creation failed due to the validation I'd written! So my guess is that the only test which makes sense is the one I've already written in `spec/requests/schemes_controller_spec.rb'. Shout if I'm wrong though!
* Some other places I thought it **might** make sense to add tests (all in `spec/requests/schemes_controller_spec.rb`) were as follows. Interested to know if these sound sensible or not (I couldn't work out how to check the contents of an accessible autocomplete box, so thought I'd wait to see if it was even a good idea before spending more time on this).
  * describe "#edit_name"
    * When signed in as a data coordinator, a test could check that the scheme's owning org (i.e. the user's organisation) is displayed in the accessible autocomplete box.
    * When signed in as a support user, a test could check that the scheme's owning org  is displayed in the accessible autocomplete box.

Thanks for reviewing!

### DIFF
--- a/app/models/scheme.rb
+++ b/app/models/scheme.rb
@@ -213,7 +213,7 @@ class Scheme < ApplicationRecord
   end
 
   def validate_owning_organisation
-    if !owning_organisation.holds_own_stock?
+    unless owning_organisation.holds_own_stock?
       errors.add(:owning_organisation_id, :does_not_own_stock, message: I18n.t("validations.scheme.owning_organisation.does_not_own_stock"))
     end
   end
--- a/app/models/scheme.rb
+++ b/app/models/scheme.rb
@@ -215,7 +215,6 @@ class Scheme < ApplicationRecord
   def validate_owning_organisation
     if !owning_organisation.holds_own_stock?
       errors.add(:owning_organisation_id, :does_not_own_stock, message: I18n.t("validations.scheme.owning_organisation.does_not_own_stock"))
-      owning_organisation = nil
     end
   end
 
--- a/app/models/scheme.rb
+++ b/app/models/scheme.rb
@@ -21,7 +21,7 @@ class Scheme < ApplicationRecord
   scope :order_by_completion, -> { order("confirmed ASC NULLS FIRST") }
   scope :order_by_service_name, -> { order(service_name: :asc) }
 
-  validate :validate_confirmed
+  validate :validate_confirmed, :validate_owning_organisation
 
   auto_strip_attributes :service_name
 
@@ -209,6 +209,13 @@ class Scheme < ApplicationRecord
           self.confirmed = false
         end
       end
+    end
+  end
+
+  def validate_owning_organisation
+    if !owning_organisation.holds_own_stock?
+      errors.add(:owning_organisation_id, :does_not_own_stock, message: I18n.t("validations.scheme.owning_organisation.does_not_own_stock"))
+      owning_organisation = nil
     end
   end
 
--- a/app/models/scheme.rb
+++ b/app/models/scheme.rb
@@ -21,7 +21,8 @@ class Scheme < ApplicationRecord
   scope :order_by_completion, -> { order("confirmed ASC NULLS FIRST") }
   scope :order_by_service_name, -> { order(service_name: :asc) }
 
-  validate :validate_confirmed, :validate_owning_organisation
+  validate :validate_confirmed
+  validate :validate_owning_organisation
 
   auto_strip_attributes :service_name
 
--- a/spec/models/scheme_spec.rb
+++ b/spec/models/scheme_spec.rb
@@ -217,7 +217,7 @@ RSpec.describe Scheme, type: :model do
 
     context "when the owning organisation is set as a non-stock-owning organisation" do
       it "throws the correct validation error" do
-        expect { scheme.update!({ owning_organisation_id: non_stock_owning_org.id }) }.to raise_error(ActiveRecord::RecordInvalid, /#{I18n.t("validations.scheme.owning_organisation.does_not_own_stock")}/)
+        expect { scheme.update!({ owning_organisation_id: non_stock_owning_org.id }) }.to raise_error(ActiveRecord::RecordInvalid, /Enter an organisation that owns housing stock/)
       end
     end
   end
--- a/spec/models/scheme_spec.rb
+++ b/spec/models/scheme_spec.rb
@@ -217,7 +217,7 @@ RSpec.describe Scheme, type: :model do
 
     context "when the owning organisation is set as a non-stock-owning organisation" do
       it "throws the correct validation error" do
-        expect { scheme.update!({ owning_organisation_id: non_stock_owning_org.id }) }.to raise_error(ActiveRecord::RecordInvalid, /Enter an organisation that owns housing stock/)
+        expect { scheme.update!({ owning_organisation: non_stock_owning_org }) }.to raise_error(ActiveRecord::RecordInvalid, /Enter an organisation that owns housing stock/)
       end
     end
   end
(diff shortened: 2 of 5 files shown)


spec/requests/schemes_controller_spec.rb
```diff
@@ -575,13 +589,18 @@
         end
       end
 
-      context "when required organisation id param is missing" do
```
SamSeed-Softwire, line 589: The test that was here seemed to be redundant as it was covered by the test above it.

kosiakkatrina (approved):
Looks good

to answer your question, we could test the model by running the validation directly on the record (here's an example https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/blob/main/spec/models/validations/date_validations_spec.rb#L11), you'd need to create a new file tho, I don't think we have one for schemes yet 

_edit: Ah just realised the validations are defined differently there, so wouldn't be able to do that.
But something like this could've worked:_
 ```
  expect { scheme.update!({ organisation: non_stock_owning_org }) }
          .to raise_error(ActiveRecord::RecordInvalid, /#{I18n.t("validations...")}/)
```

File: spec/models/scheme_spec.rb [hunk 2]
```diff
@@ -209,4 +209,16 @@
       end
     end
   end
+
```
SamSeed-Softwire, line 212: @kosiakkatrina I've added this test in response to you comment - thanks for the suggested solution!

kosiakkatrina (approved):
Looks good!

app/models/scheme.rb
```diff
@@ -212,6 +212,12 @@ def validate_confirmed
     end
   end
 
+  def validate_owning_organisation
+    unless owning_organisation.holds_own_stock?
+      errors.add(:owning_organisation_id, :does_not_own_stock, message: I18n.t("validations.scheme.owning_organisation.does_not_own_stock"))
```
jamesrose, line 218: Do you need the second argument - `:does_not_own_stock` here?

spec/models/scheme_spec.rb
```diff
@@ -209,4 +209,16 @@
...
+  describe "owning organisation" do
+    let(:stock_owning_org) { FactoryBot.create(:organisation, holds_own_stock: true) }
+    let(:non_stock_owning_org) { FactoryBot.create(:organisation, holds_own_stock: false) }
+    let(:scheme) { FactoryBot.create(:scheme, owning_organisation_id: stock_owning_org.id) }
+
+    context "when the owning organisation is set as a non-stock-owning organisation" do
+      it "throws the correct validation error" do
+        expect { scheme.update!({ owning_organisation_id: non_stock_owning_org.id }) }.to raise_error(ActiveRecord::RecordInvalid, /#{I18n.t("validations.scheme.owning_organisation.does_not_own_stock")}/)
```
jamesrose: How do you know that this is throwing the correct validation error?

app/models/scheme.rb
SamSeed-Softwire: If I don't then I get a different validation error: "Enter the name of the organisation that owns the housing stock". (The validation I want to display is "Enter an organisation that owns housing stock"). Do you have a suggestion for a better way to deal with this? Thanks!

spec/models/scheme_spec.rb
SamSeed-Softwire: When I change the expected validation (e.g. to `activerecord.errors.models.scheme.attributes.owning_organisation_id.invalid`) the test fails:

`expected ActiveRecord::RecordInvalid with message matching /Enter the name of the organisation that owns the housing stock/, got #<ActiveRecord::RecordInvalid: Validation failed: Owning organisation Enter an organisation that owns housing stock>`

app/models/scheme.rb
```diff
@@ -21,7 +21,7 @@ class Scheme < ApplicationRecord
   scope :order_by_completion, -> { order("confirmed ASC NULLS FIRST") }
   scope :order_by_service_name, -> { order(service_name: :asc) }
 
-  validate :validate_confirmed
+  validate :validate_confirmed, :validate_owning_organisation
```
jamesrose:
```suggestion
  validate :validate_confirmed
  validate :validate_owning_organisation
```

spec/requests/schemes_controller_spec.rb
```diff
@@ -586,13 +600,18 @@
...
-        let(:params) { { "scheme" => { "service_name" => "qweqwer", "sensitive" => "Yes", "owning_organisation_id" => "", "scheme_type" => "Foyer", "registered_under_care_act" => "Yes – part registered as a care home" } } }
+      context "when organisation id param refers to a non-stock-owning organisation" do
+        let(:organisation_which_does_not_own_stock) { FactoryBot.create(:organisation, holds_own_stock: false) }
+        let(:params) do
+          { scheme: {
+            owning_organisation_id: organisation_which_does_not_own_stock.id,
```
SamSeed-Softwire: @jamesrose changing this to `owning_organisation: organisation_which_does_not_own_stock` gives the following error:

```
Failure/Error: unless owning_organisation.holds_own_stock?
     
     NoMethodError:
       undefined method `holds_own_stock?' for nil:NilClass
     
           unless owning_organisation.holds_own_stock?
                                     ^^^^^^^^^^^^^^^^^
     # ./app/models/scheme.rb:217:in `validate_owning_organisation'
     # ./app/controllers/schemes_controller.rb:86:in `create'
     # ./spec/requests/schemes_controller_spec.rb:612:in `block (5 levels) in <top (required)>'
```

spec/requests/schemes_controller_spec.rb
```diff
@@ -489,6 +489,20 @@
           expect(page).to have_content(I18n.t("activerecord.errors.models.scheme.attributes.service_name.invalid"))
         end
       end
+
+      context "when the organisation id param is included" do
+        let(:organisation) { FactoryBot.create(:organisation) }
+        let(:params) do
+          { scheme: {
+            owning_organisation_id: organisation.id,
```
SamSeed-Softwire: @jamesrose unlike here https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/1027/files#r1042145393, changing this to `owning_organisation: organisation_which_does_not_own_stock` does not give an error. Why is this?

jamesrose (approved):

spec/requests/schemes_controller_spec.rb
jamesrose: Ah, apologies. This is `params` which isn't applicable. Please ignore :)

jamesrose (approved):

PR #1024, opened 2022-11-23
[1712] Start of bulk upload journey

asmega:
# Context

- https://digital.dclg.gov.uk/jira/browse/CLDC-1712
- This is the start of the bulk upload logs journey

# The changes

- Adds new links styled as buttons to lettings and sales logs pages
- These are behind a feature toggle and enabled on non prod environments
- Depending if you are in crossover period or not you are redirected to relevant page
- Updated nav so after starting bulk upload primary nav selects correct nav item
- Implemented `Form#inspect` so it doesn't echo out the entire (which is humongous) object to console

# Notes for review

- I have left the current bulk upload code untouched we can re-route at the end if this feature supersedes that one

# Screenshots

Screenshot 2022-11-23 at 16 49 49

Screenshot 2022-11-23 at 16 50 01

## Mobile version

Screenshot 2022-11-23 at 16 51 25

PR #1028, opened 2022-11-24
[1713] Bulk upload part deux

asmega:
# Context

- https://digital.dclg.gov.uk/jira/browse/CLDC-1713
- Builds upon https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/1024 so you may want to review by commit
- Adds the second page, picking the year, to bulk upload journey

# The changes

- Added second page to select year if in crossover period
- Introduce the concept of form objects to the app
- The "wizard" uses url params to pass state forwards and backwards
- Added Devise test helpers to feature specs so `sign_in` can be used

# Screenshots

## Initial state

Screenshot 2022-11-24 at 15 18 50

## Error state

Screenshot 2022-11-24 at 15 19 21

## Initial state mobile

Screenshot 2022-11-24 at 15 20 11

## Error state mobile

Screenshot 2022-11-24 at 15 20 58


app/models/forms/bulk_upload_lettings/year.rb
```diff
@@ -0,0 +1,26 @@
...
+      validates :year, presence: true
+
+      def view_path
+        "bulk_upload_lettings_logs/forms/year"
+      end
+
+      def options
+        [
+          OpenStruct.new(id: "2022", name: "2022/2023"),
+          OpenStruct.new(id: "2021", name: "2021/2022"),
```
jamesrose:
```suggestion
          OpenStruct.new(id: "2022", name: "2022/2023"),
          OpenStruct.new(id: "2021", name: "2021/2022"),
```

Ideally these are derived from `FormHandler` so we don't need to change them every year.

app/models/forms/bulk_upload_sales/year.rb
```diff
@@ -0,0 +1,26 @@
...
+      validates :year, presence: true
+
+      def view_path
+        "bulk_upload_sales_logs/forms/year"
+      end
+
+      def options
+        [
+          OpenStruct.new(id: "2022", name: "2022/2023"),
+          OpenStruct.new(id: "2021", name: "2021/2022"),
```
jamesrose: Same as other comment.

asmega: See other thread

app/models/forms/bulk_upload_lettings/year.rb
asmega: I've updated this. Not sure if my implementation meets our needs?

app/models/form.rb
```diff
@@ -223,4 +223,12 @@ def depends_on_met(depends_on, log)
       end
     end
   end
+
+  def in_crossover_period?(now: Time.zone.now)
+    ((end_date - 3.months) < now) && (now < end_date)
```
kosiakkatrina, line 228: I'd say we could use start dates of the forms to determine this instead of subtracting 3 months.
We could potentially get all the `lettings_forms` or `sales_forms` from form handler and check if today is between start_date and end_date for more than one of them.
Maybe this belongs in FormHandler instead of the model too?

app/controllers/bulk_upload_logs_controller.rb
```diff
@@ -0,0 +1,37 @@
...
+private
+
+  def in_crossover_period?
+    FormHandler.instance.forms.values.any?(&:in_crossover_period?)
+  end
+
+  def bulk_upload_path(id:)
+    case log_type
+    when "lettings"
+      bulk_upload_lettings_log_path(id:)
```
kosiakkatrina: Could we have something like "bulk_upload_#{log_type}_log_path" instead? Wouldn't need the method. Happy to omit if it makes it less clear tho.
edit: sorry was looking at this by commit 🤦 feel free to ignore

asmega: No worries, yeah I went back and made a few tweaks. I think your suggestion would be for the latest version:
```ruby
public_send("bulk_upload_#{log_type_for_controller(controller)}_log_path", id:)
```
I'd say its about 50/50 and you could go either way. You'd get rid of the if switch for a single line which but arguably is less easier to read. I don't think this code is high churn will remain fairly stable once implemented so don't think the implementation here matters much

app/models/form.rb
asmega: I see, yeah I wasn't aware they overlapped that way. Let me have a go and I'll get back you.

asmega: I've done this in another PR https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/1043 the commit is https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/1043/commits/0528cd572137c1b59474b08766883f16f1d84c26

asmega:
This work has been merged further downstream in https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/1032 and can therefore be safely closed off

app/controllers/bulk_upload_lettings_logs_controller.rb
```diff
@@ -0,0 +1,44 @@
...
+  end
+
+  def form
+    @form ||= case params[:id]
+              when "year"
+                Forms::BulkUploadLettings::Year.new(form_params)
+              when "prepare-your-file"
+                Forms::BulkUploadLettings::PrepareYourFile.new(form_params)
+              else
+                raise "Page not found for path #{params[:id]}"
```
kosiakkatrina, line 37: Would it make sense to `render_not_found` for this case at some point?

PR #1029, opened 2022-11-24
WIP CLDC-1688 Update logs page + banner

CheshireSwift:
E-mails, banner and update logs page, along with the column to mark the affected logs.

E-mail template is already configured in Notify :)

Approach for the actual questions when they click through to update is up in the air - probably wants to share behaviour with existing form, but the unusual flow (specific subset of questions, flash notice to return to Update logs when done) means we can't just deep link into the normal flow. Haven't found a good way to do that yet.

### DIFF
--- a/app/controllers/locations_controller.rb
+++ b/app/controllers/locations_controller.rb
@@ -43,8 +43,16 @@ class LocationsController < ApplicationController
   end
 
   def deactivate
-    if @location.location_deactivation_periods.create!(deactivation_date: params[:deactivation_date]) && reset_location_and_scheme_for_logs!
+    if @location.location_deactivation_periods.create!(deactivation_date: params[:deactivation_date])
+      logs = reset_location_and_scheme_for_logs!
+
       flash[:notice] = deactivate_success_notice
+      LocationOrSchemeDeactivationMailer.new.send_deactivation_mails(
+        logs.to_a,
+        url_for(controller: "update_logs", action: "show"),
+        @location.scheme.service_name,
+        @location.postcode,
+      )
     end
     redirect_to scheme_location_path(@scheme, @location)
   end
@@ -214,7 +222,9 @@ private
   end
 
   def reset_location_and_scheme_for_logs!
-    @location.lettings_logs.filter_by_before_startdate(params[:deactivation_date].to_time).update!(location: nil, scheme: nil)
+    logs = @location.lettings_logs.filter_by_before_startdate(params[:deactivation_date].to_time)
+    logs.update!(location: nil, scheme: nil, impacted_by_scheme_deactivation: true)
+    logs
   end
 
   def toggle_date(key)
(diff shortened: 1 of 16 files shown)


app/controllers/locations_controller.rb
```diff
@@ -43,8 +43,16 @@ def deactivate_confirm
   end
 
   def deactivate
-    if @location.location_deactivation_periods.create!(deactivation_date: params[:deactivation_date]) && reset_location_and_scheme_for_logs!
+    if @location.location_deactivation_periods.create!(deactivation_date: params[:deactivation_date])
+      logs = reset_location_and_scheme_for_logs!
```
CheshireSwift, line 47: This technically removes `reset_location_...` from the conditional, but since it's using `update!` it already either returns true or throws, so behaviour should be identical.

app/controllers/locations_controller.rb
```diff
@@ -43,8 +43,16 @@ def deactivate_confirm
...
+
       flash[:notice] = deactivate_success_notice
+      LocationOrSchemeDeactivationMailer.new.send_deactivation_mails(
+        logs.to_a,
+        url_for(controller: "update_logs", action: "show"),
```
CheshireSwift, line 52: Not sure if there's a neater helper for this; we want the absolute path (`https://` and all) for the E-mail.

app/controllers/schemes_controller.rb
```diff
@@ -44,8 +44,15 @@ def deactivate_confirm
   end
 
   def deactivate
-    if @scheme.scheme_deactivation_periods.create!(deactivation_date: params[:deactivation_date]) && reset_location_and_scheme_for_logs!
+    if @scheme.scheme_deactivation_periods.create!(deactivation_date: params[:deactivation_date])
+      logs = reset_location_and_scheme_for_logs!
```
CheshireSwift, line 48: (See `location_controller` comment for why I'm okay with moving this out of the condition.)

app/views/logs/index.html.erb
```diff
@@ -4,6 +4,18 @@
...
+  <% if @logs_to_update > 0 %>
+    <%= govuk_notification_banner(
+      title_text: "Important",
+      title_heading_level: 3,
+      title_id: "impacted-logs-banner",
+    ) do |notification_banner| %>
+      <% notification_banner.heading(text: "A scheme has changed and it has affected #{@logs_to_update} #{@logs_to_update > 1 ? 'logs' : 'log'}") %>
+      <div class="govuk-notification-banner__heading">
+        <a class="govuk-notification-banner__link" href="./lettings-logs/update-logs">Update logs</a>
+      </div>
```
CheshireSwift, line 16: Using explicit divs rather than the `heading_href` option on `notification_banner.heading` to force the link onto a newline, per Slack discussion with Nicky.

CheshireSwift:
Closing in favour of #1040.

PR #1030, opened 2022-11-25
Add a default ordering for schemes

kosiakkatrina:
Add order for schemes that puts incomplete schemes first and keeps the service name ordering within that.
image

### DIFF
--- a/app/controllers/schemes_controller.rb
+++ b/app/controllers/schemes_controller.rb
@@ -9,7 +9,7 @@ class SchemesController < ApplicationController
 
   def index
     redirect_to schemes_organisation_path(current_user.organisation) unless current_user.support?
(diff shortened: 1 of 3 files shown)


app/controllers/schemes_controller.rb
```diff
@@ -9,7 +9,7 @@ class SchemesController < ApplicationController
 
   def index
     redirect_to schemes_organisation_path(current_user.organisation) unless current_user.support?
-    all_schemes = Scheme.all.order("service_name ASC")
+    all_schemes = Scheme.all.order(confirmed: :asc, service_name: :asc)
```
```suggestion
    all_schemes = Scheme.order(confirmed: :asc, service_name: :asc)
```

nit: `all` not required.

jamesrose (approved):

PR #1031, opened 2022-11-25
Fix typo on managing agents page

jamesrose:
We had a missing apostrophe.

### DIFF
--- a/app/views/organisation_relationships/managing_agents.html.erb
+++ b/app/views/organisation_relationships/managing_agents.html.erb
@@ -11,7 +11,7 @@
     <p class="govuk-body">This organisation does not currently have any managing agents.</p>
   <% end %>
 <% else %>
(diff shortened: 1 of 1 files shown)


app/views/organisation_relationships/managing_agents.html.erb
```diff
@@ -11,7 +11,7 @@
     <p class="govuk-body">This organisation does not currently have any managing agents.</p>
   <% end %>
 <% else %>
-  <%= render partial: "organisations/headings", locals: { main: "This organisation managing agents", sub: current_user.organisation.name } %>
+  <%= render partial: "organisations/headings", locals: { main: "This organisation's managing agents", sub: current_user.organisation.name } %>
```
kosiakkatrina: We should use apostrophe instead of a quote for user displayed content (’)

PR #1034, opened 2022-11-25
Cldc 1669 add location redesign

natdeanlewissoftwire:
This ticket splits the add a location flow into multiple pages, and removes it from the add a scheme flow (agreed to be the right thing to do with @nick-golding-softwire)

tested locally, and tests added for each controller method
risk if there are other areas the add a location flow is used that I'm not aware of, though I don't think there are (!)
no docs required

ticket: https://digital.dclg.gov.uk/jira/browse/CLDC-1669


app/controllers/locations_controller.rb
```diff
@@ -15,7 +15,143 @@ def index
...
+      end
+    end
+  end
+
+  def availability
+    if params[:location].present?
+      day = params[:location]["startdate(3i)"]
+      month = params[:location]["startdate(2i)"]
+      year = params[:location]["startdate(1i)"]
+      if [day, month, year].none?(&:blank?)
```
natdeanlewissoftwire: validation here doesn't happen on the model as we can't create a valid date object to add to `@location` unless the day, month and year are already valid
The only validation we could do on the model is of the date being out of range, but there's no mention of that in the designs

app/helpers/locations_helper.rb
```diff
@@ -25,18 +25,18 @@ def selection_options(resource)
...
-      { name: "Postcode", value: location.postcode },
-      { name: "Local authority", value: location.location_admin_district },
-      { name: "Location name", value: location.name, edit: true },
-      { name: "Total number of units at this location", value: location.units },
-      { name: "Common type of unit", value: location.type_of_unit },
-      { name: "Mobility type", value: location.mobility_type },
-      { name: "Code", value: location.location_code },
-      { name: "Availability", value: location_availability(location) },
+      { name: "Postcode", value: location.postcode, attribute: "postcode" },
+      { name: "Location name", value: location.name, attribute: "name" },
```
natdeanlewissoftwire, line 29: Name is above Local authority in the designs, even though they're the other order in the flow. This seems a bit odd to me but I can see why you'd want postcode and name at the top as they are more identifying (and show up in the details page as well as CYA)

app/views/locations/name.html.erb
```diff
@@ -0,0 +1,39 @@
+<% content_for :title, "Add a location to this scheme" %>
+
+<% content_for :before_content do %>
+  <%= govuk_back_link(
+        text: "Back",
+        href: if params[:referrer] == "check_answers"
+                scheme_location_check_answers_path(@scheme, @location)
+              else
+                params[:referrer] == "details" ? scheme_location_path(@scheme, @location) : scheme_location_postcode_path(@scheme, @location)
```
natdeanlewissoftwire: Even if the local authority is explicitly entered (this case should be rare), we go back to the postcode page (which would clear LA on postcode resubmission and redirect again to the LA page), the designs weren't clear and it seems unneccessarily fiddly to add a referrer to this route. I think this is fine as MVP but could be persuaded otherwise :)

app/controllers/locations_controller.rb
natdeanlewissoftwire: Update - I've added validation to the model for checking the date is between 1900 and 2200

app/controllers/locations_controller.rb
```diff
@@ -15,7 +15,147 @@ def index
...
+        else
+          redirect_to scheme_location_availability_path(@scheme, @location)
+        end
+      else
+        render :mobility_standards, status: :unprocessable_entity
+      end
+    end
+  end
+
+  def availability
```
natdeanlewissoftwire: this is now a manadatory field acording to the designs - it wasn't previously, will we have to update previous locations?

app/controllers/locations_controller.rb
```diff
@@ -15,7 +15,149 @@ def index
...
+        @location.errors.add :startdate, error_message
+        render :availability, status: :unprocessable_entity
+      end
+    end
+  end
+
+  def check_answers
+    if params[:location].present?
+      @location.confirmed = true
+      @location.save!
```
natdeanlewissoftwire: having discussed with the design team - I think this is the right place to mark the location as confirmed

app/controllers/locations_controller.rb
```diff
@@ -15,7 +15,165 @@ def index
   end
 
   def new
-    @location = Location.new
+    @location = Location.new(scheme: @scheme)
+    @location.save!
+    redirect_to scheme_location_postcode_path(@scheme, @location, route: params[:route])
+  end
```
jamesrose:
```suggestion
  def create
    @location = @scheme.locations.create!
    redirect_to scheme_location_postcode_path(@scheme, @location, route: params[:route])
  end
```

It's very usual to have `new` create a resource. Are you sure you don't want to `POST /locations` ?

app/controllers/locations_controller.rb
```diff
@@ -14,8 +15,144 @@ def index
...
+
+  def postcode
+    if params[:location].present?
+      @location.postcode = PostcodeService.clean(params[:location][:postcode])
+      if @location.postcode_changed?
+        @location.location_admin_district = nil
+        @location.location_code = nil
+      end
```
jamesrose: The latter half of this block especially belongs on the model.

app/controllers/locations_controller.rb
```diff
@@ -14,8 +15,144 @@ def index
...
+  end
+
+  def postcode; end
+
+  def update_postcode
+    @location.postcode = PostcodeService.clean(params[:location][:postcode])
+    if @location.postcode_changed?
+      @location.location_admin_district = nil
+      @location.location_code = nil
+    end
```
jamesrose:
```suggestion
    @location.postcode = PostcodeService.clean(params[:location][:postcode])
    if @location.postcode_changed?
      @location.location_admin_district = nil
      @location.location_code = nil
    end
```

This all belongs in the model.

app/controllers/locations_controller.rb
```diff
@@ -181,25 +233,15 @@ def authenticate_scope!
   end
 
   def authenticate_action!
-    if %w[new edit update create index edit_name edit_local_authority new_deactivation deactivate_confirm deactivate].include?(action_name) && !((current_user.organisation == @scheme&.owning_organisation) || current_user.support?)
+    if %w[create update index new_deactivation deactivate_confirm deactivate postcode local_authority name units type_of_unit mobility_standards availability check_answers].include?(action_name) && !((current_user.organisation == @scheme&.owning_organisation) || current_user.support?)
       render_not_found and return
     end
   end
 
-  def location_params
```
jamesrose, line 198: Re-introduce this method so that you don't need to extract params individually.

app/controllers/locations_controller.rb
```diff
@@ -14,8 +15,144 @@ def index
...
+
+  def postcode; end
+
+  def update_postcode
+    @location.postcode = PostcodeService.clean(params[:location][:postcode])
+    if @location.postcode_changed?
+      @location.location_admin_district = nil
+      @location.location_code = nil
+    end
+    if @location.save(context: :postcode)
```
jamesrose, line 27:
```suggestion
    if @location.update!(postcode: params[:location][:postcode], context: :postcode)
```

Once we re-introduce permitted params then we can send the location directly into the update.

app/controllers/locations_controller.rb
```diff
@@ -14,8 +15,144 @@ def index
...
+  end
+
+  def availability; end
+
+  def update_availability
+    day = params[:location]["startdate(3i)"]
+    month = params[:location]["startdate(2i)"]
+    year = params[:location]["startdate(1i)"]
+    if [day, month, year].none?(&:blank?)
+      if Date.valid_date?(year.to_i, month.to_i, day.to_i)
```
jamesrose: These validations belong on the model. Chat to @asmega

app/controllers/locations_controller.rb
```diff
@@ -14,8 +15,144 @@ def index
...
+      @location.errors.add :startdate, error_message
+      render :availability, status: :unprocessable_entity
+    end
+  end
+
+  def check_answers; end
+
+  def update_check_answers
+    @location.confirmed = true
+    @location.save!
```
jamesrose:
```suggestion
    @location.confirmed!
```

```ruby
class Location
  ...
  
  def confirmed!
     update!(confirmed: true)
   end
end
```

We should also have validations on `Location` that prevent confirmed from being set if it is not actually a complete record.

app/models/location.rb
```diff
@@ -402,17 +407,38 @@ def reactivating_soon?
...
 
-private
-
-  PIO = PostcodeService.new
-
   def validate_postcode
-    if postcode.nil? || !postcode&.match(POSTCODE_REGEXP)
+    if postcode.blank?
+      error_message = I18n.t("validations.location.postcode_blank")
+      errors.add :postcode, error_message
```
jamesrose: This can be extracted out into `validates :postcode, presence: true ...`

app/controllers/locations_controller.rb
jamesrose:
```ruby
required_params[:location_admin_district] = nil if required_params[:location_admin_district] == "Select an option"
```

app/models/location.rb
```diff
@@ -402,17 +407,38 @@ def reactivating_soon?
...
       errors.add :postcode, error_message
     end
   end
 
+  def validate_location_admin_district
+    if location_admin_district == "Select an option"
+      error_message = I18n.t("validations.location_admin_district")
+      errors.add :location_admin_district, error_message
+    end
+  end
```
jamesrose: Remove in favour of standard `presence: true` validation once you're rejecting the select option value at the params level.

app/views/locations/check_answers.html.erb
```diff
@@ -0,0 +1,47 @@
...
+                       details_html(attr)
+                     end %>
+                  <% end %>
+                  <% row.action(text: action_text_helper(attr, @location), href: location_edit_path(@location, attr[:attribute])) unless attr[:attribute] == "status" %>
+                  <% end %>
+              <% end %>
+          <% end %>
+      </div>
+  </div>
+  <%= f.hidden_field :confirmed %>
```
jamesrose: Remove

app/views/locations/check_answers.html.erb
```diff
@@ -0,0 +1,47 @@
...
+  ) %>
+<% end %>
+
+<%= form_for(@location, method: :patch, url: scheme_location_check_answers_path(@scheme, @location, route: params[:route])) do |f| %>
+  <%= render partial: "organisations/headings", locals: { main: "Check your answers", sub: "Add a location to #{@scheme.service_name}" } %>
+
+  <div class="govuk-grid-row">
+      <div class="govuk-grid-column-two-thirds-from-desktop">
+          <%= govuk_summary_list do |summary_list| %>
+              <% display_location_attributes(@location).each do |attr| %>
```
jamesrose: There's too much logic in this view. Can we extract it back into a/the helper?

app/views/locations/index.html.erb
```diff
@@ -50,14 +50,20 @@
...
+                                                                               end), { class: "govuk-!-font-weight-bold" }, wrapper_tag: "div")) %>
               <% row.cell(text: location.id) %>
               <% row.cell(text: status_tag(location.status)) %>
               <% end %>
           <% end %>
         <% end %>
       <% end %>
-      <%= govuk_button_link_to "Add a location", new_scheme_location_path(@scheme), secondary: true %>
+      <%= form_with model: @scheme, url: scheme_locations_path(@scheme), method: "post", local: true do |f| %>
+        <%= f.govuk_submit "Add a location", secondary: true %>
```
jamesrose: This can be a link_to with `method: :post` and Rails will largely generate what you've written here by itself.

https://apidock.com/rails/ActionView/Helpers/UrlHelper/link_to

app/views/locations/index.html.erb
```diff
@@ -95,19 +101,25 @@
...
+          <% row.cell(text: location.location_admin_district) %>
           <% row.cell(text: location.startdate&.to_formatted_s(:govuk_date)) %>
           <% end %>
       <% end %>
     <% end %>
   <% end %>
-  <%= govuk_button_link_to "Add a location", new_scheme_location_path(@scheme), secondary: true %>
+  <%= form_with model: @scheme, url: scheme_locations_path(@scheme), method: "post", local: true do |f| %>
+    <%= f.govuk_submit "Add a location", secondary: true %>
+  <% end %>
```
jamesrose: This can be a link_to with method: :post and Rails will largely generate what you've written here by itself.

https://apidock.com/rails/ActionView/Helpers/UrlHelper/link_to

app/controllers/locations_controller.rb
natdeanlewissoftwire: I don't think update takes a validation context so I've left this as a save

app/controllers/locations_controller.rb
natdeanlewissoftwire: Think he's ill today, I'll try to catch him tomorrow

app/views/locations/check_answers.html.erb
natdeanlewissoftwire: deprioritised as agreed, will work on this last

app/controllers/locations_controller.rb
```diff
@@ -2,9 +2,10 @@ class LocationsController < ApplicationController
   include Pagy::Backend
   before_action :authenticate_user!
   before_action :authenticate_scope!
-  before_action :find_location, except: %i[new create index]
+  before_action :find_location, except: %i[create index]
   before_action :find_scheme
   before_action :authenticate_action!
+  before_action :scheme_and_location_present, only: %i[postcode update_postcode local_authority update_local_authority name update_name units update_units type_of_unit update_type_of_unit mobility_standards update_mobility_standards availability update_availability check_answers update_check_answers]
```
jamesrose: `except` is better here.

jamesrose (approved):

PR #1035, opened 2022-11-25
Cldc 1668 update impacted logs

kosiakkatrina:
Mark logs as unresolved when they get impacted by scheme or location deactivation.

Display impacted logs in a table on `/lettings-logs/update-logs` page. Logs are marked as unresolved every time a deactivation happens, so this table might contain logs impacted by different deactivations.
image

When a user chooses to update the impacted log they get taken through tenancy start date, scheme and location questions.
image

After those questions are answered and the log is saved from check your answers page it gets marked as resolved, because answering those questions puts it back in the same state as it was before the deactivation happened.
image

On lettings log page there is a banner linking to `lettings-logs/update-logs`
image

If the user selects an unresolved log from the lettings log list they will get guided through the same journey of answering tenancy start date, scheme and location question until the log is resolved.


app/controllers/lettings_logs_controller.rb
```diff
@@ -83,6 +83,18 @@ def email_csv
 
   def csv_confirmation; end
 
+  def update_logs
+    respond_to do |format|
+      format.html do
+        impacted_logs = current_user.lettings_logs.where(impacted_by_deactivation: true, created_by: current_user)
```
kosiakkatrina: could be scope

jamesrose: I don't think we want this to be scoped by the user that created the log. Any user can fix the logs, but the creator is the one that will be notified.

kosiakkatrina: so the logs view would be the same for all the users within the org?

jamesrose: I think so, but worth checking that with Jemma et al.

CheshireSwift:
All lgtm (and mostly quite familiar 😁)

app/controllers/lettings_logs_controller.rb
```diff
@@ -37,6 +38,7 @@ def update
   def show
     respond_to do |format|
       # We don't have a dedicated non-editable show view
+      mark_logs_resolved
```
jamesrose:
```suggestion
      mark_logs_resolved!
```

app/controllers/lettings_logs_controller.rb
```diff
@@ -51,7 +53,11 @@ def show
   def edit
     @log = current_user.lettings_logs.find_by(id: params[:id])
     if @log
-      render "logs/edit", locals: { current_user: }
+      if @log.unresolved
+        redirect_to(send("lettings_log_#{@log.form.get_question('startdate', @log).page.id}_path", @log))
```
jamesrose:
```suggestion
        redirect_to(send("lettings_log_#{@log.form.get_question('startdate', @log).page.id}_path", @log))
```

Do we have a helper to do this? Please could you create one, if not?

app/controllers/lettings_logs_controller.rb
```diff
@@ -96,4 +115,13 @@ def find_resource
...
     lettings_log_url(log)
   end
+
+  def mark_logs_resolved
+    if @log&.unresolved == true && @log.location.present? && @log.scheme.present? && @log.update(unresolved: false)
+      notice_message = "You’ve updated all the fields affected by the scheme change."
+      unresolved_logs_count_for_user = current_user.lettings_logs.where(unresolved: true, created_by: current_user).count
+      notice_message << " <a href=\"/lettings-logs/update-logs\">Update #{unresolved_logs_count_for_user} more logs</a>" if unresolved_logs_count_for_user.positive?
+      flash[:notice] = notice_message.html_safe
+    end
```
jamesrose:
```suggestion
  def resolve_logs!
    if @log&.unresolved && @log.location.present? && @log.scheme.present? && @log.resolve!
      notice_message = "You’ve updated all the fields affected by the scheme change."
      unresolved_logs_count_for_user = current_user.lettings_logs.where(unresolved: true, created_by: current_user).count
      notice_message << " <a href=\"/lettings-logs/update-logs\">Update #{unresolved_logs_count_for_user} more logs</a>" if unresolved_logs_count_for_user.positive?
      flash[:notice] = notice_message.html_safe
    end
```

The notice generation here probably doesn't want to be in the controller.

app/controllers/lettings_logs_controller.rb
jamesrose: Ask the form for this, instead of hard-coding in controller

app/controllers/lettings_logs_controller.rb
```diff
@@ -83,6 +89,19 @@ def email_csv
 
   def csv_confirmation; end
 
+  def update_logs
+    respond_to do |format|
+      format.html do
+        flash[:notice] = nil
```
jamesrose: flash.now

app/controllers/lettings_logs_controller.rb
```diff
@@ -83,6 +89,19 @@ def email_csv
...
+        impacted_logs = current_user.lettings_logs.where(unresolved: true, created_by: current_user)
```
jamesrose: This can be scopes!

app/controllers/lettings_logs_controller.rb
```diff
@@ -13,6 +13,7 @@ def index
         @pagy, @logs = pagy(unpaginated_filtered_logs)
         @searched = search_term.presence
         @total_count = all_logs.size
+        @unresolved_count = all_logs.where(unresolved: true, created_by: current_user).count
```
jamesrose: This can be scopes!

app/controllers/lettings_logs_controller.rb
jamesrose:
```ruby
  def resolve_logs!
    if @log&.unresolved && @log.location.present? && @log.scheme.present? && @log.resolve!
      count = current_user.lettings_logs.where(unresolved: true, created_by: current_user).count
      flash.now[:notice] = flash_notice_for_resolved_logs(count)
    end
  end
```

app/models/form.rb
```diff
@@ -56,19 +56,23 @@ def subsection_for_page(page)
...
-              else
-                page_ids[page_index + 1]
-              end
-    nxt_page = get_page(page_id)
-
-    return :check_answers if nxt_page.nil?
-    return nxt_page.id if nxt_page.routed_to?(log, current_user)
-
-    next_page(nxt_page, log, current_user)
+    if log.unresolved
```
jamesrose: Return early

app/views/logs/index.html.erb
```diff
@@ -4,6 +4,18 @@
 <% content_for :title, title %>
 
 <% if current_page?(controller: 'lettings_logs', action: 'index') %>
+  <% if @unresolved_count > 0 %>
+    <%= govuk_notification_banner(
+      title_text: "Important",
+      title_heading_level: 3,
+      title_id: "impacted-logs-banner",
+    ) do |notification_banner| %>
+        <% notification_banner.heading(text: "A scheme has changed and it has affected #{@unresolved_count} #{@unresolved_count > 1 ? 'logs' : 'log'}") %>
```
jamesrose:
```suggestion
        <% notification_banner.heading(text: "A scheme has changed and it has affected #{@unresolved_count} #{@unresolved_count > 1 ? 'logs' : 'log'}") %>
```

You can use `pluralize` here!

app/views/logs/index.html.erb
```diff
@@ -4,6 +4,18 @@
...
+        <div class="govuk-notification-banner__heading">
+          <a class="govuk-notification-banner__link" href="/lettings-logs/update-logs">Update logs</a>
```
jamesrose: Rails route!

app/views/logs/update_logs.html.erb
```diff
@@ -0,0 +1,46 @@
...
+                    }) do %>
+                        <%= govuk_link_to("Update now", send("lettings_log_#{log.form.get_question('startdate', log).page.id}_path", log)) %>
+                <% end %>
+            <% end %>
+        <% end %>
+    <% end %>
+    <% end %>
+<% end %>
+
+<%== render partial: "pagy/nav", locals: { pagy: @pagy, item_name: "logs" } %>
```
jamesrose:
```suggestion
<%= render partial: "pagy/nav", locals: { pagy: @pagy, item_name: "logs" } %>
```

app/views/logs/update_logs.html.erb
```diff
@@ -0,0 +1,46 @@
...
+            <% body.row do |row| %>
+                <% row.cell(text: log.id) %>
+                <% row.cell(text: log.tenancycode) %>
+                <% row.cell(text: log.propcode) %>
+                <% row.cell(text: status_tag(log.status)) %>
+                <% row.cell(html_attributes: {
+                    scope: "row",
+                    class: "govuk-!-text-align-right",
+                    }) do %>
+                        <%= govuk_link_to("Update now", send("lettings_log_#{log.form.get_question('startdate', log).page.id}_path", log)) %>
```
jamesrose:
```suggestion
                        <%= govuk_link_to("Update now", send("lettings_log_#{log.form.get_question('startdate', log).page.id}_path", log)) %>
```

This should be in a helper because you've used it twice

app/controllers/lettings_logs_controller.rb
kosiakkatrina: Not a helper, but this is now a bit smaller

app/views/logs/update_logs.html.erb
kosiakkatrina: Same as above

app/controllers/lettings_logs_controller.rb
```diff
@@ -51,7 +53,11 @@ def show
   def edit
     @log = current_user.lettings_logs.find_by(id: params[:id])
     if @log
-      render "logs/edit", locals: { current_user: }
+      if @log.unresolved
+        redirect_to(send("lettings_log_#{@log.form.unresolved_log_redirect_page_id}_path", @log))
```
kosiakkatrina: Extract into a form method

app/controllers/lettings_logs_controller.rb
```diff
@@ -96,4 +114,11 @@ def find_resource
   def post_create_redirect_url(log)
     lettings_log_url(log)
   end
+
+  def resolve_logs!
+    if @log&.unresolved == true && @log.location.present? && @log.scheme.present? && @log.update(unresolved: false)
```
jamesrose:
```suggestion
    if @log&.unresolved && @log.location.present? && @log.scheme.present? && @log.resolve!
```

app/views/logs/index.html.erb
```diff
@@ -4,6 +4,18 @@
...
 <% if current_page?(controller: 'lettings_logs', action: 'index') %>
+  <% if @unresolved_count > 0 %>
+    <%= govuk_notification_banner(
+      title_text: "Important",
+      title_heading_level: 3,
+      title_id: "impacted-logs-banner",
+    ) do |notification_banner| %>
+        <% notification_banner.heading(text: "A scheme has changed and it has affected #{@unresolved_count} #{'log'.pluralize(@unresolved_count)}") %>
+        <div class="govuk-notification-banner__heading">
+          <a class="govuk-notification-banner__link" href="<%= update_logs_lettings_logs_path %>">Update logs</a>
```
jamesrose: Please turn `href` to `link_to`

app/views/logs/update_logs.html.erb
```diff
@@ -0,0 +1,46 @@
...
+            <% body.row do |row| %>
+                <% row.cell(text: log.id) %>
+                <% row.cell(text: log.tenancycode) %>
+                <% row.cell(text: log.propcode) %>
+                <% row.cell(text: status_tag(log.status)) %>
+                <% row.cell(html_attributes: {
+                    scope: "row",
+                    class: "govuk-!-text-align-right",
+                    }) do %>
+                        <%= govuk_link_to("Update now", send("lettings_log_#{log.form.unresolved_log_redirect_page_id}_path", log)) %>
```
jamesrose: This should use your new helper

app/controllers/lettings_logs_controller.rb
jamesrose:
```suggestion
        redirect_to(send("lettings_log_#{@log.form.unresolved_log_redirect_page_id}_path", @log))
```

Can this be constructed in a helper? Or at worse `Form` like `cancel_path` is?

jamesrose (approved):

PR #1036, opened 2022-11-25
[CLDC-1514] Add buyer 2 income question

ghost:
https://digital.dclg.gov.uk/jira/browse/CLDC-1514


app/models/form/sales/questions/buyer2_relationship_to_buyer1.rb
```diff
@@ -1,3 +1,4 @@
+# this works
```
asmega: is this comment needed?

asmega (approved):

app/models/form/sales/questions/buyer2_relationship_to_buyer1.rb
ghost: no! will remove

PR #1039, opened 2022-11-29
Add sorting per org

kosiakkatrina:
Pull incomplete schemes to the top for specific organisations: 
￼

So far we're only doing that for support user view (all schemes) and not for specific orgs, this should fix it
Also pull schemes with confirmed nil values to the top as they are also incomplete

### DIFF
--- a/app/controllers/organisations_controller.rb
+++ b/app/controllers/organisations_controller.rb
@@ -19,7 +19,7 @@ class OrganisationsController < ApplicationController
   end
 
   def schemes
-    all_schemes = Scheme.where(owning_organisation: @organisation).order("confirmed ASC NULLS FIRST", service_name: :asc)
+    all_schemes = Scheme.where(owning_organisation: @organisation).order_by_completion.order_by_service_name
 
     @pagy, @schemes = pagy(filtered_collection(all_schemes, search_term))
     @searched = search_term.presence
--- a/app/controllers/organisations_controller.rb
+++ b/app/controllers/organisations_controller.rb
@@ -19,7 +19,7 @@ class OrganisationsController < ApplicationController
   end
 
   def schemes
-    all_schemes = Scheme.where(owning_organisation: @organisation)
+    all_schemes = Scheme.where(owning_organisation: @organisation).order(confirmed: :asc, service_name: :asc)
 
     @pagy, @schemes = pagy(filtered_collection(all_schemes, search_term))
     @searched = search_term.presence
--- a/app/controllers/organisations_controller.rb
+++ b/app/controllers/organisations_controller.rb
@@ -19,7 +19,7 @@ class OrganisationsController < ApplicationController
   end
 
   def schemes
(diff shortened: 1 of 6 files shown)


spec/models/scheme_spec.rb
```diff
@@ -180,14 +180,17 @@
...
     it "can sort the schemes by status" do
-      all_schemes = described_class.all.order(confirmed: :asc, service_name: :asc)
-      expect(all_schemes.count).to eq(7)
+      all_schemes = described_class.all.order("confirmed ASC NULLS FIRST", service_name: :asc)
+      expect(all_schemes.count).to eq(9)
       expect(all_schemes[0].status).to eq(:incomplete)
       expect(all_schemes[1].status).to eq(:incomplete)
       expect(all_schemes[2].status).to eq(:incomplete)
+      expect(all_schemes[3].status).to eq(:incomplete)
+      expect(all_schemes[4].status).to eq(:incomplete)
```
jamesrose: What does this actually test?

app/controllers/schemes_controller.rb
```diff
@@ -9,7 +9,7 @@ class SchemesController < ApplicationController
 
   def index
     redirect_to schemes_organisation_path(current_user.organisation) unless current_user.support?
-    all_schemes = Scheme.order(confirmed: :asc, service_name: :asc)
+    all_schemes = Scheme.order("confirmed ASC NULLS FIRST", service_name: :asc)
```
```suggestion
    all_schemes = Scheme.order("confirmed ASC NULLS FIRST", service_name: :asc)
```

Why do we have `nil` confirmed values? It shouldn't be possible.

kosiakkatrina: Hm, we only set confirmed to true if it's confirmed, otherwise we don't set it at all 🤷‍♀️

spec/models/scheme_spec.rb
kosiakkatrina: Good question. I think this was more of a test for me, which doesn't really mean much. I'll fix this. 😅

app/controllers/schemes_controller.rb
jamesrose: I think what you had previously would work. What do you think?

kosiakkatrina: The tests would fail if I removed this. I think by default when ordering the nulls with ASC they go in the back `SQL considers NULL to be a value less than any present value`

kosiakkatrina: Which after reading this sentence it doesn't make sense (but that's how it works)

app/controllers/organisations_controller.rb
```diff
@@ -19,7 +19,7 @@ def index
   end
 
   def schemes
-    all_schemes = Scheme.where(owning_organisation: @organisation)
+    all_schemes = Scheme.where(owning_organisation: @organisation).order("confirmed ASC NULLS FIRST", service_name: :asc)
```
kosiakkatrina: extract scope

jamesrose (approved):

PR #1040, opened 2022-11-29
CLDC-1688 scheme change mailer

CheshireSwift:
Extracted just the E-mail parts of #1029, updated URL to match #1035.

As per #1029, E-mail template exists and is configured to match designs 🙂

### DIFF
--- a/app/controllers/locations_controller.rb
+++ b/app/controllers/locations_controller.rb
@@ -54,7 +54,7 @@ class LocationsController < ApplicationController
       flash[:notice] = deactivate_success_notice
       LocationOrSchemeDeactivationMailer.new.send_deactivation_mails(
         logs.to_a,
-        url_for(controller: "update_logs", action: "show"),
+        url_for(controller: "lettings_logs", action: "update_logs"),
         @location.scheme.service_name,
         @location.postcode,
       )
--- a/app/controllers/locations_controller.rb
+++ b/app/controllers/locations_controller.rb
@@ -48,8 +48,16 @@ class LocationsController < ApplicationController
   end
 
   def deactivate
-    if @location.location_deactivation_periods.create!(deactivation_date: params[:deactivation_date]) && reset_location_and_scheme_for_logs!
+    if @location.location_deactivation_periods.create!(deactivation_date: params[:deactivation_date])
+      logs = reset_location_and_scheme_for_logs!
+
       flash[:notice] = deactivate_success_notice
+      LocationOrSchemeDeactivationMailer.new.send_deactivation_mails(
+        logs.to_a,
+        url_for(controller: "update_logs", action: "show"),
+        @location.scheme.service_name,
+        @location.postcode,
+      )
     end
     redirect_to scheme_location_path(@scheme, @location)
   end
@@ -219,7 +227,9 @@ private
   end
 
   def reset_location_and_scheme_for_logs!
-    @location.lettings_logs.filter_by_before_startdate(params[:deactivation_date].to_time).update!(location: nil, scheme: nil, unresolved: true)
+    logs = @location.lettings_logs.filter_by_before_startdate(params[:deactivation_date].to_time)
+    logs.update!(location: nil, scheme: nil, unresolved: true)
+    logs
   end
 
   def toggle_date(key)
--- a/app/controllers/locations_controller.rb
+++ b/app/controllers/locations_controller.rb
@@ -52,12 +52,13 @@ class LocationsController < ApplicationController
       logs = reset_location_and_scheme_for_logs!
 
       flash[:notice] = deactivate_success_notice
-      LocationOrSchemeDeactivationMailer.new.send_deactivation_mails(
-        logs,
-        url_for(controller: "lettings_logs", action: "update_logs"),
-        @location.scheme.service_name,
-        @location.postcode,
-      )
+      logs.group_by(&:created_by).transform_values(&:count).compact.each do |user, count|
+        LocationOrSchemeDeactivationMailer.send_deactivation_mail(user,
+                                                                  count,
+                                                                  url_for(controller: "lettings_logs", action: "update_logs"),
+                                                                  @location.scheme.service_name,
+                                                                  @location.postcode).deliver_later
+      end
     end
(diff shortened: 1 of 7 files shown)


app/controllers/locations_controller.rb
```diff
@@ -48,8 +48,16 @@ def deactivate_confirm
   end
 
   def deactivate
-    if @location.location_deactivation_periods.create!(deactivation_date: params[:deactivation_date]) && reset_location_and_scheme_for_logs!
+    if @location.location_deactivation_periods.create!(deactivation_date: params[:deactivation_date])
```
CheshireSwift, line 51: This technically removes `reset_location_...` from the conditional, but since it's using `update!` it already either returns true or throws, so behaviour should be identical.

app/controllers/schemes_controller.rb
```diff
@@ -49,8 +49,15 @@ def deactivate_confirm
   end
 
   def deactivate
-    if @scheme.scheme_deactivation_periods.create!(deactivation_date: params[:deactivation_date]) && reset_location_and_scheme_for_logs!
+    if @scheme.scheme_deactivation_periods.create!(deactivation_date: params[:deactivation_date])
```
CheshireSwift, line 52: (Same comment as `locations_controller` re: removing `reset_location_and_scheme_for_logs!` from condition.)

app/mailers/location_or_scheme_deactivation_mailer.rb
```diff
@@ -0,0 +1,38 @@
+class LocationOrSchemeDeactivationMailer < NotifyMailer
+  DEACTIVATION_TEMPLATE_ID = "8d07a8c3-a4e3-4102-8be7-4ee79e4183fd".freeze
+
+  def send_deactivation_mail(user, log_count, update_logs_url, scheme_name, postcode = nil)
+    send_email(
+      user.email,
+      DEACTIVATION_TEMPLATE_ID,
+      {
+        log_count:,
+        log_or_logs: log_count == 1 ? "log" : "logs",
```
jamesrose, line 10: You can use https://apidock.com/rails/ActionView/Helpers/TextHelper/pluralize directly in the template here :)

CheshireSwift: When you say "in the template"; this is going out to Notify rather than ERB internally. Do we have access to it actually in the Notify template (as configured in the web interface), or do you just mean we can put `log_or_logs: pluralize(log_count, "log")` here?

tl;dr do we have access to these sort of helpers in Notify templates?

app/mailers/location_or_scheme_deactivation_mailer.rb
```diff
@@ -0,0 +1,38 @@
...
+    counts_by_user(logs).each do |user, count|
+      send_deactivation_mail(user, count, update_logs_url, scheme_name, postcode) if user
+    end
+  end
+
+private
+
+  def counts_by_user(logs)
+    logs.each_with_object(Hash.new(0)) do |log, counts|
+      counts[log.created_by] += 1
```
kosiakkatrina: Is this going to create a hash with a key as an entire user or am I missing something?

kosiakkatrina (approved):
Looks good!

app/controllers/locations_controller.rb
```diff
@@ -48,8 +48,16 @@ def deactivate_confirm
...
+      logs = reset_location_and_scheme_for_logs!
+
       flash[:notice] = deactivate_success_notice
+      LocationOrSchemeDeactivationMailer.new.send_deactivation_mails(
+        logs.to_a,
```
jamesrose:
```suggestion
        logs,
```

Why `to_a`?

app/controllers/schemes_controller.rb
```diff
@@ -49,8 +49,15 @@ def deactivate_confirm
...
+      logs = reset_location_and_scheme_for_logs!
+
       flash[:notice] = deactivate_success_notice
+      LocationOrSchemeDeactivationMailer.new.send_deactivation_mails(
+        logs.to_a,
+        url_for(controller: "lettings_logs", action: "update_logs"),
+        @scheme.service_name,
+      )
```
```suggestion
      LocationOrSchemeDeactivationMailer.new.send_deactivation_mails(
        logs,
        url_for(controller: "lettings_logs", action: "update_logs"),
        @scheme.service_name,
      )
```

app/mailers/location_or_scheme_deactivation_mailer.rb
```diff
@@ -0,0 +1,38 @@
...
+    end
+  end
```
jamesrose:
```suggestion
  def send_deactivation_mails(logs, update_logs_url, scheme_name, postcode = nil)
        logs.group(:created_by).count.compact.each do |user, count|
      send_deactivation_mail(user, count, update_logs_url, scheme_name, postcode)
    end
  end

private
```

You can do the whole thing in ActiveRecord :)

config/routes.rb
```diff
@@ -111,6 +111,7 @@
       get "csv-download", to: "lettings_logs#download_csv"
       post "email-csv", to: "lettings_logs#email_csv"
       get "csv-confirmation", to: "lettings_logs#csv_confirmation"
+      get "update-logs", to: "lettings_logs#update_logs"
```
jamesrose: Does this route anywhere? I don't see the corresponding action.

app/mailers/location_or_scheme_deactivation_mailer.rb
CheshireSwift: It is, but James has highlighted an alternate approach for this anyway :)

config/routes.rb
CheshireSwift: It doesn't exist yet; I pulled this from Kat's PR to actually add the page. Without at least adding it to the routing, the URL construction for the E-mail fails. This seemed like the tidiest place to draw the line.

app/controllers/locations_controller.rb
CheshireSwift: I think at one point during development it made the ergonomics of the tests easier, but that no longer appears to be the case. Happy to remove.

jamesrose (approved):

PR #1043, opened 2022-11-29
[1716] [1717] Bulk upload part four

asmega:
# Context

- https://digital.dclg.gov.uk/jira/browse/CLDC-1716
- https://digital.dclg.gov.uk/jira/browse/CLDC-1717
- Further work around bulk upload logs journey
- This work builds upon https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/1032 so may want to review on commit basis

# The changes

- Add upload your file page and checking your file page to bulk upload journey
- Renamed existing `BulkUpload` class to `LegacyBulkUpload`. I assume this feature once completed will supersede the existing bulk upload code
- Create a record of the bulk upload when user uploads a file
- Upload CSV file to S3
- Fix several typos in various parts of the journey

# Notes for review

- I did attempt to use the `ruby-filemagic` gem. However had difficulty installing required dependencies on the environments due to our use of buildpacks which are a touch constraining. It's probably do-able but not worth the time or effort. At some point switching to deploy via Dockerfile may ease this
- For testing I've had to stub the file csv check. This means we lack coverage of this call. I did this as I'm not aware how to get the test and deploy environments to align. Switching to docker builds would rectify this lack of coverage 

# Screenshots

## Desktop

Screenshot 2022-11-30 at 15 44 28

## Mobile

Screenshot 2022-11-30 at 15 45 25


.github/workflows/production_pipeline.yml
```diff
@@ -253,4 +253,5 @@ jobs:
           cf set-env $APP_NAME S3_CONFIG $S3_CONFIG
           cf set-env $APP_NAME CSV_DOWNLOAD_PAAS_INSTANCE $CSV_DOWNLOAD_PAAS_INSTANCE
           cf set-env $APP_NAME SENTRY_DSN $SENTRY_DSN
+          cf set-env $APP_NAME CSV_BUCKET "dluhc-core-production-csv-bucket"
```
jamesrose: Apologies, I should have mentioned that this is already defined per-env above. See:

```
 "CSV_DOWNLOAD_PAAS_INSTANCE"=>"dluhc-core-production-csv-bucket",
```

asmega: that's done now

app/controllers/bulk_upload_lettings_logs_controller.rb
```diff
@@ -14,7 +14,7 @@ def show
   end
 
   def update
-    if form.valid?
+    if form.valid? && form.save!
```
kosiakkatrina, line 17: Wouldn't running this achieve the same result?
```suggestion
    if form.save
```

app/controllers/bulk_upload_sales_logs_controller.rb
```diff
@@ -14,7 +14,7 @@ def show
   end
 
   def update
-    if form.valid?
+    if form.valid? && form.save!
```
kosiakkatrina, line 17: Same as above?
Would it make sense to have a parent controller? Seems like a lot of the methods could be shared

app/controllers/bulk_upload_sales_logs_controller.rb
```diff
@@ -28,7 +28,7 @@ def current_year
   end
 
   def in_crossover_period?
-    FormHandler.instance.forms.values.any?(&:in_crossover_period?)
+    FormHandler.instance.in_crossover_period?
```
kosiakkatrina: Would it make sense to only check the sales forms crossover?

I suspect the crossover dates for lettings and sales would be the same, but in case for some reason they aren't

app/models/forms/bulk_upload_lettings/upload_your_file.rb
```diff
@@ -14,6 +21,49 @@ def view_path
       def back_path
         bulk_upload_lettings_log_path(id: "prepare-your-file", form: { year: })
       end
+
+      def year_combo
+        "#{year}/#{year + 1 - 2000}"
+      end
```
kosiakkatrina, line 27: Would it make sense to extract this somewhere? We're also using it in app/models/forms/bulk_upload_lettings/checking_file.rb and a couple of other places

app/models/forms/bulk_upload_lettings/upload_your_file.rb
```diff
@@ -14,6 +21,49 @@ def view_path
...
+        end
+
+        true
+      end
+
+    private
+
+      def upload_enabled?
+        !Rails.env.development?
+      end
```
kosiakkatrina: Should this be in the `FeatureToggle`?

jamesrose (approved):

app/models/forms/bulk_upload_lettings/upload_your_file.rb
asmega: Yep, I've just re-housed this to `FeatureToggle`

app/controllers/bulk_upload_sales_logs_controller.rb
asmega: These now check the relevant sales/lettings crossover period

app/controllers/bulk_upload_lettings_logs_controller.rb
asmega: It would if it was a `ActiveRecord` thing but unfortunately it's not, unless we decide to implement it to be that way

app/models/forms/bulk_upload_lettings/upload_your_file.rb
asmega: Personally I think this level of duplication is fine. I don't really see this code changing much. If there were many more pages being added I think I would consider a base class or maybe some kind of decorator thing. I don't think we stand to gain much if we were to DRY this up at this moment in time

PR #1044, opened 2022-11-29
Cldc 1211 close collection period

kosiakkatrina:
Display a different text if collection window is closed. Use `end_date` form attribute to determine if the logs can still be edited for that collection. 
image
image
Do not display change links in review and check your answers paged for logs with closed collection windows. Redirect to the tasklist if the user navigates directly to the question for such log.
image

Stub `end_date ` for forms in failing tests so that the collection period is returned as open.

### DIFF
--- a/app/controllers/form_controller.rb
+++ b/app/controllers/form_controller.rb
@@ -50,7 +50,7 @@ class FormController < ApplicationController
       page_id = request.path.split("/")[-1].underscore
       @page = @log.form.get_page(page_id)
       @subsection = @log.form.subsection_for_page(@page)
-      if @page.routed_to?(@log, current_user)
+      if @page.routed_to?(@log, current_user) && @log.collection_period_open?
         render "form/page"
       else
         redirect_to lettings_log_path(@log)
--- a/app/controllers/form_controller.rb
+++ b/app/controllers/form_controller.rb
@@ -2,6 +2,7 @@ class FormController < ApplicationController
   before_action :authenticate_user!
   before_action :find_resource, only: %i[submit_form review]
   before_action :find_resource_by_named_id, except: %i[submit_form review]
+  before_action :check_collection_period, only: %i[submit_form show_page]
 
   def submit_form
     if @log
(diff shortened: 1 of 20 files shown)


app/controllers/form_controller.rb
```diff
@@ -50,7 +50,7 @@ def show_page
       page_id = request.path.split("/")[-1].underscore
       @page = @log.form.get_page(page_id)
       @subsection = @log.form.subsection_for_page(@page)
-      if @page.routed_to?(@log, current_user)
+      if @page.routed_to?(@log, current_user) && @log.collection_period_open?
```
jamesrose: I suspect that you might want this for both `show_page` and `submit_form` — did you consider a `before_action` so stop execution and redirect if the collection period is closed?

app/helpers/tasklist_helper.rb
```diff
@@ -36,4 +36,12 @@ def subsection_link(subsection, log, current_user)
       subsection.label
     end
   end
+
+  def review_log_text(log)
+    if log.collection_period_open?
+      "You can #{govuk_link_to 'review and make changes to this log', "/lettings-logs/#{log.id}/review"} until #{(log.form.end_date + 1.day).to_formatted_s(:govuk_date)}.".html_safe
```
jamesrose:
```suggestion
      "You can #{govuk_link_to 'review and make changes to this log', "/lettings-logs/#{log.id}/review"} until #{(log.form.end_date + 1.day).to_formatted_s(:govuk_date)}.".html_safe
```

Can you change this to a route helper? Should it be `+1 day`?

kosiakkatrina: Hm, good question, am double checking, the ticket says 2nd June 2022, but maybe it should be updated on the form level

jamesrose: I think whatever date it is, it belongs on the form.

jamesrose (approved):
Please could you check whether this should apply to support users?

PR #1046, opened 2022-12-01
Update hidden_in_check_answers for managing organisation

kosiakkatrina:
Hide managing organisation question in check answers if it was not routed to.
This is because we cannot access the managing organisation page if it's not routed to (it gets redirected to tasklist) and the user has not answered the managing organisation question if it was not routed to, it was most likely inferred.

### DIFF
--- a/app/models/form/lettings/questions/managing_organisation.rb
+++ b/app/models/form/lettings/questions/managing_organisation.rb
@@ -45,9 +45,9 @@ class Form::Lettings::Questions::ManagingOrganisation < ::Form::Question
     true
   end
 
-  def hidden_in_check_answers?(_log, user = nil)
+  def hidden_in_check_answers?(log, user = nil)
     @current_user = user
-    @current_user.nil?
+    @current_user.nil? || !@page.routed_to?(log, user)
   end
 
   def enabled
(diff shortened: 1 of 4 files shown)


app/models/form/lettings/questions/managing_organisation.rb
```diff
@@ -45,9 +45,9 @@ def derived?
     true
   end
 
-  def hidden_in_check_answers?(_log, user = nil)
+  def hidden_in_check_answers?(log, user = nil)
     @current_user = user
-    @current_user.nil?
+    @current_user.nil? || !@page.routed_to?(log, user)
```
jamesrose, line 50: Is there _ever_ a time that we want a value that can't be routed to in the CYA page?

kosiakkatrina: Hm, good question.
I think if we might have an opposite situation (If we have 2 questions on the page and one of them is conditional.
We still want to route to the page even if one of the questions doesn't show up)

But I can't think of any case where the situation you describe would happen. 
I've added the check to the default method too now.

jamesrose (approved):

PR #1060, opened 2022-12-06
Update toggle scheme/location button

kosiakkatrina:
Only display deactivate/reactivate this scheme/location button for active and deactivated schemes.
The "reactivate" or "deactivate" button should not be present, until the date where change takes place, in order to stop the scheduling of more than one change at a time.

Buttons displayed for deactivated and active:
image
image

Buttons not displayed for reactivating soon, deactivating soon, activating soon
image
image
image

### DIFF
--- a/app/models/location.rb
+++ b/app/models/location.rb
@@ -383,7 +383,11 @@ class Location < ApplicationRecord
     location_deactivation_periods.order("created_at").last
   end
 
-  def status(date = Time.zone.now)
+  def status
+    @status ||= status_at(Time.zone.now)
+  end
+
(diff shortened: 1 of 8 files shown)


app/models/location.rb
```diff
@@ -398,6 +398,10 @@ def active?
     status == :active
   end
 
+  def deactivated?
+    status == :deactivated
```
jamesrose, line 405: We should memoize status — otherwise you are executing up to 3 queries per `deactivated?` and so on.

jamesrose (approved):

PR #1063, opened 2022-12-07
Add `providertype` to lettings exports and export all possible fields

jamesrose:
We previously determining the CSV columns to export by checking the first record that we were about to export. This is not practical because later rows might have more columns, so we would not export some data.

This PR exports all possible columns while maintaining the order that we've created them. 

We also add the `providertype` field where appropriate.


app/services/exports/lettings_log_export_service.rb
```diff
@@ -241,7 +242,7 @@ def build_export_csv(lettings_logs)
         lettings_logs.each do |lettings_log|
           attribute_hash = apply_cds_transformation(lettings_log, EXPORT_MODE[:csv])
           if attribute_keys.nil?
-            attribute_keys = attribute_hash.keys
+            attribute_keys = EXPORT_FIELDS.sort_by { |field| LettingsLog.attribute_names.index(field) || Float::INFINITY }
```
kosiakkatrina, line 246: Is there any reason we're sorting it like this? As in, are we sure it's the order we want?

jamesrose: It maintains the original order, but with extra fields. I checked though: CDS will accept it in any order.

PR #1064, opened 2022-12-07
Fix flaky seed test

asmega:
# Context

- There is a flaky test when testing the seeding process
- The theory is that by invoking rake tasks in the seeds it is able to break out of the transaction which does not get rolled back
- This may or may not cause a failure depending on the running ordering of test execution

# Changes

- Extract importing of rent ranges to a class
- Start using that class in relevant places instead of invoking as a rake task


kosiakkatrina (approved):
🚢

PR #1068, opened 2022-12-08
Cldc 1627 update age in check answers and put persons on different check answers cards

kosiakkatrina:
## Part of the ticket
- Change the "Buyer 1's age" label in check your answers to "Lead buyer's age" and only display one label for `age known` and `age` questions.
image

- When lead buyer's age is known display the age in check your answers:
image

- When lead buyer's age is not known display the `Not known` label in check your answers:
image

- When lead buyer's prefers not to say the age display `Prefers not to say` label in check your answers (wording updated in #1077):
image

## General clean up
- Update the form to work the same way for the age questions for the rest of the people.
- Add `check_answers_card number` to the questions in household characteristics section in order to display different people on different cards (there's a different ticket to update the card labels):
image

- Update some buyer 2 questions to only be routed to if it is a joint purchase

### DIFF
--- a/app/models/form/sales/pages/person1_gender_identity_joint_purchase.rb
+++ b/app/models/form/sales/pages/person1_gender_identity_joint_purchase.rb
@@ -5,10 +5,12 @@ class Form::Sales::Pages::Person1GenderIdentityJointPurchase < ::Form::Page
     @header = ""
     @description = ""
     @subsection = subsection
-    @depends_on = [{
-      "jointpur" => 1,
-      "details_known_3" => 1,
-    }]
+    @depends_on = [
+      { "hholdcount" => 1, "jointpur" => 1 },
+      { "hholdcount" => 2, "jointpur" => 1 },
+      { "hholdcount" => 3, "jointpur" => 1 },
+      { "hholdcount" => 4, "jointpur" => 1 },
+    ]
   end
 
   def questions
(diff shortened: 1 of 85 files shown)


app/models/form/sales/pages/person2_age.rb
```diff
@@ -6,9 +6,7 @@ def initialize(id, hsh, subsection)
     @description = ""
     @subsection = subsection
     @depends_on = [
-      { "hholdcount" => 2, "details_known_2" => 1 },
-      { "hholdcount" => 3, "details_known_2" => 1 },
-      { "hholdcount" => 4, "details_known_2" => 1 },
+      { "details_known_2" => 1}
```
kosiakkatrina: We don't need household count here because the `details_known_2` question wouldn't get routed to in the first place if the household count was wrong

app/models/form/sales/pages/person1_gender_identity_joint_purchase.rb
```diff
@@ -5,10 +5,12 @@ def initialize(id, hsh, subsection)
...
-    @depends_on = [{
-      "jointpur" => 1,
-      "details_known_3" => 1,
-    }]
+    @depends_on = [
+      { "hholdcount" => 1, "jointpur" => 1 },
+      { "hholdcount" => 2, "jointpur" => 1 },
+      { "hholdcount" => 3, "jointpur" => 1 },
+      { "hholdcount" => 4, "jointpur" => 1 },
+    ]
```
kosiakkatrina, line 13: There is currently no `details_known_1` so depending on the household count to conditionally display the question. Might need to change when/if `details_known_1` gets added

natdeanlewissoftwire (approved):
LGTM :D

PR #1071, opened 2022-12-08
Store filename on bulk upload

asmega:
# Context

- At the moment we are not storing the filename when using bulk upload
- It will be needed later to playback back to the user

# Changes

- Store the filename of the file the user uploaded for bulk upload

### DIFF
--- a/db/migrate/20221207113235_add_bulk_upload_filename.rb
+++ b/db/migrate/20221207113235_add_bulk_upload_filename.rb
@@ -1,0 +1,5 @@
+class AddBulkUploadFilename < ActiveRecord::Migration[7.0]
+  def change
+    add_column :bulk_uploads, :filename, :text
+  end
+end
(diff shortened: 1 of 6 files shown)


db/migrate/20221207113235_add_bulk_upload_filename.rb
```diff
@@ -0,0 +1,5 @@
+class AddBulkUploadFilename < ActiveRecord::Migration[7.0]
+  def change
+    add_column :bulk_uploads, :filename, :text
```
jamesrose, line 3:
```suggestion
    add_column :bulk_uploads, :filename, :text, null: false
```

Can it be null?

jamesrose (approved):

db/migrate/20221207113235_add_bulk_upload_filename.rb
asmega: So i think I'd have to do some reading. My concern would be a browser for whatever reason doesn't send us the filename i think we therefore don't want to hard block here. We just store the name if we can otherwise let it be blank and could render back to the user as `unknown file` or something to that degree

jamesrose: Ah yes, that makes sense. We could render the S3 file name, but that's probably not useful either.

PR #1073, opened 2022-12-08
CLDC-1730-stock-owning-validation

natdeanlewissoftwire:
ticket: https://digital.dclg.gov.uk/jira/browse/CLDC-1730

tested locally
minimal risk
no docs required

This PR also includes moving organisation relationship errors previously in the controller to the model

Copy is pending so I've just used the same error as is displayed in the scheme if a user tries to add a non-stock owning org there for now.

error message: 
image


app/controllers/organisation_relationships_controller.rb
```diff
@@ -51,6 +51,11 @@ def create_housing_provider
         @organisations = Organisation.where.not(id: child_organisation.id).pluck(:id, :name)
         render "organisation_relationships/add_housing_provider"
         return
+      elsif !parent_organisation.holds_own_stock
+        @organisation.errors.add :related_organisation_id, I18n.t("validations.scheme.owning_organisation.does_not_own_stock")
+        @organisations = Organisation.where.not(id: child_organisation.id).pluck(:id, :name)
+        render "organisation_relationships/add_housing_provider"
+        return
```
natdeanlewissoftwire: will add the errors in this controller to the org relationship model

app/models/organisation_relationship.rb
```diff
@@ -1,4 +1,28 @@
...
+
+private
+
+  def validate_housing_provider_relationship
+    if parent_organisation_id.blank?
+      child_organisation.errors.add :related_organisation_id, "You must choose a housing provider"
+    elsif OrganisationRelationship.exists?(child_organisation:, parent_organisation:)
+      child_organisation.errors.add :related_organisation_id, "You have already added this housing provider"
+    elsif !parent_organisation.holds_own_stock
+      errors.add :parent_organisation_id
```
natdeanlewissoftwire: I had to add an error to the org relationship model as well here so the relationship wasn't incorrectly saved, as the displayed error is being added to the org so doesn't technically invalidate the relationship - is there a better way of doing this?

natdeanlewissoftwire: done by using an org relationship model in the view rather than an org

jamesrose (approved):

PR #1077, opened 2022-12-12
CLDC-1627 Enable multiple inferred answers

kosiakkatrina:
Builds on top of #1068 

This enables different inferred check answers values to be displayed based on a different condition.
Previously we were only able to set a single inferred value with one condition per question.

This is because for the lead tenant's age question we want to display an inferred `Not known` when age known is answered as `No`, and `Prefers not to say` when the question is answered `Buyer prefers not to say`
image
image

We also want to display `Prefers not to say` instead of `Buyer prefers not to say` in check your answers for all the other tenant questions, because the text wraps better on mobile screens:
image

### DIFF
--- a/app/models/form/question.rb
+++ b/app/models/form/question.rb
@@ -53,10 +53,7 @@ class Form::Question
     answer = label_from_value(log[id]) if log[id].present?
     answer_label = [prefix, format_value(answer), suffix_label(log)].join("") if answer
 
-    inferred = inferred_answer_value(log)
-    return inferred if inferred.present?
-
-    answer_label
+    inferred_answer_value(log) || answer_label
   end
 
   def get_inferred_answers(log)
@@ -113,7 +110,7 @@ class Form::Question
     return unless inferred_check_answers_value
 
     inferred_answer = inferred_check_answers_value.find { |inferred_value| inferred_value["condition"].values[0] == log[inferred_value["condition"].keys[0]] }
-    inferred_answer.present? ? inferred_answer["value"] : ""
+    inferred_answer["value"] if inferred_answer.present?
   end
 
   def displayed_answer_options(log, _current_user = nil)
--- a/app/models/form/question.rb
+++ b/app/models/form/question.rb
@@ -106,13 +106,6 @@ class Form::Question
     false
   end
 
-  def inferred_answer_value(log)
-    return unless inferred_check_answers_value
-
-    inferred_answer = inferred_check_answers_value.find { |inferred_value| inferred_value["condition"].values[0] == log[inferred_value["condition"].keys[0]] }
-    inferred_answer["value"] if inferred_answer.present?
-  end
-
   def displayed_answer_options(log, _current_user = nil)
     answer_options.select do |_key, val|
       !val.is_a?(Hash) || !val["depends_on"] || form.depends_on_met(val["depends_on"], log)
@@ -311,6 +304,13 @@ private
 
   def enabled_inferred_answers(inferred_answers, log)
     inferred_answers.filter { |_key, value| value.all? { |condition_key, condition_value| log[condition_key] == condition_value } }
+  end
+
+  def inferred_answer_value(log)
+    return unless inferred_check_answers_value
+
+    inferred_answer = inferred_check_answers_value.find { |inferred_value| inferred_value["condition"].values[0] == log[inferred_value["condition"].keys[0]] }
+    inferred_answer["value"] if inferred_answer.present?
   end
 
   RADIO_YES_VALUE = {
--- a/app/models/form/question.rb
+++ b/app/models/form/question.rb
@@ -53,7 +53,7 @@ class Form::Question
     answer = label_from_value(log[id]) if log[id].present?
     answer_label = [prefix, format_value(answer), suffix_label(log)].join("") if answer
 
-    inferred = get_inferred_answer_value(log)
+    inferred = inferred_answer_value(log)
     return inferred if inferred.present?
 
     answer_label
@@ -109,7 +109,7 @@ class Form::Question
     false
   end
 
-  def get_inferred_answer_value(log)
+  def inferred_answer_value(log)
     return unless inferred_check_answers_value
 
     inferred_answer = inferred_check_answers_value.find { |inferred_value| inferred_value["condition"].values[0] == log[inferred_value["condition"].keys[0]] }
--- a/app/models/form/question.rb
+++ b/app/models/form/question.rb
@@ -53,7 +53,7 @@ class Form::Question
     answer = label_from_value(log[id]) if log[id].present?
     answer_label = [prefix, format_value(answer), suffix_label(log)].join("") if answer
 
-    inferred = inferred_check_answers_value["value"] if inferred_check_answers_value && has_inferred_check_answers_value?(log)
+    inferred = get_inferred_answer_value(log)
     return inferred if inferred.present?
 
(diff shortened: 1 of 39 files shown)


app/models/form/question.rb
```diff
@@ -104,11 +101,18 @@ def derived?
 
   def has_inferred_check_answers_value?(log)
     return true if selected_answer_option_is_derived?(log)
-    return inferred_check_answers_value["condition"].values[0] == log[inferred_check_answers_value["condition"].keys[0]] if inferred_check_answers_value.present?
+    return inferred_check_answers_value&.any? { |inferred_value| inferred_value["condition"].values[0] == log[inferred_value["condition"].keys[0]] } if inferred_check_answers_value.present?
 
     false
   end
 
+  def inferred_answer_value(log)
```
ghost: this should be a private method?

kosiakkatrina: Good point!

app/models/form/question.rb
```diff
@@ -53,10 +53,7 @@ def answer_label(log)
     answer = label_from_value(log[id]) if log[id].present?
     answer_label = [prefix, format_value(answer), suffix_label(log)].join("") if answer
 
-    inferred = inferred_check_answers_value["value"] if inferred_check_answers_value && has_inferred_check_answers_value?(log)
-    return inferred if inferred.present?
-
-    answer_label
+    inferred_answer_value(log) || answer_label
```
SamSeed-Softwire, line 56: Does this not return a true/false value as it's a boolean evaluation? If so, how does that work given previously this method output an formatted answer to a question?

app/models/form/sales/questions/buyer1_ethnic_group.rb
```diff
@@ -7,12 +7,12 @@ def initialize(id, hsh, page)
     @type = "radio"
     @answer_options = ANSWER_OPTIONS
     @hint_text = "Buyer 1 is the person in the household who does the most paid work. If it’s a joint purchase and the buyers do the same amount of paid work, buyer 1 is whoever is the oldest."
-    @inferred_check_answers_value = {
+    @inferred_check_answers_value = [{
```
SamSeed-Softwire, line 10: Is this missing a test?

app/models/form/question.rb
kosiakkatrina: Hm, in this situation I think it acts like a fallback value. If inferred_answer_value(log) is nil, it will return the value of answer_label. I think we use this syntax in other places too. I think this is using .presence  but I think it works without https://apidock.com/rails/Object/presence

SamSeed-Softwire: Fair enough, thanks!

app/models/form/sales/questions/buyer1_ethnic_group.rb
kosiakkatrina: Nice spot!

app/models/form/sales/questions/buyer1_age_known.rb
```diff
@@ -18,6 +18,9 @@ def initialize(id, hsh, page)
         {
           "age1_known" => 1,
         },
+        {
+          "age1_known" => 2,
```
SamSeed-Softwire, line 22: Does this need a matching definition in `config/forms/2022_2023.json`?

kosiakkatrina: Hmm, I'd say out of scope for this ticket, but merging this would make it easier to adjust the lettings logs ones as well if we need to

SamSeed-Softwire: Ah I hadn't clocked that `config/forms/2022_2023.json` was only relevant for lettings - I assumed it had stuff for both lettings and sales in it.

app/models/form/question.rb
```diff
@@ -271,7 +268,7 @@ def selected_answer_option_is_derived?(log)
   end
 
   def has_inferred_display_value?(log)
-    inferred_check_answers_value.present? && log[inferred_check_answers_value["condition"].keys.first] == inferred_check_answers_value["condition"].values.first
+    inferred_check_answers_value.present? && inferred_check_answers_value.any? { |inferred_value| log[inferred_value["condition"].keys.first] == inferred_value["condition"].values.first }
```
SamSeed-Softwire, line 271: Why have you used `.first` here but `[0]` in `inferred_check_answers_value` and `inferred_answer_value`?

app/models/form/question.rb
SamSeed-Softwire: I can't find this documented anywhere. Could you point me to some other examples in the codebase please? :) (Or actual documentation if you can find any. Thanks!)

app/models/form/sales/questions/person_gender_identity.rb
```diff
@@ -6,6 +6,12 @@ def initialize(id, hsh, page, person_index:)
     @type = "radio"
     @answer_options = ANSWER_OPTIONS
     @check_answers_card_number = person_index
+    @inferred_check_answers_value = [{
+      "condition" => {
+        id => "R",
```
SamSeed-Softwire, line 11: Just so I can understand this - what is id here? Bit confused as to how this works. Thanks!

app/models/form/question.rb
kosiakkatrina: Hmm, maybe this https://www.visuality.pl/posts/what-does-the-operator-actually-mean-in-ruby `The || operator` part.
Some examples in the codebase:
https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/blob/main/app/controllers/schemes_controller.rb#L262
https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/blob/main/app/components/log_summary_component.html.erb#L62
https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/blob/main/app/controllers/users_controller.rb#L49

Let me know if you wanted more, but they're all similar

app/models/form/sales/questions/person_gender_identity.rb
kosiakkatrina: Ah, so id is the id of this question. Because we're reusing this question for all the people, the id might be different each time. 
What this says is that if this question is answered with "R", show "Prefers not to say" in the check your answers instead of the actual answer text

app/models/form/question.rb
kosiakkatrina: No reason, it does the same thing,  I can change the ones in `inferred_answer_value` if that's more readable

kosiakkatrina: Not sure which bit you meant by `inferred_check_answers_value` tho?

app/models/form/question.rb
SamSeed-Softwire: Amazing, thanks! Good to see it in action. Wish I could find it in the official docs for Ruby, but oh well 😁

app/models/form/question.rb
SamSeed-Softwire: Sorry that was a typo, I meant `has_inferred_check_answers_value?`

app/models/form/sales/questions/person_gender_identity.rb
SamSeed-Softwire: Thanks, makes sense! :)

SamSeed-Softwire (approved):
Ok I'm happy with everything! :) Feel free to get a review from someone else more experienced if you fancy though.

PR #1083, opened 2022-12-12
Cldc 1537 details for person 1

natdeanlewissoftwire:
This adds a "Do you know the details for person 1?" page and question as required by the newly clarified requirements for this ticket: https://digital.dclg.gov.uk/jira/browse/CLDC-1537

### DIFF
--- a/app/models/form/sales/pages/person2_known.rb
+++ b/app/models/form/sales/pages/person2_known.rb
@@ -7,9 +7,9 @@ class Form::Sales::Pages::Person2Known < ::Form::Page
     @description = ""
     @subsection = subsection
     @depends_on = [
-      { "hholdcount" => 2 },
-      { "hholdcount" => 3 },
-      { "hholdcount" => 4 },
+      { "hholdcount" => 2, "details_known_1" => 1  },
+      { "hholdcount" => 3, "details_known_1" => 1  },
+      { "hholdcount" => 4, "details_known_1" => 1  },
     ]
   end
 
--- a/app/views/form/headers/_person_1_known_page.erb
+++ b/app/views/form/headers/_person_1_known_page.erb
@@ -1,0 +1,1 @@
+You have given us the details for 0 of the <%= log.hholdcount %> other people in the household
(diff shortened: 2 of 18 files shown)


app/models/form/sales/pages/person2_known.rb
```diff
@@ -7,9 +7,9 @@ def initialize(id, hsh, subsection)
     @description = ""
     @subsection = subsection
     @depends_on = [
-      { "hholdcount" => 2 },
-      { "hholdcount" => 3 },
-      { "hholdcount" => 4 },
+      { "hholdcount" => 2, "details_known_1" => 1  },
+      { "hholdcount" => 3, "details_known_1" => 1  },
+      { "hholdcount" => 4, "details_known_1" => 1  },
```
kosiakkatrina, line 12: Hmm, would we still want to ask this question even if we don't know the details for person 1? 🤔

app/views/form/headers/_person_1_known_page.erb
```diff
@@ -0,0 +1 @@
+You have given us the details for 0 of the <%= log.hholdcount %> other people in the household
```
kosiakkatrina, line 1: Should this be there? Maybe should skip for the first person?

kosiakkatrina (approved):

app/models/form/sales/pages/person2_known.rb
natdeanlewissoftwire: I don't think so - see https://miro.com/app/board/uXjVPcAfS04=/

app/views/form/headers/_person_1_known_page.erb
natdeanlewissoftwire: Yeah could probably remove this, good point 👍

app/models/form/sales/pages/person2_known.rb
kosiakkatrina: Hmm, we display it for the lettings, so that might be a bit inconsistent, maybe worth checking with design team?
image

PR #1089, opened 2022-12-14
Cldc 1530 working situation

kosiakkatrina:
Add person1/2/3/4 working situation questions and pages
image
image

Add a working situation page that determines the question id (which is also the database field) based on whether it is a joint purchase.
Add a working situation question class, that determines the display content based on it's ID and whether it is a joint purchase (get it from the page)


app/models/form/sales/questions/person_working_situation.rb
```diff
@@ -0,0 +1,45 @@
...
+    "6" => { "value" => "Not seeking work" },
+    "8" => { "value" => "Unable to work due to long term sick or disability" },
+    "5" => { "value" => "Retired" },
+    "0" => { "value" => "Other" },
+    "10" => { "value" => "Buyer prefers not to say" },
+    "7" => { "value" => "Full-time student" },
+    "9" => { "value" => "Child under 16" },
+  }.freeze
+
+  def person_database_number
```
natdeanlewissoftwire: can we put the three methods in this file that are duplicated in question and page in a general helper file?

spec/factories/sales_log.rb
```diff
@@ -63,6 +63,7 @@
       prevown { 1 }
       sex3 { "X" }
       mortgage { 20_000 }
+      ecstat3 { 10 }
```
natdeanlewissoftwire, line 66: is it worth setting hholdcount to 4 here so we can test all the person x pages?

natdeanlewissoftwire (approved):
Just a couple of comments but happy to merge as is

app/models/form/sales/questions/person_working_situation.rb
kosiakkatrina: Hmm, if it's a helper we'd need to pass in the `PERSON_INDEX` into it every time... I'm not against the idea, but need to think what is the most elegant way to do it, thank you!

PR #1090, opened 2022-12-14
Ignore `created_by` when computing differences between imported logs

jamesrose:
This field is not set on the previous service. If we don't do this then the importer will complain that the value has changed, where in fact it _is_ always nil.


kosiakkatrina (approved):

kosiakkatrina:
Do we need a test for this?

jamesrose:
@kosiakkatrina It's not a critical path — just logging, but I will return and add one.

PR #1091, opened 2022-12-14
[1778] Bulk upload errors

asmega:
# Context

- https://digital.dclg.gov.uk/jira/browse/CLDC-1778
- After a user has bulk uploaded some logs we need to process those logs and generate relevant error messages
- This PR kicks off that process and adds certain mechanisms to achieve that but are not fully implemented and missing features and functionality that can be added on at a later date

# Changes

- After the user has uploaded a file enqueue an async job to process the file
- Async job downloads the file and process it to persist any errors to the database
- It then deletes the temporary downloaded file from local disk but leaves it in S3 
- Updated active job queue adapter for dev to `inline` and test to `test` so jobs firing up automatically in dev and tests can be called without the need to run sidekiq + redis
- Added a local disk storage service that runs in dev. This way developers do not need access to S3 whilst running locally and will instead upload to `./tmp/storage` instead and also download from there
- Errors are grouped by row as according to the designs

# Notes for review

- This does not add all possible validations these will be buffed up later
- Pagination has not been added to error page which may be potentially very long, this can be added at a later date
- No emails have been added to these flows and can be added at a later date
- There lacks coverage around the validation code. These can be buffed up when proper validations are actually introduced

# Out of scope for this PR

- We have a few things still to do but I think should be completed independent of this PR
- Need to exit early in the validation process to handle other scenarios such as it CSV is malformed instead of proceeding to attempt validation. This is so we can show relevant error messages to users for these specific problems
- Need more testing to handle lines endings. Windows seems to generate slightly different endings which will need to be handled
- No code has been added to handle byte order mark aka `BOM`

# Screenshots

## Lettings errors

Screenshot 2022-12-16 at 17 19 27


app/models/bulk_upload.rb
```diff
@@ -2,9 +2,14 @@ class BulkUpload < ApplicationRecord
...
 
   belongs_to :user
+  has_many :bulk_upload_errors
 
   after_initialize :generate_identifier, unless: :identifier
 
+  def year_combo
+    "#{year}/#{year - 2000 + 1}"
+  end
+
```
kosiakkatrina, line 12: This is in quite a few places now. Anyway we could extract it? Would having it in the form handler make sense?

app/models/forms/bulk_upload_sales/upload_your_file.rb
```diff
@@ -38,21 +38,28 @@ def save!
...
+        storage_service.write_file(bulk_upload.identifier, File.read(file.path))
+
+        ProcessBulkUploadJob.perform_later(bulk_upload:)
 
         true
       end
 
     private
 
       def upload_enabled?
```
kosiakkatrina: Does this need its own method? Could we just call FeatureToggle.upload_enabled?

app/services/bulk_upload/lettings/row_parser.rb
```diff
@@ -0,0 +1,185 @@
...
+
+  def field_mapping
+    {
+      field_134: :renewal,
+    }
+  end
+
+  def validate_possible_answers
+    field_mapping.each do |field, attribute|
+      possible_answers = FormHandler.instance.current_lettings_form.questions.find { |q| q.id == attribute.to_s }.answer_options.keys
```
kosiakkatrina, line 163: Should we be getting the form based on the `year` in `bulk_upload` table?

kosiakkatrina: Unless it's out of scope, in which case feel free to ignore

kosiakkatrina: would it make sense to map through the native_object and get the form from there?
edit: never mind, we need the field number 🤦

app/services/bulk_upload/lettings/validator.rb
```diff
@@ -0,0 +1,250 @@
+require "csv"
+
+class BulkUpload::Lettings::Validator
+  include ActiveModel::Validations
+
+  QUESTIONS = {
+    field_1: "What is the letting type?",
```
kosiakkatrina, line 7: This feels unnecessary (?) Is it different from the questions in the form? Could we get it based on the field id mapping (from `app/services/bulk_upload/lettings/row_parser.rb`)

app/models/bulk_upload.rb
asmega: Where would you extract to? You suggested `FormHandler` however `BulkUpload` has zero knowledge of `FormHandler` or forms in general, nor should they have knowledge of what a form even is?

app/models/forms/bulk_upload_sales/upload_your_file.rb
asmega: I've removed this now. Not really too fussed if it's private implementation

app/services/bulk_upload/lettings/row_parser.rb
asmega: this code will be removed in favour of running validations on the `Log` and then proxying them thru to the relevant fields for the CSV upload. although those changes are outside the scope of this PR as the main focus of this PR is to simply get one error which can be surfaced in the UI

jamesrose (approved):

PR #1092, opened 2022-12-15
Refactor age to reuse pages and questions

kosiakkatrina:
Refactor age page to reuse the same page and determine the question id based on whether the sale is joint purchase.
Add a check to the view to check if check answers card is present because if it's not a joint purchase, person 6 doesn't exist.


app/components/check_answers_summary_list_card_component.html.erb
```diff
@@ -1,39 +1,41 @@
...
-    <%= govuk_summary_list do |summary_list| %>
-      <% applicable_questions.each do |question| %>
-        <% summary_list.row do |row| %>
-          <% row.key { question.check_answer_label.to_s.presence || question.header.to_s } %>
-          <% row.value do %>
-            <span class="govuk-!-margin-right-4"><%= get_answer_label(question) %></span>
-            <% extra_value = question.get_extra_check_answer_value(log) %>
-            <% if extra_value %>
-              <span class="govuk-!-font-weight-regular app-!-colour-muted"><%= extra_value %></span>
+<% if applicable_questions.first %>
```
kosiakkatrina, line 1: Only adding this line here, the rest of the file is unchanged

natdeanlewissoftwire (approved):
LGTM :)

PR #1094, opened 2022-12-15
Refactor out person classes

kosiakkatrina:
Create new Person parent classes for person pages and questions.
To contain the shared logic that is relevant only for person 1-4 questions


app/models/form/sales/pages/person.rb
```diff
@@ -0,0 +1,14 @@
+class Form::Sales::Pages::Person < ::Form::Page
+  def initialize(id, hsh, subsection, person_index)
```
jamesrose: As `person_index` is just a magic number, please could you make it a keyword argument?

app/models/form/sales/pages/person_age.rb
```diff
@@ -1,11 +1,11 @@
...
-  def initialize(id, hsh, subsection)
+class Form::Sales::Pages::PersonAge < Form::Sales::Pages::Person
+  def initialize(id, hsh, subsection, person_index)
     super
     @header = ""
     @description = ""
     @subsection = subsection
     @depends_on = [
-      { "details_known_#{person_display_number(PERSON_INDEX)}" => 1, "jointpur" => joint_purchase? ? 1 : 2 },
+      { "details_known_#{person_display_number}" => 1, "jointpur" => joint_purchase? ? 1 : 2 },
```
jamesrose: Could we extract the field interpolation out as we do this quite a lot?

app/models/form/sales/pages/person.rb
```diff
@@ -0,0 +1,18 @@
...
+    joint_purchase? ? @person_index - 2 : @person_index - 1
+  end
+
+  def joint_purchase?
+    id.include?("_joint_purchase")
+  end
+
+  def details_known_question_id
+    "details_known_#{person_display_number}"
+  end
```
jamesrose, line 17:
```suggestion
  def field_for_person(field)
    [field, person_display_number].join
  end
  
  def field_for_person_index(field)
      [person_index, field].join
  end
```

Was there a motivation for not making it flexible for any other field?

kosiakkatrina: Oh, yes, sorry, that's fine. Have a look now. I've added another method for that tho, because details_known uses display number instead of the database field. That's because details_known actually starts counting from 1 in all cases (because we only ask details known about people and not buyers, no matter how many of them we have)

jamesrose (approved):

PR #1098, opened 2022-12-16
CLDC-1450 Update declaration question

kosiakkatrina:
Edit the privacy notice question:
- Display the correct error for privacy notice
- Display the partial_guidance at the top by default
- Display the privacy notice question unconditionally
- Disable the rest of the housing characteristics questions until this question is answered

image
image
image

### DIFF
--- a/app/models/form/question.rb
+++ b/app/models/form/question.rb
@@ -14,11 +14,11 @@ class Form::Question
   def initialize(id, hsh, page)
     @id = id
     @page = page
+    @guidance_position = GuidancePosition::TOP
     if hsh
       @check_answer_label = hsh["check_answer_label"]
       @header = hsh["header"]
       @guidance_partial = hsh["guidance_partial"]
-      @guidance_position = GuidancePosition::TOP
       @hint_text = hsh["hint_text"]
       @type = hsh["type"]
       @min = hsh["min"]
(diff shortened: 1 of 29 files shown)


app/models/form/question.rb
```diff
@@ -209,6 +209,7 @@ def display_label
 
   def unanswered_error_message
     return I18n.t("validations.declaration.missing") if id == "declaration"
+    return I18n.t("validations.privacynotice.missing") if id == "privacynotice"
```
ghost, line 209: do we have a test for this on question_spec.rb ?

app/models/form/question.rb
```diff
@@ -14,11 +14,11 @@ module GuidancePosition
   def initialize(id, hsh, page)
     @id = id
     @page = page
+    @guidance_position = GuidancePosition::TOP
```
asmega, line 17: why cant this be injected like everything else? and instead must be overridden?

kosiakkatrina: It can, just setting a default here so that we don't have to set it every time (only have to set it when it's at the bottom)
edit: we're already doing this, but only if it's a question from a json, this change is just moving it out of the `if hsh` condition on line 21

app/models/form/question.rb
kosiakkatrina: there's a test here `spec/models/form/sales/questions/privacy_notice_spec.rb`

asmega (approved):

PR #1099, opened 2022-12-16
Refactor details known to reuse pages and questions

kosiakkatrina:


app/models/form/sales/pages/person_known.rb
```diff
@@ -0,0 +1,22 @@
...
+  end
+
+  def questions
+    @questions ||= [
+      Form::Sales::Questions::PersonKnown.new(field_for_person("details_known_"), nil, self, person_index: @person_index),
+    ]
+  end
+
+  def page_depends_on
+    return (@person_index..4).map { |index| { "hholdcount" => index } } if @person_index == 1
```
natdeanlewissoftwire, line 18: should this be for `person_display_number` not `@person_index`?

kosiakkatrina: Hm, I think person_index. Cause the index starts at 1 (and then 2, 3, 4). So it is the same as household count.
I don't think we use the `person_display_number` for `details_known` fields at all, because they would always be from 1 to 4, no matter if it's a joint purchase or not.

natdeanlewissoftwire: slightly odd that in other cases `@person_index` is 2-5 or 3-6 depending on if it's a joint purchase but here it's 1-4, not sure if there's a clearer way of doing this?

natdeanlewissoftwire (approved):

PR #1102, opened 2022-12-19
Cldc 855 savings validation

kosiakkatrina:
Add soft validations for savings (Must be less than £100000)
And Deposit (Must be less than 4/3 * Savings)

### DIFF
--- a/app/models/form/sales/questions/deposit_value_check.rb
+++ b/app/models/form/sales/questions/deposit_value_check.rb
@@ -1,0 +1,25 @@
+class Form::Sales::Questions::DepositValueCheck < ::Form::Question
+  def initialize(id, hsh, page)
+    super
+    @id = "deposit_value_check"
+    @check_answer_label = "Deposit confirmation"
+    @header = "Are you sure that the deposit is this much higher than the buyer's savings?"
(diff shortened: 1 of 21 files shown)


app/models/form/sales/questions/deposit_value_check.rb
```diff
@@ -0,0 +1,25 @@
+class Form::Sales::Questions::DepositValueCheck < ::Form::Question
+  def initialize(id, hsh, page)
+    super
+    @id = "deposit_value_check"
+    @check_answer_label = "Deposit confirmation"
+    @header = "Are you sure that the deposit is this much higher than the buyer's savings?"
```
natdeanlewissoftwire, line 6: "this much higher" seems weird phrasing to me - shouldn't we be showing the user how much higher it is?

natdeanlewissoftwire (approved):

app/models/form/sales/questions/deposit_value_check.rb
kosiakkatrina: Yup! I agree, I think all of the validation messages will be rewritten at some point and this is just a placeholder

PR #1103, opened 2022-12-19
CLDC-1629-update-card-titles

natdeanlewissoftwire:
this branch includes changes from refactor-details-known as that was relevant to this ticket.

tested locally
minimal risk
no docs required

ticket: https://digital.dclg.gov.uk/jira/browse/CLDC-1629

image


app/components/check_answers_summary_list_card_component.rb
```diff
@@ -15,4 +15,26 @@ def applicable_questions
...
+        "Lead tenant"
+      when 2..8
+        "Person #{question.check_answers_card_number}"
+      end
+    elsif question.form.type == "sales"
+      case question.check_answers_card_number
+      when 1..number_of_buyers
+          "Buyer #{question.check_answers_card_number}"
+      when (number_of_buyers + 1)..(number_of_buyers + 4)
+          "Person #{question.check_answers_card_number - number_of_buyers}"
```
natdeanlewissoftwire: using `person_display_name` caused issues for the `person_known` questions as the `person_index` follows a different pattern (1-4) rather than 2-5 or 3-6 depending on jointpur. This was the simplest solution to that

app/components/check_answers_summary_list_card_component.rb
```diff
@@ -15,4 +15,26 @@ def applicable_questions
...
+      end
+    end
+  end
+
+  def number_of_buyers
```
natdeanlewissoftwire, line 39: should this be private? or does that not matter

ghost: if it's not used outside of this class, then private would be best

ghost (approved):
Great work!

PR #1113, opened 2022-12-21
Cldc 1545 about the deposit

natdeanlewissoftwire:
image
 ticket: https://digital.dclg.gov.uk/jira/browse/CLDC-1545


app/models/form/sales/subsections/outright_sale.rb
```diff
@@ -8,7 +8,9 @@ def initialize(id, hsh, section)
   end
 
   def pages
-    @pages ||= []
+    @pages ||= [
+      Form::Sales::Pages::AboutDeposit.new(nil, nil, self),
+    ]
```
natdeanlewissoftwire, line 13: redirects to page not found

spec/models/form/sales/pages/about_deposit_spec.rb
```diff
@@ -0,0 +1,33 @@
+require "rails_helper"
+
+RSpec.describe Form::Sales::Pages::AboutDeposit, type: :model do
+  subject(:page) { described_class.new(page_id, page_definition, subsection) }
+
+  let(:page_id) { nil }
```
kosiakkatrina, line 6: Maybe worth testing that id gets set if we pass it in?

kosiakkatrina (approved):
Just one comment, looks good otherwise!

PR #1120, opened 2022-12-22
feat: add question page and subsection

natdeanlewissoftwire:
image

image


app/models/form/sales/sections/household.rb
```diff
@@ -7,6 +7,7 @@ def initialize(id, hsh, form)
     @form = form
     @subsections = [
       Form::Sales::Subsections::HouseholdCharacteristics.new(nil, nil, self),
+      Form::Sales::Subsections::HouseholdSituation.new(nil, nil, self),
```
natdeanlewissoftwire, line 10: status tag isn't updating correctly, tests need to be added for this and the question and page

spec/models/form/sales/subsections/household_needs_spec copy.rb
```diff
@@ -0,0 +1,34 @@
+require "rails_helper"
```
natdeanlewissoftwire, line 1: added this file in even though it's not technically related to this ticket as it was efficient to do it while I'm making another similar spec

app/models/form/subsection.rb
```diff
@@ -30,7 +30,7 @@ def status(log)
 
     qs = applicable_questions(log)
     qs_optional_removed = qs.reject { |q| log.optional_fields.include?(q.id) }
-    return :not_started if qs.count.positive? && qs.all? { |question| log[question.id].blank? || question.read_only? || question.derived? }
+    return :not_started if qs.count.positive? && qs.all? { |question| !question.completed?(log) || question.read_only? || question.derived? }
```
natdeanlewissoftwire, line 33: this means the status tag behaviour works correctly for the subsection, as the question id is `buyers_organisations` but actually we want to know if any of the 4 `preg...` ids (one for each checkbox) are yes in the log.

app/models/form/subsection.rb
```diff
@@ -30,7 +30,7 @@ def status(log)
 
     qs = applicable_questions(log)
     qs_optional_removed = qs.reject { |q| log.optional_fields.include?(q.id) }
-    return :not_started if qs.count.positive? && qs.all? { |question| log[question.id].blank? || question.read_only? || question.derived? }
+    return :not_started if qs.count.positive? && qs.all? { |question| question.unanswered?(log) || question.read_only? || question.derived? }
```
natdeanlewissoftwire, line 33: this means the status tag behaviour works correctly for the subsection, as the question id is `buyers_organisations` but actually we want to know if any of the 4 `preg...` ids (one for each checkbox) are yes in the log.

spec/models/form/sales/subsections/household_needs_spec.rb
```diff
@@ -0,0 +1,34 @@
+require "rails_helper"
```
natdeanlewissoftwire, line 1: added this file as it's easy to add in here and should already exist even though it's technically not part of this ticket

app/models/form/subsection.rb
kosiakkatrina: It seems to be working correct for the lettings logs checkbox questions. Could we make it work the same way?

natdeanlewissoftwire: Ah - I think it only looks like it's working for lettings logs because there are other questions in those subsections - `:not-started` won't be triggered incorrectly in that case because it won't pass the qs.all? (even though the log[question.id] will be incorrectly blank when individual checkbox ids are filled in)

natdeanlewissoftwire: There's an argument that it doesn't matter because we're always going to be adding other questions to the section but I think it's still good to fix it (can't think of a case it matters but there may be one)

kosiakkatrina: Okay, that's fine. I was worried we're looping a lot, but we don't really have that many checkbox questions. It is actually a nice change, well done!

kosiakkatrina (approved):

PR #1122, opened 2022-12-23
[CLDC-857] Add household wheelchair check

ghost:
https://digital.dclg.gov.uk/jira/browse/CLDC-857

### DIFF
--- a/app/models/form/sales/questions/household_wheelchair_check.rb
+++ b/app/models/form/sales/questions/household_wheelchair_check.rb
@@ -10,7 +10,16 @@ class Form::Sales::Questions::HouseholdWheelchairCheck < ::Form::Question
       "0" => { "value" => "Yes" },
       "1" => { "value" => "No" },
     }
-    @hidden_in_check_answers = true
+    @hidden_in_check_answers = {
+      "depends_on" => [
+        {
+          "wheel_value_check" => 0,
+        },
+        {
+          "wheel_value_check" => 1,
+        },
+      ],
+    }
     @page = page
   end
 end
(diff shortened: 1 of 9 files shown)


app/models/validations/sales/soft_validations.rb
```diff
@@ -20,4 +20,10 @@ def mortgage_over_soft_max?
     income_used_for_mortgage = (income1_used_for_mortgage? ? income1 : 0) + (income2_used_for_mortgage? ? income2 : 0)
     mortgage > income_used_for_mortgage * 5
   end
+
+  def wheelchair_when_not_disabled?
+    return false unless disabled == 2
+
+    wheel == 1
```
natdeanlewissoftwire, line 27: what's `wheel` used for?

natdeanlewissoftwire: oh I see it's the previously given wheelchair value, nvm :D

natdeanlewissoftwire (approved):
failing test but lgtm otherwise :)

app/models/form/sales/questions/household_wheelchair_check.rb
```diff
@@ -0,0 +1,16 @@
...
+    @id = "wheel_value_check"
+    @check_answer_label = "Does anyone in the household use a wheelchair?"
+    @header = "Are you sure? You said previously that somebody in household uses a wheelchair"
+    @type = "interruption_screen"
+    @page = page
+    @answer_options = {
+      "0" => { "value" => "Yes" },
+      "1" => { "value" => "No" },
+    }
+    @hidden_in_check_answers = true
```
kosiakkatrina: We only want it to be hidden in check answers if it's answered, otherwise if it's skipped, there's no way to get back to it

app/models/validations/sales/soft_validations.rb
ghost: in the question that asks if someone uses a wheelchair

PR #1125, opened 2022-12-23
Cldc 1488 last accomodation

kosiakkatrina:
Add a previous postcode question:
image

image

Reuse some of the lettings logs methods to infer the previous location.


app/models/sales_log.rb
```diff
@@ -111,4 +114,25 @@ def income1_used_for_mortgage?
...
+  end
+
+  def process_postcode(postcode, postcode_known_key, la_inferred_key, la_key)
+    return if postcode.blank?
+
+    self[postcode_known_key] = 0
+    inferred_la = get_inferred_la(postcode)
+    self[la_inferred_key] = inferred_la.present?
+    self[la_key] = inferred_la if inferred_la.present?
+  end
```
kosiakkatrina, line 137: These are the same as the lettings logs, but I didn't reuse that one because `ppcodenk` == 1 == "Yes", which means that postcode is known in the lettings log if it's set to 1.

In sales log 1 == "No" which meant that the postcode isn't known.

I suspect that the lettings log is wrong and will need to be fixed eventually.

natdeanlewissoftwire: Sounds good, seems like something worth making a separate ticket for investigating but I agree with your suspicion 🕵️

natdeanlewissoftwire (approved):

PR #1129, opened 2023-01-03
Cldc 1521 living before purchase

natdeanlewissoftwire:
image

CYA page - should this be "year" when `proplen` = 1 or is it fine to always be "years" as shown:
image

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema[7.0].define(version: 2022_12_22_133600) do
+ActiveRecord::Schema[7.0].define(version: 2023_01_03_094948) do
   # These are extensions that must be enabled in order to support this database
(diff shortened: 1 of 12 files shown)


db/schema.rb
```diff
@@ -415,6 +415,14 @@
...
     t.integer "ecstat5"
     t.integer "ecstat6"
+    t.integer "pregyrha"
+    t.integer "pregla"
+    t.integer "pregghb"
+    t.integer "pregother"
+    t.integer "disabled"
+    t.integer "wheel_value_check"
+    t.integer "hhregres"
+    t.integer "hhregresstill"
```
natdeanlewissoftwire: should these be included?

kosiakkatrina: I don't think so, try dropping and recreating the db, it should update it

natdeanlewissoftwire: Cheers!

kosiakkatrina (commented):
Looks good, just needs a rebase!

PR #1134, opened 2023-01-03
Cldc 1547 leasehold charges

natdeanlewissoftwire:
bug - numeric field doesn't reflect db precision with rounding, should I try to fix here or make a separate ticket as it's the component that's causing it? image


app/models/form/sales/questions/leasehold_charges_known.rb
```diff
@@ -0,0 +1,27 @@
...
+        {
+          "mscharge_known" => 1,
+        },
+      ],
+    }
+  end
+
+  ANSWER_OPTIONS = {
+    "1" => { "value" => "Yes" },
+    "0" => { "value" => "No" },
```
natdeanlewissoftwire, line 25: There's no consistent rule in the codebase for what numbers correspond to No and Yes (normally 0, 1 or 2 in some order), but I think there should be... worth making a ticket for? Or is it just an artifact from how we want to be exporting data?

app/models/sales_log.rb
```diff
@@ -20,6 +20,7 @@ class SalesLog < Log
   validates_with SalesLogValidator
   before_validation :set_derived_fields!
   before_validation :reset_invalidated_dependent_fields!
+  before_validation :process_mscharge_known_changes!, if: :mscharge_known_changed?
```
natdeanlewissoftwire: should we be using a `mscharge_known_inferred` field like we do for `la_inferred`? or is it fine to infer like this?

app/models/form/sales/subsections/shared_ownership_scheme.rb
```diff
@@ -13,14 +13,15 @@ def pages
       Form::Sales::Pages::Staircase.new(nil, nil, self),
       Form::Sales::Pages::AboutStaircase.new(nil, nil, self),
       Form::Sales::Pages::Resale.new(nil, nil, self),
+      Form::Sales::Pages::ExchangeDate.new(nil, nil, self),
```
natdeanlewissoftwire: think this should be here according to its ticket

app/models/form/sales/questions/leasehold_charges_known.rb
kosiakkatrina: Yup, we're saving it the way it's expected to be exported

app/models/sales_log.rb
kosiakkatrina: Do we need to set it to 0? With other fields (for example age fields) we leave it blank/nil

natdeanlewissoftwire: Yep, it's specified in the ticket :)

kosiakkatrina: Could we add this into the `set_derived_fields!` method instead, cause that's where most stuff like this gets set I think?

kosiakkatrina (approved):
Looks good, just 1 comment 👌

PR #1137, opened 2023-01-04
Validate radio values are possible

asmega:
# Context

- We currently do not validate radio buttons for options that are not actually possible
- e.g only `1` or `2` is possible put could actually set to `100` as there is no validation to prevent against this
- The reason for this change is so we will want to bubble these up for CSV uploads to prevent unpermitted values from seeping thru and return relevant error messages

# Changes

- Added validation for radio buttons so users cannot select options that are not possible. e.g. via HTML injection

# Notes for review

- These changes are only for radio buttons
- We may also need these changes for checkboxes which we can address next outside the scope of this PR
- This may impact imports. As this essentially tightens validations it may be that previously that may have been able to get thru are now halt

# Screenshots

Screenshot 2023-01-04 at 14 46 59


spec/fixtures/exports/general_needs_log.xml
```diff
@@ -54,7 +54,7 @@
     <waityear>1</waityear>
     <postcode_full>NW1 5TY</postcode_full>
     <reasonpref>1</reasonpref>
-    <cbl>1</cbl>
+    <cbl>2</cbl>
```
kosiakkatrina, line 57: cbl is a checkbox in real form, so it should only be 1 or 0. Is this not using the real form?

spec/fixtures/files/lettings_logs_download.csv
```diff
@@ -1,2 +1,2 @@
...
+{id},in_progress,2022-02-08 16:52:15 +0000,2022-02-08 16:52:15 +0000,Danny Rojas,No,DLUHC,DLUHC,2021,Supported housing,,2 October 2021,London Affordable Rent,,,,,,,,,,,,,,,,,,,,No,,,,,No,Westminster,E09000033,,SE1 1TE,,No,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,2,8,0,0,0,,0,,,,,,,,,,,,,,,,,,,,,,,,0,,,,,,,0,,,,,,,,,,,,,,,,,,,,9,1,,,,,,,,6,{scheme_code},{scheme_service_name},{scheme_sensitive},Missing,No,DLUHC,DLUHC,{scheme_primary_client_group},,{scheme_secondary_client_group},{scheme_support_type},{scheme_intended_stay},2021-04-01 00:00:00 +0100,{location_code},SE1 1TE,Downing Street,20,Bungalow,Fitted with equipment and adaptations,Westminster,{location_startdate}
```
kosiakkatrina, line 2: Do we know what this `No` is a value for and why it changed?

spec/fixtures/exports/general_needs_log.xml
asmega: i see. i wasn't aware. it'll be because i changed the factory which has global implications. let me change that back and take a look

asmega: there seems to be explicit code to set to `2`. see https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/blob/main/app/services/exports/lettings_log_export_service.rb#L186

i'm guessing this is what the consumer wants?

spec/fixtures/files/lettings_logs_download.csv
asmega: it'll be to do with https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/1137/files#diff-bb2cb9148b0c268bac3f33ed9a23fc1b4aaa190477e1e1c1fa679ac508ea3b10R573

I'm guessing we added to `wchair` a possible value of `No` which was previously keyed differently. as a result when it does the look up for the human version as it is now actually present, it uses the human version.

spec/fixtures/exports/general_needs_log.xml
kosiakkatrina: Oh, okay, you're right. That's odd. It should only really be true or false (set to 1 if true and 0 or nothing if false). I'd expect it to be set to 0 if it's 0 in the import 🤔

spec/fixtures/files/lettings_logs_download.csv
kosiakkatrina: Okay, that makes sense!

spec/fixtures/exports/general_needs_log.xml
kosiakkatrina: Don't listen to me, the checkboxes are 0-1 in our domain, but looking at the data dictionary it seems like they need to be exported as 1-2. This looks fine

app/models/validations/shared_validations.rb
```diff
@@ -68,4 +68,15 @@ def inactive_status(date, resource)
 
     { scope: status, date: date&.to_formatted_s(:govuk_date), deactivation_date: closest_reactivation&.deactivation_date&.to_formatted_s(:govuk_date) }
   end
+
+  def validate_valid_radio_option(record)
```
jamesrose, line 72: @jamesrose To test this

jamesrose, line 72: Unfortunately we need to resolve these issues before we can move forward with this PR: https://communities-govuk.slack.com/archives/C01QGR78H9S/p1673357370076599

jamesrose: The good news is that they are straight forward data migrations.

jamesrose (approved):

PR #1140, opened 2023-01-04
CLDC-1723 Overhaul letting log owning & managing org questions & tests

SamSeed-Softwire:
## The context
- This ticket https://digital.dclg.gov.uk/jira/browse/CLDC-1723 is basically about making sure that if a log has an owning organisation assigned to it, and then the organisation relationship which makes that owning organisation a valid answer is deleted, then the owning org remain assigned to the log and that it is displayed correctly in the check your answers page. Similarly for managing org.
- In reality this PR goes beyond that, as it fixes other bugs found with these questions as well as adding missing tests where appropriate.

## The changes
- For the original ticket https://digital.dclg.gov.uk/jira/browse/CLDC-1723:
  - Now explicitly passing `user` and `log` to various methods to ensure they're available where required. (This was only required for the stock owner and managing organisation question classes, but was rolled out to equivalent questions as well to ensure method signatures were consistent.) This helped fix the issue where deleting an org relationship and then refreshing the log CYA page led to users seeing "You didn't answer this question" for the owning/managing organisation questions, even if the data still existed in the database.
  - If a stock owner is assigned as a log's owning organisation, then that stock owner now remains selectable even if the relationship which made it a valid option is deleted (e.g. say the stock owner contracted out management to the user's organisation, but then this agreement ended so the relationship was deleted). This means if a user goes back to this question, they are able to see the currently selected stock owner (i.e. what's in the database) in the list of options. However if they then change the stock owner to a be different organisation, the original stock owner is then no longer selectable as the log's owning organisation. Basically the change ensures that what a log currently has saved as its owning organisation in the database is selectable in the list of allowed owning organisations in the front end. I know this is a potentially contentious choice - happy to discuss!
  - An equivalent change was made for a log's managing organisation.
  - Updated and added tests to account for the above changes.
- Changes not part of the original ticket:
  - Previously when signed in as a coordinator, when creating a log the owning_organisation for the log immediately got set as the user’s organisation. This happened even if the organisation doesn’t own stock. The change I've made means that if the user's org doesn't own stock, then it doesn't get automatically set as the log's owning org. There were no tests for this logic even before making the change, so I added tests in `spec/requests/lettings_logs_controller_spec.rb`.
  - Tests that the stock owner and managing org questions in the log setup section were correctly shown/hidden based on various conditions were either missing or could be improved, so I overhauled this (`spec/features/lettings_log_spec.rb`).


app/controllers/logs_controller.rb
```diff
@@ -60,8 +60,9 @@ def api_log_params
   end
 
   def org_params
+    owning_organisation_id = current_user.organisation.holds_own_stock? ? current_user.organisation.id : nil
     {
-      "owning_organisation_id" => current_user.organisation.id,
+      "owning_organisation_id" => owning_organisation_id,
```
SamSeed-Softwire, line 65: This change was because, as I understand it, we shouldn't set a log's owning organisation as an organisation which doesn't hold stock.

app/models/form/lettings/questions/managing_organisation.rb
```diff
@@ -12,10 +12,15 @@ def initialize(id, hsh, page)
...
     opts = { "" => "Select an option" }
+
     return opts unless ActiveRecord::Base.connected?
     return opts unless current_user
     return opts unless log
 
+    if log.managing_organisation.present?
+      opts = opts.merge({ log.managing_organisation.id => log.managing_organisation.name })
+    end
+
```
SamSeed-Softwire, line 23: Whatever is currently in the database, we want to make sure is in the list of answer options.

app/models/form/lettings/questions/managing_organisation.rb
```diff
@@ -34,7 +39,10 @@ def displayed_answer_options(log, user)
     answer_options
   end
 
-  def label_from_value(value)
+  def label_from_value(value, log = nil, user = nil)
+    @log = log
+    @current_user = user
```
SamSeed-Softwire, line 44: This part was necessary to ensure that the log setup CYA page didn't show "You didn't answer this question" for the managing org question (despite the data existing in the db) if you deleted the org relationship and then refreshed the log setup CYA page.

app/models/form/lettings/questions/stock_owner.rb
```diff
@@ -11,8 +11,14 @@ def initialize(id, hsh, page)
...
   def answer_options
     answer_opts = { "" => "Select an option" }
+
     return answer_opts unless ActiveRecord::Base.connected?
     return answer_opts unless current_user
+    return answer_opts unless log
+
+    if log.owning_organisation_id.present?
+      answer_opts = answer_opts.merge({ log.owning_organisation.id => log.owning_organisation.name })
+    end
```
SamSeed-Softwire, line 21: Whatever is currently in the database, we want to make sure is in the list of answer options.

app/models/form/lettings/questions/stock_owner.rb
```diff
@@ -28,7 +34,10 @@ def displayed_answer_options(log, user = nil)
     answer_options
   end
 
-  def label_from_value(value)
+  def label_from_value(value, log = nil, user = nil)
+    @log = log
+    @current_user = user
+
```
SamSeed-Softwire, line 40: This part was necessary to ensure that the log setup CYA page didn't show "You didn't answer this question" for the stock owner question (despite the data existing in the db) if you deleted the org relationship and then refreshed the log setup CYA page.

db/seeds.rb
```diff
@@ -120,7 +120,7 @@
   end
 
   User.find_or_create_by!(
-    name: "Coordinator Owns Stock",
+    name: "Coordinator No Stock",
     email: "coordinator.nostock1@example.com",
```
SamSeed-Softwire: QQ update email too?

db/seeds.rb
```diff
@@ -110,7 +110,7 @@
   )
 
   User.find_or_create_by!(
-    name: "Provider Owns Stock",
+    name: "Provider No Stock",
     email: "provider.nostock1@example.com",
```
SamSeed-Softwire: QQ update email too?

spec/features/lettings_log_spec.rb
```diff
@@ -110,15 +110,55 @@
...
     context "when completing the setup log section" do
-      it "does not include the organisation and created by questions" do
-        visit("/lettings-logs")
-        click_button("Create a new lettings log")
-        click_link("Set up this lettings log")
-        log_id = page.current_path.scan(/\d/).join
-        expect(page).to have_current_path("/lettings-logs/#{log_id}/needs-type")
-        visit("lettings-logs/#{log_id}/setup/check-answers")
-        expect(page).not_to have_content("Owning organisation #{user.organisation.name}")
-        expect(page).not_to have_content("Log owner #{user.name}")
```
SamSeed-Softwire, line 121: Not showing the owning organisation and created by questions does not apply in all cases - it depends on how many potential owning organisations there are

spec/features/lettings_log_spec.rb
```diff
@@ -110,15 +110,55 @@
...
+          visit("lettings-logs/#{log_id}/setup/check-answers")
+          expect(page).to have_content("Stock owner User org")
+        end
+
+        context "and there are more than 2 potential stock owners" do
+          let!(:owning_org2) { create(:organisation, name: "Owning org 2") }
+          let!(:org_rel2) { create(:organisation_relationship, child_organisation: user.organisation, parent_organisation: owning_org2) }
+
+          context "and the organisation relationship for the selected stock owner is deleted" do
+            it "doesn't change the CYA page text to be 'You didn't answer this question'" do
```
SamSeed-Softwire, line 248: This test is for the bug where deleting the organisation relationship then refreshing the log setup CYA page would show "You didn't answer this question" for the owning organisation question, even if the question had been answered.

spec/features/lettings_log_spec.rb
```diff
@@ -110,15 +110,55 @@
...
+              log_id = page.current_path.scan(/\d/).join
+              expect(page).to have_current_path("/lettings-logs/#{log_id}/stock-owner")
+              select(owning_org1.name, from: "lettings-log-owning-organisation-id-field")
+              click_button("Save and continue")
+              visit("lettings-logs/#{log_id}/setup/check-answers")
+              expect(page).to have_content("Stock owner Owning org 1")
+              org_rel1.destroy!
+              visit("lettings-logs/#{log_id}/setup/check-answers")
+              expect(page).to have_content("Stock owner Owning org 1")
+              expect(user.organisation.stock_owners).to eq([org_rel2.parent_organisation])
```
SamSeed-Softwire, line 261: This line is a way of artificially including `org_rel2` in the test - otherwise there will be a lint offence from defining something with `let!` which then doesn't get explicitly used in the test. Although not _explicitly_ used in the test, it's existence is essential because it influences the answer to the stock owner question (if there is only one possible answer for that question then the answer gets auto-selected, whereas if there are multiple possible answers then none get auto-selected).

spec/requests/lettings_logs_controller_spec.rb
```diff
@@ -143,6 +143,61 @@
         expect(whodunnit_actor).to be_a(User)
         expect(whodunnit_actor.id).to eq(user.id)
       end
+
```
SamSeed-Softwire, line 146: These tests are about making sure the log gets the correct managing and owning orgs correctly auto-assigned (or not auto-assigned where appropriate).

spec/features/lettings_log_spec.rb
```diff
@@ -88,14 +88,115 @@
...
+          click_link("Set up this lettings log")
+          log_id = page.current_path.scan(/\d/).join
+          select(support_user.organisation.name, from: "lettings-log-owning-organisation-id-field")
+          click_button("Save and continue")
+          expect(page).to have_current_path("/lettings-logs/#{log_id}/managing-organisation")
+          select(managing_org.name, from: "lettings-log-managing-organisation-id-field")
+          click_button("Save and continue")
+          visit("lettings-logs/#{log_id}/setup/check-answers")
+          expect(page).to have_content("Managing agent Managing org")
+          expect(support_user.organisation.managing_agents).to eq(org_rel.child_organisation)
```
SamSeed-Softwire: This line is pretty artificial/contrived way of explicitly referencing the organisation relationship in the test. Without it, Rubocop complains that something being defined with `let!` is not being explicitly used in the test. The reason I have defined the organisation relationship at all is because its existence influences the logic around the answer options for the question, and in particular whether the question gets auto-assigned an answer.

spec/features/lettings_log_spec.rb
```diff
@@ -88,14 +88,115 @@
...
+                click_button("Save and continue")
+                expect(page).to have_current_path("/lettings-logs/#{log_id}/managing-organisation")
+                select(managing_org1.name, from: "lettings-log-managing-organisation-id-field")
+                click_button("Save and continue")
+                visit("lettings-logs/#{log_id}/setup/check-answers")
+                expect(page).to have_content("Managing agent Managing org 1")
+                org_rel1.destroy!
+                visit("lettings-logs/#{log_id}/setup/check-answers")
+                expect(page).to have_content("Managing agent Managing org 1")
+                expect(support_user.organisation.managing_agents).to eq(org_rel2.child_organisation)
```
SamSeed-Softwire: Similarly to above, this line is pretty artificial/contrived way of explicitly referencing the organisation relationship in the test. Without it, Rubocop complains that something being defined with `let!` is not being explicitly used in the test. The reason I have defined the organisation relationship at all is because its existence influences the logic around the answer options for the question, and in particular whether the question gets auto-assigned an answer.

spec/features/lettings_log_spec.rb
```diff
@@ -110,15 +211,115 @@
...
+              log_id = page.current_path.scan(/\d/).join
+              expect(page).to have_current_path("/lettings-logs/#{log_id}/stock-owner")
+              select(owning_org1.name, from: "lettings-log-owning-organisation-id-field")
+              click_button("Save and continue")
+              visit("lettings-logs/#{log_id}/setup/check-answers")
+              expect(page).to have_content("Stock owner Owning org 1")
+              org_rel1.destroy!
+              visit("lettings-logs/#{log_id}/setup/check-answers")
+              expect(page).to have_content("Stock owner Owning org 1")
+              expect(user.organisation.stock_owners).to eq(org_rel2.parent_organisation)
```
SamSeed-Softwire: Similarly to above, this line is pretty artificial/contrived way of explicitly referencing the organisation relationship in the test. Without it, Rubocop complains that something being defined with `let!` is not being explicitly used in the test. The reason I have defined the organisation relationship at all is because its existence influences the logic around the answer options for the question, and in particular whether the question gets auto-assigned an answer.

spec/features/lettings_log_spec.rb
```diff
@@ -110,15 +211,115 @@
...
+            log_id = page.current_path.scan(/\d/).join
+            if page.current_path == "/lettings-logs/#{log_id}/owning-organisation"
+              click_link("Skip for now")
+            end
+            expect(page).to have_current_path("/lettings-logs/#{log_id}/managing-organisation")
+            select(managing_org.name, from: "lettings-log-managing-organisation-id-field")
+            click_button("Save and continue")
+            visit("lettings-logs/#{log_id}/setup/check-answers")
+            expect(page).to have_content("Managing agent Managing org")
+            expect(user.organisation.managing_agents).to eq(org_rel.child_organisation)
```
SamSeed-Softwire: Similarly to above, this line is pretty artificial/contrived way of explicitly referencing the organisation relationship in the test. Without it, Rubocop complains that something being defined with `let!` is not being explicitly used in the test. The reason I have defined the organisation relationship at all is because its existence influences the logic around the answer options for the question, and in particular whether the question gets auto-assigned an answer.

jamesrose (approved):

app/models/form/lettings/questions/stock_owner.rb
asmega: although not introduced by this PR. line 2 `attr_accessor :current_user, :log` should be removed and those passed into the method calls. otherwise this singleton will suffer from side effects. would it be an easy change for line 12 from
```
def answer_options
```
to
```
def answer_options(log, current_user)
```
or something to that effect?

app/models/form/lettings/questions/managing_organisation.rb
asmega: same thing see https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/1140/files#r1072052592

so `answer_options` could become `answer_options(log, current_user)`?

app/components/check_answers_summary_list_card_component.rb
```diff
@@ -13,7 +13,7 @@ def applicable_questions
   end
 
   def get_answer_label(question)
-    question.answer_label(log).presence || "<span class=\"app-!-colour-muted\">You didn’t answer this question</span>".html_safe
+    question.answer_label(log, user).presence || "<span class=\"app-!-colour-muted\">You didn’t answer this question</span>".html_safe
```
kosiakkatrina, line 16: Is this meant to be @user?

app/helpers/check_answers_helper.rb
```diff
@@ -43,6 +43,6 @@ def total_applicable_questions(subsection, lettings_log, current_user)
   end
 
   def get_answer_label(question, lettings_log)
-    question.answer_label(lettings_log).presence || "<span class=\"app-!-colour-muted\">You didn’t answer this question</span>".html_safe
+    question.answer_label(lettings_log, current_user).presence || "<span class=\"app-!-colour-muted\">You didn’t answer this question</span>".html_safe
```
kosiakkatrina, line 46: Where is currently_user coming from here?

app/components/check_answers_summary_list_card_component.rb
asmega: looking at the code the result should be the same. its just going thru the getter

app/helpers/check_answers_helper.rb
asmega: my guess is it's the devise controller method, which is also a helper method. this module is mixed in so is likely in scope

app/models/form/lettings/questions/managing_organisation.rb
asmega: to be addressed later

app/models/form/lettings/questions/stock_owner.rb
asmega: to be addressed later

asmega (approved):

app/components/check_answers_summary_list_card_component.rb
SamSeed-Softwire: I won't change this then unless it's better practice to do so - @kosiakkatrina @asmega any thoughts on that?

Also presumably if we use `@user` instead of `user` we should also use `@log` instead of `log`?

app/helpers/check_answers_helper.rb
SamSeed-Softwire: Given this seems to be working fine and `current_user` was already being used in many places in this file, I think this is fine.

PR #1144, opened 2023-01-05
feat: PO fix and tests

natdeanlewissoftwire:
responds to: https://digital.dclg.gov.uk/jira/browse/CLDC-1545

### DIFF
--- a/app/models/form/sales/pages/about_deposit_without_discount.rb
+++ b/app/models/form/sales/pages/about_deposit_without_discount.rb
@@ -5,7 +5,6 @@ class Form::Sales::Pages::AboutDepositWithoutDiscount < ::Form::Page
     @description = ""
     @subsection = subsection
     @depends_on = [{ "is_type_discount?" => false }]
-
   end
 
   def questions
--- a/app/models/form/sales/pages/about_deposit_without_discount.rb
+++ b/app/models/form/sales/pages/about_deposit_without_discount.rb
@@ -1,0 +1,31 @@
+class Form::Sales::Pages::AboutDepositWithoutDiscount < ::Form::Page
+  def initialize(id, hsh, subsection)
+    super
+    @header = "About the deposit"
+    @description = ""
+    @subsection = subsection
+    @depends_on = [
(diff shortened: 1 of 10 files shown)


ghost (approved):

app/models/form/sales/pages/about_deposit_without_discount.rb
```diff
@@ -0,0 +1,31 @@
...
+      { "type" => 30 },
+      { "type" => 8 },
+      { "type" => 14 },
+      { "type" => 27 },
+      { "type" => 9 },
+      { "type" => 29 },
+      { "type" => 21 },
+      { "type" => 22 },
+      { "type" => 10 },
+      { "type" => 12 },
```
kosiakkatrina: Is it possible for this to be a method on the sales_log model?
```
@depends_on = [{is_type_discount? => false}]
```

and in the model 
```
def is_type_discount?
    type == 18
end
```
And could use it in the other question as ` [{is_type_discount? => true}]`

kosiakkatrina: Idk if there's a better name for this

natdeanlewissoftwire: I don't think that's possible - I think depends_on has to be an array with a db field and value otherwise it won't get parsed correctly

kosiakkatrina: Hmm, I thought we were doing methods on interruption screens, for example:
```
"depends_on": [
                {
                  "female_in_pregnant_household_in_soft_validation_range?": true
                }
              ],
```

kosiakkatrina: Happy for you to merge as is tho

natdeanlewissoftwire: hm ok I'll have another go, thanks! Will merge either way

natdeanlewissoftwire: Got it, cheers!

PR #1145, opened 2023-01-05
[CLDC-1470] Add wheelchair accessible question

ghost:
https://digital.dclg.gov.uk/jira/browse/CLDC-1470


db/schema.rb
```diff
@@ -416,6 +416,22 @@
     t.integer "ecstat5"
     t.integer "ecstat6"
     t.integer "disabled"
+    t.integer "wheel_value_check"
```
ghost: to fix

spec/models/form/sales/pages/property_local_authority_spec.rb
```diff
@@ -0,0 +1,34 @@
+require "rails_helper"
```
kosiakkatrina, line 1: Hm, maybe needs a rebase onto the main? Cause I think this change is on a different branch as well as `spec/models/form/sales/questions/property_local_authority_spec.rb`

ghost: I just discovered some files without tests and added them, if they are on another branch I can delete them from mine

kosiakkatrina (approved):

PR #1148, opened 2023-01-05
CLDC-1779 Bulk upload lettings validation

asmega:
# Context

- https://digital.dclg.gov.uk/jira/browse/CLDC-1779
- We need to add validation rules for CSV bulk upload for lettings
- The majority of the validation rules already exist and on `Log` and we simply bubble there errors upwards for the CSV validation

# Changes

- Added methods for developer happiness to easily toggle between fake and real forms
- Instantiate a log for each CSV row and validate according to existing validation rules and pass any errors back along to CSV validation so they can be bubbled back to the user
- Add mechanism to be able to add validation rules solely for the CSV upload
- Add extra validation rules for CSV validation to handle certain edge cases that are CSV specific

# Missing things

- I've cut a PR so we can work on things further downstream but there a few gaps that will need patching
- field `111` `Which organisation owns this property?` we need extra validation to ensure these are associated to only permitted orgs
- same as above for field `113` `Which organisation manages this property?`
- field `112` `Username field` yet to be decided if this is needed and if so how the lookup is performed
- field `52` `What is the tenant's main reason for the household leaving their last settled home?`. There seems to be a mismatch from the data dictionary so needs to verified before work can be done
- fields `55-60` which are field around disabled access needs. the CSV has them as checkboxes which mismatches with the webform of grouped radios. need to determine what should happen here
- field `4-5` `Management group code` and `Scheme code` we need to validate these are permitted and how to perform the look ups
- field `1` `renttype` need to check how this ties in with question we added to the bulk upload journey and how that data affects this field

# Notes for review

- Depends on https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/1137 which should probably be merged first

# Screenshots

Screenshot 2023-01-12 at 16 55 49


jamesrose (approved):

app/models/validations/shared_validations.rb
```diff
@@ -68,4 +68,15 @@ def inactive_status(date, resource)
 
     { scope: status, date: date&.to_formatted_s(:govuk_date), deactivation_date: closest_reactivation&.deactivation_date&.to_formatted_s(:govuk_date) }
   end
+
+  def validate_valid_radio_option(record)
```
jamesrose: Do you need to rebase so that the feature flag is here as well?

app/services/bulk_upload/lettings/row_parser.rb
```diff
@@ -137,49 +139,633 @@ class BulkUpload::Lettings::RowParser
...
+    mrcdate.present? ? 1 : 0
+  end
+
+  def mrcdate
+    Date.new(field_94 + 2000, field_93, field_92) if field_94.present? && field_93.present? && field_92.present?
+  end
+
+  def prevloc
+    field_62
+  end
```
kosiakkatrina, line 605: does this need a method?

asmega: well it's basically an `extract method` and gives an alias to what `field_62` actually is as it's been used elsewhere. I can remove if you feel strongly about it

app/models/validations/shared_validations.rb
asmega: that's done

PR #1164, opened 2023-01-09
CLDC-1749 Add mortgage lender question

kosiakkatrina:
Add mortgage lender question:
image
image
image

### DIFF
--- a/app/models/form/sales/questions/mortgage_lender.rb
+++ b/app/models/form/sales/questions/mortgage_lender.rb
@@ -1,0 +1,56 @@
+class Form::Sales::Questions::MortgageLender < ::Form::Question
+  def initialize(id, hsh, page)
+    super
+    @id = "mortgagelender"
+    @check_answer_label = "Mortgage Lender"
+    @header = "What is the name of the mortgage lender?"
+    @type = "select"
+    @hint_text = ""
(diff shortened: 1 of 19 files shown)


ghost (approved):

app/models/form/sales/questions/mortgage_lender.rb
```diff
@@ -0,0 +1,58 @@
...
+    "31" =>	"Skipton Building Society",
+    "32" =>	"Teachers Building Society",
+    "33" =>	"The Co-operative Bank",
+    "34" =>	"Tipton & Coseley Building Society",
+    "35" =>	"TSB",
+    "36" =>	"Ulster Bank",
+    "37" =>	"Virgin Money",
+    "38" =>	"West Bromwich Building Society",
+    "39" =>	"Yorkshire Building Society",
+    "40" =>	"Other",
```
asmega, line 56: are these values already defined somewhere? if not i feel `Other` should be `0`

asmega (approved):

app/models/form/sales/questions/mortgage_lender.rb
kosiakkatrina: These are defined :<

PR #1169, opened 2023-01-10
CLDC-1765 Remove managing organisation for sales log from ui and backend

natdeanlewissoftwire:
Sales logs don't have managing orgs, so that text is removed from the log list: 
image

In removing this from the db table, some general log logic that referred to managing orgs had to be separated to be in both the lettings log model (with managing orgs) and the sales log model (without managing orgs)


app/components/log_summary_component.html.erb
```diff
@@ -37,16 +37,18 @@
...
           <dl class="app-metadata">
             <div class="app-metadata__item">
               <dt class="app-metadata__term">Owned by</dt>
               <dd class="app-metadata__definition"><%= log.owning_organisation&.name %></dd>
             </div>
-            <div class="app-metadata__item">
-              <dt class="app-metadata__term">Managed by</dt>
-              <dd class="app-metadata__definition"><%= log.managing_organisation&.name %></dd>
-            </div>
+            <% if log.instance_of?(LettingsLog) && log.managing_organisation %>
```
jamesrose:
```suggestion
            <% if log.lettings? && log.managing_organisation %>
```

app/components/log_summary_component.html.erb
```diff
@@ -37,16 +37,18 @@
       <% end %>
 
       <% if current_user.support? %>
-        <% if log.owning_organisation or log.managing_organisation %>
+        <% if log.owning_organisation %>
```
jamesrose: Does this not hide managed lettings logs from the summary list?

app/services/filter_service.rb
```diff
@@ -17,6 +17,14 @@ def self.filter_logs(logs, search_term, filters, all_orgs, user)
       logs = logs.public_send("filter_by_#{category}", values, user)
     end
     logs = logs.order(created_at: :desc)
-    user.support? ? logs.all.includes(:owning_organisation, :managing_organisation) : logs
+    if user.support?
+      if logs.first.instance_of?(LettingsLog)
```
SamSeed-Softwire: (Assuming that James's [suggestion](https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/1169/files#r1067915639) works.) 
```suggestion
      if logs.first.lettings?
```

spec/components/log_summary_component_spec.rb
```diff
@@ -19,7 +19,7 @@
       expect(result).to have_text("Created 8 February 2022")
       expect(result).to have_text("by Danny Rojas")
       expect(result).to have_content("Owned by\n              DLUHC")
-      expect(result).to have_content("Managed by\n              DLUHC")
+      expect(result).not_to have_content("Managed by\n              DLUHC")
```
SamSeed-Softwire: Would the following be better? It would be good to get a test failure if you see the 'managed by' question pop up with _any_ organisation, rather than just failing for DLUHC only.
```suggestion
      expect(result).not_to have_content("Managed by")
```

spec/models/sales_log_spec.rb
```diff
@@ -80,21 +80,20 @@
...
-      create(:sales_log, :in_progress, owning_organisation: organisation_1, managing_organisation: organisation_1)
-      create(:sales_log, :completed, owning_organisation: organisation_1, managing_organisation: organisation_2)
-      create(:sales_log, :completed, owning_organisation: organisation_2, managing_organisation: organisation_1)
-      create(:sales_log, :completed, owning_organisation: organisation_2, managing_organisation: organisation_2)
+      create(:sales_log, :in_progress, owning_organisation: organisation_1)
+      create(:sales_log, :completed, owning_organisation: organisation_1)
+      create(:sales_log, :completed, owning_organisation: organisation_2)
     end
 
     it "filters by given organisation id" do
```
SamSeed-Softwire: Is it necessary to test that it works with both the org and the org id? AFAICT the fact that it works with both is due to some Rails magic rather than because of how the the scope has been written, so seems odd to test that.

natdeanlewissoftwire: good shout, will remove

app/components/log_summary_component.html.erb
natdeanlewissoftwire: Ah yep, I've tweaked this slightly, thanks

app/services/filter_service.rb
```diff
@@ -17,6 +17,14 @@ def self.filter_logs(logs, search_term, filters, all_orgs, user)
       logs = logs.public_send("filter_by_#{category}", values, user)
     end
     logs = logs.order(created_at: :desc)
-    user.support? ? logs.all.includes(:owning_organisation, :managing_organisation) : logs
+    if user.support?
+      if logs.first.is_a?(LettingsLog)
```
arfacamble: `logs.first.lettings?`

arfacamble (approved):

PR #1170, opened 2023-01-10
CLDC-1796 Remove illness_type_0 from lettings log

ghost:
Remove illness_type_0 answer option from condition_effects question in Lettings Log. 
This option should not exist because it's not in the data dictionary/ service map.

https://digital.dclg.gov.uk/jira/browse/CLDC-1796

Another PR will follow up with rake task to extract the data and another one for removing the column from the database.


natdeanlewissoftwire (approved):
just the test but it looks pretty simple :D

PR #1172, opened 2023-01-11
CLDC-1796 Add rake task to extract logs data

ghost:
Retrieve user and org data for completed logs where illness_type_0 == 1

https://digital.dclg.gov.uk/jira/browse/CLDC-1796


natdeanlewissoftwire (approved):
I don't have a load of experience with rake tasks but as far as I can be confident this lgtm!

PR #1176, opened 2023-01-12
CLDC-1734 Use managing_organisation for rent period validations

kosiakkatrina:
Currently, there is a validation for "How often does the household pay rent and other charges?"
The validation triggers if the user selects a rent period which is not listed in the owning organisation's listed rent periods on the "about your organisation" page.
For organisations in parent/child relationships, the validation should use the managing agents' rent periods.

### DIFF
--- a/app/models/validations/financial_validations.rb
+++ b/app/models/validations/financial_validations.rb
@@ -98,11 +98,11 @@ module Validations::FinancialValidations
   end
 
   def validate_rent_period(record)
-    if record.owning_organisation.present? && record.owning_organisation.rent_periods.present? &&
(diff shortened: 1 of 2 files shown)


ghost (approved):

app/models/validations/financial_validations.rb
```diff
@@ -98,11 +98,11 @@ def validate_rent_amount(record)
   end
 
   def validate_rent_period(record)
-    if record.owning_organisation.present? && record.owning_organisation.rent_periods.present? &&
```
jamesrose, line 101: Is the `managing_organisation` always set?

kosiakkatrina: Not when it's created by the support user

PR #1184, opened 2023-01-13
CLDC-808 Update derived sales variables

kosiakkatrina:
We had some derived household variables missing, this PR adds them in.
We don't need to collect total elder number separately for sales logs, but we need to be able to check how many elders live in the property to determine household type


spec/models/sales_log_spec.rb
```diff
@@ -127,6 +127,44 @@
       record_from_db = ActiveRecord::Base.connection.execute("select deposit from sales_logs where id=#{sales_log.id}").to_a[0]
       expect(record_from_db["deposit"]).to eq(nil)
     end
+
+    context "when deriving household variables" do
+      let!(:household_lettings_log) do
```
SamSeed-Softwire: I haven't worked on sales yet so this may be a silly question, but why are we calling this object a lettings log when it's an instance of a sales log?

app/models/derived_variables/sales_log_variables.rb
```diff
@@ -15,5 +15,74 @@ def set_derived_fields!
...
+    relationships = [relat2, relat3, relat4, relat5, relat6]
+    relationships.count("C")
+  end
+
+  def total_adult
+    total = !age1.nil? && age1 >= 16 && age1 < 60 ? 1 : 0
+    total + (2..6).count do |i|
+      age = public_send("age#{i}")
+      relat = public_send("relat#{i}")
+      !age.nil? && ((age >= 16 && age < 18 && %w[P X].include?(relat)) || age >= 18 && age < 60)
```
ghost: `!age.nil?` => `age.present?`

app/models/derived_variables/sales_log_variables.rb
```diff
@@ -15,5 +15,74 @@ def set_derived_fields!
...
+    self.totadult = total_adult + total_elder
+    self.hhmemb = totchild + totadult
+    self.hhtype = household_type
+  end
+
+private
+
+  def total_elder
+    ages = [age1, age2, age3, age4, age5, age6]
+    ages.count { |x| !x.nil? && x >= 60 }
```
ghost: `!x.nil?` => `x.present?`

ghost (approved):

app/models/derived_variables/sales_log_variables.rb
SamSeed-Softwire: Could bracket `age >= 18 && age < 60` to make the logic clearer here? That's just personal preference though - I like to be super explicit.

SamSeed-Softwire: Where is the logic for teenagers (`age >= 16 && age < 18 && %w[P X].include?(relat)`) specified? Can't see it in the ticket.

app/models/derived_variables/sales_log_variables.rb
```diff
@@ -15,5 +15,74 @@ def set_derived_fields!
...
+    end
+  end
+
+  def household_type
+    return unless total_elder && total_adult && totchild
+
+    if only_one_elder?
+      1
+    elsif two_adults_including_elders?
+      2
```
SamSeed-Softwire, line 55: I read the table in the ticket to mean household_type would be 2 if and only if there were 2 elderly people in the house. Are we sure we're allowing one of the people to be a non-elderly adult?

app/models/derived_variables/sales_log_variables.rb
```diff
@@ -15,5 +15,74 @@ def set_derived_fields!
...
+    elsif one_adult_with_at_least_one_child?
+      5
+    elsif two_adults_with_at_least_one_child?
+      6
+    else
+      9
+    end
+  end
+
+  def two_adults_with_at_least_one_child?
```
SamSeed-Softwire: Would it be better to call this `at_least_two_adults_with_at_least_one_child`?

spec/models/sales_log_spec.rb
kosiakkatrina: Not a silly question! I've copied the tests from lettings logs and didn't notice this 🙃 
We're deriving same (similar?) variables for both types of logs

app/models/derived_variables/sales_log_variables.rb
kosiakkatrina: Yup, you're right!

app/models/derived_variables/sales_log_variables.rb
kosiakkatrina: Fixed this now, it's different on sales, good spot!

app/models/derived_variables/sales_log_variables.rb
```diff
@@ -16,5 +16,77 @@ def set_derived_fields!
...
+  def total_elder
+    ages = [age1, age2, age3, age4, age5, age6]
+    ages.count { |age| age.present? && age >= 60 }
+  end
+
+  def total_child
+    (2..6).count do |i|
+      age = public_send("age#{i}")
+      relat = public_send("relat#{i}")
+      age.present? && ((age < 20 && %w[C].include?(relat)))
```
SamSeed-Softwire, line 36: @kosiakkatrina in the ticket Jemma wrote the following:

> For tenants that aren’t the lead tenant, we define adult as 18 or more, though if the person has RELAT = C and is less than 20, we define them as a child. 

Is the code above not missing the logic for people who are 16 or 17 but don't have RELAT = C (and who aren't the lead tenant of course).

app/models/derived_variables/sales_log_variables.rb
```diff
@@ -16,5 +16,77 @@ def set_derived_fields!
...
+    end
+  end
+
+  def total_adult
+    total = age1.present? && age1.between?(16, 59) ? 1 : 0
+    total + (2..6).count do |i|
+      age = public_send("age#{i}")
+      relat = public_send("relat#{i}")
+      age.present? && ((age.between?(20, 59) || age.between?(18, 19) && relat != "C"))
```
SamSeed-Softwire, line 45: Are the double brackets necessary? Or was the second open bracket meant to be before `age.between?(18, 19)` rather than before `age.between?(20, 59)` (presumably in order to help make clearer the order of operation)?

app/models/derived_variables/sales_log_variables.rb
kosiakkatrina: Sorry, merged this without seeing the comment, but will look at this now

kosiakkatrina: https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/1195

PR #1188, opened 2023-01-13
CLDC-859 Sales validation - bedsits cannot have more than 1 bed

SamSeed-Softwire:
## Context
- https://digital.dclg.gov.uk/jira/browse/CLDC-859
- Adding a sales validation to ensure bedsits have at most one bedroom.

## The changes
- Added a validation to the property type page, so that if the number of bedrooms is >1 and then bedsit (proptype = 2) is selected as the property type, then the following message is displayed: "Bedsit maximum 1 bedroom".
image

- Added a validation to the 'number of bedrooms' page, so that if the property type is selected (as bedsit) and then the number of beds is entered as >1 then the following message is displayed: "Bedsit bedroom maximum 1".
image

### DIFF
--- a/app/models/sales_log.rb
+++ b/app/models/sales_log.rb
@@ -223,6 +223,12 @@ class SalesLog < Log
     type == 24
   end
 
+  def is_bedsit?
+    return if proptype.nil?
+
+    proptype == 2
+  end
+
   def shared_ownership_scheme?
     ownershipsch == 1
   end
--- a/app/models/sales_log.rb
+++ b/app/models/sales_log.rb
@@ -224,8 +224,6 @@ class SalesLog < Log
   end
 
   def is_bedsit?
-    return if proptype.nil?
-
     proptype == 2
   end
 
--- a/app/models/validations/sales/property_validations.rb
+++ b/app/models/validations/sales/property_validations.rb
@@ -9,9 +9,9 @@ module Validations::Sales::PropertyValidations
   end
 
   def validate_bedsit_number_of_beds(record)
-    return if record.proptype.blank? || record.beds.blank?
+    return if record.is_bedsit?.blank? || record.beds.blank?
 
-    if record.proptype == 2 && record.beds > 1
+    if record.is_bedsit? && record.beds > 1
       record.errors.add :proptype, I18n.t("validations.property.proptype.bedsits_have_max_one_bedroom")
       record.errors.add :beds, I18n.t("validations.property.beds.bedsits_have_max_one_bedroom")
     end
--- a/app/models/validations/sales/property_validations.rb
+++ b/app/models/validations/sales/property_validations.rb
@@ -15,4 +15,12 @@ module Validations::Sales::PropertyValidations
       record.errors.add :proptype, I18n.t("validations.property.proptype.bedsits_have_max_one_bedroom")
     end
   end
+
+  def validate_property_number_of_bedrooms(record)
+    return if record.proptype.blank? || record.beds.blank?
+
+    unless record.proptype != 2 || record.beds <= 1
+      record.errors.add :beds, I18n.t("validations.property.beds.bedsits_have_max_one_bedroom")
+    end
+  end
 end
--- a/app/models/validations/sales/property_validations.rb
+++ b/app/models/validations/sales/property_validations.rb
@@ -8,7 +8,7 @@ module Validations::Sales::PropertyValidations
     end
   end
 
-  def validate_propert_unit_type(record)
+  def validate_property_unit_type(record)
     return if record.proptype.blank? || record.beds.blank?
 
     unless record.proptype != 2 || record.beds <= 1
--- a/app/models/validations/sales/property_validations.rb
+++ b/app/models/validations/sales/property_validations.rb
@@ -8,18 +8,11 @@ module Validations::Sales::PropertyValidations
     end
   end
 
-  def validate_property_unit_type(record)
+  def validate_bedsit_number_of_beds(record)
     return if record.proptype.blank? || record.beds.blank?
 
     unless record.proptype != 2 || record.beds <= 1
       record.errors.add :proptype, I18n.t("validations.property.proptype.bedsits_have_max_one_bedroom")
-    end
-  end
-
-  def validate_property_number_of_bedrooms(record)
-    return if record.proptype.blank? || record.beds.blank?
-
-    unless record.proptype != 2 || record.beds <= 1
       record.errors.add :beds, I18n.t("validations.property.beds.bedsits_have_max_one_bedroom")
     end
   end
--- a/app/models/validations/sales/property_validations.rb
+++ b/app/models/validations/sales/property_validations.rb
@@ -11,7 +11,7 @@ module Validations::Sales::PropertyValidations
   def validate_bedsit_number_of_beds(record)
     return if record.proptype.blank? || record.beds.blank?
 
-    unless record.proptype != 2 || record.beds <= 1
+    if record.proptype == 2 && record.beds > 1
       record.errors.add :proptype, I18n.t("validations.property.proptype.bedsits_have_max_one_bedroom")
       record.errors.add :beds, I18n.t("validations.property.beds.bedsits_have_max_one_bedroom")
     end
--- a/app/models/validations/sales/property_validations.rb
+++ b/app/models/validations/sales/property_validations.rb
@@ -7,4 +7,12 @@ module Validations::Sales::PropertyValidations
       record.errors.add :ppostcode_full, I18n.t("validations.property.postcode.must_match_previous")
     end
   end
+
+  def validate_propert_unit_type(record)
+    return if record.proptype.blank? || record.beds.blank?
+
+    unless record.proptype != 2 || record.beds <= 1
+      record.errors.add :proptype, I18n.t("validations.property.proptype.bedsits_have_max_one_bedroom")
+    end
+  end
 end
--- a/app/models/validations/sales/property_validations.rb
+++ b/app/models/validations/sales/property_validations.rb
@@ -9,7 +9,7 @@ module Validations::Sales::PropertyValidations
   end
 
   def validate_bedsit_number_of_beds(record)
-    return if record.is_bedsit?.blank? || record.beds.blank?
+    return unless record.proptype.present? && record.beds.present?
 
     if record.is_bedsit? && record.beds > 1
       record.errors.add :proptype, I18n.t("validations.property.proptype.bedsits_have_max_one_bedroom")
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -202,9 +202,9 @@ en:
       beds:
         non_positive: "Number of bedrooms has to be greater than 0"
         over_max: "Number of bedrooms cannot be more than 12"
-        bedsits_have_max_one_bedroom: "Bedsits have at most one bedroom"
+        bedsits_have_max_one_bedroom: "Bedsit bedroom maximum 1"
       proptype:
-        bedsits_have_max_one_bedroom: "Properties with 2 or more bedrooms cannot be bedsits"
+        bedsits_have_max_one_bedroom: "Bedsit maximum 1 bedroom"
       postcode:
         must_match_previous: "Buyer's last accommodation and discounted ownership postcodes must match"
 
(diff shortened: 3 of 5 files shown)


spec/models/validations/sales/property_validations_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+RSpec.describe Validations::Sales::PropertyValidations do
+  subject(:property_validator) { validator_class.new }
+
+  let(:validator_class) { Class.new { include Validations::Sales::PropertyValidations } }
+
+  describe "#validate_property_unit_type" do
+    context "when number of bedrooms is <= 1" do
+      let(:record) { FactoryBot.build(:sales_log, beds: 1, proptype: 2) }
+
+      it "does not add an error if it's a bedsit" do
```
SamSeed-Softwire: Is it overkill to test that the error is _not_ added when not a bedsit?

spec/models/validations/sales/property_validations_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+
+      it "does add an error if it's a bedsit" do
+        property_validator.validate_property_unit_type(record)
+
+        expect(record.errors).to be_present
+      end
+    end
+  end
+
+  describe "#validate_property_number_of_bedrooms" do
```
SamSeed-Softwire: Would it be simpler to combine this test and the above (so the following 2 lines would be sequential)?
`property_validator.validate_property_unit_type(record)`
`property_validator.validate_property_number_of_bedrooms(record)`

app/models/validations/sales/property_validations.rb
```diff
@@ -0,0 +1,17 @@
+module Validations::Sales::PropertyValidations
+  def validate_property_unit_type(record)
+    return if record.proptype.blank? || record.beds.blank?
+
+    unless record.proptype != 2 || record.beds <= 1
+      record.errors.add :proptype, I18n.t("validations.property.proptype.bedsits_have_max_one_bedroom")
+    end
+  end
+
+  def validate_property_number_of_bedrooms(record)
```
natdeanlewissoftwire: Can't we combine these into one method?

spec/models/validations/sales/property_validations_spec.rb
natdeanlewissoftwire: I can see both sides, I don't feel strongly enough to break the convention in the code personally but I'd probably still approve this PR without this test 😅

spec/models/validations/sales/property_validations_spec.rb
natdeanlewissoftwire: I think so!

natdeanlewissoftwire (approved):
lgtm!

natdeanlewissoftwire:
(I'm not sure about the answer to your question regarding tests in `spec/features` / `spec/requests`, maybe one of the other devs can advise better there)

app/models/validations/sales/property_validations.rb
```diff
@@ -0,0 +1,17 @@
...
+    return if record.proptype.blank? || record.beds.blank?
+
+    unless record.proptype != 2 || record.beds <= 1
+      record.errors.add :beds, I18n.t("validations.property.beds.bedsits_have_max_one_bedroom")
+    end
+  end
+end
```
arfacamble, line 19: I agree this can be one method.
The condition on line 13 or 5 also feels to me it could be easier to read. Might just be the way I think about bedsits or I might not quite grasp what this line is expressing, but it reads like 'unless it's not a bedsit or unless one bedroom or fewer' is that equivalent to ' if it is a bedsit && beds > 1' that was what I was imagining it would be doing
also on readability I have a soft preference for writing an is_bedsit? method on SalesLog rather than the slightly more obscure .proptype == magic_number

arfacamble: also on readability of my comment I seem to have forgotten about punctuation

spec/models/validations/sales/property_validations_spec.rb
arfacamble: I think it's good to test at least one situation where it is and is not added.

if you wanted to be really comprehensive you could even have tests for one field undefined, eg is_bedsit but beds: nil

kosiakkatrina (approved):

config/locales/en.yml
```diff
@@ -195,6 +195,9 @@ en:
       beds:
         non_positive: "Number of bedrooms has to be greater than 0"
         over_max: "Number of bedrooms cannot be more than 12"
+        bedsits_have_max_one_bedroom: "Bedsit bedroom maximum 1"
+      proptype:
+        bedsits_have_max_one_bedroom: "Bedsit maximum 1 bedroom"
```
jamesrose: Can we improve on these error messages? Neither are sentences!

SamSeed-Softwire: @jamesrose I agree I'm not a fan of these - I originally wrote something more verbose, but then changed them to what they are now as I was told Beth had written them so I should use them. Do you think I should query these with her?

app/models/validations/sales/property_validations.rb
```diff
@@ -7,4 +7,13 @@ def validate_postcodes_match_if_discounted_ownership(record)
       record.errors.add :ppostcode_full, I18n.t("validations.property.postcode.must_match_previous")
     end
   end
+
+  def validate_bedsit_number_of_beds(record)
+    return if record.proptype.blank? || record.beds.blank?
+
+    unless record.proptype != 2 || record.beds <= 1
```
kosiakkatrina: I try to avoid using unless not, it might be easier to read if you change this into an if statement 👀

kosiakkatrina (approved):

app/models/validations/sales/property_validations.rb
SamSeed-Softwire: Thanks @kosiakkatrina! Changed now :)

app/models/validations/sales/property_validations.rb
SamSeed-Softwire: Thanks @arfacamble. I think you must have written these before some recent commits I did - I've already condensed the methods into one and have changed `unless...not` to `if`. Shout if you disagree.

Also nice idea about writing a separate method `is_bedsit?` - have done that now, can you let me know what you think please?

Cheers!

spec/models/validations/sales/property_validations_spec.rb
```diff
@@ -48,4 +48,40 @@
...
+    context "when number of bedrooms is > 1" do
+      let(:record) { FactoryBot.build(:sales_log, beds: 2, proptype: 2) }
+
+      it "does add an error if it's a bedsit" do
+        property_validator.validate_bedsit_number_of_beds(record)
+        expect(record.errors.added?(:proptype, "Bedsit maximum 1 bedroom")).to be true
+        expect(record.errors.added?(:beds, "Bedsit bedroom maximum 1")).to be true
+      end
+
+      it "does not add an error if proptype is undefined" do
```
SamSeed-Softwire, line 71: @arfacamble you wrote a comment about potentially making these tests more comprehensive by checking for when one of the fields is undefined. What do you think about these (this test and the one below)?

app/models/validations/sales/property_validations.rb
arfacamble: floats my boat

spec/models/validations/sales/property_validations_spec.rb
arfacamble: looks great

arfacamble (approved):

app/models/sales_log.rb
```diff
@@ -219,6 +219,12 @@ def old_persons_shared_ownership?
     type == 24
   end
 
+  def is_bedsit?
+    return if proptype.nil?
```
arfacamble: I don't think this is necessary - if proptype is nil, then 225 returns nil == 2 => false anyway

app/models/validations/sales/property_validations.rb
```diff
@@ -7,4 +7,13 @@ def validate_postcodes_match_if_discounted_ownership(record)
       record.errors.add :ppostcode_full, I18n.t("validations.property.postcode.must_match_previous")
     end
   end
+
+  def validate_bedsit_number_of_beds(record)
+    return if record.is_bedsit?.blank? || record.beds.blank?
```
arfacamble: aha. I see related to my commetn above, that it would return nil is proptype is nil and then this would work as a guard clause.
I think it would work better in being consistent with the other validations for this line to be ` return if record.proptype.blank? || ...`

separately this could be `return unless record.proptype && record.beds` which is slightly more concise but they would both be doing the same thing so it's really a subjective readability choice and I don't see a clear convention on the project.

arfacamble (changes requested):
Can see one fairly small thing that wants changing to be consistent with other validations

app/models/validations/sales/property_validations.rb
SamSeed-Softwire: Thanks @arfacamble. It seems odd to me to go to the trouble of writing an `is_bedsit?` method and then use it only on line 14 here and not in the logical evaluation on line 12 - what do you think?

arfacamble: I think it's good to have it for line 14 as it makes that line more readable, even if it's only used in that one place.

I think it's a bit misleading on line 12 because the convention for a method ending in ? is that it returns true or false.
Due to the thruthy falsey nature of ruby, I think it's just about fine for that method to return nil when it means false, but I don't think that should be intended to be used as a 'nil' because then it goes against that contravention.

On that line 12 I'm kind of expecting to be calling `false.blank?` or `true.blank?` which again will work but is on the limits of having a bit of a code smell. I have to think harder to work out what will happen in each situation. In the end it boils down to returning if proptype is not bedsit, si I think it would be better if you want to use it, to have !record.is_bedsit? but that would be caught in the if lower down anyway.

kosiakkatrina: Could it be `if record.is_bedsit? &&  record.beds &&  record.beds > 1` on line 14 and we remove the guard clause?

arfacamble: more concise formulation:

in general for these validations and this follows a pattern more generally, these guard clauses aren't really for the proper logic, they're guarding against errors (and excluding cases we don't care about by so doing).

generally these guard clauses across these methods are ways of avoiding get nil reference exceptions by eg trying to compare nil to 1 below

db/schema.rb
```diff
@@ -500,14 +500,14 @@
...
-    t.integer "extrabor_value_check"
-    t.integer "deposit_and_mortgage_value_check"
     t.integer "shared_ownership_deposit_value_check"
+    t.integer "extrabor_value_check"
     t.integer "grant_value_check"
     t.integer "old_persons_shared_ownership_value_check"
+    t.integer "deposit_and_mortgage_value_check"
     t.index ["bulk_upload_id"], name: "index_sales_logs_on_bulk_upload_id"
     t.index ["created_by_id"], name: "index_sales_logs_on_created_by_id"
     t.index ["owning_organisation_id"], name: "index_sales_logs_on_owning_organisation_id"
```
kosiakkatrina: Does this file need to be here?

app/models/validations/sales/property_validations.rb
arfacamble: I was also thinking you could collapse as Kat says above, which I think is another reason it's not necessary to call is_bedsit more than once.
I slightly lean towards guard clause and then slightly more concise if below, but it's much of a muchness

db/schema.rb
SamSeed-Softwire: I thought that in general if any migrations were required (e.g. I try to run the app and it tells me to run a migration) then I should commit any changes due to that migration. Is that not true? Thanks!

app/models/validations/sales/property_validations.rb
SamSeed-Softwire: Thanks both :) I like the idea of keeping the guard clause, but of course making it a bit more readable and in line with convention. Have done so - lmk if you have any other suggestions. Thanks!

app/models/validations/sales/property_validations.rb
```diff
@@ -7,4 +7,13 @@ def validate_postcodes_match_if_discounted_ownership(record)
...
   end
+
+  def validate_bedsit_number_of_beds(record)
+    return unless record.proptype.present? && record.beds.present?
+
+    if record.is_bedsit? && record.beds > 1
+      record.errors.add :proptype, I18n.t("validations.property.proptype.bedsits_have_max_one_bedroom")
+      record.errors.add :beds, I18n.t("validations.property.beds.bedsits_have_max_one_bedroom")
+    end
+  end
```
arfacamble, line 18: 🎉

arfacamble (approved):

PR #1189, opened 2023-01-16
CLDC-1750 Add missing mortgage routing behaviour

natdeanlewissoftwire:
Adds missing routing behaviour, spotted in po review of https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/1163


jamesrose:
@natdeanlewissoftwire Can you add detail to the title/description re: what you're doing here? Thanks!

kosiakkatrina (commented):
I think this change is already on main, I've added it as part of a routing fix I was working on

PR #1195, opened 2023-01-17
CLDC-808 Consider everyone below 18 a child

kosiakkatrina:
Consider everyone below 18 a child, regardless of their relationship to buyer 1
(addressing comments from https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/1184)


SamSeed-Softwire (approved):
Looks good to me!

PR #1197, opened 2023-01-18
CLDC-1782 Create bulk upload logs

asmega:
# Context

- https://digital.dclg.gov.uk/jira/browse/CLDC-1782
- This change enables logs from bulk upload to actually be created if they are valid

# Changes

- Added associations between logs and bulk upload
- When logs are created they now reference the bulk upload
- The step to check the threshold of whether or not we should create logs has been skipped out for this PR and can be slot in later. At the moment as long as every row in the CSV is considered valid it will create all the logs
- For whatever reason a log cannot be saved to log this to sentry for further analysis later
- Added class in tests providing tooling to create csv rows from a log i.e. reversing the process to create test data. This class is no way complete but offers convenience and lowers effort needed to create test data
- Fixed a bug where renewal was not correctly inferred from correct field if not provided


spec/services/bulk_upload/lettings/log_creator_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+      it "blanks invalid field" do
+        service.call
+
+        record = LettingsLog.last
+        expect(record.age1).to be_blank
+      end
+    end
+
+    context "when valid csv with existing log" do
+      xit "what should happen?"
```
ghost, line 65: missed this test?

asmega: Outside of the scope of this PR, should be covered by https://digital.dclg.gov.uk/jira/browse/CLDC-1888

There will be quite a few holes unfortunately. I think we'll patch these as time goes by.

spec/support/bulk_upload/log_to_csv.rb
```diff
@@ -0,0 +1,254 @@
...
+    case log.leftreg
+    when 3
+      3
+    when 1
+      4
+    when 2
+      5
+    when 0
+      6
+    end
```
kosiakkatrina, line 191: that's odd that it's not matching 👀

spec/support/bulk_upload/log_to_csv.rb
```diff
@@ -0,0 +1,254 @@
...
+  end
+
+  def previous_postcode_known
+    case log.ppcodenk
+    when 1
+      1
+    when 0
+      2
+    end
```
kosiakkatrina: can this be a method? Like `checkbox_value(field)` etc?

asmega: I find they're always different which is not ideal. look at `cbl` for example which does the inverse...

asmega: oh i misread, sure

asmega: done

asmega:
This has been merged via https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/1208

PR #1191, opened 2023-01-16
CLDC-845 Add sales household details validations

kosiakkatrina:
Add the following validations:
- [x] Buyer 1 age must be between 16 - 110
image

- [x] If male and ECSTAT1, (ECSTAT2, ECSTAT3 etc) is 5 (Retired), AGE1 (AGE2, AGE3, etc) must be 65 or over (Soft validation)  
image

- [x] If female and ECSTAT1, (ECSTAT2, ECSTAT3 etc) is 5 (Retired), AGE1 (AGE2, AGE3 etc) must be 60 or over (Soft validation)
- [x] TYPE cannot be 24 if no purchasers are over 64
image

PR #1199, opened 2023-01-18
CLDC-848 Add household details validations

kosiakkatrina:
- [x]  Buyer 2's age must be below 110 [person 2's age / person 3's age etc etc] 
image

- [ ] If male and ECSTAT2, (ECSTAT3 etc) is 5, AGE2  (AGE3, etc) must be 65 or over - covered in https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/1191

- [ ] If female and ECSTAT2, (ECSTAT3 etc), AGE2 (AGE3 etc) must be 60 or over - covered in https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/1191
- [x] Must be under 16 if RELAT2 is "Child" - covered in CLDC-850

- [x] If 16-19 and ECSTAT2 is 7, RELAT2 must be child - covered in CLDC-850

- [x] RELAT2 cannot be child if AGE2 is 20 or over  - covered in CLDC-850

- [x] ECSTAT2 Must be 9 if AGE2 is under 16
image

- [x] ECSTAT2 cannot be 9 if Age2 is 16 or over
image

- [ ] TYPE cannot be 24 if no purchasers are over 64 - covered in https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/1191

### DIFF
--- a/app/models/validations/sales/household_validations.rb
+++ b/app/models/validations/sales/household_validations.rb
@@ -13,6 +13,7 @@ module Validations::Sales::HouseholdValidations
     (2..6).each do |n|
       validate_person_age_matches_relationship(record, n)
       validate_person_age_and_relationship_matches_economic_status(record, n)
+      validate_person_age_matches_economic_status(record, n)
     end
     shared_validate_partner_count(record, 6)
   end
@@ -46,6 +47,21 @@ private
     end
   end
 
+  def validate_person_age_matches_economic_status(record, person_num)
+    age = record.public_send("age#{person_num}")
+    economic_status = record.public_send("ecstat#{person_num}")
+    return unless age && economic_status
+
+    if age < 16 && !tenant_is_economic_child?(economic_status)
+      record.errors.add "ecstat#{person_num}", I18n.t("validations.household.ecstat.child_under_16", person_num:)
+      record.errors.add "age#{person_num}", I18n.t("validations.household.age.child_under_16", person_num:)
+    end
+    if tenant_is_economic_child?(economic_status) && age > 16
+      record.errors.add "ecstat#{person_num}", I18n.t("validations.household.ecstat.child_over_16", person_num:)
+      record.errors.add "age#{person_num}", I18n.t("validations.household.age.child_over_16", person_num:)
+    end
+  end
+
   def person_is_partner?(relationship)
     relationship == "P"
   end
@@ -60,5 +76,9 @@ private
 
   def person_is_child?(relationship)
     relationship == "C"
+  end
+
+  def tenant_is_economic_child?(economic_status)
+    economic_status == 9
   end
 end
--- a/app/models/validations/sales/household_validations.rb
+++ b/app/models/validations/sales/household_validations.rb
@@ -40,10 +40,26 @@ private
     relationship = record.public_send("relat#{person_num}")
     return unless age && economic_status && relationship
 
-    if age >= 16 && age <= 19 && person_is_fulltime_student?(economic_status) && !person_is_child?(relationship)
-      record.errors.add "age#{person_num}", I18n.t("validations.household.age.student_16_19")
-      record.errors.add "ecstat#{person_num}", I18n.t("validations.household.ecstat.student_16_19")
-      record.errors.add "relat#{person_num}", I18n.t("validations.household.relat.student_16_19")
+    age_between_16_19 = age.between?(16, 19)
+    student = person_is_fulltime_student?(economic_status)
+    child = person_is_child?(relationship)
+
+    if age_between_16_19 && student && !child
+      record.errors.add "age#{person_num}", I18n.t("validations.household.age.student_16_19.cannot_be_16_19.student_not_child")
+      record.errors.add "ecstat#{person_num}", I18n.t("validations.household.ecstat.student_16_19.cannot_be_student.16_19_not_child")
+      record.errors.add "relat#{person_num}", I18n.t("validations.household.relat.student_16_19.must_be_child")
+    end
+
+    if age_between_16_19 && !student && child
+      record.errors.add "age#{person_num}", I18n.t("validations.household.age.student_16_19.cannot_be_16_19.child_not_student")
+      record.errors.add "ecstat#{person_num}", I18n.t("validations.household.ecstat.student_16_19.must_be_student")
+      record.errors.add "relat#{person_num}", I18n.t("validations.household.relat.student_16_19.cannot_be_child.16_19_not_student")
+    end
+
+    if !age_between_16_19 && student && child
+      record.errors.add "age#{person_num}", I18n.t("validations.household.age.student_16_19.must_be_16_19")
+      record.errors.add "ecstat#{person_num}", I18n.t("validations.household.ecstat.student_16_19.cannot_be_student.child_not_16_19")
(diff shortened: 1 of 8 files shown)


app/models/validations/sales/household_validations.rb
```diff
@@ -46,6 +47,21 @@ def validate_person_age_and_relationship_matches_economic_status(record, person_
     end
   end
 
+  def validate_person_age_matches_economic_status(record, person_num)
+    age = record.public_send("age#{person_num}")
+    economic_status = record.public_send("ecstat#{person_num}")
```
arfacamble, line 68: so much better than ecstat, floats my boat
🙏

app/models/validations/sales/household_validations.rb
```diff
@@ -61,4 +77,8 @@ def person_economic_status_refused?(economic_status)
   def person_is_child?(relationship)
     relationship == "C"
   end
+
+  def tenant_is_economic_child?(economic_status)
+    economic_status == 9
+  end
```
arfacamble, line 99: how do we feel about endless method syntax for one liners like these?
`def tenant_is_economic_child?(economic_status) = economic_status == 9`
probably golf code for no benefit since they are not widely used as far as I know
I think I'm just in the habit of one liners for small things coming over from C# land

app/models/validations/sales/household_validations.rb
```diff
@@ -46,6 +47,21 @@ def validate_person_age_and_relationship_matches_economic_status(record, person_
...
   end
 
+  def validate_person_age_matches_economic_status(record, person_num)
+    age = record.public_send("age#{person_num}")
+    economic_status = record.public_send("ecstat#{person_num}")
+    return unless age && economic_status
+
+    if age < 16 && !tenant_is_economic_child?(economic_status)
+      record.errors.add "ecstat#{person_num}", I18n.t("validations.household.ecstat.child_under_16", person_num:)
+      record.errors.add "age#{person_num}", I18n.t("validations.household.age.child_under_16", person_num:)
```
arfacamble, line 73: I saw that there are variables in the strings in the yml
I would have expected here that it would read `person_num: person_num` with the first symbol being the variable name for the yml string and the second variable being the value 1,2,3,etc
I guess there is some ruby/rails magic going on here. Can you shed any light on that for me?

app/models/validations/sales/household_validations.rb
```diff
@@ -46,6 +47,21 @@ def validate_person_age_and_relationship_matches_economic_status(record, person_
...
+    end
+    if tenant_is_economic_child?(economic_status) && age > 16
+      record.errors.add "ecstat#{person_num}", I18n.t("validations.household.ecstat.child_over_16", person_num:)
+      record.errors.add "age#{person_num}", I18n.t("validations.household.age.child_over_16", person_num:)
+    end
+  end
```
arfacamble, line 79: given this is adding the same errors as far as I can see, is there a reason you have chosen to have two separate ifs rather than `if (under 16 and !child) || (child and over 16)`
Also as I write this I realise the conditions are for over 16 and under 16, is it the spec that 16 years old can be either?

spec/factories/sales_log.rb
```diff
@@ -66,7 +66,7 @@
       sex5 { "X" }
       sex6 { "X" }
       mortgage { 20_000 }
-      ecstat3 { 10 }
+      ecstat3 { 9 }
```
arfacamble, line 69: I just looked in this file and it looks like a load of magic numbers to me since I don't know how this works fully.
Would be cool to have it explained briefly.

spec/models/validations/sales/household_validations_spec.rb
```diff
@@ -86,6 +86,35 @@
...
+        record.age2 = 14
+        record.ecstat2 = 9
+        household_validator.validate_household_number_of_other_members(record)
+        expect(record.errors["ecstat2"]).to be_empty
+        expect(record.errors["age2"]).to be_empty
+      end
+
+      it "validates that a person with economic status 'child' must be under 16" do
+        record.age2 = 21
+        record.relat2 = "C"
```
arfacamble: since the new validation doesn't mention `.relat` is this needed?

arfacamble (commented):
looks great.
I think I'm only suggesting one change, otherwise I am just interested in finding out about some new bits of the codebase

spec/factories/sales_log.rb
natdeanlewissoftwire: This file is a sort of model sales log that we use in a lot of tests (there are similar factories for other objects) - the magic numbers are just values that in theory should be part of a valid sales log in combination with each other, so usually when you're adding any new fields to the database/updating validation, you'll need to add/update a field to this file too.

app/models/validations/sales/household_validations.rb
kosiakkatrina: Hm, I don't don't mind too much either way. I personally find it a bit easier to read when it's on separate lines and it's currently consistent with what we have everywhere else, but if you can convince the team that this is worth a refactor (cause this happens across the whole codebase), I can live with that!

app/models/validations/sales/household_validations.rb
kosiakkatrina: Hm, yup, I think we can omit it if the argument name is the same as parameter name: https://rubyreferences.github.io/rubychanges//3.1.html#values-in-hash-literals-and-keyword-arguments-can-be-omitted

app/models/validations/sales/household_validations.rb
kosiakkatrina: Hmm, so the first validation is if the tenant is under 16 but not marked as a child, the second one is it they are over 16 and marked as a child. I guess it's different situations/different error messages etc

app/models/validations/sales/household_validations.rb
arfacamble: I am very much not sure I prefer it, but I was excited to learn about this syntax as I was not previously aware of it

spec/models/validations/sales/household_validations_spec.rb
kosiakkatrina: Good point!

app/models/validations/sales/household_validations.rb
arfacamble: very nice

app/models/validations/sales/household_validations.rb
arfacamble: my mistake. I thought I had read the yml directions carefully and they were the same

app/models/validations/sales/household_validations.rb
kosiakkatrina: hm, I don't think I've used one liners for methods before! So this is sorta new to me!

spec/factories/sales_log.rb
arfacamble: aha so there is one standard example sales log that is kind of like a version we default to, and then in any tests we just override that?
so if I call `FactoryBot.build(:sales_log)` I get a rich thing with ownership scheme 2, mortgage 20_000, person 3 with economic status 9 etc
and if I call `FactoryBot.build(:sales_log, ecstat3: 10)` I get the same thing but with person 3 economic status changed?

arfacamble (approved):

PR #1200, opened 2023-01-18
CLDC-1461 add validation ensure postcodes match if discounted ownership purchase

arfacamble:
https://digital.dclg.gov.uk/jira/browse/CLDC-1461

If log type is discounted ownership, their current postcode and last postcode should match.

image

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -429,6 +429,7 @@ GEM
 
 PLATFORMS
   arm64-darwin-21
+  arm64-darwin-22
   x86_64-darwin-19
   x86_64-darwin-20
   x86_64-darwin-21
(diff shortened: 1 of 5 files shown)


Gemfile.lock
```diff
@@ -429,6 +429,7 @@ GEM
 
 PLATFORMS
   arm64-darwin-21
+  arm64-darwin-22
```
natdeanlewissoftwire, line 432: do we want to merge this in this PR?

jamesrose (approved):

Gemfile.lock
jamesrose: It would be be better separate, but I don't think it's harmful.

PR #1207, opened 2023-01-20
CLDC 1457 Add registered org checkbox validation

ghost:
Tell users they need to select at least one checkbox

Screenshot 2023-01-20 at 10 46 29

https://digital.dclg.gov.uk/jira/browse/CLDC-1457

### DIFF
--- a/app/models/form/sales/questions/buyers_organisations.rb
+++ b/app/models/form/sales/questions/buyers_organisations.rb
@@ -15,4 +15,8 @@ class Form::Sales::Questions::BuyersOrganisations < ::Form::Question
     "pregla" => { "value" => "Local Authority" },
     "pregghb" => { "value" => "Help to Buy Agent" },
   }.freeze
+
+  def unanswered_error_message
+    "At least one option must be selected of these four"
+  end
 end
--- a/spec/models/form/sales/questions/buyers_organisations_spec.rb
+++ b/spec/models/form/sales/questions/buyers_organisations_spec.rb
@@ -35,6 +35,10 @@ RSpec.describe Form::Sales::Questions::BuyersOrganisations, type: :model do
     expect(question.hint_text).to eq("Select all that apply")
   end
 
+  it "has the correct unanswered_error_message" do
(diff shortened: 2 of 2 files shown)


app/models/form/sales/questions/buyers_organisations.rb
```diff
@@ -15,4 +15,8 @@ def initialize(id, hsh, page)
     "pregla" => { "value" => "Local Authority" },
     "pregghb" => { "value" => "Help to Buy Agent" },
   }.freeze
+
+  def unanswered_error_message
+    "At least one option must be selected of these four"
+  end
```
arfacamble, line 21: whoa. how is this working? I'm guessing there is some inherited behaviour from `::Form::Question`?
From playing in the app it looks to me like the default behaviour is to show "There is a problem \n You must answer #{@check_answer_label}" and this method just changes the second line of that. Is that right?

arfacamble (approved):

app/models/form/sales/questions/buyers_organisations.rb
ghost: You are right, the default behaviour is calling `unanswered_error_message` on the question object. See https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/blob/main/app/controllers/form_controller.rb#L19

Here I am [overriding](https://en.wikipedia.org/wiki/Method_overriding) that method so that it returns the string I want and does not call the method on the parent class.

arfacamble:
hot stuff
last comment in case you are not aware of them (I am only just)
alternative syntax available:
`def unanswered_error_message = "At least one option ... "`
not sure it's better unless it's more widely used.

app/models/form/sales/questions/buyers_organisations.rb
arfacamble: hot stuff
last comment in case you are not aware of them (I am only just)
alternative syntax available:
`def unanswered_error_message = "At least one option ... "`
not sure it's better unless it's more widely used.

PR #1214, opened 2023-01-23
CLDC-886 add validations for discounts

arfacamble:
Caution, rebased onto Nat's PR [1583](https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/1201) - If reviewing before that is merged, review only the last 4 commits.

[ticket](https://digital.dclg.gov.uk/jira/browse/CLDC-886)

**Validation 1 - done in this PR**
image
___
Compound Validation 1 - separate to own ticket/clarify
    cash discount only applies in case of social homebuy (type 18), which is only selectable in case of ownershipsch = shared ownership, but validation is specific to discounted ownership
___
Compound validation 2 - covered by @natdeanlewissoftwire in 1583
___
**Compound validation 3 - done in this PR**
image


app/models/sales_log.rb
```diff
@@ -171,6 +172,10 @@ def mortgage_not_used?
     mortgageused == 2
   end
 
+  def set_mortgage_value_zero
+    self.mortgage = 0
+  end
```
jamesrose: We generally do this kind of thing in `DerivedVariables::SalesLogVariables` — could you look to move this there?

app/models/form/sales/questions/deposit_and_mortgage_value_check.rb
```diff
@@ -0,0 +1,23 @@
+class Form::Sales::Questions::DepositAndMortgageValueCheck < ::Form::Question
+  def initialize(id, hsh, page)
+    super
+    @id = "deposit_and_mortgage_value_check"
+    @check_answer_label = "Deposit and mortgage against discount confirmation" # is this meant to come from somewhere or do I just write something sensible?
```
jamesrose: Is this displayed anywhere? Can you omit it entirely?

config/locales/en.yml
```diff
@@ -272,6 +272,8 @@ en:
       carehome:
         out_of_range: "Household rent and other charges must be between %{min_chcharge} and %{max_chcharge} if paying %{period}"
         not_provided: "Enter how much rent and other charges the household pays %{period}"
+      percentage:
+        invalid_percentage: "Percentage discount must be 0 - 100"
```
```suggestion
        invalid_percentage: "Percentage discount must be between 0 - 100"
```

app/models/sales_log.rb
arfacamble: I was looking at this for a while with Nat this morning. Perhaps you know something that we don't that will let us do this.
We first wanted to put it in derived variables - but if mortgageused == 2 (no), then mortgage value question is not routed, therefore on the line after set_derived_fields! we hit reset_invalidated_dependent_fields! inside of which as far as I can see mortgage is then reset to nil as it is no longer routed to.
We tried changing the order of those two methods because on a vague feeling level it seems like they shouldn't mind that, but according to a handful of tests they do mind that (we didn't investigate those tests and why they are failing at this point but could revisit that)

app/models/form/sales/questions/deposit_and_mortgage_value_check.rb
arfacamble: whoops I meant to put this comment in to remind me to investigate this properly. I don't know yet what check_answer_label is or why I was just copying the format from other files.

arfacamble: I see what it is. Yes it can be displayed
image
I will assume that this label is acceptable on the basis that a label has not been provided in the ticket. But it seems there are quite a few places where devs are being allowed to make up copy

config/locales/en.yml
arfacamble: in this case the copy is coming directly from the ticket, so I'm surprised that would be incorrect. Should I definitely change this?

app/models/sales_log.rb
kosiakkatrina: You could try changing `selected_answer_option_is_derived?` question method specifically for that question. I think we do it for some other questions already

config/locales/en.yml
kosiakkatrina: I usually set @min and @max values on the question for any numeric questions and that automatically adds a validation, the wording for it is a bit different tho

jamesrose: @arfacamble Our validations should be sentences! You can do as @kosiakkatrina suggested and get it for free.

app/models/sales_log.rb
jamesrose: Could you not set it to zero anyway, even if it's previously made nil?

jamesrose:
@arfacamble I've changed the base of this PR so that it can be easily reviewed sans Nat's changes.

app/models/sales_log.rb
kosiakkatrina: It might clear not routed answers

kosiakkatrina (approved):
Just the rebase!

PR #1220, opened 2023-01-23
CLDC 882 add validation against grant

arfacamble:
[TICKET](https://digital.dclg.gov.uk/jira/browse/CLDC-882)

Hard validation against grant value:
image
___

Soft validation against grant value:
image


app/models/validations/sales/financial_validations.rb
```diff
@@ -11,4 +11,12 @@ def validate_income1(record)
...
     end
   end
+
+  def validate_grant(record)
+    return unless record.grant
+
+    unless (0..999_999).cover?(record.grant)
+      record.errors.add :grant, I18n.t("validations.financial.grant_out_of_range")
+    end
+  end
```
kosiakkatrina: Is this here only so that the error message could be different from the numeric validations?

app/models/validations/sales/soft_validations.rb
```diff
@@ -42,4 +42,10 @@ def hodate_3_years_or_more_saledate?
 
     ((saledate.to_date - hodate.to_date).to_i / 365) >= 3
   end
+
+  def grant_outside_common_range?
+    return unless grant
+
+    !(9_000..16_000).cover? grant
```
kosiakkatrina: grant.between?(9_000, 16_000) might make more sense for numeric values maybe

app/models/validations/sales/financial_validations.rb
arfacamble: Yes I have been working on the assumption that the error message given in the ticket should be the error message displayed, but there seems to be some inconsistency on that form the tickets so I'm happy to take direction.
I assume you're asking as an error will get added anyway since I have applied @min and @max in the Question

db/schema.rb
```diff
@@ -501,6 +502,9 @@
     t.integer "totchild"
     t.integer "hhtype"
     t.integer "hodate_check"
+    t.integer "extrabor_value_check"
+    t.integer "deposit_and_mortgage_value_check"
```
kosiakkatrina: These columns and the one on line 490 shouldn't be here? I think that's why the tests fail on remote

app/models/validations/sales/financial_validations.rb
kosiakkatrina: Asked a question about it in slack, cause I think @bibblobcode did something similar to get a different wording as well. Feels a bit redundant to run same validations twice

db/schema.rb
arfacamble: where the hell have they come from?
I've really tied myself in knots with rebases today, not recommended.

kosiakkatrina (approved):

PR #1223, opened 2023-01-24
CLDC-1764 Bulk upload filter

asmega:
# Context

- https://digital.dclg.gov.uk/jira/browse/CLDC-1764
- After a bulk upload assuming it proceeded to create the logs but a handful were invalid. We need to be able to playback these handful of logs to the user so they can be corrected

# Changes

- When passing the correct url eg `lettings-logs/bulk-upload-results/84/resume` it will redirect and load the lettings index screen showing only logs related to the specified bulk upload
- The LHS filter bar adds an additional area to signify a bulk upload filter is applied
- It also gives the user the ability to remove this applied filter
- The filter bar also has year and status filter removed as uploads relate to a single year and status is fixed to in progress
- When this filter is applied the buttons to create a new log or further do bulk uploads is hidden
- The existing mechanism of persisting filters to session remains, therefore after fixing a log and returning the bulk upload filter will continue to be applied
- The headings on the page have been updated to reflect the state of the bulk upload
- An additional banner displays context help text and offering meta info about the upload
- When all logs are resolved the interstitial page is displayed prompting that bulk upload process is complete
- Should the user revisit the link from the email they will be shown the same interstitial to say the process is complete

# Screenshots

Screenshot 2023-01-25 at 13 00 57

## Interstitial

Screenshot 2023-01-25 at 13 02 04

### DIFF
--- a/app/controllers/bulk_upload_lettings_results_controller.rb
+++ b/app/controllers/bulk_upload_lettings_results_controller.rb
@@ -6,4 +6,24 @@ class BulkUploadLettingsResultsController < ApplicationController
   def show
     @bulk_upload = current_user.bulk_uploads.lettings.find(params[:id])
   end
+
+  def resume
+    @bulk_upload = current_user.bulk_uploads.lettings.find(params[:id])
+
+    set_bulk_upload_logs_filters
+
+    redirect_to(lettings_logs_path(bulk_upload_id: [@bulk_upload.id]))
+  end
+
+private
+
+  def set_bulk_upload_logs_filters
+    hash = {
+      years: [""],
+      status: ["", "in_progress"],
+      user: "all",
+    }
+
+    session["logs_filters"] = hash.to_json
+  end
 end
--- a/app/controllers/bulk_upload_lettings_results_controller.rb
+++ b/app/controllers/bulk_upload_lettings_results_controller.rb
@@ -10,12 +10,20 @@ class BulkUploadLettingsResultsController < ApplicationController
   def resume
     @bulk_upload = current_user.bulk_uploads.lettings.find(params[:id])
 
-    set_bulk_upload_logs_filters
+    if @bulk_upload.lettings_logs.in_progress.count.positive?
+      set_bulk_upload_logs_filters
 
-    redirect_to(lettings_logs_path(bulk_upload_id: [@bulk_upload.id]))
+      redirect_to(lettings_logs_path(bulk_upload_id: [@bulk_upload.id]))
+    else
+      reset_logs_filters
+    end
   end
 
 private
+
+  def reset_logs_filters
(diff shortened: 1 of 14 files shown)


app/controllers/bulk_upload_lettings_results_controller.rb
```diff
@@ -6,4 +6,32 @@ class BulkUploadLettingsResultsController < ApplicationController
...
+  end
+
+  def set_bulk_upload_logs_filters
+    hash = {
+      years: [""],
+      status: ["", "in_progress"],
+      user: "all",
+    }
+
+    session["logs_filters"] = hash.to_json
```
asmega, line 35: we should add an interface or service for filtering next time round, otherwise we have lots of ad-hoc random controller code

app/models/user.rb
```diff
@@ -146,9 +146,9 @@ def assignable_roles
 
   def logs_filters(specific_org: false)
     if (support? && !specific_org) || organisation.has_managing_agents?
-      %w[status years user organisation]
+      %w[status years user organisation bulk_upload_id]
     else
-      %w[status years user]
+      %w[status years user bulk_upload_id]
```
asmega, line 150: if we have some kind of filter service that wasn't a module we could add this code there. however it seems odd to have to change the user model for filtering which is rather unrelated

app/views/logs/_log_filters.erb
```diff
@@ -3,13 +3,49 @@
...
     <div class="app-filter__content">
       <%= form_with html: { method: :get } do |f| %>
-        <% years = {"2021": "2021/22", "2022": "2022/23"} %>
-        <% all_or_yours = {"all": { label: "All" }, "yours": { label: "Yours" } } %>
-        <%= render partial: "filters/checkbox_filter", locals: { f: f, options: years, label: "Collection year", category: "years" } %>
-        <%= render partial: "filters/checkbox_filter", locals: { f: f, options: status_filters, label: "Status", category: "status" } %>
-        <%= render partial: "filters/radio_filter", locals: { f: f, options: all_or_yours, label: "Logs", category: "user", } %>
+        <% years = { "2021": "2021/22", "2022": "2022/23" } %>
+        <% all_or_yours = { "all": { label: "All" }, "yours": { label: "Yours" } } %>
+        <% bulk_upload_options = (@session_filters["bulk_upload_id"] || []).reject(&:blank?).index_with { |_bulk_upload_id| "With logs from bulk upload" } %>
```
ghost: this could go in a view helper method

ghost (approved):

app/views/logs/_log_filters.erb
asmega: i've just extracted this to a view helper

PR #1225, opened 2023-01-24
CLDC-868 Add purchase price validations

natdeanlewissoftwire:
image


app/models/validations/sales/soft_validations.rb
```diff
@@ -37,9 +37,29 @@ def deposit_over_soft_max?
...
+  end
+
+  def purchase_price_soft_min_or_soft_max
+    value < purchase_price_range.soft_min ? purchase_price_range.soft_min : purchase_price_range.soft_max
+  end
+
+private
+
+  def purchase_price_range
+    LaPurchasePriceRange.find_by(start_year: collection_start_year, la:, bedrooms: beds)
```
natdeanlewissoftwire: should there be a default year for sales before the 2022/3 collection year? Or do we not care about validating/will not encounter old data?

app/helpers/interruption_screen_helper.rb
```diff
@@ -10,7 +10,7 @@ def display_informative_text(informative_text, lettings_log)
               else
                 lettings_log.public_send(argument["key"])
               end
-      translation_params[argument["i18n_template"].to_sym] = value
+      translation_params[argument["i18n_template"].to_sym] = argument["prefix"] == "£" ? "£#{ActionController::Base.helpers.number_to_currency(value, delimiter: ',', format: '%n')}" : value
```
natdeanlewissoftwire: these allow currency to display correctly even when not from a field in a log

app/models/form/sales/pages/handover_date_check.rb
```diff
@@ -1,7 +1,7 @@
 class Form::Sales::Pages::HandoverDateCheck < ::Form::Page
   def initialize(id, hsh, subsection)
     super
-    @depends_on = [{ "hodate_3_years_or_more_saledate?" => true }]
+    @depends_on = [{ "hodate_3_years_or_more_exdate?" => true }]
```
natdeanlewissoftwire: things related to this are fixes from another PR that wasn't doing what I think it should have been

app/helpers/interruption_screen_helper.rb
```diff
@@ -33,7 +33,7 @@ def display_title_text(title_text, lettings_log)
...
```
kosiakkatrina: `[argument["prefix"], {ActionController::Base.helpers.number_to_currency(value, delimiter: ',', format: '%n')}"].join(")` would remove the condition?

config/locales/en.yml
```diff
@@ -429,10 +429,13 @@ en:
...
       min:
         title_text: "You told us the rent is %{brent}"
-        hint_text: "The minimum rent expected for this type of property in this local authority is £%{soft_min_for_period}."
+        hint_text: "The minimum rent expected for this type of property in this local authority is £%{soft_min_for_period}"
       max:
         title_text: "You told us the rent is %{brent}"
-        hint_text: "The maximum rent expected for this type of property in this local authority is £%{soft_max_for_period}."
+        hint_text: "The maximum rent expected for this type of property in this local authority is £%{soft_max_for_period}"
+    purchase_price:
+      title_text: "You told us the purchase price is %{value}"
```
kosiakkatrina, line 459: Can't we have `title_text: "You told us the purchase price is £%{value}"` instead? Would remove the need for prefixes

db/schema.rb
```diff
@@ -490,20 +501,21 @@
...
+    t.string "pcode1"
+    t.string "pcode2"
+    t.integer "pcodenk"
+    t.string "postcode_full"
+    t.boolean "is_la_inferred"
     t.bigint "bulk_upload_id"
-    t.index ["bulk_upload_id"], name: "index_sales_logs_on_bulk_upload_id"
     t.integer "retirement_value_check"
+    t.integer "hodate_check"
+    t.integer "value_value_check"
```
kosiakkatrina, line 521: Is this meant to be here?

app/models/validations/sales/sale_information_validations.rb
```diff
@@ -7,10 +7,10 @@ def validate_deposit_range(record)
     end
   end
 
-  def validate_pratical_completion_date_before_saledate(record)
-    return if record.saledate.blank? || record.hodate.blank?
+  def validate_pratical_completion_date_before_exdate(record)
+    return if record.exdate.blank? || record.hodate.blank?
 
-    unless record.saledate > record.hodate
+    unless record.exdate > record.hodate
```
kosiakkatrina: Is this part of this ticket?

app/models/validations/sales/soft_validations.rb
kosiakkatrina: I think we're unlikely to have 21/22 data, but we will have to add ranges for 23/24 before April

app/models/validations/sales/sale_information_validations.rb
natdeanlewissoftwire: ah no, I had a pending comment addressing these, they're fixes for another PR that was already approved - seems easier just to fix them here

db/schema.rb
natdeanlewissoftwire: same as other hodate comment

app/helpers/interruption_screen_helper.rb
natdeanlewissoftwire: this is to avoid turning a value that is just a number e.g. age to a currency format, I think it's necessary

config/locales/en.yml
natdeanlewissoftwire: see above comment, I think prefixes are necessary to differentiate formatting between currency and non-currency types

app/models/validations/sales/soft_validations.rb
natdeanlewissoftwire: Ok, that's fine - I'll chase that up 👍

app/helpers/interruption_screen_helper.rb
ghost: this change should be covered by tests

config/locales/en.yml
kosiakkatrina: But wouldn't this validation always be £? Cause it's price?

app/helpers/interruption_screen_helper.rb
kosiakkatrina: But the prefix wouldn't be "£" for age right?

natdeanlewissoftwire: yep - maybe I'm missing something, but I think your suggestion would display an age of 5 as 5.00

app/models/log.rb
```diff
@@ -30,10 +30,19 @@ class Log < ApplicationRecord
...
-    @start_year = startdate < window_end_date ? startdate.year - 1 : startdate.year
+    if lettings?
+      return unless startdate
+
+      log_start_date = startdate
+    else
+      return unless saledate
+
+      log_start_date = saledate
+    end
```
jamesrose: Are you sure that you need to do this? `SalesLog#startdate` is already `saledate`.

app/helpers/interruption_screen_helper.rb
jamesrose: I think there's a bit of indirection here because the first condition isn't the primary path. Shall we talk this through to see if we can come up with something a bit more cleaner?

app/models/la_purchase_price_range.rb
```diff
@@ -0,0 +1,2 @@
+class LaPurchasePriceRange < ApplicationRecord
```
jamesrose: I would have gone with `LaSaleRange` to match `LaRentRange` — what do you think?

app/models/validations/sales/soft_validations.rb
```diff
@@ -37,9 +37,29 @@ def deposit_over_soft_max?
...
+
+  def purchase_price_out_of_soft_range?
+    return unless value && beds && la
+
+    purchase_price_range.present? && !value.between?(purchase_price_range.soft_min, purchase_price_range.soft_max)
+  end
+
+  def purchase_price_min_or_max_text
+    value < purchase_price_range.soft_min ? "minimum" : "maximum"
+  end
```
jamesrose: Is there a reason why we couldn't copy the rent error format with separate `min` and `max` keys?

```yaml
    rent:
      min:
        title_text: "You told us the rent is %{brent}"
        hint_text: "The minimum rent expected for this type of property in this local authority is £%{soft_min_for_period}."
        hint_text: "The minimum rent expected for this type of property in this local authority is £%{soft_min_for_period}"
      max:
        title_text: "You told us the rent is %{brent}"
        hint_text: "The maximum rent expected for this type of property in this local authority is £%{soft_max_for_period}."
        hint_text: "The maximum rent expected for this type of property in this local authority is £%{soft_max_for_period}"
```

natdeanlewissoftwire: that was what I initially did, but this would mean we would have two separate pages that contain the same question which I thought was more confusing than just dynamic text

app/models/la_purchase_price_range.rb
natdeanlewissoftwire: Yep, love it 👍

app/models/log.rb
natdeanlewissoftwire: Ah, I didn't know that, this is probably unnecessary then I agree

app/helpers/interruption_screen_helper.rb
natdeanlewissoftwire: Yep, I'm busy til EOD but I'll book some time in tomorrow morning :)

app/helpers/interruption_screen_helper.rb
```diff
@@ -7,6 +7,8 @@ def display_informative_text(informative_text, lettings_log)
       value = if argument["label"]
                 pre_casing_value = lettings_log.form.get_question(argument["key"], lettings_log).answer_label(lettings_log)
                 pre_casing_value.downcase
+              elsif argument["currency"]
+                ["£", ActionController::Base.helpers.number_to_currency(lettings_log.public_send(argument["key"]), delimiter: ",", format: "%n")].join("")
```
asmega: do you need to prefix with `ActionController::Base.helpers`, I would of thought default view helpers would be available in this context?

natdeanlewissoftwire: Interesting, I'll check - just copied this from somewhere else we do a similar thing.

natdeanlewissoftwire: Yep, works great, thanks

app/helpers/interruption_screen_helper.rb
```diff
@@ -7,6 +7,8 @@ def display_informative_text(informative_text, lettings_log)
       value = if argument["label"]
                 pre_casing_value = lettings_log.form.get_question(argument["key"], lettings_log).answer_label(lettings_log)
                 pre_casing_value.downcase
+              elsif argument["currency"]
+                ["£", number_to_currency(lettings_log.public_send(argument["key"]), delimiter: ",", format: "%n")].join("")
```
kosiakkatrina: Could it not be a label true and a currency? _Or a label false and currency?_ edit: never mind, it's executing when false

kosiakkatrina: I don't necessarily see the need for the currency keyword.
If it's a currency and we're getting it from the label, that should set the "£" for us and format it in the way that the answer would be displayed in the answer options. If it's not a label, that means it's most likely a method, so you could define the formatting in that method that's specific to your error

natdeanlewissoftwire: I suppose I'm imagining possible other soft validations that have currency displayed numbers for methods. Happy to do the formatting in the method, though I think it makes a bit more sense in the helper rather than on the sales log object. What do you think?

kosiakkatrina: I guess it felt like this should already be happening somewhere and we wouldn't have to explicitly define it, but happy for this to go in if it's more readable and makes more sense!

app/models/form/sales/pages/about_price_shared_ownership_value_check.rb
```diff
@@ -0,0 +1,43 @@
...
+      "arguments" => [
+        {
+          "key" => "purchase_price_soft_min_or_soft_max",
+          "label" => false,
+          "i18n_template" => "soft_min_or_soft_max",
+          "currency" => true,
+        },
+        {
+          "key" => "purchase_price_min_or_max_text",
+          "label" => false,
```
arfacamble, line 30: `"label" => false` not necessary, do we want it for readability for any reason?

app/models/la_sale_range.rb
```diff
@@ -0,0 +1,2 @@
+class LaSaleRange < ApplicationRecord
+end
```
arfacamble, line 2: what's this? Doesn't look like it's doing much

app/models/validations/sales/sale_information_validations.rb
```diff
@@ -1,9 +1,18 @@
...
-    return if record.saledate.blank? || record.hodate.blank?
+  def validate_deposit_range(record)
+    return if record.deposit.blank?
 
-    unless record.saledate > record.hodate
-      record.errors.add :hodate, "Practical completion or handover date must be before exchange date"
+    unless record.deposit >= 0 && record.deposit <= 999_999
+      record.errors.add :deposit, "Cash deposit must be £0 - £999,999"
+    end
+  end
```
arfacamble: shouldn't this just be done by putting min and max on the deposit question now?

app/models/validations/sales/soft_validations.rb
```diff
@@ -43,6 +43,18 @@ def extra_borrowing_expected_but_not_reported?
     extrabor != 1 && mortgage + deposit > value - value * discount / 100
   end
 
+  def hodate_3_years_or_more_exdate?
+    return unless hodate && exdate
+
+    ((exdate.to_date - hodate.to_date).to_i / 365) >= 3
```
arfacamble: I bet ruby has a fancy way of doing this

app/models/validations/sales/soft_validations.rb
```diff
@@ -43,6 +43,18 @@ def extra_borrowing_expected_but_not_reported?
...
+  end
+
+  def purchase_price_out_of_soft_range?
+    return unless value && beds && la
+
+    sale_range.present? && !value.between?(sale_range.soft_min, sale_range.soft_max)
```
arfacamble: how come sale_range isn't just on line 53 with the rest of the required fields?

app/services/imports/sale_ranges_service.rb
```diff
@@ -0,0 +1,31 @@
...
+            soft_max: row["soft_max"] },
+          unique_by: %i[start_year bedrooms la],
+        )
+        self.count = count + 1
+      end
+    end
+
+  private
+
+    attr_writer :count
```
arfacamble: I don't think you need this if you change 23 to `@count  = @count + 1` or even `@count += 1`

spec/models/form_handler_spec.rb
```diff
@@ -52,14 +52,14 @@
     it "is able to load a current sales form" do
       form = form_handler.get_form("current_sales")
       expect(form).to be_a(Form)
-      expect(form.pages.count).to eq(191)
+      expect(form.pages.count).to eq(192)
```
arfacamble: this number is always incorrect.

arfacamble (changes requested):

app/models/form/sales/pages/about_price_shared_ownership_value_check.rb
natdeanlewissoftwire: I think so, this is the format being used in the year .json forms

app/models/la_sale_range.rb
natdeanlewissoftwire, line 2: this is needed so we can add the La Sale Range data to the La Sale Range table in seeds.rb

app/models/validations/sales/sale_information_validations.rb
natdeanlewissoftwire: not related to this ticket, will be removed after rebase

app/models/validations/sales/soft_validations.rb
natdeanlewissoftwire: stellar point

app/services/imports/sale_ranges_service.rb
natdeanlewissoftwire: love it, looks like it works to me, have updated the rent ranges file i got this from too

app/models/validations/sales/soft_validations.rb
natdeanlewissoftwire: not related to this ticket, this change was removed after rebase
agreed though, have railsified

spec/models/form_handler_spec.rb
natdeanlewissoftwire: fixed now but not for long...

config/locales/en.yml
natdeanlewissoftwire: have removed prefixes by using unit in number_to_currency 👍

arfacamble (approved):

PR #1227, opened 2023-01-25
Change numeric validation to display prefixes in error message

kosiakkatrina:
Update the numeric validation method to display prefixes, suffixes and appropriate delimiters for large values
image
image

### DIFF
--- a/app/models/validations/shared_validations.rb
+++ b/app/models/validations/shared_validations.rb
@@ -19,17 +19,19 @@ module Validations::SharedValidations
       next unless record[question.id]
 
       field = question.check_answer_label || question.id
+      min = [question.prefix, question.min].join("")
+      max = [question.prefix, question.max].join("")
 
       begin
         answer = Float(record.public_send("#{question.id}_before_type_cast"))
       rescue ArgumentError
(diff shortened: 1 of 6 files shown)


app/models/validations/shared_validations.rb
```diff
@@ -19,17 +19,19 @@ def validate_numeric_min_max(record)
       next unless record[question.id]
 
       field = question.check_answer_label || question.id
+      min = [question.prefix, question.min].join("")
```
SamSeed-Softwire: Nice, although it might be good to also add a comma delimiter between the 1000s?

The below works but you also need to add `include ActionView::Helpers::NumberHelper` to the module.

```suggestion
      min = [question.prefix, number_with_delimiter(question.min, delimiter: ',')].join("")
```

arfacamble: similarly - if we are prepending prefixes eg currency, do we want to append suffixes eg %?

kosiakkatrina: yup! good spot!

kosiakkatrina: Nice one!

SamSeed-Softwire (approved):
Looks good to me!

jamesrose (approved):

PR #1228, opened 2023-01-25
CLDC-878 Validate basic monthly rent

kosiakkatrina:
Validate shared ownership basic monthly rent if the type of sale is not old persons shared ownership
image

### DIFF
--- a/app/models/validations/sales/sale_information_validations.rb
+++ b/app/models/validations/sales/sale_information_validations.rb
@@ -54,4 +54,13 @@ module Validations::Sales::SaleInformationValidations
       end
     end
   end
+
+  def validate_basic_monthly_rent(record)
+    return unless record.mrent && record.ownershipsch && record.type
+
+    if record.shared_owhership_scheme? && !record.old_persons_shared_ownership? && record.mrent > 9999
+      record.errors.add :mrent, I18n.t("validations.sale_information.monthly_rent.higher_than_expected")
+      record.errors.add :type, I18n.t("validations.sale_information.monthly_rent.higher_than_expected")
+    end
+  end
 end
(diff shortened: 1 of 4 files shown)


app/models/validations/sales/sale_information_validations.rb
```diff
@@ -23,4 +23,13 @@ def validate_previous_property_unit_type(record)
       record.errors.add :fromprop, I18n.t("validations.sale_information.previous_property_type.property_type_bedsit")
     end
   end
+
+  def validate_basic_monthly_rent(record)
+    return unless record.mrent && record.ownershipsch && record.type
+
+    if record.shared_owhership_scheme? && !record.old_persons_shared_ownership? && record.mrent > 9999
```
arfacamble, line 61: I would expect rubocop would complain about 9999 and say it needs to be 9_999 
does it only start complaining after 5 digits or more?

spec/models/validations/sales/sale_information_validations_spec.rb
```diff
@@ -130,4 +130,61 @@
...
+
+      it "does not add an error" do
+        sale_information_validator.validate_basic_monthly_rent(record)
+
+        expect(record.errors[:mrent]).not_to be_present
+        expect(record.errors[:type]).not_to be_present
+      end
+    end
+
+    context "when the type is 24" do
```
arfacamble: It feels more readable to me that in the title of the test we would say what type 24 represents, and then type:24 on l158 is implicit. This is the older persons thing right?

arfacamble: this?

arfacamble (approved):

app/models/validations/sales/sale_information_validations.rb
kosiakkatrina: I think 4 9s is reasonable to read

spec/models/validations/sales/sale_information_validations_spec.rb
kosiakkatrina: Good spot!

PR #1229, opened 2023-01-25
CLDC-1899 Add discounted ownership value validation

kosiakkatrina:
image

### DIFF
--- a/app/models/validations/sales/sale_information_validations.rb
+++ b/app/models/validations/sales/sale_information_validations.rb
@@ -40,12 +40,15 @@ module Validations::Sales::SaleInformationValidations
   end
 
   def validate_discounted_ownership_value(record)
-    return unless record.mortgage && record.value && record.deposit && record.ownershipsch && (record.discount || record.grant)
+    return unless record.value && record.deposit && record.ownershipsch
+    return unless record.mortgage || record.mortgageused == 2
+    return unless record.discount || record.grant || record.type == 29
 
     discount_amount = record.discount ? record.value * record.discount / 100 : 0
     grant_amount = record.grant || 0
+    mortgage_amount = record.mortgage || 0
     value_with_discount = (record.value - discount_amount)
-    if record.mortgage + record.deposit + grant_amount != value_with_discount && record.discounted_ownership_sale?
+    if mortgage_amount + record.deposit + grant_amount != value_with_discount && record.discounted_ownership_sale?
       %i[mortgage deposit grant value discount ownershipsch].each do |field|
         record.errors.add field, I18n.t("validations.sale_information.discounted_ownership_value", value_with_discount: sprintf("%.2f", value_with_discount))
       end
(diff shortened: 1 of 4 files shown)


app/models/validations/sales/sale_information_validations.rb
```diff
@@ -23,4 +23,17 @@ def validate_previous_property_unit_type(record)
       record.errors.add :fromprop, I18n.t("validations.sale_information.previous_property_type.property_type_bedsit")
     end
   end
+
+  def validate_discounted_ownership_value(record)
+    return unless record.mortgage && record.value && record.deposit && record.ownershipsch && (record.discount || record.grant)
+
```
natdeanlewissoftwire, line 46: don't we also want this validation to run in the case that there is neither a discount nor a grant (i.e. `rent_to_buy_full_ownership?` == true) or is that covered by a separate validation?

natdeanlewissoftwire (approved):
lgtm, one query

app/models/validations/sales/sale_information_validations.rb
kosiakkatrina: hmm, good question! Lemme check!

kosiakkatrina: Have a look now?

app/models/validations/sales/sale_information_validations.rb
```diff
@@ -23,4 +23,20 @@ def validate_previous_property_unit_type(record)
       record.errors.add :fromprop, I18n.t("validations.sale_information.previous_property_type.property_type_bedsit")
     end
   end
+
+  def validate_discounted_ownership_value(record)
+    return unless record.value && record.deposit && record.ownershipsch
+    return unless record.mortgage || record.mortgageused == 2
```
natdeanlewissoftwire, line 44: I think as part of one of Arthur's PRs mortgageuser == 2 => mortgage = 0, but I'm not sure if that behaviour was decided against in the end. Either way this should be unaffected but maybe unnecessary 👍

natdeanlewissoftwire (approved):

app/models/validations/sales/sale_information_validations.rb
kosiakkatrina: Ooh, @arfacamble do you know which change/PR that is?

arfacamble: yes it's in 886
link in [PR diff](https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/1214/files#diff-140301ddbd5f381586fc3d0fbae7f21d15ef45776bcc72ce90597c5c782d07d0)
it's the work I was doing that you helped me with yesterday when I needed to move the set_derived_fields to below reset_invalidated_fields
I am waiting to finalise that PR once Nat's 1583 moves through PO review which I was expecting to happen today. (due to rebase conflict annoyances)

the upshot is that you should be able to just return unless mortgage as on line 28 and the fix making that happy will be merged in soon.

kosiakkatrina: Ah nice, okay, will update it when it's merged

PR #1232, opened 2023-01-25
CLDC-1464 Add completion date validations

natdeanlewissoftwire:
image

image


app/models/validations/sales/sale_information_validations.rb
```diff
@@ -3,18 +3,22 @@ def validate_pratical_completion_date_before_saledate(record)
...
-    record.errors.add(:exdate, I18n.t("validations.sale_information.exdate.must_be_before_saledate")) if record.exdate > record.saledate
-
-    return if (record.saledate.to_date - record.exdate.to_date).to_i / 365 < 1
+    if record.exdate > record.saledate
+      record.errors.add :exdate, I18n.t("validations.sale_information.exdate.must_be_before_saledate")
+      record.errors.add :saledate, I18n.t("validations.sale_information.saledate.must_be_after_exdate")
+    end
 
-    record.errors.add(:exdate, I18n.t("validations.sale_information.exdate.must_be_less_than_1_year_from_saledate"))
+    if record.saledate - record.exdate > 1.year
```
natdeanlewissoftwire: the previous version of thos (on old line 15) rounds to integers, which means if the difference is between 1 and 2 years the validation won't work as it should - I think this way is also easier to read which is nice

spec/requests/sales_logs_controller_spec.rb
```diff
@@ -186,31 +186,31 @@
           end
 
           context "with year filter" do
-            let!(:sales_log_2021) do
+            let!(:sales_log_2022) do
```
natdeanlewissoftwire, line 182: now we only allow logs from the 22/23 collection year, filtering by years doesn't really make as much sense but I've left it in as you can see

app/models/validations/sales/sale_information_validations.rb
arfacamble: I think your new line is much nicer anyway for readability. But in what case didn't the old one work? I would expect it to round down the number of years in the division if the number of days is less than 365 so if the time span was under 1 year it would resolve to 0 < 1

that's not really important given your line is nicer anyway.

As a minor point that looks to me it does want changing, it sounds from the description than you want >= rather than >

app/models/validations/sales/setup_validations.rb
```diff
@@ -0,0 +1,9 @@
+module Validations::Sales::SetupValidations
+  def validate_saledate(record)
+    return unless record.saledate
+
+    if record.saledate < Time.zone.local(2022, 4, 1)
+      record.errors.add :saledate, I18n.t("validations.setup.saledate.financial_year")
+    end
+  end
+end
```
arfacamble, line 9: would we not need to test also that the saledate is before (2023, 4, 1)
I wonder if ruby comes with some exciting date range inclusion methods

edit: it does
```
[12] pry(main)> new_year = Date.new(2023, 1, 1)
=> #<Date: 2023-01-01 ((2459946j,0s,0n),+0s,2299161j)>
[13] pry(main)> last_week = Date.new(2023, 1, 20)
=> #<Date: 2023-01-20 ((2459965j,0s,0n),+0s,2299161j)>
[14] pry(main)> tomorrow = Date.new(2023, 1, 30)
=> #<Date: 2023-01-30 ((2459975j,0s,0n),+0s,2299161j)>
[15] pry(main)> Date.today.between?(last_week, tomorrow)
=> true
[16] pry(main)> Date.today.between?(new_year, last_week)
=> false
```

app/models/validations/sales/sale_information_validations.rb
natdeanlewissoftwire: ah good point, yep I changed it to unless... >  rather than if... < at some point but I see that's different, good spot about >=

ghost (approved):

PR #1235, opened 2023-01-25
CLDC-1786 add validations for staircasing transaction

arfacamble:
[JIRA TICKET](https://digital.dclg.gov.uk/jira/browse/CLDC-1786)
***
Validations 1 & 3 - percentage bought and percentage owned must be between zero and one hundred
image
***
Validation 2 (soft) - confirm if percentage bought is greater than 50%
image
***
Validation 4 - percentage bought cannot exceed percentage owned after purchase
image
***
Validation 5 - percentage bought cannot exceed 75 for older persons hared ownership type
image


app/models/form/sales/questions/staircase_bought.rb
```diff
@@ -8,6 +8,6 @@ def initialize(id, hsh, page)
     @width = 5
     @min = 0
     @max = 100
-    @suffix = " percent"
+    @suffix = "%"
```
arfacamble, line 11: @kosiakkatrina I changed this for consistency with other questions on the assumption that symbols are more concise and also a bit more consistent with eg £ prefix. Let me know if it should definitely be "percent"

kosiakkatrina: % looks okay, you might need to rebase and fix some tests in `validate_numeric_min_max` tests

app/models/sales_log.rb
```diff
@@ -134,6 +134,10 @@ def is_type_discount?
     type == 18
   end
 
+  def is_older_persons_shared_ownership?
+    type == 24
+  end
+
```
kosiakkatrina: `old_persons_shared_ownership?` already exists on line 202

db/schema.rb
```diff
@@ -501,9 +501,13 @@
     t.integer "totchild"
     t.integer "hhtype"
     t.integer "hodate_check"
+    t.integer "extrabor_value_check"
+    t.integer "deposit_and_mortgage_value_check"
+    t.integer "grant_value_check"
```
kosiakkatrina: Is this meant to be here?

app/models/sales_log.rb
arfacamble: good catch thanks

db/schema.rb
arfacamble: as a more experienced rails developer please let me know if you know why this keeps happening in my branches and what the solution is broadly.
It was my understanding that the schema file is auto-generated and generally I should just let it do what it wants, when I migrate it should add/remove/amend the relevant things, and sometimes it reorders some and that isn't really a problem.
I was taught when first using rails that it shouldn't be necessary and is generally a cad idea to edit the schema file yourself as it is auto generated. I've rerun db:migrate to give it a chance to fix itself and it seems to be happy and the server runs, the tests run and pass etc etc so that made me inclined to just leave it.
Thoughts?

spec/models/validations/shared_validations_spec.rb
```diff
@@ -71,10 +71,11 @@
 
     context "when validating percent" do
       it "validates that % suffix is added in the error message" do
-        sales_record.stairbought = "random"
+        question_under_test = Form::Sales::Questions::StaircaseBought.new(nil, nil, self)
+        sales_record.stairbought = question_under_test.max + 1
         shared_validator.validate_numeric_min_max(sales_record)
         expect(sales_record.errors["stairbought"])
-          .to include(match I18n.t("validations.numeric.valid", field: "Percentage bought in this staircasing transaction", min: "0 percent", max: "100 percent"))
+          .to include(match I18n.t("validations.numeric.valid", field: question_under_test.check_answer_label, min: "#{question_under_test.min}#{question_under_test.suffix}", max: "#{question_under_test.max}#{question_under_test.suffix}"))
```
arfacamble: @kosiakkatrina let me know what you think of this test.
I think it also has the advantage of having fewer magic numbers and strings floating around in it that could theoretically need updating in future when the structure of the test could still be valid.
I feel like I should amend some of the other test in this file to follow this pattern if we like it.

db/schema.rb
kosiakkatrina: If you drop and recreate (db:drop, db:create, db:migrate) the database that should fix itself. I think it might be happening when you've had different migrations on different branches. If you had ran the migration on your db on one branch,  switched branches and ran migration on another branch, that would update the schema file to represent the state of your db, which now would have changes from both branches (I think). The reason it didn't fix itself when migrating is probably because the migration number in your db is the same as the migration number in your schema file, so it probably thinks it's up to date

spec/models/validations/shared_validations_spec.rb
kosiakkatrina: I think I'd keep it the way it was because then we'd be testing that `question_under_test.check_answer_label` and min/max methods get called correctly and return what we expect as well as testing that the error got added

arfacamble: I think this specifically doesn't care what the max is, it just cares that we set the field to be one greater than whatever the max is.
The way I see it either formulation makes the assumption that the question for the the field has a max set, and writing the test with a lot of hard-coded numbers or strings makes more assumptions that those numbers in the question are what we think they are.

kosiakkatrina: How do you know that `question_under_test.max`, `question_under_test.suffix` and `question_under_test.check_answer_label` aren't nil?

kosiakkatrina (approved):
Just needs a rebase 🎉

PR #1239, opened 2023-01-26
Persist column for bulk upload errors

asmega:
# Context

- At the moment we do not persist the column for bulk upload errors
- For the summary for bulk upload errors we need to display the column
- Persisting the column makes the query for the summary trivial
- If not persisted we'd have to select a substring in query to extract the column then add it to the grouping which makes things a little more complicated and brittle in my opinion

# Changes

- Add a new column called `col` to `bulk_upload_errors` table
- When extracting errors persist the column to database


kosiakkatrina (approved):
Do we need to persist cell? Could it be a method on the model?

asmega:
> Do we need to persist cell? Could it be a method on the model?

Might be able to drop it. as I wasn't on planning on having `col` but then turned out it was useful. i don't think the redundancy/duplication is particularly expensive and kind of happy to keep it for now and drop it later if it turns out its not very useful

PR #1241, opened 2023-01-26
CLDC-1880 Add mortgage length validations

ghost:
https://digital.dclg.gov.uk/jira/browse/CLDC-1880

### DIFF
--- a/app/models/validations/sales/sale_information_validations.rb
+++ b/app/models/validations/sales/sale_information_validations.rb
@@ -54,4 +54,12 @@ module Validations::Sales::SaleInformationValidations
       end
     end
   end
+
+  def validate_mortgage_length(record)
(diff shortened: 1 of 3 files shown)


app/models/validations/sales/sale_information_validations.rb
```diff
@@ -38,4 +38,12 @@ def validate_previous_property_unit_type(record)
...
     end
   end
+
+  def validate_mortgage_length(record)
+    return unless record.mortlen
+
+    return if record.mortlen >= 0 && record.mortlen <= 60
+
+    record.errors.add :mortlen, I18n.t("validations.sale_information.mortlen.range")
+  end
```
kosiakkatrina, line 64: Can't we use min/max for this?

PR #1242, opened 2023-01-26
CLDC-1780 Bulk upload summary

asmega:
# Context

- https://digital.dclg.gov.uk/jira/browse/CLDC-1780
- In order to get a glimpse of what the errors are for a bulk upload we offer a summary page which groups and counts errors together based on all the errors

# Changes

- Add a page that summarises the bulk upload errors
- This will be the page linked to from the email that will act as the landing page
- Therefore the full report page has a back link added to return back to the summary page

# Notes for review

- This PR also contains https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/1239 which can be reviewed and merged independently but is needed for this PR

# Screenshots

Screenshot 2023-01-27 at 14 38 45

### DIFF
--- a/app/components/bulk_upload_error_summary_table_component.rb
+++ b/app/components/bulk_upload_error_summary_table_component.rb
@@ -1,0 +1,17 @@
+class BulkUploadErrorSummaryTableComponent < ViewComponent::Base
+  attr_reader :bulk_upload
+
+  def initialize(bulk_upload:)
+    @bulk_upload = bulk_upload
+
+    super
+  end
+
+  def sorted_errors
+    @sorted_errors ||= bulk_upload
+      .bulk_upload_errors
+      .group(:col, :field, :error)
+      .count
+      .sort_by { |el| el[0][0].rjust(3, "0") }
+  end
+end
--- a/app/controllers/bulk_upload_lettings_results_controller.rb
+++ b/app/controllers/bulk_upload_lettings_results_controller.rb
@@ -19,6 +19,10 @@ class BulkUploadLettingsResultsController < ApplicationController
     end
   end
 
+  def summary
+    @bulk_upload = current_user.bulk_uploads.lettings.find(params[:id])
+  end
+
 private
 
   def reset_logs_filters
--- a/app/components/bulk_upload_error_summary_table_component.html.erb
+++ b/app/components/bulk_upload_error_summary_table_component.html.erb
@@ -1,0 +1,25 @@
+<%= govuk_table do |table| %>
+  <% table.caption(size: "m", text: bulk_upload.filename) %>
+
+  <% table.head do |head| %>
+    <% head.row do |row| %>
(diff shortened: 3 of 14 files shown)


app/components/bulk_upload_error_summary_table_component.rb
```diff
@@ -0,0 +1,17 @@
...
+
+    super
+  end
+
+  def sorted_errors
+    @sorted_errors ||= bulk_upload
+      .bulk_upload_errors
+      .group(:col, :field, :error)
+      .count
+      .sort_by { |el| el[0][0].rjust(3, "0") }
```
kosiakkatrina, line 15: can this/these be a scope(s)?

app/controllers/bulk_upload_lettings_results_controller.rb
```diff
@@ -19,6 +19,10 @@ def resume
     end
   end
 
+  def summary
+    @bulk_upload = current_user.bulk_uploads.lettings.find(params[:id])
+  end
+
```
kosiakkatrina, line 25: Could we extract this to a `before_action :find_bulk_upload, only: %I[summary show resume]` etc?

asmega: I'm not really a fan of this pattern. For example like https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/blob/main/app/controllers/locations_controller.rb#L3 you end up with a load of intermingled `before_action`s and takes a while to work out state for a specific method. whereas here its super obvious and zero doubt

app/components/bulk_upload_error_summary_table_component.rb
asmega: the `sort_by` can't be a scope as its ruby and SQL. I'm not aware of how to change or apply a different sort algorithm to postgres. I don't see the benefit of extracting the whole query to a scope as it's not used anywhere else

asmega: or maybe it could be a scope but with my latter point feel it's kind of moot

kosiakkatrina: I guess it might be quicker if it's sql, but not sure how easy it would be to translate, so will trust you on that one

kosiakkatrina (approved):

jamesrose (approved):

app/components/bulk_upload_error_summary_table_component.rb
asmega: i would also prefer to do it in SQL the following is the issue:
```SQL
SELECT * FROM (VALUES(1, 'A'), (2, 'B'), (3, 'AA')) AS t (id, field) order by field;

 id | field
----+-------
  1 | A
  3 | AA
  2 | B
(3 rows)
```

the select would have to be padded in order for the default sort to work. This can be done but don't feel it'll be worth the effort for the return we'd gain

PR #1244, opened 2023-01-26
CLDC-1875 Create bulk upload Notify email templates and corresponding methods

SamSeed-Softwire:
## Context
- https://digital.dclg.gov.uk/jira/browse/CLDC-1875
- Automated emails need to get sent when using the bulk upload service (e.g. to inform users if the upload has succeeded or failed). This ticket covers:
  - Creating appropriate templates in GOV UK Notify
  - Writing methods which tell Notify to send emails using the templates mentioned above.

## The changes
- Creation of 5 templates in the `Bulk upload` folder in Notify (https://www.notifications.service.gov.uk/services/ccdf7d34-7aa1-4aa9-9481-46c89a34db57/templates).
- Creation of a new mailer class `BulkUploadMailer` which contains 5 methods and corresponding template IDs which match the 5 templates mentioned above.

## Note
- This ticket doesn't cover passing the correct personalised information to the aforementioned methods - currently they just send dummy text strings to Notify. Separate tickets will cover inserting the correct information.
- Currently the `bulk_upload` parameter is being used in each method as a text string. This is just to avoid a lint failure (lint complains if I'm not using a parameter defined in a method signature). I'm not using the parameter in the way it will be used in the future (as an actual bulk upload object presumably) because that will be done in other tickets, as mentioned in the above point.

## Example
- A template in Notify:
image

- The corresponding method:
image

- The email received after manually triggering the aforementioned method from the Rails console (with the user set as me and `bulk_upload` set as the text string "dummy"):
image

### DIFF
--- a/app/mailers/bulk_upload_mailer.rb
+++ b/app/mailers/bulk_upload_mailer.rb
@@ -1,6 +1,23 @@
 class BulkUploadMailer < NotifyMailer
+  BULK_UPLOAD_COMPLETE_TEMPLATE_ID = "83279578-c890-4168-838b-33c9cf0dc9f0".freeze
   BULK_UPLOAD_FAILED_CSV_ERRORS_TEMPLATE_ID = "e27abcd4-5295-48c2-b127-e9ee4b781b75".freeze
+  BULK_UPLOAD_FAILED_FILE_SETUP_ERROR_TEMPLATE_ID = "24c9f4c7-96ad-470a-ba31-eb51b7cbafd9".freeze
+  BULK_UPLOAD_FAILED_SERVICE_ERROR_TEMPLATE_ID = "c3f6288c-7a74-4e77-99ee-6c4a0f6e125a".freeze
   BULK_UPLOAD_WITH_ERRORS_TEMPLATE_ID = "eb539005-6234-404e-812d-167728cf4274".freeze
+
+  def send_bulk_upload_complete_mail(user, bulk_upload)
+    send_email(
+      user.email,
+      BULK_UPLOAD_COMPLETE_TEMPLATE_ID,
+      {
+        title: "[dummy title]",
+        filename: "[dummy filename]",
+        upload_timestamp: "[dummy upload_timestamp]",
+        success_description: "[dummy success_description]",
+        logs_link: "[dummy logs_link]"
+      },
+    )
+  end
 
   def send_bulk_upload_failed_csv_errors_mail(user, bulk_upload)
     send_email(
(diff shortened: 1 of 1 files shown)


ghost (changes requested):
This mailer could come with tests? 
You can look at this for inspiration:
https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/blob/main/spec/mailers/csv_download_mailer_spec.rb

app/mailers/bulk_upload_mailer.rb
```diff
@@ -0,0 +1,79 @@
+class BulkUploadMailer < NotifyMailer
+  BULK_UPLOAD_COMPLETE_TEMPLATE_ID = "83279578-c890-4168-838b-33c9cf0dc9f0".freeze
+  BULK_UPLOAD_FAILED_CSV_ERRORS_TEMPLATE_ID = "e27abcd4-5295-48c2-b127-e9ee4b781b75".freeze
+  BULK_UPLOAD_FAILED_FILE_SETUP_ERROR_TEMPLATE_ID = "24c9f4c7-96ad-470a-ba31-eb51b7cbafd9".freeze
+  BULK_UPLOAD_FAILED_SERVICE_ERROR_TEMPLATE_ID = "c3f6288c-7a74-4e77-99ee-6c4a0f6e125a".freeze
+  BULK_UPLOAD_WITH_ERRORS_TEMPLATE_ID = "eb539005-6234-404e-812d-167728cf4274".freeze
+
+  def send_bulk_upload_complete_mail(user, bulk_upload)
```
arfacamble, line 8: wouldn't these methods need to consume different variables for each field in the email template?

app/mailers/bulk_upload_mailer.rb
```diff
@@ -0,0 +1,79 @@
...
+    send_email(
+      user.email,
+      BULK_UPLOAD_COMPLETE_TEMPLATE_ID,
+      {
+        title: "[#{bulk_upload} title]",
+        filename: "[#{bulk_upload} filename]",
+        upload_timestamp: "[#{bulk_upload} upload_timestamp]",
```
arfacamble, line 15: if the things being interpolated/inserted here is a `Time` object, how would we like it formatted? And is that going to be part of another ticket or is the time formatting linked enough to the templating in this ticket to be worth doing now?
Also is there a standard gov or CORE set of date presentation formats?

app/mailers/bulk_upload_mailer.rb
asmega: it's possible some of them will need additional parameters. some can be pulled from `bulk_upload` eg `filename` would be `bulk_upload.filename` and `upload_timestamp` from `bulk_upload.created_at`

app/mailers/bulk_upload_mailer.rb
asmega: > Also is there a standard gov or CORE set of date presentation formats?

there are, please see the style guide https://www.gov.uk/guidance/style-guide/a-to-z-of-gov-uk-style#dates 

> if the things being interpolated/inserted here is a Time object, how would we like it formatted?

it'll be one of the above. these should already be available in `Time::DATE_FORMATS` specified by `config/initializers/date_formats.rb`

> And is that going to be part of another ticket or is the time formatting linked enough to the templating in this ticket to be worth doing now?

yea this will be plumbed in later. this is just to open the integration point so the next piece of work can be done

asmega:
> This mailer could come with tests? You can look at this for inspiration: https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/blob/main/spec/mailers/csv_download_mailer_spec.rb

I think we'd write the tests when we come to plumb in this work. at the moment this sits as dead code and is not actually used

SamSeed-Softwire:
> > This mailer could come with tests? You can look at this for inspiration: https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/blob/main/spec/mailers/csv_download_mailer_spec.rb
> 
> I think we'd write the tests when we come to plumb in this work. at the moment this sits as dead code and is not actually used

Hi @bibblobcode, thanks for the review comments. Given what Phil wrote above, are you happy with the state of this PR and happy to approve it?

ghost (approved):

PR #1247, opened 2023-01-27
CLDC-1924 Add monthly charges validation

kosiakkatrina:
image


app/models/sales_log.rb
```diff
@@ -34,7 +34,7 @@ class SalesLog < Log
   scope :search_by, ->(param) { filter_by_id(param) }
   scope :filter_by_organisation, ->(org, _user = nil) { where(owning_organisation: org) }
 
-  OPTIONAL_FIELDS = %w[purchid].freeze
+  OPTIONAL_FIELDS = %w[purchid monthly_charges_value_check].freeze
```
arfacamble: I'm interested what this does.

app/models/validations/sales/soft_validations.rb
```diff
@@ -54,4 +54,11 @@ def grant_outside_common_range?
 
     !grant.between?(9_000, 16_000)
   end
+
+  def monthly_charges_over_soft_max?
+    return unless type && mscharge && proptype
+
+    soft_max = old_persons_shared_ownership? ? 550 : 300
+    mscharge > soft_max
+  end
```
arfacamble, line 84: smashing

config/locales/en.yml
```diff
@@ -459,6 +459,8 @@ en:
       title_text: "You told us the time between the start of the tenancy and the major repairs completion date is more than 2 years"
     void_date:
       title_text: "You told us the time between the start of the tenancy and the void date is more than 2 years"
+    monthly_charges_over_soft_max: 
+      title_text: "The amount of monthly charges is high for this type of property and sale type"
```
arfacamble, line 474: I don't imagine it matters either way, but given that there is only one string related to monthly_charges_over_soft_max is there a benefit to having one more or one fewer layers of nesting in the yml.
eg here just having monthly_charges_over_soft_max_title_text

db/migrate/20230127102334_add_monthly_charges_value_check.rb
```diff
@@ -0,0 +1,7 @@
+class AddMonthlyChargesValueCheck < ActiveRecord::Migration[7.0]
+  def change
+    change_table :sales_logs, bulk: true do |t|
+      t.column :monthly_charges_value_check, :integer
+    end
+  end
+end
```
arfacamble, line 7: purely out of interest, is there a benefit in migrations to doing a change_table over just an add_column?
also, is the bulk: true doing anything here? When I investigated it looked to me from docs that all it does is group together the changes within a change_table, so if there's only one then it won't be doing anything as far as I can tell.

arfacamble (approved):
I put a few comments for my own interest that I would be interested to know your opinions/knowledge on, but looks great.

app/models/sales_log.rb
kosiakkatrina: Ah, right, so because this validation depends on a question in setup section (type), whenever this validation gets triggered this question also appears in the setup section. 
When the setup section isn't completed we don't allow the rest of the sections to be started, so whenever this validation gets triggered, setup section goes from completed to in_progress and locks and clears all the other sections. Making this questions optional basically prevents that.

config/locales/en.yml
kosiakkatrina: Hmm, dunno, this is somewhat consistent and I don't think it's too difficult to read

db/migrate/20230127102334_add_monthly_charges_value_check.rb
kosiakkatrina: Yeah, it doesn't really matter when it's just one change, it's just a different syntax at that point

PR #1249, opened 2023-01-30
CLD-1857 Update household characteristics section

kosiakkatrina:
Change the headers for person/buyer questions.

Previously if it was a joint purchase we would refer to people in the household as:
- Buyer 1
- Buyer 2
- Person 1 
- Person 2
- Person 3
- ...

If it was a non-joint purchase, they'd be referred to as:
- Buyer 1
- Person 1 
- Person 2
- Person 3
- ...

This created a misalignment between how we store the details for the person, vs how we display it to the user, which didn't make much sense for bulk upload.

Now we refer to people in the household, if it's a joint purchase:
- Buyer 1
- Buyer 2
- Person 3 
- Person 4
- Person 5
- ...

If it's a non-joint purchase:
- Buyer 1
- Person 2
- Person 3 
- Person 4
- Person 5
- ...

Now the `details_known_` indexes align with the other field indexes in the database.
For example, `details_known_3` means that `age3`, `ecstat3` etc questions will be asked. 
That means that we no longer have `details_known_1` and only ask `details_known_2` question if it's a non-joint purchase.
We also display the same pages for person questions for joint and non-joint purchases, as numbering is now the same for both.


db/schema.rb
```diff
@@ -502,10 +501,11 @@
     t.boolean "is_la_inferred"
     t.bigint "bulk_upload_id"
     t.integer "retirement_value_check"
-    t.integer "deposit_and_mortgage_value_check"
```
ghost: did you mean to drop this column?

kosiakkatrina: Yeah, I think it was committed accidentally, cause it's nowhere else in the codebase

spec/features/form/check_answers_page_sales_logs_spec.rb
```diff
@@ -43,8 +43,10 @@
           visit("/sales-logs/#{completed_sales_log_joint_purchase.id}/#{subsection}/check-answers")
           assert_selector ".x-govuk-summary-card__title", text: "Buyer 1", count: 1
           assert_selector ".x-govuk-summary-card__title", text: "Buyer 2", count: 1
-          assert_selector ".x-govuk-summary-card__title", text: "Person 1", count: 1
+          assert_selector ".x-govuk-summary-card__title", text: "Person 1", count: 0
           assert_selector ".x-govuk-summary-card__title", text: "Person 2", count: 0
+          assert_selector ".x-govuk-summary-card__title", text: "Person 3", count: 1
+          assert_selector ".x-govuk-summary-card__title", text: "Person 4", count: 1
```
SamSeed-Softwire, line 51: Do we want to add a line for Person 5 here too?

spec/features/form/check_answers_page_sales_logs_spec.rb
```diff
@@ -53,8 +55,10 @@
           visit("/sales-logs/#{completed_sales_log_non_joint_purchase.id}/#{subsection}/check-answers")
           assert_selector ".x-govuk-summary-card__title", text: "Buyer 1", count: 1
           assert_selector ".x-govuk-summary-card__title", text: "Buyer 2", count: 0
-          assert_selector ".x-govuk-summary-card__title", text: "Person 1", count: 1
-          assert_selector ".x-govuk-summary-card__title", text: "Person 2", count: 0
+          assert_selector ".x-govuk-summary-card__title", text: "Person 1", count: 0
+          assert_selector ".x-govuk-summary-card__title", text: "Person 2", count: 1
+          assert_selector ".x-govuk-summary-card__title", text: "Person 3", count: 1
+          assert_selector ".x-govuk-summary-card__title", text: "Person 4", count: 1
```
SamSeed-Softwire, line 64: Do we want to add a line for Person 5 here too?

spec/models/form/sales/pages/person_age_spec.rb
```diff
@@ -20,171 +20,85 @@
...
-      it "has the correct id" do
-        expect(page.id).to eq("person_1_age")
-      end
-
-      it "has correct depends_on" do
-        expect(page.depends_on).to eq(
-          [{ "details_known_1" => 1, "jointpur" => 2 }],
-        )
-      end
+  context "and person 2" do
```
SamSeed-Softwire: I find the "and" at the start of this context description slightly weird (in a grammatical sense) because it's the outermost context block. What do you think?

spec/models/form/sales/pages/person_age_spec.rb
```diff
@@ -20,171 +20,85 @@
...
-
-      it "has correct depends_on" do
-        expect(page.depends_on).to eq(
-          [
-            { "details_known_4" => 1, "jointpur" => 2 },
-          ],
-        )
-      end
+    it "has the correct id" do
+      expect(page.id).to eq("person_3_age")
```
SamSeed-Softwire, line 51: Do we need to repeat these tests ("has correct questions", "has the correct id", and "has correct depends_on") for each person? I feel like these tests check whether the `PersonAge` question is correctly initialised given certain parameters, so if it works for one person it should work for all of them.

I don't think these tests are at all related to whether the number of people is valid (e.g. I can create a new context block which runs these tests for person_7 and it still passes absolutely fine). So going up to 5 people doesn't seem to make sense to me.

Lmk your thoughts!

spec/features/form/check_answers_page_sales_logs_spec.rb
SamSeed-Softwire: I would have assumed that Person 6 would be allowed to (I thought for a joint purchase there would be Buyer 1, Buyer 2, Person 3, Person 4, Person 5, Person 6) since you can have up to 4 non-buyers. However testing this locally leads to a test failure. Am I missing something?

SamSeed-Softwire (commented):
Not finished reviewing but thought it would be useful to share my thoughts so far!

spec/models/form/sales/pages/person_age_spec.rb
kosiakkatrina: That's a good point. For the first part, I think it's worth checking the values for each person, cause we don't know if they're hardcoded, maybe we don't need to check it for 4 different people, but it's only 3 assertions per context, so I wouldn't rush to delete them

spec/models/form/sales/pages/person_age_spec.rb
kosiakkatrina: Good spot!

spec/features/form/check_answers_page_sales_logs_spec.rb
kosiakkatrina: Added that

spec/features/form/check_answers_page_sales_logs_spec.rb
kosiakkatrina: Added person 5, person 6 fails for me too. But that's cause I don't think we're asking questions about person 6. I've added a field for it but the implementation might be part of 1847(?)

SamSeed-Softwire: The reason I thought 6 might be valid is because in `schema.rb` there are fields for the 6th person, for example these can be seen in the sales log table definition:
- t.integer "age6"
- t.integer "age6_known"
- t.integer "ecstat6"
- t.string "relat6"

spec/models/form/sales/pages/person_gender_identity_spec.rb
```diff
@@ -7,172 +7,85 @@
...
-        expect(page.questions.map(&:id)).to eq(%w[sex3])
-      end
+  it "has the correct description" do
+    expect(page.description).to be_nil
+  end
 
-      it "has the correct id" do
-        expect(page.id).to eq("person_2_gender_identity")
-      end
+  context "with person 2  " do
```
SamSeed-Softwire, line 24: Why aren't we doing these tests (has correct questions etc) for person 1 as well?

db/schema.rb
```diff
@@ -509,6 +508,9 @@
     t.integer "deposit_and_mortgage_value_check"
     t.integer "shared_ownership_deposit_value_check"
     t.integer "old_persons_shared_ownership_value_check"
+    t.integer "grant_value_check"
```
SamSeed-Softwire: `grant_value_check` also appears a few lines above here.

spec/models/form/sales/pages/person_known_spec.rb
```diff
@@ -21,34 +21,9 @@
       expect(page.description).to be_nil
```
SamSeed-Softwire, line 21: Not a big deal but for consistency you could standardise which person index you use in the initial section in this (and equivalent) files. I.e. before the first `context "with person 2" do` block. For example above `page_id` has index 2 but `person_index` is 3. Tests still pass but perhaps a bit confusing?

spec/models/form/sales/pages/person_known_spec.rb
```diff
@@ -86,15 +65,27 @@
       end
 
       it "has correct depends_on" do
-        expect(page.depends_on).to eq(
-          [{ "hholdcount" => 3, "jointpur" => 2 }, { "hholdcount" => 4, "jointpur" => 2 }],
-        )
+        expect(page.depends_on).to eq([
+          {
+            "hholdcount" => {
```
SamSeed-Softwire, line 71: Again not a huge deal but the order of the items in these hashes are reversed compared with in `person_known.rb`. In there `jointpur` comes before `hholdcount` and `operator` comes before `operand`. Whereas here `jointpur` comes _after_ `hholdcount` and `operator` comes _after_ `operand`. Probs not worth changing unless you think so.

SamSeed-Softwire (commented):
A few more comments - will continue looking tomorrow :) Sorry I didn't finish today, been lots of meetings!

spec/models/form/sales/pages/person_gender_identity_spec.rb
kosiakkatrina: Cause person 1 would always be buyer, and we're using different classes for those questions

spec/models/form/sales/questions/person_age_known_spec.rb
```diff
@@ -35,375 +35,187 @@
...
-
-      it "has the correct check_answers_card_number" do
-        expect(question.check_answers_card_number).to eq(2)
-      end
+  context "with person 2" do
+    let(:question_id) { "age2_known" }
+    let(:person_index) { 2 }
+
+    before do
+      allow(page).to receive(:id).and_return("person_2_age")
```
SamSeed-Softwire: I'm not sure any of the `allow(page).to receive(:id).and_return...` lines on this page are necessary (can't see where they'd be used, and the tests still pass without them).

app/models/form/sales/questions/person_age_known.rb
```diff
@@ -1,24 +1,24 @@
...
     }
     @hidden_in_check_answers = {
       "depends_on" => [
         {
-          field_for_person("age", "_known") => 0,
+          "age#{person_index}_known" => 0,
         },
         {
-          field_for_person("age", "_known") => 1,
+          "age#{person_index}_known" => 1,
```
SamSeed-Softwire, line 17: I'm probably just misunderstanding `depends_on`, but what's the point in specifying that it's shown whether it's either 0 or 1? Is it because 0 and 1 are all the non-nil values we'd expect `age#{person_index}_known` to take (so we hide the 'is person X's age known' question from the CYA page unless it hasn't been answered)?

spec/models/form/sales/questions/person_age_spec.rb
```diff
@@ -40,263 +40,131 @@
...
+    end
 
-      it "has the correct inferred check answers value" do
-        expect(question.inferred_check_answers_value).to eq([{
-          "condition" => { "age5_known" => 1 },
-          "value" => "Not known",
-        }])
-      end
+    it "has the correct width" do
+      expect(question.width).to eq(3)
```
SamSeed-Softwire: There's an inconsistency here: some instance vars like `@header` are not dynamic and are only tested once, at the top of the file, whereas others like `@width` are also not dynamic but are tested for every person. Is there a reason for that?

spec/models/form/sales/questions/person_age_spec.rb
```diff
@@ -40,263 +40,131 @@
...
-          "condition" => { "age3_known" => 1 },
-          "value" => "Not known",
-        }])
-      end
-
-      it "has the correct check_answers_card_number" do
-        expect(question.check_answers_card_number).to eq(3)
-      end
+    before do
+      allow(page).to receive(:id).and_return("person_2_age")
```
SamSeed-Softwire: Again like with `spec/models/form/sales/questions/person_age_known_spec.rb`, is this redundant?

spec/models/form/sales/questions/person_gender_identity_spec.rb
```diff
@@ -37,259 +37,127 @@
...
-      end
+  context "when person 2" do
+    let(:question_id) { "sex2" }
+    let(:person_index) { 2 }
 
-      it "has the correct check_answer_label" do
-        expect(question.check_answer_label).to eq("Person 1’s gender identity")
-      end
+    before do
+      allow(page).to receive(:id).and_return("person_2_gender_identity")
```
SamSeed-Softwire: Again like with spec/models/form/sales/questions/person_age_known_spec.rb, is this redundant?

spec/models/form/sales/questions/person_gender_identity_spec.rb
```diff
@@ -37,259 +37,127 @@
...
-      end
+    it "has the correct check_answer_label" do
+      expect(question.check_answer_label).to eq("Person 2’s gender identity")
     end
 
-    context "and joint purchase" do
-      let(:question_id) { "sex4" }
-      let(:person_index) { 4 }
+    it "has expected check answers card number" do
+      expect(question.check_answers_card_number).to eq(2)
```
SamSeed-Softwire, line 49: The same test is included at the top of this file along with tests that are not dependent on the person index. Worth removing there (but keeping here)?

(I would have put a comment on the first example of this test in the file, but since you didn't change those lines I can't leave a comment there. Shout if you know a way around this!)

spec/models/form/sales/questions/person_known_spec.rb
```diff
@@ -35,287 +35,143 @@
...
-        expect(question.hidden_in_check_answers).to eq(
-          "depends_on" => [{ "details_known_2" => 1 }],
-        )
-      end
-
-      it "has the correct check_answers_card_number" do
-        expect(question.check_answers_card_number).to eq(3)
-      end
+    before do
+      allow(page).to receive(:id).and_return("person_2_known")
```
SamSeed-Softwire: Again like with spec/models/form/sales/questions/person_age_known_spec.rb, is this redundant?

spec/models/form/sales/questions/person_known_spec.rb
```diff
@@ -35,287 +35,143 @@
...
-      end
+    it "has the correct check_answer_label" do
+      expect(question.check_answer_label).to eq("Details known for person 2?")
+    end
 
-      it "has correct conditional for" do
-        expect(question.conditional_for).to be_nil
-      end
+    it "has correct conditional for" do
+      expect(question.conditional_for).to be_nil
```
SamSeed-Softwire: Do we need to check this is nil for every person? That hasn't been done on other questions where `conditional_for` isn't defined.

spec/models/form/sales/questions/person_relationship_to_buyer1_spec.rb
```diff
@@ -37,259 +37,127 @@
...
-      end
+    it "has the correct check_answer_label" do
+      expect(question.check_answer_label).to eq("Person 2’s relationship to Buyer 1")
     end
 
-    context "and joint purchase" do
-      let(:question_id) { "relat4" }
-      let(:person_index) { 4 }
+    it "has expected check answers card number" do
+      expect(question.check_answers_card_number).to eq(2)
```
SamSeed-Softwire, line 53: Like in `spec/models/form/sales/questions/person_gender_identity_spec.rb`, we probably want to keep this but remove the duplicate test further up this file?

spec/models/form/sales/questions/person_relationship_to_buyer1_spec.rb
```diff
@@ -37,259 +37,127 @@
...
-      end
+  context "when person 2" do
+    let(:question_id) { "relat2" }
+    let(:person_index) { 2 }
 
-      it "has the correct check_answer_label" do
-        expect(question.check_answer_label).to eq("Person 1’s relationship to Buyer 1")
-      end
+    before do
+      allow(page).to receive(:id).and_return("person_2_relationship_to_buyer_1")
```
SamSeed-Softwire: Again like with spec/models/form/sales/questions/person_age_known_spec.rb, is this redundant?

spec/models/form/sales/questions/person_working_situation_spec.rb
```diff
@@ -9,7 +9,7 @@
   let(:person_index) { 2 }
 
   before do
-    allow(page).to receive(:id).and_return("person_1_working_situation")
+    allow(page).to receive(:id).and_return("person_2_working_situation")
```
SamSeed-Softwire: Again like with `spec/models/form/sales/questions/person_age_known_spec.rb`, is this redundant?

spec/models/form/sales/questions/person_working_situation_spec.rb
```diff
@@ -44,259 +44,127 @@
...
-        expect(question.check_answers_card_number).to eq(3)
-      end
-
-      it "has the correct inferred_check_answers_value" do
-        expect(question.inferred_check_answers_value).to eq([
-          { "condition" => { "ecstat3" => 10 }, "value" => "Prefers not to say" },
-        ])
-      end
+    it "has the correct check_answer_label" do
+      expect(question.check_answer_label).to eq("Person 2’s working situation")
```
SamSeed-Softwire, line 52: Again like in `spec/models/form/sales/questions/person_gender_identity_spec.rb`, we probably want to keep this but remove the duplicate test further up this file?

SamSeed-Softwire (commented):
Ok, reviewed everything now, phew!

app/models/form/sales/questions/person_age_known.rb
kosiakkatrina: That's exactly it!

spec/models/form/sales/questions/person_age_known_spec.rb
kosiakkatrina: Yup, they shouldn't be there, leftover from us checking the page IDs to determine if it's a joint purchase or not. Will remove!

spec/models/form/sales/questions/person_age_spec.rb
kosiakkatrina: Yup, that's just a mistake

spec/models/form/sales/questions/person_known_spec.rb
kosiakkatrina: Nope! Will extract

SamSeed-Softwire (approved):
Nice one Kat, sorry it took ages for me to review this. Would recommend removing the extra tests as mentioned here for example https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/1249#discussion_r1093374499 but that's personal preference.

spec/models/validations/shared_validations_spec.rb
```diff
@@ -67,6 +67,25 @@
         shared_validator.validate_numeric_min_max(record)
         expect(record.errors["age6"]).to be_empty
       end
+
+      context "with sales log" do
+        it "validates that person 2's age is between 16 and 120 for non joint purchase" do
+          sales_record.jointpur = 2
+          sales_record.hholdcount = 1
+          sales_record.details_known_2 = 1
+          sales_record.jointmore = 1
```
SamSeed-Softwire: I haven't seen `jointmore` before but looking at `app/models/form/sales/questions/number_joint_buyers.rb` it seems that it refers to the question "Are there more than 2 joint buyers of this property?". `jointmore = 1` would mean that there are more than 2 joint buyers. However just a few lines above we have that `jointpur = 2` i.e. it's not a joint purchase. Why is this?

SamSeed-Softwire: P.s. the tests still pass if you comment out the `jointmore = 1` line.

spec/models/validations/shared_validations_spec.rb
```diff
@@ -67,6 +67,25 @@
...
+          sales_record.age2 = 130
+          shared_validator.validate_numeric_min_max(sales_record)
+          expect(sales_record.errors["age2"].first).to eq("Person 2’s age must be between 0 and 110")
```
SamSeed-Softwire, line 78: The context description doesn't match what this test is actually doing (age range is incorrect).

spec/models/validations/shared_validations_spec.rb
```diff
@@ -67,6 +67,25 @@
...
+        end
+
+        it "validates that buyer 2's age is between 16 and 120 for joint purchase" do
+          sales_record.jointpur = 1
+          sales_record.age2 = 130
+          shared_validator.validate_numeric_min_max(sales_record)
+          expect(sales_record.errors["age2"].first).to eq("Buyer 2’s age must be between 0 and 110")
```
SamSeed-Softwire, line 85: Same here - the context description doesn't match what this test is actually doing (age range is incorrect).

app/models/validations/shared_validations.rb
```diff
@@ -18,7 +18,7 @@ def validate_other_field(record, value_other = nil, main_field = nil, other_fiel
   def validate_numeric_min_max(record)
     record.form.numeric_questions.each do |question|
       next unless question.min || question.max
-      next unless record[question.id]
+      next unless record[question.id] && question.page.routed_to?(record, nil)
```
SamSeed-Softwire, line 21: This does fix the issue identified in the PO review but I can't work out why... AFAICT the person 2 age / buyer 2 age questions are distinct (different `Pages`sub class, different `Questions` subclass, different fields in the db) so I'm not sure why there would be a bleedover between the two.

kosiakkatrina: So the field in the db is the same (age2), and we're adding the error to the field. So the error would always be added to the same field, but we'd use a different page/question to determine the wording (buyer vs person), depending whether it is routed to

spec/models/validations/shared_validations_spec.rb
kosiakkatrina: Ah good catch, will remove

spec/models/validations/shared_validations_spec.rb
kosiakkatrina: Hmm, I can change it, but it sounds like it is doing that, I guess it depends on how it's read 😅

kosiakkatrina: Oh, sorry, the range! Good spot!

app/models/validations/shared_validations.rb
SamSeed-Softwire: Ah gotcha - I think I got confused between different page ids/headers. I can see that the `PersonAge` question gets instantiated (in `app/models/form/sales/pages/person_age.rb`) with `id = "age#{@person_index}"`, which matches the id in the `Age2` question class.

PR #1251, opened 2023-01-30
CLDC-1828 amend copy on soft validation screen for rent value

arfacamble:
[TICKET](https://digital.dclg.gov.uk/jira/browse/CLDC-1828)

***

image

### DIFF
--- a/config/forms/2021_2022.json
+++ b/config/forms/2021_2022.json
@@ -8411,7 +8411,7 @@
                 }
               ],
               "title_text": {
-                "translation": "soft_validations.rent.min.title_text",
+                "translation": "soft_validations.rent.outside_range_title",
                 "arguments": [
                   {
                     "key": "brent",
@@ -8421,7 +8421,7 @@
                 ]
               },
               "informative_text": {
-                "translation": "soft_validations.rent.min.hint_text",
+                "translation": "soft_validations.rent.min_hint_text",
                 "arguments": [
                   {
                     "key": "soft_min_for_period",
@@ -8463,7 +8463,7 @@
                 }
               ],
               "title_text": {
-                "translation": "soft_validations.rent.max.title_text",
+                "translation": "soft_validations.rent.outside_range_title",
                 "arguments": [
                   {
                     "key": "brent",
@@ -8473,7 +8473,7 @@
                 ]
               },
               "informative_text": {
-                "translation": "soft_validations.rent.max.hint_text",
+                "translation": "soft_validations.rent.max_hint_text",
                 "arguments": [
                   {
                     "key": "soft_max_for_period",
--- a/config/forms/2022_2023.json
+++ b/config/forms/2022_2023.json
@@ -8400,7 +8400,7 @@
                     ]
                   },
                   "header": "This rent is lower than expected for this property type, in this area. Check:",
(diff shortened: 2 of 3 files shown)


config/locales/en.yml
```diff
@@ -436,12 +436,9 @@ en:
       in_soft_max_range:
         message: "Net income is higher than expected based on the lead tenant’s working situation. Are you sure this is correct?"
     rent:
-      min:
-        title_text: "You told us the rent is %{brent}"
-        hint_text: "The minimum rent expected for this type of property in this local authority is £%{soft_min_for_period}."
-      max:
-        title_text: "You told us the rent is %{brent}"
-        hint_text: "The maximum rent expected for this type of property in this local authority is £%{soft_max_for_period}."
+      outside_range_title: "You told us the rent is %{brent}"
```
arfacamble, line 439: the title text for min and max are the same, so I just made them both into the same translation in the yml rather than have two identical lines.

config/locales/en.yml
```diff
@@ -436,12 +436,9 @@ en:
...
+      min_hint_text: "The minimum rent expected for this type of property in this local authority is £%{soft_min_for_period}."
+      max_hint_text: "The maximum rent expected for this type of property in this local authority is £%{soft_max_for_period}."
```
arfacamble, line 441: I had removed these, but I reinstated them as they are still referenced in the 21/22 form. I don't have a great grip on what happens in terms of moving between different years.
I would slightly prefer these lines to no longer exist if they are no longer used. Alternative options:
- allow the 21/22 form to be broken because no one sees it any more?
- update the 21/22 form to the new copy
- remove these strings from the yml and just put them directly in the json, I assume I can't actually do this since they take variables which are retrieved on the step in between the json and the yml

config/forms/2022_2023.json
```diff
@@ -8460,7 +8443,8 @@
                       }
                     ]
                   },
-                  "header": "Are you sure this is correct?",
+                  "header": "This rent is higher than expected for this property type, in this area. Check:",
+                  "hint_text": "<ul><li>the decimal point is not missing (£X.XX)</li><li>the frequency is correct, for example weekly, monthly</li><li>the rent type is correct, for example affordable or social rent</li></ul><p>Once checked, confirm below.</p>",
```
arfacamble: as this is identical to the string above, I would ideally like this to be in the yml file. For design reasons it should be the hint text though, so as I understand it it will just have to be here as hint text can only be a plain string at this stage.

config/forms/2021_2022.json
```diff
@@ -8411,7 +8411,7 @@
                 }
```
arfacamble, line 8411: further comments below on the yml file. But can the changes I make to 21/22 mean anythign any more?

config/forms/2022_2023.json
```diff
@@ -8437,16 +8429,7 @@
...
-                "translation": "soft_validations.rent.max.hint_text",
-                "arguments": [
-                  {
-                    "key": "soft_max_for_period",
-                    "label": false,
-                    "i18n_template": "soft_max_for_period"
-                  }
-                ]
-              },
+              "informative_text": {},
```
kosiakkatrina, line 8432: I know we spoke about this, but why did we decide to put it in the hint_text vs informative_text?

kosiakkatrina: Ah, nevermind, I just looked at the designs

kosiakkatrina (approved):

PR #1252, opened 2023-01-30
CLDC-1781 Bulk upload thresholds

asmega:
# Context

- https://digital.dclg.gov.uk/jira/browse/CLDC-1781
- When bulk uploading logs we only want to create logs when a certain number pass validation

# Changes

- Prevent log creation if there are more than 16 errors in a column
- Only create logs if error rate is below a certain percentage threshold, this has been set to 60%
- There is also a final check to ensure `some_log.valid?` is true so no log from the bulk upload is lost


jamesrose (approved):

jamesrose:
Nice!

PR #1253, opened 2023-01-30
CLDC-1923 amend validation message for single ended ranges

arfacamble:


app/models/validations/sales/household_validations.rb
```diff
@@ -5,7 +5,7 @@ def validate_number_of_other_people_living_in_the_property(record)
     return if record.hholdcount.blank?
 
     unless record.hholdcount >= 0 && record.hholdcount <= 4
-      record.errors.add :hholdcount, I18n.t("validations.numeric.valid", field: "Number of other people living in the property", min: 0, max: 4)
+      record.errors.add :hholdcount, I18n.t("validations.numeric.within_range", field: "Number of other people living in the property", min: 0, max: 4)
     end
   end
```
arfacamble: Can anyone see any reason why I should not remove this validation? I assume this was done before, or in innocent ignorance of, the validate_numeric_min_max method.

app/models/validations/shared_validations.rb
```diff
@@ -20,20 +20,16 @@ def validate_numeric_min_max(record)
       next unless question.min || question.max
       next unless record[question.id]
 
-      field = question.check_answer_label || question.id
-      min = [question.prefix, number_with_delimiter(question.min, delimiter: ","), question.suffix].join("")
-      max = [question.prefix, number_with_delimiter(question.max, delimiter: ","), question.suffix].join("")
-
       begin
         answer = Float(record.public_send("#{question.id}_before_type_cast"))
```
arfacamble, line 24: I couldn't actually tell what this is doing. It doesn't stop me from being able to do the work, but I would be interested.
My best guess is that it's a funny kind of way of checking the type? We get the value before deserializing and then see if we can convert it to a float. That implies we have questions with a min max where the db type is not numeric???

app/models/validations/shared_validations.rb
```diff
@@ -108,4 +104,18 @@ def shared_validate_partner_count(record, max_people)
...
+    min = [question.prefix, number_with_delimiter(question.min, delimiter: ","), question.suffix].join("") if question.min
+    max = [question.prefix, number_with_delimiter(question.max, delimiter: ","), question.suffix].join("") if question.max
+
+    if min && max
+      record.errors.add question.id.to_sym, I18n.t("validations.numeric.within_range", field:, min:, max:)
+    elsif min
+      record.errors.add question.id.to_sym, I18n.t("validations.numeric.above_min", field:, min:)
+    else
+      record.errors.add question.id.to_sym, I18n.t("validations.numeric.below_max", field:, max:)
+    end
```
arfacamble, line 117: I am not actually aware of a situation where there is a max but not a min after global searching `@max = ` and `"max": ` Is there any other way than those?
In which case I'm inclined to remove the else but keep the previous condition as elsif min
Agree?

config/locales/en.yml
```diff
@@ -131,7 +131,9 @@ en:
     other_field_missing: "If %{main_field_label} is other then %{other_field_label} must be provided"
     other_field_not_required: "%{other_field_label} must not be provided if %{main_field_label} was not other"
     numeric:
-      valid: "%{field} must be between %{min} and %{max}"
+      within_range: "%{field} must be between %{min} and %{max}"
+      above_min: "%{field} must be at least %{min}"
+      below_max: "%{field} must be at most %{max}"
```
arfacamble: as above, this should probably be removed

app/models/validations/sales/household_validations.rb
natdeanlewissoftwire: fine by me, you'll need to add the min and max to the question object instead of course (and tests) but I agree that's cleaner

app/models/validations/shared_validations.rb
natdeanlewissoftwire: I agree with your best guess - I don't know what question this specific behaviour was needed for but yep seems like there is some question that has a numeric min/max but is cast to a string or something...

app/models/validations/shared_validations.rb
natdeanlewissoftwire: I don't think there's any other way - only possible exception I can think of are financial questions e.g. max salary that only display the max in an error message (rather than also saying min = 0). If that doesn't affect this then agree

natdeanlewissoftwire (approved):

PR #1255, opened 2023-01-30
CLDC 1865 Refactor questions to avoid race conditions

ghost:
https://digital.dclg.gov.uk/jira/browse/CLDC-1865

- Refactor StockOwner option question that it does not rely on instance variables
- Refactor ManagingOrganisation question so that it does not rely on instance variables
- Extract FormHandler methods in helper
- Add missing specs to BulkUploadController#start

### DIFF
--- a/app/models/form/lettings/questions/managing_organisation.rb
+++ b/app/models/form/lettings/questions/managing_organisation.rb
@@ -1,48 +1,45 @@
 class Form::Lettings::Questions::ManagingOrganisation < ::Form::Question
-  attr_accessor :current_user, :log
-
   def initialize(id, hsh, page)
     super
     @id = "managing_organisation_id"
     @check_answer_label = "Managing agent"
     @header = "Which organisation manages this letting?"
     @type = "select"
-    @answer_options = answer_options
   end
 
-  def answer_options
+  def answer_options(log = nil, user = nil)
     opts = { "" => "Select an option" }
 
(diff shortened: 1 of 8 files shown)


app/models/form/lettings/questions/managing_organisation.rb
```diff
@@ -1,48 +1,45 @@
 class Form::Lettings::Questions::ManagingOrganisation < ::Form::Question
```
jamesrose, line 1: We use `answer_options` in `FormController`, and likely a few other places. Do they need adjusting to inject the user and log?

spec/requests/bulk_upload_controller_spec.rb
```diff
@@ -50,6 +58,40 @@
...
+      before do
+        Timecop.freeze(time)
+        get start_bulk_upload_lettings_logs_path
+      end
+
+      after do
+        Timecop.unfreeze
+      end
+
+      context "when not crossover period" do
```
asmega, line 71: there should be coverage in `spec/features/bulk_upload_lettings_logs_spec.rb` but I have no problems with extra coverage

asmega (approved):

app/models/form/lettings/questions/managing_organisation.rb
ghost: no need for that as both parameter values default to nil, thus calling `answer_options` in FormController will just work

PR #1262, opened 2023-02-01
CLDC-1775 Bulk upload check your answers treatment

asmega:
# Context

- https://digital.dclg.gov.uk/jira/browse/CLDC-1775
- When checking answers for a bulk upload log, it get different treatment

# Changes

- When checking answers for a bulk upload log and the question is yet to be answered the copy is different and also coloured red
- `FilterService` class can now be instantiated. This is the start of having a better API for filtering and in my opinion beats having to include the `Modules::LogsFilter` which uses inheritance rather than composition

# Screenshots

Screenshot 2023-02-01 at 16 56 55


app/components/check_answers_summary_list_card_component.rb
```diff
@@ -1,10 +1,12 @@
 class CheckAnswersSummaryListCardComponent < ViewComponent::Base
-  attr_reader :questions, :log, :user
+  attr_reader :questions, :log, :user, :bulk_upload
 
-  def initialize(questions:, log:, user:)
+  def initialize(questions:, log:, user:, bulk_upload: nil)
     @questions = questions
     @log = log
     @user = user
+    @bulk_upload = bulk_upload
```
kosiakkatrina: Not against doing it like this, but checking if @log has bulk_upload_id might mean that we pass in less arguments here

kosiakkatrina (approved):

app/components/check_answers_summary_list_card_component.rb
asmega: The behaviour would be different if it were implemented that way. Here the red version is only shown when the user is filtering logs via bulk upload. In your implementation it would always display the red version if a log was uploaded via bulk upload regardless if they were filtering by bulk upload. I will check in with designers to see which it is they are after

jamesrose (approved):

app/components/check_answers_summary_list_card_component.rb
asmega: Looks they were after what you suggested. I'll update this PR later to reflect that. I might keep the filtering stuff though and use it elsewhere as think that's worth keeping

asmega: I've tweaked this now, so it's always red if the log was bulk uploaded and ignores any filtering the user has chosen

PR #1263, opened 2023-02-02
CLDC-1911 Initialise missing forms using the new format

kosiakkatrina:
Initialise missing forms using the new format (non JSON)
Disable 14 days validation on non production
Do not use the next lettings form in validations and as a bulk upload option

### DIFF
--- a/app/models/form.rb
+++ b/app/models/form.rb
@@ -7,7 +7,7 @@ class Form
     if type == "sales" || (start_year && start_year.to_i > 2022)
       @setup_sections = type == "sales" ? [Form::Sales::Sections::Setup.new(nil, nil, self)] : [Form::Lettings::Sections::Setup.new(nil, nil, self)]
       @form_sections = sections_in_form.map { |sec| sec.new(nil, nil, self) }
-      @type = type == "sales" ? "sales" : "lettings"
+      @type = type
       @sections = setup_sections + form_sections
       @subsections = sections.flat_map(&:subsections)
       @pages = subsections.flat_map(&:pages)
--- a/app/models/form.rb
+++ b/app/models/form.rb
@@ -4,10 +4,10 @@ class Form
               :setup_sections, :form_sections, :unresolved_log_redirect_page_id
 
   def initialize(form_path, start_year = "", sections_in_form = [], type = "lettings")
-    if type == "sales"
-      @setup_sections = [Form::Sales::Sections::Setup.new(nil, nil, self)]
+    if type == "sales" || (start_year && start_year.to_i > 2022)
+      @setup_sections = type == "sales" ? [Form::Sales::Sections::Setup.new(nil, nil, self)] : [Form::Lettings::Sections::Setup.new(nil, nil, self)]
       @form_sections = sections_in_form.map { |sec| sec.new(nil, nil, self) }
-      @type = "sales"
+      @type = type == "sales" ? "sales" : "lettings"
       @sections = setup_sections + form_sections
       @subsections = sections.flat_map(&:subsections)
       @pages = subsections.flat_map(&:pages)
@@ -20,6 +20,7 @@ class Form
         "end_date" => end_date,
         "sections" => sections,
       }
+      @unresolved_log_redirect_page_id = "tenancy_start_date"
     else
       raise "No form definition file exists for given year".freeze unless File.exist?(form_path)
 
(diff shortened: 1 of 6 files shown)


app/models/form.rb
```diff
@@ -4,10 +4,10 @@ class Form
               :setup_sections, :form_sections, :unresolved_log_redirect_page_id
 
   def initialize(form_path, start_year = "", sections_in_form = [], type = "lettings")
-    if type == "sales"
-      @setup_sections = [Form::Sales::Sections::Setup.new(nil, nil, self)]
+    if type == "sales" || (start_year && start_year.to_i > 2022)
+      @setup_sections = type == "sales" ? [Form::Sales::Sections::Setup.new(nil, nil, self)] : [Form::Lettings::Sections::Setup.new(nil, nil, self)]
       @form_sections = sections_in_form.map { |sec| sec.new(nil, nil, self) }
-      @type = "sales"
+      @type = type == "sales" ? "sales" : "lettings"
```
```suggestion
      @type = type
```

?

app/models/form_handler.rb
```diff
@@ -41,6 +43,23 @@ def lettings_forms
...
+
+    if forms["previous_lettings"].blank? && current_collection_start_year >= 2022
+      forms["previous_lettings"] = Form.new(nil, current_collection_start_year - 1, lettings_sections, "lettings")
+    end
+    if forms["current_lettings"].blank? && current_collection_start_year >= 2022
+      forms["current_lettings"] = Form.new(nil, current_collection_start_year, lettings_sections, "lettings")
+    end
+    if forms["next_lettings"].blank? && current_collection_start_year >= 2022
+      forms["next_lettings"] = Form.new(nil, current_collection_start_year + 1, lettings_sections, "lettings")
+    end
```
jamesrose: Why do we need `current_collection_start_year >= 2022` conditions?

app/models/form.rb
```diff
@@ -4,10 +4,10 @@ class Form
               :setup_sections, :form_sections, :unresolved_log_redirect_page_id
 
   def initialize(form_path, start_year = "", sections_in_form = [], type = "lettings")
-    if type == "sales"
-      @setup_sections = [Form::Sales::Sections::Setup.new(nil, nil, self)]
+    if type == "sales" || (start_year && start_year.to_i > 2022)
+      @setup_sections = type == "sales" ? [Form::Sales::Sections::Setup.new(nil, nil, self)] : [Form::Lettings::Sections::Setup.new(nil, nil, self)]
```
jamesrose, line 8: I think if it were me I'd have three conditions here:

```ruby
if type == "sales"
 ..
elsif start_year == 2022

else # .. this one will be depreciated

end
```

I don't think you need to convert `start_year` to an integer because as far as I can see, it's either an empty string, or an integer year.

kosiakkatrina: "" > 2022 would fail with argument error, but if we're separating it into 3 ifs that wouldn't matter

app/models/form_handler.rb
kosiakkatrina: Cause if it's 2021 Form.new will try to create previous_lettings for 20/21 from a JSON that we don't have. Needs a refactor tho

jamesrose: Yes, so sorry, I didn't realise this was in draft. Safely ignore my comments until you're ready!

app/models/form.rb
jamesrose: Ah yes, you're completely right.

kosiakkatrina: Ready for review now, but not sure if it makes sense to have 2 ifs, cause the content is very similar

spec/models/form_handler_spec.rb
```diff
@@ -168,5 +168,44 @@
       end
     end
   end
+
+  describe "lettings_forms" do
+    context "when current and previous forms are defined in JSON (current collection start year before 2023)" do
+      it "creates a next_lettings form from ruby form objects" do
```
ghost, line 176: can you set the `let(:now)` block here so that it's easier to review/understand for the next person that has to work on this file in the future?

app/models/form.rb
```diff
@@ -4,10 +4,10 @@ class Form
               :setup_sections, :form_sections, :unresolved_log_redirect_page_id
 
   def initialize(form_path, start_year = "", sections_in_form = [], type = "lettings")
-    if type == "sales"
-      @setup_sections = [Form::Sales::Sections::Setup.new(nil, nil, self)]
+    if type == "sales" || (start_year && start_year.to_i > 2022)
```
ghost: can this check moved to a private method so it's easier to follow? 

something like:
```
def sales_after_2022?
   type == "sales" || (start_year && start_year.to_i > 2022)
end
```

ghost (approved):

PR #1265, opened 2023-02-02
CLDC-1900 improve expired join links

natdeanlewissoftwire:
Adds the following flow: 
image

ticket: https://digital.dclg.gov.uk/jira/browse/CLDC-1900

- [x] This ticket adds 2 new emails to notify as well - they should be renamed to remove the (dev) in their names once this is merged (and the old template marked as old)


app/controllers/auth/passwords_controller.rb
```diff
@@ -1,5 +1,6 @@
 class Auth::PasswordsController < Devise::PasswordsController
   include Helpers::Email
+  before_action :set_minimum_password_length, only: %i[edit create]
```
natdeanlewissoftwire: this shows the password hint text that was previously nil - added to this ticket as it's a small improvement, see ticket

app/models/user.rb
```diff
@@ -121,6 +124,7 @@ def send_confirmation_instructions
     return unless active?
 
     super
+    update!(initial_confirmation_sent: true)
```
natdeanlewissoftwire, line 127: not certain this is the best way of showing a different form when the user has already been invited before, happy to take suggestions if you can think of a better way!

app/controllers/auth/passwords_controller.rb
ghost: can't we directly access `Devise.password_length.min` rather than setting its value to a variable?

ghost (approved):

app/controllers/auth/passwords_controller.rb
natdeanlewissoftwire: Yep good point, was trying to avoid repetition but I've ended up just writing more code!

PR #1268, opened 2023-02-03
CLDC-1826 lettings log codes only download

arfacamble:
https://digital.dclg.gov.uk/jira/browse/CLDC-1826


db/seeds.rb
```diff
@@ -207,6 +207,16 @@
...
+    User.find_or_create_by!(
+      name: "Arthur",
+      email: "arthur.campbell@softwire.com",
+      organisation: org,
+      role: "support",
+    ) do |user|
+      user.password = "password"
+      user.confirmed_at = Time.zone.now
+    end
+
```
arfacamble: - [ ] needs reverting once testing on review app complete
- [ ] does Jemma need a review env user to test?

spec/models/lettings_log_spec.rb
```diff
@@ -2493,40 +2493,69 @@ def previous_postcode_fields
...
+        expected_content.sub!(/\{scheme_sensitive\}/, scheme["sensitive"].to_s)
+        expected_content.sub!(/\{scheme_primary_client_group\}/, scheme["primary_client_group"].to_s)
+        expected_content.sub!(/\{scheme_secondary_client_group\}/, scheme["secondary_client_group"].to_s)
+        expected_content.sub!(/\{scheme_support_type\}/, scheme["support_type"].to_s)
+        expected_content.sub!(/\{scheme_intended_stay\}/, scheme["intended_stay"].to_s)
+        expected_content.sub!(/\{location_code\}/, location["id"].to_s)
+        expected_content.sub!(/\{location_startdate\}/, location["startdate"].to_s)
+        expected_content.sub!(/\{scheme_id\}/, scheme["service_name"].to_s)
+        expected_content.sub!(/\{location_id\}/, location["id"].to_s)
+        expected_content.gsub!("\r\n", "\n")
```
arfacamble: - [x] I should discuss this and the fact that I am calling strip on everything below as I shouldn't need to have a fight with line endings. I found myself unable to force the pooter to save these csv files with LF line endings

spec/requests/lettings_logs_controller_spec.rb
```diff
@@ -523,54 +523,59 @@
...
             FactoryBot.create(:lettings_log, created_by: user, owning_organisation: user.organisation, tenancycode: "foo")
-            get "/lettings-logs?search=#{search_term}", headers: headers, params: {}
-            expect(page).to have_link("Download (CSV)", href: "/lettings-logs/csv-download?search=#{search_term}")
+            get "/lettings-logs?search=#{search_term}", headers:, params: {}
+            download_link = page.find_link("Download (CSV)")
+            download_link_params = CGI.parse(URI.parse(download_link[:href]).query)
+            codes_only_download_link = page.find_link("Download (CSV, codes only)")
+            codes_only_download_link_params = CGI.parse(URI.parse(codes_only_download_link[:href]).query)
+            expect(download_link_params).to include("search" => [search_term])
+            expect(codes_only_download_link_params).to include("search" => [search_term])
```
arfacamble: - [x] add test: includes codes_only=true/false in link
at this level of nesting/another/multiple?

spec/requests/lettings_logs_controller_spec.rb
```diff
@@ -1270,33 +1276,33 @@
...
       end
 
       it "passes a combination of search term and filter parameters" do
         postcode = "XX1 1TG"
 
         expect {
-          post "/lettings-logs/email-csv?status[]=completed&search=#{postcode}", headers:, params: {}
-        }.to enqueue_job(EmailCsvJob).with(user, postcode, { "status" => %w[completed] }, false)
+          post "/lettings-logs/email-csv?status[]=completed&search=#{postcode}&is_codes_only_export=false", headers:, params: {}
+        }.to enqueue_job(EmailCsvJob).with(user, postcode, { "status" => %w[completed] }, false, nil, false)
```
arfacamble, line 1419: - [x] add test that codes_only parameter is passed in isolation and in conjunction with other parameters

spec/requests/organisations_controller_spec.rb
```diff
@@ -1158,15 +1159,15 @@
...
 
           expect(page).to have_text("You've selected 3 logs.")
         end
 
         it "provides the organisation to the mail job" do
           expect {
-            post "/organisations/#{organisation.id}/logs/email-csv?status[]=completed", headers:, params: {}
-          }.to enqueue_job(EmailCsvJob).with(user, nil, { "status" => %w[completed] }, false, organisation)
+            post "/organisations/#{organisation.id}/logs/email-csv?status[]=completed&is_codes_only_export=false", headers:, params: {}
+          }.to enqueue_job(EmailCsvJob).with(user, nil, { "status" => %w[completed] }, false, organisation, false)
```
arfacamble, line 1170: organisations controller spec doesn't test download_csv or email_csv methods?
 - [x] check this
 - [x]  add provides codes_only flag to mail job

app/controllers/lettings_logs_controller.rb
```diff
@@ -80,13 +80,15 @@ def destroy
 
   def download_csv
     unpaginated_filtered_logs = filtered_logs(current_user.lettings_logs, search_term, @session_filters)
+    is_codes_only_export = params.require(:codes_only) == "true"
 
-    render "download_csv", locals: { search_term:, count: unpaginated_filtered_logs.size, post_path: email_csv_lettings_logs_path }
+    render "download_csv", locals: { search_term:, count: unpaginated_filtered_logs.size, post_path: email_csv_lettings_logs_path, is_codes_only_export: }
```
arfacamble: write tests for this method in controller
 - [x] parses is_codes_only_export correctly
 - [x] displays page with eg button

db/seeds.rb
kosiakkatrina: do we send emails on review? If so, can a user be created by sending an invite by a support user?

.github/workflows/review_pipeline.yml
```diff
@@ -120,7 +120,7 @@ jobs:
           cf set-env $APP_NAME IMPORT_PAAS_INSTANCE $IMPORT_PAAS_INSTANCE
           cf set-env $APP_NAME EXPORT_PAAS_INSTANCE $EXPORT_PAAS_INSTANCE
           cf set-env $APP_NAME S3_CONFIG $S3_CONFIG
-          cf set-env $APP_NAME CSV_DOWNLOAD_PAAS_INSTANCE "dluhc-core-review-csv-bucket"
+          cf set-env $APP_NAME CSV_DOWNLOAD_PAAS_INSTANCE "dluhc-core-staging-csv-bucket"
```
kosiakkatrina, line 126: Do we want to change it?

asmega: probably not. can we get by and not change it? you tested it with the correct bucket and everything worked then i don't see why we need to make this change. it'll probably cause problems elsewhere if we change this now

app/services/csv/lettings_log_csv_service.rb
```diff
@@ -7,35 +7,45 @@ def initialize(user)
...
+      elsif is_codes_only_export && att.start_with?("location_", "scheme_")
+        att += "_before_type_cast" unless %w[location_code scheme_code scheme_owning_organisation_name scheme_created_at location_startdate].include? att
+        record.send(att)
       else
-        record.form.get_question(att.remove("_label"), record)&.label_from_value(record.send(att.remove("_label"))) || label_from_boolean_value(record.send(att.remove("_label")))
+        att = att.remove("_label", "_detail") # a couple of csv column headers have suffixes for the user that are not reflected in the app domain
+        field_value = record.send(att)
+        answer_label = record.form
+                             .get_question(att, record)
+                             &.label_from_value(field_value)
```
kosiakkatrina: It doesn't look like we need to execute this is it is a codes only export (?) can we bring it down into the if statement after line 40?

app/views/logs/_log_list.html.erb
```diff
@@ -1,7 +1,8 @@
 <h2 class="govuk-body">
   <%= render(SearchResultCaptionComponent.new(searched:, count: pagy.count, item_label:, total_count:, item: "logs", path: request.path)) %>
   <% if logs&.first&.lettings? %>
-    <%= govuk_link_to "Download (CSV)", csv_download_url, type: "text/csv" %>
+    <%= govuk_link_to "Download (CSV)", csv_download_url, type: "text/csv", class: "govuk-!-margin-right-4" %>
+    <%= govuk_link_to "Download (CSV, codes only)", csv_codes_only_download_url, type: "text/csv" %>
```
kosiakkatrina: would this display for all users (providers and coordinators etc)?

spec/fixtures/files/lettings_logs_download.csv
```diff
@@ -1,2 +1,2 @@
...
+{id},in_progress,2022-02-08 16:52:15 +0000,2022-02-08 16:52:15 +0000,Danny Rojas,No,DLUHC,DLUHC,2021,Supported housing,,2 October 2021,London Affordable Rent,,,,,,,,,,,,,,,,,,,,No,,,,,No,Westminster,E09000033,,SE1 1TE,,No,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,2,8,0,0,0,,0,,,,,,,,,,,,,,,,,,,,,,,,0,,,,,,,0,,,,,,,,,,,,,,,,,,,9,1,,,,,,,,6,{scheme_code},{scheme_service_name},{scheme_sensitive},Missing,No,DLUHC,{scheme_primary_client_group},,{scheme_secondary_client_group},{scheme_support_type},{scheme_intended_stay},2021-04-01 00:00:00 +0100,{location_code},SE1 1TE,Downing Street,20,Bungalow,Fitted with equipment and adaptations,Westminster,{location_startdate}
```
kosiakkatrina: New line at the end of the file

spec/fixtures/files/lettings_logs_download_codes_only.csv
```diff
@@ -0,0 +1,2 @@
...
+{id},in_progress,2022-02-08 16:52:15 +0000,2022-02-08 16:52:15 +0000,Danny Rojas,false,DLUHC,DLUHC,2021,2,,2 October 2021,2,,,,,,,,,,,,,,,,,,,,false,,,,,false,E09000033,E09000033,,SE1 1TE,,2,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,2,8,0,0,0,,0,,,,,,,,,,,,,,,,,,,,,,,,0,,,,,,,0,,,,,,,,,,,,,,,,,,,9,1,,,,,,,,6,{scheme_code},{scheme_service_name},{scheme_sensitive},0,1,DLUHC,{scheme_primary_client_group},,{scheme_secondary_client_group},{scheme_support_type},{scheme_intended_stay},2021-04-01 00:00:00 +0100,{location_code},SE1 1TE,Downing Street,20,6,A,Westminster,{location_startdate}
```
kosiakkatrina: Same here

spec/models/form/sales/subsections/discounted_ownership_scheme_spec.rb
```diff
@@ -12,7 +12,6 @@
   end
 
   it "has correct pages" do
-    puts discounted_ownership_scheme.pages.map(&:id)
```
kosiakkatrina: nice

.github/workflows/review_pipeline.yml
arfacamble: yes this similar to seeds needs changing back after I have done final test on review app

app/models/derived_variables/lettings_log_variables.rb
```diff
@@ -1,4 +1,5 @@
 module DerivedVariables::LettingsLogVariables
+  # this is opaque enough to deserve a comment
```
arfacamble: @kosiakkatrina it sounded when we discussed that you had an idea of why this is this way. I feel this could do with a comment so at least hopefully the next person to engage with this code knows that we all have sympathy with them and the sinking feeling in their stomach

arfacamble: @kosiakkatrina it sounded when we discussed that you had an idea of why this is this way. I feel this could do with a comment so at least hopefully the next person to engage with this code knows that we all have sympathy with them and the sinking feeling in their stomach

app/views/logs/_log_list.html.erb
arfacamble: The ticket doesn't mention anything about this so I assumed that this type of download would be available with the same permissions as the old download

.github/workflows/review_pipeline.yml
arfacamble: yes I meant to chagne it back after final testing on review app

arfacamble: test

app/services/csv/lettings_log_csv_service.rb
arfacamble: good spot

db/seeds.rb
arfacamble: I am not aware. This was my first idea and it worked. I believe I ran it by James but I don't remember. As above, I intended to remove anyway once tested

spec/models/lettings_log_spec.rb
arfacamble: this has resolved itself

app/models/lettings_log.rb
```diff
@@ -427,15 +429,17 @@ def scheme_owning_organisation_name
...
+  delegate :location_admin_district_before_type_cast, to: :location, allow_nil: true
 
   # This is not the location_code in the db, location.id is just called code in the UI
   def location_code
     location&.id
   end
 
-  def self.to_csv(user = nil)
-    Csv::LettingsLogCsvService.new(user).to_csv
+  def self.to_csv(user = nil, is_codes_only_export:)
```
asmega: would you not default this to `false` to prevent breaking the existing API?

app/models/lettings_log.rb
```diff
@@ -416,7 +416,9 @@ def is_dpo
   end
 
   delegate :service_name, :sensitive, :registered_under_care_act, :primary_client_group, :has_other_client_group, :secondary_client_group, :owning_organisation, :managing_organisation, :support_type, :intended_stay, :created_at, prefix: "scheme", to: :scheme, allow_nil: true
+  delegate :service_name_before_type_cast, :sensitive_before_type_cast, :registered_under_care_act_before_type_cast, :primary_client_group_before_type_cast, :has_other_client_group_before_type_cast, :secondary_client_group_before_type_cast, :owning_organisation_before_type_cast, :managing_organisation_before_type_cast, :support_type_before_type_cast, :intended_stay_before_type_cast, :created_at_before_type_cast, prefix: "scheme", to: :scheme, allow_nil: true
```
asmega: if we take one of these eg `service_name_before_type_cast`, where is that used? I don't see the method being used anywhere? or have i missed it?

app/models/lettings_log.rb
arfacamble: I'm happy to add it here as well. This method is only called from within the EmailCsvJob as far as I can see and the parameter defaults to false there and is passed straight through. Let me know if you think it's best to add it here as well and if so why as it doesn't seem obvious to me I would want more default values.
For interest I would also be interested to discuss and understand ruby best practices with regards something in the EmailCsvJob. I will write something on that.

app/jobs/email_csv_job.rb
```diff
@@ -5,14 +5,14 @@ class EmailCsvJob < ApplicationJob
...
-  def perform(user, search_term = nil, filters = {}, all_orgs = false, organisation = nil) # rubocop:disable Style/OptionalBooleanParameter - sidekiq can't serialise named params
+  def perform(user, search_term = nil, filters = {}, all_orgs = false, organisation = nil, is_codes_only_export = false) # rubocop:disable Style/OptionalBooleanParameter - sidekiq can't serialise named params
     unfiltered_logs = organisation.present? && user.support? ? LettingsLog.where(owning_organisation_id: organisation.id) : user.lettings_logs
     filtered_logs = FilterService.filter_logs(unfiltered_logs, search_term, filters, all_orgs, user)
 
     filename = organisation.present? ? "logs-#{organisation.name}-#{Time.zone.now}.csv" : "logs-#{Time.zone.now}.csv"
 
     storage_service = Storage::S3Service.new(Configuration::EnvConfigurationService.new, ENV["CSV_DOWNLOAD_PAAS_INSTANCE"])
-    storage_service.write_file(filename, BYTE_ORDER_MARK + filtered_logs.to_csv(user))
+    storage_service.write_file(filename, BYTE_ORDER_MARK + filtered_logs.to_csv(user, is_codes_only_export:))
```
arfacamble: I would like to discuss and understand more regarding best practices in ruby/rails.
Here we call a method .to_csv on a collection of logs and following through in the debugger we end up in a method_missing and then it works out that we would like to call the to_csv method on each log individually.

arfacamble: @asmega this is what I meant in the other comment I replied to

app/services/csv/lettings_log_csv_service.rb
```diff
@@ -7,35 +7,45 @@ def initialize(user)
...
 
-    def label_from_value(record, att)
+    def label_from_value(record, att, is_codes_only_export:)
       if %w[la prevloc].include? att
-        label_from_boolean_value(record.send(att))
+        record.send(att)
       elsif %w[mrcdate startdate voiddate].include? att
         record.send(att)&.to_formatted_s(:govuk_date)
+      elsif is_codes_only_export && att.start_with?("location_", "scheme_")
+        att += "_before_type_cast" unless %w[location_code scheme_code scheme_owning_organisation_name scheme_created_at location_startdate].include? att
```
arfacamble: @asmega  this is where we call those delgated methods.

app/models/lettings_log.rb
arfacamble: I have tagged you in a comment where they are called.

app/services/csv/lettings_log_csv_service.rb
jamesrose: What's the motivation for the second condition here excluding `location_code` and so on?

arfacamble: there is existing inconsistency in lettings_log as to how and to what these methods are delegated to.
location_cod, scheme_code and scheme_owning_organisation_name are method defined in lettings_log that call in turn location.id, scheme.id (with an 'S' on the start) and scheme_owning_organisation.name, so I have the choice of defining a _before_type_cast method in lettings_log that is doing something not the same as it looks to maintain some look of consistency in this file, or not add _before_type_cast to the method name to call conditionally.
The date ones may be worth discussing, I did not investigate that in full but preferred consistency with the existing implementation, but calling before typecast offset them by one hour to fail tests. From basic investigation this is time zone related, so it owuld be good to know what the general rule for handling time zones is in this project.

app/views/logs/_log_list.html.erb
arfacamble: this is now done

app/views/logs/_log_list.html.erb
```diff
@@ -1,7 +1,10 @@
 <h2 class="govuk-body">
   <%= render(SearchResultCaptionComponent.new(searched:, count: pagy.count, item_label:, total_count:, item: "logs", path: request.path)) %>
   <% if logs&.first&.lettings? %>
-    <%= govuk_link_to "Download (CSV)", csv_download_url, type: "text/csv" %>
+    <%= govuk_link_to "Download (CSV)", csv_download_url, type: "text/csv", class: "govuk-!-margin-right-4" %>
+    <% if @current_user.support? %>
```
jamesrose, line 5: Is there a test for this?

app/models/derived_variables/lettings_log_variables.rb
kosiakkatrina: So I think the column that CDS cares about is `renttype`, which has these options:
```
    "1" => "Social Rent" 
    "2" => "Affordable Rent" 
    "3" => "Intermediate Rent"
```
 But we ask `rent_type` question which is something we defined and it has these options:
```
    "1" => "Affordable Rent" 
    "2" => "London Affordable Rent" 
    "4" => "London Living Rent" 
    "3" => "Rent to Buy" 
    "0" => "Social Rent" 
    "5" => "Other intermediate rent product" 
```

We could potentially add something like this here:
```
RENT_TYPE_MAPPING = { 
0 => 1, # "Social Rent"  =>  "Social Rent" 
1 => 2, # "Affordable Rent" => "Affordable Rent" 
2 => 2, # "London Affordable Rent"  =>  "Affordable Rent" 
3 => 3, # "Rent to Buy"  => "Intermediate Rent"
4 => 3, #  "London Living Rent"  => "Intermediate Rent"
5 => 3 # "Other intermediate rent product"  => "Intermediate Rent"
 }.freeze
```

app/jobs/email_csv_job.rb
arfacamble: @jamesrose 
this is the quirky thing I was referring to

app/services/csv/lettings_log_csv_service.rb
```diff
@@ -12,30 +13,126 @@ def to_csv
...
     end
 
   private
 
-    def label_from_value(record, att)
-      if %w[la prevloc].include? att
-        label_from_boolean_value(record.send(att))
-      elsif %w[mrcdate startdate voiddate].include? att
-        record.send(att)&.to_formatted_s(:govuk_date)
+    ATTRIBUTES_OF_RELATED_OBJECTS = {
```
arfacamble, line 23: does it matter where this is? ie down here or at the top of the class?

jamesrose (approved):

app/services/csv/lettings_log_csv_service.rb
jamesrose: The convention is at the top, yes.

app/controllers/organisations_controller.rb
```diff
@@ -107,12 +107,14 @@ def lettings_logs
   def download_csv
     organisation_logs = LettingsLog.all.where(owning_organisation_id: @organisation.id)
     unpaginated_filtered_logs = filtered_logs(organisation_logs, search_term, @session_filters)
+    codes_only = params.require(:codes_only) == "true"
 
-    render "logs/download_csv", locals: { search_term:, count: unpaginated_filtered_logs.size, post_path: logs_email_csv_organisation_path }
+    render "logs/download_csv", locals: { search_term:, count: unpaginated_filtered_logs.size, post_path: logs_email_csv_organisation_path, codes_only: }
   end
```
kosiakkatrina, line 113: Do we need to authorise for this too?

arfacamble: how did you write a comment on a line that hadn't been changed? I can't seem to be able to.

line 150 you have an existing authenticate_scope that covers this only being available for support users. This is fine since if you're not support you are directed through the lettings log controller anyway

app/services/csv/lettings_log_csv_service.rb
kosiakkatrina: What's the benefit of using this over delegating in the log?

app/controllers/organisations_controller.rb
kosiakkatrina: You can select several lines, but maybe because some of the lines I've selected have been changed that's why it let me comment on them

Nice

app/services/csv/lettings_log_csv_service.rb
arfacamble: the delegating was a little confusing already, and then in particular when we needed to get the codes instead of the human readable values there was an extra layer of weirdness on top because we had to call bla_before_type_cast on some of the values etc.
you can see at the below commit what it had become before moving to this way of doing.
https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/1268/commits/66f5f379fbc8d7fc1618847fbbf5f9c8f23fcb04

in general, it was decided that since there is some fairly case by case mapping between some of these things as they exist in the app and how they should be exported, it would be better to restrict that to this file and make it really explicit if overly verbose, than let it spill into the lettings log class

app/models/lettings_log.rb
```diff
@@ -418,27 +418,13 @@ def is_dpo
...
-    scheme_owning_organisation&.name
-  end
-
-  delegate :postcode, :name, :units, :type_of_unit, :mobility_type, :startdate, prefix: "location", to: :location, allow_nil: true
-  delegate :location_admin_district, to: :location, allow_nil: true
-
-  # This is not the location_code in the db, location.id is just called code in the UI
-  def location_code
-    location&.id
-  end
```
kosiakkatrina, line 445: Are we sure we're not using this somewhere else?

kosiakkatrina (approved):

app/models/lettings_log.rb
arfacamble: I remember cmd-shift-effing these things at the time so I am confident (never sure with the ol' meta-programming)

PR #1269, opened 2023-02-03
CLDC-1809 Add prevshared question

ghost:
this needs to be added to the new year form rather than the current ones https://digital.dclg.gov.uk/jira/browse/CLDC-1809

Ask this question if user owned previous property


app/models/form/sales/questions/prevshared.rb
```diff
@@ -0,0 +1,17 @@
+class Form::Sales::Questions::Prevshared < ::Form::Question
+  def initialize(id, hsh, page)
+    super
+    @id = "prevshared"
+    @check_answer_label = "Buyers previously owned a shared property"
```
natdeanlewissoftwire: should this be Buyer/Buyers depending if it's a joint purchase? or does this only occcur for shared ownerships already

natdeanlewissoftwire (approved):
lgtm, just one query

app/models/form/sales/questions/prevshared.rb
ghost: good point, I will figure that out!

app/models/form/sales/subsections/income_benefits_and_savings.rb
```diff
@@ -22,6 +22,15 @@ def pages
...
       Form::Sales::Pages::PreviousOwnership.new(nil, nil, self),
-    ]
+      previous_shared_page,
+    ].compact
+  end
+
+private
+
+  def previous_shared_page
+    if form.start_date.year >= 2023
```
natdeanlewissoftwire, line 32: should this be the `collection_start_year` rather than `start_date.year`?

natdeanlewissoftwire: ah wait no you're right I was thinking of `log.startdate`

PR #1270, opened 2023-02-06
Bulk upload validation tweaks

asmega:
# Context

- We should be persisting full validation error messages for bulk uploads
- We simply play these back to the user

# Changes

- In some places we were not persisting the full error message and instead just the error type
- These have now been changes to persist the full error message meaning correct error messages are now played back to the user rather than just messages that read `blank` or `invalid`

# Screenshots

## Before change with duplicates and incorrect message 

Screenshot 2023-02-06 at 10 49 39

## After change with dupes removed and full error messages

Screenshot 2023-02-06 at 10 51 23

### DIFF
--- a/app/services/bulk_upload/lettings/row_parser.rb
+++ b/app/services/bulk_upload/lettings/row_parser.rb
@@ -139,7 +139,8 @@ class BulkUpload::Lettings::RowParser
   attribute :field_133, :integer
   attribute :field_134, :integer
 
-  validates :field_1, presence: true, inclusion: { in: (1..12).to_a }
+  validates :field_1, presence: { message: I18n.t("validations.not_answered", question: "letting type") },
+                      inclusion: { in: (1..12).to_a, message: I18n.t("validations.invalid_option", question: "letting type") }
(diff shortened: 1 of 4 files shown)


app/services/bulk_upload/lettings/row_parser.rb
```diff
@@ -139,13 +139,10 @@ class BulkUpload::Lettings::RowParser
...
   attribute :field_134, :integer
 
-  validates :field_1, presence: true, inclusion: { in: (1..12).to_a }
+  validates :field_1, presence: { message: I18n.t("validations.not_answered", question: "letting type") },
+                      inclusion: { in: (1..12).to_a, message: I18n.t("validations.invalid_option", question: "letting type") }
   validates :field_4, presence: { if: proc { [2, 4, 6, 8, 10, 12].include?(field_1) } }
 
-  validates :field_96, presence: true
-  validates :field_97, presence: true
-  validates :field_98, presence: true
```
asmega, line 147: these validations are already present on the log and therefore causing duplicate errors to appear

ghost (approved):

PR #1272, opened 2023-02-07
CLDC-1939 Bugfix: Only add error if attribute is present

ghost:
Bugfix: Only add error if attribute is present

### DIFF
--- a/app/controllers/form_controller.rb
+++ b/app/controllers/form_controller.rb
@@ -71,7 +71,7 @@ private
     end
     if session["fields"]
       session["fields"].each do |field, value|
-        if form.get_question(field, @log)&.type != "date" && @log.respond_to?(field)
+        if form.get_question(field, @log)&.type != "date" && @log.attributes.key?(field)
           @log[field] = value
         end
       end


app/controllers/form_controller.rb
```diff
@@ -71,7 +71,7 @@ def restore_error_field_values
     end
     if session["fields"]
       session["fields"].each do |field, value|
-        if form.get_question(field, @log)&.type != "date" && @log.respond_to?(field)
+        if form.get_question(field, @log)&.type != "date" && @log.attributes.key?(field)
```
natdeanlewissoftwire, line 74: Will this fix the issue if there is an attribute in both sales and lettings logs e.g. age1? Couldn't there be an error added to age1 in a lettings log that would then appear on a sales log if the sales log age1 question was open in a different tab? Chance of this happening is small to be fair

natdeanlewissoftwire (commented):
(I'm assuming the tests turned out not to be worth adding time-wise)

natdeanlewissoftwire (approved):
Approving as a short term fix, can you add a ticket for what needs to be done long term Jack (you've probably got the most context around it)?

app/controllers/form_controller.rb
ghost: yes that is correct, we'd need to find a way to tell what type of form the user is updating

PR #1273, opened 2023-02-07
CLDC-1918 Add 2023/24 sales form

kosiakkatrina:
Add sales form for 23/24 as the next_sales form and remove the collection window validation on review and staging

### DIFF
--- a/app/models/form.rb
+++ b/app/models/form.rb
@@ -5,6 +5,8 @@ class Form
 
   def initialize(form_path, start_year = "", sections_in_form = [], type = "lettings")
     if sales_or_start_year_after_2022?(type, start_year)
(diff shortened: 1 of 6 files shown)


app/models/form.rb
```diff
@@ -5,15 +5,15 @@ class Form
 
   def initialize(form_path, start_year = "", sections_in_form = [], type = "lettings")
     if type == "sales"
+      @start_date = Time.zone.local(start_year, 4, 1)
+      @end_date = Time.zone.local(start_year + 1, 7, 1)
```
kosiakkatrina, line 9: Added this to the top, because we pass in `self` as a form to the sections when initializing them, and we would want the start and end date to be set before we do that

ghost (approved):

PR #1276, opened 2023-02-08
CLDC-1963 Stop password reset story turning into create password story

natdeanlewissoftwire:


app/views/devise/passwords/reset_password.html.erb
```diff
@@ -1,8 +1,4 @@
-<% content_for :title, @confirmation ? I18n.t("user.create_password") : I18n.t("user.reset_password") %>
-
-<% content_for :before_content do %>
-  <%= govuk_back_link(href: :back) %>
-<% end %>
```
natdeanlewissoftwire, line 5: I assume we don't want a back button - this page is reached externally from email links in both the create password and reset password stories

PR #1277, opened 2023-02-08
CLDC-1888 Check for existing duplicate logs when bulk uploading

SamSeed-Softwire:
## Context
- https://digital.dclg.gov.uk/jira/browse/CLDC-1888
- When doing a bulk upload, we want to check whether any logs already exist in the db which are duplicates of logs in the bulk upload file. We don't care about them being duplicated across _every_ field - just specific fields (these are given in the ticket linked above).

## The changes
- Added a validation in `app/services/bulk_upload/lettings/row_parser.rb` (will do for sales next) which:
  - Defines which fields in the log we care about checking.
  - Defines what fields in the bulk upload these map to (sometimes they map to a method which acts on 1 or more fields in the bulk upload).
  - Checks for any existing logs which match the current row being processed in the bulk upload (where matches are found based on the field mapping described above).
  - If any matches are found, an error is added to _all_ relevant fields in the bulk upload.

## Remaining work
- Confirm which email should get sent to users if a duplicate is detected, and plumb connection in so that the email actually gets sent.
- Potentially make the error message more descriptive.
- Repeat everything for sales (after confirming whether exactly the same fields should be used for matching).

### DIFF
--- a/app/services/bulk_upload/lettings/row_parser.rb
+++ b/app/services/bulk_upload/lettings/row_parser.rb
@@ -182,7 +182,7 @@ class BulkUpload::Lettings::RowParser
   end
 
   def log_already_exists?
-    fields_for_duplicity_check = %w(
+    fields_for_duplicity_check = %w[
       startdate
       postcode_full
       brent
@@ -194,9 +194,9 @@ class BulkUpload::Lettings::RowParser
       sex1
       ecstat1
       ethnic
-    )
+    ]
 
-    LettingsLog.exists?(Hash[fields_for_duplicity_check.collect { |field| [field, log[field]] }])
+    LettingsLog.exists?(fields_for_duplicity_check.index_with { |field| log[field] })
   end
 
 private
--- a/app/services/bulk_upload/lettings/row_parser.rb
+++ b/app/services/bulk_upload/lettings/row_parser.rb
@@ -329,7 +329,7 @@ private
       "ethnic" => field_43,
     }
 
-    LettingsLog.where(field_mappings).present?
+    LettingsLog.exists?(field_mappings)
   end
 
   def field_mapping_for_errors
--- a/app/services/bulk_upload/lettings/row_parser.rb
+++ b/app/services/bulk_upload/lettings/row_parser.rb
@@ -315,21 +315,21 @@ private
   end
 
   def log_already_exists?
-    field_mappings = {
-      "startdate" => startdate,
-      "postcode_full" => postcode_full,
-      "brent" => brent,
-      "scharge" => scharge,
-      "pscharge" => pscharge,
-      "supcharg" => supcharg,
-      "tenancycode" => field_7,
-      "age1" => field_12,
-      "sex1" => field_20,
-      "ecstat1" => field_35,
-      "ethnic" => field_43,
-    }
+    fields_for_duplicity_check = %w(
+      startdate
+      postcode_full
+      brent
+      scharge
+      pscharge
+      supcharg
+      tenancycode
+      age1
+      sex1
+      ecstat1
+      ethnic
+    )
 
-    LettingsLog.exists?(field_mappings)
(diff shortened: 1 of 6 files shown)


app/services/bulk_upload/lettings/row_parser.rb
```diff
@@ -804,6 +840,30 @@ def ethnic_group_from_ethnic
     end
   end
 
+  def brent
+    if %i[field_80 field_81 field_82 field_83].any? { |f| public_send(f).present? }
+      self.field_80 ||= 0
```
SamSeed-Softwire: I wrote this logic (and the equivalent logic for `scharge`, `pscharge` and `supcharg`) based on the derived variable logic in `app/models/derived_variables/lettings_log_variables.rb:23`. Would be interested to see if it could be done in a better way.

kosiakkatrina: I don't think we need the if statement, because we assume that these logs are completed, so the values in `field_80 field_81 field_82 field_83` should be what they're meant to be. Except if it's a care home, then these won't get set to 0 so they'd look like they're different...
What If you create a log in memory and run .valid? on it? And then use the fields from that log to check if an old log exists. That way all the values would get inferred and you wouldn't need to do it yourself

app/services/bulk_upload/lettings/row_parser.rb
```diff
@@ -230,6 +231,41 @@ def validate_nulls
...
+      "scharge" => scharge,
+      "pscharge" => pscharge,
+      "supcharg" => supcharg,
+      "tenancycode" => field_7,
+      "age1" => field_12,
+      "sex1" => field_20,
+      "ecstat1" => field_35,
+      "ethnic" => field_43,
+    }
+    duplicate_log_exists = LettingsLog.where(field_mappings).present?
```
ghost: this does what you need without pulling data from the database https://apidock.com/rails/ActiveRecord/Base/exists%3F/class

app/services/bulk_upload/lettings/row_parser.rb
SamSeed-Softwire: Hey @kosiakkatrina, thanks for this. I've tried implementing your idea doing something like the following: 

```
field_mappings = {
  "startdate" => startdate,
  "postcode_full" => postcode_full,
  "brent" => field_80,
  "scharge" => field_81,
  "pscharge" => field_82,
  "supcharg" => field_83,
  "tenancycode" => field_7,
  "age1" => field_12,
  "sex1" => field_20,
  "ecstat1" => field_35,
  "ethnic" => field_43,
}
log = LettingsLog.new(field_mappings)
log.valid?
field_mappings.update(field_mappings) { |key, value| log[key] } # I found this online as a way to easily reassign the values of the hash
duplicate_log_exists = LettingsLog.exists?(field_mappings)
```

However I'm having trouble - calling `log.valid?` seems to change `brent`, `scharge` etc from non-nil values to nil, and I'm not sure why. I thought it might be due to `app/models/lettings_log.rb#validate_rent_range` but this isn't getting triggered. Is it because I haven't filled in some essential fields for the log I'm creating in memory? If so, wouldn't I need to give values for almost _all_ the fields for the log I'm creating? (I tried to find which fields would be essential, and it seems to be most of them based on `app/models/lettings_log.rb#optional_fields`.) Specifying values for almost all the fields for the log I'm creating in memory doesn't feel like the right approach, but it might be!

app/services/bulk_upload/lettings/row_parser.rb
```diff
@@ -180,6 +181,24 @@ def log
...
+      pscharge
+      supcharg
+      tenancycode
+      age1
+      sex1
+      ecstat1
+      ethnic
+    ]
+
+    LettingsLog.exists?(fields_for_duplicity_check.index_with { |field| log[field] })
```
arfacamble, line 199: I don't think duplicity is the word here, unless you think the uploader is doing it on purpose to test us.
Maybe duplicate/replica

PR #1278, opened 2023-02-08
CLDC-1882 Bulk upload validate leaving reason if renewal

asmega:
# Context

- https://digital.dclg.gov.uk/jira/browse/CLDC-1882
- If the lettings is a renewal the the leaving reason should be checked so it makes sense for a renewal

# Changes

- If the lettings is a renewal then we check the leaving reason is limited a certain subset

# Screenshots

Screenshot 2023-02-08 at 13 02 44

### DIFF
--- a/app/services/bulk_upload/lettings/row_parser.rb
+++ b/app/services/bulk_upload/lettings/row_parser.rb
@@ -148,6 +148,7 @@ class BulkUpload::Lettings::RowParser
   validate :validate_relevant_collection_window
   validate :validate_la_with_local_housing_referral
   validate :validate_cannot_be_la_referral_if_general_needs
+  validate :leaving_reason_for_renewal
 
   def valid?
     errors.clear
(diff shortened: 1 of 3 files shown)


asmega:
I'm thinking should this validation belong as is just on bulk upload? or be moved to belong with all the other validations on `Log`?

app/services/bulk_upload/lettings/row_parser.rb
```diff
@@ -168,6 +169,14 @@ def log
 
 private
 
+  def leaving_reason_for_renewal
+    if field_134 == 1 && ![40, 42].include?(field_52)
+      message = 'The reason for leaving must be "End of assured shorthold tenancy - no fault" or "End of fixed term tenancy - no fault" if the letting is a renewal'
```
kosiakkatrina: Translations file? 👀

kosiakkatrina (approved):

app/services/bulk_upload/lettings/row_parser.rb
asmega: done

PR #1284, opened 2023-02-09
CLDC-1883 bulk upload validate LA and local housing referral

asmega:
# Context

- https://digital.dclg.gov.uk/jira/browse/CLDC-1883

# Changes

- Validate for bulk upload when owning organisation is an LA, the referral cannot be a PRP nominated by LA

# Notes for review

- There does already seem to be such an validation directly on the log. However I could never get this to fire for bulk upload. I believe this to be because the question is not routed to and therefore blanked
- We do still want this validation to appear though so therefore it has been attached as a validation thats exists on the bulk upload process

# Screenshots

Screenshot 2023-02-09 at 12 36 36

### DIFF
--- a/app/services/bulk_upload/lettings/row_parser.rb
+++ b/app/services/bulk_upload/lettings/row_parser.rb
@@ -171,7 +171,7 @@ private
 
   def validate_la_with_local_housing_referral
     if field_78 == 3 && owning_organisation && owning_organisation.la?
-      errors.add(:field_78, "The source of the referral cannot be Nominated by local housing authority as your organisation is a local authority")
+      errors.add(:field_78, I18n.t("validations.household.referral.nominated_by_local_ha_but_la"))
     end
   end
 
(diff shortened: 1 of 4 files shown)


app/services/bulk_upload/lettings/row_parser.rb
```diff
@@ -168,6 +169,12 @@ def log
 
 private
 
+  def validate_la_with_local_housing_referral
+    if field_78 == 3 && owning_organisation && owning_organisation.la?
+      errors.add(:field_78, "The source of the referral cannot be Nominated by local housing authority as your organisation is a local authority")
```
kosiakkatrina: Would it make sense to have this in translation file?

kosiakkatrina (approved):
I think this should be caught by the validation that checks if the selected question option exists, but we are indeed clearing it because we route to a different referral question version with different options...

app/services/bulk_upload/lettings/row_parser.rb
asmega: done

PR #1288, opened 2023-02-10
CLDC-1885 Bulk upload type must match given needs type

asmega:
# Context

- https://digital.dclg.gov.uk/jira/browse/CLDC-1885
- For the bulk upload journey users can select whether they are uploading for `general needs` or `supported housing`

# Changes

- When a user bulk uploads we ensure the data they upload matches that selected in the journey, otherwise return a validation error that there is a mismatch

# Screenshots

Screenshot 2023-02-10 at 16 08 36

### DIFF
--- a/app/services/bulk_upload/lettings/row_parser.rb
+++ b/app/services/bulk_upload/lettings/row_parser.rb
@@ -148,7 +148,8 @@ class BulkUpload::Lettings::RowParser
   validate :validate_relevant_collection_window
   validate :validate_la_with_local_housing_referral
   validate :validate_cannot_be_la_referral_if_general_needs
-  validate :leaving_reason_for_renewal
+  validate :validate_leaving_reason_for_renewal
+  validate :validate_lettings_type_matches_bulk_upload
 
   def valid?
     errors.clear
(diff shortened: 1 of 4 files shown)


app/services/bulk_upload/lettings/row_parser.rb
```diff
@@ -171,6 +172,16 @@ def log
 
 private
 
+  def validate_lettings_type_matches_bulk_upload
+    if [1, 3, 5, 7, 9, 11].include?(field_1) && bulk_upload.supported_housing?
```
kosiakkatrina: Is there no field for needs type specifically in the template?

config/locales/en.yml
```diff
@@ -170,6 +170,9 @@ en:
         invalid: "Please select owning organisation or managing organisation that you belong to"
       created_by:
         invalid: "Please select owning organisation or managing organisation that you belong to"
+      lettype:
+        general_needs_mismatch: Lettings type must be a general needs type
+        supported_housing_mismatch: Lettings type must be a supported housing type
```
kosiakkatrina: Do we want to update the content to say something like `...because you selected general needs before uploading a file` or something similar?

app/services/bulk_upload/lettings/row_parser.rb
kosiakkatrina: Also, would it make more sense to check that those aren't supported housing options as opposed to checking that they are general needs?

asmega: needs type is selected as part of the user journey when they select the file to upload. therefore we check to make sure the data from the csv matches the value they selected from the user journey. the needs type for the csv is derived from field 1 which is the `lettype`

asmega: > Also, would it make more sense to check that those aren't supported housing options as opposed to checking that they are general needs?

sure i can do that. don't think it will make any real difference though

asmega: I've now inverted the logic

config/locales/en.yml
asmega: i've now tweaked this

kosiakkatrina (approved):

PR #1292, opened 2023-02-13
CLDC-1968 Add review page for Sales logs

ghost:
- update link to point to sales log page
- update review template to also handle sales logs 
- hide empty sections in review template
- include factorybot in development so that we can use factories when testing locally

Can test here: https://dluhc-core-review-1292.london.cloudapps.digital/sales-logs/4/review?sales_log=true

Screenshot 2023-02-13 at 16 29 06

### DIFF
--- a/app/helpers/tasklist_helper.rb
+++ b/app/helpers/tasklist_helper.rb
@@ -39,7 +39,9 @@ module TasklistHelper
 
   def review_log_text(log)
     if log.collection_period_open?
-      "You can #{govuk_link_to 'review and make changes to this log', review_lettings_log_path(log)} until #{log.form.end_date.to_formatted_s(:govuk_date)}.".html_safe
+      link = log.is_a?(SalesLog) ? review_sales_log_path(id: log, sales_log: true) : review_lettings_log_path(log)
+
+      "You can #{govuk_link_to 'review and make changes to this log', link} until #{log.form.end_date.to_formatted_s(:govuk_date)}.".html_safe
     else
       "This log is from the #{log.form.start_date.year}/#{log.form.start_date.year + 1} collection window, which is now closed."
     end
--- a/app/helpers/tasklist_helper.rb
+++ b/app/helpers/tasklist_helper.rb
@@ -39,7 +39,7 @@ module TasklistHelper
 
   def review_log_text(log)
     if log.collection_period_open?
-      link = log.is_a?(SalesLog) ? review_sales_log_path(id: log, sales_log: true) : review_lettings_log_path(log)
+      link = log.sales? ? review_sales_log_path(id: log, sales_log: true) : review_lettings_log_path(log)
 
       "You can #{govuk_link_to 'review and make changes to this log', link} until #{log.form.end_date.to_formatted_s(:govuk_date)}.".html_safe
     else
--- a/app/models/log.rb
+++ b/app/models/log.rb
@@ -40,11 +40,11 @@ class Log < ApplicationRecord
   end
 
   def lettings?
-    is_a? LettingsLog
+    false
   end
 
   def sales?
-    is_a? SalesLog
+    false
   end
 
   def ethnic_refused?
--- a/app/models/log.rb
+++ b/app/models/log.rb
@@ -40,7 +40,11 @@ class Log < ApplicationRecord
   end
 
   def lettings?
-    false
+    is_a? LettingsLog
+  end
+
+  def sales?
+    is_a? SalesLog
   end
 
   def ethnic_refused?
--- a/app/views/form/review.html.erb
+++ b/app/views/form/review.html.erb
@@ -1,4 +1,4 @@
-<% if @log.is_a? SalesLog %>
+<% if @log.sales? %>
   <% content_for :title, "Review sales log" %>
   <% content_for :breadcrumbs, govuk_breadcrumbs(breadcrumbs: {
     "Logs" => "/logs",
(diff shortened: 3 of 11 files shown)


app/views/form/review.html.erb
```diff
@@ -1,9 +1,18 @@
...
+    "Review sales log" => "",
+  }) %>
+<% else %>
+  <% content_for :title, "Review lettings log" %>
+  <% content_for :breadcrumbs, govuk_breadcrumbs(breadcrumbs: {
+    "Logs" => "/logs",
+    "Log #{@log.id}" => "/lettings-logs/#{@log.id}",
+    "Review lettings log" => "",
+  }) %>
+<% end %>
```
jamesrose:
```suggestion
<% if @log.sales? %>
  <% content_for :title, "Review sales log" %>
  <% content_for :breadcrumbs, govuk_breadcrumbs(breadcrumbs: {
    "Logs" => "/logs",
    "Log #{@log.id}" => "/sales-logs/#{@log.id}",
    "Review sales log" => "",
  }) %>
<% else %>
  <% content_for :title, "Review lettings log" %>
  <% content_for :breadcrumbs, govuk_breadcrumbs(breadcrumbs: {
    "Logs" => "/logs",
    "Log #{@log.id}" => "/lettings-logs/#{@log.id}",
    "Review lettings log" => "",
  }) %>
<% end %>
```

We already have `Log#lettings?` and `LettingsLog#lettings?` — could you create the equivalent?

spec/requests/form_controller_spec.rb
```diff
@@ -314,6 +314,12 @@
           get "/lettings-logs/#{setup_complete_lettings_log.id}/review", headers: headers, params: {}
           expect(response.body).to match("Review lettings log")
         end
+
+        it "renders the review page for the sales log" do
+          log = create(:sales_log, :completed, created_by: user)
+          get "/sales-logs/#{log.id}/review", headers: headers, params: { sales_log: true }
```
asmega, line 320: are we passing query params to determine the log type? is the log type not already given earlier in the path eg `/sales-log` or `/lettings-log`?

app/helpers/tasklist_helper.rb
```diff
@@ -39,7 +39,9 @@ def subsection_link(subsection, log, current_user)
 
   def review_log_text(log)
     if log.collection_period_open?
-      "You can #{govuk_link_to 'review and make changes to this log', review_lettings_log_path(log)} until #{log.form.end_date.to_formatted_s(:govuk_date)}.".html_safe
+      link = log.is_a?(SalesLog) ? review_sales_log_path(id: log, sales_log: true) : review_lettings_log_path(log)
```
jamesrose:
```suggestion
      link = log.sales? ? review_sales_log_path(id: log) : review_lettings_log_path(log)
```

Why do you need the `sales_log: true` here?

natdeanlewissoftwire: sales?* presumably in the suggested improvement

jamesrose: @natdeanlewissoftwire Good spot! Thanks - updated my suggestion.

app/views/form/review.html.erb
ghost: missed that, will update

app/helpers/tasklist_helper.rb
ghost: https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/blob/a52b97b86629313a942fc599f8b560b30b8f180d/app/controllers/form_controller.rb#L110

I agree this is not ideal but I did not want to change this logic because I am not 100% it won't affect other parts of the application that might not be tested.

spec/requests/form_controller_spec.rb
ghost: https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/blob/a52b97b86629313a942fc599f8b560b30b8f180d/app/controllers/form_controller.rb#L110

I agree this isn't ideal but there's already some preexistent logic for this and I am not sure the form controller is covered well enough by tests to change that logic

app/models/log.rb
```diff
@@ -40,7 +40,11 @@ def collection_start_year
   end
 
   def lettings?
-    false
+    is_a? LettingsLog
```
jamesrose: What's the motivation for this change? I somewhat prefer the inheritance here vs type-checking, otherwise `Log` needs to have knowledge of its children.

ghost: The reasoning was that we wouldn't be calling this method on the Log object itself but on the children and having a method type check seemed a neater solution.
It is inherited by all the children classes and thought it would less error prone than just having a false true value set in a method like we had previously. Happy to switch to something like that if you prefer.

spec/requests/form_controller_spec.rb
asmega: sure, i feel like this will stick around and fester until it's addressed though

spec/models/sales_log_spec.rb
```diff
@@ -12,11 +12,16 @@
...
 
-  it "is a sales log" do
+  it "is a not a lettings log" do
     sales_log = build(:sales_log, created_by: created_by_user)
     expect(sales_log.lettings?).to be false
   end
 
+  it "is a sales log" do
+    sales_log = build(:sales_log, created_by: created_by_user)
+    expect(sales_log.sales?).to be true
```
asmega, line 22: you can also do
```ruby
expect(sales_log).to be_sales
expect(sales_log).not_to be_lettings

```

asmega (approved):

PR #1293, opened 2023-02-13
CLDC-1913 Change date question hint text

arfacamble:
https://digital.dclg.gov.uk/jira/browse/CLDC-1913

I have made an assumption about the form changing between years, I may be wrong in which case I will learn something I assume!


app/views/form/_date_question.html.erb
```diff
@@ -3,7 +3,7 @@
 <%= f.govuk_date_field question.id.to_sym,
   caption: caption(caption_text, page_header, conditional),
   legend: legend(question, page_header, conditional),
-  hint: { text: question.hint_text&.html_safe || "For example, 1 9 2022." },
+  hint: { text: question.hint_text&.html_safe || "For example, #{Date.today.to_formatted_s(:govuk_date_number_month)}" },
```
arfacamble: Do I understand correctly that the form rolls over each year automatically, so today's date should always be in the "current" tax year? Or is there a period in early April where today is not in the tax year of the form currently being filled out?
If the latter, what is it that defines the tax year of the form currently being worked on? If it's the sale completion date we will have a circular dependency here!

davidm-m: There's a period of overlap in the collection windows. The window for entering logs for a particular tax year is about 15 months, from 1st April to June/July of the next year. Which version of the form is used depends on which tax year the tenancy start date is in. For instance, if I'm filling out a log on 10th April 2023, but my tenancy start date is 10th *March* 2023, I should be shown the previous version of the form.
A lot of organisations (against DLUHC's advice) wait until the end of the year to submit all of their data, so there is a lot of data that is not submitted for the current tax year.

app/helpers/question_view_helper.rb
```diff
@@ -13,6 +13,18 @@ def legend(question, page_header, conditional)
     }
   end
 
+  def example_date_in_tax_year_of(date)
```
asmega: api of this method is a bit quirky. as `date` is a required argument if you want to call this without a date one would have to write
```ruby
example_date_in_tax_year_of(nil)
```

you probably want people to write
```ruby
example_date_in_tax_year_of
```

so signature might be better with something like the below
```ruby
def example_date_in_tax_year_of(date = nil)

OR

def example_date_in_tax_year_of(date = Time.zone.today)
```

asmega: just a minor thing for people who want to consume this api

app/helpers/question_view_helper.rb
```diff
@@ -13,6 +13,18 @@ def legend(question, page_header, conditional)
...
+    return Time.zone.today if date.nil?
+
+    year = if date.month > 4 || (date.month == 4 && date.day > 5)
+             date.year
+           else
+             date.year - 1
+           end
```
jamesrose: Are you trying to get the current collection year? If so, please us the methods in `CollectionTimeHelper`.

asmega (approved):

app/views/form/_date_question.html.erb
```diff
@@ -3,7 +3,7 @@
 <%= f.govuk_date_field question.id.to_sym,
   caption: caption(caption_text, page_header, conditional),
   legend: legend(question, page_header, conditional),
-  hint: { text: question.hint_text&.html_safe || "For example, 1 9 2022." },
+  hint: { text: question.hint_text&.html_safe || "For example, #{example_date_in_tax_year_of(lettings_log.startdate).to_formatted_s(:govuk_date_number_month)}" },
```
jamesrose:
```suggestion
  hint: { text: question.hint_text&.html_safe || "For example, #{example_date_in_collection_year_of(lettings_log.startdate)}" },
```

You've made a helper to do this, which is great, but you can help yourself even more by formatting the date within it :)

app/helpers/question_view_helper.rb
arfacamble: It wasn't my intention that anyone would want to call this method without a date to call it with. I onyl had it dealing with nils since it would have a log.startdate passed in which may be nil.
That was also why the method name ended with of, so that it read fairly semantically.
In any case for now this method is be delete since I didn't know about tax year vs collection year and already existsing CollecitonTimeHelper class

app/helpers/question_view_helper.rb
arfacamble: ta.

app/views/form/_date_question.html.erb
arfacamble: yes I have moved formatting into method in collecitontimehelper now.

PR #1300, opened 2023-02-14
CLDC-1903 Infer first time property let question

ghost:
To backfill the data, we will need to run the following script after deploy:

```
    LettingsLog.where(renewal: 1).update_all(first_time_property_let_as_social_housing: 0)
```

### DIFF
--- a/app/models/derived_variables/lettings_log_variables.rb
+++ b/app/models/derived_variables/lettings_log_variables.rb
@@ -46,6 +46,8 @@ module DerivedVariables::LettingsLogVariables
       self.referral = 1
       self.waityear = 2
       self.offered = 0
+      self.voiddate = startdate
+      self.first_time_property_let_as_social_housing = 0
       if is_general_needs?
         # fixed term
         self.prevten = 32 if managing_organisation&.provider_type == "PRP"
@@ -61,7 +63,6 @@ module DerivedVariables::LettingsLogVariables
     if is_supported_housing? && location
       self.wchair = location.mobility_type_before_type_cast == "W" ? 1 : 2
     end
-    self.voiddate = startdate if is_renewal?
     self.vacdays = property_vacant_days
 
     set_housingneeds_fields if housingneeds?
--- a/spec/models/lettings_log_spec.rb
+++ b/spec/models/lettings_log_spec.rb
@@ -1454,6 +1454,15 @@ RSpec.describe LettingsLog do
         expect(record_from_db["vacdays"]).to eq(0)
         expect(lettings_log["vacdays"]).to eq(0)
       end
+
+      it "correctly derives and saves first_time_property_let_as_social_housing" do
+        record_from_db = ActiveRecord::Base.connection.execute(
+          "select first_time_property_let_as_social_housing" \
+          " from lettings_logs where id=#{lettings_log.id}",
+        ).to_a[0]
+        expect(record_from_db["first_time_property_let_as_social_housing"]).to eq(0)
+        expect(lettings_log["first_time_property_let_as_social_housing"]).to eq(0)
+      end
     end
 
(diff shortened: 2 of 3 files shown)


spec/models/lettings_log_spec.rb
```diff
@@ -1454,6 +1454,15 @@ def previous_postcode_fields
         expect(record_from_db["vacdays"]).to eq(0)
         expect(lettings_log["vacdays"]).to eq(0)
       end
+
+      it "correctly derives and saves first_time_property_let_as_social_housing" do
+        record_from_db = ActiveRecord::Base.connection.execute(
```
asmega, line 1459: why do we hand craft SQL queries when we have active record?

```suggestion
record_from_db = LettingsLog.find(id: lettings_log.id)
```

spec/models/lettings_log_spec.rb
```diff
@@ -1454,6 +1454,15 @@ def previous_postcode_fields
...
+          "select first_time_property_let_as_social_housing" \
+          " from lettings_logs where id=#{lettings_log.id}",
+        ).to_a[0]
```
asmega, line 1462: why do we hand craft SQL queries when we have active record?

```suggestion
        record_from_db = LettingsLog.find(lettings_log.id)
```

ghost: By looking at the rest of the file and what it is testing, I _think_ this is to bypass the ActiveRecord model and retrieve data straight from the database.

asmega: you can add a `select` if you really want to restrict the data we request and that comes back

asmega: well at this moment in time i'm gonna say the rest of the file is wrong. we don't interface with the database directly with SQL queries it goes through active record. therefore validating the database has the correct data with SQL, only to then interface with it through active record seems absurd

ghost: I totally agree with that. I will create a tech debt ticket to refactor this test file 👍

ghost: https://digital.dclg.gov.uk/jira/browse/CLDC-2033

natdeanlewissoftwire (approved):

spec/models/lettings_log_spec.rb
jamesrose: Agreed!

kosiakkatrina: Context to why it is the way it is: 
At some point we weren't saving the inferred values to the db, but we had methods that would return that value on the model. After that we made the decision that we actually wanted to persist the value to the db. This code was a consequence of basically testing that.
If we had it as `expect(LettingsLog.find(lettings_log.id).field_we_are_testing_for).to eq(expected_value)` it wold call the method on the model and wouldn't fail as expected
I don't love how this test looks, but in terms of confidence I think it is testing the fact that the data is actually persisted and I'm not sure how else we could do that

app/models/derived_variables/lettings_log_variables.rb
```diff
@@ -60,7 +62,6 @@ def set_derived_fields!
     if is_supported_housing? && location
       self.wchair = location.mobility_type_before_type_cast == "W" ? 1 : 2
     end
-    self.voiddate = startdate if is_renewal?
```
kosiakkatrina, line 64: nice

kosiakkatrina (approved):

spec/models/lettings_log_spec.rb
asmega: We have a ticket now so at least someone can take a look at it. i think at least a few of us thing its less than ideal

PR #1301, opened 2023-02-14
Only load session fiters when getting or exporting the logs

kosiakkatrina:
When creating a new organisation we would pass through params including "organisation". 
That's the same param we use to set organisation session filters.

Previously we would set session filters on every organisation or log load.
When creating a new organisation it would set the organisation filter to the new org params.
When loading lettings logs page we would get an error because we would try to filter by using an unexpected format. Now we set session filters only when we need to


jamesrose:
> a new organisation it would set the organisation filter to the new org params

Where does it actually do this, out of interest?

jamesrose (approved):

PR #1304, opened 2023-02-15
CLDC-1917 add collection year validations for crossover and non-crossover periods

natdeanlewissoftwire:
the ticket describes this work as showing the 23/24 form - I think this is all handled dynamically by individual 23/24 form components depending on startdate, so all the work in this PR is just making sure the validations allow dates in the right collection depending on if the log is created in the crossover period or not


config/locales/en.yml
```diff
@@ -136,7 +136,7 @@ en:
       above_min: "%{field} must be at least %{min}"
     date:
       invalid_date: "Enter a date in the correct format, for example 31 1 2022"
-      outside_collection_window: Enter a date within the 22/23 financial year, which is between 1st April 2022 and 31st March 2023
+      outside_collection_window: "Enter a date within the 22/23 or 23/24 financial years, which is between 1st April 2022 and 31st March 2024"
```
jamesrose: Ideally we would not need to replace this every year!

PR #1307, opened 2023-02-15
Generate old_visible_id for organisations

asmega:
# Context

- We need a mechanism to be able to find organisations based on an identifier
- We will be importing orgs from old core which will already have an identifier which is available at `Organisation#old_visible_id`
- These are used by external third parties to identify records
- New organisations created in new core will not have such an ID
- Therefore we generate this value so we can continue to identify records
- These need to be prefixed with something so they do not clash with an existing identifiers
- We have not fully imported from old core there we do not know all the existing identifier ahead of time
- A prefix will get around this by having a totally different key space

# Changes

- As an active record `after_create` callback on an org, generate `old_visible_id` if it does not already have one
- This value is prefixed to prevent any clashes

### DIFF
--- a/app/models/organisation.rb
+++ b/app/models/organisation.rb
@@ -35,6 +35,8 @@ class Organisation < ApplicationRecord
   validates :name, presence: { message: I18n.t("validations.organisation.name_missing") }
   validates :provider_type, presence: { message: I18n.t("validations.organisation.provider_type_missing") }
 
+  after_create :persist_missing_old_visible_id
+
   def lettings_logs
     LettingsLog.filter_by_organisation(self)
   end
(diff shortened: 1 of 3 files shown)


app/models/organisation.rb
```diff
@@ -96,4 +98,13 @@ def display_organisation_attributes
   def has_managing_agents?
     managing_agents.count.positive?
   end
+
+private
+
+  def persist_missing_old_visible_id
+    if old_visible_id.blank?
+      self.old_visible_id ||= "ORG#{id}"
+      update_column(:old_visible_id, old_visible_id)
```
jamesrose, line 107: On reflection I think it's nice to have one field to distinguish whether an org was migrated or not. Did you consider doing this on the lookup? For example, if `ORG` is appended look at `id`, otherwise `old_visible_id`.

jamesrose:
```ruby
  scope :filter_by_id, ->(id) { id.start_with?("ORG") ? where(id: id[3..]) : where(old_visible_id: id) }
```

Not as pretty, but something like this?

asmega: I've done something along those lines and is present in https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/1311

asmega:
Closing this PR in favour of https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/1311 which contains a change that mitigates the need for this change

PR #1309, opened 2023-02-16
Change CDS export archive names from quarter to financial year

jamesrose:
Follows on from a set of changes to the CDS export (e.g., #1306)

# Context

- We previously pushed logs into archives categorised by the quarter that they were created for.
- CDS requested that instead we push everything into a larger bucket separated by FY.

This PR changes the format of archives to to be categorised by financial year. For example:

`core_2021_2022_jan_mar_f0002_inc0001.zip `becomes: `core_2021_2022_apr_mar_f0002_inc0001.zip`

I have also made a push to start standardising collection start and end date calculations in `CollectionTimeHelper`. This is just a stop-gap to prevent a growing list of hard-coded values for collection start and end dates. We should start _asking_ the form what these dates are. There's a Slack thread [about this here](https://communities-govuk.slack.com/archives/C03QXAE3Y3W/p1676476617031739).

### DIFF
--- a/app/helpers/collection_time_helper.rb
+++ b/app/helpers/collection_time_helper.rb
@@ -1,16 +1,26 @@
 module CollectionTimeHelper
+  def collection_start_year(date)
+    window_end_date = Time.zone.local(date.year, 4, 1)
+    date < window_end_date ? date.year - 1 : date.year
+  end
+
   def current_collection_start_year
-    today = Time.zone.now
-    window_end_date = Time.zone.local(today.year, 4, 1)
-    today < window_end_date ? today.year - 1 : today.year
+    collection_start_year(Time.zone.now)
   end
 
   def collection_start_date(date)
-    window_end_date = Time.zone.local(date.year, 4, 1)
-    date < window_end_date ? Time.zone.local(date.year - 1, 4, 1) : Time.zone.local(date.year, 4, 1)
+    Time.zone.local(collection_start_year(date), 4, 1)
   end
 
   def current_collection_start_date
     Time.zone.local(current_collection_start_year, 4, 1)
+  end
+
+  def collection_end_date(date)
+    Time.zone.local(collection_start_year(date) + 1, 3, 31)
+  end
+
+  def current_collection_end_date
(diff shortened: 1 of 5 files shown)


app/helpers/collection_time_helper.rb
```diff
@@ -1,16 +1,26 @@
...
-    window_end_date = Time.zone.local(today.year, 4, 1)
-    today < window_end_date ? today.year - 1 : today.year
+    collection_start_year(Time.zone.now)
   end
 
   def collection_start_date(date)
-    window_end_date = Time.zone.local(date.year, 4, 1)
-    date < window_end_date ? Time.zone.local(date.year - 1, 4, 1) : Time.zone.local(date.year, 4, 1)
+    Time.zone.local(collection_start_year(date), 4, 1)
   end
```
kosiakkatrina, line 13: Would it make more sense to leave this as is and get the `collection_start_year` method call `collection_start_date(date).year` instead? Less hardcoded numbers

app/helpers/collection_time_helper.rb
```diff
@@ -1,16 +1,26 @@
...
 
   def current_collection_start_date
     Time.zone.local(current_collection_start_year, 4, 1)
   end
+
+  def collection_end_date(date)
+    Time.zone.local(collection_start_year(date) + 1, 3, 31)
```
kosiakkatrina, line 20: collection_start_year(date) + 1.year - 1.day, less hardcoded, still looks like it makes sense? Maybe personal preference, so no need to change if you disagree

kosiakkatrina: or +364.days 🫠

kosiakkatrina (approved):
I take it we don't actually split all the logs into quarters when exporting?

app/helpers/collection_time_helper.rb
jamesrose: Yes, I considered both those options, but then decided what I came up with was less confusing. This definitely should live in the form itself, as discussed!

jamesrose:
@kosiakkatrina Not anymore!

PR #1311, opened 2023-02-16
CLDC-1891 Bulk upload handles owning organisation

asmega:
# Context

- https://digital.dclg.gov.uk/jira/browse/CLDC-1891
- This PR addresses several issues around handling owning organisation for bulk upload
- We need to validate data from CSV on the owning organisation
- We need users to be able to reference owning organisation either via their old core id stored as `old_visible_id` or via their new core `id` which will be prefixed with `ORG` eg. `ORG123`

# Changes

- When given the owning organisation id in bulk upload we look up on both the `old_visible_id` and `id` to attempt to find the correct organisation
- Present back relevant validations based on the data entered for the owning organisation
- If for whatever reason the user has managed to use an ID that references an existing organisation but not permitted to use them we block log creation so invalid data does not enter the system

### DIFF
--- a/app/models/organisation.rb
+++ b/app/models/organisation.rb
@@ -17,8 +17,21 @@ class Organisation < ApplicationRecord
   has_many :managing_agent_relationships, foreign_key: :parent_organisation_id, class_name: "OrganisationRelationship"
   has_many :managing_agents, through: :managing_agent_relationships, source: :child_organisation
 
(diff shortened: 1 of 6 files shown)


asmega:
The coded has been incorporated via https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/1315

PR #1314, opened 2023-02-16
CLDC-1909 Add 2023-2024 collection year filter

ghost:
Screenshot 2023-02-16 at 14 18 11

### DIFF
--- a/app/helpers/filters_helper.rb
+++ b/app/helpers/filters_helper.rb
@@ -27,4 +27,12 @@ module FiltersHelper
     organisation_options = user.support? ? Organisation.all : [user.organisation] + user.organisation.managing_agents
     [OpenStruct.new(id: "", name: "Select an option")] + organisation_options.map { |org| OpenStruct.new(id: org.id, name: org.name) }
   end
+
+  def collection_year_options
+    if FeatureToggle.collection_2023_2024_year_enabled?
+      { "2023": "2023/24", "2022": "2022/23", "2021": "2021/22" }
+    else
+      { "2022": "2022/23", "2021": "2021/22" }
+    end
+  end
 end
--- a/app/views/logs/_log_filters.erb
+++ b/app/views/logs/_log_filters.erb
@@ -6,7 +6,6 @@
 
     <div class="app-filter__content">
       <%= form_with html: { method: :get } do |f| %>
-        <% years = { "2021": "2021/22", "2022": "2022/23" } %>
         <% all_or_yours = { "all": { label: "All" }, "yours": { label: "Yours" } } %>
 
         <% if bulk_upload_options(@bulk_upload).present? %>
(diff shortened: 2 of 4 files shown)


app/helpers/filters_helper.rb
```diff
@@ -27,4 +27,12 @@ def organisations_filter_options(user)
...
     [OpenStruct.new(id: "", name: "Select an option")] + organisation_options.map { |org| OpenStruct.new(id: org.id, name: org.name) }
   end
+
+  def collection_year_options
+    if FeatureToggle.collection_2023_2024_year_enabled?
+      { "2023": "2023/24", "2022": "2022/23", "2021": "2021/22" }
+    else
+      { "2022": "2022/23", "2021": "2021/22" }
+    end
+  end
```
kosiakkatrina, line 37: Looks good, but I'm wondering if we can get these years from form handler

kosiakkatrina: And we can always return the current and previous form, without having to check the toggle

ghost: but in [this](https://digital.dclg.gov.uk/jira/browse/CLDC-1916) ticket seems that we want to be able to filter for all the forms not just the previous and current ones? 
I do not find reusable logic in the FormHandler nor I can see it reused if I were to move this in the FormHandler. What do you mean precisely?

jamesrose: Agreed. We should try and find out how to dynamically generate this list.

ghost: I do not think this is dynamic list but it's instead a set list of years we're allowing in the system. 
In fact, we are not dynamically generating a similar list in the FormHandler and manually building it (see `sales_forms` and `lettings_forms`).

One way I can think of doing this dynamically (just for the filter, not the FormHandler) is to query the all the logs collection years and group them. This means running an additional db query (probably fine) but will change the filter behaviour

kosiakkatrina: So what I was thinking was something like:
`previous_year = FormHandler.instance.forms["previous_lettings"].start_date.year`
`current_year = FormHandler.instance.forms["current_lettings"].start_date.year`

And then construct the options from those.
It's for a specific type of form (lettings) but I think our collection years should align.

We shouldn't need a filter for the next collection window, because there wouldn't be any logs yet and I'm not sure if we should be showing the logs before the previous collection year in general, because I wouldn't expect to keep a record of `previous_previous_form` etc. Hence only previous and current year filters

The point around keeping/displaying logs from 2 years ago needs some more general discussion I think

kosiakkatrina: In regards to your point about dynamically generating forms, I think the current form depends on the current time, so it will switch to 23/24 form in April automatically, leaving us with 22/23 and 23/24 forms only (and a 24/25 next form I think which will look exactly as 23/24 but we shouldn't use that anyway)

jamesrose (approved):

kosiakkatrina (approved):

PR #1317, opened 2023-02-17
CLDC-1916 Handle sales forms and validations during comparison period

ghost:
- dynamically generate error messages for sales financial periods
- use crossover period to decide which error message to display
- add capybara screenshot
https://digital.dclg.gov.uk/jira/browse/CLDC-1916

### DIFF
--- a/app/helpers/collection_time_helper.rb
+++ b/app/helpers/collection_time_helper.rb
@@ -24,6 +24,10 @@ module CollectionTimeHelper
     Time.zone.local(current_collection_start_year + 1, 3, 31)
   end
 
+  def next_collection_end_date
+    current_collection_end_date + 1.year
+  end
+
   def next_collection_start_year
     current_collection_start_year + 1
   end
--- a/app/helpers/collection_time_helper.rb
+++ b/app/helpers/collection_time_helper.rb
@@ -23,4 +23,12 @@ module CollectionTimeHelper
   def current_collection_end_date
     Time.zone.local(current_collection_start_year + 1, 3, 31)
   end
+
+  def next_collection_start_year
+    current_collection_start_year + 1
+  end
+
+  def previous_collection_start_year
+    current_collection_start_year - 1
+  end
 end
--- a/app/helpers/collection_time_helper.rb
+++ b/app/helpers/collection_time_helper.rb
@@ -24,8 +24,8 @@ module CollectionTimeHelper
     Time.zone.local(current_collection_start_year + 1, 3, 31)
   end
 
-  def next_collection_end_date
-    current_collection_end_date + 1.year
+  def previous_collection_end_date
+    current_collection_end_date - 1.year
   end
 
   def next_collection_start_year
@@ -34,5 +34,9 @@ module CollectionTimeHelper
 
   def previous_collection_start_year
     current_collection_start_year - 1
+  end
+
+  def previous_collection_start_date
+    current_collection_start_date - 1.year
   end
 end
--- a/app/models/validations/sales/setup_validations.rb
+++ b/app/models/validations/sales/setup_validations.rb
@@ -1,11 +1,44 @@
 module Validations::Sales::SetupValidations
   include Validations::SharedValidations
+  include CollectionTimeHelper
 
   def validate_saledate(record)
     return unless record.saledate && date_valid?("saledate", record)
 
-    unless record.saledate.between?(current_collection_start_date, current_collection_end_date) || !FeatureToggle.saledate_collection_window_validation_enabled?
-      record.errors.add :saledate, I18n.t("validations.setup.saledate.financial_year")
+    unless record.saledate.between?(current_collection_start_date, active_collection_end_date)
+      record.errors.add :saledate, validation_error_message
+    end
+  end
+
+private
+
+  def active_collection_end_date
+    if FeatureToggle.saledate_next_collection_year_validation_enabled?
+      next_collection_end_date
+    else
+      current_collection_end_date
+    end
+  end
+
+  def validation_error_message
+    start_date = current_collection_start_date
+    if FeatureToggle.saledate_next_collection_year_validation_enabled?
+      I18n.t(
+        "validations.setup.saledate.current_and_next_financial_year",
+        current_start_year_short: start_date.strftime("%y"),
+        current_end_year_short: current_collection_end_date.strftime("%y"),
+        current_start_year_long: start_date.strftime("%Y"),
+        next_end_year_short: next_collection_end_date.strftime("%y"),
+        next_end_year_long: next_collection_end_date.strftime("%Y"),
+      )
+    else
+      I18n.t(
+        "validations.setup.saledate.current_financial_year",
+        current_start_year_short: start_date.strftime("%y"),
+        current_end_year_short: current_collection_end_date.strftime("%y"),
+        current_start_year_long: start_date.strftime("%Y"),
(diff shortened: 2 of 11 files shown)


app/helpers/collection_time_helper.rb
```diff
@@ -13,4 +13,16 @@ def collection_start_date(date)
...
+    Time.zone.local((current_collection_start_year + 1), 3, 31)
+  end
+
+  def next_collection_start_year
+    current_collection_start_year + 1
+  end
+
+  def previous_collection_start_year
+    current_collection_start_year - 1
+  end
```
jamesrose, line 37: Might be worth a rebase as I'd already implemented `current_collection_end_date` in a commit merged earlier this week.

app/models/validations/sales/setup_validations.rb
```diff
@@ -1,10 +1,11 @@
 module Validations::Sales::SetupValidations
   include Validations::SharedValidations
+  include CollectionTimeHelper
 
   def validate_saledate(record)
     return unless record.saledate && date_valid?("saledate", record)
 
-    unless record.saledate.between?(Time.zone.local(2022, 4, 1), Time.zone.local(2023, 3, 31)) || !FeatureToggle.saledate_collection_window_validation_enabled?
+    unless record.saledate.between?(current_collection_start_date, current_collection_end_date) || !FeatureToggle.saledate_collection_window_validation_enabled?
```
jamesrose: Nice!!

app/models/validations/sales/setup_validations.rb
```diff
@@ -1,11 +1,44 @@
...
+  end
+
+private
+
+  def active_collection_end_date
+    if FeatureToggle.saledate_next_collection_year_validation_enabled?
+      next_collection_end_date
+    else
+      current_collection_end_date
+    end
```
kosiakkatrina: I think there's a method `sales_in_crossover_period?` that we could potentially use instead of the feature flag

ghost: nice, missed that! Will update shortly

jamesrose (approved):

app/models/validations/sales/setup_validations.rb
```diff
@@ -1,11 +1,43 @@
...
 
   def validate_saledate(record)
     return unless record.saledate && date_valid?("saledate", record)
 
-    unless record.saledate.between?(Time.zone.local(2022, 4, 1), Time.zone.local(2023, 3, 31)) || !FeatureToggle.saledate_collection_window_validation_enabled?
-      record.errors.add :saledate, I18n.t("validations.setup.saledate.financial_year")
+    unless record.saledate.between?(current_collection_start_date, active_collection_end_date)
+      record.errors.add :saledate, validation_error_message
+    end
+  end
```
kosiakkatrina: Shouldn't this be .between?(active_collection_start_date, current_collection_start_date)?

app/models/validations/sales/setup_validations.rb
```diff
@@ -1,11 +1,43 @@
...
+
+private
+
+  def active_collection_end_date
+    if FormHandler.instance.sales_in_crossover_period?
+      next_collection_end_date
+    else
+      current_collection_end_date
+    end
+  end
```
kosiakkatrina: If it's in crossover period, shouldn't we be returning previous form? and current if not in crossover?

spec/models/validations/sales/setup_validations_spec.rb
```diff
@@ -6,43 +6,95 @@
   let(:validator_class) { Class.new { include Validations::Sales::SetupValidations } }
 
   describe "#validate_saledate" do
-    context "when saledate is blank" do
-      let(:record) { FactoryBot.build(:sales_log, saledate: nil) }
+    context "with sales_in_crossover_period == true" do
+      before do
+        allow(FormHandler.instance).to receive(:sales_in_crossover_period?).and_return(true)
+      end
```
kosiakkatrina: Maybe mock the date to be in crossover period instead? Because I think the previous, current and next forms will change when we enter the crossover period

spec/models/validations/sales/setup_validations_spec.rb
```diff
@@ -6,43 +6,100 @@
...
+      # end
+
+      around do |example|
+        Timecop.freeze(Time.zone.local(2022, 5, 1)) do
+          Singleton.__init__(FormHandler)
+          example.run
+        end
+        Timecop.return
+        Singleton.__init__(FormHandler)
+      end
```
kosiakkatrina: These tests might start failing when we go into the next year, because we will no longer have the 21/22 form so `sales_in_crossover_period?` will return false

Maybe there's a way you can set those dates dynamically, or at least test with the next years crossover date?

config/locales/en.yml
```diff
@@ -150,7 +150,11 @@ en:
       intermediate_rent_product_name:
         blank: "Enter name of other intermediate rent product"
       saledate:
-        financial_year: "Date must be from 22/23 financial year, which is between 1st April 2022 and 31st March 2023"
+        current_financial_year:
+          Enter a date within the %{current_start_year_short}/%{current_end_year_short} financial year, which is between 1st April %{current_start_year_long} and 31st March %{current_end_year_long}
```
jamesrose: Should we inject inject 1st April and 31st March dynamically from `active_collection_start_date` etc?

config/locales/en.yml
```diff
@@ -150,7 +150,11 @@ en:
...
+        previous_and_current_financial_year:
+          "Enter a date within the %{previous_start_year_short}/%{previous_end_year_short} or %{previous_end_year_short}/%{current_end_year_short} financial years, which is between 1st April %{previous_start_year_long} and 31st March %{current_end_year_long}"
+
```
kosiakkatrina: would it make sense to pass in those dates fully as opposed to hardcode April 1st and March 31st in the error message?

spec/helpers/tasklist_helper_spec.rb
```diff
@@ -1,100 +1,55 @@
...
-          expect(get_subsections_count(empty_lettings_log, :not_started)).to eq(8)
-        end
-
-        it "returns the number of sections in progress" do
-          expect(get_subsections_count(lettings_log, :in_progress)).to eq(3)
-        end
-
-        it "returns 0 for invalid state" do
-          expect(get_subsections_count(lettings_log, :fake)).to eq(0)
-        end
```
kosiakkatrina, line 44: Any reason we're removing all of these?

config/initializers/feature_toggle.rb
```diff
@@ -7,10 +7,6 @@ def self.startdate_collection_window_validation_enabled?
     Rails.env.production? || Rails.env.test? || Rails.env.staging?
   end
 
-  def self.saledate_collection_window_validation_enabled?
-    Rails.env.production? || Rails.env.test? || Rails.env.staging?
-  end
-
```
kosiakkatrina: This makes sense so that the validation could be checked, but we probably want the validation to be disabled on review app afterwards, so that 23/24 logs could be reviewed

config/locales/en.yml
kosiakkatrina: Oh, lol I left the same comment. Twins ✨

spec/helpers/tasklist_helper_spec.rb
ghost: made that method private and the calling method is already tested

kosiakkatrina (approved):

PR #1318, opened 2023-02-17
 CLDC-1917 Fix startdate validations for crossover/not

natdeanlewissoftwire:
the ticket describes this work as showing the 23/24 form - I think this is all handled dynamically by individual 23/24 form components depending on startdate, so all the work in this PR is just making sure the validations allow dates in the right collection depending on if the log is created in the crossover period or not


app/models/validations/date_validations.rb
```diff
@@ -68,20 +68,32 @@ def validate_startdate(record)
...
-  def second_collection_start_date
-    @second_collection_start_date ||= FormHandler.instance.lettings_forms["current_lettings"].start_date
+  def previous_collection_start_date
+    @previous_collection_start_date ||= FormHandler.instance.lettings_forms["previous_lettings"].start_date
   end
 
-  def second_collection_end_date
-    @second_collection_end_date ||= FormHandler.instance.lettings_forms["current_lettings"].end_date
+  def previous_collection_end_date
+    @previous_collection_end_date ||= FormHandler.instance.lettings_forms["previous_lettings"].end_date
```
natdeanlewissoftwire: these instantated start and end dates are the issues causing the tests to fail - their state is carried on between tests, e.g. in tasklist_helper_spec.rb: if "with 2023 deadline/returns relevant text" runs before "with 2024 deadline/returns relevant text", the 2024 one will fail because these start and end dates are still set from the previous test

app/models/validations/date_validations.rb
```diff
@@ -68,20 +68,32 @@ def validate_startdate(record)
 
 private
 
-  def first_collection_start_date
-    @first_collection_start_date ||= FormHandler.instance.lettings_forms["previous_lettings"].start_date
+  def previous_collection_start_suffix
+    previous_collection_start_date.strftime("%y").to_i
```
jamesrose:
```suggestion
    previous_collection_start_date.year
```

:)

app/models/validations/date_validations.rb
```diff
@@ -68,20 +68,32 @@ def validate_startdate(record)
...
   end
 
-  def first_collection_end_date
-    @first_collection_end_date ||= FormHandler.instance.lettings_forms["previous_lettings"].end_date
+  def current_collection_start_suffix
+    current_collection_start_date.strftime("%y").to_i
```
jamesrose:
```suggestion
    current_collection_start_date.year
```

natdeanlewissoftwire: `strftime("%y")` gives me e.g. 22 from 2022 which is what I want - I could do` .year % 100` I suppose, that's probably clearer

natdeanlewissoftwire: `strftime("%y")` gives me e.g. 22 from 2022 which is what I want - I could do` .year % 100` I suppose, that's probably clearer

PR #1326, opened 2023-02-20
CLDC-1895 pluralise various copy when joint purchase

arfacamble:
https://digital.dclg.gov.uk/jira/browse/CLDC-1895


app/models/form/sales/questions/prevown.rb
```diff
@@ -1,9 +1,9 @@
 class Form::Sales::Questions::Prevown < ::Form::Question
-  def initialize(id, hsh, page)
-    super
+  def initialize(id, hsh, page, joint_purchase:)
+    super(id, hsh, page)
     @id = "prevown"
-    @check_answer_label = "Buyers previously owned a property"
-    @header = "Has the buyer previously owned a property?"
+    @check_answer_label = "Buyer#{'s' if joint_purchase} previously owned a property"
```
jamesrose: Rails is clever! You get it to do it for you, for example:

```ruby
"buyer".pluralize(joint_purchase ? 2 : 1)
```

arfacamble: I knew about pluralize, but hadn't remembered the number argument thing.
Unless you feel strongly about this I find using pluralize marginally less readable just because I wouldn't know what the two and the 1 are doing without going and looking it up.
It's definitely a margin call though!

natdeanlewissoftwire: I didn't know about pluralize so thanks @jamesrose even though this was for Arthur haha, it is clever!

jamesrose:
@arfacamble Shall we discuss this one? I think we can improve how it reads.

jamesrose (approved):

PR #1327, opened 2023-02-20
CLDC-2035 Fix search bug

natdeanlewissoftwire:
Fixes search text and routing for sales logs, also adds correct filters for sales logs so users can find logs by id, postcode or purchaser code
image
image

ticket: https://digital.dclg.gov.uk/jira/browse/CLDC-2035


app/components/search_component.rb
```diff
@@ -23,6 +23,8 @@ def path(current_user)
       user_path(current_user)
     elsif request.path.include?("organisations")
       organisations_path
+    elsif FeatureToggle.sales_log_enabled? && request.path.include?("sales-logs")
```
ghost: the should be a test covering for this new behaviour

app/views/logs/index.html.erb
```diff
@@ -64,9 +64,24 @@
...
-    <%= render partial: "log_list", locals: { logs: @logs, title: "Logs", pagy: @pagy, searched: @searched, item_label:, total_count: @total_count, csv_download_url: csv_download_lettings_logs_path(search: @search_term) } %>
+
+    <% if current_page?(controller: 'lettings_logs', action: 'index') %>
+      <%= render partial: "log_list", locals: { logs: @logs, title: "Logs", pagy: @pagy, searched: @searched, item_label:, total_count: @total_count, csv_download_url: csv_download_lettings_logs_path(search: @search_term) } %>
+    <% end %>
+
+    <% if FeatureToggle.sales_log_enabled? && current_page?(controller: 'sales_logs', action: 'index') %>
+      <%= render partial: "log_list", locals: { logs: @logs, title: "Logs", pagy: @pagy, searched: @searched, item_label:, total_count: @total_count } %>
+    <% end %>
+
```
kosiakkatrina: Is the only difference between these the csv_download_url? could we only make that conditional? Also, do we need to check the feature toggle?

app/views/logs/index.html.erb
```diff
@@ -64,9 +64,24 @@
   <%= render partial: "log_filters" %>
 
   <div class="app-filter-layout__content">
-    <%= render SearchComponent.new(current_user:, search_label: "Search by log ID, tenant code, property reference or postcode", value: @searched) %>
+    <% if current_page?(controller: 'lettings_logs', action: 'index') %>
+      <%= render SearchComponent.new(current_user:, search_label: "Search by log ID, tenant code, property reference or postcode", value: @searched) %>
+    <% end %>
+
+    <% if FeatureToggle.sales_log_enabled? && current_page?(controller: 'sales_logs', action: 'index') %>
+      <%= render SearchComponent.new(current_user:, search_label: "Search by log ID, purchaser code or postcode", value: @searched) %>
```
kosiakkatrina: Could we extract search_label into a method (maybe in a helper?) and lose the if statements in the erb?

app/models/log.rb
```diff
@@ -18,6 +18,7 @@ class Log < ApplicationRecord
     years.each { |year| query = query.or(filter_by_year(year)) }
     query.all
   }
+  scope :filter_by_postcode, ->(postcode_full) { where("REPLACE(postcode_full, ' ', '') ILIKE ?", "%#{postcode_full.delete(' ')}%") }
```
kosiakkatrina, line 21: nice!

app/components/search_component.rb
kosiakkatrina: Do we need the feature toggle here?

app/views/logs/index.html.erb
natdeanlewissoftwire: Good idea :D

app/components/search_component.rb
natdeanlewissoftwire: eh probably not

natdeanlewissoftwire: I agree, there aren't tests for the old behaviours so I'll see what makes the most sense to add

natdeanlewissoftwire: have added tests for log filtering by log specific values (for sales and lettings), not sure there's much else that provides value but lmk if you can think of anything!

ghost (approved):

PR #1330, opened 2023-02-21
CLDC-2021 ensure that status is up to date when records are called from the DB

arfacamble:
After discussion with Kat we think the most likely reason for inconsistent data on staging is that there were logs that were completed at the time and subsequently questions were added to the form.

that has been dealt with in a console, here are just some small changes that I find more readable/efficient looking

### DIFF
--- a/app/models/log.rb
+++ b/app/models/log.rb
@@ -111,14 +111,12 @@ private
   end
 
   def all_fields_completed?
-    subsection_statuses = form.subsections.map { |subsection| subsection.status(self) if subsection.displayed_in_tasklist?(self) }.uniq.compact
-    subsection_statuses == [:completed]
+    form.subsections.all? { |subsection| subsection.complete?(self) || subsection.not_displayed_in_tasklist?(self) }
   end
 
   def all_fields_nil?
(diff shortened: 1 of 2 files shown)


jamesrose (approved):

app/models/log.rb
```diff
@@ -7,6 +7,7 @@ class Log < ApplicationRecord
   belongs_to :bulk_upload, optional: true
 
   before_save :update_status!
+  after_find :verify_status!
```
jamesrose: We should not mutate objects on find. What are you trying to do here?

jamesrose: If we would like to ensure that all existing logs have an up to date status, we can touch them all and `update_status!` will run. I'd be curious to inspect whether this would actually change the status on logs, though?

```ruby
class Log
  def calculate_status
    if all_fields_completed? && errors.empty?
      "completed"
    elsif all_fields_nil?
      "not_started"
    else
      "in_progress"
    end
  end
end

LettingsLog.all.select { |log| log.status != log.calculate_status }.count
```

I ran this on production for you:

```ruby
=> 34
```

jamesrose (approved):

PR #1331, opened 2023-02-22
CLDS-1938 Add sales logs importer

kosiakkatrina:
Create a Sales log Import service.

Extract shared Lettings log and Sales log import service behaviour into Log Import service parent class. There is also Import service class but that seemed to generic to stored log import specific behaviour.

Service maps incoming old CORE data from XML to our domain objects.

A lot of the fields were optional on the old CORE, so if imported log is completed (submitted), we infer some of the missing values, such as age, gender and armed forces questions as "Don't know".

The test XML files are anonymised examples of real logs. Some parts of those logs are commented out because I am still waiting to find out how to import them (some of them are optional on old core and not on this service).


app/services/imports/sales_logs_import_service.rb
```diff
@@ -0,0 +1,405 @@
...
+          attributes["sex#{index}"] == "R" &&
+          attributes["relat#{index}"] == "R" &&
+          attributes["ecstat#{index}"] == 10
+        2 # No
+      else
+        1 # Yes
+      end
+    end
+
+    MORNTGAGE_LENDER_OPTIONS = {
```
jamesrose:
```suggestion
    MORTGAGE_LENDER_OPTIONS = {
```

jamesrose (approved):

PR #1334, opened 2023-02-22
CLDC-1888 Bulk upload handles duplicates

asmega:
# Context

- https://digital.dclg.gov.uk/jira/browse/CLDC-1888
- This is a continuation of https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/1277
- When bulk uploading we want to check users are not uploading data that already exists to prevent them submitting duplicate 

# Changes

- This feature is behind a feature toggle. it has been disabled for staging for testing purposes but available in all other environments
- If a log already exists based off certain fields add errors to the associated fields
- We discount any hidden logs and only check "active" logs
- Added memoization to `#valid?` as an optimisation

### DIFF
--- a/spec/factories/bulk_upload_error.rb
+++ b/spec/factories/bulk_upload_error.rb
@@ -9,7 +9,7 @@ FactoryBot.define do
     tenant_code { SecureRandom.hex(4) }
     property_ref { SecureRandom.hex(4) }
     purchaser_code { SecureRandom.hex(4) }
-    field { "field_#{rand(134) + 1}" }
+    field { "field_#{rand(1..134)}" }
     error { "some error" }
   end
 end
--- a/spec/factories/bulk_upload_error.rb
+++ b/spec/factories/bulk_upload_error.rb
@@ -9,7 +9,7 @@ FactoryBot.define do
     tenant_code { SecureRandom.hex(4) }
     property_ref { SecureRandom.hex(4) }
     purchaser_code { SecureRandom.hex(4) }
-    field { "field_#{rand(134)}" }
+    field { "field_#{rand(134) + 1}" }
     error { "some error" }
   end
 end
--- a/spec/mailers/bulk_upload_mailer_spec.rb
+++ b/spec/mailers/bulk_upload_mailer_spec.rb
@@ -90,11 +90,10 @@ RSpec.describe BulkUploadMailer do
             year_combo: bulk_upload.year_combo,
             lettings_or_sales: bulk_upload.log_type,
             summary_report_link: "http://localhost:3000/lettings-logs/bulk-upload-results/#{bulk_upload.id}",
-            error_description: "Please correct your data export and upload again.\n",
           },
         )
 
-        mailer.send_correct_and_upload_again_mail(bulk_upload:, errors: {})
+        mailer.send_correct_and_upload_again_mail(bulk_upload:)
       end
     end
   end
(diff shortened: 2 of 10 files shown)


jamesrose (approved):

spec/mailers/bulk_upload_mailer_spec.rb
```diff
@@ -109,13 +109,8 @@
...
-    context "when 2 columns with errors" do
-      before do
-        create(:bulk_upload_error, bulk_upload:, col: "A")
-        create(:bulk_upload_error, bulk_upload:, col: "B")
-      end
-
-      it "sends correctly formed email with A, B" do
+    context "when there are no errors" do
+      it "sends correctly formed email" do
         expect(notify_client).to receive(:send_email).with(
```
SamSeed-Softwire: Does this expect line get called after the mail has been sent (line 127)? If so, why is it before it in the file?

spec/mailers/bulk_upload_mailer_spec.rb
```diff
@@ -144,6 +139,8 @@
       end
 
       it "sends correctly formed email with A, B, C and more" do
+        error_description = "We noticed the following issues with your upload:\n- logs where the setup sections were incomplete\n- logs with a lot of similar errors in column A, B, C and more\n- logs you are trying to upload have been created previously\n"
+
         expect(notify_client).to receive(:send_email).with(
```
SamSeed-Softwire: Same as above comment - is this expect statement not the final thing to be called in this it block?

SamSeed-Softwire (approved):

spec/factories/bulk_upload_error.rb
```diff
@@ -9,7 +9,7 @@
     tenant_code { SecureRandom.hex(4) }
     property_ref { SecureRandom.hex(4) }
     purchaser_code { SecureRandom.hex(4) }
-    field { "field_#{rand(134)}" }
+    field { "field_#{rand(1..134)}" }
```
SamSeed-Softwire, line 12: `rand(134)` generates a number in the range 0-133. We want a number in the range 1-134.

app/services/bulk_upload/lettings/row_parser.rb
```diff
@@ -403,6 +422,29 @@ def setup_question?(question)
     log.form.setup_sections[0].subsections[0].questions.include?(question)
   end
 
+  def validate_if_log_already_exists
+    if log_already_exists?
+      error_message = "This is a duplicate log"
+
+      errors.add(:field_5, error_message) # location
```
SamSeed-Softwire: Is `field_5` correct here? In the bulk upload template it seems that `field_5` is the scheme, but the logic in the `location` method seems to find the location using this field.

asmega: my understanding is old core doesn't have the concept of a location. so scheme here actually relates to new core location and `Management group code` is the scheme

spec/mailers/bulk_upload_mailer_spec.rb
asmega: - if you want it read so its after you have to use `have_received` see https://rubydoc.info/github/rspec/rspec-mocks/RSpec%2FMocks%2FExampleMethods:have_received
- this is just how the `receive` part of the DSL works. you set up the expectation before, call the thing. this way it knows if the expectation happened or not.
- `have_received` works the same way except you write it afterwards
- i don't actually know how it works under the hood but i would guess it setups all the expectations up first then calls the code so it knows if mocks received messages or not

spec/mailers/bulk_upload_mailer_spec.rb
asmega: see other reply

jamesrose (approved):

PR #1335, opened 2023-02-22
CLDC-1788 Add bulk upload guidance page

natdeanlewissoftwire:
Adds this page with different template links depending on if the user has come from sales/lettings prepare file page: 

image

Happy to take suggestions on what to add tests for, it's mainly adding a page of copy/reusing the template though so not certain how much value tests would add to this.

Ticket: https://digital.dclg.gov.uk/jira/browse/CLDC-1788


app/views/bulk_upload_shared/guidance.html.erb
```diff
@@ -0,0 +1,52 @@
...
+    <div class="govuk-!-padding-bottom-4">
+      <h2 class="govuk-heading-s">Exporting CSV data</h2>
+      <p class="govuk-body">Export CSV data directly from your current systems, or export then adjust it to CSV.</p>
+      <p class="govuk-body">You can then upload it via a button at the top of the lettings and sales logs pages.</p>
+      <%= govuk_details(summary_text: "My organisation has a CMS") do %>
+        <p class="govuk-body">Some HMS providers sell an add-on "eCORE" module, which exports CSV data for you.</p>
+        <p class="govuk-!-font-weight-bold">It can take HMS providers a while to update these per new collection year, so you may have to wait for updates to export, or adjust your data manually post-export.</p>
+      <% end %>
+      <%= govuk_details(summary_text: "My organisation does not have a CMS") do %>
+        <p class="govuk-body">Your organisation’s IT team may be able to export CSV data for you - find out more about data specifications. This document outlines:</p>
```
natdeanlewissoftwire: "Find out more about data specs" should be a link, but that page hasn't been built yet it seems, so this sounds a bit passive aggressive atm. Could link to the spec form as a quick fix, or just wait til the specs are ready. I've added a comment to [this ticket](https://digital.dclg.gov.uk/jira/browse/CLDC-1774) about this.

app/views/bulk_upload_shared/guidance.html.erb
```diff
@@ -0,0 +1,52 @@
...
+          <li>each field's valid response</li>
+          <li>if/when certain fields can be left blank</li>
+        </ul>
+        <p class="govuk-body">Fields can appear in any order, as long as you include the <%= govuk_link_to "template document", @template_path %> headers, to easily identify what each column represents. You can rearrange data columns to match your system exports, copy-pasting multiple columns at once. For data stored in multiple systems, you can copy-paste all columns for one system next to each other, repeating this for subsequent system exports.</p>
+      <% end %>
+    </div>
+
+    <div class="govuk-!-padding-bottom-4">
+      <h2 class="govuk-heading-s">Getting help</h2>
+      <p class="govuk-body">There is no step-by-step bulk upload guide like there is with single log upload. However, you can download <%= govuk_link_to "our template", @template_path %>, which you can copy-paste data into from your systems column-by-column. You can also view a post-upload report showing any data errors, and our data specification can help fix these.</p>
```
natdeanlewissoftwire: should also be a link on "data specification", same as above

app/views/bulk_upload_shared/guidance.html.erb
jamesrose: Is this ready for review or should we hold off until you get an answer about that page?

natdeanlewissoftwire: This is ready for review now, those spec links are now in this PR too :)

app/models/forms/bulk_upload_lettings/prepare_your_file.rb
```diff
@@ -29,6 +29,10 @@ def template_path
         "/files/bulk-upload-lettings-template-v1.xlsx"
       end
 
+      def specification_path
+        "/files/Lettings 2022-23 Bulk Upload Specification.csv"
```
asmega: I recommended replacing the spaces for underscores or hyphens

app/models/forms/bulk_upload_sales/prepare_your_file.rb
```diff
@@ -27,6 +27,10 @@ def template_path
         "/files/bulk-upload-sales-template-v1.xlsx"
       end
 
+      def specification_path
+        "/files/Sales 2022-23 Bulk Upload Specification.csv"
```
asmega: these files don't actually appear to be CSV files, they're showing up as binary for this review and not text. therefore my guess is that are actually excel spreadsheets and should therefore honour the extension rather than masquerading the file as a CSV when it is not

app/models/forms/bulk_upload_lettings/prepare_your_file.rb
natdeanlewissoftwire: Ah good point, forgot to put this in, cheers

app/models/forms/bulk_upload_sales/prepare_your_file.rb
natdeanlewissoftwire: Yep, they're direct from https://core.communities.gov.uk/content/GuidesAndManuals.html - you're right, they're not real csvs... was initially hesitant to change these files to be different to what they are on that page though, and afaik I don't have access to change that one. I've just seen the templates are different in our service to what they are on that site however so I'll go ahead and make that change

kosiakkatrina (approved):
You could add a test for routing and one just to check that the page exists, but looks good otherwise

PR #1339, opened 2023-02-23
CDLC-1818 add new answer options to nationality

arfacamble:
https://digital.dclg.gov.uk/jira/browse/CLDC-1818

### DIFF
--- a/app/models/form/lettings/questions/nationality.rb
+++ b/app/models/form/lettings/questions/nationality.rb
@@ -1,9 +1,9 @@
-class Form::Lettings::Questions::National < ::Form::Question
+class Form::Lettings::Questions::Nationality < ::Form::Question
   def initialize(id, hsh, page)
     super
     @id = "national"
     @check_answer_label = "Lead tenant’s nationality"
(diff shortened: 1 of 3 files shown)


app/models/form/lettings/questions/nationality.rb
```diff
@@ -14,6 +14,8 @@ def initialize(id, hsh, page)
     "18" => { "value" => "United Kingdom" },
     "17" => { "value" => "Republic of Ireland" },
     "19" => { "value" => "European Economic Area (EEA) country, excluding Ireland" },
+    "20" => { "value" => "Afghanistan" },
+    "21" => { "value" => "Ukraine" },
```
jamesrose, line 18: Are these applicable to previous forms as well?

arfacamble: meaning previous years? I don't believe so. But I think to add it to previous years I would have to add them to the json right? These ruby definitions for lettings will only be active from 23/24

jamesrose (approved):

PR #1340, opened 2023-02-23
Bulk upload handles invalid byte sequences

asmega:
# Context

- Sometimes CSV is not exported as utf8 by spreadsheet software therefore invalid byte sequences are present

# Changes

- if invalid byte sequence is found, remove this before parsing as CSV

# Notes for review

- Not sure if I'm doing this right or if there's a better way?

### DIFF
--- a/app/services/bulk_upload/lettings/csv_parser.rb
+++ b/app/services/bulk_upload/lettings/csv_parser.rb
@@ -52,6 +52,7 @@ private
 
     @normalised_string = File.read(path, encoding: "bom|utf-8")
     @normalised_string.gsub!("\r\n", "\n")
+    @normalised_string.scrub!("")
 
     @normalised_string
   end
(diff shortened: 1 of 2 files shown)


ghost (approved):

app/services/bulk_upload/lettings/csv_parser.rb
```diff
@@ -52,6 +52,7 @@ def normalised_string
 
     @normalised_string = File.read(path, encoding: "bom|utf-8")
```
davidm-m, line 53: Do you know what the invalid byte sequence was? This sounds like a BOM issue (we've hit this a few times on the old service) but looks like this line should handle it without any additional scrubbing.

asmega: i see it as `<92>` so i'm guessing it's some kind of apostrophe or single smart quote. the box recognised the file as
```sh
file THE_FILE
THE_FILE: Non-ISO extended-ASCII text, with very long lines, with CRLF, LF line terminators
```

asmega: Sorry there are also more sequences:
- `<A0>` - non-breaking space
- `<93>` - start smart double quote
- `<94>` - end smart double quote

asmega: ~actually let me get back to you it's possible it maybe something else~

EDIT: correction, it does seem related to those sequences

PR #1343, opened 2023-02-23
CLDC-1821 tweak wheelchair question depends on

arfacamble:
https://digital.dclg.gov.uk/jira/browse/CLDC-1821

amend wheelchair question class name, tweak depends on for readability, add test file


app/models/form/lettings/pages/property_wheelchair_accessible.rb
```diff
@@ -2,10 +2,10 @@ class Form::Lettings::Pages::PropertyWheelchairAccessible < ::Form::Page
   def initialize(id, hsh, subsection)
     super
     @id = "property_wheelchair_accessible"
-    @depends_on = [{ "needstype" => 1 }]
+    @depends_on = [{ "is_supported_housing?" => false }]
```
jamesrose, line 5: Is there a test for this?

arfacamble: there is now!

jamesrose (approved):

PR #1349, opened 2023-02-27
CLDC-1864 Introduce question numbering

kosiakkatrina:
Number sales log questions to match the paper form numbering
Can check mapping here: [CLDC-1864](https://digital.dclg.gov.uk/jira/browse/CLDC-1864)


app/models/form/sales/pages/privacy_notice.rb
```diff
@@ -2,7 +2,7 @@ class Form::Sales::Pages::PrivacyNotice < ::Form::Page
   def initialize(id, hsh, subsection)
     super
     @id = "privacy_notice"
-    @header = "Department for Levelling Up, Housing and Communities privacy notice"
+    @header = "Q19 - Department for Levelling Up, Housing and Communities privacy notice"
```
asmega, line 5: would it not have been easier to leave the header as is then at view level do something like?
```
<%= "Q#{@question_number} - #{@header}" %>
```
plus any additional null guards

you get the benefit of not coupling the header to the question number

kosiakkatrina: Yeah, I can see that

PR #1354, opened 2023-02-27
CLDC-1808 add previous tenure question for buyer two

arfacamble:
https://digital.dclg.gov.uk/jira/browse/CLDC-1808


app/models/form/sales/pages/buyer2_previous_housing_situation.rb
```diff
@@ -0,0 +1,11 @@
+class Form::Sales::Pages::Buyer2PreviousHousingSituation < ::Form::Page
+  def initialize(id, hsh, subsection)
+    super
+    @id = "buyer_2_previous_housing_situation"
+    @depends_on = [{ "buyer_two_not_already_living_in?" => true }]
```
kosiakkatrina: `buyer_two_not_already_living_in` sounds like they're living in the new property, but from what I'm gathering it's buyers lived together before this sale? Fine if it makes sense to you, but I didn't quite get the meaning of the method without having to go and look what it actually checks

spec/models/form/sales/questions/buyer2_living_in_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+  it "has correct page" do
+    expect(question.page).to eq(page)
+  end
+
+  it "has the correct id" do
+    expect(question.id).to eq("buy2living")
+  end
+
+  it "has the correct header" do
+    expect(question.header).to eq("Were all buyers living at the same address at the time of purchase?")
```
jamesrose: Does this title accurately match the name of the variable `buy2living` here?

app/models/sales_log.rb
```diff
@@ -132,6 +132,14 @@ def income1_used_for_mortgage?
     inc1mort == 1
   end
 
+  def buyer_two_will_live_in_property?
+    buy2livein == 1
+  end
```
jamesrose, line 156: The title of this question is: Were all buyers living at the same address at the time of purchase?

Does answering "Yes" here match this method name?

arfacamble:
The names of things in this PR are generally a bit confusing.
I don't believe I have much choice over that but I may have misunderstood what my constraints are.
@kosiakkatrina @jamesrose I will reply to each of your comments, but you may need to read the answers to all of them to understand my answer to each!

The flow:
in household characteristics we ask "Will buyer 2 live in the property?" (buy2livein)
if yes (`buyer_two_will_live_in_property?`), in household situation we ask "Were all buyers living at the same address at the time of purchase?" (buy2living)
if no(`buyer_two_not_already_living_in?`), we ask for buyer 2 previous tenure

app/models/sales_log.rb
arfacamble: There are two confusingly similarly named sales log fields here called buy2living which is the answer to the question you quote here. and buy2livein which refers to the question "Will buyer 2 live in the property?"

I believe it is strongly recommended against the fields having more readable names, so my instinct is to wrap/hide that obscurely names field inside descriptive methods as soon as possible

app/models/form/sales/pages/buyer2_previous_housing_situation.rb
arfacamble: We should ask about buyer 2's previous housing situation if we do not know it, which is if buyer 2 is not already living in the house that is being bought.
I didn't find it obvious what to call this method, so I'm open to looking for a better name, considering it has failed the acid test of my fellow dev being able to easily read what it means.
I considered `buyer_two_already_living_in => false` (`buy2living == 1`), but I believe that if `buy2living => nil` and they navigate to the check answers page they would then see a question they shouldn't

spec/models/form/sales/questions/buyer2_living_in_spec.rb
arfacamble: I believe this is answered in response to other comment about 2 very similar looking fields buy2livein and buy2living

app/models/form/sales/pages/buyer2_previous_housing_situation.rb
kosiakkatrina: "At the time of purchase, was buyer 2 living at the same address as buyer 1?" `buyer_two_lived_at_the_same_address?` maybe?

arfacamble: hello. This has already changed a bit after chatting with James to be `buyer_two_not_already_living_in_property?` what do you think about that?
As per previous comment, unless you disagree, I think that the method name has to be phrased in the negative, which is ever so slightly more awkward.
If you don't like `buyer_two_not_already_living_in_property?` then it could be `buyer_two_did_not_live_at_the_same_address?`, I'm fine with that if you prefer it, I find them equivalent.

kosiakkatrina: I don't mind what it it right now 👍

app/models/form/sales/subsections/household_situation.rb
```diff
@@ -12,6 +12,16 @@ def pages
       Form::Sales::Pages::LastAccommodation.new(nil, nil, self),
       Form::Sales::Pages::LastAccommodationLa.new(nil, nil, self),
       Form::Sales::Pages::BuyersOrganisations.new(nil, nil, self),
-    ]
+      buyer_2_situation_pages,
+    ].flatten.compact
```
natdeanlewissoftwire, line 16: not a suggestion but thought this might be inefficient, turns out it's fine, interesting graph: https://stackoverflow.com/questions/34400670/flatten-and-compact-an-array-more-efficiently

natdeanlewissoftwire (approved):

natdeanlewissoftwire:
lgtm! (assuming you've dealt with James + Kat's points)

PR #1357, opened 2023-02-28
Blank out incompatible data when importing from CDS

jamesrose:
While importing data from CDS we might encounter hard validations where the new service is more strict. We do have a pattern for blanking values where they are incompatible so that the user can enter them properly. This expands on that pattern for three new conditions described below.

This covers the following errors:

- Where the income is 0, set earnings and income to blank and set incref to refused
- Removing invalid tenancylength and tenancy values where tenancylength is invalid
- Removing prevten and age1 where incompatible

### DIFF
--- a/app/models/validations/financial_validations.rb
+++ b/app/models/validations/financial_validations.rb
@@ -24,11 +24,11 @@ module Validations::FinancialValidations
   def validate_net_income(record)
     if record.ecstat1 && record.weekly_net_income
       if record.weekly_net_income > record.applicable_income_range.hard_max
-        record.errors.add :earnings, I18n.t("validations.financial.earnings.over_hard_max", hard_max: record.applicable_income_range.hard_max)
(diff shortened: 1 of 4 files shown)


kosiakkatrina (approved):

app/models/validations/financial_validations.rb
```diff
@@ -24,11 +24,11 @@ def validate_net_income_uc_proportion(record)
   def validate_net_income(record)
     if record.ecstat1 && record.weekly_net_income
       if record.weekly_net_income > record.applicable_income_range.hard_max
-        record.errors.add :earnings, I18n.t("validations.financial.earnings.over_hard_max", hard_max: record.applicable_income_range.hard_max)
+        record.errors.add :earnings, :over_hard_min, message: I18n.t("validations.financial.earnings.over_hard_max", hard_max: record.applicable_income_range.hard_max)
```
davidm-m: Guessing that's supposed to be over_hard_**max**?

jamesrose: Great spot. Fixed!

PR #1336, opened 2023-02-23
CLDC-1813 when buyer is a company, do not route to household characteristics

arfacamble:

PR #1368, opened 2023-03-02
CLDC-2073 app crashing if needs type is changed

arfacamble:
https://digital.dclg.gov.uk/jira/browse/CLDC-2073

There are comments in line with some changes. A broader question - I would expect that postcode_known would be set to nil when the relevant question is no longer routed to. It's not immediately clear why it isn't or whether that definitely is desired behaviour.
Open to discussion on whether the solution provided is enough or whether that also warrants further investigation.


app/models/form.rb
```diff
@@ -111,8 +111,8 @@ def next_incomplete_section_redirect_path(subsection, log)
     when :in_progress
       "#{next_subsection.id}/check_answers".dasherize
     when :not_started
-      first_question_in_subsection = next_subsection.pages.find { |page| page.routed_to?(log, nil) }.id
-      first_question_in_subsection.to_s.dasherize
+      first_question_in_subsection = next_subsection.pages.find { |page| page.routed_to?(log, nil) }
+      first_question_in_subsection ? first_question_in_subsection.id.to_s.dasherize : next_incomplete_section_redirect_path(next_subsection, log)
```
arfacamble, line 115: the is the site of the original error found.
It is possible that in the next subsection there are not questions routed to, in this case we should link the next subsection that does have questions routed to.

app/models/form/lettings/subsections/property_information.rb
```diff
@@ -27,4 +27,8 @@ def pages
       Form::Lettings::Pages::PropertyMajorRepairsValueCheck.new(nil, nil, self),
     ].compact
   end
+
+  def displayed_in_tasklist?(log)
+    !log.supported_housing_and_renewal?
+  end
```
arfacamble, line 33: This was the location of another error.
If we answer postcode_known and then alter the setup such that none of the questions in property information are routed to, tasklist.html.erb (l.12) tries to render the subsection as there is a question now considered applicable, however it then tries to make a link to the first question in that section that is routed to (none)
I have prevented that crash here by failing an earlier condition, but we may want to add safeguards elsewhere also, eg. in tasklist_helper:subsection_link

app/models/form/lettings/subsections/property_information.rb
```diff
@@ -3,7 +3,7 @@ def initialize(id, hsh, section)
     super
     @id = "property_information"
     @label = "Property information"
-    @depends_on = [{ "non_location_setup_questions_completed?" => true }]
+    @depends_on = [{ "non_location_setup_questions_completed?" => true, "supported_housing_and_renewal?" => false }]
```
kosiakkatrina: Hm, so this depends_on is used to determine if the section should be unlocked or not. Is this the correct place to put `supported_housing_and_renewal?` condition?
I'd imagine we don't want to display it at all/don't want to route to any of the pages in this section as opposed to just lock it

arfacamble: I'm not sure I follow. Do you mean having the displayed_in_tasklist? method below should be enough?

arfacamble: Also relevantly I went through all the pages in the section and they all depend on being general needs or not renewal, although perhaps that could change in future?

app/models/form.rb
kosiakkatrina: Hmm, I wonder if it would make sense so that `next_subsection` returns the next subsection that is routed to, but I'm not sure where else this method is used, so happy for you to go with what makes more sense

app/models/form/lettings/subsections/property_information.rb
arfacamble: if this has a smell for you then perhaps the "proper" solution would be to make sure postcode_known is set to nil, did you see that comment? and do you know if setting it to nil definitely would be desired behaviour once it's no longer routed to?

kosiakkatrina: Yeah I'm not sure in terms of setting the postcode known. I'm also not a 100% sure why it being false impacts anything. 
Also, what is this line meant to achieve? Sorry, I'm just confused

app/models/form/lettings/pages/first_time_property_let_as_social_housing.rb
```diff
@@ -2,7 +2,7 @@ class Form::Lettings::Pages::FirstTimePropertyLetAsSocialHousing < ::Form::Page
   def initialize(id, hsh, subsection)
     super
     @id = "first_time_property_let_as_social_housing"
-    @depends_on = [{ "renewal" => 0 }]
+    @depends_on = [{ "not_renewal?" => true }]
```
jamesrose:
```suggestion
    @depends_on = [{ "renewal" => 0 }]
```

```suggestion
    @depends_on = [{ "is_renewal?" => false }]
```

Why do you need a method for this?

app/models/lettings_log.rb
```diff
@@ -201,6 +201,10 @@ def is_renewal?
     renewal == 1
   end
 
+  def not_renewal?
+    renewal.zero?
+  end
```
jamesrose: I don't think this method is required.

app/models/form/lettings/subsections/property_information.rb
jamesrose:
```suggestion
    @depends_on = [{ "non_location_setup_questions_completed?" => true, "is_supported_housing?" => false, "is_renewal?" => false }]
```

Remove the method — just be explicit.

app/models/form/lettings/pages/first_time_property_let_as_social_housing.rb
arfacamble: I considered using `is_renewal? => false` but in theory renewal could have values 1, 0, nil and I would like to target 0 rather than 0 or nil.
However, in practice renewal is a required question in the setup section so practically `is_renewal? => false` is probably fine?

app/models/form/lettings/subsections/property_information.rb
arfacamble: It should show if either supported housing is false or if renewal is false, not only if both are false

arfacamble: https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/1336/files

linking this here for discussion. It appeared to me to be a similar situation, where we don't wish to display an entire subsection, dependent on these conditions

app/models/form/lettings/pages/first_time_property_let_as_social_housing.rb
jamesrose: How about?

```ruby
@depends_on = [{ "renewal" => 0 }]
```

app/models/form/lettings/subsections/property_information.rb
jamesrose: You can stack them for `OR` statements:

```ruby
@depends_on = [
  { "non_location_setup_questions_completed?" => true, "is_supported_housing?" => false },
  { "non_location_setup_questions_completed" => true, "is_renewal?" => false }
]
```

arfacamble: yes I forgot about this.
I will change to this for now until I discuss with Kat whether this should be here at all

app/models/form/lettings/pages/first_time_property_let_as_social_housing.rb
arfacamble: I can do that if you feel strongly. I'm on a small crusade to remove all conditions that look like that that I come across as I find it pretty obscure to know what a particular code represents.
In my dreams we would replace all of them only with things that should be true and @depends_on would be a 2d array rather than an array of hashes, but I'll try not to get too over-excited.

jamesrose (approved):

PR #1372, opened 2023-03-02
CLDC-1938 Fix household members import 

kosiakkatrina:
Logs for 21/22 don't have joint purchase field, so they fail when setting hhmemb during the import. 
We're not going to import those logs, but we don't want the import to fail when testing

### DIFF
--- a/app/services/imports/sales_logs_import_service.rb
+++ b/app/services/imports/sales_logs_import_service.rb
@@ -432,9 +432,11 @@ module Imports
     end
 
     def household_members(_xml_doc, attributes)
-      return attributes["hholdcount"] + 2 if attributes["jointpur"] == 1
-
-      attributes["hholdcount"] + 1 if attributes["jointpur"] == 2
+      if attributes["jointpur"] == 2
+        attributes["hholdcount"] + 1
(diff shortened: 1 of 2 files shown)


app/services/imports/sales_logs_import_service.rb
```diff
@@ -432,9 +432,11 @@ def other_people_with_details(xml_doc, attributes)
     end
 
     def household_members(_xml_doc, attributes)
-      return attributes["hholdcount"] + 2 if attributes["jointpur"] == 1
-
-      attributes["hholdcount"] + 1 if attributes["jointpur"] == 2
+      if attributes["jointpur"] == 2
```
ghost, line 435: do we have a test for this?

kosiakkatrina: Existing tests didn't break 🫠
It's a good point tho, I couldn't think of how to test it, but I've just added a test. 
It basically hangs because it tries to complete other calculations involving hhmemb without it being set. So that test hangs with the old version, but runs fine with the changes in this PR

jamesrose (approved):

PR #1377, opened 2023-03-03
CLDC-2001 Bulk upload stacked errors

asmega:
# Context

- https://digital.dclg.gov.uk/jira/browse/CLDC-2001
- There are times that if there is an field is incorrect several errors are added a result
- We want to reduce these so there is less noise and users can focus on the issues that they can action to resolve problematic data

# Changes

- Only include existing log errors if the field does not already have errors on it. this allows the CSV errors to supersede the existing log errors 
- Shim in additional errors for ages for bulk upload to prevent quirky existing errors from bubbling up 

# Notes for review

- We're now entering more into the territory where existing log validations are simply insufficient and do not work with the desired effect for bulk upload although are applicable for the web flow
- As a result we have to shim in desired validations at CSV level and only include existing validation errors if they shimmed rules are not triggered

### DIFF
--- a/app/services/bulk_upload/lettings/row_parser.rb
+++ b/app/services/bulk_upload/lettings/row_parser.rb
@@ -414,7 +414,7 @@ private
 
       if setup_question?(question)
         fields.each do |field|
-          unless errors.any? { |e| fields.include?(e.attribute) }
+          if errors[field].present?
             errors.add(field, I18n.t("validations.not_answered", question: question.check_answer_label&.downcase), category: :setup)
           end
         end
--- a/app/services/bulk_upload/lettings/row_parser.rb
+++ b/app/services/bulk_upload/lettings/row_parser.rb
@@ -911,6 +911,8 @@ private
       0
     when nil
       rsnvac == 14 ? 1 : 0
+    else
+      field_134
     end
   end
 
--- a/app/services/bulk_upload/lettings/row_parser.rb
+++ b/app/services/bulk_upload/lettings/row_parser.rb
@@ -144,6 +144,15 @@ class BulkUpload::Lettings::RowParser
                       inclusion: { in: (1..12).to_a, message: I18n.t("validations.invalid_option", question: "letting type") }
   validates :field_4, presence: { if: proc { [2, 4, 6, 8, 10, 12].include?(field_1) } }
 
+  validates :field_12, format: { with: /\A\d{1,3}\z|\AR\z/, message: "Age of person 1 must be a number or the letter R" }
+  validates :field_13, format: { with: /\A\d{1,3}\z|\AR\z/, message: "Age of person 2 must be a number or the letter R" }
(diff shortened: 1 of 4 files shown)


app/services/bulk_upload/lettings/row_parser.rb
```diff
@@ -387,7 +408,11 @@ def validate_nulls
       next if log.optional_fields.include?(question.id)
       next if question.completed?(log)
 
-      fields.each { |field| errors.add(field, I18n.t("validations.not_answered", question: question.check_answer_label&.downcase)) }
+      fields.each do |field|
+        unless errors.any? { |e| fields.include?(e.attribute) }
```
kosiakkatrina: What is this doing?

spec/services/bulk_upload/lettings/row_parser_spec.rb
```diff
@@ -531,6 +543,18 @@
         end
       end
 
+      context "when one of these fields is blank" do
```
kosiakkatrina, line 546: Would it make sense to have a more descriptive test name? I assume this is a compound validation that would otherwise set an error on each field?

app/services/bulk_upload/lettings/row_parser.rb
asmega: - yep so before if there was an error it was always get added to the relevant fields
- now if any one of the relevant fields has one error on it we no longer keeping adding errors to that grouping

eg `startdate` consists of `field_dd`, `field_mm`, `field_yy` from the mapping
if there is an error that already on `field_dd` the code will no longer surface any additional errors on those fields. this should prevent the user from being overwhelmed by too many errors

app/services/bulk_upload/lettings/row_parser.rb
```diff
@@ -337,7 +356,7 @@ def validate_leaving_reason_for_renewal
   end
 
   def validate_relevant_collection_window
-    return unless start_date && bulk_upload.form
+    return if start_date.blank? || bulk_upload.form.blank?
```
natdeanlewissoftwire, line 363: I'm with you on not using unless UNLESS it's actually useful, which I agree it isn't here

natdeanlewissoftwire (approved):
Looks good, my domain knowledge on bulk upload isn't as strong so I'm deferring to yours as to why these validations are necessary to be separate, why we're allowing R for age etc.

asmega:
> Looks good, my domain knowledge on bulk upload isn't as strong so I'm deferring to yours as to why these validations are necessary to be separate, why we're allowing R for age etc.

Yeah for bulk upload some of the fields are multi-purpose. eg in this case `R` denotes they don't know. this means they must be explicit is passing this information and leaving it blank it considered that they forgot to answer the question. i didn't have input on the format/spec so it's just the agreed way that it is

they are separate as they need to sit on top of the existing `Log` validation which doesn't quite cover all the scenarios of bulk upload such as the `R`

it works but yes i don't think its particularly elegant

PR #1378, opened 2023-03-03
CLDC-1917 Add Startdate Validation

natdeanlewissoftwire:
This adds the startdate validation for crossover and non-crossover periods. I've added a feature flag to stop this validation causing failures in unrelated tests that have assumed the startdate to be a certain time. There were almost 400 of these, and fixing them one by one was a time sink that @kosiakkatrina and I agreed wasn't worth getting sucked into - I had it down to 320 or so but progress was slowing so opted for the feature flag approach instead.
image

A lot of the tests I've fixed for this ticket assumed time was in 22/23 collection year (sometimes in 21/22 collection year). I've enforced this to be true to fix tests and avoid testing overlapping behaviour (some tests relied on features from both collection years simultaneously), but this means we should probably add some more tests for the 23/24 collection year/current time than there currently are. I'd rather get this merged and make that a separate ticket though for size/ease of review reasons however.


app/models/form_handler.rb
```diff
@@ -79,12 +79,12 @@ def in_crossover_period?(now: Time.zone.now)
 
   def lettings_in_crossover_period?(now: Time.zone.now)
     forms = lettings_forms.values
-    forms.count { |form| form.start_date < now && now < form.end_date } > 1
+    forms.count { |form| form.start_date <= now && now < form.end_date } > 1
```
natdeanlewissoftwire: think this was a bug that got missed previously, can't think of a reason this shouldn't be <=

kosiakkatrina: `.between?` ?

config/locales/en.yml
```diff
@@ -156,6 +156,10 @@ en:
           "Enter a date within the %{previous_start_year_short}/%{previous_end_year_short} or %{previous_end_year_short}/%{current_end_year_short} financial years, which is between %{previous_start_year_long} and %{current_end_year_long}"
 
       startdate:
+        current_financial_year:
+          Enter a date within the %{current_start_year_short}/%{current_end_year_short} financial year, which is between %{current_start_year_long} and %{current_end_year_long}
```
jamesrose: > Enter a date within the %{current_start_year_short}/%{current_end_year_short} financial year

Is that a financial year? Is there a reason we are specific about it being one here?

config/initializers/feature_toggle.rb
```diff
@@ -4,11 +4,11 @@ def self.saledate_collection_window_validation_enabled?
     Rails.env.production? || Rails.env.test? || Rails.env.staging?
   end
 
-  def self.startdate_two_week_validation_enabled?
-    Rails.env.production? || Rails.env.test? || Rails.env.staging?
+  def self.startdate_collection_window_validation_enabled?
+    !Rails.env.test?
   end
```
jamesrose: This is a bad idea. Our tests should reflect the reality of the logic in production :)

natdeanlewissoftwire:
@jamesrose to advise as to whether we want to go through the best-practice process of making all previous time-assuming tests pass, or whether the current feature flag fix is acceptable (with tech debt to come back to it later)

Update: Decision is to fix tests.

config/locales/en.yml
natdeanlewissoftwire: Good point, it's not actually a financial year, I think we assume in copy that "financial year" for our purposes just means 1st April -> 31st March. I'll check with design.

app/models/form_handler.rb
natdeanlewissoftwire: between is inclusive at both ends of a range I believe, think I could use ... though

natdeanlewissoftwire: Ah no ignore me, between is fine here I think, thanks!

config/locales/en.yml
natdeanlewissoftwire: changed to collection year ^

app/models/validations/setup_validations.rb
```diff
@@ -1,5 +1,14 @@
 module Validations::SetupValidations
   include Validations::SharedValidations
+  include CollectionTimeHelper
+
+  def validate_startdate_setup(record)
+    return unless record.startdate && date_valid?("startdate", record)
+
+    unless record.startdate.between?(active_collection_start_date, current_collection_end_date) || !FeatureToggle.startdate_collection_window_validation_enabled?
```
asmega: minor thing, can:
```
!FeatureToggle.startdate_collection_window_validation_enabled?
```
be made into a short circuit at the top of the method eg:
```
return if !FeatureToggle.startdate_collection_window_validation_enabled?
```
if so, it should read more easily?

app/models/validations/setup_validations.rb
```diff
@@ -27,6 +36,37 @@ def validate_organisation(record)
...
 
+  def active_collection_start_date
+    if FormHandler.instance.lettings_in_crossover_period?
+      previous_collection_start_date
+    else
+      current_collection_start_date
+    end
+  end
+
+  def validation_error_message
```
asmega: i'd probably opt for a less generic name. because this apps loves modules, everything will get mixed into one class therefore to prevent collisions a more specific name may be wiser

config/locales/en.yml
```diff
@@ -141,7 +141,7 @@ en:
       above_min: "%{field} must be at least %{min}"
     date:
       invalid_date: "Enter a date in the correct format, for example 31 1 2022"
-      outside_collection_window: Enter a date within the 22/23 financial year, which is between 1st April 2022 and 31st March 2023
+      outside_collection_window: Enter a date within the 22/23 collection year, which is between 1st April 2022 and 31st March 2023
```
asmega, line 144: may be for next time i think it would be great it we could pass these values in

spec/models/lettings_log_spec.rb
```diff
@@ -1497,6 +1520,23 @@ def previous_postcode_fields
...
+            after do
+              Timecop.unfreeze
+            end
+
+            around do |example|
+              Timecop.freeze(Time.zone.local(2023, 5, 1)) do
+                Singleton.__init__(FormHandler)
+                example.run
+              end
+              Timecop.return
```
asmega, line 1536: here i don't think you will need the `before` and `after` as you've covered this in the `around`. also this line should be able to go? as the `freeze` should only happen to anything with the block

EDIT: there seem to be other occurrences of this same issue in this PR

spec/requests/lettings_logs_controller_spec.rb
```diff
@@ -333,6 +333,14 @@
           end
 
           context "with year filter" do
+            before do
+              Timecop.freeze(2022, 3, 1)
```
asmega: personally i would just change all these to use `around` and switch to the block syntax, maybe that a preference thing though

asmega (approved):

app/models/validations/setup_validations.rb
natdeanlewissoftwire: good idea, will do this for sales too 👍

config/locales/en.yml
natdeanlewissoftwire: this is only used for bulk upload, but yep I agree - probably a separate ticket

app/models/validations/setup_validations.rb
natdeanlewissoftwire: good shout, will do for sales logs too 👍

spec/models/lettings_log_spec.rb
natdeanlewissoftwire: These are actually required, it's not ideal but the around wraps around global before and afters which is required for initialising logs that are reused in different time periods.

spec/requests/lettings_logs_controller_spec.rb
natdeanlewissoftwire: some of these are necessary but have changed the ones that can be

PR #1379, opened 2023-03-03
CLDC-1938 Import LA and mortgage used fixes

kosiakkatrina:
- Import `staircase` if it's shared ownership only, because it is only asked for shared ownership
- Import LA manually if it cannot be inferred, this will likely result in differences when importing if it can be inferred, but no data loss
- Infer mortgage used as yes if any of the mortgage information is given, as we don't want to lose that data
- Only import 22/23 sales collection onwards, because we don't have 21/22 form on our service and it is not necessary to import those logs from the data perspective

### DIFF
--- a/app/services/imports/sales_logs_import_service.rb
+++ b/app/services/imports/sales_logs_import_service.rb
@@ -110,7 +110,7 @@ module Imports
       attributes["prevten"] = unsafe_string_as_integer(xml_doc, "Q6PrevTenure")
       attributes["mortlen"] = mortgage_length(xml_doc, attributes)
       attributes["extrabor"] = borrowing(xml_doc, attributes)
-      attributes["mortgageused"] = unsafe_string_as_integer(xml_doc, "MORTGAGEUSED")
+      attributes["mortgageused"] = mortgage_used(xml_doc, attributes)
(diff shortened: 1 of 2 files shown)


app/services/imports/sales_logs_import_service.rb
```diff
@@ -16,7 +16,9 @@ def create_logs(folder)
   private
 
     def create_log(xml_doc)
+      # only import sales logs from 22/23 collection period onwards
       return unless meta_field_value(xml_doc, "form-name").include?("Sales")
+      return unless compose_date(xml_doc, "DAY", "MONTH", "YEAR") > Time.zone.local(2022, 4, 1)
```
jamesrose: Is it possible to infer this date somehow?

kosiakkatrina: Hm, I thought about that, but not really. I think our form dates are based on `current collection period`, so even if I got the `previous collection start year` it would change in a month. This should always be the same tho

kosiakkatrina: It should be >= tho, so let me change that

jamesrose (approved):

PR #1381, opened 2023-03-06
CLDC-1864 single log numbering

natdeanlewissoftwire:
adds question numbering to CYA pages and single question pages for lettings and sales logs (also means numbers show in review page which isn't mentioned in the ticket but is the correct behaviour to my knowledge):

Questions not numbered as they are being destructively changed/added in other tickets: Lettings 11 and 12, Sales 14, 15, 32, 60 and 61

image
image

review page: 
image


app/models/form/question.rb
```diff
@@ -258,6 +259,12 @@ def is_derived_or_has_inferred_check_answers_value?(log)
     selected_answer_option_is_derived?(log) || has_inferred_check_answers_value?(log)
   end
 
+  def question_number_string(conditional)
```
asmega: can we give this a default? it'll save you from passing in flags all the time

natdeanlewissoftwire:
One thing I should add is enforcing this behaviour only for 23/24 - as some qs are different, thought it wouldn't be an issue but it looks like it is

Update: done

app/models/form/lettings/questions/age.rb
```diff
@@ -11,5 +11,6 @@ def initialize(id, hsh, page, person_index:)
     @max = 120
     @min = 0
     @step = 1
+    @question_number = 31 + 4 * person_index
```
asmega: i think i would of added brackets here so its obvious where the multiplication happens

asmega (approved):

app/models/form/lettings/questions/age.rb
natdeanlewissoftwire: sure, good idea!

PR #1387, opened 2023-03-07
CLDC-1862 Copy change for local authority known question

asmega:
# Context

- https://digital.dclg.gov.uk/jira/browse/CLDC-1862

# Changes

- Copy changes for LA known question

### DIFF
--- a/app/models/form/sales/questions/property_local_authority_known.rb
+++ b/app/models/form/sales/questions/property_local_authority_known.rb
@@ -3,7 +3,7 @@ class Form::Sales::Questions::PropertyLocalAuthorityKnown < ::Form::Question
     super
     @id = "la_known"
     @check_answer_label = "Local authority known"
-    @header = "Do you know the local authority of the property?"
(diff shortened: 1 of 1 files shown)


app/models/form/sales/questions/property_local_authority_known.rb
```diff
@@ -3,7 +3,7 @@ def initialize(id, hsh, page)
     super
     @id = "la_known"
     @check_answer_label = "Local authority known"
-    @header = "Do you know the local authority of the property?"
+    @header = "Do you know the property’s local authority?"
```
arfacamble, line 6: won't there be a test file for this given it's in sales?

jamesrose (approved):

app/models/form/sales/questions/property_local_authority_known.rb
asmega: nothing failed so it's not covered. do we need tests to cover copy?

PR #1392, opened 2023-03-08
CLDC-2006 Add 23/24 collection year LAs

kosiakkatrina:
Add new LAs for 23/24 and remove LAs for 22/23 in:

- Lettings property information section: Local authority of current property
- Lettings Household Situation: local authority of previous property
- Sales property information section: Local authority of current property
- Sales Household Situation: local authority of previous property
- Scheme Locations: What is the local authority of [postcode]? - based on today's date

List of local authority changes is in [CLDC-2006](https://digital.dclg.gov.uk/jira/browse/CLDC-2006)


app/models/form/sales/questions/prevloc.rb
```diff
@@ -14,390 +13,396 @@ def initialize(id, hsh, page)
...
-    "W92000004" => "Wales",
-    "9300000XX" => "Outside UK",
-  }.freeze
+  def answer_options
+    {
+      "" => "Select an option",
+      "S12000033" => "Aberdeen City",
+      "S12000034" => "Aberdeenshire",
+      "E07000223" => "Adur",
+      "E07000026" => ("Allerdale" if form.start_date.year < 2023),
```
jamesrose: How would you do this if we didn't have local authorities statically defined, and they were in the database instead?

kosiakkatrina: I guess similar to what I've done in locations. Have the list of all and then filter down based on the year. I can change that here too if that seems more appropriate

jamesrose: What do you think about something like this? We might use fixed dates instead of years, but the principle remains the same.

```ruby
LocalAuthority.active(2022) 
# or
LocalAuthority.active(Time.today)
```

```diff
diff --git a/app/models/local_authority.rb b/app/models/local_authority.rb
new file mode 100644
index 00000000..0014f357
--- /dev/null
+++ b/app/models/local_authority.rb
@@ -0,0 +1,3 @@
+class LocalAuthority < ApplicationRecord
+  scope :active, ->(year) { where("? BETWEEN start_year AND end_year", year) }
+end
diff --git a/db/migrate/20230308091405_create_local_authorities.rb b/db/migrate/20230308091405_create_local_authorities.rb
new file mode 100644
index 00000000..3581ce7b
--- /dev/null
+++ b/db/migrate/20230308091405_create_local_authorities.rb
@@ -0,0 +1,12 @@
+class CreateLocalAuthorities < ActiveRecord::Migration[7.0]
+  def change
+    create_table :local_authorities do |t|
+      t.string :code
+      t.string :name
+      t.integer :start_year
+      t.integer :end_year
+
+      t.timestamps
+    end
+  end
+end
```

app/models/form/lettings/questions/la.rb
```diff
@@ -7,326 +7,9 @@ def initialize(id, hsh, page)
...
-    "E08000031" => "Wolverhampton",
-    "E07000237" => "Worcester",
-    "E07000229" => "Worthing",
-    "E07000238" => "Wychavon",
-    "E07000128" => "Wyre",
-    "E07000239" => "Wyre Forest",
-    "E06000014" => "York",
-  }.freeze
+  def answer_options
+    { "" => "Select an option" }.merge(LocalAuthority.active(Time.zone.local(2023, 4, 1)).england.map { |la| [la.code, la.name] }.to_h)
```
arfacamble: Do I understand correctly that this is getting all the LA's that are active at a given time?
Therefore wouldn't we want the date here to be something more meaningful? Maybe the lettings start date?

kosiakkatrina: You're right, I replaced it with form start date

kosiakkatrina: You're right, replaced it with a form start date

app/models/form/lettings/questions/la.rb
```diff
@@ -7,326 +7,9 @@ def initialize(id, hsh, page)
...
-    "E07000217" => "Woking",
-    "E06000041" => "Wokingham",
-    "E08000031" => "Wolverhampton",
-    "E07000237" => "Worcester",
-    "E07000229" => "Worthing",
-    "E07000238" => "Wychavon",
-    "E07000128" => "Wyre",
-    "E07000239" => "Wyre Forest",
-    "E06000014" => "York",
-  }.freeze
```
natdeanlewissoftwire, line 331: Glad to see these go, may they never return 🥳

natdeanlewissoftwire (approved):
lgtm!

PR #1395, opened 2023-03-08
CLDC-2006 Add and populate local authorities table

kosiakkatrina:
add and populate local authorities table
this will be used for populating answer options for LA questions for different years instead of hardcoding them in the service


app/services/imports/local_authorities_service.rb
```diff
@@ -0,0 +1,26 @@
...
+    def initialize(path:)
+      @path = path
+      @count = 0
+    end
+
+    def call
+      CSV.foreach(path, headers: true) do |row|
+        LocalAuthority.upsert(
+          { code: row["code"],
+            la_name: row["la_name"],
```
jamesrose:
```suggestion
            name: row["la_name"],
```

db/schema.rb
```diff
@@ -289,6 +289,17 @@
     t.index ["updated_by_id"], name: "index_lettings_logs_on_updated_by_id"
   end
 
+  create_table "local_authorities", force: :cascade do |t|
+    t.string "code"
+    t.string "la_name"
```
jamesrose:
```suggestion
    t.string "name", null: false
```

db/schema.rb
```diff
@@ -289,6 +289,17 @@
...
+    t.datetime "start_date"
```
jamesrose:
```suggestion
    t.datetime "start_date", null: false
```

db/schema.rb
```diff
@@ -289,6 +289,17 @@
     t.index ["updated_by_id"], name: "index_lettings_logs_on_updated_by_id"
   end
 
+  create_table "local_authorities", force: :cascade do |t|
+    t.string "code"
```
jamesrose:
```suggestion
    t.string "code", null: false
```

app/services/imports/local_authorities_service.rb
```diff
@@ -0,0 +1,26 @@
...
+            start_date: Time.zone.local(row["start_year"], 4, 1),
+            end_date: (Time.zone.local(row["end_year"], 3, 31) if row["end_year"]),
+            previous_location_only: row["previous_location_only"] || false },
```
jamesrose: What does this field represent?

kosiakkatrina: Well, some of the fields are non England as we discussed earlier, so they shouldn't be shown in current la questions but should be shown in previous la questions. I figured marking them in the db would be the easiest way to filter them out when needed

jamesrose (approved):

PR #1396, opened 2023-03-08
CLDC-1687 allow deactivation or reactivation of last year schemes and locations in crossover period

arfacamble:
https://digital.dclg.gov.uk/jira/browse/CLDC-1687

this shows a scheme that has existed since 2020, it now cannot be deactivated before the oldest open collection period
|image
|-

In this screenshot it is currently April 2023.
This scheme was started in the 23/24 collection period, so despite the other copy on the page, it cannot be deactivated before the collection period in which it was created.
|image
|-


app/helpers/collection_time_helper.rb
```diff
@@ -44,4 +44,16 @@ def previous_collection_start_year
...
+    # generally the crossover period ends on the first Friday in June, but there may be arbitrary exceptions such as in 2023
+    today = Time.zone.today
+    if today.year == 2023
+      today.between?(Time.zone.local(2023, 4, 1).to_date, Time.zone.local(2023, 6, 9).to_date)
+    else
+      crossover_end_date = Time.zone.local(today.year, 6, 1).to_date
+      crossover_end_date += 1 until crossover_end_date.friday?
+      today.between?(Time.zone.local(today.year, 4, 1).to_date, crossover_end_date)
+    end
+  end
```
jamesrose: Is this the same as `FormHandler#in_crossover_period?`?

arfacamble: aha. hadn't seen that.
very good.
Can I get a sense check though as this might be important for the future depending on what we're using this method for (the FormHandler one). I would expect that that method would return false for all times on the final day of the crossover period.
i.e. Time.zone.now returns the time now, but the end_date for the forms appears to be 00:00 on that day, so now < end_date will be false.
Shall I make a tech debt/bug ticket to change to now.to_date <= end_date.to_date?
I am fairly confident in that change, but I may not want to slow down a 23/24 ticket at this time writing tests.

app/models/location_deactivation_period.rb
```diff
@@ -24,6 +26,8 @@ def validate_reactivation(record, recent_deactivation, location)
   end
 
   def validate_deactivation(record, location)
+    earliest_possible_deactivation = FormHandler.instance.in_crossover_period? ? previous_collection_start_date : current_collection_start_date
```
jamesrose: We're duplicating this logic four times. Could this live on the form handler itself? What would be an appropriate name to describe it?

arfacamble: knock and the door shall be opened unto you...
(I agree and did this thing)

app/models/location_deactivation_period.rb
```diff
@@ -32,10 +34,10 @@ def validate_deactivation(record, location)
       end
     elsif location.location_deactivation_periods.any? { |period| period.reactivation_date.present? && record.deactivation_date.between?(period.deactivation_date, period.reactivation_date - 1.day) }
       record.errors.add(:deactivation_date, message: I18n.t("validations.location.deactivation.during_deactivated_period"))
-    else
-      unless record.deactivation_date.between?(location.available_from, Time.zone.local(2200, 1, 1))
-        record.errors.add(:deactivation_date, message: I18n.t("validations.location.toggle_date.out_of_range", date: location.available_from.to_formatted_s(:govuk_date)))
-      end
+    elsif record.deactivation_date.before? FormHandler.instance.start_date_of_earliest_open_collection_period
```
SamSeed-Softwire, line 37: Do we not want to write tests for this new case?

app/models/scheme_deactivation_period.rb
```diff
@@ -32,10 +34,10 @@ def validate_deactivation(record, scheme)
       end
     elsif scheme.scheme_deactivation_periods.any? { |period| period.reactivation_date.present? && record.deactivation_date.between?(period.deactivation_date, period.reactivation_date - 1.day) }
       record.errors.add(:deactivation_date, message: I18n.t("validations.scheme.deactivation.during_deactivated_period"))
-    else
-      unless record.deactivation_date.between?(scheme.available_from, Time.zone.local(2200, 1, 1))
-        record.errors.add(:deactivation_date, message: I18n.t("validations.scheme.toggle_date.out_of_range", date: scheme.available_from.to_formatted_s(:govuk_date)))
-      end
+    elsif record.deactivation_date.before? FormHandler.instance.start_date_of_earliest_open_collection_period
```
SamSeed-Softwire, line 37: Do we not want to write tests for this new case?

SamSeed-Softwire (commented):
All good except for potentially needing some tests for new error cases, I'm guessing there might be a good reason they're not required though!

app/models/location_deactivation_period.rb
arfacamble: I certainly considered this, but the existing validations did not yet have tests so I didn't add a whole new file.
Given we need to get this out ASAP, @jamesrose would you prefer:
- no tests
- tests just for the things I've added
- tests for all the validations here, do them now
- tests for all the validations here, raise a tech debt ticket for the next/future sprint

jamesrose (approved):

PR #1397, opened 2023-03-08
CLDC-1998 Correctly infer 22/23 las if a NEW one is returned by postcodes.io

natdeanlewissoftwire:
If postcodes.io returns a NEW 23/24 La but the log is for 22/23, set LA as nil and make user manually select LA


app/models/lettings_log.rb
```diff
@@ -593,7 +593,7 @@ def process_postcode(postcode, postcode_known_key, la_inferred_key, la_key)
     self[postcode_known_key] = 1
     inferred_la = get_inferred_la(postcode)
     self[la_inferred_key] = inferred_la.present?
-    self[la_key] = inferred_la if inferred_la.present?
+    self[la_key] = inferred_la
```
natdeanlewissoftwire: This is so if a postcode is changed to a different LA that can't be inferred (or is a new 2023 LA) we clear the old LA as it no longer matches the postcode

app/models/log.rb
```diff
@@ -137,7 +137,9 @@ def process_previous_postcode_changes!
 
   def get_inferred_la(postcode)
     result = PIO.lookup(postcode)
-    result[:location_code] if result
+    if result && !(form.start_date.year < 2023 && new_2023_admin_district_codes.include?(result[:location_code]))
```
natdeanlewissoftwire: only add LA to log if it's not a new 2023 LA being added to a non-2023 log, otherwise make it nil and the manual LA input page displays

jamesrose: Can you think of a way of doing this that doesn't hard-code conditions for 2023 and a method explicitly for 2023?

app/models/log.rb
```diff
@@ -135,9 +135,32 @@ def process_previous_postcode_changes!
...
+    "E07000188" => "E06000066", # Sedgemoor => Somerset
+    "E07000246" => "E06000066", # Somerset West and Taunton => Somerset
+    "E07000189" => "E06000066", # South Somerset => Somerset
+  }.freeze
+
   def get_inferred_la(postcode)
     result = PIO.lookup(postcode)
-    result[:location_code] if result
+    location_code = result[:location_code] if result
+    unless LA_CHANGES.value?(location_code) && form.start_date.year < 2023
```
jamesrose: It’s a bit difficult to parse this because it involves dates. Could you change it to an `if ... `?

natdeanlewissoftwire: ah yep, there are actually relevant conflicts that affect this, updated code incoming...

jamesrose (approved):

PR #1400, opened 2023-03-09
CLDC-2025 Add equity range validation to sales logs

natdeanlewissoftwire:
Adds min and max validations for equity, with ranges specific to type and collection year

image

https://digital.dclg.gov.uk/jira/browse/CLDC-2025


app/models/validations/sales/financial_validations.rb
```diff
@@ -79,4 +100,26 @@ def is_relationship_child?(relationship)
...
+    },
+    2023 => {
+      2 => { "min" => 10, "max" => 75 },
+      30 => { "min" => 25, "max" => 75 },
+      18 => { "min" => 25, "max" => 75 },
+      16 => { "min" => 10, "max" => 75 },
+      24 => { "min" => 25, "max" => 75 },
+      # 28 => ranges don't apply
+      31 => { "min" => 0, "max" => 75 },
+      32 => { "min" => 0, "max" => 75 },
```
natdeanlewissoftwire: happy to hear suggestions on a better way to represent this, only differences are the mins for type == 2 and 30 and that the 2023 year has type 32 as well as the 2022 types

app/models/validations/sales/financial_validations.rb
```diff
@@ -70,6 +70,27 @@ def validate_percentage_owned_not_too_much_if_older_person(record)
...
   end
 
+  def validate_equity_in_range_for_year_and_type(record)
+    return unless record.type && record.equity && record.collection_start_year
+
+    if record.collection_start_year == 2022
+      ranges = EQUITY_RANGES_FOR_TYPES_BY_COLLECTION_START_YEAR[2022]
+    elsif record.collection_start_year >= 2023
+      ranges = EQUITY_RANGES_FOR_TYPES_BY_COLLECTION_START_YEAR[2023]
+    end
```
jamesrose:
```suggestion
    ranges = EQUITY_RANGES_BY_YEAR.fetch(record.collection_start_year, DEFAULT_EQUITY_RANGES)
```

app/models/validations/sales/financial_validations.rb
```diff
@@ -79,4 +100,26 @@ def is_relationship_child?(relationship)
...
+      2 => { "min" => 10, "max" => 75 },
+      30 => { "min" => 25, "max" => 75 },
+      18 => { "min" => 25, "max" => 75 },
+      16 => { "min" => 10, "max" => 75 },
+      24 => { "min" => 25, "max" => 75 },
+      # 28 => ranges don't apply
+      31 => { "min" => 0, "max" => 75 },
+      32 => { "min" => 0, "max" => 75 },
+    },
+  }.freeze
```
jamesrose:
```suggestion
  EQUITY_RANGES_BY_YEAR = {
    2022 => {
      2   => 25..75
      30 => 10..75,
      18 => 25..75,
      16 => 10..75,
      24 => 25..75,
      31 => 0..75,
    },
 }
 
 DEFAULT_EQUITY_RANGES =  {
      2   => 10..75,
      30 => 25..75,
      18 => 25..75,
      16 => 10..75,
      24 => 25..75,
      31 => 0..75,
      32 => 0..75,
    },
  }
```

app/models/validations/sales/financial_validations.rb
```diff
@@ -70,6 +70,27 @@ def validate_percentage_owned_not_too_much_if_older_person(record)
...
+    return unless record.type && record.equity && record.collection_start_year
+
+    if record.collection_start_year == 2022
+      ranges = EQUITY_RANGES_FOR_TYPES_BY_COLLECTION_START_YEAR[2022]
+    elsif record.collection_start_year >= 2023
+      ranges = EQUITY_RANGES_FOR_TYPES_BY_COLLECTION_START_YEAR[2023]
+    end
+
+    if ranges.key?(record.type)
+      range = ranges[record.type]
```
jamesrose:
```suggestion
    if range = ranges[record.type]
```

This is the same thing :)

app/models/validations/sales/financial_validations.rb
```diff
@@ -70,6 +70,27 @@ def validate_percentage_owned_not_too_much_if_older_person(record)
...
+      if record.equity < range["min"]
+        record.errors.add :type, I18n.t("validations.financial.equity.under_min", min_equity: range["min"])
+        record.errors.add :equity, I18n.t("validations.financial.equity.under_min", min_equity: range["min"])
+      elsif record.equity > range["max"]
+        record.errors.add :type, I18n.t("validations.financial.equity.over_max")
+        record.errors.add :equity, I18n.t("validations.financial.equity.over_max")
+      end
```
jamesrose:
```suggestion
      if record.equity < range.min
        record.errors.add :type, I18n.t("validations.financial.equity.under_min", min_equity: range.min)
        record.errors.add :equity, I18n.t("validations.financial.equity.under_min", min_equity: range.min)
      elsif record.equity > range.max
        record.errors.add :type, I18n.t("validations.financial.equity.over_max", max_equity: range.max)
        record.errors.add :equity, I18n.t("validations.financial.equity.over_max", max_equity: range.max)
      end
```

Are you sure that these should be `<` and `>` and not `<=` and `>=`?

config/locales/en.yml
```diff
@@ -297,6 +297,9 @@ en:
       staircasing:
         percentage_bought_must_be_greater_than_percentage_owned: "Total percentage buyer now owns must be more than percentage bought in this transaction"
         older_person_percentage_owned_maximum_75: "Percentage cannot be above 75% under Older Person's Shared Ownership"
+      equity:
+        under_min: "The minimum initial equity stake for this type of shared ownership sale is %{min_equity}%"
+        over_max: "The maximum initial equity stake is 75%"
```
```suggestion
        over_max: "The maximum initial equity stake is %{max_equity}%"
```

We're assuming that this won't change here.

app/models/validations/sales/financial_validations.rb
arfacamble: I'm still not sure how I feel about assignment inside a condition. It's very easy to read it in passing as `range == ranges[kxdfs]`

app/models/validations/sales/financial_validations.rb
natdeanlewissoftwire: yep!

app/models/validations/sales/financial_validations.rb
```diff
@@ -85,6 +85,22 @@ def validate_percentage_owned_not_too_much_if_older_person(record)
     end
   end
 
+  def validate_equity_in_range_for_year_and_type(record)
+    return unless record.type && record.equity && record.collection_start_year
+
+    ranges = EQUITY_RANGES_BY_YEAR.fetch(record.collection_start_year, DEFAULT_EQUITY_RANGES)
+
+    if (range = ranges[record.type])
```
arfacamble: nested ifs
could reduce nesting by changing 93 to return unless (range = ranges[record.type])

spec/models/validations/sales/financial_validations_spec.rb
```diff
@@ -266,4 +266,86 @@
...
+  describe "#validate_equity_in_range_for_year_and_type" do
+    let(:record) { FactoryBot.create(:sales_log, saledate: now) }
+
+    before do
+      Timecop.freeze(now)
+    end
+
+    after do
+      Timecop.unfreeze
+    end
```
arfacamble, line 278: the before and after can be swapped for an around
you can search the syntax in the project but something like 
```
before do |test|
Timecop.freeze
test.run
Timecop.unfreeze
end
```

arfacamble (approved):

spec/models/validations/sales/financial_validations_spec.rb
natdeanlewissoftwire: yeah true, in case you're interested an around actually behaves slightly differently to a before/after in terms of when it is called but in this case you're right they're interchangeable

PR #1401, opened 2023-03-09
CLDC-1838 update copy for both offered questions

arfacamble:
ensure copy is changed for both questions relating to how many times a property has been offered

extract to en.yml so that both questions keep identical copy in the future

https://digital.dclg.gov.uk/jira/browse/CLDC-1838

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -560,7 +560,7 @@ en:
       toggle_active: "If the date is before %{date}, select ‘From the start of the current collection period’ because the previous period has now closed."
     bulk_upload:
       needstype: "General needs housing includes both self-contained and shared housing without support or specific adaptations. Supported housing can include direct access hostels, group homes, residential care and nursing homes."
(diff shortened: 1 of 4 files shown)


config/locales/en.yml
```diff
@@ -559,6 +560,7 @@ en:
       toggle_active: "If the date is before %{date}, select ‘From the start of the current collection period’ because the previous period has now closed."
     bulk_upload:
       needstype: "General needs housing includes both self-contained and shared housing without support or specific adaptations. Supported housing can include direct access hostels, group homes, residential care and nursing homes."
+    offered: "Do not include the offer that led to this letting.This is after the last tenancy ended. If the property is being offered for let for the first time, enter 0."
```
natdeanlewissoftwire: Space missing after the .

natdeanlewissoftwire (approved):
little tweak but otherwise ready for staging

config/locales/en.yml
arfacamble: good catch

PR #1402, opened 2023-03-09
CLDC-1999 Add local authority links and select correct LA from location

kosiakkatrina:
- Add a new table to link local authorities from local_authorities table so that we could track which authorities changed into which over the years.

- Display all linked LAs in location information:
image

- Choose the correct LA for supported housing logs based on the year


app/helpers/locations_helper.rb
```diff
@@ -107,4 +107,15 @@ def is_nested?(inner, outer)
...
+  def formatted_local_authority_timeline(location, field)
+    sorted_linked_authorities = location.linked_local_authorities.sort_by(&:start_date)
+    return sorted_linked_authorities.first[field] if sorted_linked_authorities.count == 1
+
+    sorted_linked_authorities.map { |linked_local_authority|
+      formatted_start_date = linked_local_authority.start_date&.to_formatted_s(:govuk_date)
+      formatted_end_date = linked_local_authority.end_date&.to_formatted_s(:govuk_date)
+      "#{linked_local_authority[field]} (#{formatted_start_date} - #{formatted_end_date || 'present'})"
+    }.join("\n")
+  end
```
jamesrose:
```suggestion
  def formatted_local_authority_timeline(location, field)
    sorted_linked_authorities = location.linked_local_authorities.sort(:start_date)
    
    return sorted_linked_authorities.first[field] if sorted_linked_authorities.count.one?

    sorted_linked_authorities.map do |linked_local_authority|
      formatted_start_date = linked_local_authority.start_date&.to_formatted_s(:govuk_date)
      formatted_end_date = linked_local_authority.end_date&.to_formatted_s(:govuk_date) || "present"
      "#{linked_local_authority[field]} (#{formatted_start_date} - #{formatted_end_date})"
    end.join("\n")
  end
```

https://github.com/rubocop/ruby-style-guide#single-line-blocks-delimiters

However! I'd challenge whether we need that return statement:

```suggestion
  def formatted_local_authority_timeline(location, field)
    sorted_linked_authorities = location.linked_local_authorities.sort(:start_date)

    sorted_linked_authorities.map do |linked_local_authority|
      formatted_start_date = linked_local_authority.start_date&.to_formatted_s(:govuk_date)
      formatted_end_date = linked_local_authority.end_date&.to_formatted_s(:govuk_date) || "present"
      "#{linked_local_authority[field]} (#{formatted_start_date} - #{formatted_end_date})"
    end.join("\n")
  end
```

Thoughts?

app/models/lettings_log.rb
```diff
@@ -84,7 +84,7 @@ def self.editable_fields
 
   def la
     if location
-      location.location_code
+      location.linked_local_authorities.active(form.start_date).first.code
```
jamesrose:
```suggestion
      location.linked_local_authorities.active(form.start_date).first&.code
```

app/models/location.rb
```diff
@@ -130,6 +130,14 @@ def validate_confirmed
     self.confirmed = [postcode, location_admin_district, location_code, units, type_of_unit, mobility_type].all?(&:present?)
   end
 
+  def linked_local_authorities
+    la = LocalAuthority.find_by(code: location_code)
+    return [] unless la
+
+    location_local_authorities = [la] + la.linked_local_authorities
+    LocalAuthority.where(id: location_local_authorities.map(&:id))
+  end
```
jamesrose: Can this not be an association?

app/helpers/locations_helper.rb
kosiakkatrina: Hm, yes to the first comment, not too sure about removing the return. This will impact a very small number of LAs, so maybe we shouldn't display the dates by default

jamesrose: Your call!

jamesrose: (Note the AR sorter vs Enumerator.)

app/models/location.rb
kosiakkatrina: hm, I'm not too sure, I tried to make it one, but we don't save the LA code as key in the links table and from location we only have a code. Maybe possible but not sure if clearer

app/helpers/locations_helper.rb
kosiakkatrina: linked_local_authorities is an array and not a relation, so that doesn't work 😬

kosiakkatrina: That's okay tho methinks, cause there'd be at most 2 records there anyway

app/models/location.rb
jamesrose:
```suggestion
  def linked_local_authorities
    local_authority = LocalAuthority.where(code: location_code).pluck(:id)
    return local_authority if local_authority.empty?

    LocalAuthority.where(id: local_authority + la.linked_local_authorities_ids)
  end
```

jamesrose:
```suggestion
  def linked_local_authorities
    la = LocalAuthority.find_by(code: location_code)
    return [] unless la

    LocalAuthority.where(id: [la.id] + la.linked_local_authorities_ids)
  end
```

If you don't map the `linked_local_authorities` you can just pluck out the IDs. You can even do:

```ruby
LocalAuthority.where(id: la.linked_local_authorities_ids + la.id)
```

... if you don't mind about the ordering.

app/models/location.rb
```diff
@@ -130,6 +130,13 @@ def validate_confirmed
     self.confirmed = [postcode, location_admin_district, location_code, units, type_of_unit, mobility_type].all?(&:present?)
   end
 
+  def linked_local_authorities
+    la = LocalAuthority.find_by(code: location_code)
```
arfacamble, line 134: This looks confusing to me, but do I remember a comment somewhere else that location_code means something to do with `Location`s but also something else unrelated to do with LAs. Is it just like a unique id for them or something?

Separately, would this be the same as `LocalAuthority.find_by_code(location_code)`?

jamesrose (approved):

app/models/location.rb
kosiakkatrina: Hm, so I think `lettings_log.location_code` == `location.id`, but `location.code` = `local_authority.location_code` = admin district E code, which is what we're searching by here.

I believe it's the same, yes

PR #1403, opened 2023-03-09
CLDC-2034 Skip fixed term tenancy length question when teancy type is other

natdeanlewissoftwire:
Pretty self explanatory, haven't covered this by tests as it seems none of the details of json/ruby files lettings log behaviour is. Might be worth addressing that as a whole separately at some point 🤔 

https://digital.dclg.gov.uk/jira/browse/CLDC-2034


app/models/form/lettings/pages/tenancy_length.rb
```diff
@@ -2,7 +2,7 @@ class Form::Lettings::Pages::TenancyLength < ::Form::Page
   def initialize(id, hsh, subsection)
     super
     @id = "tenancy_length"
-    @depends_on = [{ "tenancy" => 4 }, { "tenancy" => 6 }, { "tenancy" => 3 }]
+    @depends_on = [{ "tenancy" => 4 }, { "tenancy" => 6 }]
```
ghost: seems like this question is missing a spec?

natdeanlewissoftwire: Yep, mentioned this in the PR description - although I think I might have been wrong, and it's the lettings ruby files that are missing a load of tests. I'll add a spec to this ticket, thanks :D

jamesrose (approved):

PR #1404, opened 2023-03-09
CLDC-2026 Add minimum staircasing validation for shared ownership sales

SamSeed-Softwire:
## Context
- https://digital.dclg.gov.uk/jira/browse/CLDC-2026

## The changes
- Added 10% threshold validation for shared ownership types 2, 16, 18, 24.
- Added 1% threshold validation for all other shared ownership types.
- Added tests for all shared ownership types - might have been overkill, but maybe fine.

### DIFF
--- a/app/models/validations/sales/financial_validations.rb
+++ b/app/models/validations/sales/financial_validations.rb
@@ -52,6 +52,16 @@ module Validations::Sales::FinancialValidations
     end
   end
 
+  def validate_percentage_bought_at_least_threshold(record)
+    return unless record.stairbought && record.type
+
+    if ([2, 18, 16, 24].include? record.type) && record.stairbought < 10
+      record.errors.add :stairbought, I18n.t("validations.financial.staircasing.percentage_bought_must_be_at_least_threshold", percentage: 10)
+    elsif record.type == 30 && record.stairbought < 1
+      record.errors.add :stairbought, I18n.t("validations.financial.staircasing.percentage_bought_must_be_at_least_threshold", percentage: 1)
+    end
+  end
+
   def validate_child_income(record)
     return unless record.income2 && record.ecstat2
 
--- a/app/models/validations/sales/financial_validations.rb
+++ b/app/models/validations/sales/financial_validations.rb
@@ -55,11 +55,11 @@ module Validations::Sales::FinancialValidations
   def validate_percentage_bought_at_least_threshold(record)
     return unless record.stairbought && record.type
 
-    if [2, 16, 18, 24].include? record.type
-      threshold = 10
-    else
-      threshold = 1
-    end
+    threshold = if [2, 16, 18, 24].include? record.type
+                  10
+                else
+                  1
+                end
 
     if threshold && record.stairbought < threshold
       record.errors.add :stairbought, I18n.t("validations.financial.staircasing.percentage_bought_must_be_at_least_threshold", threshold:)
--- a/app/models/validations/sales/financial_validations.rb
+++ b/app/models/validations/sales/financial_validations.rb
@@ -63,6 +63,7 @@ module Validations::Sales::FinancialValidations
 
     if threshold && record.stairbought < threshold
       record.errors.add :stairbought, I18n.t("validations.financial.staircasing.percentage_bought_must_be_at_least_threshold", threshold:)
+      record.errors.add :type, I18n.t("validations.setup.type.percentage_bought_must_be_at_least_threshold", threshold:)
     end
   end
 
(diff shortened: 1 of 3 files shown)


app/models/validations/sales/financial_validations.rb
```diff
@@ -52,6 +52,20 @@ def validate_percentage_bought_not_greater_than_percentage_owned(record)
...
+    return unless record.stairbought && record.type
+
+    if [2, 16, 18, 24].include? record.type
+      threshold = 10
+    elsif record.type == 30
+      threshold = 1
+    end
+
+    if threshold && record.stairbought < threshold
+      record.errors.add :stairbought, I18n.t("validations.financial.staircasing.percentage_bought_must_be_at_least_threshold", threshold:)
```
SamSeed-Softwire, line 65: I think we should also add an error message to `:type` (so that if you set a valid value for stairbought and then go back and change the shared ownership type such that the stairbought value is now invalid, then an error message is shown on the shared ownership type page.

Alternatively do we want to just clear the stairbought value if the shared ownership type changes? Presumably not.

Any other options?

jamesrose: @SamSeed-Softwire Yes, I would add the message to type as well.

app/models/validations/sales/financial_validations.rb
```diff
@@ -52,6 +52,21 @@ def validate_percentage_bought_not_greater_than_percentage_owned(record)
...
   end
 
+  def validate_percentage_bought_at_least_threshold(record)
+    return unless record.stairbought && record.type
+
+    threshold = if [2, 16, 18, 24].include? record.type
+                  10
+                else
+                  1
+                end
```
arfacamble, line 62: do types 2, 16, 18, 24 fit inside an easily expressed category?
if so I would support making a method on SalesLog so you can do:
`threshold = record.that_category_type? ? 10 : 1`

spec/models/validations/sales/financial_validations_spec.rb
```diff
@@ -160,6 +160,118 @@
...
+      record.errors.clear
+    end
+
+    it "doesn't add an error to stairbought and type if the percentage bought is less than the threshold (which is 1% by default, but higher for some shared ownership types)" do
+      record.stairbought = 10
+
+      record.type = 2
+      financial_validator.validate_percentage_bought_at_least_threshold(record)
+      expect(record.errors["stairbought"]).to be_empty
+      expect(record.errors["type"]).to be_empty
```
arfacamble: for all of these tests (sidenote, I'm not sure you need to test every one, but it doesn't hurt I suppose) you can replace these two lines with just `expect(record.errors).to be_empty`

spec/models/validations/sales/financial_validations_spec.rb
```diff
@@ -160,6 +160,118 @@
...
+      record.errors.clear
+
+      record.type = 24
+      financial_validator.validate_percentage_bought_at_least_threshold(record)
+      expect(record.errors["stairbought"]).to be_empty
+      expect(record.errors["type"]).to be_empty
+      record.errors.clear
```
arfacamble: this repetition does make think you could be iterating also
```
[2, 16, ..].each do |type|
  record.type = type
  financial_validator.validate
  expect empty
  clear errors
end
```

app/models/validations/sales/financial_validations.rb
SamSeed-Softwire: I can't see an obvious category but will check with Jemma, thanks for the suggestion.

SamSeed-Softwire: @arfacamble Jemma has confirmed there's no obvious category:

"Yeah sorry I don't think there is one really. And I can't see anything in common with some of the other conditional validations/displaying of questions depending on shared ownership type that we've implemented elsewhere"

One option could therefore be for us to create our own category e.g. `higher_staircasing_threshold_types` or `10_percent_staircasing_threshold_types` but I think those categories are a bit inflexible / wouldn't work well if new thresholds were introduced. Lmk if you can think of something better, otherwise I'll leave it as just an array of types.

arfacamble: no I hadn't looked at what the type codes represented. I agree there's not much point putting them in a category only named after what is happening here.

spec/models/validations/sales/financial_validations_spec.rb
```diff
@@ -160,6 +160,48 @@
...
+
+      record.stairbought = 1
+      [28, 30, 31, 32].each do |type|
+        record.type = type
+        financial_validator.validate_percentage_bought_at_least_threshold(record)
+        expect(record.errors).to be_empty
+        record.errors.clear
+      end
+    end
+  end
```
arfacamble, line 203: looks great. I think this is a great combo of concise + full coverage
⛵ 
that is my boat, and it's floating

arfacamble (approved):

PR #1408, opened 2023-03-13
CLDC-1886 Add sales paper form download

kosiakkatrina:
image


app/controllers/start_controller.rb
```diff
@@ -4,4 +4,12 @@ def index
       redirect_to(lettings_logs_path)
     end
   end
+
+  def download_pdf
```
jamesrose: Perhaps this could be called something more descriptive? We might want to add a lettings one, for example.

kosiakkatrina: Yes, I think I had an idea doing 1 path for all downloads, but that's not very smart, so I shall change this

jamesrose (approved):

PR #1410, opened 2023-03-13
CLDC-2074 Change care home charges validation

kosiakkatrina:
Add care home charges soft validation and make care home charges field optional

image


db/migrate/20230313124749_add_carehome_charges_value_check.rb
```diff
@@ -0,0 +1,5 @@
+class AddCarehomeChargesValueCheck < ActiveRecord::Migration[7.0]
+  def change
+    add_column :lettings_logs, :carehome_charges_value_check, :integer
```
jamesrose, line 3: Do we need to backdate values for this?

kosiakkatrina: I don't think so. I don't think it is currently possible to have the answers to `is_carehome` as "Yes" and `chcharge` as nil, because it would throw a validation (?)

jamesrose (approved):

PR #1412, opened 2023-03-14
Run tests in parallel

kosiakkatrina:
Run unit tests in parallel using parallel_test gem
Brings down the running time in CI from ~10min to ~6min
Remove --fail-fast from the unit tests, which means the tests will have to run fully before gihtub actions marks it as failed, but looking at some examples, our tests would still often take longer than 6 minutes to fail even with --fail-fast


natdeanlewissoftwire (approved):
lgtm 👍

PR #1413, opened 2023-03-14
Fix: get real validation messages for rent range soft validations

SamSeed-Softwire:
The soft validation banners shown when the rent ranges for a lettings log are unusually high or low were not displaying correctly. The validation logic was looking up things in `en.yml` which didn't exist. This PR fixes that.


kosiakkatrina (approved):
Nice spot

PR #1420, opened 2023-03-16
CLDC-2032 test log creation in crossover period

arfacamble:
https://digital.dclg.gov.uk/jira/browse/CLDC-2032

one small fix necessary, explanation in line.

gifs below:
|23_24_lettings_log
|-

|22_23_lettings_log
|-

|22_23_sales_log
|-

|23_24_sales_log
|-

### DIFF
--- a/app/models/form_handler.rb
+++ b/app/models/form_handler.rb
@@ -47,7 +47,7 @@ class FormHandler
         form = Form.new(form_path)
 
         form_to_set = form_name_from_start_year(form.start_date.year, "lettings")
-        forms[form_to_set] = form if forms[form_to_set].blank?
+        forms[form_to_set] = form if form_to_set && forms[form_to_set].blank?
       end
     end
 
--- a/spec/models/form_handler_spec.rb
+++ b/spec/models/form_handler_spec.rb
@@ -9,8 +9,6 @@ RSpec.describe FormHandler do
       Singleton.__init__(described_class)
       example.run
     end
-    Timecop.return
-    Singleton.__init__(described_class)
   end
 
   context "when accessing a form in a different year" do
@@ -44,7 +42,6 @@ RSpec.describe FormHandler do
           Singleton.__init__(described_class)
           example.run
         end
-        Singleton.__init__(described_class)
       end
 
       it "does not load outdated forms" do
--- a/spec/models/form_handler_spec.rb
+++ b/spec/models/form_handler_spec.rb
@@ -35,6 +35,23 @@ RSpec.describe FormHandler do
       expect(all_forms.count).to be >= 1
       expect(all_forms["current_sales"]).to be_a(Form)
     end
+
+    context "when in 23/24 period or later" do
+      let(:now) { Time.utc(2023, 6, 7) }
+
+      around do |example|
+        Timecop.freeze(now) do
+          Singleton.__init__(described_class)
+          example.run
+        end
+        Singleton.__init__(described_class)
+      end
+
+      it "does not load outdated forms" do
+        all_forms = form_handler.forms
+        expect(all_forms.keys).not_to include nil
+      end
+    end
   end
 
   describe "Get specific form" do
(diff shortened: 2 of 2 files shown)


app/models/form_handler.rb
arfacamble, line 1: when creating a new log, that log should default to the current year's form until it can tell its form name from its saledate/startdate
in the crossover period, the 21_22 lettings form is no longer current or previous and would be added to the @forms hash under the key nil, this is prevented by the change below.

kosiakkatrina: What issue does not having this cause? I thought we can no longer update 21/22 logs?
Maybe adding a test that fails without this change would be useful?

app/models/form_handler.rb
```diff
@@ -47,7 +47,7 @@ def lettings_forms
         form = Form.new(form_path)
 
         form_to_set = form_name_from_start_year(form.start_date.year, "lettings")
-        forms[form_to_set] = form if forms[form_to_set].blank?
+        forms[form_to_set] = form if form_to_set && forms[form_to_set].blank?
       end
```
arfacamble, line 51: @kosiakkatrina 
`Dir.glob("#{directory}/*.json") => ["config/2021_2022.json", "config/2022_2023.json"]`
for 21/22 we will call `form_name_from_start_year(2021, "lettings")`
in form_name_from_start_year below line 71 you will see that will return
`form_mappings[2023 - 2021]` => `form_mappings[2]` => `nil`
so back here on line 50 we will call `forms[nil] = bla bla`

This is a problem because when we create a log, we go and get the relevant form. we use the saledate/startdate to get the right form name and call get_form (above line 23).
if there is no startdate/saledate, then currently we call `get_form(nil)` which returns `nil` and we then default in the log `:form` method to the current form for that type:
`FormHandler.instance.get_form(form_name) || FormHandler.instance.current_lettings_form`
In the crossover period, if we allow `FormHandler.instance.forms` to have a nil key in it, then until saledate/startdate is set we default to the 21/22 lettings form. This crashes the site for creating new sales logs entirely for obvious reasons and the lettings log section doesn't allow the setup subsection link to function.

I think that we don't want to be adding things to our forms hash under the key `nil`
We may also want to make our `:form` methods in `SalesLog` and `LettingsLog` more robust to deal with this case? I didn't immediately feel that was necessary as long as the forms hash is kept tidy.

app/models/form_handler.rb
arfacamble: test now also written

spec/models/form_handler_spec.rb
```diff
@@ -34,6 +34,23 @@
...
+
+    context "when in 23/24 period or later" do
+      let(:now) { Time.utc(2023, 6, 7) }
+
+      around do |example|
+        Timecop.freeze(now) do
+          Singleton.__init__(described_class)
+          example.run
+        end
+        Singleton.__init__(described_class)
```
kosiakkatrina: I don't think we need this line because it's added in spec_helper after hook

app/models/form_handler.rb
kosiakkatrina: Okay, yes, this makes sense!
I've got a different question tho, in the crossover period, the form should default to 23/24, because that would be the `current_form` for that time. Is that not what's happening?

arfacamble: that will be what will happen with the change that I have made because 
`FormHandler.instance.get_form(form_name) || FormHandler.instance.current_lettings_form`
becomes `FormHandler.instance.get_form(nil) || FormHandler.instance.current_lettings_form`
becomes `nil || FormHandler.instance.current_lettings_form`
but without the change (ie `FormHandler.instance.forms => {nil => <#Form 21-22>, "current_sales" => <#Form bla>, etc}`:
`FormHandler.instance.get_form(form_name) || FormHandler.instance.current_lettings_form`
becomes `FormHandler.instance.get_form(nil) || FormHandler.instance.current_lettings_form`
becomes `<#Form 21-22> || FormHandler.instance.current_lettings_form`
and this makes crashy crashy

spec/models/form_handler_spec.rb
arfacamble: excellent ta. removed it somewhere else too

kosiakkatrina (approved):

PR #1425, opened 2023-03-16
CLDC-2143 Address import validations

kosiakkatrina:
Address errors happening during the import:
- [x] Buyer's last accommodation and discounted ownership postcodes must match - Clear previous postcode
- [x] Child's income must be £0 - remove the validation for 22/23
- [x] Contract exchange date must be less than 1 year before completion date - Clear exchange date
- [x] Income must not exceed £80,000 for properties outside London local authorities - Clear income

Also do not run the discounted ownership validation when mortgageused is don't know (3) only impacts imports, because don't know is a hidden value in single log submission flow.

Errors to be addressed in a separate PRs:

- Mortgage, deposit, and grant total must equal x
- Person cannot be aged 16-19 if they are a student but don't have relationship child

### DIFF
--- a/app/services/imports/sales_logs_import_service.rb
+++ b/app/services/imports/sales_logs_import_service.rb
@@ -191,11 +191,12 @@ module Imports
           attributes.delete(error.attribute.to_s)
         end
         @logs_overridden << sales_log.old_id
-        if sales_log.errors.of_kind?(:postcode_full, :postcodes_not_matching)
-          @logger.warn("Log #{sales_log.old_id}: Removing postcode known and previous postcode known as the postcodes are invalid")
-          attributes.delete("pcodenk")
-          attributes.delete("ppcodenk")
-        end
+        save_sales_log(attributes, previous_status)
+      elsif sales_log.errors.of_kind?(:postcode_full, :postcodes_not_matching)
+        @logger.warn("Log #{sales_log.old_id}: Removing previous postcode known and previous postcode as the postcode is invalid")
+        @logs_overridden << sales_log.old_id
+        attributes.delete("ppcodenk")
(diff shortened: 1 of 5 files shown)


app/services/imports/sales_logs_import_service.rb
```diff
@@ -191,11 +191,22 @@ def rescue_validation_or_raise(sales_log, attributes, previous_status, exception
...
+        save_sales_log(attributes, previous_status)
+      elsif sales_log.errors.of_kind?(:exdate, :over_a_year_from_saledate)
+        @logger.warn("Log #{sales_log.old_id}: Removing exchange date as the exchange date is invalid")
+        @logs_overridden << sales_log.old_id
+        attributes.delete("exdate")
+        save_sales_log(attributes, previous_status)
+      elsif sales_log.errors.of_kind?(:income1, :over_hard_max_for_outside_london)
+        @logger.warn("Log #{sales_log.old_id}: Removing income1 as the income1 is invalid")
+        @logs_overridden << sales_log.old_id
+        attributes.delete("income1")
```
natdeanlewissoftwire, line 209: is there a similar behaviour we should add for income2?

natdeanlewissoftwire: and should we clear both if the combined incomes are over the combined threshold?

natdeanlewissoftwire (approved):
Nice approach, just one question but all good!

app/services/imports/sales_logs_import_service.rb
kosiakkatrina: Possibly, I'm only addressing the cases that are currently causing an issue when importing. We might never hit the other cases

PR #1428, opened 2023-03-16
Update service charge validations

jamesrose:
- [x] Update service charge ranges
- [x] Make sure the location is confirmed correctly
- [x] Do not set scheme end date, create a deactivation period for location instead

### DIFF
--- a/app/services/imports/scheme_location_import_service.rb
+++ b/app/services/imports/scheme_location_import_service.rb
@@ -97,7 +97,7 @@ module Imports
     end
 
     def add_location(scheme, attributes)
-      location = Location.new(
+      location = Location.create!(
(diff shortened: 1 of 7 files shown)


app/services/imports/scheme_location_import_service.rb
```diff
@@ -110,6 +108,11 @@ def add_location(scheme, attributes)
         startdate: attributes["start_date"],
         scheme:,
       )
+      confirm_scheme_or_location(location)
```
jamesrose: Not required anymore.

kosiakkatrina (approved):

jamesrose:
Thanks @kosiakkatrina !!

PR #1429, opened 2023-03-17
Value validation

kosiakkatrina:
Builds on top of https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/1425 so might be easier to merge that first
Change discounted ownership validation to a soft validation that will run for 22/23 and 23/24. Change the hard validation to run from 24/25 onwards.
image


app/models/form/sales/pages/discounted_sale_value_check.rb
natdeanlewissoftwire, line 1: I think we're naming all the other shared person pages starting with person_, if that's right I think we should keep that going here

app/models/form/sales/questions/discounted_sale_value_check.rb
```diff
@@ -0,0 +1,23 @@
+class Form::Sales::Questions::DiscountedSaleValueCheck < ::Form::Question
+  def initialize(id, hsh, page)
+    super
+    @id = "discounted_sale_value_check"
+    @check_answer_label = "Discounted sale value confirmation"
+    @header = "Are you sure this is correct?"
+    @type = "interruption_screen"
+    @answer_options = {
+      "0" => { "value" => "Yes" },
+      "1" => { "value" => "No" },
```
natdeanlewissoftwire, line 10: oo look, more 0 == yes, 1 == no 🥳

app/models/sales_log.rb
```diff
@@ -316,4 +316,19 @@ def field_formatted_as_currency(field_name)
   def should_process_uprn_change?
     uprn_changed? && saledate && saledate.year >= 2023
   end
+
+  def value_with_discount
+    return if value.blank?
+
+    discount_amount = discount ? value * discount / 100 : 0
+    (value - discount_amount)
```
natdeanlewissoftwire: are these parentheses needed?

app/models/validations/sales/sale_information_validations.rb
```diff
@@ -45,17 +47,14 @@ def validate_previous_property_unit_type(record)
   end
 
   def validate_discounted_ownership_value(record)
+    return unless record.saledate && collection_start_year(record.saledate) > 2023
```
natdeanlewissoftwire: `>= 2024 `would be slightly clearer imo but not a big deal

config/locales/en.yml
```diff
@@ -526,6 +526,9 @@ en:
     staircase_bought_seems_high: "You said %{percentage}% was bought in this staircasing transaction, which seems high. Are you sure?"
     monthly_charges_over_soft_max:
       title_text: "The amount of monthly charges is high for this type of property and sale type"
+    discounted_sale_value:
+      title_text: "Mortgage, deposit, and grant total must equal £%{value_with_discount}"
+      informative_text: "Your given mortgage, deposit and grand total is £%{mortgage_deposit_and_grand_total}"
```
natdeanlewissoftwire: I'm not sure the grammar of this sentence is that clear - given as in it's been input by the user? More of a copy question but there feels like a better way of saying this

natdeanlewissoftwire (approved):
couple of thoughts, nothing not worth approving over though :)

config/locales/en.yml
kosiakkatrina: You're right, I'll get it PO reviewed and will see if any comments come out of it!

app/models/sales_log.rb
kosiakkatrina: nope

PR #1430, opened 2023-03-17
Filter on the subset of logs in tests

kosiakkatrina:
Filter on the subset of logs in tests instead of all the logs in the db


asmega:
Might fix it. I would of thought everything would be inside a transaction (transactional fixtures) therefore this would not make a difference?

asmega (approved):

kosiakkatrina:
> Might fix it. I would of thought everything would be inside a transaction (transactional fixtures) therefore this would not make a difference?

@asmega yeah, I know what you mean. I mean, can't hurt 🤷

PR #1435, opened 2023-03-17
Seed la data before links

kosiakkatrina:


ghost (approved):

spec/spec_helper.rb
```diff
@@ -113,6 +113,9 @@
 
   config.before do
     RequestHelper.stub_http_requests
+  end
+
+  config.before(:all) do
```
asmega: should this not be `config.before(:suite)`?

PR #1436, opened 2023-03-17
CLDC-2134 Bulk upload tabs

asmega:
# Context

- https://digital.dclg.gov.uk/jira/browse/CLDC-2134

# Changes

- Copy change to bulk upload email, swaps question and column ordering and precedence
- Bulk upload summary report now uses tabs and contains the full report in the other tab
- Summary report now uses multiple tables rather than one
- Can no longer click back to summary report from full report. As when viewing the full report there should not be enough errors to populate the summary report
- Various copy changes due to the changes in the journey

# Screenshots

NOTE: for testing I set to threshold to 5 errors as I didn't have enough test data for 16 which is what the value should be

Screenshot 2023-03-17 at 14 04 32

Screenshot 2023-03-17 at 14 04 46

### DIFF
--- a/app/components/bulk_upload_error_summary_table_component.html.erb
+++ b/app/components/bulk_upload_error_summary_table_component.html.erb
@@ -1,24 +1,16 @@
-<%= govuk_table do |table| %>
-  <% table.caption(size: "m", text: bulk_upload.filename) %>
+<% sorted_errors.each do |error| %>
+  <%= govuk_table do |table| %>
+    <% table.head do |head| %>
+      <% head.row do |row| %>
+        <% row.cell(text: question_for_field(error[0][1].to_sym), header: true) %>
+        <% row.cell(text: "Column #{error[0][0]}", header: true, numeric: true) %>
+      <% end %>
 
-  <% table.head do |head| %>
-    <% head.row do |row| %>
-      <% row.cell(text: "Column", header: true) %>
-      <% row.cell(text: "Number of rows", header: true) %>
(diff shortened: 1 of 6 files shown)


app/components/bulk_upload_error_summary_table_component.html.erb
```diff
@@ -1,24 +1,16 @@
...
-      <% row.cell(text: "Question", header: true) %>
-      <% row.cell(text: "Error", header: true) %>
-      <% row.cell(text: "Specification", header: true) %>
-    <% end %>
-  <% end %>
+<% sorted_errors.each do |error| %>
+  <%= govuk_table do |table| %>
+    <% table.head do |head| %>
+      <% head.row do |row| %>
+        <% row.cell(text: question_for_field(error[0][1].to_sym), header: true) %>
```
jamesrose, line 5: What's `error[0][1]` and below `error[0][2]` in this instance?

jamesrose (approved):

app/components/bulk_upload_error_summary_table_component.html.erb
asmega: sorry it's not great. it because its doing an aggregate in sql so the output is something like
```ruby
[
  [["H", "field_7", "This is a duplicate log"], 9],
  [["M", "field_12", "This is a duplicate log"], 9],
  ...
]
```
the aggregate is the count or in this case `9` in the above example the other part is the group by part

PR #1441, opened 2023-03-20
Student child validation

kosiakkatrina:
Change student, aged between 16 and 19 and not a child validation to be a soft validation instead:
image


app/models/form/sales/pages/student_not_child_value_check.rb
natdeanlewissoftwire, line 1: same as my other comment about starting with `person_` but it feels even more tentative here

app/models/form/sales/questions/student_not_child_value_check.rb
```diff
@@ -0,0 +1,24 @@
...
+        {
+          "student_not_child_value_check" => 0,
+        },
+        {
+          "student_not_child_value_check" => 1,
+        },
+      ],
+    }
+    @check_answers_card_number = person_index
+    @header = "Are you sure this person is not child?"
```
natdeanlewissoftwire: grammar seems odd - maybe "Are you sure this person is not a child?"

app/services/imports/sales_logs_import_service.rb
```diff
@@ -242,7 +243,8 @@ def fields_not_present_in_softwire_data
          staircase_bought_value_check
          monthly_charges_value_check
          hodate_check
-         saledate_check]
+         saledate_check
+         student_not_child_value_check]
```
natdeanlewissoftwire, line 247: nit: can we put the ] on the next line for tidiness

natdeanlewissoftwire (commented):
small grammar point, otherwise looks good!

natdeanlewissoftwire (approved):

app/services/imports/sales_logs_import_service.rb
kosiakkatrina: Oh, rubocop don't like that :<

PR #1442, opened 2023-03-20
CLDC-1945 numeric questions decimal letter entry

arfacamble:
https://digital.dclg.gov.uk/jira/browse/CLDC-1945

This ticket is to deal with the situation currently where questions relating to integer fields in the db would accept decimals and round/truncate them silently.
The user will now see a validation indicating the degree of accuracy required for the question.

Relatedly users are able to enter the letter 'e' in number input fields, since HTML allows representing numbers in exponent form (eg 1e2 = 1 * 10^2) - users who enter an e in an invalid exponent form (eg 'e' or 'e1') we will allow HTML to treat that as invalid, empty the field and the standard question is required validation will show. users who enter a number in valid exponent form (eg '1e2') will see a validation asking them to enter the number in normal number digits form

|image
|-

---

N.B.
the values for @step for each question have been checked and cleared with Jemma. Worth noting that several fields representing currency are decimal with scale 2 in the db but we are now only allowing integer. There may be "invalid" values in the db and we may want to do tech debt to change the column type but I imagine both of these things are really low priority


spec/models/validations/shared_validations_spec.rb
arfacamble, line 1: apologies there is a lot of renaming diff here.
I will comment in line where actual changes are in this file and please ignore everything else

spec/models/validations/shared_validations_spec.rb
```diff
@@ -15,119 +15,177 @@
...
-        sales_record.income1 = -5
-        shared_validator.validate_numeric_min_max(sales_record)
-        expect(sales_record.errors["income1"])
+        sales_log.income1 = -5
+        shared_validator.validate_numeric_min_max(sales_log)
+        expect(sales_log.errors["income1"])
           .to include(match I18n.t("validations.numeric.within_range", field: "Buyer 1’s gross annual income", min: "£0", max: "£999,999"))
       end
     end
   end
```
arfacamble, line 116: HERE ARE THE CHANGES!!!!!

app/models/validations/shared_validations.rb
```diff
@@ -34,6 +34,35 @@ def validate_numeric_min_max(record)
     end
   end
 
+  def validate_numeric_step(record)
+    record.form.numeric_questions.each do |question|
+      next unless question.step
+      next unless record[question.id] && question.page.routed_to?(record, nil)
```
kosiakkatrina, line 40: Wouldn't the value get cleared if the page is not routed to?

app/models/validations/shared_validations.rb
```diff
@@ -34,6 +34,35 @@ def validate_numeric_min_max(record)
...
+
+      begin
+        input = record.public_send("#{question.id}_before_type_cast")
+        answer = Float(input)
+      rescue ArgumentError
+        add_step_error(record, question)
+      end
```
kosiakkatrina: is there a test for this?

app/models/validations/shared_validations.rb
```diff
@@ -34,6 +34,35 @@ def validate_numeric_min_max(record)
...
+  def validate_numeric_normal_format(record)
+    record.form.numeric_questions.each do |question|
+      next unless record[question.id] && question.page.routed_to?(record, nil)
+
+      user_input = record.public_send("#{question.id}_before_type_cast")
+      if user_input.is_a?(String) && user_input.include?("e")
+        record.errors.add question.id.to_sym, I18n.t("validations.numeric.normal_format")
+      end
+    end
+  end
```
jamesrose: I don't think we should do this. Let's discuss!

config/locales/en.yml
```diff
@@ -139,6 +139,11 @@ en:
     numeric:
       within_range: "%{field} must be between %{min} and %{max}"
       above_min: "%{field} must be at least %{min}"
+      whole_number: "%{field} must be a whole number"
+      nearest_ten: "%{field} must be given to the nearest ten"
+      nearest_penny: "%{field} must be given to the nearest penny"
```
jamesrose: Are you sure this is always a monetary amount?

arfacamble: I did check that currently all questions with step 0.01 are representing currency, but I agree this made me pause

jamesrose: I think we need to keep in mind that numeric fields aren't limited to currency, and might not be in the future. Suggest changing `nearest_penny` to something that works for all numeric types.

app/models/validations/shared_validations.rb
arfacamble: hmm. I might try and have a little experiment with this. What you say makes sense, I wonder if there's a chance because we're in a before-save lifecycle call back if there might be funny edge cases.
Really here, I copied the format of the above method for validate_numeric_min_max so I could get started and didn't come back round to question this.
Now I look in git blame it looks like you last touched the corresponding line on l21 above, maybe you can enlighten me as to why I need to do this! I will see what happens if I get rid of it

arfacamble: hmmm. I am happy to remove this if you think it's fine, I wonder if there's an edge case where it makes sense though.
A page can be not routed to if the subsection it is in is not enabled, and perhaps that answer could be derived? I should hope that we don't derive invalid values, but maybe it's not crazy to have an extra safeguard that a validation error gets added to a question that the user can't even get to? although in that case we will be calling a value from the db that has already passed this validation, so I really can't imagine a case where this is necessary.

OK I see that you did add it in the other method and discussed it with sam in review
https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/1249/files#diff-86ebd101437a30c8df76e2e300c9c61a9cdecefd09e7594fe02cad46e7234aa2
maybe when you're free you can call/message if you remember that thing and whether it applies here (I don't tihnk I have the context for that conversation)

jamesrose (approved):

arfacamble:
comment

PR #1448, opened 2023-03-20
CLDC-2017-monthly-charges-validation

arfacamble:
https://digital.dclg.gov.uk/jira/browse/CLDC-2017

after discussing with Jemma, it is not the case that this can be covered by the standard error messages for any field with values outside min max range given.
The error message specifically directs them to change their answer to another question in the case they feel the value should be 0.

The existing range validation error message still applies for non zero but non valid inputs

### DIFF
--- a/app/models/validations/sales/financial_validations.rb
+++ b/app/models/validations/sales/financial_validations.rb
@@ -41,7 +41,7 @@ module Validations::Sales::FinancialValidations
   end
 
   def validate_monthly_leasehold_charges(record)
-    record.errors.add :mscharge, I18n.t("validations.financial.monthly_leasehold_charges.not_zero") if record.mscharge == 0
+    record.errors.add :mscharge, I18n.t("validations.financial.monthly_leasehold_charges.not_zero") if record.mscharge&.zero?
   end
 
   def validate_percentage_bought_not_greater_than_percentage_owned(record)
--- a/app/models/validations/sales/financial_validations.rb
+++ b/app/models/validations/sales/financial_validations.rb
@@ -40,12 +40,8 @@ module Validations::Sales::FinancialValidations
     record.errors.add :mortgage, :cannot_be_0, message: I18n.t("validations.financial.mortgage") if record.mortgage_used? && record.mortgage&.zero?
   end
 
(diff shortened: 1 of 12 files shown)


spec/models/validations/sales/financial_validations_spec.rb
arfacamble, line 1: removed the validate cash discount as it is doing the exact same thing as the shared range validation

app/models/validations/sales/financial_validations.rb
```diff
@@ -36,12 +36,8 @@ def validate_combined_income(record)
...
   end
 
-  def validate_cash_discount(record)
-    return unless record.cashdis
-
-    unless record.cashdis.between?(0, 999_999)
-      record.errors.add :cashdis, I18n.t("validations.financial.cash_discount_invalid")
-    end
+  def validate_monthly_leasehold_charges(record)
+    record.errors.add :mscharge, I18n.t("validations.financial.monthly_leasehold_charges.not_zero") if record.mscharge&.zero?
```
kosiakkatrina, line 44: Is this so that the validation message is different from min/max?

app/models/validations/sales/financial_validations.rb
```diff
@@ -36,12 +36,8 @@ def validate_combined_income(record)
     end
   end
 
-  def validate_cash_discount(record)
-    return unless record.cashdis
-
-    unless record.cashdis.between?(0, 999_999)
-      record.errors.add :cashdis, I18n.t("validations.financial.cash_discount_invalid")
-    end
```
kosiakkatrina, line 48: Is this covered in min/max validations?
edit: dw, just saw your comment

kosiakkatrina (approved):

natdeanlewissoftwire (approved):

app/models/validations/sales/financial_validations.rb
arfacamble: exactly

app/models/validations/sales/financial_validations.rb
arfacamble: very good

PR #1450, opened 2023-03-21
CLDC-2143 Set more default fields and don't clear setup values

kosiakkatrina:
- [x] Set default hhregresstill to don't know
- [x] Do not clear type or ownerhshipsch if there's an error attached to them, because they're in a setup section and we would end up losing data
- [x] Add `Don't know` for `mortgagelender`
- [x] Make `mortgagelender` mandatory again and infer as `Don't know` by default


natdeanlewissoftwire (approved):
lgtm!

PR #1452, opened 2023-03-21
CLDC-2148-household-count-validation

arfacamble:
https://digital.dclg.gov.uk/jira/browse/CLDC-2148

not much to explain here beyond the ticket description. Max has gone from 4 or 5 up to 15.
Took the liberty to polish a tiny bit of copy and have cleared that with Jemma but when you see it you will see that it's _obviously_ fine

### DIFF
--- a/app/models/form/sales/pages/number_of_others_in_property.rb
+++ b/app/models/form/sales/pages/number_of_others_in_property.rb
@@ -4,11 +4,11 @@ class Form::Sales::Pages::NumberOfOthersInProperty < ::Form::Page
     @depends_on = [
       {
         "privacynotice" => 1,
-        "jointpur" => joint_purchase ? 1 : 2,
+        "joint_purchase?" => joint_purchase,
       },
       {
-        "noint" => 1,
-        "jointpur" => joint_purchase ? 1 : 2,
+        "buyer_not_interviewed?" => true,
+        "joint_purchase?" => joint_purchase,
       },
     ]
     @joint_purchase = joint_purchase
--- a/spec/models/form/sales/questions/number_of_others_in_property_spec.rb
+++ b/spec/models/form/sales/questions/number_of_others_in_property_spec.rb
@@ -9,7 +9,7 @@ RSpec.describe Form::Sales::Questions::NumberOfOthersInProperty, type: :model do
   let(:joint_purchase) { true }
 
   it "has correct page" do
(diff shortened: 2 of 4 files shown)


kosiakkatrina (approved):

app/models/form/sales/pages/number_of_others_in_property.rb
```diff
@@ -4,11 +4,11 @@ def initialize(id, hsh, subsection, joint_purchase:)
...
       {
         "privacynotice" => 1,
-        "jointpur" => joint_purchase ? 1 : 2,
+        "joint_purchase?" => joint_purchase,
       },
       {
-        "noint" => 1,
-        "jointpur" => joint_purchase ? 1 : 2,
+        "buyer_not_interviewed?" => true,
+        "joint_purchase?" => joint_purchase,
```
natdeanlewissoftwire, line 11: nice

spec/models/form/sales/questions/number_of_others_in_property_spec.rb
```diff
@@ -9,15 +9,15 @@
   let(:joint_purchase) { true }
 
   it "has correct page" do
-    expect(question.page).to eq(page)
+    expect(question.page).to be(page)
```
natdeanlewissoftwire, line 12: are the parentheses needed here? you've removed them below
also I'm not sure be is easier as we only really care about equality, not object instances right? and we can't replace it for strings anyway so I think it adds complexity for not much gain imo

natdeanlewissoftwire (approved):
one suggestion but approving as it's not an essential change 👍

spec/models/form/sales/questions/number_of_others_in_property_spec.rb
arfacamble: parentheses not needed. Kat asked me to leave them in in general for consistency with the rest of the project. I sometimes don't type them if I only have one hand at the time.

In this case I think we do care about the object specifically as we want to be able to call question.page.bla but also in practice == equality for custom objects checks they are the same class and that they have the same id, so ultimately it will be the same test.
How do you think it adds copmlexity?

arfacamble: In general my instinct is to be more specific where possible, so here be rather than eq

PR #1454, opened 2023-03-22
CLDC-2143 Resolve more import validations

kosiakkatrina:
- [x] Remove failing equity field over max on import
- [x] Remove failing equity field under min on import
- [x] Use LA instead of postcode if postcode is in invalid format
- [x] Do not backfill information about people indexed 7 and higher (depends on https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/1452, will need to uncomment xit test when this is merged in)

### DIFF
--- a/app/services/imports/sales_logs_import_service.rb
+++ b/app/services/imports/sales_logs_import_service.rb
@@ -214,6 +214,11 @@ module Imports
         @logs_overridden << sales_log.old_id
         attributes.delete("income1")
         save_sales_log(attributes, previous_status)
+      elsif sales_log.errors.of_kind?(:equity, :over_max)
+        @logger.warn("Log #{sales_log.old_id}: Removing equity as the equity is invalid")
+        @logs_overridden << sales_log.old_id
+        attributes.delete("equity")
+        save_sales_log(attributes, previous_status)
       else
         @logger.error("Log #{sales_log.old_id}: Failed to import")
         sales_log.errors.each do |error|
--- a/app/services/imports/sales_logs_import_service.rb
+++ b/app/services/imports/sales_logs_import_service.rb
@@ -219,6 +219,12 @@ module Imports
         @logs_overridden << sales_log.old_id
         attributes.delete("equity")
         save_sales_log(attributes, previous_status)
+      elsif sales_log.errors.of_kind?(:postcode_full, :wrong_format)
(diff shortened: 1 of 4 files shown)


app/services/imports/sales_logs_import_service.rb
```diff
@@ -544,7 +555,7 @@ def set_default_values(attributes)
       end
 
       # other household members characteristics
-      (2..attributes["hhmemb"]).each do |index|
+      (2..[attributes["hhmemb"], 6].min).each do |index|
```
ghost, line 558: can this ever happen?
Screenshot 2023-03-23 at 08 39 38

ghost (approved):

app/services/imports/sales_logs_import_service.rb
kosiakkatrina: Hmm, I don't think so, cause `other_people_with_details` always sets it to a number, but I'll keep an eye out!

app/services/imports/sales_logs_import_service.rb
```diff
@@ -214,6 +214,17 @@ def rescue_validation_or_raise(sales_log, attributes, previous_status, exception
...
+      elsif sales_log.errors.of_kind?(:equity, :over_max) || sales_log.errors.of_kind?(:equity, :under_min)
+        @logger.warn("Log #{sales_log.old_id}: Removing equity as the equity is invalid")
+        @logs_overridden << sales_log.old_id
+        attributes.delete("equity")
+        save_sales_log(attributes, previous_status)
+      elsif sales_log.errors.of_kind?(:postcode_full, :wrong_format)
+        @logger.warn("Log #{sales_log.old_id}: Removing postcode as the postcode is invalid")
+        @logs_overridden << sales_log.old_id
+        attributes.delete("postcode_full")
+        attributes["pcodenk"] = attributes["la"].present? ? 1 : nil
```
jamesrose, line 226: I think we should keep this kind of logic in the main block — it will be easy to miss, otherwise.

kosiakkatrina: What do you mean the main block? This is only meant to set it if the postcode is cleared as a result of validation, cause then we'd want to set it to not known  and get the la persisted instead

jamesrose: Can we not set `pcodenk` earlier in `create_log`?

PR #1455, opened 2023-03-22
CLDC-2140 Scheme and location updates

natdeanlewissoftwire:
this adds a few new features:

new complete/incomplete location breakdown: 
image

validation if a scheme has no completed locations: 
image

also, only active locations show on the locations question now and schemes and locations are only confirmed when they are complete and the user clicks "save"/"create" on the cya page

https://digital.dclg.gov.uk/jira/browse/CLDC-2140


app/models/scheme.rb
```diff
@@ -166,7 +166,7 @@ def synonyms
   end
 
   def appended_text
-    "#{locations.count { |location| location.startdate.blank? || location.startdate <= Time.zone.today }} locations"
+    "#{confirmed_locations_count} completed #{'location'.pluralize(confirmed_locations_count)}, #{unconfirmed_locations_count} incomplete #{'location'.pluralize(unconfirmed_locations_count)}"
```
natdeanlewissoftwire: we use language completed/incomplete on the frontend but actually this is just the same as confirmed/unconfirmed for purposes in the code. It made the most sense to juxtapose these here rather than earlier on e.g. by making a completed scope that was actually just checking for confirmed

spec/models/form/lettings/questions/location_id_spec.rb
```diff
@@ -70,17 +72,19 @@
...
           FactoryBot.create(:location, scheme:, startdate: Time.utc(2023, 1, 1))
+          lettings_log.update!(scheme:)
         end
 
         it "the displayed_answer_options is an empty hash" do
           expect(question.displayed_answer_options(lettings_log)).to eq({})
         end
       end
 
-      context "and the locations have a no startdate" do
```
natdeanlewissoftwire, line 80: 🇮🇹

app/controllers/locations_controller.rb
```diff
@@ -134,7 +134,9 @@ def update_availability
   def check_answers; end
 
   def confirm
-    flash[:notice] = "#{@location.postcode} #{@location.startdate.blank? || @location.startdate < Time.zone.now ? 'has been' : 'will be'} added to this scheme"
+    if @location.confirmed
+      flash[:notice] = "#{@location.postcode} #{@location.startdate.blank? || @location.startdate < Time.zone.now ? 'has been' : 'will be'} added to this scheme"
```
jamesrose: This logic should be in a helper.

app/models/scheme.rb
jamesrose: This logic should not be in the model — look at moving to a helper.

app/models/scheme.rb
```diff
@@ -264,4 +264,8 @@ def has_confirmed_locations?
   def confirmed_locations_count
     locations.confirmed.size
   end
+
+  def unconfirmed_locations_count
+    locations.unconfirmed.size
+  end
```
jamesrose: You shouldn't need to extract these methods out.

app/models/validations/setup_validations.rb
```diff
@@ -20,6 +20,14 @@ def validate_location(record)
     location_during_startdate_validation(record, :location_id)
   end
 
+  def validate_scheme_has_confirmed_locations_validation(record)
+    return unless record.scheme
+
+    unless record.scheme.has_confirmed_locations?
```
jamesrose:
```suggestion
    unless record.scheme.locations.confirmed.any?
```

app/models/validations/soft_validations.rb
```diff
@@ -66,6 +66,12 @@ def all_tenants_age_and_gender_information_completed?
     end
   end
 
+  def all_tenants_gender_information_completed?
+    (1..hhmemb).all? do |n|
+      public_send("sex#{n}").present? && details_known_or_lead_tenant?(n)
+    end
+  end
```
jamesrose: Are these changes related to this PR?

app/controllers/locations_controller.rb
```diff
@@ -134,7 +134,7 @@ def update_availability
   def check_answers; end
 
   def confirm
-    flash[:notice] = "#{@location.postcode} #{@location.startdate.blank? || @location.startdate < Time.zone.now ? 'has been' : 'will be'} added to this scheme"
+    flash[:notice] = helpers.location_creation_success_notice(@location)
```
natdeanlewissoftwire, line 137: moved to helper

app/models/scheme.rb
```diff
@@ -161,18 +161,6 @@ def check_support_attributes
...
-  end
-
-  def appended_text
-    "#{locations.count { |location| location.startdate.blank? || location.startdate <= Time.zone.today }} locations"
-  end
-
-  def hint
-    [primary_client_group, secondary_client_group].filter(&:present?).join(", ")
-  end
-
```
natdeanlewissoftwire, line 175: moved to helper

app/models/validations/soft_validations.rb
natdeanlewissoftwire: nope, just because diff is vs 1854 not main

app/views/form/_select_question.html.erb
```diff
@@ -9,9 +9,9 @@
   hint: { text: question.hint_text&.html_safe }) do %>
   <% answers.each do |answer| %>
     <option value="<%= answer.id %>"
-      data-synonyms="<%= question.answer_option_synonyms(answer.resource) %>"
-      data-append="<%= question.answer_option_append(answer.resource) %>"
-      data-hint="<%= question.answer_option_hint(answer.resource) %>"
+      data-synonyms="<%= answer_option_synonyms(answer.resource) %>"
+      data-append="<%= answer_option_append(answer.resource) %>"
+      data-hint="<%= answer_option_hint(answer.resource) %>"
```
jamesrose: What can `answer.resource` be in all instances?

natdeanlewissoftwire: The resource is either "Select an option" or a scheme for the scheme typeahead, and for org typeaheads etc it's just the string of the org name so passing the resource allows us to test if it's an instance of a Scheme before we display Scheme specific synonyms/append/hint text

app/helpers/locations_helper.rb
```diff
@@ -78,6 +78,12 @@ def toggle_location_link(location)
     return govuk_button_link_to "Reactivate this location", scheme_location_new_reactivation_path(location.scheme, location) if location.deactivated?
   end
 
+  def location_creation_success_notice(location)
+    if location.confirmed
+      "#{location.postcode} #{location.startdate.blank? || location.startdate < Time.zone.now ? 'has been' : 'will be'} added to this scheme"
+    end
```
arfacamble, line 84: I believe the condition here can be slightly more concise `location.startdate&.before? Time.zone.now`

app/helpers/question_view_helper.rb
```diff
@@ -13,6 +13,26 @@ def legend(question, page_header, conditional)
     }
   end
 
+  def answer_option_synonyms(resource)
+    return unless resource.instance_of?(Scheme)
```
arfacamble, line 17: I find `resource.is_a? Scheme` slightly more readable, but I'm pretty sure they are aliases, I jsut mention it in case you know of one and not the other

arfacamble, line 17: I find `resource.is_a? Scheme` slightly more readable, but I'm pretty sure they are aliases, I jsut mention it in case you know of one and not the other

app/helpers/question_view_helper.rb
```diff
@@ -13,6 +13,26 @@ def legend(question, page_header, conditional)
...
+
+    confirmed_locations_count = resource.locations.confirmed.size
+    unconfirmed_locations_count = resource.locations.unconfirmed.size
+    "#{confirmed_locations_count} completed #{'location'.pluralize(confirmed_locations_count)}, #{unconfirmed_locations_count} incomplete #{'location'.pluralize(unconfirmed_locations_count)}"
+  end
+
+  def answer_option_hint(resource)
+    return unless resource.instance_of?(Scheme)
+
+    [resource.primary_client_group, resource.secondary_client_group].filter(&:present?).join(", ")
```
arfacamble: `.filter(&:present?)` does the same as `.compact` I believe.
Also, I may not have the full picture yet, but if both primary and secondary client groups are nil, then this will return an empty string. Is that fine?

arfacamble (changes requested):
have got as far as _select_question.html.erb
note for me to remember, or kind of a question for you because I probably just don't understand. It feels weird to me that we have view helpers doing things that are model specific. Is there any way we can do this in a controller before we pass to a view?

spec/models/form/lettings/questions/location_id_spec.rb
arfacamble: this is top quality bants,
turns out 🇱🇻 though

app/helpers/locations_helper.rb
natdeanlewissoftwire: will use the before but nil coalescing isn't equivalent here

app/helpers/question_view_helper.rb
natdeanlewissoftwire: I prefer `instance_of?` in this case as it doesn't allow subclasses which I think is preferable conceptually

app/helpers/question_view_helper.rb
natdeanlewissoftwire: only difference is .compact won't remove empty strings, but that's fine for this purpose - I'll change to .compact

app/helpers/locations_helper.rb
arfacamble: how comesies not equivalent?
location.startdate type is DateTime, so surely it's either a date or nil?

app/helpers/question_view_helper.rb
arfacamble: rad

app/helpers/question_view_helper.rb
arfacamble: wiki-wiki-wiki-wickeeeeed

app/helpers/locations_helper.rb
natdeanlewissoftwire: yep but we want nil to pass the ternary condition

arfacamble: pleasure doing business sir

natdeanlewissoftwire: 🤝

arfacamble (approved):

PR #1456, opened 2023-03-22
CLDC-1854 Make schemes with 0 locations be incomplete rather than active

natdeanlewissoftwire:
https://digital.dclg.gov.uk/jira/browse/CLDC-1854
the status of this scheme would previously have been active, this PR just changes location-less schemes to be incomplete: 
image


app/models/location.rb
```diff
@@ -23,6 +23,7 @@ class Location < ApplicationRecord
   scope :search_by, ->(param) { search_by_name(param).or(search_by_postcode(param)) }
   scope :started, -> { where("startdate <= ?", Time.zone.today).or(where(startdate: nil)) }
   scope :active, -> { where(confirmed: true).and(started) }
+  scope :confirmed, -> { where(confirmed: true) }
```
arfacamble, line 26: I would like to know what you understand about this because I don't know it.
These are ... procs?
They have things inside them that looks a bit like SQL, so it's probably something like a bit fo a DB query that you can tag onto other active record calls?

app/models/scheme.rb
```diff
@@ -254,4 +254,14 @@ def reactivating_soon?
...
+
+  def has_confirmed_locations?
+    confirmed_locations_count.positive?
+  end
+
+private
+
+  def confirmed_locations_count
+    locations.confirmed.size
+  end
```
arfacamble: if this one line method is only being called in one other one line method a few lines above I feel they could just be folded into one.
Personal preference is around 3 or 4 chained method calls to put it on separate lines
```
locations
  .confirmed
  .count
  .positive?
```

spec/helpers/schemes_helper_spec.rb
```diff
@@ -110,166 +110,213 @@
...
+          { name: "Primary client group", value: "Rough sleepers" },
+          { name: "Has another client group", value: "Yes" },
+          { name: "Secondary client group", value: "Refugees (permanent)" },
+          { name: "Level of support given", value: "High level" },
+          { name: "Intended length of stay", value: "Permanent" },
+          { name: "Availability", value: "Active from 1 April 2022" },
+          { name: "Status", value: status_tag(:incomplete) },
+        ]
+        expect(display_scheme_attributes(scheme, coordinator_user)).to eq(attributes)
+      end
```
arfacamble, line 151: this is a really disappointing effort at diffing from github

spec/helpers/schemes_helper_spec.rb
arfacamble, line 1: are there particular parts of this file that are worth looking at in more detail? It mostly looks like the a lot of the same things moved around a bit under more sensible context blocks.

spec/models/validations/date_validations_spec.rb
```diff
@@ -197,6 +198,7 @@
       let(:scheme) { create(:scheme) }
 
       before do
+        FactoryBot.create(:location, scheme:)
         create(:scheme_deactivation_period, deactivation_date: Time.zone.local(2022, 6, 4), reactivation_date: Time.zone.local(2022, 8, 4), scheme:)
```
arfacamble, line 194: I haven't looked into this but seen it around. What is the difference between `FactoryBot.create(:syntax)` and other test files eg here where it just says `create(:same_syntax_in_here)`?

arfacamble (approved):
I think only one place where I have requested a small optional change. Mostly a couple of questions

app/models/scheme.rb
arfacamble: actually if that proc above is the only place we're using this .confirmed then I feel like this would only be more readable and probably equally efficient if we just used active record directly?
`locations.count(confirmed: true).positive?

app/models/location.rb
arfacamble: is this used in more than one place? or just the confirmed_locations_count method? If so is it worth not jsut leaving in that method and extracting when it wants to be used in more places?
I feel like it might be easier for that method to just be `locations.where(confirmed: true).count.positive?` and not have extra procs and private methods

natdeanlewissoftwire: I just see them as active record helper methods/filters - totally agree with your second comment, but I put this in as I'll be using it in a couple more places in the [CLDC-2140 PR ](https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/1455) so that's why I've defined this here to avoid unneccessary diffs

app/models/scheme.rb
natdeanlewissoftwire: again for readability I've done this separately as in CLDC-2140 I need unconfirmed_locations_count as well and pass them into a string so will keep this too

spec/helpers/schemes_helper_spec.rb
natdeanlewissoftwire: yeah it's mostly rearrangement, have just tested status is different for schemes with/without locations

spec/models/validations/date_validations_spec.rb
natdeanlewissoftwire: I think they're identical, I've tried to just be consistent with whatever's in a given file (but I wasn't here, I'll change it)

jamesrose:
@natdeanlewissoftwire The alert in the screenshot above should be captalized. Please could you update that?

app/models/location.rb
jamesrose: We generally do keep status-like queries as scoped @arfacamble

app/models/scheme.rb
jamesrose: You don't need either of these methods. You can just do:

```ruby
def status_at(date)
  return :incomplete unless confirmed && locations.confirmed.any?
  # ... 
```

arfacamble: 💣 💣 💣 💣 💣 💣

natdeanlewissoftwire: that is much tidier @jamesrose I agree, I still stand by the point about readability in connection to CLDC-2140 however - specifically https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/1455/files#diff-0cd65118d88b34f32b919f37a20a1024d03620f8a892f3f020fe70b0193e5a61R169 where I do think this having confirmed and unconfirmed counts is the best way of working (unless there's a VERY brief active record query I could use?), and given that I think it makes sense to use those methods here too

PR #1457, opened 2023-03-22

asmega:
# Context

- We now let users decide for bulk upload when there are errors whether to re-upload the csv again or create and fix inline

# Changes

- New mini journey if we can't create logs and there are errors let user decide how to fix these error
- For performance we create the logs ahead of time where we can which are not visible
- to handle this `default_scope` added to `LettingsLog`

# Review notes

- We create the logs synchronously on the web box which is less than ideal. We can improve this in a later iteration to create them in advance then hide them

# Screenshots

## Landing page from email

## Confirm action


app/services/bulk_upload/processor.rb
```diff
@@ -15,11 +15,16 @@ def call
...
-      send_fix_errors_mail if created_logs_but_incompleted?
-      send_success_mail if created_logs_and_all_completed?
+      if bulk_upload.bulk_upload_errors.count.zero?
+        create_logs
+
+        send_fix_errors_mail if created_logs_but_incompleted?
+        send_success_mail if created_logs_and_all_completed?
+      else
+        send_how_fix_upload_mail
+      end
```
asmega: this branching strategy is not perfect as ideally we would just create the logs assuming no errors. however as the log validation is not perfect it's possible we don't capture all the errors ahead of time therefore could create the logs but still end up with them in progress. therefore i have left the old system of fixing inline present

jamesrose (approved):

asmega:
closing this in favour of https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/1497 which has some the un-used migrations removed

PR #1460, opened 2023-03-22
CLDC-2139 Show location details under scheme when inferred

natdeanlewissoftwire:
Shows an inferred location's postcode and name in grey under the scheme.

when inferred we show scheme in black and then location postcode and name in grey:
image

when not inferred we show postcode in black, and then LA and name in grey:
image

ticket: https://digital.dclg.gov.uk/jira/browse/CLDC-2139


app/models/form/question.rb
```diff
@@ -317,7 +317,7 @@ def evaluate_condition(condition, log)
   end
 
   def enabled_inferred_answers(inferred_answers, log)
-    inferred_answers.filter { |_key, value| value.all? { |condition_key, condition_value| log[condition_key] == condition_value } }
+    inferred_answers.filter { |_key, value| value.all? { |condition_key, condition_value| log.public_send(condition_key) == condition_value } }
```
natdeanlewissoftwire: needed so the inferred_answers can acccept log methods as well as attributes

kosiakkatrina: Would it work if we delegated the location name to the log? Wouldn't need to change any methods

natdeanlewissoftwire: I don't think so - it's the `:scheme_has_multiple_locations?` method that I need the public send for

app/models/form/lettings/questions/scheme_id.rb
```diff
@@ -9,6 +9,11 @@ def initialize(_id, hsh, page)
     @guidance_position = GuidancePosition::BOTTOM
     @guidance_partial = "scheme_selection"
     @question_number = 9
+    @inferred_answers = {
+      "location.name": {
+        "scheme_has_multiple_locations?": false,
+      },
+    }
```
arfacamble, line 16: I haven't used this before. I assume it means that if the scheme does not have multiple locations, then we infer the location when the scheme is selected.
I assume this is so that we can clear inferred answers if this is cleared?

app/models/form/lettings/questions/scheme_id.rb
```diff
@@ -41,6 +46,10 @@ def hidden_in_check_answers?(lettings_log, _current_user = nil)
     !supported_housing_selected?(lettings_log)
   end
 
+  def get_extra_check_answer_value(lettings_log)
+    lettings_log.form.get_question("postcode_full", nil).label_from_value(lettings_log.postcode_full) unless lettings_log.scheme_has_multiple_locations?
+  end
```
arfacamble, line 51: given the content of this method seems to be specific to postcode, I feel like the name of the method could be specific also. perhaps `get_postcode_if_inferable` / `get_postcode_if_inferred`

app/models/form/question.rb
arfacamble: I feel pretty strongly that we shouldn't have nested one line code blocks, I think this should be spread across multiple lines.

I also don't know how these inferred answers things work, their form or what they're for exactly, so it would be good to rename `|_key, value|` with something a little more descriptive.

app/models/form/lettings/questions/scheme_id.rb
arfacamble: I see now. that is a generic method called from view

app/models/form/lettings/questions/scheme_id.rb
arfacamble: is location.name used somewhere? As above I haven't used this part of the codebase. I assume it must be, otehrwise why wouldn't the label just be location_name?

arfacamble: oh i see. it's the slightly dubiously written `Question:71`

app/models/form/question.rb
natdeanlewissoftwire: I think these are good ideas

arfacamble (approved):

PR #1469, opened 2023-03-24
CLDC-2072 changing startdate to another collection year clear invalid answers

arfacamble:
https://digital.dclg.gov.uk/jira/browse/CLDC-2072

Having previously written a small method that did this independently. On reflection it was better to incorporate this behaviour into some adjacent logic, take that opportunity to revamp the existing code to improve readability, and write more comprehensive tests to go with it.

I recommend reading the tests first as they describe the intended behaviour well I think.
I have written comments in line in the form.rb diff to explain changes there
I have ended up with around 30 lines more code, but I hope that is at the gain of much more readability, I am happy to talk through why the original `invalidated_page_questions` method was confusing if that would help

### DIFF
--- a/app/models/form.rb
+++ b/app/models/form.rb
@@ -239,9 +239,9 @@ class Form
 
   def routed_and_not_routed_questions_by_type(log, type: nil, current_user: nil)
     questions_by_type = if type
-                          questions.reject { |q| q.type != type || q.do_not_clear }
+                          questions.reject { |q| q.type != type || q.disable_clearing_if_not_routed_or_dynamic_answer_options }
                         else
-                          questions.reject { |q| %w[radio checkbox].include?(q.type) || q.do_not_clear }
+                          questions.reject { |q| %w[radio checkbox].include?(q.type) || q.disable_clearing_if_not_routed_or_dynamic_answer_options }
                         end
     routed, not_routed = questions_by_type.partition { |q| q.page.routed_to?(log, current_user) || q.derived? }
     { routed:, not_routed: }
--- a/app/models/form.rb
+++ b/app/models/form.rb
@@ -179,21 +179,21 @@ class Form
   end
 
   def reset_not_routed_questions_and_invalid_answers(log)
-    reset_not_routed_checkbox_questions(log)
+    reset_checkbox_questions_if_not_routed(log)
 
     reset_radio_questions_if_not_routed_or_invalid_answers(log)
 
-    reset_not_routed_free_user_input_questions(log)
+    reset_free_user_input_questions_if_not_routed(log)
   end
 
-  def reset_not_routed_checkbox_questions(log)
+  def reset_checkbox_questions_if_not_routed(log)
     checkbox_questions = routed_and_not_routed_questions_by_type(log, type: "checkbox")
-    checkbox_questions[:invalid].each do |invalidated_question|
-      valid_options = checkbox_questions[:valid]
-                                        .select { |q| q.id == invalidated_question.id }
+    checkbox_questions[:not_routed].each do |not_routed_question|
+      valid_options = checkbox_questions[:routed]
+                                        .select { |q| q.id == not_routed_question.id }
                                         .flat_map { |q| q.answer_options.keys }
-      invalidated_question.answer_options.each_key do |invalid_option|
-        if !log.respond_to?(invalid_option) || valid_options.include?(invalid_option)
+      not_routed_question.answer_options.each_key do |invalid_option|
+        if !log.respond_to?(invalid_option) || valid_options.include?(invalid_option) || log.public_send(invalid_option).nil?
           next
         else
           clear_attribute(log, invalid_option)
@@ -204,47 +204,47 @@ class Form
 
   def reset_radio_questions_if_not_routed_or_invalid_answers(log)
     radio_questions = routed_and_not_routed_questions_by_type(log, type: "radio")
-    valid_radio_options = radio_questions[:valid]
-                                   .group_by(&:id)
-                                   .transform_values! { |q_array| q_array.flat_map { |q| q.answer_options.keys } }
-    radio_questions[:invalid].each do |invalidated_question|
-      q_id = invalidated_question.id
-      if !log.respond_to?(q_id) || log.public_send(q_id).nil? || valid_radio_options.key?(q_id)
+    valid_radio_options = radio_questions[:routed]
+                                         .group_by(&:id)
+                                         .transform_values! { |q_array| q_array.flat_map { |q| q.answer_options.keys } }
+    radio_questions[:not_routed].each do |not_routed_question|
+      question_id = not_routed_question.id
+      if !log.respond_to?(question_id) || log.public_send(question_id).nil? || valid_radio_options.key?(question_id)
         next
       else
-        clear_attribute(log, q_id)
+        clear_attribute(log, question_id)
       end
     end
-    valid_radio_options.each do |q_id, valid_options|
-      if !log.respond_to?(q_id) || valid_options.include?(log.public_send(q_id).to_s)
+    valid_radio_options.each do |question_id, valid_options|
+      if !log.respond_to?(question_id) || valid_options.include?(log.public_send(question_id).to_s)
         next
       else
-        clear_attribute(log, q_id)
+        clear_attribute(log, question_id)
       end
     end
   end
 
-  def reset_not_routed_free_user_input_questions(log)
+  def reset_free_user_input_questions_if_not_routed(log)
     non_radio_checkbox_questions = routed_and_not_routed_questions_by_type(log)
-    enabled_question_ids = non_radio_checkbox_questions[:valid].map(&:id)
-    non_radio_checkbox_questions[:invalid].each do |invalidated_question|
-      q_id = invalidated_question.id
-      if log.public_send(q_id).nil? || enabled_question_ids.include?(q_id)
+    enabled_question_ids = non_radio_checkbox_questions[:routed].map(&:id)
+    non_radio_checkbox_questions[:not_routed].each do |not_routed_question|
+      question_id = not_routed_question.id
+      if log.public_send(question_id).nil? || enabled_question_ids.include?(question_id)
         next
       else
-        clear_attribute(log, q_id)
+        clear_attribute(log, question_id)
       end
     end
   end
 
   def routed_and_not_routed_questions_by_type(log, type: nil, current_user: nil)
-    questions_to_categorise = if type
-                                questions.reject { |q| q.type != type || q.do_not_clear }
-                              else
-                                questions.reject { |q| %w[radio checkbox].include?(q.type) || q.do_not_clear }
-                              end
-    valid, invalid = questions_to_categorise.partition { |q| q.page.routed_to?(log, current_user) || q.derived? }
-    { valid:, invalid: }
+    questions_by_type = if type
+                          questions.reject { |q| q.type != type || q.do_not_clear }
+                        else
+                          questions.reject { |q| %w[radio checkbox].include?(q.type) || q.do_not_clear }
+                        end
+    routed, not_routed = questions_by_type.partition { |q| q.page.routed_to?(log, current_user) || q.derived? }
+    { routed:, not_routed: }
   end
 
   def clear_attribute(log, attribute)
--- a/app/models/form.rb
+++ b/app/models/form.rb
@@ -178,48 +178,80 @@ class Form
     pages.reject { |p| p.routed_to?(log, current_user) }
   end
 
-  def invalidated_questions(log)
-    invalidated_page_questions(log) + invalidated_conditional_questions(log)
+  def reset_not_routed_questions_and_invalid_answers(log)
+    reset_not_routed_checkbox_questions(log)
+
+    reset_radio_questions_if_not_routed_or_invalid_answers(log)
+
+    reset_not_routed_free_user_input_questions(log)
   end
 
-  def invalidated_page_questions(log, current_user = nil)
-    # we're already treating these fields as a special case and reset their values upon saving a log
-    callback_questions = %w[postcode_known la ppcodenk previous_la_known prevloc postcode_full ppostcode_full location_id address_line1 address_line2 town_or_city county]
-    questions.reject { |q| q.page.routed_to?(log, current_user) || q.derived? || callback_questions.include?(q.id) } || []
-  end
-
-  def reset_not_routed_questions(log)
-    enabled_questions = enabled_page_questions(log)
-    enabled_question_ids = enabled_questions.map(&:id)
-
-    invalidated_page_questions(log).each do |question|
-      if %w[radio checkbox].include?(question.type)
-        enabled_answer_options = enabled_question_ids.include?(question.id) ? enabled_questions.find { |q| q.id == question.id }.answer_options : {}
-        current_answer_option_valid = enabled_answer_options.present? ? enabled_answer_options.key?(log.public_send(question.id).to_s) : false
-
-        if !current_answer_option_valid && log.respond_to?(question.id.to_s)
-          Rails.logger.debug("Cleared #{question.id} value")
-          log.public_send("#{question.id}=", nil)
+  def reset_not_routed_checkbox_questions(log)
+    checkbox_questions = routed_and_not_routed_questions_by_type(log, type: "checkbox")
+    checkbox_questions[:invalid].each do |invalidated_question|
+      valid_options = checkbox_questions[:valid]
+                                        .select { |q| q.id == invalidated_question.id }
+                                        .flat_map { |q| q.answer_options.keys }
+      invalidated_question.answer_options.keys.each do |invalid_option|
+        if !log.respond_to?(invalid_option) || valid_options.include?(invalid_option)
+          next
         else
-
-          (question.answer_options.keys - enabled_answer_options.keys).map do |invalid_answer_option|
-            Rails.logger.debug("Cleared #{invalid_answer_option} value")
-            log.public_send("#{invalid_answer_option}=", nil) if log.respond_to?(invalid_answer_option)
-          end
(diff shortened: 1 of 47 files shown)


app/models/lettings_log.rb
```diff
@@ -79,6 +80,10 @@ def form_name
     FormHandler.instance.form_name_from_start_year(collection_start_year, "lettings")
   end
 
+  def clear_invalid_radio_answers!
+    super if startdate_was && @start_year != FormHandler.instance.collection_start_year(startdate_was)
```
natdeanlewissoftwire: you might have tried this already, but sales logs have a startdate method that just sets startdate = saledate, so you might be able to put all this logic on the log model without separating out for lettings and sales

natdeanlewissoftwire: didn't know about attribute_was though, cool!

jamesrose: Will this clear answers if `startdate_was` is nil? Is it possible that start year is not set before other answers?

arfacamble: I forgot this, but given that the attribute_was thing is coming from rails magic I think it would be a bit weird to make a parallel method for just this situation.
```
def startdate_was
  saledate_was
end
```

arfacamble: this is an awkwardness yes. hmmm
part of my thinking was that for things like bulk upload we might rather that the user is "allowed" the invalid answer but gets an invalid answer validation, rather than just voiding it and making them revisit. Maybe that's not the most important thought

arfacamble: Actually now I fiddle with it that wasn't my thought!
If I change it to:
`super unless startdate_was && @start_year == FormHandler.instance.collection_start_year(startdate_was)`
then I fail tests:
`BulkUpload::Lettings::Year2023::RowParser validations #field_30 when unpermitted values returns an error`
ie tests that check explicitly that if a value is incorrect they get this validation.
I will check with Phil what his knowledge is from the Bulk Upload side, but yes if we don't check when startdate_was is nil we can definitely find a niche edge case where this bug can be reproduced. My sales log tests on this PR refer to a field whose question is in the setup section

natdeanlewissoftwire (dismissed):
lgtm post addressing comments 👍

app/models/lettings_log.rb
natdeanlewissoftwire: yeah probably fair enough 👍

jamesrose: Please could you update this and add a test case for it?

app/models/form.rb
```diff
@@ -154,48 +154,80 @@ def invalidated_pages(log, current_user = nil)
     pages.reject { |p| p.routed_to?(log, current_user) }
   end
 
-  def invalidated_questions(log)
-    invalidated_page_questions(log) + invalidated_conditional_questions(log)
-  end
```
arfacamble, line 183: sense check on this, but it did not appear to be being called anywhere in the codebase. very slight risk that the method is being called through some metaprogramming whatnot?

spec/models/form_spec.rb
```diff
@@ -205,32 +205,140 @@ def answer_local_authority(lettings_log)
...
+      end
+    end
+
+    context "when an attribute is related to a callback question with no set answer options, and no questions for that attribute are routed to" do
+      let(:location) { FactoryBot.create(:location) }
+      let(:log) { FactoryBot.create(:lettings_log, :startdate_today) }
 
-      it "does not route to the page" do
-        expect(form.invalidated_pages(lettings_log).map(&:id)).to include("outstanding_amount_known")
+      # Pages::PropertyPostcode and questions inside have been removed from form. do we not want to delete and migration delete_column?
```
arfacamble, line 335: this question is mostly directed at @bibblobcode 
On reflection I'm assuming that the columns we may want to keep for existing data/migrated data, but since the lettings forms made from `Questions::BlaBla` are only for 23/24 maybe they can be deleted?

ghost: - yes we should probably keep the data
- I am not sure I understand, are you suggesting to delete 23/24 questions from Lettings?

arfacamble:
yes the latter. the Pages::PropertyPostcode is no longer rendered anywhere.
and then therefore the Questions::Bla classes inside that

spec/models/form_spec.rb
arfacamble: yes the latter. the Pages::PropertyPostcode is no longer rendered anywhere. and then therefore the Questions::Bla classes inside that

app/models/form.rb
```diff
@@ -154,48 +154,80 @@ def invalidated_pages(log, current_user = nil)
...
+  def routed_and_not_routed_questions_by_type(log, type: nil, current_user: nil)
+    # we're already treating these fields as a special case and reset their values upon saving a log
+    callback_questions = %w[postcode_known la ppcodenk previous_la_known prevloc postcode_full ppostcode_full location_id address_line1 address_line2 town_or_city county]
+    questions_to_categorise = if type
+                                questions.reject { |q| q.type != type || callback_questions.include?(q.id) }
+                              else
+                                questions.reject { |q| %w[radio checkbox].include?(q.type) || callback_questions.include?(q.id) }
+                              end
+    valid, invalid = questions_to_categorise.partition { |q| q.page.routed_to?(log, current_user) || q.derived? }
+    { valid:, invalid: }
```
arfacamble: this method is a development on `invalidated_page_questions` on old line 160

app/models/form.rb
```diff
@@ -154,48 +154,80 @@ def invalidated_pages(log, current_user = nil)
...
-    enabled_questions = enabled_page_questions(log)
-    enabled_question_ids = enabled_questions.map(&:id)
+    reset_radio_questions_if_not_routed_or_invalid_answers(log)
 
-    invalidated_page_questions(log).each do |question|
-      if %w[radio checkbox].include?(question.type)
-        enabled_answer_options = enabled_question_ids.include?(question.id) ? enabled_questions.find { |q| q.id == question.id }.answer_options : {}
-        current_answer_option_valid = enabled_answer_options.present? ? enabled_answer_options.key?(log.public_send(question.id).to_s) : false
+    reset_not_routed_free_user_input_questions(log)
+  end
```
arfacamble, line 187: the above method is a development on `reset_not_routed_questions` and now delegates to methods for questions of each type

app/models/form.rb
arfacamble, line 1: Improvements for readability should be:
- old method did not separate out between radio and checkbox so you have strange control flow eg l.176 `if !current_answer_option_valid && log.respond_to?(question.id.to_s)` sends all checkbox questions down the else as their id is not an attribute on the log, but also sends all valid radio options with them.
- hard to process what 173 and 174 are doing exactly for radio questions and checkbox questions since their keys are not the same thing at all, respectively values of attributes and the attributes themselves. I believe `l:174 log.public_send(question.id).to_s` would crash the app if there was a situation in the future where we had two checkbox questions with the same question id
- old method printed 'cleared X value' lots of times when nothing was being cleared (the value was already nil)

arfacamble: Can anyone see more ways to commonise the three reset_not_routed_x_questions beyond the clear_attribute method I have pulled out of them?
I can't really, and I think that is my main justification for why they should be separated out despite many more lines of code.

spec/services/bulk_upload/lettings/year2022/row_parser_spec.rb
arfacamble, line 1: @asmega I believe the four deleted tests in the two row parser spec files are the kinds of thing we have discussed where these will have to be tested in some other way for bulk upload

app/models/form.rb
```diff
@@ -154,48 +154,80 @@ def invalidated_pages(log, current_user = nil)
...
+    end
+  end
+
+  def reset_radio_questions_if_not_routed_or_invalid_answers(log)
+    radio_questions = routed_and_not_routed_questions_by_type(log, type: "radio")
+    valid_radio_options = radio_questions[:valid]
+                                   .group_by(&:id)
+                                   .transform_values! { |q_array| q_array.flat_map { |q| q.answer_options.keys } }
+    radio_questions[:invalid].each do |invalidated_question|
+      q_id = invalidated_question.id
```
```suggestion
      question_id = invalidated_question.id
```

app/models/form.rb
```diff
@@ -154,48 +154,80 @@ def invalidated_pages(log, current_user = nil)
...
+        clear_attribute(log, q_id)
+      end
+    end
+  end
+
+  def reset_not_routed_free_user_input_questions(log)
+    non_radio_checkbox_questions = routed_and_not_routed_questions_by_type(log)
+    enabled_question_ids = non_radio_checkbox_questions[:valid].map(&:id)
+    non_radio_checkbox_questions[:invalid].each do |invalidated_question|
+      q_id = invalidated_question.id
```
```suggestion
      question_id = invalidated_question.id
```

app/models/form.rb
```diff
@@ -154,48 +154,80 @@ def invalidated_pages(log, current_user = nil)
...
+        clear_attribute(log, q_id)
       end
     end
   end
 
-  def enabled_page_questions(log)
-    questions - invalidated_page_questions(log)
+  def routed_and_not_routed_questions_by_type(log, type: nil, current_user: nil)
+    # we're already treating these fields as a special case and reset their values upon saving a log
+    callback_questions = %w[postcode_known la ppcodenk previous_la_known prevloc postcode_full ppostcode_full location_id address_line1 address_line2 town_or_city county]
```
jamesrose: The `Form` model should have no specific knowledge of the questions within an instance of a form. Can you ask the form for this information instead of hard-coding it?

app/models/form.rb
arfacamble: very good. I did make ti this at some point and I felt like it made the if line a bit harder to read just because it got so much longer!
What is your feeling about putting condition statements over multiple lines? That was common in my old c# project but I feel like it's not generally encouraged in RoR?

app/models/form.rb
arfacamble: I support this and I think that this list is different categories of things and it might be better to interrogate what they are and why.
eg. they are mostly radio questions where the options are populated on render I believe. For those questions the best way to get that information form the Form that occurs to me is questions that have no answer options when this code runs. Do you think that is robust?
then I will look into the other questions , but for instance I do not know why postcode_known is in this list, I assume that at some point for some other ticket someone didn't want to it to be cleared and just put it here without thinking further.

app/models/form.rb
```diff
@@ -154,48 +154,80 @@ def invalidated_pages(log, current_user = nil)
...
-    callback_questions = %w[postcode_known la ppcodenk previous_la_known prevloc postcode_full ppostcode_full location_id address_line1 address_line2 town_or_city county]
-    questions.reject { |q| q.page.routed_to?(log, current_user) || q.derived? || callback_questions.include?(q.id) } || []
-  end
+  def reset_not_routed_questions_and_invalid_answers(log)
+    reset_not_routed_checkbox_questions(log)
 
-  def reset_not_routed_questions(log)
-    enabled_questions = enabled_page_questions(log)
-    enabled_question_ids = enabled_questions.map(&:id)
+    reset_radio_questions_if_not_routed_or_invalid_answers(log)
```
natdeanlewissoftwire, line 184: inconsistent naming :( maybe this is a clue there should be more function separation in this case?

app/models/form.rb
```diff
@@ -154,48 +154,80 @@ def invalidated_pages(log, current_user = nil)
...
-          log.public_send("#{question.id}=", nil)
+  def reset_not_routed_checkbox_questions(log)
+    checkbox_questions = routed_and_not_routed_questions_by_type(log, type: "checkbox")
+    checkbox_questions[:invalid].each do |invalidated_question|
+      valid_options = checkbox_questions[:valid]
+                                        .select { |q| q.id == invalidated_question.id }
+                                        .flat_map { |q| q.answer_options.keys }
+      invalidated_question.answer_options.each_key do |invalid_option|
+        if !log.respond_to?(invalid_option) || valid_options.include?(invalid_option)
+          next
```
natdeanlewissoftwire, line 197: is this (and the other few) next needed? can't we just say 
```
clear_attribute(log, invalid_option) unless !log.respond_to?(invalid_option) || valid_options.include?(invalid_option)  
```

spec/services/bulk_upload/lettings/year2022/row_parser_spec.rb
```diff
@@ -840,16 +840,6 @@
       end
     end
 
-    describe "#field_134" do
```
natdeanlewissoftwire, line 855: why are we getting rid of these tests?

app/models/form.rb
```diff
@@ -154,48 +154,80 @@ def invalidated_pages(log, current_user = nil)
...
+    reset_not_routed_free_user_input_questions(log)
+  end
 
-        if !current_answer_option_valid && log.respond_to?(question.id.to_s)
-          Rails.logger.debug("Cleared #{question.id} value")
-          log.public_send("#{question.id}=", nil)
+  def reset_not_routed_checkbox_questions(log)
+    checkbox_questions = routed_and_not_routed_questions_by_type(log, type: "checkbox")
+    checkbox_questions[:invalid].each do |invalidated_question|
+      valid_options = checkbox_questions[:valid]
```
natdeanlewissoftwire: should this be `valid_checkbox_options` for consistency?

app/models/form.rb
```diff
@@ -154,48 +154,80 @@ def invalidated_pages(log, current_user = nil)
...
+      invalidated_question.answer_options.each_key do |invalid_option|
+        if !log.respond_to?(invalid_option) || valid_options.include?(invalid_option)
+          next
         else
-
-          (question.answer_options.keys - enabled_answer_options.keys).map do |invalid_answer_option|
-            Rails.logger.debug("Cleared #{invalid_answer_option} value")
-            log.public_send("#{invalid_answer_option}=", nil) if log.respond_to?(invalid_answer_option)
-          end
+          clear_attribute(log, invalid_option)
```
natdeanlewissoftwire, line 199: mabye we could separate into more general methods that:

app/models/form.rb
```diff
@@ -154,48 +154,80 @@ def invalidated_pages(log, current_user = nil)
...
+    enabled_question_ids = non_radio_checkbox_questions[:valid].map(&:id)
+    non_radio_checkbox_questions[:invalid].each do |invalidated_question|
+      q_id = invalidated_question.id
+      if log.public_send(q_id).nil? || enabled_question_ids.include?(q_id)
+        next
+      else
+        clear_attribute(log, q_id)
       end
     end
   end
```
natdeanlewissoftwire, line 238: Do you think we could separate these three resetting methods into more general ones that:

1. get all questions of a type
2. get all valid questions of a type
3. clear attributes unless condition depending on type

then we could run one method that calls all these for each relevant question type.

There could be a cost of readability but I think with good naming it could be pretty clean.

app/models/form.rb
arfacamble: we could but James and Phil have shared various things with me about being careful about overusing unless, with one example being unless with compound conditions.
I think it does make parsing easier to have it separate.

spec/services/bulk_upload/lettings/year2022/row_parser_spec.rb
arfacamble: previously for bulk upload, if there was an invalid radio option, we were adding a validation for that.
now that there are radio options different between years, we really kind of need to clear invalid answers unless we go down a fiddly rabbit hole where we try to check what year we are going from and to and selectively clear individual values.
 Phil is now going to do something else with bulk upload where he checks for fields that have been cleared in the saving process. That's probably more robust as well sense there are already values being cleared under various conditions.

app/models/form.rb
arfacamble: can be

app/models/form.rb
arfacamble: let's pair on that in a min you want, happy to hear your ideas

arfacamble: I think what you're describing sounds like what the original method tried to be, but I would challenge you to discern exactly how that method is working without spending quite a lot of time with it.
I would be happy to combine them if we can work out what to combine, but I think that as the name of the attribute(s) to clear is different for checkbox questions and the conditions under which we clear attributes is somewhat different in each case and particularly different in the case of radio questions, so I can't see how to combine them without having a lot of ifs within that method that either re-check what type of question we have or have other conditions that seem more general but are in fact underneath it checking what type of question we have.
As above let's pair on it briefly this morning and see if your fresh eyes can spot something

app/models/form/lettings/questions/address_line1.rb
```diff
@@ -7,6 +7,7 @@ def initialize(id, hsh, page)
     @type = "text"
     @plain_label = true
     @check_answer_label = "Q12 - Address"
+    @do_not_clear = true
```
jamesrose: Is there a more descriptive name than this? It's not self-evident what it does to me!

spec/services/bulk_upload/lettings/year2022/row_parser_spec.rb
asmega: sorry missed this, but yeah i'll get that sorted

jamesrose (approved):

app/models/form.rb
natdeanlewissoftwire: Fine by me, I'm also anti-unless a lot of places we use it but think it would be better here. Subjective though, your PR so your call :)

spec/services/bulk_upload/lettings/year2022/row_parser_spec.rb
natdeanlewissoftwire: Sounds sensible!

app/models/form.rb
arfacamble: The socially appropriate thing here is clearly to say "OK cool" and move on, ... but I can't because I'm great like that!!!
`unless not one_thing or something_else` is gross.
double negative and unless with compound condition is 💩

PR #1472, opened 2023-03-27
Clear mortgage and outstanding amount when importing

kosiakkatrina:
- Clear mortgage if it's £0
- Clear outstanding charges if are there any outstanding charges is answered as no

### DIFF
--- a/spec/services/imports/lettings_logs_import_service_spec.rb
+++ b/spec/services/imports/lettings_logs_import_service_spec.rb
@@ -710,7 +710,7 @@ RSpec.describe Imports::LettingsLogsImportService do
 
         expect(lettings_log).not_to be_nil
         expect(lettings_log.tshortfall).to be_nil
-        expect(lettings_log.hbrentshortfall).to eq(2)
+        expect(lettings_log.hbrentshortfall).to be_nil
       end
     end
   end
--- a/spec/services/imports/lettings_logs_import_service_spec.rb
+++ b/spec/services/imports/lettings_logs_import_service_spec.rb
@@ -671,5 +671,47 @@ RSpec.describe Imports::LettingsLogsImportService do
         expect(lettings_log.hhmemb).to eq(1)
       end
     end
+
+    context "when there are no outstanding charges but outstanding amount is given" do
+      let(:lettings_log_id) { "0b4a68df-30cc-474a-93c0-a56ce8fdad3b" }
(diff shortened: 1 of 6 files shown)


spec/services/imports/lettings_logs_import_service_spec.rb
```diff
@@ -671,5 +671,47 @@ def open_file(directory, filename)
...
     end
+
+    context "when there are no outstanding charges but outstanding amount is given" do
+      let(:lettings_log_id) { "0b4a68df-30cc-474a-93c0-a56ce8fdad3b" }
+      let(:lettings_log_file) { open_file(fixture_directory, lettings_log_id) }
+      let(:lettings_log_xml) { Nokogiri::XML(lettings_log_file) }
+
+      before do
+        FormHandler.instance.use_fake_forms!
+        Singleton.__init__(FormHandler)
```
asmega, line 682: minor thing but i'm pretty sure you don't need to reset the instance every time and calling `use_fake_forms!` should be enough to mutate the instance. but you will need to call `use_real_forms!` afterwards as you are already doing so

spec/services/imports/lettings_logs_import_service_spec.rb
```diff
@@ -671,5 +671,47 @@ def open_file(directory, filename)
...
+        lettings_log_xml.at_xpath("//xmlns:Q18d").content = "2"
+        lettings_log_xml.at_xpath("//xmlns:Q18dyes").content = "20"
+        lettings_log_xml.at_xpath("//xmlns:_2cYears").content = ""
+        lettings_log_xml.at_xpath("//xmlns:Inj").content = ""
+        lettings_log_xml.at_xpath("//xmlns:LeftAF").content = ""
+      end
+
+      after do
+        FormHandler.instance.use_real_forms!
+        Singleton.__init__(FormHandler)
```
asmega, line 697: same as above and probably don't need this line

asmega (approved):

spec/services/imports/lettings_logs_import_service_spec.rb
kosiakkatrina: Yeah, I was a bit confused, it didn't work without me resetting it. So `FormHandler.instance.current_lettings_form` worked as expected, but the `FormHandler.instance.get_form("current_lettings")` didn't. Even if I put the content of the `FormHandler.instance.current_lettings_form` in `FormHandler.instance.get_form(name)`, it still didn't return the updated form

PR #1477, opened 2023-03-27
CLDC-2188 If dont know joint tenancy dont validate hhmemb

asmega:
# Context

- https://digital.dclg.gov.uk/jira/browse/CLDC-2188

# Changes

- Fix validation where if don't know if joint tenancy don't validate `hhmemb`
- This fix affects both normal web flow and bulk upload

### DIFF
--- a/app/models/validations/tenancy_validations.rb
+++ b/app/models/validations/tenancy_validations.rb
@@ -42,7 +42,7 @@ module Validations::TenancyValidations
   def validate_joint_tenancy(record)
     return unless record.collection_start_year && record.joint
 
-    if record.hhmemb == 1 && record.joint != 2 && record.collection_start_year >= 2022
+    if record.hhmemb == 1 && record.joint == 1 && record.collection_start_year >= 2022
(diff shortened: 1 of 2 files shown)


app/models/validations/tenancy_validations.rb
```diff
@@ -40,6 +40,7 @@ def validate_other_tenancy_type(record)
   end
 
   def validate_joint_tenancy(record)
+    return if record.joint == 3
```
kosiakkatrina: return unless joint is yes and remove the check for != 2 on line 46?

kosiakkatrina (approved):

app/models/validations/tenancy_validations.rb
asmega: ok i think i get you, i've tweaked

kosiakkatrina: Do we need this return at all now?

asmega: you are correct, removing

kosiakkatrina (approved):

jamesrose (approved):

PR #1478, opened 2023-03-28
Resolve cloud foundry key issue

asmega:
# Context

- Cannot install cloud foundry CLI
- Because https://packages.cloudfoundry.org/debian/cli.cloudfoundry.org.key seems to be blocking curl/wget requests

# Changes

- Cache the above key in version control and use this instead

# Review notes

- Note sure if they is a way to have global environment variables for all workflows which would reduce the duplication
- But this will work for now
- Please feedback if you have a more elegant solution to resolve this issue


asmega:
closing in favour of https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/1480

PR #1482, opened 2023-03-28
CLDC-1893 Support username as email for bulk upload

asmega:
# Context

- https://digital.dclg.gov.uk/jira/browse/CLDC-1893
- Support `created_by` for bulk upload

# Changes

- Given an email for username in bulk upload assign them as the `created_by`
- If no user is found we user the uploader
- Validations for error reporting
- Also prevent log creation if there is bad data

### DIFF
--- a/app/services/bulk_upload/lettings/year2022/row_parser.rb
+++ b/app/services/bulk_upload/lettings/year2022/row_parser.rb
@@ -325,6 +325,9 @@ class BulkUpload::Lettings::Year2022::RowParser
   validate :validate_location_exists
   validate :validate_location_data_given
 
+  validate :validate_created_by_exists
+  validate :validate_created_by_related
+
   def self.question_for_field(field)
     QUESTIONS[field]
   end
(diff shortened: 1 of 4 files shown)


app/services/bulk_upload/lettings/year2022/row_parser.rb
```diff
@@ -382,6 +385,28 @@ def property_ref
 
 private
 
+  def validate_created_by_exists
+    return if field_112.blank?
+
+    unless created_by
+      block_log_creation!
+      errors.add(:field_112, "User with the specified email could not be found")
```
jamesrose: Do we want to block log creation if the user can't be found? My impression was that we would just import them silently and assign attribution to the author?

asmega: just removed that now

ghost (approved):

jamesrose (approved):

PR #1493, opened 2023-03-30
CLDC-1975 Format money values with 2 decimals

ghost:
- render error messages with 2 decimals
- render input fields with 2 decimals

### DIFF
--- a/app/helpers/money_formatting_helper.rb
+++ b/app/helpers/money_formatting_helper.rb
@@ -1,0 +1,23 @@
+module MoneyFormattingHelper
+  include ActionView::Helpers::NumberHelper
+
+  def format_money_input(log:, question:)
+    value = log[question.id]
+
+    return unless value
+    return value unless question.prefix == "£"
+
+    number_with_precision(
+      value,
+      precision: 2,
+    )
+  end
+
+  def format_as_currency(num_string)
+    number_to_currency(
+      num_string,
+      unit: "£",
+      precision: 2,
+    )
+  end
+end
--- a/app/models/lettings_log.rb
+++ b/app/models/lettings_log.rb
@@ -149,7 +149,7 @@ class LettingsLog < Log
   def weekly_to_value_per_period(field_value)
     num_of_weeks = NUM_OF_WEEKS_FROM_PERIOD[period]
 
-    ((field_value * 52) / num_of_weeks).round(2)
+    sprintf("%.2f", (field_value * 52) / num_of_weeks)
   end
 
   def applicable_income_range
@@ -650,7 +650,7 @@ private
     num_of_weeks = NUM_OF_WEEKS_FROM_PERIOD[period]
     return "" unless value && num_of_weeks
 
-    (value * 52 / num_of_weeks).round(2)
+    sprintf("%.2f", (value * 52 / num_of_weeks))
   end
 
   def fully_wheelchair_accessible?
--- a/app/models/lettings_log.rb
+++ b/app/models/lettings_log.rb
@@ -20,6 +20,7 @@ class LettingsLog < Log
   include DerivedVariables::LettingsLogVariables
   include Validations::DateValidations
   include Validations::FinancialValidations
+  include MoneyFormattingHelper
 
   has_paper_trail
 
@@ -149,7 +150,7 @@ class LettingsLog < Log
   def weekly_to_value_per_period(field_value)
     num_of_weeks = NUM_OF_WEEKS_FROM_PERIOD[period]
 
-    sprintf("%.2f", (field_value * 52) / num_of_weeks)
+    format_as_currency((field_value * 52) / num_of_weeks)
   end
 
   def applicable_income_range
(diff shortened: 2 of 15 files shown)


app/models/lettings_log.rb
```diff
@@ -149,7 +149,7 @@ def weekly_value(field_value)
   def weekly_to_value_per_period(field_value)
     num_of_weeks = NUM_OF_WEEKS_FROM_PERIOD[period]
 
-    ((field_value * 52) / num_of_weeks).round(2)
+    sprintf("%.2f", (field_value * 52) / num_of_weeks)
```
asmega: do you really want to be converting these to strings at this stage?

ghost: I don't see why not, the method is only used in error messages https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/blob/345898e599bf631aa4fca21f7a8a224b9e1720e2/app/models/validations/financial_validations.rb#L121

spec/models/validations/financial_validations_spec.rb
```diff
@@ -919,39 +919,39 @@
           record.chcharge = 1001
           financial_validator.validate_care_home_charges(record)
           expect(record.errors["chcharge"])
-            .to include(match I18n.t("validations.financial.carehome.out_of_range", min_chcharge: 10, period: "weekly for 52 weeks", max_chcharge: 1000))
+            .to include("Household rent and other charges must be between 10.00 and 1000.00 if paying weekly for 52 weeks")
```
natdeanlewissoftwire: I think all these messages should have £ signs prefixing all the financial values

ghost: good point!

natdeanlewissoftwire (changes requested):
I know my comment's not strictly to do with your PR but seems worth fixing while we're here, rest of it looks good!

app/models/lettings_log.rb
arfacamble: is this just taking a number and formatting it right?
We have things that do this already I believe
eg `log:format_as_currency` (if using this you could change its parameter from num_string to num)
if that doesn't work for you because you don't want the '£' then you could use rails `number_with_delimeter` or `number_with_precision`

app/helpers/money_input_helper.rb
```diff
@@ -0,0 +1,10 @@
+module MoneyInputHelper
+  def format_money_input(log:, question:)
+    value = log[question.id]
+
+    return unless value
+    return value unless question.prefix == "£"
+
+    sprintf("%.2f", value)
```
arfacamble:
```suggestion
    number_with_precision(value, precision: 2, delimiter: ',')
```

natdeanlewissoftwire: maybe even https://apidock.com/rails/ActionView/Helpers/NumberHelper/number_to_currency

arfacamble: that's the method being used in `log:format_as_currency` but I dunnae think he wants the £ prefix here

ghost: nice, thank you both. Didn't know about these helpers. I've updated the PR to use them @natdeanlewissoftwire @arfacamble

app/helpers/money_formatting_helper.rb
```diff
@@ -0,0 +1,21 @@
...
+    return unless value
+    return value unless question.prefix == "£"
+
+    number_with_precision(
+      value,
+      precision: 2,
+    )
+  end
+
+  def format_as_currency(num_string)
```
arfacamble, line 16: a similar method already exists on `Log` - and precision: 2 is already the default value for that parameter.

app/helpers/money_input_helper.rb
natdeanlewissoftwire: yep, think the unit is optional if you give it ""

app/models/log.rb
```diff
@@ -203,8 +203,4 @@ def reset_location(is_inferred, la_key, is_inferred_key, postcode_key, is_la_kno
     self[is_inferred_key] = false
     self[postcode_key] = nil
   end
-
-  def format_as_currency(num_string)
-    ActionController::Base.helpers.number_to_currency(num_string, unit: "£")
-  end
```
arfacamble, line 207: Oh I see you're removing this. I believe it's being called a couple of times from in the sales log?

natdeanlewissoftwire (approved):
Nice one!

app/helpers/money_formatting_helper.rb
```diff
@@ -0,0 +1,21 @@
...
+    ActionController::Base.helpers.number_to_currency(
```
arfacamble: if this is in a helpers folder, does it definitely need the ActionController::Base.helpers prefix?

app/models/sales_log.rb
```diff
@@ -223,7 +224,7 @@ def postcode_full=(postcode)
   def expected_shared_ownership_deposit_value
     return unless value && equity
 
-    (value * equity / 100).round(2)
+    sprintf("%.2f", (value * equity / 100))
```
arfacamble: wouldn't this also just like to use something from the MoneyForattingHelper rather than sprintf?

app/models/log.rb
ghost: moved to the helper

app/helpers/money_formatting_helper.rb
ghost: was moved from log to here

PR #1497, opened 2023-03-30

asmega:
# Context

- We now let users decide for bulk upload when there are errors whether to re-upload the csv again or create and fix inline

# Changes

- New mini journey if we can't create logs and there are errors let user decide how to fix these errors
- For performance we create the logs ahead of time where we can. these logs are not visible
- to handle this `default_scope` added to `LettingsLog`

# Screenshots

## Landing page from email

## Confirm action


app/models/lettings_log.rb
```diff
@@ -21,6 +21,8 @@ class LettingsLog < Log
   include Validations::DateValidations
   include Validations::FinancialValidations
 
+  default_scope { where(visible: true) }
```
jamesrose, line 24: This probably worth a discussion! 😬 

I'm generally not of default_scope because it impacts initialization amongst other funky things. e.g. 

```ruby
LettingsLog.new.visible # => true

scheme = Scheme.last
scheme.lettings_logs.destroy_all # Only deletes visible logs
```

Thoughts?

jamesrose: Did you consider adding another state to `Log#status` that is before `not_started` ?

asmega: > I'm generally not of default_scope because it impacts initialization amongst other funky things

Yeah i already considered this and not also not a fan of `default_scope`. the alternative is to add explicitly add the scope `visible` to pretty much every existing AR thing that involves lettings logs. Thats the only alternative i can think of.

asmega: > Did you consider adding another state to `Log#status` that is before `not_started`?

I can have a look at this but would expect it would be a similar issue?

jamesrose: I don't think it will be too difficult. Searching for `lettings_logs.*` and `LettingsLog.` doesn't yield that many results. I suggested using `status` because it's more likely you won't have to make as any many changes, because some places scope out `in_progress` or `completed` specifically.

jamesrose: But yeah, I think `default_scope` is more trouble than it's worth.

asmega:
Closing this in favour of https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/1503 which has the `default_scope` removed and using a new `status` of `pending` instead

PR #1503, opened 2023-03-31

asmega:
# Context

- We now let users decide for bulk upload when there are errors whether to re-upload the CSV again or create and fix inline

# Changes

- New mini journey if we can't create logs and there are errors let user decide how to fix these errors
- For performance we create the logs ahead of time where we can. these logs are not visible
- to handle this there is a new `status` of `pending`
- we now have to explicitly call `visible` to hide any `pending` logs from views and other use cases 

# Screenshots

## Landing page from email

## Confirm action


jamesrose:
@asmega I found these other instances that we might want to scope. Could you have a look?

- `FormController#find_resource`
- `FormController#find_resource_by_named_id`
- `LocationsController#deactivate_confirm`
- `LocationsController#reset_location_and_scheme_for_logs!`
- `SchemesController#deactivate_confirm`
- `SchemesController#reset_location_and_scheme_for_logs!`
- `SalesLogsController#index`
- `Organisation#completed_lettings_logs`
- `Organisation#not_completed_lettings_logs`
- `User#completed_lettings_logs`
- `User#not_completed_lettings_logs`

asmega:
> I found these other instances that we might want to scope. Could you have a look?

I've updated to handle these. Some were dead code and have pruned, others simple updates. For locations and schemes I'm not 100% what the behaviour should be and have made them so they only affect visible logs

jamesrose (approved):

PR #1506, opened 2023-04-03
Fix collection year shift issues

kosiakkatrina:
The options were 2023/24 and 2024/25 - possibly for testing reasons. This sets them back to 2022/23 and 2023/24
Sales options are correct
Reset routes after reinitialising the FormHandler to fix the tests, because the routes are dynamically created from form pages. Different years have different form pages and therefore different routes. They need to be set depending on the forms we're using in specific tests.

xit out 2 tests that are/will be addressed separately.
Schemes test to be fixed as part of CLDC-1687 
Collection window close date to be fixed as part of CLDC-2240


jamesrose (approved):

app/models/scheme.rb
```diff
@@ -219,7 +219,11 @@ def validate_owning_organisation
   end
 
   def available_from
```
arfacamble, line 221: there might be a more succinct way of writing this, see https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/1396/files#diff-21a9bef9fe959a51cb7dc2afeb12b6ebafa4b89436172a40997c48f5d44e7813

quite possibly not though since we're anchoring everything at created_at

PR #1507, opened 2023-04-03
CLDC-1852 add weekly rent period and reorder options

arfacamble:
https://digital.dclg.gov.uk/jira/browse/CLDC-1852

1. add in a new option for the period question and reorder existing answer options
2. ensure that the lettings log calculates values correctly that relate to the period

|image
|-

|image
|-

### DIFF
--- a/app/models/form/lettings/pages/rent_weekly.rb
+++ b/app/models/form/lettings/pages/rent_weekly.rb
@@ -4,30 +4,8 @@ class Form::Lettings::Pages::RentWeekly < ::Form::Page
     @id = "rent_weekly"
     @header = "Household rent and charges"
     @depends_on = [
-      { "period" => 1, "household_charge" => 0, "is_carehome" => 0 },
-      { "period" => 1, "household_charge" => nil, "is_carehome" => 0 },
-      { "period" => 5, "household_charge" => 0, "is_carehome" => 0 },
-      { "period" => 5, "household_charge" => nil, "is_carehome" => 0 },
-      { "period" => 6, "household_charge" => 0, "is_carehome" => 0 },
-      { "period" => 6, "household_charge" => nil, "is_carehome" => 0 },
-      { "period" => 7, "household_charge" => 0, "is_carehome" => 0 },
-      { "period" => 7, "household_charge" => nil, "is_carehome" => 0 },
-      { "period" => 8, "household_charge" => 0, "is_carehome" => 0 },
-      { "period" => 8, "household_charge" => nil, "is_carehome" => 0 },
-      { "period" => 9, "household_charge" => 0, "is_carehome" => 0 },
-      { "period" => 9, "household_charge" => nil, "is_carehome" => 0 },
-      { "period" => 1, "household_charge" => 0, "is_carehome" => nil },
-      { "period" => 1, "household_charge" => nil, "is_carehome" => nil },
-      { "period" => 5, "household_charge" => 0, "is_carehome" => nil },
-      { "period" => 5, "household_charge" => nil, "is_carehome" => nil },
-      { "period" => 6, "household_charge" => 0, "is_carehome" => nil },
-      { "period" => 6, "household_charge" => nil, "is_carehome" => nil },
-      { "period" => 7, "household_charge" => 0, "is_carehome" => nil },
-      { "period" => 7, "household_charge" => nil, "is_carehome" => nil },
-      { "period" => 8, "household_charge" => 0, "is_carehome" => nil },
-      { "period" => 8, "household_charge" => nil, "is_carehome" => nil },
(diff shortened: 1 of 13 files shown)


SamSeed-Softwire (commented):
The stuff you've written all looks fine.

Are there other places in the codebase however which need to reference the new answer option? E.g. the following look like they might need it:
- `app/models/form/lettings/pages/care_home_weekly.rb`
- `app/models/form/lettings/pages/rent_weekly.rb`
- `app/models/form/lettings/questions/tshortfall.rb`
- `config/forms/2022_2023.json`, assuming it's relevant to 22/23

Is it worth adding some tests for the new answer option in `spec/models/lettings_log_spec.rb` as well?

arfacamble:
> The stuff you've written all looks fine.
> 
> Are there other places in the codebase however which need to reference the new answer option? E.g. the following look like they might need it:
> 
> * `app/models/form/lettings/pages/care_home_weekly.rb`
> * `app/models/form/lettings/pages/rent_weekly.rb`
> * `app/models/form/lettings/questions/tshortfall.rb`
> * `config/forms/2022_2023.json`, assuming it's relevant to 22/23
> 
> Is it worth adding some tests for the new answer option in `spec/models/lettings_log_spec.rb` as well?

great catch

app/models/lettings_log.rb
```diff
@@ -379,6 +379,10 @@ def receives_uc_with_housing_element_excl_housing_benefit?
     hb == 6
   end
 
+  def receives_any_housing_benefit?
```
SamSeed-Softwire: Could this new method also be used to refactor `receives_housing_related_benefits?`? Or did you even need to create this method and could you have just used `receives_housing_related_benefits?` instead?

spec/fixtures/forms/2021_2022.json
SamSeed-Softwire, line 1: Is the new answer option necessary here, given this file is for 2021/22?

app/models/form/lettings/pages/rent_weekly.rb
```diff
@@ -4,30 +4,8 @@ def initialize(id, hsh, subsection)
...
-      { "period" => 5, "household_charge" => nil, "is_carehome" => nil },
-      { "period" => 6, "household_charge" => 0, "is_carehome" => nil },
-      { "period" => 6, "household_charge" => nil, "is_carehome" => nil },
-      { "period" => 7, "household_charge" => 0, "is_carehome" => nil },
-      { "period" => 7, "household_charge" => nil, "is_carehome" => nil },
-      { "period" => 8, "household_charge" => 0, "is_carehome" => nil },
-      { "period" => 8, "household_charge" => nil, "is_carehome" => nil },
-      { "period" => 9, "household_charge" => 0, "is_carehome" => nil },
-      { "period" => 9, "household_charge" => nil, "is_carehome" => nil },
+      { "rent_and_charges_paid_weekly?" => true, "household_charge" => 0, "is_carehome?" => false },
```
SamSeed-Softwire, line 7: There are various other places that `"is_carehome" => nil or 0` could be replaced with `"is_carehome?" => false`, for example in `app/models/form/lettings/pages/rent_4_weekly.rb`. I'd recommend either doing that as part of this ticket, or raising a new tech debt ticket.

spec/fixtures/forms/2021_2022.json
arfacamble: this is a test fixture. If I don't put it here I have to rewrite the whole lettings_log_spec and that would really stretch the scope of this ticket!

app/models/lettings_log.rb
arfacamble: just use that! good catch

app/models/form/lettings/pages/rent_weekly.rb
arfacamble: Not sure, I don't want to change half the conditions in rent_4_weekly, so then I change them all with some new methods, and then I have to search to for everywhere that household_charge is nil or 0 and everywhere period is 3 and then write new methods for the other conditions in those depends on, and so on and so on.
Where do I cut the scope of the ticket off?

arfacamble: ok I relented on this in a big way

spec/fixtures/forms/2021_2022.json
SamSeed-Softwire: Ah gotcha

app/models/form/lettings/pages/rent_weekly.rb
SamSeed-Softwire: > ok I relented on this in a big way

What do you mean?

arfacamble: I made the changes

SamSeed-Softwire (approved):
The time has come!

PR #1508, opened 2023-04-03
CLDC-1856 Add housing question flow validation

ghost:
https://digital.dclg.gov.uk/jira/browse/CLDC-1856

Trigger a validation when selecting an invalid combination of user needs (user said that there are housing needs but no user need was selected in follow up questions)

record.housingneeds == 1 && record.housingneeds_type == 3 && record.housingneeds_other&.zero?

Don't think we need to test all possible answers to these questions because they're too many and should be covered by other validation tests

### DIFF
--- a/app/models/validations/household_validations.rb
+++ b/app/models/validations/household_validations.rb
@@ -121,6 +121,14 @@ module Validations::HouseholdValidations
     end
   end
 
+  def validate_housing_needs(record)
+    if record.housingneeds == 1 && record.housingneeds_type == 3 && record.housingneeds_other&.zero?
+      record.errors.add :housingneeds, I18n.t("validations.household.housingneeds.invalid")
+      record.errors.add :housingneeds_type, I18n.t("validations.household.housingneeds.invalid")
+      record.errors.add :housingneeds_other, I18n.t("validations.household.housingneeds.invalid")
+    end
+  end
+
 private
 
   def household_no_illness?(record)
--- a/app/models/validations/household_validations.rb
+++ b/app/models/validations/household_validations.rb
@@ -121,7 +121,7 @@ module Validations::HouseholdValidations
     end
   end
 
-  def validate_housing_needs(record)
+  def validate_combination_of_housing_needs_responses(record)
     if record.housingneeds == 1 && record.housingneeds_type == 3 && record.housingneeds_other&.zero?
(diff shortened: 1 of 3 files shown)


asmega (approved):

spec/models/validations/household_validations_spec.rb
```diff
@@ -637,4 +637,20 @@
...
+
+      household_validator.validate_housing_needs(record)
+
+      error_message = ["If somebody in the household has disabled access needs, they must have the access needs listed, or other access needs"]
+
+      expect(record.errors["housingneeds"]).to eq(error_message)
+      expect(record.errors["housingneeds_type"]).to eq(error_message)
+      expect(record.errors["housingneeds_other"]).to eq(error_message)
+    end
+  end
```
arfacamble, line 655: can the name of the test be written in english not obscure codes, I don't know off the top of my head what housingneeds == 1 means in terms of what the user answered
and slight pickiness error_message should be a string and the expects can be `to include error_message`

app/models/validations/household_validations.rb
```diff
@@ -121,6 +121,14 @@ def validate_layear(record)
     end
   end
 
+  def validate_housing_needs(record)
+    if record.housingneeds == 1 && record.housingneeds_type == 3 && record.housingneeds_other&.zero?
+      record.errors.add :housingneeds, I18n.t("validations.household.housingneeds.invalid")
+      record.errors.add :housingneeds_type, I18n.t("validations.household.housingneeds.invalid")
+      record.errors.add :housingneeds_other, I18n.t("validations.household.housingneeds.invalid")
+    end
+  end
```
arfacamble, line 130: my preference also in these validations is to have methods on the log class that make these codes human readable.
I don't know what these codes represent without going and checking elsewhere

arfacamble (dismissed):
comments in line

spec/models/validations/household_validations_spec.rb
ghost: I personally find more confusing to read the english translation of those codes, I've amended the copy to include both

> slight pickiness error_message should be a string and the expects can be to include error_message

this isn't equivalent to what I have right now. I do want to check that there's a single error under that key and testing for inclusion won't cover that

PR #1511, opened 2023-04-04
CLDC-2211 Correctly save care home charges

kosiakkatrina:
When care home charges are given in bulk upload:
 - set is_carehome to yes
 - set care home charge to the given value

When care home charges are not given:
 - set is_carehome to no

### DIFF
--- a/spec/services/bulk_upload/lettings/year2022/row_parser_spec.rb
+++ b/spec/services/bulk_upload/lettings/year2022/row_parser_spec.rb
@@ -599,6 +599,28 @@ RSpec.describe BulkUpload::Lettings::Year2022::RowParser do
       end
     end
 
+    describe "#field_85" do
+      let(:bulk_upload) { create(:bulk_upload, :lettings, user:, needstype: "2") }
+
+      context "when care home charge is given" do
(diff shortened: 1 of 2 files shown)


arfacamble (approved):

spec/services/bulk_upload/lettings/year2022/row_parser_spec.rb
```diff
@@ -599,6 +599,28 @@
       end
     end
 
+    describe "#field_85" do
+      let(:bulk_upload) { create(:bulk_upload, :lettings, user:, needstype: "2") }
```
arfacamble, line 603: out of interest does create() implicitly call FactoryBot.create or is it something else?
why do we use one over the other in different test contexts?

kosiakkatrina: I think they're doing the same thing, I'm not a 100% sure but I think we can use the shorter syntax because of this in the spec_helper `config.include FactoryBot::Syntax::Methods`. I don't think there's a reason why we would use one over the other. The inconsistency probably happened due to different preferences of different people writing the tests

PR #1512, opened 2023-04-04
CLDC-2163 add validations on whether buyers are living in the property

arfacamble:
https://digital.dclg.gov.uk/jira/browse/CLDC-2163

the naming in this ticket is a little bit confusing in places since the fields being checked against each other are really close in meaning.
There is a question in the setup section 'buylivein' that asks "Will the buyers live in the property?" that is only asked under certain conditions and then there are the fields 'buy1livein' and 'buy2livein' that ask respectively whether buyer 1 and buyer 2 specifically will live in the property.

In most situations the values in the setup questions logically allow buy1livein and buy2livein to be derived.
Iff buylivein is yes and joint purchase is yes we only know that at least one buyer must live in so marking both as no triggers a validation.

### DIFF
--- a/app/models/form/sales/pages/buyer1_live_in_property.rb
+++ b/app/models/form/sales/pages/buyer1_live_in_property.rb
@@ -5,9 +5,21 @@ class Form::Sales::Pages::Buyer1LiveInProperty < ::Form::Page
     @depends_on = [
       {
         "buyer_has_seen_privacy_notice?" => true,
+        "outright_sale?" => false,
       },
       {
         "buyer_not_interviewed?" => true,
+        "outright_sale?" => false,
+      },
+      {
+        "buyer_has_seen_privacy_notice?" => true,
+        "joint_purchase?" => true,
+        "buyers_will_live_in?" => true,
+      },
+      {
+        "buyer_not_interviewed?" => true,
+        "joint_purchase?" => true,
+        "buyers_will_live_in?" => true,
       },
     ]
   end
--- a/app/models/form/sales/pages/buyer2_live_in_property.rb
+++ b/app/models/form/sales/pages/buyer2_live_in_property.rb
@@ -4,12 +4,24 @@ class Form::Sales::Pages::Buyer2LiveInProperty < ::Form::Page
     @id = "buyer_2_live_in_property"
     @depends_on = [
       {
-        "joint_purchase?" => true,
         "buyer_has_seen_privacy_notice?" => true,
+        "outright_sale?" => false,
+        "joint_purchase?" => true,
       },
       {
-        "joint_purchase?" => true,
         "buyer_not_interviewed?" => true,
+        "outright_sale?" => false,
+        "joint_purchase?" => true,
+      },
+      {
+        "buyer_has_seen_privacy_notice?" => true,
+        "joint_purchase?" => true,
+        "buyers_will_live_in?" => true,
+      },
+      {
+        "buyer_not_interviewed?" => true,
+        "joint_purchase?" => true,
+        "buyers_will_live_in?" => true,
       },
     ]
   end
--- a/app/models/lettings_log.rb
+++ b/app/models/lettings_log.rb
@@ -531,7 +531,6 @@ class LettingsLog < Log
 
 private
 
-  # show in diff
   def reset_derived_questions
     dependent_questions = { waityear: [{ key: :renewal, value: 0 }],
                             referral: [{ key: :renewal, value: 0 }],
@@ -549,7 +548,6 @@ private
       end
     end
   end
-  # show in diff
 
   def reset_invalid_unresolved_log_fields!
     return unless unresolved?
--- a/app/models/lettings_log.rb
+++ b/app/models/lettings_log.rb
@@ -531,6 +531,7 @@ class LettingsLog < Log
 
 private
 
+  # show in diff
   def reset_derived_questions
     dependent_questions = { waityear: [{ key: :renewal, value: 0 }],
                             referral: [{ key: :renewal, value: 0 }],
@@ -548,6 +549,7 @@ private
       end
     end
   end
+  # show in diff
 
   def reset_invalid_unresolved_log_fields!
     return unless unresolved?
--- a/app/models/sales_log.rb
+++ b/app/models/sales_log.rb
@@ -161,8 +161,24 @@ class SalesLog < Log
     inc1mort == 1
   end
 
+  def buyers_will_live_in?
+    buylivein == 1
+  end
+
+  def buyers_will_not_live_in?
+    buylivein == 2
+  end
+
   def buyer_two_will_live_in_property?
     buy2livein == 1
+  end
+
+  def buyer_two_will_not_live_in_property?
+    buy2livein == 2
+  end
+
+  def buyer_one_will_not_live_in_property?
+    buy1livein == 2
   end
 
   def buyer_two_not_already_living_in_property?
--- a/app/models/sales_log.rb
+++ b/app/models/sales_log.rb
@@ -403,8 +403,8 @@ private
       any_parent_attributes_changed = dependency[:parent_conditions].any? { |parent_condition| send("#{parent_condition[:attribute]}_changed?") }
       next unless any_parent_attributes_changed
 
-      were_in_derived_state = dependency[:parent_conditions].all? { |parent_condition| send("#{parent_condition[:attribute]}_was") == parent_condition[:value] }
-      next unless were_in_derived_state
+      previously_in_derived_state = dependency[:parent_conditions].all? { |parent_condition| send("#{parent_condition[:attribute]}_was") == parent_condition[:value] }
+      next unless previously_in_derived_state
 
       dependency[:derived_attributes].each do |derived_attribute|
         Rails.logger.debug("Cleared derived #{derived_attribute} value")
--- a/app/models/sales_log.rb
+++ b/app/models/sales_log.rb
@@ -385,25 +385,25 @@ private
   def reset_derived_questions!
     dependencies = [
       {
-        parent_conditions: [
-          { attribute: :buylivein, value: 2 },
-        ],
+        conditions: {
+          buylivein: 2,
+        },
         derived_attributes: %i[buy1livein buy2livein],
       },
       {
-        parent_conditions: [
-          { attribute: :buylivein, value: 1 },
-          { attribute: :jointpur, value: 1 },
-        ],
+        conditions: {
+          buylivein: 1,
+          jointpur: 1,
+        },
         derived_attributes: [:buy1livein],
       },
     ]
 
     dependencies.each do |dependency|
-      any_parent_attributes_changed = dependency[:parent_conditions].any? { |parent_condition| send("#{parent_condition[:attribute]}_changed?") }
-      next unless any_parent_attributes_changed
+      any_primary_attributes_changed = dependency[:conditions].any? { |attribute, _value| send("#{attribute}_changed?") }
+      next unless any_primary_attributes_changed
 
-      previously_in_derived_state = dependency[:parent_conditions].all? { |parent_condition| send("#{parent_condition[:attribute]}_was") == parent_condition[:value] }
+      previously_in_derived_state = dependency[:conditions].all? { |attribute, value| send("#{attribute}_was") == value }
       next unless previously_in_derived_state
 
       dependency[:derived_attributes].each do |derived_attribute|
--- a/app/models/sales_log.rb
+++ b/app/models/sales_log.rb
@@ -368,12 +368,48 @@ class SalesLog < Log
 
   def ownership_scheme
     case ownershipsch
-    when 1
-      "shared ownership"
-    when 2
-      "discounted ownership"
-    when 3
-      "outright sale"
+    when 1 then "shared ownership"
+    when 2 then "discounted ownership"
+    when 3 then "outright sale"
+    end
+  end
+
+private
+
+  def reset_invalidated_dependent_fields!
+    super
+
+    reset_derived_questions!
+  end
+
+  def reset_derived_questions!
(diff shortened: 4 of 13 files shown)


app/models/validations/sales/household_validations.rb
```diff
@@ -20,6 +20,27 @@ def validate_previous_postcode(record)
...
+        record.errors.add :buylivein, I18n.t("validations.household.buyers_will_live_in_property.buyers_live_but_no_buyers_live.joint.buyers")
+        record.errors.add :buy1livein, I18n.t("validations.household.buyers_will_live_in_property.buyers_live_but_no_buyers_live.joint.each_buyer")
+        record.errors.add :buy2livein, I18n.t("validations.household.buyers_will_live_in_property.buyers_live_but_no_buyers_live.joint.each_buyer")
+      elsif record.not_joint_purchase? && record.buyer_one_will_not_live_in_property?
+        record.errors.add :buylivein, I18n.t("validations.household.buyers_will_live_in_property.buyers_live_but_no_buyers_live.not_joint.buyers")
+        record.errors.add :jointpur, I18n.t("validations.household.buyers_will_live_in_property.buyers_live_but_no_buyers_live.not_joint.jointpur")
+        record.errors.add :buy1livein, I18n.t("validations.household.buyers_will_live_in_property.buyers_live_but_no_buyers_live.not_joint.each_buyer")
+      end
+    elsif record.buyer_one_will_live_in_property? || record.buyer_two_will_live_in_property?
```
arfacamble, line 37: I had 37-41 as an else with a separate if inside it which I think read a little more clearly as the branch for when buyers_live_in? is false, but the linter made me collapse the nested if

config/locales/en.yml
arfacamble, line 1: this is the place where I think the naming is particularly gross, but I really struggled to come up with something better 😬

jamesrose (approved):

app/models/sales_log.rb
arfacamble, line 1: It was possible to get the form into a state where if the answers are wrong it is not possible to enter the correct answers since there is a circular validation.
E.G.: If the user says buyers will not live in the property in the setup and also in household characteristics, but this is not true and they need to change it, they cannot change either without already changing the other (a validation triggers on either one due to the other), so they can't change anything to move the form to the truth state.
This solves that by clearing the inconsistent values.
It is important that the user only clears values later in the form if they try to create an invalid state by changing the answers to a question in the setup question. Questions in setup are never cleared due to changes later in the form, this would be dangerous since it could end up clearing large sections of the rest of the form as well

natdeanlewissoftwire: I agree that your solution makes sense within the scope of the ticket, but it seems a bit band-aidy to me rather than solving the actual issue. Imo users shouldn't be able to enter the same data in two places - can't we just infer some of the data to avoid this situation? if a user says buyers won't live in the property in setup I would argue we should just not display the buy1livein and buy2livein questions at all and/or infer them to be false. I think just doing that would be enough to break this circular dependency, wdyt?

arfacamble: the complication with that is that if they say that buyers do live in the property, we cannot infer from that that both buyers do live in the property

app/models/lettings_log.rb
```diff
@@ -521,6 +521,7 @@ def is_carehome?
 
 private
 
+  # show in diff
```
jamesrose: Could you remove these comments please?

app/models/sales_log.rb
```diff
@@ -348,12 +364,48 @@ def beds_for_la_sale_range
...
+      any_parent_attributes_changed = dependency[:parent_conditions].any? { |parent_condition| send("#{parent_condition[:attribute]}_changed?") }
+      next unless any_parent_attributes_changed
+
+      were_in_derived_state = dependency[:parent_conditions].all? { |parent_condition| send("#{parent_condition[:attribute]}_was") == parent_condition[:value] }
+      next unless were_in_derived_state
+
+      dependency[:derived_attributes].each do |derived_attribute|
+        Rails.logger.debug("Cleared derived #{derived_attribute} value")
+        send("#{derived_attribute}=", nil)
+      end
```
arfacamble: This is worth thinking about and discussing.
I have put comments next to the corresponding method in lettings log so you can see it above.
You will see that I have done it slightly differently, which in itself is undesirable, so for consistency, I propose that either
- we decide the old way was better and I change this to reflect that
- we decide this is better and we mark on the lettings log method to update it to reflect this model during the ticket to clear variables derived from renewal coming up in the next sprint.

this method:
defines the state in which variables are defined, and associated attributes whose values are derived.
for each of this dependencies, the value of the derived attribute is cleared iff the form has started in the state where the derivation occurs && one of the attributes on which the derived attribute depends has changed.
eg the first dependency above says if "Will the buyers live in the property?" is answered "No" (buylivein: 2), then the values for buy1livein and buy2livein are derived, so if we are transitioning away from that parent state, the child attributes should be cleared.

the old method:
defines the state the form would be moving to in order for the derived variable to be cleared. It does not take advantage of the `attribute_was` methods to check if we are transitioning _from_ a state where the derivation would have taken place.
I have put a more concrete example next to the above method or why this is less good.

app/models/form/sales/pages/buyer2_live_in_property.rb
```diff
@@ -4,12 +4,24 @@ def initialize(id, hsh, subsection)
     @id = "buyer_2_live_in_property"
     @depends_on = [
       {
-        "joint_purchase?" => true,
         "privacynotice" => 1,
+        "outright_sale?" => false,
```
natdeanlewissoftwire, line 8: is making this conditional on not-outright sale relevant?
don't we just care about joint purchase?

app/models/sales_log.rb
```diff
@@ -348,12 +364,48 @@ def beds_for_la_sale_range
 
   def ownership_scheme
     case ownershipsch
-    when 1
-      "shared ownership"
-    when 2
-      "discounted ownership"
-    when 3
-      "outright sale"
+    when 1 then "shared ownership"
```
natdeanlewissoftwire, line 370: I like this

app/models/sales_log.rb
```diff
@@ -348,12 +364,48 @@ def beds_for_la_sale_range
...
+
+private
+
+  def reset_invalidated_dependent_fields!
+    super
+
+    reset_derived_questions!
+  end
+
+  def reset_derived_questions!
```
natdeanlewissoftwire: any reason this is a bang method compared to lettings_log's not being one?

app/models/sales_log.rb
```diff
@@ -348,12 +364,48 @@ def beds_for_la_sale_range
...
+        ],
+        derived_attributes: [:buy1livein],
+      },
+    ]
+
+    dependencies.each do |dependency|
+      any_parent_attributes_changed = dependency[:parent_conditions].any? { |parent_condition| send("#{parent_condition[:attribute]}_changed?") }
+      next unless any_parent_attributes_changed
+
+      were_in_derived_state = dependency[:parent_conditions].all? { |parent_condition| send("#{parent_condition[:attribute]}_was") == parent_condition[:value] }
```
natdeanlewissoftwire: I would prefer this to be called `previously_in_derived_state` or something even more descriptive but I accept that's subjective

app/models/validations/sales/household_validations.rb
```diff
@@ -20,6 +20,18 @@ def validate_previous_postcode(record)
     end
   end
 
+  def validate_buyers_living_in_property(record)
+    return unless record.form.start_date.year >= 2023
+
+    if record.buyers_will_live_in? &&
+        record.joint_purchase? &&
```
natdeanlewissoftwire: isn't this redundant if we're also testing buyer one and buyer 2 after?

app/models/lettings_log.rb
```diff
@@ -538,6 +539,7 @@ def reset_derived_questions
       end
     end
   end
+  # show in diff
```
arfacamble: the first entry in dependent questions above says that:
if we change "Is this letting a renewal?" to "No" (renewal == 0) then we can assume that it was previously "Yes" (1) and clear the field that would have been derived from that state.
So far, this way works, however if we are ever to make derivations from an answer to a radio question with more than two answers, we will need to know what state we are moving away from, rather than moving to in order to know whether we can clear the dependent fields.
eg if we derive a value from something like relat3 being "C" for child in the future, then in order to clear the derived value, we need to know in future if `relat3_was == "C"` rather than `relat3 == "P"/"X"/etc`

config/locales/en.yml
```diff
@@ -437,6 +437,8 @@ en:
         no_choices: "You cannot answer this question as you told us nobody in the household has a physical or mental health condition (or other illness) expected to last 12 months or more"
       postcode:
         discounted_ownership: "Last settled accommodation and discounted ownership property postcodes must match"
+      buyers_will_live_in_property:
+        buyers_live_but_no_buyers_live: "You have already told us that the buyers will live in the property. Either buyer 1 or buyer 2 must live in the property"
```
natdeanlewissoftwire: can we name this something less confusing?

app/models/form/sales/pages/buyer1_live_in_property.rb
arfacamble, line 1: the depends_on for the following two pages have become quite complicated, I wonder at what point we start adding comments to the depends_on field!?!?
these questions show if
(privacy notice OR buyer not interviewed) AND (not outright sale OR ( joint purchase AND buyers will live in ) )

app/models/sales_log.rb
natdeanlewissoftwire: Makes sense to me, I think your justification for the new method is sound. To be clear, are there definite examples where this old/new method distinction will matter?

app/models/lettings_log.rb
arfacamble: yes I definitely will, they're just so this method is visible in the diff as I wanted to discuss it as part of the review

app/models/form/sales/pages/buyer2_live_in_property.rb
arfacamble: if it is not an outright sale, it must be a joint purchase (and buylivein will be nil because routing)

if it is an outright sale, buyers_will_live_in must be true (buylivein) and joint purchase true for this page to show

app/models/sales_log.rb
arfacamble: I believe it should be a bang method because it is !dangerous! ie it makes changes to the model.
I would suggest that the lettings log method also be updated in this way.

app/models/sales_log.rb
arfacamble: I approve that rename

app/models/sales_log.rb
arfacamble: there are not currently situations where this improvement is necessary, however since I was writing this method new I thought it made sense to write it as well as possible.
I think this way is truer to the reality which does give it better future flexibility, but also for me makes it slightly more readable since it better reflects the conditions under which we want to clear things.

app/models/validations/sales/household_validations.rb
arfacamble: Yes I think this is redundant given routing. good spot.

config/locales/en.yml
arfacamble: suggestion? I found some of the naming awkward in this ticket since the things that the variables concerned represent are almost exactly the same thing.

app/models/validations/sales/household_validations.rb
arfacamble: I think that also makes this condition short enough to go on one line which is an added bonus

app/models/sales_log.rb
arfacamble: @jamesrose 
if Nat is happy with this way are you happy that I put a to-do comment on the corresponding method on lettings_log that it should be updated in line with this method as part of ticket 2288 and make a note of this also on the JIRA ticket for 2288?
I ask because this was something we did occasionally on a previous project but I've not seen it on this one.

config/locales/en.yml
natdeanlewissoftwire: how about `inconsistent_buyers_live_in_answers`?

app/models/sales_log.rb
natdeanlewissoftwire: go on then

app/models/form/sales/pages/buyer2_live_in_property.rb
natdeanlewissoftwire: coolio

natdeanlewissoftwire (approved):
pending last few conversations

app/models/sales_log.rb
```diff
@@ -348,12 +364,48 @@ def beds_for_la_sale_range
...
+        parent_conditions: [
+          { attribute: :buylivein, value: 1 },
+          { attribute: :jointpur, value: 1 },
+        ],
+        derived_attributes: [:buy1livein],
+      },
+    ]
+
+    dependencies.each do |dependency|
+      any_parent_attributes_changed = dependency[:parent_conditions].any? { |parent_condition| send("#{parent_condition[:attribute]}_changed?") }
```
kosiakkatrina:
```suggestion
      any_parent_attributes_changed = dependency[:parent_conditions].any? { |attribute, _value| send("#{attribute}_changed?") }
```
Would that work?

kosiakkatrina: Also, maybe just conditions instead of parent_conditions?

arfacamble: conditions I like.
but conditions is an array, so it won't take a block with two parameters.
But you do give me the idea that actually it could be simplified to
```
conditions: {
  buylivein: 2
}
```
then yours will work and 402 will {need/benefit from} simplifying also

arfacamble: changes made. great shout

kosiakkatrina (approved):

PR #1513, opened 2023-04-04
CLDC-2222 Set housingneeds as yes unless field_59 is given

kosiakkatrina:
If `fields_59 - Disabled access needs g) No disabled access needs` is not set assume that the household has housing needs and set housingneeds to Yes


asmega:
because each year is different do you need to also implement the same thing for the new collection year ie `app/services/bulk_upload/lettings/year2023/row_parser.rb`

asmega (approved):

PR #1515, opened 2023-04-04
CLDC-1985 Fix error message persistence bug

natdeanlewissoftwire:
Bug described in the ticket: https://digital.dclg.gov.uk/jira/browse/CLDC-1985

After much digging (this bug was being caused in certain browsers only in dev envs), the issue was js back links that were using cached versions of the page. Browsers were also resistant to cached data being cleared or not set with meta tags etc., but the final solution i arrived at works pretty well - the rails back link generates a new link to the referrer to redirect to rather than navigating to a previous cached page which resulst in infinite looping, but generating previous pages explicitly works better. Have also removed tautologous occurences of "Back" text for govuk back links elsewhere.


app/models/form.rb
```diff
@@ -88,6 +88,31 @@ def next_page_redirect_path(page, log, current_user)
     end
   end
 
+  def previous_page(page, log, current_user)
+    page_ids = subsection_for_page(page).pages.map(&:id)
```
arfacamble, line 92: this seems weird. I haven't seen this subsection_for_page method before but how is it doing anything different to `page.subsection`?

app/models/form.rb
```diff
@@ -67,7 +67,7 @@ def next_page(page, log, current_user)
     page_ids = subsection_for_page(page).pages.map(&:id)
     page_index = page_ids.index(page.id)
```
arfacamble, line 68: it seems like these lines don't need to be here if you've moved the logic they are for into the other method?

app/models/form.rb
```diff
@@ -88,6 +88,31 @@ def next_page_redirect_path(page, log, current_user)
...
+    page_ids = subsection_for_page(page).pages.map(&:id)
+    page_index = page_ids.index(page.id)
+    page_id = if page_index.zero?
+                :check_answers
+              else
+                page_ids[page_index - 1]
+              end
+    prvs_page = get_page(page_id)
+
+    return :check_answers if prvs_page.nil?
```
arfacamble:
```suggestion
    return :check_answers if page_index.zero?
    page_id = page_ids[page_index - 1]
    prvs_page = get_page(page_id)
```

app/models/form.rb
```diff
@@ -88,6 +88,31 @@ def next_page_redirect_path(page, log, current_user)
...
+    return prvs_page.id if prvs_page.routed_to?(log, current_user)
+
+    previous_page(prvs_page, log, current_user)
```
arfacamble: I think at the end of a method like here it's slightly more readable to have a standard if else, since it parses for me more like there are two fairly likely outcomes and we do one or the other:
"if the previous page is routed to, then return that, otherwise keep going back"
rather than guard clause format which slightly implies to me that there's this edge case where the page is routed to where we want to return early, otherwise we do the normal thing of keep going back.
It's pretty similar and not complex control flow so I don't feel strongly about it!

app/models/form.rb
arfacamble: the only way I can see that these are not equivalent is if get page returns nil, but considering you have built this list of ids from the pages that exist just lines earlier, I don't think this is a real risk.

app/models/form.rb
```diff
@@ -88,6 +88,31 @@ def next_page_redirect_path(page, log, current_user)
...
 
+  def previous_page(page, log, current_user)
+    page_ids = subsection_for_page(page).pages.map(&:id)
+    page_index = page_ids.index(page.id)
+    page_id = if page_index.zero?
+                :check_answers
+              else
+                page_ids[page_index - 1]
+              end
+    prvs_page = get_page(page_id)
```
arfacamble: no need to abbreviate previous for me. No need to be letterist against vowels.

app/views/form/page.html.erb
```diff
@@ -1,7 +1,11 @@
 <% content_for :title, @page.header.presence || @page.questions.first.header.html_safe %>
 
 <% content_for :before_content do %>
-  <%= govuk_back_link(href: "javascript:history.back()") %>
+  <% if url_for(:back).include?("check_answers") %>
+    <%= govuk_back_link(href: :back) %>
+  <% else %>
+    <%= govuk_back_link(href: send(@log.form.previous_page_redirect_path(@page, @log, current_user), @log)) %>
```
arfacamble: I wonder if there is some clever way that we can use url_for() here
https://api.rubyonrails.org/v5.2/classes/ActionDispatch/Routing/UrlFor.html#method-i-url_for

arfacamble (changes requested):
half reviewed, need to pick up from page.html.erb

app/views/form/page.html.erb
arfacamble: I don't believe it is possible.

jamesrose: Two questions! Did you know that `:back` returns HTTP referrer or `javascript:history.back() ` if it's not set? Why do you need to use `send` here?

natdeanlewissoftwire: 1) yes! the js history back is what was causing the issue this bug ticket seeks to fix so the http referrer would be a solution, but:
2) `:back` gets us stuck in infinite loops between two pages if you go back more than once. where the http referrer is set

app/models/form.rb
natdeanlewissoftwire: yep this was to mirror nxt in the corresponding next method, I agree though I've changed added the vowels in both domains since this

natdeanlewissoftwire: (and made method names more descriptive when they actually get ids not pages)

app/models/form.rb
natdeanlewissoftwire: i disagree, this is for the `next-page` method, the one I've added is `previous_page`

app/models/form.rb
natdeanlewissoftwire: Good spot, looks like subsection_for_page is a redundant method now... Can you think of any cases where it wouldn't be? I can't so will delete if we've got no reason not to between us

arfacamble: only if the subsection is not passed into initialise correctly

app/models/form.rb
arfacamble: I mean lines 67/68
I can't see page_ids or page_index being used in the rest of next_page_id

arfacamble (approved):
approved pending the comments from James, I would be interesting in understanding that re http referrer js history etc. Not an area I am familiar with.

app/models/form.rb
natdeanlewissoftwire: they're used on line 72 unless I'm misunderstanding you?

app/models/form.rb
natdeanlewissoftwire: yep my thinking exactly

app/models/form.rb
arfacamble: they are. consider me blind

PR #1518, opened 2023-04-05
CLDC-2010 Add discounted ownership max income soft validations

natdeanlewissoftwire:
ticket: https://digital.dclg.gov.uk/jira/browse/CLDC-2010
image

note: as part of this ticket, I prepended  the occurences of la in `interruption_screen_question_ids` with uprn and postcode_full as they can sometimes infer la, causing the la question to be hidden in the interruption screen. We could also fix this by forcing interruption screen questions to display, but that seems like a new ticket to me rather than just updating these few interruption screens to be in line with the format of the others.


app/models/form/sales/pages/buyer1_income_max_value_check.rb
```diff
@@ -0,0 +1,27 @@
...
+          "i18n_template" => "income",
+        },
+      ],
+    }
+    @informative_text = {}
+  end
+
+  def questions
+    @questions ||= [
+      Form::Sales::Questions::Buyer1IncomeValueCheck.new(nil, nil, self),
```
natdeanlewissoftwire: My thinking is it's fine to share this for min and max as a buyer can't fail both soft checks simultaneously

app/models/validations/sales/soft_validations.rb
```diff
@@ -19,6 +19,24 @@ def income2_under_soft_min?
...
+  def income2_over_soft_max?
+    return unless income2 && la && discounted_ownership_sale?
+
+    income_over_soft_max?(income2)
+  end
+
+  def combined_income_over_soft_max?
+    return unless income1 && income2 && la && discounted_ownership_sale?
+
+    income_over_soft_max?(income1 + income2)
```
natdeanlewissoftwire, line 37: think it's clearer not to use combined_income here

app/models/form/sales/pages/buyer1_income_max_value_check.rb
```diff
@@ -0,0 +1,28 @@
...
+      "arguments" => [
+        {
+          "key" => "field_formatted_as_currency",
+          "arguments_for_key" => "income1",
+          "i18n_template" => "income",
+        },
+      ],
+    }
+    @informative_text = {}
+    @check_answers_card_number = check_answers_card_number
```
kosiakkatrina, line 20: Wouldn't this always be 1? and 2 on buyer 2?

app/models/sales_log.rb
```diff
@@ -350,4 +350,8 @@ def mortgage_deposit_and_grant_total
   def beds_for_la_sale_range
     beds.nil? ? nil : [beds, LaSaleRange::MAX_BEDS].min
   end
+
+  def combined_income
+    income1 + income2
```
kosiakkatrina: what if one of them is nil?

natdeanlewissoftwire: this is only for displaying the combined income in the soft validation interruption screen, and we only run the combined validation when both incomes are present

app/models/form/sales/pages/buyer1_income_max_value_check.rb
natdeanlewissoftwire: no - needs to be nil in some cases when the value check is displayed in other subsections

app/models/sales_log.rb
natdeanlewissoftwire: I'd rather not but I'll put an explicit nil check

arfacamble: I agree it seems sensible to catch potential errors in case someone else ends up using this in future without thinking too hard.

sorry about the half written comment earlier, I started editing it then forgot, the above is the only important bit

natdeanlewissoftwire: no problemo

kosiakkatrina (approved):

kosiakkatrina (approved):

PR #1521, opened 2023-04-05
CLDC-2023 Add livein soft validation

kosiakkatrina:
image


app/models/form/sales/pages/buyer_live_in_value_check.rb
```diff
@@ -0,0 +1,21 @@
+class Form::Sales::Pages::BuyerLiveInValueCheck < Form::Sales::Pages::Person
+  def initialize(id, hsh, subsection, person_index:)
+    super
+    @depends_on = [
+      {
+        "buyer#{person_index}_not_livein?" => true,
```
arfacamble: this is so readable I am sooo confident what this method means I'm not going to bother scrolling down. 🎉

app/models/form/sales/questions/buyer_live_in_value_check.rb
arfacamble, line 1: I might be misremembering how some of these things work, but doesn't this need a dynamic id in case there is more than one of these question instances, ie for buyers 1 and 2?

app/models/sales_log.rb
```diff
@@ -42,7 +42,7 @@ class SalesLog < Log
   }
   scope :filter_by_organisation, ->(org, _user = nil) { where(owning_organisation: org) }
 
-  OPTIONAL_FIELDS = %w[saledate_check purchid monthly_charges_value_check old_persons_shared_ownership_value_check othtype discounted_sale_value_check].freeze
+  OPTIONAL_FIELDS = %w[saledate_check purchid monthly_charges_value_check old_persons_shared_ownership_value_check othtype discounted_sale_value_check buyer_livein_value_check].freeze
```
arfacamble, line 46: are all soft validation interruption screens optional fields? If so is there a way of doing this dynamically since they all end in _value_check?
I'm guessing the answer to the first question is no, otherwise this list would be quite a bit longer!

app/models/sales_log.rb
```diff
@@ -349,4 +349,15 @@ def mortgage_deposit_and_grant_total
   def beds_for_la_sale_range
     beds.nil? ? nil : [beds, LaSaleRange::MAX_BEDS].min
   end
+
+  def ownership_scheme
+    case ownershipsch
+    when 1
+      "Shared owhership"
+    when 2
+      "Discounted owhership"
```
arfacamble: typos

app/models/validations/sales/soft_validations.rb
```diff
@@ -123,6 +123,19 @@ def discounted_ownership_value_invalid?
...
+    (discounted_ownership_sale? || shared_ownership_scheme?) && buy1livein == 2
+  end
+
+  def buyer2_not_livein?
+    return unless ownershipsch && buy2livein
+    return unless joint_purchase?
+
+    (discounted_ownership_sale? || shared_ownership_scheme?) && buy2livein == 2
+  end
+
```
arfacamble, line 138: https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/1512/files#diff-896f324e0f3c60b69ebafa248b8305404cea7cb90b44c1632c2f007b9973160f

this hasn't been PO'ed and merged yet, but if it is in time there will be methods to reuse here

app/models/validations/sales/soft_validations.rb
```diff
@@ -123,6 +123,19 @@ def discounted_ownership_value_invalid?
...
 
+  def buyer1_not_livein?
+    return unless ownershipsch && buy1livein
+
+    (discounted_ownership_sale? || shared_ownership_scheme?) && buy1livein == 2
+  end
+
+  def buyer2_not_livein?
+    return unless ownershipsch && buy2livein
+    return unless joint_purchase?
```
arfacamble, line 134: I don't think it matters, but I'm fairly confident that this line isn't necessary
since `if !joint_purchase?` then `buy2livein` is not routed to and therefore cleared.
perfectly fine to have a safety net though!

app/services/imports/sales_logs_import_service.rb
```diff
@@ -277,7 +278,8 @@ def fields_not_present_in_softwire_data
          hodate_check
          saledate_check
          student_not_child_value_check
-         discounted_sale_value_check]
+         discounted_sale_value_check
+         buyer_livein_value_check]
```
arfacamble, line 283: It's not relevant to this ticket directly but I'm not sure why softwire is in the name of any variables!

config/locales/en.yml
```diff
@@ -547,6 +547,8 @@ en:
       informative_text: "Your given mortgage, deposit and grant total is £%{mortgage_deposit_and_grant_total}"
     care_home_charges:
       title_text: "Care home charges should be provided if this is a care home accommodation"
+    buyer1_not_livein:
+      title_text: "You told us that buyer 1 will not live in the property. For %{ownership_scheme} types, the buyer usually lives in the property."
```
arfacamble, line 555: It seemed to me above that there can be interruption screens for buyer 2 as well? If that's so surely this message needs to have `".... buyer {buyer_num} will not ..."`

spec/models/form_spec.rb
```diff
@@ -247,9 +247,9 @@ def answer_local_authority(lettings_log)
       expect(form.sections[0].class).to eq(Form::Sales::Sections::Setup)
       expect(form.subsections.count).to eq(1)
       expect(form.subsections.first.id).to eq("setup")
-      expect(form.pages.count).to eq(16)
+      expect(form.pages.count).to eq(18)
       expect(form.pages.first.id).to eq("organisation")
-      expect(form.questions.count).to eq(17)
+      expect(form.questions.count).to eq(19)
```
arfacamble, line 252: I'm scared these tests are going to become like that hated FormHandler test that I eventually killed.

db/migrate/20230405140343_add_buyer_livin_value_check.rb
```diff
@@ -0,0 +1,5 @@
+class AddBuyerLivinValueCheck < ActiveRecord::Migration[7.0]
+  def change
+    add_column :sales_logs, :buyer_livein_value_check, :integer
```
arfacamble, line 3: similar to other comments I'm a little confused as to how this copes with buyer 1 and buyer 2, I would expect them to need two independent fields in case both are triggered and only one has been dealt with by the user.

spec/models/validations/sales/soft_validations_spec.rb
arfacamble, line 1: In this context the method names for the soft validation look a little odd, but I see looking in the soft validations file that we don't really have a naming convention that makes clear they are validation methods.
It just read here that we are calling a method to check whether buyer 1 is not going to live in the property, rather than a soft validation checking whether that result is unexpected.
I'm not sure I can think if a better name that isn't very long though! `buyer1_will_not_live_in_unexpectedly?`,  maybe it would be good to move towards have a similar naming convention to the hard validations and go for `check_buyer1_will_not_live_in?`

arfacamble (changes requested):
typo somewhere and most other comments are just questions/clarifications

app/models/form/sales/pages/buyer_live_in_value_check.rb
arfacamble: lol. on reflection, this method didn't actually do what I thought it was going to when I wrote this 😬

app/models/form/sales/questions/buyer_live_in_value_check.rb
kosiakkatrina: I thought it was? The page id? I think we're passing it through when constructing the page

app/models/sales_log.rb
kosiakkatrina: Not all, only the ones that appear in setup section, because if they weren't optional the log would clear every time the validation triggers as all setup fields need to be completed before the rest of the sections are unlocked

app/models/validations/sales/soft_validations.rb
kosiakkatrina: Yes!

app/services/imports/sales_logs_import_service.rb
kosiakkatrina: I think it's cause maybe we were getting the import data from a Softwire team at the time. I'm not sure, it is a bit misleading

config/locales/en.yml
kosiakkatrina: You're absolutely right!

spec/models/form_spec.rb
kosiakkatrina: Hah, hopefully not, cause it's just the setup section

db/migrate/20230405140343_add_buyer_livin_value_check.rb
kosiakkatrina: I think we do the same thing for all the validations for different people, it would work as expected, but if you've answered this check before as yes, it would have it preselected, so you'd have to either change or confirm again. It's maybe not the most expected behaviour but the check would still show up for both.
It avoids adding a column for each check, plus with soft validations redesign coming up, we will change the flow anyway

spec/models/validations/sales/soft_validations_spec.rb
kosiakkatrina: How does it sound now?

app/models/sales_log.rb
arfacamble: makes sense thanks

arfacamble: although I still wonder if it might not be nicer to amend the `setup_completed?` method rather than keep a static list here if all the pages we ignore end in _check

db/migrate/20230405140343_add_buyer_livin_value_check.rb
arfacamble: ok cool. I think I can imagine a case where it wouldn't make sense in the current setup but it's edge enough to probably not worry about.
1. trigger soft validation on buyer 1 with invalid answer
2. don't answer soft validation, just go back
3. go to question related to buyer 2 and input valid answer
4. see soft validation despite valid answer?

app/services/imports/sales_logs_import_service.rb
arfacamble: do we want to rename it? what is this list of attributes?
I'm guessing since it's a list of value checks then it's actually attributes not present in old core or dluhc data?

spec/models/validations/sales/soft_validations_spec.rb
arfacamble: LGTM

app/models/form/sales/questions/buyer_live_in_value_check.rb
arfacamble: very good. I thought as with the other comments that the questions would need different ids to refer to different attributes, I didn't appreciate we were reusing the same attribute

db/migrate/20230405140343_add_buyer_livin_value_check.rb
kosiakkatrina: Hm, I think it wouldn't route to that if the second buyer answer is valid. Cause it would be a different soft validation

arfacamble (approved):
I petition you a last time to rename the variable with 'softwire' inside it, but you can ignore me if you really don't want to

app/models/sales_log.rb
kosiakkatrina: Does the rule to end it in `_check` still exists?

kosiakkatrina:
> I petition you a last time to rename the variable with 'softwire' inside it, but you can ignore me if you really don't want to

@arfacamble There are a few other things that mention softwire, like `was_migrated_from_softwire?`, so I'll leave it for now, cause I don't know contexts for all of them

PR #1524, opened 2023-04-06
CLDC-2226 Update UPRN validations in bulk upload

SamSeed-Softwire:
## Context
- https://digital.dclg.gov.uk/jira/browse/CLDC-2226

## The changes
- If the UPRN question is not answered and either of address line 1 and town/city are unanswered, make sure that a bulk upload error is created for the UPRN question (not just the unanswered address field question).
- If any compulsory field is missing in a bulk upload, normally the error message in the error report just references the question's `check_answer_label`. Sometimes a question doesn't have a `check_answer_label` though - in which case we now display the question's `header` instead.
- Added appropriate tests for the above.
- Also made an improvement to the existing "UPRN must be 12 digits or less" test.

## Note
- @kosiakkatrina and I paired on trying to write a test for the setup questions part of the ´validate_nulls' method in the row parser, but it ended up being too complex and we didn't think it was worth the time investment in the end.

## Remaining to do
- There are other bits of error message logic that need updating I believe, but I'll spin them out into separate tickets. Thread for context: https://softwire.slack.com/archives/C01QGR78H9S/p1681397427245309

### DIFF
--- a/app/services/bulk_upload/lettings/year2023/row_parser.rb
+++ b/app/services/bulk_upload/lettings/year2023/row_parser.rb
@@ -334,7 +334,7 @@ class BulkUpload::Lettings::Year2023::RowParser
 
   validate :validate_valid_radio_option, on: :before_log
 
-  validate :validate_uprn_exists_if_any_key_adddress_fields_are_blank
+  validate :validate_uprn_exists_if_any_key_adddress_fields_are_blank, on: :after_log
 
   def self.question_for_field(field)
     QUESTIONS[field]
--- a/app/services/bulk_upload/lettings/year2023/row_parser.rb
+++ b/app/services/bulk_upload/lettings/year2023/row_parser.rb
@@ -292,6 +292,7 @@ class BulkUpload::Lettings::Year2023::RowParser
   validates :field_72, format: { with: /\A\d{1,3}\z|\AR\z/, message: "Age of person 7 must be a number or the letter R" }, allow_blank: true, on: :after_log
   validates :field_76, format: { with: /\A\d{1,3}\z|\AR\z/, message: "Age of person 8 must be a number or the letter R" }, allow_blank: true, on: :after_log
 
+  validates :field_4, presence: { message: I18n.t("validations.not_answered", question: "needs type") }, on: :after_log
   validates :field_6, presence: { message: I18n.t("validations.not_answered", question: "property renewal") }, on: :after_log
   validates :field_7, presence: { message: I18n.t("validations.not_answered", question: "tenancy start date (day)") }, on: :after_log
   validates :field_8, presence: { message: I18n.t("validations.not_answered", question: "tenancy start date (month)") }, on: :after_log
@@ -299,7 +300,6 @@ class BulkUpload::Lettings::Year2023::RowParser
 
   validates :field_9, format: { with: /\A\d{2}\z/, message: I18n.t("validations.setup.startdate.year_not_two_digits") }, on: :after_log
 
-  validate :validate_needs_type_present, on: :after_log
   validate :validate_data_types, on: :after_log
   validate :validate_nulls, on: :after_log
   validate :validate_relevant_collection_window, on: :after_log
@@ -436,12 +436,6 @@ private
   def validate_uprn_exists_if_any_key_adddress_fields_are_blank
     if field_18.blank? && (field_19.blank? || field_21.blank?)
       errors.add(:field_18, I18n.t("validations.not_answered", question: "UPRN"))
-    end
-  end
-
-  def validate_needs_type_present
-    if field_4.blank?
-      errors.add(:field_4, I18n.t("validations.not_answered", question: "needs type"))
     end
   end
 
--- a/app/services/bulk_upload/lettings/year2023/row_parser.rb
+++ b/app/services/bulk_upload/lettings/year2023/row_parser.rb
@@ -548,13 +548,15 @@ private
       if setup_question?(question)
         fields.each do |field|
           if errors[field].present?
-            errors.add(field, I18n.t("validations.not_answered", question: question.check_answer_label&.downcase), category: :setup)
+            question_text = question.check_answer_label.presence || question.header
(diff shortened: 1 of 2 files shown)


app/services/bulk_upload/lettings/year2023/row_parser.rb
```diff
@@ -535,13 +543,15 @@ def validate_nulls
       if setup_question?(question)
         fields.each do |field|
           if errors[field].present?
-            errors.add(field, I18n.t("validations.not_answered", question: question.check_answer_label&.downcase), category: :setup)
+            question_text = question.check_answer_label.presence || question.header
+            errors.add(field, I18n.t("validations.not_answered", question: question_text&.downcase), category: :setup)
```
kosiakkatrina: Do we want any tests for this?

kosiakkatrina (approved):

app/services/bulk_upload/lettings/year2023/row_parser.rb
ghost: does the "validations.not_answered" translation handle a `nil` value?

app/services/bulk_upload/lettings/year2023/row_parser.rb
```diff
@@ -535,13 +543,15 @@ def validate_nulls
...
           end
         end
       else
         fields.each do |field|
           unless errors.any? { |e| fields.include?(e.attribute) }
-            errors.add(field, I18n.t("validations.not_answered", question: question.check_answer_label&.downcase))
+            question_text = question.check_answer_label.presence || question.header
+            errors.add(field, I18n.t("validations.not_answered", question: question_text&.downcase))
```
ghost: same question as above

app/services/bulk_upload/lettings/year2023/row_parser.rb
SamSeed-Softwire: @kosiakkatrina does the test at `spec/services/bulk_upload/lettings/year2023/row_parser_spec.rb:240` (see below) not test this functionality? Or do you mean adding another test but for the setup questions (the test mentioned deals with non-setup qus). Thanks!

SamSeed-Softwire: @bibblobcode not currently, no. My understanding is that I wouldn't be able to add that nil-handling logic to `en.yml` itself, instead I could update the code in this file so that if `question_text` is nil then I just pass "this question" or something to `validations.not_answered`. Does that sound sensible to you? Thanks!

kosiakkatrina: I meant a test for a setup section, it makes sense that this way of adding error is here, but I'm not a 100% sure if we have any cases where this would actually trigger. Happy for you to leave as is tho

ghost: if `question_text&.downcase == nil` then we shouldn't add the error

SamSeed-Softwire: @kosiakkatrina I've been trying to write an equivalent test for the setup section but am struggling a bit. All setup questions have a `check_answer_label` so the ` || question.header` logic is not being invoked for any existing questions. Therefore I was trying to use something like `allow().to receive().and_return(nil)' to make it seem like one of the existing questions actually didn't have a `check_answer_label`. Does this sound like the right approach to take, and if so how would we fill in the arguments for `allow` in `allow().to receive().and_return(nil)`? I've had a think and am a bit stumped!

Before we continue it's worth saying this might not be worth investing any more time in - I'm really curious about what would be the right thing to do here, but I'm also happy to just leave it if it's not worth it!

app/services/bulk_upload/lettings/year2023/row_parser.rb
SamSeed-Softwire: @bibblobcode I've added ` || "this question"` to the end of the line. I think this is fine because we know `question` exists (it was checked earlier in the method) so this just catches the case when the question doesn't have a header or check_answer_label. Let me know what you think! Thanks 🙂

ghost: 👍 now you don't need the & in `question_text&.downcase`

SamSeed-Softwire: Makes sense, thanks!

spec/services/bulk_upload/lettings/year2023/row_parser_spec.rb
```diff
@@ -806,8 +818,50 @@
...
+      context "when neither UPRN nor address fields are given" do
+        let(:attributes) do
+          {
+            bulk_upload:,
+            field_1: "1",
+          }
+        end
+
+        it "adds appropriate errors" do
+          expect(parser.errors[:field_18]).to eql(["You must answer UPRN"])
```
SamSeed-Softwire, line 835: This is failing because `parser.errors[:field_18]` is actually returning `["You must answer UPRN", "You must answer UPRN"]`. I can't work out why though!

natdeanlewissoftwire: Is validations.not_answered also being triggered? (in the row parser's validate_nulls method)

SamSeed-Softwire: Thanks for looking at this @natdeanlewissoftwire! For the record, no it wasn't being triggered (that was my first thought too though). I've had a look at why and this is what I've worked out:

It wasn't being triggered because when the UPRN is blank in a bulk upload, there's a line in the row parser which automatically sets `uprn_known` as 0. Then the answer to the UPRN question is inferred (as "Not known") as can be seen in `app/models/form/lettings/questions/uprn.rb`, and because it's inferred it's understood to be completed. So the final if statement in the `validate_nulls` method doesn't get reached because it ignores completed questions. This is fine because the UPRN question is allowed to be null - that's why the UPRN known? question exists. The reason we want to see a "You must answer UPRN" error message is because UPRN can only be null if certain other address fields (address line 1 and town/city) are not null.

PR #1525, opened 2023-04-06
CLDC-1975 format money amounts with currency and 2 decimals

ghost:
- add missing spec files for several value check pages
- move up formatted value helper method to Log model
- format interruption screen pages


spec/helpers/interruption_screen_helper_spec.rb
```diff
@@ -191,5 +238,35 @@
...
+          "translation" => "test.title_text.one_argument",
+          "arguments" => [{ "key" => "brent", "label" => true, "i18n_template" => "argument", money: false }],
+        }
+        expect(display_title_text(informative_text_hash, lettings_log)).to eq("You said this: 12345.0")
+      end
+    end
+
+    context "when money set to true" do
+      it "formats the value as money with currency and 2 decimals" do
+        informative_text_hash = {
```
arfacamble: `title_text_hash`
and above.

arfacamble (changes requested):
I put in one minor correction in a test file.
But more importantly, there is existing logic that does this.
If you see about_price_value_check informative_text or buyer1_value_check title_text it is already possible to insert values in I18n formatted as currency using field_formatted_as_currency on the sales log (this will need lifting up to the log so we can call it for lettings as well)

config/locales/en.yml
```diff
@@ -283,7 +283,7 @@ en:
             general_needs: "Enter a value for the support charge between £0 and £60 per week if the landlord is a local authority and it is a general needs letting"
             supported_housing: "Enter a value for the support charge between £0 and £120 per week if the landlord is a local authority and it is a supported housing letting"
       ecstat:
-        over_hard_max: "Net income of £%{hard_max} per week is too high for given the tenant’s working situation"
+        over_hard_max: "Net income of %{hard_max} per week is too high for given the tenant’s working situation"
```
arfacamble:
```suggestion
        over_hard_max: "Net income of %{hard_max} per week is too high given the tenant’s working situation"
```

arfacamble (approved):
I think I just put a comment for one small typo

PR #1533, opened 2023-04-12
[CLDC-2202] Allow coordinators to set created_by

ghost:
https://digital.dclg.gov.uk/jira/browse/CLDC-2202

- Allow coordinators to set created_by, scoped only to user's organisation
- Show user email when selecting created_by in sales questions
- move sales question and page from `/common/` to `/sales/`

### DIFF
--- a/app/models/form/common/questions/created_by_id.rb
+++ b/app/models/form/common/questions/created_by_id.rb
@@ -11,8 +11,8 @@ class Form::Common::Questions::CreatedById < ::Form::Question
     answer_opts = { "" => "Select an option" }
     return answer_opts unless ActiveRecord::Base.connected?
 
-    User.select(:id, :name).each_with_object(answer_opts) do |user, hsh|
-      hsh[user.id] = user.name
+    User.select(:id, :name, :email).each_with_object(answer_opts) do |user, hsh|
+      hsh[user.id] = "#{user.name} (#{user.email})"
       hsh
     end
   end
@@ -31,7 +31,10 @@ class Form::Common::Questions::CreatedById < ::Form::Question
   end
 
   def hidden_in_check_answers?(_log, current_user)
-    !current_user.support?
+    return false if current_user.support?
+    return false if current_user.data_coordinator?
+
+    true
   end
 
   def derived?
--- a/app/models/form/common/questions/created_by_id.rb
+++ b/app/models/form/common/questions/created_by_id.rb
@@ -1,4 +1,6 @@
 class Form::Common::Questions::CreatedById < ::Form::Question
+  ANSWER_OPTS = { "" => "Select an option" }.freeze
+
   def initialize(id, hsh, page)
     super
     @id = "created_by_id"
@@ -8,19 +10,23 @@ class Form::Common::Questions::CreatedById < ::Form::Question
   end
 
   def answer_options
-    answer_opts = { "" => "Select an option" }
-    return answer_opts unless ActiveRecord::Base.connected?
+    return ANSWER_OPTS unless ActiveRecord::Base.connected?
 
-    User.select(:id, :name, :email).each_with_object(answer_opts) do |user, hsh|
+    User.select(:id, :name, :email).each_with_object(ANSWER_OPTS.dup) do |user, hsh|
       hsh[user.id] = "#{user.name} (#{user.email})"
       hsh
     end
   end
 
-  def displayed_answer_options(log, _user = nil)
-    return answer_options unless log.owning_organisation
+  def displayed_answer_options(log, user = nil)
+    return ANSWER_OPTS unless log.owning_organisation
+    return ANSWER_OPTS unless user
+    return ANSWER_OPTS unless user.support? || user.data_coordinator?
 
-    user_ids = log.owning_organisation.users.pluck(:id) + [""]
+    users = user.support? ? log.owning_organisation.users : user.organisation.users
+
+    user_ids = users.pluck(:id) + [""]
+
     answer_options.select { |k, _v| user_ids.include?(k) }
   end
 
--- a/app/models/form/lettings/questions/created_by_id.rb
+++ b/app/models/form/lettings/questions/created_by_id.rb
@@ -1,4 +1,6 @@
 class Form::Lettings::Questions::CreatedById < ::Form::Question
+  ANSWER_OPTS = { "" => "Select an option" }.freeze
+
   def initialize(id, hsh, page)
     super
     @id = "created_by_id"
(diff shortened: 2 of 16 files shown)


app/models/form/common/questions/created_by_id.rb
```diff
@@ -11,8 +11,8 @@ def answer_options
     answer_opts = { "" => "Select an option" }
     return answer_opts unless ActiveRecord::Base.connected?
 
-    User.select(:id, :name).each_with_object(answer_opts) do |user, hsh|
-      hsh[user.id] = user.name
+    User.select(:id, :name, :email).each_with_object(answer_opts) do |user, hsh|
```
asmega: pretty sure you will want different scope here depending on the user viewing this question otherwise the data coordinator will see everyone

ghost: pretty sure about it too, I have left a question about this on the ticket and will only merge when answered and addressed

jamesrose: I'm pretty sure this should be restricted to users within their org, or perhaps child/parent orgs.

app/models/form/lettings/questions/created_by_id.rb
```diff
@@ -8,21 +10,27 @@ def initialize(id, hsh, page)
   end
 
   def answer_options
-    answer_opts = { "" => "Select an option" }
-    return answer_opts unless ActiveRecord::Base.connected?
+    return ANSWER_OPTS unless ActiveRecord::Base.connected?
```
asmega: if we are touching any `ActiveRecord::Base.connected?` can we start removing them? i don't understand why we code for this, if we can't connect to the DB we won't know who the user is any way

app/models/form/lettings/questions/created_by_id.rb
```diff
@@ -8,21 +10,27 @@ def initialize(id, hsh, page)
...
-  def displayed_answer_options(log, _user = nil)
-    return answer_options unless log.owning_organisation && log.managing_organisation
+  def displayed_answer_options(log, user = nil)
+    return ANSWER_OPTS unless user
+    return ANSWER_OPTS unless user.support? || user.data_coordinator?
 
     user_ids = [""]
-    user_ids += log.owning_organisation.users.pluck(:id)
-    user_ids += log.managing_organisation.users.pluck(:id)
+    user_ids += if user.support?
```
asmega: why are we selecting every single user in the database then removing the vast majority of them leaving the interesting ones? surely we'd want just simply select the ones we're concerned with for our scope? this seems like the most inefficient way of getting the data we want

app/models/form/lettings/questions/created_by_id.rb
```diff
@@ -8,21 +10,27 @@ def initialize(id, hsh, page)
...
       hsh[user.id] = "#{user.name} (#{user.email})"
       hsh
     end
   end
 
-  def displayed_answer_options(log, _user = nil)
-    return answer_options unless log.owning_organisation && log.managing_organisation
+  def displayed_answer_options(log, user = nil)
+    return ANSWER_OPTS unless user
+    return ANSWER_OPTS unless user.support? || user.data_coordinator?
```
kosiakkatrina: We're not routing to this page unless it's support or data_coordinator anyway. Do we need this line?

app/models/form/lettings/questions/created_by_id.rb
ghost: To be honest, I do not know why these checks are here. My best guess is that at some point this method is called and we do not care about the value returned by this method and just return `{ "" => "Select an option" }`. This wasn't changed, just moved the value into a constant.

Do you know whether they can be safely removed?

app/models/form/lettings/questions/created_by_id.rb
```diff
@@ -8,21 +10,27 @@ def initialize(id, hsh, page)
...
     user_ids = [""]
-    user_ids += log.owning_organisation.users.pluck(:id)
-    user_ids += log.managing_organisation.users.pluck(:id)
+    user_ids += if user.support?
+                  [
+                    (log.owning_organisation&.users&.pluck(:id) if log.owning_organisation),
+                    (log.managing_organisation&.users&.pluck(:id) if log.managing_organisation),
+                  ]
+                else
+                  user.organisation.users.pluck(:id)
```
kosiakkatrina: What is the data coordinator want to assign to someone from managing organisation? Should they be able to do that?

app/models/form/lettings/questions/created_by_id.rb
```diff
@@ -34,7 +42,10 @@ def label_from_value(value, _log = nil, _user = nil)
   end
 
   def hidden_in_check_answers?(_log, current_user)
-    !current_user.support?
+    return false if current_user.support?
+    return false if current_user.data_coordinator?
+
+    true
   end
```
kosiakkatrina: Do we need this at all? It will be hidden in CYA if it's not routed to

app/models/form/sales/questions/created_by_id.rb
```diff
@@ -0,0 +1,55 @@
...
+
+    answer_options[value]
+  end
+
+  def hidden_in_check_answers?(_log, current_user)
+    return false if current_user.support?
+    return false if current_user.data_coordinator?
+
+    true
+  end
```
kosiakkatrina, line 39: Same as above

app/models/form/sales/questions/created_by_id.rb
```diff
@@ -0,0 +1,55 @@
...
+      hsh
+    end
+  end
+
+  def displayed_answer_options(log, user = nil)
+    return ANSWER_OPTS unless log.owning_organisation
+    return ANSWER_OPTS unless user
+    return ANSWER_OPTS unless user.support? || user.data_coordinator?
+
+    users = user.support? ? log.owning_organisation.users : user.organisation.users
```
kosiakkatrina: Do we need to do this? Can it always be `log.owning_organisation.users` ? Cause data provider won't be able to see any logs other than their own org. Also, can it be assigned to someone from managing org?
edit: Ignore my second question, it's sales..

app/models/form/lettings/questions/created_by_id.rb
asmega: i don't see why not, if we can't connect to the database you won't even be able to authenticate the user

app/models/form/lettings/questions/created_by_id.rb
ghost: no, this isn't what we need right now

app/models/form/sales/questions/created_by_id.rb
ghost: we want to be able to select users  only from `user.organisation.users` if the current user is data coordinator

kosiakkatrina (approved):

PR #1519, opened 2023-04-05
CLDC-2054 Add request a merge page

kosiakkatrina:
Add a before you start page for merge organisations.
image
image

PR #1535, opened 2023-04-13
CLDC-2055 Add merging organisations merge request page

kosiakkatrina:
This PR build on top of https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/1519

When we click Start Now from the merge-request page, a new merge request it created and persisted with the current user organisation as the organisation requesting the merge. We didn't want to persist the merge request based on the ticket description, but it's easier to persist it and mark it as "submitted" etc. later on.

This PR also adds a table for storing merging organisations against the merge request and other merging organisations, which are just free text (in case those organisations don't exist on the service).

image

There are some validation errors:
- When a merging organisation that is involved in another merge is added we display this error:
image

- When add organisation is clicked without selecting an organisation we get this: 
image

- When we try to add an organisation that is already added, nothing happens, we don't add the organisation twice.

### DIFF
--- a/app/controllers/merge_requests_controller.rb
+++ b/app/controllers/merge_requests_controller.rb
@@ -1,14 +1,9 @@
 class MergeRequestsController < ApplicationController
   before_action :authenticate_user!
-  # before_action :authenticate_scope!
+  before_action :find_resource, only: %i[organisations update_organisations]
 
-  def create
-    @merge_request = MergeRequest.new
-  end  
-  
   def create
     @merge_request = MergeRequest.create!(merge_request_params)
-
     redirect_to merge_request_organisations_path(@merge_request)
   end
 
@@ -31,7 +26,28 @@ class MergeRequestsController < ApplicationController
     render "organisations"
   end
 
-  private
+  def update_organisations
+    if @merge_request.merging_organisation_ids
+      @merge_request.merging_organisation_ids << params[:merge_request][:merging_organisation].to_i
+      @merge_request.save!
+    else
+      @merge_request.update!(merging_organisation_ids: [params[:merge_request][:merging_organisation]])
+    end
+    @answer_options = organisations_answer_options
+    @merging_organisations_list = [@merge_request.requesting_organisation] + @merge_request.merging_organisations
+    render "organisations"
+  end
+
+private
+
+  def organisations_answer_options
+    answer_options = { "" => "Select an option" }
+
+    Organisation.all.pluck(:id, :name).each do |organisation|
+      answer_options[organisation[0]] = organisation[1]
+    end
+    answer_options
+  end
 
   def answer_options
     answer_options = { "" => "Select an option" }
@@ -43,10 +59,13 @@ class MergeRequestsController < ApplicationController
   end
   
   def merge_request_params
-    required_params = params.fetch(:merge_request, {}).permit(:requesting_organisation)
+    merge_params = params.fetch(:merge_request, {}).permit(:requesting_organisation)
 
-    required_params[:requesting_organisation] = current_user.organisation
-    required_params
+    merge_params[:requesting_organisation] = current_user.organisation
+    merge_params
   end
 
+  def find_resource
+    @merge_request = MergeRequest.find(params[:merge_request_id])
+  end
 end
--- a/app/controllers/merge_requests_controller.rb
+++ b/app/controllers/merge_requests_controller.rb
@@ -1,6 +1,6 @@
 class MergeRequestsController < ApplicationController
   before_action :authenticate_user!
-  before_action :find_resource, only: %i[organisations update_organisations remove_merging_organsiation]
+  before_action :find_resource, only: %i[organisations update_organisations remove_merging_organisation update_other_merging_organisations]
 
   def create
     @merge_request = MergeRequest.create!(merge_request_params)
@@ -26,7 +26,15 @@ class MergeRequestsController < ApplicationController
     end
   end
 
-  def remove_merging_organsiation
+  def update_other_merging_organisations
+    if @merge_request.update(other_merging_organisaitions_params)
+      redirect_to merge_request_absorbing_organisation_path(@merge_request)
+    else
+      render :organisations, status: :unprocessable_entity
+    end
+  end
+
+  def remove_merging_organisation
     MergeRequestOrganisation.find_by(merge_request_organisation_params).destroy!
     @merge_request.reload
     @answer_options = organisations_answer_options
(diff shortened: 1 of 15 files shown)


app/models/merge_request_organisation.rb
```diff
@@ -0,0 +1,31 @@
+class MergeRequestOrganisation < ApplicationRecord
+  belongs_to :merge_request, class_name: "MergeRequest"
+  belongs_to :merging_organisation, class_name: "Organisation"
+  validates :merge_request_id, presence: { message: I18n.t("validations.merge_request.merge_request_id.blank") }
```
asmega: why do we choose to validate the column `merge_request_id` as opposed to the presence of the association `merge_request`?

app/models/merge_request_organisation.rb
```diff
@@ -0,0 +1,31 @@
...
+  validates :merging_organisation_id, presence: { message: I18n.t("validations.merge_request.merging_organisation_id.blank") }
+  validate :validate_merging_organisations
+
+  has_paper_trail
+
+private
+
+  def validate_merging_organisations
+    if MergeRequestOrganisation.where(merge_request_id:, merging_organisation_id:).count.positive?
```
asmega: same question as above on why do we chose to use the ids over the object?

app/controllers/merge_requests_controller.rb
```diff
@@ -0,0 +1,77 @@
...
+    @merge_request = MergeRequest.find(params[:merge_request_id])
+    @answer_options = organisations_answer_options
+    @merging_organisations_list = [@merge_request.requesting_organisation] + @merge_request.merging_organisations
+  end
+
+  def update_organisations
+    merge_request_organisation = MergeRequestOrganisation.new(merge_request_organisation_params)
+    @answer_options = organisations_answer_options
+    if merge_request_organisation.save
+      @merge_request.reload
```
asmega: why do we have to `reload` here? our object should already by in sync?

app/models/merge_request_organisation.rb
kosiakkatrina: No reason, good spot!

app/controllers/merge_requests_controller.rb
kosiakkatrina: They are, not sure why I had them there, thanks!

app/controllers/merge_requests_controller.rb
```diff
@@ -0,0 +1,75 @@
...
+  before_action :authenticate_user!
+  before_action :find_resource, only: %i[update organisations update_organisations remove_merging_organisation]
+
+  def create
+    @merge_request = MergeRequest.create!(merge_request_params)
+    redirect_to organisations_merge_request_path(@merge_request)
+  end
+
+  def organisations
+    @merge_request = MergeRequest.find(params[:id])
```
asmega: i think this line is now redundant with your before filter

app/views/merge_requests/organisations.html.erb
```diff
@@ -0,0 +1,54 @@
...
+    <%= table.body do |body| %>
+        <%= body.row do |row| %>
+            <% row.cell(text: @merge_request.requesting_organisation.name) %>
+            <% row.cell(html_attributes: {
+                scope: "row",
+                class: "govuk-!-text-align-right",
+                }) %>
+        <% end %>
+    <% end %>
+  <% @merge_request.merging_organisations&.each do |merging_organisation| %>
```
asmega: minor thing, shouldn't need a nil guard as the return will always be a collection

app/views/merge_requests/organisations.html.erb
```diff
@@ -0,0 +1,54 @@
...
+          }) do %>
+            <%= govuk_link_to("Remove", organisations_remove_merge_request_path(merge_request: { merging_organisation: merging_organisation.id })) %>
+          <% end %>
+        <% end %>
+      <% end %>
+    <% end %>
+  <% end %>
+<% end %>
+<%= form_with model: @merge_request, url: merge_request_path(id: @merge_request.id), method: :patch do |f| %>
+  <%= govuk_details(summary_text: "Can't find the managing agent you're looking for?") do %>
```
asmega: smart quote

spec/requests/merge_requests_controller_spec.rb
```diff
@@ -0,0 +1,205 @@
+require "rails_helper"
+
+RSpec.describe MergeRequestsController, type: :request do
+  let(:organisation) { user.organisation }
+  let!(:other_organisation) { FactoryBot.create(:organisation, name: "Other Test Org") }
```
asmega: think this can be a lazy `let`?

asmega (approved):

app/controllers/merge_requests_controller.rb
```diff
@@ -5,6 +5,8 @@ class MergeRequestsController < ApplicationController
 
   def create
     @merge_request = MergeRequest.create!(merge_request_params)
+    MergeRequestOrganisation.create!({ merge_request: @merge_request, merging_organisation_id: merge_request_params.fetch(:requesting_organisation_id) })
```
asmega: I'd say wrap this in a transaction block so if one create fails it'll rollback the other

asmega: `merging_organisation_id` we're using ids again can these not be the objects?

app/models/merge_request.rb
```diff
@@ -2,4 +2,11 @@ class MergeRequest < ApplicationRecord
   belongs_to :requesting_organisation, class_name: "Organisation"
   has_many :merge_request_organisations
   has_many :merging_organisations, through: :merge_request_organisations, source: :merging_organisation
+  scope :unsubmitted, -> { where.not(status: "unsubmitted") }
```
asmega: is this right? a scope of `unsubmitted` is where the `status` is not `unsubmitted`

app/views/organisations/merge_request.html.erb
```diff
@@ -43,6 +43,7 @@
 
     <%= form_for @merge_request, url: merge_requests_path do |f| %>
       <%= f.hidden_field :requesting_organisation_id, value: @organisation.id %>
+            <%= f.hidden_field :status, value: "unsubmitted" %>
```
asmega: i don't see why we want to expose this on the frontend. can we not just do this on the backend when they `POST` to the `create`?

app/models/merge_request.rb
asmega: also by using an enum does it not create these scopes by default out of the box? or maybe i'm thinking of the predicates?

kosiakkatrina: Yup, you get that by default, in fact I was using it without realising 😬

asmega (approved):

PR #1537, opened 2023-04-13
[CLDC-2260] Move UPRN Known and UPRN questions to same page

ghost:
- move UPRN know and UPRN to a shared page
- amend CYA page to hide UPRN Known and show UPRN with "not known" value when not known
- set uprn known to 1 when updating UPRN


app/models/form/lettings/pages/uprn.rb
```diff
@@ -6,14 +6,13 @@ def initialize(id, hsh, subsection)
...
+      Form::Lettings::Questions::UprnKnown.new(nil, nil, self),
       Form::Lettings::Questions::Uprn.new(nil, nil, self),
     ]
   end
 
   def routed_to?(log, _current_user = nil)
-    return false if log.is_supported_housing?
-
-    log.uprn_known == 1
+    log.is_supported_housing? ? false : true
```
asmega:
```suggestion
    !log.is_supported_housing?
```

asmega:
The help text seems to be duplicated which I assume is not desired?

Screenshot 2023-04-14 at 11 22 58

ghost:
@asmega good catch, updated!

app/models/log.rb
```diff
@@ -50,6 +50,7 @@ def process_uprn_change!
 
       presenter = UprnDataPresenter.new(service.result)
 
+      self.uprn_known = 1
```
kosiakkatrina, line 53: How come this needs to be set separately? I expected it would have to be filled in if the uprn is given

kosiakkatrina (approved):

PR #1538, opened 2023-04-13
CLDC-2199 Fix LA/rent soft validation bugs

natdeanlewissoftwire:
ticket: https://digital.dclg.gov.uk/jira/browse/CLDC-2199

image
(the double £ display is a known bug being fixed by another ticket)

This issue is just caused by the soft validation pages not being by the la question. Have also added them after the beds question as that would cause a similar issue. Have put them after relevant setup questions too: 
image


app/models/form/lettings/subsections/setup.rb
```diff
@@ -10,14 +10,22 @@ def pages
     @pages ||= [
       organisation_page,
       stock_owner_page,
+      Form::Lettings::Pages::MinRentValueCheck.new(nil, nil, self, check_answers_card_number: nil),
```
jamesrose: You could just make the default value nil here, if required.

natdeanlewissoftwire: True, I think there's a tendency to do this more explicitly but I'm not that sold on it myself, will make this change

app/models/form/lettings/subsections/setup.rb
```diff
@@ -10,14 +10,22 @@ def pages
...
       Form::Lettings::Pages::Location.new(nil, nil, self),
+      Form::Lettings::Pages::MinRentValueCheck.new(nil, nil, self),
+      Form::Lettings::Pages::MaxRentValueCheck.new(nil, nil, self),
       Form::Lettings::Pages::Renewal.new(nil, nil, self),
       Form::Lettings::Pages::TenancyStartDate.new(nil, nil, self),
+      Form::Lettings::Pages::MinRentValueCheck.new(nil, nil, self),
+      Form::Lettings::Pages::MaxRentValueCheck.new(nil, nil, self),
       Form::Lettings::Pages::RentType.new(nil, nil, self),
+      Form::Lettings::Pages::MinRentValueCheck.new(nil, nil, self),
+      Form::Lettings::Pages::MaxRentValueCheck.new(nil, nil, self),
```
natdeanlewissoftwire: It seems ugly to have to add the value check so many times to the setup section but all are relevant, have made this field optional to be consistent with how this is handled for sales logs

spec/models/form_handler_spec.rb
arfacamble, line 1: 🥇

spec/models/form/lettings/subsections/setup_spec.rb
arfacamble, line 1: 🥉

app/models/form/lettings/subsections/setup.rb
```diff
@@ -10,14 +10,22 @@ def pages
     @pages ||= [
       organisation_page,
       stock_owner_page,
+      Form::Lettings::Pages::MinRentValueCheck.new("stock_owner_min_rent_value_check", nil, self),
+      Form::Lettings::Pages::MaxRentValueCheck.new("stock_owner_max_rent_value_check", nil, self),
```
arfacamble, line 14: teh rent soft validations can get triggered by the stock owner?

arfacamble (approved):
pending question

app/models/form/lettings/subsections/setup.rb
natdeanlewissoftwire: yep, it owning org provider type affects lettype which affects rent ranges and hence the validation triggering/not

spec/models/form/lettings/subsections/setup_spec.rb
natdeanlewissoftwire: fair

PR #1539, opened 2023-04-13
Speed up bulk upload with missing memoization

asmega:
# Context

- Bulk upload has been acting slow and spent some time actually investigating why

# Changes

- Add missing memoization to bulk upload which causes performance boost
- See evidence below

```
Finished in 4 minutes 0.9 seconds (files took 5.27 seconds to load)
210 examples, 0 failures, 2 pending

Finished in 30.85 seconds (files took 5.23 seconds to load)
210 examples, 0 failures, 2 pending
```

### DIFF
--- a/app/services/bulk_upload/lettings/year2022/row_parser.rb
+++ b/app/services/bulk_upload/lettings/year2022/row_parser.rb
@@ -340,9 +340,9 @@ class BulkUpload::Lettings::Year2022::RowParser
 
     return true if blank_row?
 
-    super
-
     log.valid?
+
+    super
 
     log.errors.each do |error|
       fields = field_mapping_for_errors[error.attribute] || []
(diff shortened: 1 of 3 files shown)


jamesrose (approved):

arfacamble (approved):

arfacamble:
I wondered when I saw the lines that failed the linter if that was best practice, but I also know that chaining on a do end is also gross.
Interested to know what is the nicest way to write that.

asmega:
> I wondered when I saw the lines that failed the linter if that was best practice, but I also know that chaining on a do end is also gross. Interested to know what is the nicest way to write that.

turns out i didn't need it so i deleted it. i think i would have multi lined it with do end blocks although that would probably still failed the linter

PR #1546, opened 2023-04-14
CLDC-2274 Fix buyer 2 nationality routing bug

natdeanlewissoftwire:
ticket: https://digital.dclg.gov.uk/jira/browse/CLDC-2274

Buyer 2 nationality was displaying regardless of buyer being interviewed or not - it should follow the same pattern as every other question in the subsection, in that it should only display when the bueyr has seen the privacy notice or was not interviewed. This is for the same reason as every other question in this subsection.

fixed behaviour: (previously buyer 2 nationality was displaying at the bottom of this page in this case)
image


app/models/form/sales/pages/buyer2_nationality.rb
```diff
@@ -2,7 +2,16 @@ class Form::Sales::Pages::Buyer2Nationality < ::Form::Page
   def initialize(id, hsh, subsection)
     super
     @id = "buyer_2_nationality"
-    @depends_on = [{ "joint_purchase?" => true }]
+    @depends_on = [
```
natdeanlewissoftwire, line 5: this contains the actual behaviour for the ticket, rest is just refactoring

jamesrose:
@natdeanlewissoftwire Please could you include what was fixed and why?

natdeanlewissoftwire:
> @natdeanlewissoftwire Please could you include what was fixed and why?

@jamesrose done 👍

app/models/form/sales/pages/housing_benefits.rb
```diff
@@ -2,7 +2,14 @@ class Form::Sales::Pages::HousingBenefits < ::Form::Page
   def initialize(id, hsh, subsection, joint_purchase:)
     super(id, hsh, subsection)
     @joint_purchase = joint_purchase
-    @depends_on = [{ "jointpur" => @joint_purchase ? 1 : 2 }]
+    @depends_on =
```
asmega: it works but think i would extract to a method at this point. not sure how i feel having an `if` within an array?

db/schema.rb
```diff
@@ -519,6 +519,7 @@
     t.integer "prevten"
     t.integer "mortgageused"
     t.integer "wchair"
+    t.integer "income2_value_check"
```
asmega: i don't recommend checking in the schema file if you haven't actually made any schema changes

asmega (approved):

db/schema.rb
natdeanlewissoftwire: good point

app/models/form/sales/pages/housing_benefits.rb
natdeanlewissoftwire: yep agreed

PR #1553, opened 2023-04-17
CLDC-2272 Fix UPRN income validation bug

natdeanlewissoftwire:
Adds uprn to lettings rent range and sales income validations.

The uprn processing must happen before validation otherwise the user could get in situations where the user had failed a validation but the uprn processing had changed the data related to the failure, and so the log error could not be fixed by the user.

ticket: https://digital.dclg.gov.uk/jira/browse/CLDC-2272


ghost (approved):
very good detective work!

PR #1557, opened 2023-04-18
CLDC-2197 Improve column counting in bulk upload

SamSeed-Softwire:
## Context
- https://digital.dclg.gov.uk/jira/browse/CLDC-2197

## The changes
- Check field headers more strictly to ensure they are valid (we weren't checking if they were numbers before).
- Count the number of valid field headers and raise an error if this is not the expected amount (134).
- Update the 'max columns count' check so that it only happens when there are no headers in the file (this is so that if there are headers, extra columns without valid field headers are safely ignored).
- Remove the 'min columns count' check, because a data row could have nil values for its final field(s) and therefore seem to have too few columns despite being valid.
- Update the 2022 csv/row parsers to work in the same way as 2023 (this was arguably not that important because 2022 is almost over, but it felt cleaner to bring it into line with 2023.


app/services/bulk_upload/lettings/year2022/csv_parser.rb
```diff
@@ -1,8 +1,8 @@
 require "csv"
 
 class BulkUpload::Lettings::Year2022::CsvParser
-  MIN_COLUMNS = 134
-  MAX_COLUMNS = 136
+  FIELDS = 134
+  MAX_COLUMNS = 135
```
kosiakkatrina, line 5: Why is MAX_COLUMNS and not 134? I assume this is max without counting the offset columns?

SamSeed-Softwire: So I originally put 134 but after discussion with @asmega we decided to go for 135 to give a bit of leniency to the user. The thinking is that they might end up with an additional blank column at the end of their file depending on the software they use to prep the csv, and depending on the technical ability of the user this might not be obvious to them.

kosiakkatrina: Ah calm calm. Nice!

kosiakkatrina (approved):

asmega (approved):

kosiakkatrina (approved):

PR #1560, opened 2023-04-19
Add 23/24 fields to import mappings

kosiakkatrina:
We need to ensure that new questions/fields for 2023/24 collection period are being imported and any changes to answer options are mapped correctly.
So far we have only mapped out 2022/23 logs fields.
This PR adds any missing mappings to the import based on the annual form changes both for lettings and sales imports.

### DIFF
--- a/app/services/imports/lettings_logs_import_service.rb
+++ b/app/services/imports/lettings_logs_import_service.rb
@@ -202,15 +202,13 @@ module Imports
       attributes["first_time_property_let_as_social_housing"] = first_time_let(attributes["rsnvac"])
       attributes["declaration"] = declaration(xml_doc)
 
-      if attributes["startdate"] >= Time.zone.local(2023, 4, 1)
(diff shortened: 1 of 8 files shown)


app/services/imports/lettings_logs_import_service.rb
```diff
@@ -202,6 +202,16 @@ def create_log(xml_doc)
       attributes["first_time_property_let_as_social_housing"] = first_time_let(attributes["rsnvac"])
       attributes["declaration"] = declaration(xml_doc)
 
+      if attributes["startdate"] >= Time.zone.local(2023, 4, 1)
```
jamesrose: Why do we need this condition? I would have thought the values would just be nil if 2022.

kosiakkatrina: good point

jamesrose: Note we do this twice in this diff.

jamesrose (approved):

PR #1568, opened 2023-04-19
CLDC-1633 build feature csv download of sales logs

arfacamble:
DEPLOY-ACTION:
Jemma: 
_"When this hits production in my absence, could you just give Miguel, Alistair and Sam the heads up that it's there? I don't expect them to look at it straight away because they've got a publication deadline, but just so they're aware:"_
___

https://digital.dclg.gov.uk/jira/browse/CLDC-1633

One important change in the model for how this csv will be generated comes from the fact that I couldn't work out after poking around with James, how the old model for exporting lettings logs works exactly, there's some rails magic and some mysterious way of not losing the filtering passing between the Job, a LettingsLog class method and the LettingsLogCsvService.
I hope this way of doing it is easier to read, follow and test. It does mean that until the tech debt LettingsLogCsvService is tackled there are one or two places with a bit of messy looking inconsistency


app/jobs/email_csv_job.rb
```diff
@@ -5,14 +5,25 @@ class EmailCsvJob < ApplicationJob
...
+      unfiltered_logs = organisation.present? && user.support? ? LettingsLog.visible.where(owning_organisation_id: organisation.id) : user.lettings_logs.visible
+    else
+      unfiltered_logs = organisation.present? && user.support? ? SalesLog.visible.where(owning_organisation_id: organisation.id) : user.sales_logs.visible
+    end
     filtered_logs = FilterService.filter_logs(unfiltered_logs, search_term, filters, all_orgs, user)
 
-    filename = organisation.present? ? "logs-#{organisation.name}-#{Time.zone.now}.csv" : "logs-#{Time.zone.now}.csv"
+    filename = [log_type, "logs", organisation&.name, Time.zone.now].compact.join("-") + ".csv"
+
+    csv_string = if log_type == "sales"
```
jamesrose: This can be in one condition as per above

app/models/log.rb
```diff
@@ -136,6 +136,16 @@ def field_formatted_as_currency(field_name)
     format_as_currency(field_value)
   end
 
+  def creation_method
+    if bulk_upload_id.present?
+      "bulk upload"
+    elsif old_id.present?
+      "migrated"
```
jamesrose: Do we need this one?

spec/factories/form.rb
```diff
@@ -0,0 +1,25 @@
...
+  def build
+    form = FormFixture.new(nil, @year, [], @type)
+    @sections.each { |section| section.form = form }
+    form.sections = @sections
+    form.subsections = form.sections.flat_map(&:subsections)
+    form.pages = form.subsections.flat_map(&:pages)
+    form.questions = form.pages.flat_map(&:questions)
+    form
+  end
+end
```
arfacamble, line 25: This is where the shape of this solution is determined.
`Form` does not have accessors like section, subsection, etc and it is initialised with an array of `Section` classes, not instances, which it initialises and from that populates various instance variables. 
So to create something that `is_a? Form` with the structure we want, I have had to create FormFixture which inherits and adds setters and recreate a section of `Form:initialize` in `FormFactory:build`

spec/factories/page.rb
```diff
@@ -1,5 +1,15 @@
...
     initialize_with { new(id, nil, nil) }
+    trait :with_question do
+      transient do
+        question_id { nil }
+
+        after :build do |page, evaluator|
+          page.questions = [build(:question, id: evaluator.question_id, page:)]
+        end
+      end
+    end
```
arfacamble, line 19: the point of making all these call downwards recursively is that on line 10 we assign the page to the question (and something similar in the other factories)
which allows code of the form [sic] `question.form` to work.
This isn't actually necessary for the test I am writing here, but I think it's an important pattern to demonstrate for this model of building forms on the fly for tests

app/controllers/logs_controller.rb
```diff
@@ -28,6 +28,10 @@ def create
     end
   end
 
+  def codes_only_export?(params)
```
asmega:
```suggestion
  def codes_only_export?
```
i think you already have access to this and therefore don't need to pass it in

app/controllers/lettings_logs_controller.rb
```diff
@@ -124,6 +115,11 @@ def org_params
 
 private
 
+  def authenticate_scope!
+    codes_only_export = codes_only_export?(params)
+    head :unauthorized and return if codes_only_export && !current_user.support?
```
asmega:
```suggestion
    head :unauthorized and return if codes_only_export? && !current_user.support?
```
in tandem with my other suggestion, i feel the local variable is a bit redundant and overkill and can probably be removed

app/controllers/sales_logs_controller.rb
```diff
@@ -36,11 +37,34 @@ def edit
...
 
+  def download_csv
+    unpaginated_filtered_logs = filtered_logs(current_user.sales_logs, search_term, @session_filters)
+    codes_only = codes_only_export?(params)
+
+    render "download_csv", locals: { search_term:, count: unpaginated_filtered_logs.size, post_path: email_csv_sales_logs_path, codes_only: }
+  end
+
+  def email_csv
+    all_orgs = params["organisation_select"] == "all" # what's this for? params['organisation_select'] appears to always be nil
```
asmega: is it one of the search filters? to restrict to all orgs, or a specific org?

app/services/csv/sales_log_csv_service.rb
```diff
@@ -0,0 +1,108 @@
...
+
+        logs.find_each do |log|
+          csv << @attributes.map { |attribute| value(attribute, log) }
+        end
+      end
+    end
+
+  private
+
+    ATTRIBUTES_OF_RELATED_OBJECTS = {
```
asmega, line 20:
```
irb(main):001:0> class A
irb(main):002:1>   private
irb(main):003:1>   B = "B"
irb(main):004:1> end
=> "B"
irb(main):005:0> A::B
=> "B"
```
As an fyi declaring constants after a `private` this way does not make them private

app/controllers/logs_controller.rb
arfacamble: veyr good. Always nice to know where the bits of rails magic pop up regarding scope

app/controllers/lettings_logs_controller.rb
arfacamble: certainly given params not needed to be passed in.

app/controllers/sales_logs_controller.rb
arfacamble: I need toi investigate this. Left it in as comment since I don't have the asnwer!

app/models/form_handler.rb
```diff
@@ -40,6 +40,28 @@ def sales_forms
...
+    deprecated_questions_by_preceding_question_id(question_ids, all_questions_from_previous_forms).each do |preceding_question_id, deprecated_question|
+      index_of_preceding_question = ordered_questions.index { |q| q.id == preceding_question_id }
+      ordered_questions.insert(index_of_preceding_question + 1, deprecated_question)
+    end
+    ordered_questions
+  end
+
+  def deprecated_questions_by_preceding_question_id(current_form_question_ids, all_questions_from_previous_forms)
+    deprecated_questions = {}
+    all_questions_from_previous_forms.each_cons(2) do |preceding_question, question|
```
ghost, line 57: first time I see `each_cons` used!

app/controllers/lettings_logs_controller.rb
```diff
@@ -86,19 +81,13 @@ def destroy
...
-  def codes_only_export?(params)
-    params.require(:codes_only) == "true"
+    render "download_csv", locals: { search_term:, count: unpaginated_filtered_logs.size, post_path: email_csv_lettings_logs_path, codes_only: codes_only_export? }
   end
 
   def email_csv
     all_orgs = params["organisation_select"] == "all"
-    codes_only_export = params.require(:codes_only) == "true"
-    EmailCsvJob.perform_later(current_user, search_term, @session_filters, all_orgs, nil, codes_only_export)
+    EmailCsvJob.perform_later(current_user, search_term, @session_filters, all_orgs, nil, codes_only_export?)
```
natdeanlewissoftwire, line 90: luscious

app/controllers/organisations_controller.rb
```diff
@@ -137,6 +139,20 @@ def sales_logs
     end
   end
 
+  def download_sales_csv
+    organisation_logs = SalesLog.visible.where(owning_organisation_id: @organisation.id)
+    unpaginated_filtered_logs = filtered_logs(organisation_logs, search_term, @session_filters)
+    codes_only = params.require(:codes_only) == "true"
```
natdeanlewissoftwire, line 144: can we use the newly moved `codes_only_export?` method for this too?

app/controllers/organisations_controller.rb
```diff
@@ -137,6 +139,20 @@ def sales_logs
...
+
+    render "logs/download_csv", locals: { search_term:, count: unpaginated_filtered_logs.size, post_path: sales_logs_email_csv_organisation_path, codes_only: }
+  end
+
+  def email_sales_csv
+    codes_only_export = params.require(:codes_only) == "true"
```
natdeanlewissoftwire: ditto

natdeanlewissoftwire: (and i guess in lettings but might not be worth the bother if that's getting tech debted out later anyway)

app/helpers/logs_helper.rb
```diff
@@ -36,6 +36,15 @@ def search_label_for_controller(controller)
   def csv_download_url_for_controller(controller_type:, search:, codes_only:)
     case log_type_for_controller(controller_type)
     when "lettings" then csv_download_lettings_logs_path(search:, codes_only:)
+    when "sales" then csv_download_sales_logs_path(search:, codes_only:)
+    end
+  end
+
+  def csv_download_url_by_log_type(log_type, organisation, search:, codes_only:)
+    if log_type == :lettings
```
natdeanlewissoftwire: teeny tiny: could use a switch like no lines 37-40 to make this ever so slightly more explicit? (and more log-tyoe-symmetrical 😍)

natdeanlewissoftwire: couple of other similar places we could use switches too in this PR

app/jobs/email_csv_job.rb
```diff
@@ -5,14 +5,22 @@ class EmailCsvJob < ApplicationJob
...
+      csv_string = filtered_logs.to_csv(user, codes_only_export:)
+    else
+      unfiltered_logs = organisation.present? && user.support? ? SalesLog.visible.where(owning_organisation_id: organisation.id) : user.sales_logs.visible
+      filtered_logs = FilterService.filter_logs(unfiltered_logs, search_term, filters, all_orgs, user)
+      export_type = codes_only_export ? "codes" : "labels"
+      csv_string = Csv::SalesLogCsvService.new(export_type:).prepare_csv(filtered_logs)
+    end
 
-    filename = organisation.present? ? "logs-#{organisation.name}-#{Time.zone.now}.csv" : "logs-#{Time.zone.now}.csv"
+    filename = "#{[log_type, 'logs', organisation&.name, Time.zone.now].compact.join('-')}.csv"
```
natdeanlewissoftwire, line 21: nice

app/views/logs/csv_confirmation.html.erb
```diff
@@ -9,7 +9,8 @@
     <p class="govuk-body">Open your email inbox and click the link to download your CSV file.</p>
 
     <p class="govuk-body">
-      <%= govuk_link_to "Return to logs", lettings_logs_path %>
+      <% logs_path = controller.class.name.start_with?("Lettings") ? lettings_logs_path : sales_logs_path %>
```
natdeanlewissoftwire: could go in a view helper

spec/jobs/email_csv_job_spec.rb
```diff
@@ -39,18 +40,44 @@
...
+      it "includes the organisation name in the filename when one is provided" do
+        expect(storage_service).to receive(:write_file).with(/sales-logs-#{organisation.name}-.*\.csv/, anything)
+        job.perform(user, nil, {}, nil, organisation, nil, "sales")
+      end
+
+      it "creates a SalesLogCsvService with the correct export type" do
+        expect(Csv::SalesLogCsvService).to receive(:new).with(export_type: "labels")
+        codes_only = false
+        job.perform(user, nil, {}, nil, nil, codes_only, "sales")
+        expect(Csv::SalesLogCsvService).to receive(:new).with(export_type: "codes")
```
natdeanlewissoftwire, line 77: should these be separate tests?

natdeanlewissoftwire (approved):
lgtm pending comments

app/controllers/organisations_controller.rb
arfacamble: not quite, since it's calling a method in the logs controller, but I can recreate it in private section and make looky nicer

app/controllers/organisations_controller.rb
arfacamble: first bit done, second comment more specific please.

app/views/logs/csv_confirmation.html.erb
arfacamble: yes

app/jobs/email_csv_job.rb
arfacamble: yes

spec/jobs/email_csv_job_spec.rb
arfacamble: I don't think it really matters. Happy to be convinced otherwise

app/controllers/lettings_logs_controller.rb
arfacamble: that's a bit strong

arfacamble (commented):
submit replies to Nat.

app/controllers/organisations_controller.rb
natdeanlewissoftwire: oh actually ignore the second bit, thought this was sales specific but i see it isn't

app/views/logs/csv_confirmation.html.erb
natdeanlewissoftwire: _should_ go in a view helper

spec/jobs/email_csv_job_spec.rb
natdeanlewissoftwire: this is not the hill i choose to die on

app/helpers/logs_helper.rb
arfacamble: where???

app/views/logs/csv_confirmation.html.erb
arfacamble: why are you still talking about this?

PR #1570, opened 2023-04-20
CLDC-2262 Bulk upload clear invalid data to create forms

natdeanlewissoftwire:
This ticket clears invalid non-setup data from bulk upload logs so we can create logs in the service that the user then has to tweak.

Some errors relate to fields that the errors are not being added to, e.g. postcode being cleared means we must also clear postcode_known, but an invalid postcode error is only added to the postcode field, so we have to manually clear the postcode_known in this case as well.

Otherwise we just clear all fields that have an error and create the logs if they are now valid.

Ticket: https://digital.dclg.gov.uk/jira/browse/CLDC-2262


app/models/validations/local_authority_validations.rb
```diff
@@ -2,8 +2,8 @@ module Validations::LocalAuthorityValidations
   def validate_previous_accommodation_postcode(record)
     postcode = record.ppostcode_full
     if record.previous_postcode_known? && (postcode.blank? || !postcode.match(POSTCODE_REGEXP))
-      error_message = I18n.t("validations.postcode")
-      record.errors.add :ppostcode_full, error_message
+      record.errors.add :ppostcode_full, I18n.t("validations.postcode")
+      record.errors.add :ppcodenk, I18n.t("validations.postcode")
```
natdeanlewissoftwire: i wonder if this should be handled with derived variables instead

natdeanlewissoftwire: or maybe we should hard code a solution for bulk upload?

app/models/log.rb
```diff
@@ -136,6 +138,22 @@ def field_formatted_as_currency(field_name)
...
   end
 
+  def blank_compound_invalid_non_setup_fields!
+    validate_previous_accommodation_postcode(self)
+    self.ppcodenk = nil if errors.of_kind?(:ppostcode_full, :wrong_format)
+
+    process_uprn_change!
+    if errors.of_kind?(:uprn, :uprn_error)
+      self.uprn_known = 0
+      self.uprn_confirmed = nil
```
natdeanlewissoftwire, line 152: all these nils are to avoid a bug if there is an invalid uprn AND address data that caused the site to hang. Note to self to fully investigate why it was hanging to begin with, but this fixes it

app/models/validations/local_authority_validations.rb
natdeanlewissoftwire, line 1: moved to shared validations as there was only one of these

app/controllers/bulk_upload_lettings_resume_controller.rb
```diff
@@ -23,6 +23,11 @@ def update
     end
   end
 
+  def pluralize_logs_and_errors_warning(log_count, error_count)
```
natdeanlewissoftwire: not part of this ticket but kept seeing this so added in this fix, the previous message assumed plural logs and errors which to be fair is probably going to be the case

app/models/log.rb
```diff
@@ -98,11 +98,15 @@ def collection_period_open?
...
 
+    blank_compound_invalid_non_setup_fields!
+
     errors.each do |error|
       next if setup_ids.include?(error.attribute.to_s)
 
       public_send("#{error.attribute}=", nil)
     end
+
+    valid?
```
natdeanlewissoftwire: need to revalidate after blanking fields to remove old errors so we can set log  status correctly

config/locales/en.yml
```diff
@@ -634,6 +635,8 @@ en:
     offered: "Times previously offered since becoming available"
 
   warnings:
+    bulk_upload_log_creation_logs: "There is 1 log in this bulk upload with 1 error that still 1 need to be fixed after upload."
+    bulk_upload_log_creation_errors: "There is 1 log in this bulk upload with 1 error that still 1 need to be fixed after upload."
```
asmega: is this suppose to be here?

natdeanlewissoftwire: nope, thanks, have removed :)

asmega (approved):

PR #1574, opened 2023-04-20
CLDC-2031 Sales bulk upload

kosiakkatrina:
Update 2022/23 sales bulk upload to follow patterns similar to lettings bulk upload.
Create a log creator
Update the validator
Create csv parser
Update row parser with:
 - presence validations for setup section
 - incoming field types
 - error and attribute field mappings
 - organisation, user and date validations

Sales log bulk upload should now be able to get and create any errors that come from the logs validations themselves as well as setup errors. Any other specific errors will need to be added to row parser

Add status cache to sales logs


app/services/bulk_upload/sales/log_creator.rb
```diff
@@ -0,0 +1,70 @@
+class BulkUpload::Sales::LogCreator
```
asmega, line 1: maybe for another PR but if we don't want to duplicate this class it may be an idea to have `BulkUpload::LogCreator` where we can inject in the correct `csv_parser`

EDIT: same maybe true of `BulkUpload::Sales::Validator` too with `BulkUpload::Validator`

asmega (approved):

asmega (approved):

PR #1575, opened 2023-04-20
Replace S3 config service

asmega:
# Context

- 

# Changes

-


asmega:
this change is no longer needed

PR #1580, opened 2023-04-21
CLDC-2056 - Ask absorbing organisation

ghost:
https://digital.dclg.gov.uk/jira/browse/CLDC-2056

- Add absorbing organisation question to the flow and handle 'other' answer
- Tidy up request spec
- Add methods to aid with routing to follow up page

### DIFF
--- a/app/controllers/merge_requests_controller.rb
+++ b/app/controllers/merge_requests_controller.rb
@@ -1,7 +1,20 @@
 class MergeRequestsController < ApplicationController
-  before_action :find_resource, only: %i[update organisations update_organisations remove_merging_organisation absorbing_organisation confirm_telephone_number new_org_name]
+  before_action :find_resource, only: %i[
+    update
+    organisations
+    update_organisations
+    remove_merging_organisation
+    absorbing_organisation
+    confirm_telephone_number
+    new_org_name
+  ]
   before_action :authenticate_user!
   before_action :authenticate_scope!, except: [:create]
+  before_action :redirect_to_next_page
+
+  def absorbing_organisation; end
+  def confirm_telephone_number; end
+  def new_org_name; end
 
   def create
     ActiveRecord::Base.transaction do
@@ -13,18 +26,12 @@ class MergeRequestsController < ApplicationController
     render_not_found
   end
 
-  def absorbing_organisation; end
-  def confirm_telephone_number; end
-  def new_org_name; end
-
   def organisations
     @answer_options = organisations_answer_options
   end
 
   def update
-    if params.dig(:merge_request, :absorbing_organisation_id) == "other"
-      redirect_to next_page_path
-    elsif @merge_request.update(merge_request_params)
+    if @merge_request.update(merge_request_params)
       redirect_to next_page_path
     else
       render previous_template, status: :unprocessable_entity
@@ -48,6 +55,35 @@ class MergeRequestsController < ApplicationController
   end
 
 private
+
+  def page
+    params.dig(:merge_request, :page)
+  end
+
+  def next_page_path
+    case page
+    when "absorbing_organisation"
+      if create_new_org?
+        new_org_name_merge_request_path(@merge_request)
+      else
+        confirm_telephone_number_merge_request_path(@merge_request)
+      end
+    when "organisations"
+      absorbing_organisation_merge_request_path(@merge_request)
+    end
+  end
+
+  def previous_template
+    page
+  end
+
+  def redirect_to_next_page
+    redirect_to next_page_path if create_new_org?
+  end
+
+  def create_new_org?
+    params.dig(:merge_request, :absorbing_organisation_id) == "other"
+  end
 
   def organisations_answer_options
     answer_options = { "" => "Select an option" }
@@ -82,25 +118,6 @@ private
 
   def find_resource
     @merge_request = MergeRequest.find(params[:id])
-  end
-
-  def next_page_path
-    if params.dig(:merge_request, :absorbing_organisation_id) == "other"
-      # TODO: flow to be implemented in follow up PR
-      new_org_name_merge_request_path(@merge_request)
-    elsif params.dig(:merge_request, :absorbing_organisation_id).present?
-      confirm_telephone_number_merge_request_path(@merge_request)
-    else
-      absorbing_organisation_merge_request_path(@merge_request)
-    end
-  end
-
-  def previous_template
-    if params.dig(:merge_request, :absorbing_organisation_id).present?
-      :absorbing_organisation
-    else
-      :organisations
-    end
   end
 
   def authenticate_scope!
--- a/app/controllers/merge_requests_controller.rb
+++ b/app/controllers/merge_requests_controller.rb
@@ -10,6 +10,7 @@ class MergeRequestsController < ApplicationController
   ]
   before_action :authenticate_user!
   before_action :authenticate_scope!, except: [:create]
+  before_action :validate_response, only: %i[update]
 
   def absorbing_organisation; end
   def confirm_telephone_number; end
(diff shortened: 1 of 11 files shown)


app/controllers/merge_requests_controller.rb
```diff
@@ -13,12 +13,17 @@ def create
...
     @answer_options = organisations_answer_options
   end
 
   def update
-    if @merge_request.update(merge_request_params)
+    if params.dig(:merge_request, :absorbing_organisation_id) == "other"
+      redirect_to new_organisation_path
+    elsif @merge_request.update(merge_request_params)
       redirect_to next_page_path
     else
```
kosiakkatrina: Any reason this is not just redirecting to the next_page_path?

app/controllers/merge_requests_controller.rb
```diff
@@ -13,12 +13,17 @@ def create
     render_not_found
   end
 
+  def absorbing_organisation; end
+  def confirm_telephone_number; end
+
```
kosiakkatrina: Do we need these?
edit: I think renaming telephone_number.html.erb to confirm_telephone_number.html.erb might be enough?

app/controllers/merge_requests_controller.rb
ghost: good point, will fold into that method

app/controllers/merge_requests_controller.rb
```diff
@@ -63,19 +73,30 @@ def merge_request_params
...
     @merge_request = MergeRequest.find(params[:id])
   end
 
   def next_page_path
-    absorbing_organisation_merge_request_path(@merge_request)
+    if params.dig(:merge_request, :absorbing_organisation_id).present?
+      confirm_telephone_number_merge_request_path(@merge_request)
+    else
+      absorbing_organisation_merge_request_path(@merge_request)
+    end
```
kosiakkatrina: Hmm, maybe it's worth passing through a page/current_page in params? Similar to what we do in schemes

app/controllers/merge_requests_controller.rb
```diff
@@ -1,5 +1,5 @@
 class MergeRequestsController < ApplicationController
-  before_action :find_resource, only: %i[update organisations update_organisations remove_merging_organisation]
+  before_action :find_resource, only: %i[update organisations update_organisations remove_merging_organisation absorbing_organisation telephone_number]
```
kosiakkatrina: Should this be confirm_telephone_number?

app/controllers/merge_requests_controller.rb
ghost: yes good point, I think I will refactor this in the follow up PR where I will need to properly handle `confirm_telephone_number_merge_request_path`

config/routes.rb
```diff
@@ -137,6 +137,8 @@
       patch "organisations", to: "merge_requests#update_organisations"
       get "organisations/remove", to: "merge_requests#remove_merging_organisation"
       get "absorbing-organisation"
+      get "confirm-telephone-number"
+      get "new-org-name"
```
ghost, line 141: I do not know whether this is the best route for the feature, we can decide later at implementation stage

app/controllers/merge_requests_controller.rb
```diff
@@ -13,12 +13,18 @@ def create
     render_not_found
   end
 
+  def absorbing_organisation; end
+  def confirm_telephone_number; end
+  def new_org_name; end
+
```
kosiakkatrina: Do we need these?

app/controllers/merge_requests_controller.rb
```diff
@@ -49,6 +56,35 @@ def remove_merging_organisation
...
+    when "organisations"
+      absorbing_organisation_merge_request_path(@merge_request)
+    end
+  end
+
+  def previous_template
+    page
+  end
+
+  def redirect_to_next_page
```
ghost: this method can be reused if for some other reason we need to skip the `update` action and redirect to somewhere else

app/controllers/merge_requests_controller.rb
```diff
@@ -1,7 +1,20 @@
...
-  before_action :find_resource, only: %i[update organisations update_organisations remove_merging_organisation absorbing_organisation confirm_telephone_number new_org_name]
+  before_action :find_resource, only: %i[
+    update
+    organisations
+    update_organisations
+    remove_merging_organisation
+    absorbing_organisation
+    confirm_telephone_number
+    new_org_name
+  ]
```
ghost, line 10: after we've implemented a bit more of the merge org flow, we could decide to move from an `only` to an `except`. i think it's early to say

app/controllers/merge_requests_controller.rb
ghost: required by rubocop

app/controllers/merge_requests_controller.rb
kosiakkatrina: Shouldn't we save the fact that it's a new organisation in the merge request? For example, if we navigate back if one of the orgs is selected that persists, but if it's new org it clears the answer

kosiakkatrina: Would also reduce the need for handling different post methods differently

app/views/merge_requests/absorbing_organisation.html.erb
```diff
@@ -0,0 +1,38 @@
+<% content_for :before_content do %>
+  <% title = "Tell us if your organisation is merging" %>
+  <% content_for :title, title %>
+  <%= govuk_back_link(href: :back) %>
```
kosiakkatrina: Do we want this back button to take us back in the form flow? For example, currently if we're on `/merge-request/:id/confirm-telephone-number` and keep on pressing back, we get stuck in a loop

app/controllers/merge_requests_controller.rb
ghost: I am not sure I understand; to what endpoint or controller action are you referring to? When would the org be new in this flow?

kosiakkatrina: Oh sorry, so on the question`/merge-request/:id/absorbing-organisation` if we select `These organisations are merging into a new one` it doesn't save it, so if we navigate back to this page, nothing is selected.

Also, on the same page, if we select an organisation, but then change it to `These organisations are merging into a new one` that won't be reflected either. If we navigate back to absorbing organisation page it had the previously chosen organisation selected, I assume that's because we don't save this answer `These organisations are merging into a new one`?

app/views/merge_requests/confirm_telephone_number.html.erb
```diff
@@ -1,7 +1,7 @@
 <% content_for :before_content do %>
   <% title = "Tell us if your organisation is merging" %>
   <% content_for :title, title %>
-  <%= govuk_back_link(href: :back) %>
+  <%= govuk_back_link href: absorbing_organisation_merge_request_path(id: @merge_request.requesting_organisation_id) %>
```
kosiakkatrina: Should this point to the merge_request id instead of the requesting_organisation_id `absorbing_organisation_merge_request_path(@merge_request)`? Do we have any tests for this?

ghost: good find, updating. Tests for what precisely? The back link on a template? This isn't something I'd be testing but if you think it's required I can do it in the follow up ticket that is about this page https://digital.dclg.gov.uk/jira/browse/CLDC-2057

app/controllers/merge_requests_controller.rb
ghost: Aah yes, I see now! I will have a think on how to do this

app/views/merge_requests/confirm_telephone_number.html.erb
ghost: Is this also a bug? I think I've ended up with those broken links by reusing it https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/1580/files#diff-ac707743834564e6ddb58b9dcebcfca7f9d6d564d2cbdffe04b1f8e771c3f8ddR4

kosiakkatrina: Hmm, I think this is routing to `/organisations/:id/merge-request` which is not related to a specific merge request, so should be fine. I see where it came from now!

app/controllers/merge_requests_controller.rb
ghost: @kosiakkatrina updated as you suggested and it did get rid of this method, good catch

kosiakkatrina (approved):

PR #1584, opened 2023-04-24
CLDC-2248 Improve soft validations

kosiakkatrina:
Update interruption screen design to display all the questions that affect the soft validation:
image

User can choose to confirm, skip or correct the values. If the values are changed the user is returned to the interruption screen and can change any other relevant values if needed.
image

We no longer display question headers, as we're no longer asking a question. Question headers would often be `Are you sure this is correct?` but some of them had more information in them, such as `Are you sure the property has been vacant for this long?`, messages such as these might need to be reworded.

### DIFF
--- a/app/controllers/form_controller.rb
+++ b/app/controllers/form_controller.rb
@@ -127,13 +127,13 @@ private
 
   def referrer_from_query
     referrer = request.headers["HTTP_REFERER"]
-    return nil unless referrer
+    return unless referrer
 
     query_params = URI.parse(referrer).query
-    return nil unless query_params
+    return unless query_params
 
     parsed_params = CGI.parse(query_params)
-    return nil unless parsed_params["referrer"]
+    return unless parsed_params["referrer"]
 
     parsed_params["referrer"][0]
   end
@@ -143,7 +143,7 @@ private
   end
 
   def previous_interruption_screen_referrer
-    params[@log.model_name.param_key]["interruption_page_referrer_type"]
+    params[@log.model_name.param_key]["interruption_page_referrer_type"].presence
   end
 
   def successful_redirect_path
@@ -159,11 +159,7 @@ private
       end
     end
     if previous_interruption_screen_page_id.present?
-      if previous_interruption_screen_referrer.present?
-        return send("#{@log.class.name.underscore}_#{previous_interruption_screen_page_id}_path", @log, referrer: previous_interruption_screen_referrer)
-      else
-        return send("#{@log.class.name.underscore}_#{previous_interruption_screen_page_id}_path", @log)
-      end
+      return send("#{@log.class.name.underscore}_#{previous_interruption_screen_page_id}_path", @log, { referrer: previous_interruption_screen_referrer }.compact)
     end
 
     redirect_path = form.next_page_redirect_path(@page, @log, current_user)
--- a/app/controllers/form_controller.rb
+++ b/app/controllers/form_controller.rb
@@ -48,6 +48,7 @@ class FormController < ApplicationController
   def show_page
     if request.params["referrer"] == "interruption_screen"
       @interruption_page_id = URI.parse(request.headers["HTTP_REFERER"]).path.split("/").last.underscore
+      @interruption_page_referrer_type = referrer_from_query
     end
 
     if @log
@@ -121,18 +122,28 @@ private
   end
 
   def is_referrer_type?(referrer_type)
+    referrer_from_query == referrer_type
+  end
+
+  def referrer_from_query
     referrer = request.headers["HTTP_REFERER"]
-    return false unless referrer
+    return nil unless referrer
 
     query_params = URI.parse(referrer).query
-    return false unless query_params
+    return nil unless query_params
 
     parsed_params = CGI.parse(query_params)
-    parsed_params["referrer"].present? && parsed_params["referrer"][0] == referrer_type
+    return nil unless parsed_params["referrer"]
+
+    parsed_params["referrer"][0]
   end
 
   def previous_interruption_screen_page_id
     params[@log.model_name.param_key]["interruption_page_id"]
+  end
+
+  def previous_interruption_screen_referrer
+    params[@log.model_name.param_key]["interruption_page_referrer_type"]
   end
 
   def successful_redirect_path
@@ -148,7 +159,11 @@ private
       end
     end
     if previous_interruption_screen_page_id.present?
-      return send("#{@log.class.name.underscore}_#{previous_interruption_screen_page_id}_path", @log)
+      if previous_interruption_screen_referrer.present?
+        return send("#{@log.class.name.underscore}_#{previous_interruption_screen_page_id}_path", @log, referrer: previous_interruption_screen_referrer)
+      else
+        return send("#{@log.class.name.underscore}_#{previous_interruption_screen_page_id}_path", @log)
+      end
     end
 
     redirect_path = form.next_page_redirect_path(@page, @log, current_user)
--- a/app/controllers/form_controller.rb
+++ b/app/controllers/form_controller.rb
@@ -47,7 +47,7 @@ class FormController < ApplicationController
 
   def show_page
     if request.params["referrer"] == "interruption_screen"
-      @interruption_page_id = request.headers["HTTP_REFERER"].split("/")[-1].split("?")[0].underscore
+      @interruption_page_id = URI.parse(request.headers["HTTP_REFERER"]).path.split("/").last.underscore
     end
 
     if @log
--- a/app/controllers/form_controller.rb
+++ b/app/controllers/form_controller.rb
@@ -11,7 +11,7 @@ class FormController < ApplicationController
       mandatory_questions_with_no_response = mandatory_questions_with_no_response(responses_for_page)
 
       if mandatory_questions_with_no_response.empty? && @log.update(responses_for_page.merge(updated_by: current_user))
-        flash[:notice] = "You have successfully updated #{@page.questions.map(&:check_answer_label).join(', ')}" if interruption_screen_referrer.present?
+        flash[:notice] = "You have successfully updated #{@page.questions.map(&:check_answer_label).join(', ').downcase}" if interruption_screen_referrer.present?
         redirect_to(successful_redirect_path)
       else
         mandatory_questions_with_no_response.map do |question|
@@ -47,7 +47,7 @@ class FormController < ApplicationController
 
   def show_page
     if request.params["referrer"] == "interruption_screen"
-      @interruption_page_id = request.headers["HTTP_REFERER"].split("/")[-1].underscore
+      @interruption_page_id = request.headers["HTTP_REFERER"].split("/")[-1].split("?")[0].underscore
     end
 
     if @log
--- a/app/controllers/form_controller.rb
+++ b/app/controllers/form_controller.rb
@@ -11,7 +11,7 @@ class FormController < ApplicationController
       mandatory_questions_with_no_response = mandatory_questions_with_no_response(responses_for_page)
 
       if mandatory_questions_with_no_response.empty? && @log.update(responses_for_page.merge(updated_by: current_user))
-        flash[:notice] = "You have successfully updated #{@page.questions.map(&:check_answer_label).join(', ').downcase}" if previous_interruption_screen_page_id.present?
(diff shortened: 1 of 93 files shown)


app/views/form/_interruption_screen_question.html.erb
```diff
@@ -1,22 +1,57 @@
...
+        <% end %>
       <% end %>
     <% end %>
   <% end %>
-  <%= f.govuk_submit "Save and continue", accesskey: "s", class: "app-button--inverse govuk-!-margin-bottom-0" %>
 <% end %>
+
+<%= f.hidden_field question.id, value: "0" %>
+<div class="govuk-button-group">
+  <%= f.govuk_submit "Save and continue" %>
```
jamesrose: The designs show this as "Confirm and continue" - worth checking which this should be.

app/controllers/form_controller.rb
```diff
@@ -120,6 +121,11 @@ def is_referrer_check_answers?
     referrer.present? && CGI.parse(referrer.split("?")[-1]).present? && CGI.parse(referrer.split("?")[-1])["referrer"][0] == "check_answers"
   end
 
+  def interruprion_screen_referrer
+    referrer = request.headers["HTTP_REFERER"].presence || ""
+    return CGI.parse(referrer.split("?")[-1])["referrer"][0] if referrer.present? && CGI.parse(referrer.split("?")[-1]).present?
+  end
```
jamesrose: This is quite difficult to parse. What are you trying to do?

app/controllers/form_controller.rb
```diff
@@ -120,6 +121,11 @@ def is_referrer_check_answers?
     referrer.present? && CGI.parse(referrer.split("?")[-1]).present? && CGI.parse(referrer.split("?")[-1])["referrer"][0] == "check_answers"
   end
 
+  def interruprion_screen_referrer
```
jamesrose: Typo: `interruption`

app/controllers/form_controller.rb
```diff
@@ -132,6 +138,10 @@ def successful_redirect_path
         return send("#{@log.model_name.param_key}_#{form.subsection_for_page(@page).id}_check_answers_path", @log)
       end
     end
+    if interruprion_screen_referrer.present?
+      return send("#{@log.class.name.underscore}_#{interruprion_screen_referrer}_path", @log)
```
jamesrose: This is potentially dangerous. You're injecting a HTTP header into this method!

app/models/form/lettings/pages/no_females_pregnant_household_lead_age_value_check.rb
```diff
@@ -16,4 +16,8 @@ def initialize(id, hsh, subsection)
   def questions
     @questions ||= [Form::Lettings::Questions::PregnancyValueCheck.new(nil, nil, self)]
   end
+
+  def affected_question_ids
```
jamesrose: What does this represent? IMO the title of this method could be more descriptive.

app/controllers/form_controller.rb
kosiakkatrina: This is very similar to the method at the top (`is_referrer_check_answers?`):
When we click change we get sent to `lettings-logs/<id>/<question_that_is_being_changed>?referrer=<soft_validation_path>` 
So at this point `request.headers["HTTP_REFERER"]` could for example be `lettings-logs/1/pregnant?referrer=no_females_pregnant_household_value_check`
And `CGI.parse(referrer.split("?")[-1])["referrer"][0]` would end up being `no_females_pregnant_household_value_check`

app/models/form/question.rb
```diff
@@ -124,6 +124,10 @@ def action_href(log, page_id)
     "/#{log.model_name.param_key.dasherize}s/#{log.id}/#{page_id.to_s.dasherize}?referrer=check_answers"
   end
 
+  def interruption_action_href(log, page_id, current_page_id)
+    "/#{log.model_name.param_key.dasherize}s/#{log.id}/#{page_id.to_s.dasherize}?referrer=#{current_page_id}"
```
jamesrose: This (nor `action_href`) does not belong on the model. It should be a helper, and ideally use route helpers.

app/controllers/form_controller.rb
jamesrose: Why do we need to depend on referrer to do this routing?

app/models/form/lettings/pages/no_females_pregnant_household_lead_age_value_check.rb
kosiakkatrina: dependent_question_ids, linked_question_ids? Any other ideas?

kosiakkatrina: These are the ids of the questions that the validation depends on basically

app/controllers/form_controller.rb
kosiakkatrina: Because the interruption_screen page is no longer routed to after the relevant soft validation has been fixed. We don't really want to route to it, but we want to right after some of the relevant questions have been updated according to design

app/controllers/form_controller.rb
kosiakkatrina: How about this? [Change how we route back to interruption screen](https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/1584/commits/236a53853f7b2c9920979ff8da43d60f04ce0506)

app/models/form/lettings/pages/no_females_pregnant_household_lead_age_value_check.rb
arfacamble: `ids_of_questions_triggering_validation` if this is still a question

app/helpers/interruption_screen_helper.rb
```diff
@@ -29,6 +29,14 @@ def display_title_text(title_text, log)
     I18n.t(title_text["translation"], **translation_params).to_s
   end
 
+  def soft_validation_affected_questions(question, log)
+    question.page.affected_question_ids.map { |question_id| log.form.get_question(question_id, log) }.compact
+  end
+
+  def interruption_action_href(log, page_id)
+    send("#{log.model_name.param_key}_#{page_id}_path", log, referrer: "interruption_screen")
+  end
```
ghost, line 34: are these 2 methods covered by unit tests?

app/models/form/lettings/pages/no_females_pregnant_household_lead_age_value_check.rb
ghost: what about interruption_screen_question_ids

app/controllers/form_controller.rb
```diff
@@ -11,6 +11,7 @@ def submit_form
       mandatory_questions_with_no_response = mandatory_questions_with_no_response(responses_for_page)
 
       if mandatory_questions_with_no_response.empty? && @log.update(responses_for_page.merge(updated_by: current_user))
+        flash[:notice] = "You have successfully updated #{@page.questions.map(&:check_answer_label).join(', ').downcase}" if previous_interruption_screen_page_id.present?
```
ghost: this should probably be covered by tests

app/controllers/form_controller.rb
```diff
@@ -45,11 +46,15 @@ def review
...
+    if request.params["referrer"] == "interruption_screen"
+      @interruption_page_id = request.headers["HTTP_REFERER"].split("/")[-1].split("?")[0].underscore
+    end
+
     if @log
       page_id = request.path.split("/")[-1].underscore
       @page = form.get_page(page_id)
       @subsection = form.subsection_for_page(@page)
-      if @page.routed_to?(@log, current_user)
+      if @page.routed_to?(@log, current_user) || is_referrer_interruption_screen?
```
ghost: probably covered in `spec/features/form/validations_spec.rb` but would be nice to cover this behaviour in `spec/requests/form_controller_spec.rb`

app/controllers/form_controller.rb
```diff
@@ -132,6 +146,10 @@ def successful_redirect_path
         return send("#{@log.model_name.param_key}_#{form.subsection_for_page(@page).id}_check_answers_path", @log)
       end
     end
+    if previous_interruption_screen_page_id.present?
+      return send("#{@log.class.name.underscore}_#{previous_interruption_screen_page_id}_path", @log)
```
ghost: same as above

app/views/form/_interruption_screen_question.html.erb
```diff
@@ -1,22 +1,64 @@
...
   <% end %>
-  <%= f.govuk_submit "Save and continue", accesskey: "s", class: "app-button--inverse govuk-!-margin-bottom-0" %>
 <% end %>
+
+<h1 class="govuk-heading-m">
+  Make sure these answers are all correct
+</h1>
+<div class="x-govuk-summary-card govuk-!-margin-bottom-6">
+  <div class="x-govuk-summary-card__body">
+    <%= govuk_summary_list do |summary_list| %>
```
ghost: this view file is getting quite a bit complicated, why not extracting this into a testable view component?

app/controllers/form_controller.rb
```diff
@@ -45,11 +46,15 @@ def review
   end
 
   def show_page
+    if request.params["referrer"] == "interruption_screen"
+      @interruption_page_id = request.headers["HTTP_REFERER"].split("/")[-1].split("?")[0].underscore
```
arfacamble: I thought rails would have helpers that would go a bit further than this in terms of making it more readable
```suggestion
      @interruption_page_id = URI.parse(request.headers["HTTP_REFERER"]).path.split("/").last.underscore
```

app/helpers/interruption_screen_helper.rb
kosiakkatrina: Ahh good point, they are now

app/controllers/form_controller.rb
kosiakkatrina: I think all of the cases you mentioned now should be covered in request specs

app/views/form/_interruption_screen_question.html.erb
kosiakkatrina: Yes! I managed to reuse the check_answers_summary_list and extracted the banner

app/controllers/form_controller.rb
kosiakkatrina: Yup! Way nicer, thanks

spec/helpers/form_page_helper_spec.rb
```diff
@@ -0,0 +1,15 @@
+require "rails_helper"
+
+RSpec.describe FormPageHelper do
+  describe "#action_href" do
+    let(:lettings_log) { FactoryBot.create(:lettings_log) }
```
ghost, line 5: let's also test sales logs given the url would change?

ghost (approved):

jamesrose (approved):

app/controllers/form_controller.rb
```diff
@@ -115,13 +121,33 @@ def find_resource_by_named_id
...
+    return nil unless referrer
+
+    query_params = URI.parse(referrer).query
+    return nil unless query_params
+
+    parsed_params = CGI.parse(query_params)
+    return nil unless parsed_params["referrer"]
+
+    parsed_params["referrer"][0]
+  end
```
jamesrose:
```suggestion
  def referrer_from_query
    referrer = request.headers["HTTP_REFERER"]
    return unless referrer

    query_params = URI.parse(referrer).query
    return unless query_params

    parsed_params = CGI.parse(query_params)
    return unless parsed_params["referrer"]

    parsed_params["referrer"][0]
  end
```

app/controllers/form_controller.rb
```diff
@@ -132,6 +158,14 @@ def successful_redirect_path
         return send("#{@log.model_name.param_key}_#{form.subsection_for_page(@page).id}_check_answers_path", @log)
       end
     end
+    if previous_interruption_screen_page_id.present?
+      if previous_interruption_screen_referrer.present?
+        return send("#{@log.class.name.underscore}_#{previous_interruption_screen_page_id}_path", @log, referrer: previous_interruption_screen_referrer)
+      else
+        return send("#{@log.class.name.underscore}_#{previous_interruption_screen_page_id}_path", @log)
+      end
```
jamesrose: As this is the same route, you can compact nil values out:

```suggestion
     send("#{@log.class.name.underscore}_#{previous_interruption_screen_page_id}_path", @log, {referrer: previous_interruption_screen_referrer}.compact)
```

kosiakkatrina: Smart

PR #1590, opened 2023-04-26
CLDC-2301 Update bulk upload journey copy

asmega:
# Context

- https://digital.dclg.gov.uk/jira/browse/CLDC-2301
- Content for 2022 and 2023 bulk upload journey need to differ to give users the correct info they need to complete the journey

# Changes

- Update 2022 and 2023 `Prepare your file` and guidance pages

### DIFF
--- a/app/models/forms/bulk_upload_lettings/prepare_your_file.rb
+++ b/app/models/forms/bulk_upload_lettings/prepare_your_file.rb
@@ -12,7 +12,7 @@ module Forms
         case year
         when 2022
           "bulk_upload_lettings_logs/forms/prepare_your_file_2022"
-        when 2023
+        else
           "bulk_upload_lettings_logs/forms/prepare_your_file_2023"
         end
       end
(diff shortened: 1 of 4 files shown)


app/models/forms/bulk_upload_lettings/prepare_your_file.rb
```diff
@@ -9,7 +9,12 @@ class PrepareYourFile
       attribute :needstype, :integer
 
       def view_path
-        "bulk_upload_lettings_logs/forms/prepare_your_file"
+        case year
+        when 2022
+          "bulk_upload_lettings_logs/forms/prepare_your_file_2022"
+        when 2023
```
jamesrose: Is it worth having these fallback gracefully when year is not 2022 or 2023?

asmega: i can make it catch all to 2023. it should be noted unfortunately at this stage the form is state dependent. if year is missing the back button won't work. but your change would at least make new years compatible without a future change

jamesrose (approved):

PR #1591, opened 2023-04-26
CLDC-2294 Add authorization to bulk upload results

asmega:
# Context

- https://digital.dclg.gov.uk/jira/browse/CLDC-2294
- When bulk upload returns a results page it would be useful if colleagues of the uploader can see this page to help fix errors
- It would also be useful if support users can see these reports to help diagnose bulk upload errors

# Changes

- Added `pundit` gem to handle authorization
- Bulk upload results previously only accessible to the bulk uploader. Now they can be seen by users in the same org as the uploader and also support users

### DIFF
--- a/app/policies/bulk_upload_policy.rb
+++ b/app/policies/bulk_upload_policy.rb
@@ -1,0 +1,26 @@
+class BulkUploadPolicy
+  attr_reader :user, :bulk_upload
+
+  def initialize(user, bulk_upload)
+    @user = user
+    @bulk_upload = bulk_upload
+  end
+
+  def summary?
+    owner? || same_org? || user.support?
+  end
+
+  def show?
+    owner? || same_org? || user.support?
+  end
+
+private
+
+  def owner?
+    bulk_upload.user == user
+  end
+
+  def same_org?
+    bulk_upload.user.organisation.users.include?(user)
+  end
+end
--- a/spec/controllers/bulk_upload_lettings_results_controller_spec.rb
+++ b/spec/controllers/bulk_upload_lettings_results_controller_spec.rb
@@ -5,6 +5,19 @@ RSpec.describe BulkUploadLettingsResultsController do
     sign_in user
   end
 
+  describe "#show" do
+    let(:user) { create(:user) }
+    let(:bulk_upload) { create(:bulk_upload, :lettings, user:) }
+
+    it "passes thru pundit" do
+      allow(controller).to receive(:authorize)
+
+      get :show, params: { id: bulk_upload.id }
+
+      expect(controller).to have_received(:authorize)
+    end
+  end
+
   describe "GET #resume /lettings-logs/bulk-upload-results/:ID/resume" do
     let(:user) { create(:user) }
     let(:bulk_upload) { create(:bulk_upload, :lettings, user:) }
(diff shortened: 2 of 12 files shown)


spec/controllers/bulk_upload_lettings_results_controller_spec.rb
```diff
@@ -5,6 +5,19 @@
...
+    it "passes thru pundit" do
+      allow(controller).to receive(:authorize)
+
+      get :show, params: { id: bulk_upload.id }
+
+      expect(controller).to have_received(:authorize)
+    end
+  end
+
   describe "GET #resume /lettings-logs/bulk-upload-results/:ID/resume" do
```
SamSeed-Softwire, line 21: Do we want to add a test here for #summary as well?

app/policies/bulk_upload_policy.rb
```diff
@@ -0,0 +1,26 @@
+class BulkUploadPolicy
```
SamSeed-Softwire, line 1: It's probably not that important, but would it be worth creating an `ApplicationPolicy` class that this class inherits from?

spec/requests/bulk_upload_lettings_results_controller_spec.rb
SamSeed-Softwire, line 1: Do we want to write equivalent tests for sales too?

SamSeed-Softwire (approved):
Looks good to me! Just a few minor suggestions.

app/policies/bulk_upload_policy.rb
asmega: might need one in the future but don't really see what i would add to the base class if i added one right now?

spec/controllers/bulk_upload_lettings_results_controller_spec.rb
asmega: this should be covered by the additions to `spec/requests/bulk_upload_lettings_results_controller_spec.rb` even though its a bit asymmetrical

spec/requests/bulk_upload_lettings_results_controller_spec.rb
asmega: should be covered by `spec/controllers/bulk_upload_sales_results_controller_spec.rb`?

app/policies/bulk_upload_policy.rb
SamSeed-Softwire: Fair enough. I was wondering about whether we could write something more generic in an `ApplicationPolicy` class which the bulk upload stuff could be a special case of. Not sure whether that would make sense here - even if it did, it probably wouldn't be worth doing until we had other policies doing something similar to the bulk upload policy.

spec/requests/bulk_upload_lettings_results_controller_spec.rb
SamSeed-Softwire: Why would we have lettings and sales versions of the relevant tests in `spec/controllers` but only a lettings version of the relevant tests in `spec/requests` though?

asmega: it was just the organic way i wrote it. i wrote the request spec first to try to cover everything for lettings. when i moved on to sales i didn't really want to write full blown request specs as i think they're slow, so i opted just controller tests to ensure we delegated to the policy and the policy was working correctly

PR #1593, opened 2023-04-27
Refactor with split of log to CSV by log type

asmega:
# Context

- Partially related to https://digital.dclg.gov.uk/jira/browse/CLDC-2316
- Comprehending sales or lettings bulk upload CSV is enough, comprehending both simultaneously is rather challenging

# Changes

- Split out test helper class by log type ie lettings/sales


spec/support/bulk_upload/sales_log_to_csv.rb
```diff
@@ -0,0 +1,187 @@
...
+  def row_prefix
+    [nil] * col_offset
+  end
+
+  def to_2022_csv_row
+    (row_prefix + to_2022_row).flatten.join(",") + line_ending
+  end
+
+  def default_2022_field_numbers
+    (1..134).to_a
```
SamSeed-Softwire: Not relevant to this ticket but I think the number of fields for 2223 sales is either 125 or 128?

SamSeed-Softwire (approved):
Lgtm!

spec/support/bulk_upload/sales_log_to_csv.rb
asmega: ah good spot, you are correct it looks to be `125`, i will get that changed

asmega: that's done now

PR #1594, opened 2023-04-27
CLDC-2296 Display soft validation errors on bulk upload

kosiakkatrina:
Previously we didn't show soft validations as errors on bulk upload.
Now if any of the soft validations are routed to we add an error to the fields that cause those soft validations.
We get those fields using `interruption_screen_question_ids`. The full list of `interruption_screen_question_ids` for soft validation questions are this PR, so it will need to be merged first: https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/1584

I'll keep this PR with https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/1584 as the target and will rebase onto main when that's merged to make it easier to review.

### DIFF
--- a/app/services/bulk_upload/lettings/year2022/row_parser.rb
+++ b/app/services/bulk_upload/lettings/year2022/row_parser.rb
@@ -1,6 +1,7 @@
 class BulkUpload::Lettings::Year2022::RowParser
   include ActiveModel::Model
   include ActiveModel::Attributes
+  include InterruptionScreenHelper
 
   QUESTIONS = {
     field_1: "What is the letting type?",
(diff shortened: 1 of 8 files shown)


app/services/bulk_upload/lettings/year2022/row_parser.rb
```diff
@@ -1366,4 +1384,44 @@ def ethnic_group_from_ethnic
...
+    informative_text["arguments"].each do |argument|
+      value = get_value_from_argument(log, argument)
+      translation_params[argument["i18n_template"].to_sym] = value
+    end
+
+    begin
+      translation = I18n.t(informative_text["translation"], **translation_params)
+      translation.to_s.html_safe
+    rescue I18n::MissingInterpolationArgument => e
+      Rails.logger.error(e.message)
```
ghost: we might want to send this to Sentry rather than just logging it

app/services/bulk_upload/sales/year2022/row_parser.rb
```diff
@@ -944,4 +945,61 @@ def validate_relevant_collection_window
...
+          unless errors.any? { |e| e.options[:category] == :soft_validation && field_mapping_for_errors[interruption_screen_question_id.to_sym].include?(e.attribute) }
+            error_message = [display_title_text(question.page.title_text, log), display_informative_text(question.page.informative_text, log)].join(". ")
+            errors.add(field, message: error_message, category: :soft_validation)
+          end
+        end
+      end
+    end
+  end
+
+  def display_title_text(title_text, log)
```
ghost: too many lines in this file and the browser is misbehaving 🙈 , so I can't check myself. If `display_title_text` `display_informative_text` `get_value_from_argument ` are public methods then they should be unit tested

ghost (approved):

app/services/bulk_upload/lettings/year2022/row_parser.rb
kosiakkatrina: These were copied methods, so I ended up importing and reusing them from the helper

app/services/bulk_upload/sales/year2022/row_parser.rb
kosiakkatrina: Same as above!

app/services/bulk_upload/sales/year2022/row_parser.rb
```diff
@@ -944,4 +946,21 @@ def validate_relevant_collection_window
       errors.add(:field_4, I18n.t("validations.date.outside_collection_window"))
     end
   end
+
+  def validate_incomplete_soft_validations
```
jamesrose, line 950: Are these validations identical?

kosiakkatrina: Maybe. I'm following a pattern we're using for all the row parser validations, which is currently defining them individually for each year/type

PR #1595, opened 2023-04-27
CLDC-2057 Telephone number confirmation

ghost:
Add telephone confirmation question - if the org doesn't have a phone number set, just ask user for the number

If the org does not have a number set, ask the user to provide one
If the org has a phone number set, ask users to either confirm or provide a new one

### DIFF
--- a/app/controllers/merge_requests_controller.rb
+++ b/app/controllers/merge_requests_controller.rb
@@ -127,16 +127,20 @@ private
   def validate_response
     if page == "absorbing_organisation" && merge_request_params[:absorbing_organisation_id].blank? && merge_request_params[:new_absorbing_organisation].blank?
       @merge_request.errors.add(:absorbing_organisation_id, I18n.t("validations.merge_request.absorbing_organisation_blank"))
-      render previous_template
+      render previous_template, status: :unprocessable_entity
     end
 
     if page == "confirm_telephone_number"
       if merge_request_params[:telephone_number_correct].blank? && merge_request_params[:new_telephone_number].blank?
-        @merge_request.errors.add(:telephone_number_correct, I18n.t("validations.merge_request.telephone_number_correct_blank"))
-        render previous_template
+        if @merge_request.absorbing_organisation.phone.present?
+          @merge_request.errors.add(:telephone_number_correct, I18n.t("validations.merge_request.telephone_number_correct_blank"))
+        else
+          @merge_request.errors.add(:telephone_number_correct, I18n.t("validations.merge_request.new_telephone_number_blank"))
+        end
+        render previous_template, status: :unprocessable_entity
       elsif merge_request_params[:telephone_number_correct] == "0" && merge_request_params[:new_telephone_number].blank?
         @merge_request.errors.add(:new_telephone_number, I18n.t("validations.merge_request.new_telephone_number_blank"))
-        render previous_template
+        render previous_template, status: :unprocessable_entity
       end
     end
   end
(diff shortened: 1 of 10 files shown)


app/views/merge_requests/confirm_telephone_number.html.erb
```diff
@@ -8,6 +8,30 @@
...
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds-from-desktop">
-    <p class="govuk-body">Confirm the telephone number on file, or enter a new one.</p>
+    <% if @merge_request.absorbing_organisation.phone.present? %>
+      <p class="govuk-body">Confirm the telephone number on file, or enter a new one.</p>
+    <% end %>
+
+    <div class="govuk-inset-text">
+      <%= @merge_request.absorbing_organisation.phone %>
+    </div>
```
kosiakkatrina: Do we want to display this if the phone is empty? It has an empty inset text in that case

app/controllers/merge_requests_controller.rb
```diff
@@ -119,6 +129,16 @@ def validate_response
...
+
+    if page == "confirm_telephone_number"
+      if merge_request_params[:telephone_number_correct].blank? && merge_request_params[:new_telephone_number].blank?
+        @merge_request.errors.add(:telephone_number_correct, I18n.t("validations.merge_request.telephone_number_correct_blank"))
+        render previous_template
+      elsif merge_request_params[:telephone_number_correct] == "0" && merge_request_params[:new_telephone_number].blank?
+        @merge_request.errors.add(:new_telephone_number, I18n.t("validations.merge_request.new_telephone_number_blank"))
+        render previous_template
+      end
+    end
```
kosiakkatrina: Not sure if it's because of this or something else, but I always have `Enter a new phone number` preselected before I answer that question. Any ideas?

kosiakkatrina: Also, should a validation be a different one if there is no current number and we only ask them to enter a new one?
Cause `Select to confirm or enter a new telephone number` sounds off if the user can't confirm anything 🤔

app/views/merge_requests/confirm_telephone_number.html.erb
ghost: should be hidden! will fix

app/views/merge_requests/confirm_telephone_number.html.erb
```diff
@@ -10,19 +10,18 @@
...
+        <%= @merge_request.absorbing_organisation.phone %>
+      </div>
     <% end %>
 
-    <div class="govuk-inset-text">
-      <%= @merge_request.absorbing_organisation.phone %>
-    </div>
-
     <%= form_with model: @merge_request, url: merge_request_path, method: :patch do |f| %>
       <%= f.govuk_error_summary %>
```
kosiakkatrina: I just realised that the error should probably be above the question header, it's wrong on all pages 😭

app/controllers/merge_requests_controller.rb
```diff
@@ -132,7 +132,11 @@ def validate_response
 
     if page == "confirm_telephone_number"
       if merge_request_params[:telephone_number_correct].blank? && merge_request_params[:new_telephone_number].blank?
-        @merge_request.errors.add(:telephone_number_correct, I18n.t("validations.merge_request.telephone_number_correct_blank"))
+        if @merge_request.absorbing_organisation.phone.present?
+          @merge_request.errors.add(:telephone_number_correct, I18n.t("validations.merge_request.telephone_number_correct_blank"))
+        else
+          @merge_request.errors.add(:telephone_number_correct, I18n.t("validations.merge_request.new_telephone_number_blank"))
+        end
         render previous_template
```
kosiakkatrina: `render previous_template, status: :unprocessable_entity` ? same for the one below?

kosiakkatrina (approved):

PR #1598, opened 2023-05-02
CLDC-2058 Add merging organisation name question

kosiakkatrina:
Add a new organisation name question to the merge organisations flow:
image

### DIFF
--- a/app/controllers/merge_requests_controller.rb
+++ b/app/controllers/merge_requests_controller.rb
@@ -123,7 +123,7 @@ private
       end
     end
 
-    if merge_params[:telephone_number_correct] == "1"
+    if merge_params[:telephone_number_correct] == "true"
       merge_params[:new_telephone_number] = nil
     end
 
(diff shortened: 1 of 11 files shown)


app/controllers/merge_requests_controller.rb
```diff
@@ -125,22 +131,27 @@ def merge_request_params
...
-    if page == "absorbing_organisation" && merge_request_params[:absorbing_organisation_id].blank? && merge_request_params[:new_absorbing_organisation].blank?
-      @merge_request.errors.add(:absorbing_organisation_id, I18n.t("validations.merge_request.absorbing_organisation_blank"))
-      render previous_template, status: :unprocessable_entity
-    end
+    case page
 
-    if page == "confirm_telephone_number"
-      if merge_request_params[:telephone_number_correct].blank? && merge_request_params[:new_telephone_number].blank?
+    when "absorbing_organisation"
+      if merge_request_params[:absorbing_organisation_id].blank? && merge_request_params[:new_absorbing_organisation].blank?
```
asmega, line 136: can we not somehow utilise `ActiveModel::Errors` rather writing validations out by hand?

kosiakkatrina: Is this [Refactor error messages](https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/1598/commits/98159536408d341dcfe3672db4392bc5b839a32a) at all what you meant?

asmega: for me if we have to keep adding `if` statements the abstraction is either wrong or non-existent. would we not want to be writing code like:
```
  validates :new_organisation_name, presence: true
  validate :validate_custom_exist_check
```
to handle validations which are the apis available in `ActiveModel::Errors`

kosiakkatrina: Yeah, gotcha, so I've extracted some of them onto the model, the ones that are left depend on the pages quite a bit, so not sure how to move those

asmega (approved):

PR #1602, opened 2023-05-04
CLDC-2060 Add new organisation address question

kosiakkatrina:
Add a new organisation address question to the merge request flow. There are no validations and there's a skip button as this is not a mandatory question atm:
image


jamesrose (approved):
LGTM

PR #1610, opened 2023-05-09
Switch to original devise gem

asmega:
# Context

- We're currently using a custom fork of devise

# Changes

- Switch back to original devise gem
- This gives us the benefit of being able to more easily upgrade when the time comes
- Minor fixes to validation copy during password reset process which were not correct

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -33,7 +33,7 @@ gem "roo"
 gem "json-schema"
 # Authentication
 # Point at branch until devise is compatible with Turbo, see https://github.com/heartcombo/devise/pull/5340
-gem "devise", github: "baarkerlounger/devise", branch: "dluhc-fixes"
+gem "devise"
 # Two-factor Authentication for devise models.
 gem "devise_two_factor_authentication"
 # UK postcode parsing and validation
--- a/Gemfile
+++ b/Gemfile
@@ -32,7 +32,6 @@ gem "roo"
 # Json Schema
 gem "json-schema"
 # Authentication
-# Point at branch until devise is compatible with Turbo, see https://github.com/heartcombo/devise/pull/5340
 gem "devise"
 # Two-factor Authentication for devise models.
 gem "devise_two_factor_authentication"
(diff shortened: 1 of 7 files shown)


baarkerlounger:
You ideally want this merged upstream before you switch this: https://github.com/heartcombo/devise/pull/5454

Though whether it's important enough to continue pointing at a fork :man_shrugging:

asmega:
@baarkerlounger I've not figured out the integration point for this work yet. If one were not using stimulus is there any reason to use the fork?

baarkerlounger:
There were originally two reasons for pointing to the fork

1. Fixes for Turbo which can be seen [here](https://github.com/heartcombo/devise/compare/main...baarkerlounger:devise:dluhc-fixes). Upstream now supports Turbo and CORE doesn't use it anymore anyway so this shouldn't be needed anymore.

2. Copy change in the PR above (incorrect capitalisation of some error messages) - see the PR [here](https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/211). I think this is still needed but it's also doubtful it's important enough to continue pointing at a fork now the functionality changes aren't needed anymore.

asmega:
So I think we are fine without turbo/stimulus support. It's still plumbed in but not used in most places. RE the copy change theres a mechanism thats already exists that you can shim in via i18n which I've now added which should cater for the original need. Which means hopefully we can now track upstream without any issues

Gemfile
```diff
@@ -33,7 +33,7 @@ gem "roo"
 gem "json-schema"
 # Authentication
 # Point at branch until devise is compatible with Turbo, see https://github.com/heartcombo/devise/pull/5340
```
baarkerlounger: Can probably remove this comment now too

spec/models/user_spec.rb
```diff
@@ -268,7 +268,7 @@
 
     context "when a too short password is entered" do
       let(:password) { "123" }
-      let(:error_message) { "Validation failed: Password #{I18n.t('errors.messages.too_short', count: 8)}" }
+      let(:error_message) { "Validation failed: Password #{I18n.t('activerecord.errors.models.user.attributes.password.too_short', count: 8)}" }
```
asmega, line 271: these tests are kind of useless IMO, we never actually surface this way

asmega: but I've kept to provide minimum touch for this change

kosiakkatrina (approved):

PR #1612, opened 2023-05-10
CLDC-2304 Deduplicate sales logs in bulk upload

kosiakkatrina:
Replicate the functionality from lettings logs for deduplicating bulk uploaded logs.

### DIFF
--- a/app/services/bulk_upload/sales/year2022/row_parser.rb
+++ b/app/services/bulk_upload/sales/year2022/row_parser.rb
@@ -338,6 +338,12 @@ class BulkUpload::Sales::Year2022::RowParser
     block_log_creation
   end
 
+  def log_already_exists?
+    @log_already_exists ||= SalesLog
+      .where(status: %w[not_started in_progress completed])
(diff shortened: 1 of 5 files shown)


app/services/bulk_upload/sales/year2022/row_parser.rb
```diff
@@ -337,6 +338,12 @@ def block_log_creation?
     block_log_creation
   end
 
+  def log_already_exists?
+    @log_already_exists ||= SalesLog
+      .where(status: %w[not_started in_progress completed])
```
natdeanlewissoftwire, line 343: what's this line checking? this is all the possible statuses right?

natdeanlewissoftwire (approved):
just one question but lgtm :)

app/services/bulk_upload/sales/year2022/row_parser.rb
kosiakkatrina: I think it's checking for "pending" which is created by bulk upload, but the user hasn't yet selected how to fix it

PR #1615, opened 2023-05-10
CLDC-2334 Add journey to fix soft validation errors only

kosiakkatrina:
When only soft validations are getting triggered on bulk upload, the user gets an email with a link to this page:
image
Here they can choose to confirm all of the soft validations and create the logs, or choose to fix the fields that are triggering those validations. In which case they get put onto the fix-choice journey where the soft validation errors are treated as normal errors.

### DIFF
--- a/app/controllers/bulk_upload_lettings_data_check_controller.rb
+++ b/app/controllers/bulk_upload_lettings_data_check_controller.rb
@@ -1,0 +1,36 @@
+class BulkUploadLettingsDataCheckController < ApplicationController
+  before_action :authenticate_user!
+
+  def show
+    @bulk_upload = current_user.bulk_uploads.find(params[:id])
+
+    render form.view_path
+  end
+
+  def update
+    @bulk_upload = current_user.bulk_uploads.find(params[:id])
+
+    if form.valid? && form.save!
+      redirect_to form.next_path
+    else
+      render form.view_path
+    end
+  end
+
+private
+
+  def form
+    @form ||= case params[:page]
+              when "soft-errors-valid"
+                Forms::BulkUploadLettingsDataCheck::SoftErrorsValid.new(form_params.merge(bulk_upload: @bulk_upload))
+              when "confirm"
+                Forms::BulkUploadLettingsDataCheck::Confirm.new(form_params.merge(bulk_upload: @bulk_upload))
+              else
+                raise "invalid form"
+              end
+  end
+
+  def form_params
+    params.fetch(:form, {}).permit(:soft_errors_valid)
+  end
+end
--- a/app/controllers/bulk_upload_lettings_data_check_controller.rb
+++ b/app/controllers/bulk_upload_lettings_data_check_controller.rb
@@ -1,4 +1,6 @@
 class BulkUploadLettingsDataCheckController < ApplicationController
+  include ActionView::Helpers::TextHelper
+
   before_action :authenticate_user!
 
   def show
@@ -11,6 +13,11 @@ class BulkUploadLettingsDataCheckController < ApplicationController
     @bulk_upload = current_user.bulk_uploads.find(params[:id])
 
     if form.valid? && form.save!
+      if params[:page] == "confirm"
+        n_logs = pluralize(@bulk_upload.logs.count, "log")
+        flash[:notice] = "You’ve successfully uploaded #{n_logs}"
+      end
+
       redirect_to form.next_path
     else
       render form.view_path
--- a/app/mailers/bulk_upload_mailer.rb
+++ b/app/mailers/bulk_upload_mailer.rb
@@ -29,7 +29,7 @@ class BulkUploadMailer < NotifyMailer
   def send_check_soft_validations_mail(bulk_upload:)
     title = "Check your file data"
     description = "Some of your #{bulk_upload.year_combo} #{bulk_upload.log_type} data might not be right. Click the link below to review the potential errors, and check your file to see if the data is correct."
-    cta_link = bulk_upload.sales? ? bulk_upload_sales_data_check_url(bulk_upload) : bulk_upload_lettings_data_check_url(bulk_upload)
+    cta_link = bulk_upload.sales? ? bulk_upload_sales_data_check_url(bulk_upload, page: "soft-errors-valid") : bulk_upload_lettings_data_check_url(bulk_upload, page: "soft-errors-valid")
 
     send_email(
       bulk_upload.user.email,
--- a/app/mailers/bulk_upload_mailer.rb
+++ b/app/mailers/bulk_upload_mailer.rb
@@ -5,17 +5,17 @@ class BulkUploadMailer < NotifyMailer
   FAILED_CSV_ERRORS_TEMPLATE_ID = "e27abcd4-5295-48c2-b127-e9ee4b781b75".freeze
   FAILED_FILE_SETUP_ERROR_TEMPLATE_ID = "24c9f4c7-96ad-470a-ba31-eb51b7cbafd9".freeze
   FAILED_SERVICE_ERROR_TEMPLATE_ID = "c3f6288c-7a74-4e77-99ee-6c4a0f6e125a".freeze
-  HOW_FIX_UPLOAD_TEMPLATE_ID = "21a07b26-f625-4846-9f4d-39e30937aa24".freeze
+  HOW_TO_FIX_UPLOAD_TEMPLATE_ID = "21a07b26-f625-4846-9f4d-39e30937aa24".freeze
   CHECK_SOFT_VALIDATIONS_TEMPLATE_ID = "21a07b26-f625-4846-9f4d-39e30937aa24".freeze
 
-  def send_how_fix_upload_mail(bulk_upload:)
+  def send_how_to_fix_upload_mail(bulk_upload:)
     title = "We found #{pluralize(bulk_upload.bulk_upload_errors.count, 'error')} in your bulk upload"
     description = "There was a problem with your #{bulk_upload.year_combo} #{bulk_upload.log_type} data. Check the error report below to fix these errors."
     cta_link = bulk_upload.sales? ? start_bulk_upload_sales_resume_url(bulk_upload) : start_bulk_upload_lettings_resume_url(bulk_upload)
 
     send_email(
       bulk_upload.user.email,
-      HOW_FIX_UPLOAD_TEMPLATE_ID,
+      HOW_TO_FIX_UPLOAD_TEMPLATE_ID,
       {
         title:,
         filename: bulk_upload.filename,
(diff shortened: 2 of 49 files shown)


app/controllers/bulk_upload_lettings_data_check_controller.rb
natdeanlewissoftwire, line 1: data check to me seems like a vague name - if this is just to do with checking soft validations could we call it`bulk_upload_lettings_check_soft_validations_controller` or similar (same for sales as well ofc)? Up to you, I'm not tied to it

app/models/bulk_upload.rb
```diff
@@ -72,6 +72,19 @@ def unpend
...
 
+  def unpend_and_confirm_soft_validations
+    logs.find_each do |log|
+      log.retirement_value_check = 0
+      log.pregnancy_value_check = 0
+      log.major_repairs_date_value_check = 0
+      log.void_date_value_check = 0
+      log.rent_value_check = 0
+      log.net_income_value_check = 0
+      log.carehome_charges_value_check = 0
```
natdeanlewissoftwire: feels like there are more soft validations than this but I'm sure this is correct, kind of nice to know there are only 7

app/models/forms/bulk_upload_lettings_data_check/confirm.rb
natdeanlewissoftwire, line 1: same naming suggestion as before if you want it :)

app/views/bulk_upload_lettings_data_check/confirm.html.erb
```diff
@@ -0,0 +1,22 @@
...
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <span class="govuk-caption-l">Bulk upload for lettings (<%= @bulk_upload.year_combo %>)</span>
+    <h1 class="govuk-heading-l">Are you sure you want to upload all logs from this bulk upload?</h1>
+
+    <p class="govuk-body">There are <%= pluralize(@bulk_upload.logs.count, "log") %> in this bulk upload, and <%= pluralize(@bulk_upload.bulk_upload_errors.count, "unexpected answer") %> will be marked as correct.</p>
+
+    <%= govuk_warning_text(icon_fallback_text: "Danger") do %>
+      You can not delete logs once you create them
```
natdeanlewissoftwire: copy question - I think we're favouring cannot over can not, there's one other place I can see a can not in the codebase which could probably be changed too

config/locales/en.yml
```diff
@@ -73,6 +73,14 @@ en:
             choice:
               blank: You must select how would you like to fix errors
               inclusion: You must select one of the following options for how would like to fix errors
+        forms/bulk_upload_lettings_data_check/soft_errors_valid:
```
natdeanlewissoftwire: could we share this copy rather than duplicating?

app/mailers/bulk_upload_mailer.rb
```diff
@@ -6,6 +6,7 @@ class BulkUploadMailer < NotifyMailer
   FAILED_FILE_SETUP_ERROR_TEMPLATE_ID = "24c9f4c7-96ad-470a-ba31-eb51b7cbafd9".freeze
   FAILED_SERVICE_ERROR_TEMPLATE_ID = "c3f6288c-7a74-4e77-99ee-6c4a0f6e125a".freeze
   HOW_FIX_UPLOAD_TEMPLATE_ID = "21a07b26-f625-4846-9f4d-39e30937aa24".freeze
```
natdeanlewissoftwire: not related to your PR but I've just remembered how much it annoys me this isn't called how_to_fix... everywhere it occurs, if you agree I'd support changing it but happy to do that myself when I'm next on bulk upload

app/mailers/bulk_upload_mailer.rb
```diff
@@ -25,6 +26,24 @@ def send_how_fix_upload_mail(bulk_upload:)
     )
   end
 
+  def send_check_soft_validations_mail(bulk_upload:)
+    title = "Check your file data"
+    description = "Some of your #{bulk_upload.year_combo} #{bulk_upload.log_type} data might not be right. Click the link below to review the potential errors, and check your file to see if the data is correct."
+    cta_link = bulk_upload.sales? ? bulk_upload_sales_data_check_url(bulk_upload, page: "soft-errors-valid") : bulk_upload_lettings_data_check_url(bulk_upload, page: "soft-errors-valid")
```
natdeanlewissoftwire: I _slightly_ prefer doing this sort of thing with a switch statement to be more readable that it's sales or lettings rather than sales or anything else but I accept that should make no practical difference

also sales before lettings is not the conventional order

these are both VERY minor points haha I'm scraping the barrel

app/models/forms/bulk_upload_lettings_data_check/soft_errors_valid.rb
```diff
@@ -0,0 +1,40 @@
...
+      include Rails.application.routes.url_helpers
+
+      attribute :bulk_upload
+      attribute :soft_errors_valid, :string
+
+      validates :soft_errors_valid, presence: true
+
+      def options
+        [
+          OpenStruct.new(id: "yes", name: "Yes, some of these are errors"),
```
natdeanlewissoftwire: I don't love yes = errors, no = correct, seems at odds with the "are you sure these values are correct" format we use in the soft validations themselves

natdeanlewissoftwire (approved):
nice work, looks very clean, all my comments are stylistic/subjective/inconsequential/all of the above

app/models/bulk_upload.rb
kosiakkatrina: There are, this is just lettings. Good spot!

app/controllers/bulk_upload_lettings_data_check_controller.rb
kosiakkatrina: I renamed it to `bulk_upload_lettings_soft_validations_check_controller` cause I wanted to keep `validations` plural in paths and rails had different ideas, does it look any better?

app/views/bulk_upload_lettings_data_check/confirm.html.erb
kosiakkatrina: Hm, it was in the designs, let me check

config/locales/en.yml
kosiakkatrina: Hmm, I think it makes it easier to call the validation, cause we wouldn't have to define the message in the form. For example calling `validates :soft_errors_valid, presence: true` from soft_errors_valid.rb form is enough to get this message without having to specify the path. I think I prefer that cause it makes the actual form a bit cleaner, but let me know if you disagree

app/models/forms/bulk_upload_lettings_data_check/soft_errors_valid.rb
kosiakkatrina: Hmm, do you mean the content should be different, or how we store it on our side?

app/mailers/bulk_upload_mailer.rb
kosiakkatrina: I've changed the check to lettings and put it onto multiline if statement, cause that's consistent with the rest of this file

app/controllers/bulk_upload_lettings_data_check_controller.rb
natdeanlewissoftwire: Yeah, love it 🥳

config/locales/en.yml
natdeanlewissoftwire: ah ok, I see now, happy with this as is

app/mailers/bulk_upload_mailer.rb
natdeanlewissoftwire: sure, sounds good

app/models/forms/bulk_upload_lettings_data_check/soft_errors_valid.rb
natdeanlewissoftwire: I mean the content should be different - if Susanna's already made it though then obviously she knows best! Can't remember if she had a hand in the content for this bit or not

kosiakkatrina: Okay, I've asked about this as well. Can possibly change later if need to

config/locales/en.yml
```diff
@@ -71,20 +71,20 @@ en:
...
+        forms/bulk_upload_sales_soft_validations_check/confirm_soft_errors:
           attributes:
-            soft_errors_valid:
+            confirm_soft_errors:
               blank: You must select if there are errors in these fields
         forms/bulk_upload_sales_resume/fix_choice:
             attributes:
               choice:
-                blank: You must select how would you like to fix errors
+                blank: Select how would you like to fix these errors
```
natdeanlewissoftwire: I've just seen this, but this should be how **you would** not **would you** - missed this on my PR too, have just made that change there myself

config/locales/en.yml
```diff
@@ -71,20 +71,20 @@ en:
...
                 inclusion: You must select one of the following options for how would like to fix errors
```
natdeanlewissoftwire: also this should say "how you would" too!

natdeanlewissoftwire: and should say "these errors"!!

spec/requests/bulk_upload_lettings_resume_controller_spec.rb
```diff
@@ -49,7 +49,7 @@
 
         expect(response).to be_successful
 
-        expect(response.body).to include("You must select")
+        expect(response.body).to include("Select how would you like to fix these errors")
```
natdeanlewissoftwire: this test has the same typo as above :)

spec/requests/bulk_upload_sales_resume_controller_spec.rb
```diff
@@ -49,7 +49,7 @@
 
         expect(response).to be_successful
 
-        expect(response.body).to include("You must select")
+        expect(response.body).to include("Select how would you like to fix these errors")
```
natdeanlewissoftwire: typo

config/locales/en.yml
```diff
@@ -71,20 +71,20 @@ en:
         forms/bulk_upload_lettings_resume/fix_choice:
           attributes:
             choice:
-              blank: You must select how would you like to fix errors
+              blank: Select how would you like to fix these errors
```
natdeanlewissoftwire: same typo

config/locales/en.yml
```diff
@@ -71,20 +71,20 @@ en:
...
               inclusion: You must select one of the following options for how would like to fix errors
```
natdeanlewissoftwire: same other typo

config/locales/en.yml
```diff
@@ -81,6 +81,11 @@ en:
           attributes:
             soft_errors_valid:
               blank: You must select if there are errors in these fields
+        forms/bulk_upload_sales_resume/fix_choice:
```
natdeanlewissoftwire, line 84: same typo but I've addressed this in CLDC-2357

PR #1617, opened 2023-05-11
CLDC-2361 Extract bulk upload filter from session

asmega:
# Context

- https://digital.dclg.gov.uk/jira/browse/CLDC-2361
- When fixing inline for sales bulk uploads users need to be able to show only logs related to the bulk upload therefore have the bulk upload filter applied

# Changes

- After a sales bulk upload the session is set with bulk upload filter
- When viewing sales log ensure we extract this apply as a filter

# Issues

- The current implementation is globally shared between lettings and sales therefore there are issues when switching between the 2 areas
- There is a backlog ticket to refactor the filter and recommend fixing this issue then

### DIFF
--- a/app/controllers/sales_logs_controller.rb
+++ b/app/controllers/sales_logs_controller.rb
@@ -70,7 +70,7 @@ private
   end
 
   def redirect_if_bulk_upload_resolved
-    if @bulk_upload && @bulk_upload.sales? && @bulk_upload.sales_logs.in_progress.count.zero?
+    if @bulk_upload&.sales? && @bulk_upload.sales_logs.in_progress.count.zero?
       redirect_to resume_bulk_upload_sales_result_path(@bulk_upload)
     end
   end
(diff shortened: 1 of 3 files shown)


app/controllers/sales_logs_controller.rb
```diff
@@ -61,6 +64,17 @@ def permitted_log_params
 
 private
 
+  def extract_bulk_upload_from_session_filters
+    filter_service = FilterService.new(current_user:, session:)
+    @bulk_upload = filter_service.bulk_upload
+  end
+
+  def redirect_if_bulk_upload_resolved
+    if @bulk_upload && @bulk_upload.sales? && @bulk_upload.sales_logs.in_progress.count.zero?
```
natdeanlewissoftwire: We could replace the first two conditions with `@bulk_upload&.sales?`

natdeanlewissoftwire: or even just replace the whole line with `@bulk_upload&.sales_logs.in_progress.count.zero?`  I think

natdeanlewissoftwire (approved):
lgtm, just one suggestion for readability

app/controllers/sales_logs_controller.rb
asmega: i think we can change to the first one. second one i think will have a bug cause of the nature that bulk upload can be lettings or sales and we have yet to do work to split/scope filters out for different log types

PR #1620, opened 2023-05-12
CLDC 2286 Delete logs

ghost:
Add delete log flow:
- add new confirmation page
- add new delete button on log page
- add deleted log status
- add `discarded_at` column

Screenshot 2023-05-12 at 15 11 49


app/controllers/lettings_logs_controller.rb
```diff
@@ -68,17 +68,27 @@ def edit
...
-      if @log.delete
-        head :no_content
-      else
-        render json: { errors: @log.errors.messages }, status: :unprocessable_entity
-      end
+    render_not_found and return unless @log
+
+    authorize @log, policy_class: LogPolicy
+
+    if @log.delete
```
jamesrose: Just a friendly note that I'd expect this to change to a state-change vs. permanent delete.

ghost: thanks James!

app/policies/log_policy.rb
```diff
@@ -0,0 +1,24 @@
...
+    return false unless log && user
+
+    # Can only delete editable logs
+    return false unless log.collection_period_open?
+
+    # Only delete logs with answered questions
+    return false unless log.in_progress? || log.completed?
+
+    # Data coordinators and support users can delete any log
+    return true if user.data_coordinator? || user.support?
```
asmega: theres no scoping here or at controller level on the resource. which means as a data coordinator I could enumerate the IDs of every log in the system and delete them all

app/controllers/sales_logs_controller.rb
```diff
@@ -40,6 +40,28 @@ def edit
     end
   end
 
+  def destroy
+    @log = SalesLog.visible.find_by(id: params[:id])
+
+    render_not_found and return unless @log
```
asmega: to prevent from littering every controller method with this line you can use `rescue_from` combined with `ActiveRecord::RecordNotFound` to handle these at a higher level

app/controllers/lettings_logs_controller.rb
```diff
@@ -1,5 +1,5 @@
 class LettingsLogsController < LogsController
-  before_action :find_resource, except: %i[create index edit]
+  before_action :find_resource, except: %i[create index edit delete_confirmation]
```
asmega: maybe this should be whitelisted rather than blacklisted?

app/controllers/sales_logs_controller.rb
```diff
@@ -40,6 +40,28 @@ def edit
...
+
+    authorize @log, policy_class: LogPolicy
```
asmega: if you create the empty classes:
-  `LettingsLogPolicy < LogPolicy`
-  `SalesLogPolicy < LogPolicy`

you won't have to explicitly set the `policy_class` every time

app/policies/log_policy.rb
ghost: the AC is "Data providers can only see this button if the log is assigned to them, even if it belongs to a parent org. Data coordinators and support users can see this button on any log.", I will double check if this isn't correct

jamesrose: This line as it is would allow any data coordinator of any organisation to delete a log. I suggest we scope this to the relevant organisation, but also update this finder in the controller so that it can only find relevant logs:

```ruby
    @log = LettingsLog.visible.find(params[:id])
```

ghost: I'm addressing the scope now that I've confirmed that the AC was incorrect. 

@jamesrose  Can you point where you see the missing `@log = LettingsLog.visible.find(params[:id])`? I think all the new controller actions are already using that scope.

jamesrose: @bibblobcode Sorry, I was suggesting that those lines in the controller action might want to change, so that you can be defensive against enumerating IDs or permission failures.

ghost: I would rather not change the behaviour of  the other controller actions in this PR because I _think_ most of them aren't even used and at that point I would prefer deleting them. 

I could log a tech debt ticket, what do you think?

asmega (approved):
~I've approved with the caveat that this change introduces a vulnerability that a data co-ordinator is capable of enumerating all logs and delete them all~

ghost:
@asmega  I am struggling to see that vulnerability, can you point to it?

The policies prevent it:
- https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/1620/files#diff-af4994402cd71f1ff6094b705f060bea5e7229012e25d629eb9ae0a1a68ca103R22
- https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/1620/files#diff-4bd74a981d6c0d9f2abebef7aff7a681176006ae7cfa7c0a7a13727586b97e2fR22

Both new controller actions check whether the user is authorised to either delete or access the delete-confirmation page. If not the user is redirected to the error page.

What have I missed?

app/policies/lettings_log_policy.rb
natdeanlewissoftwire, line 1: do these need to be separate files for sales and lettings? I'm imagining the conditions will always be the same for deletion

natdeanlewissoftwire, line 1: do these need to be separate files for sales and lettings? I'm imagining the conditions will always be the same for deletion

asmega:
Apologies i retract my previous comment, I was only looking at the controller and did not see this change, sorry. Looks good now!

app/controllers/sales_logs_controller.rb
```diff
@@ -32,12 +34,26 @@ def show
...
-    if @log
-      render "logs/edit", locals: { current_user: }
-    else
-      render_not_found
-    end
+    @log = current_user.sales_logs.visible.find(params[:id])
+    render "logs/edit", locals: { current_user: }
+  end
+
+  def destroy
```
natdeanlewissoftwire, line 41: if we're actually discarding, shouldn't we call this method and the one in log policy discard too?

app/services/csv/lettings_log_csv_service.rb
natdeanlewissoftwire, line 1: is there another ticket/tech debt to add this to the `sales_log_csv_service` too? Same for export service

natdeanlewissoftwire (approved):
couple of thoughts, but looks good

app/policies/lettings_log_policy.rb
natdeanlewissoftwire: actually feel free to ignore this unless there's an easy way to do it, just seen we do actually check specifically for a type of logs on line 22

natdeanlewissoftwire: although it's probably not necessary to do that, could just test for lettings OR sales

ghost: I had a single one initially but then I added a check on `user.letting_logs` and one on `user.sales_logs` and had to find a way to separate them. also the controller will try to automatically pull the relevant policy based on the model name

app/services/csv/lettings_log_csv_service.rb
ghost: the `sales_log_csv_service` specs didn't error, so that makes me believe the discarded at field isn't being exported and I think we've happy with that... same for the export service

app/controllers/sales_logs_controller.rb
ghost: I went for this because:
- trying to keep this endpoint RESTful
- I want to make it clear that we don't want to really `destroy`, and want to prevent this method/action from being used  for a destructive action

PR #1621, opened 2023-05-12
CLDC-2344 Remove bulk upload minimum column check

asmega:
# Context

- https://digital.dclg.gov.uk/jira/browse/CLDC-2344
- So the reported bug and these change do not match 100%
- As the reported bug should have already been resolved in an earlier change as the error message that was reported has already been removed from the system
- These changes remove the check from sales plus a few minor tidy ups

# Changes

- Remove dead code around error threshold which is no longer used or needed
- Inverted negative predicate to positive one, `incorrect_field_count` => `correct_field_count`
- Remove `validate_min_columns` from sales bulk upload

### DIFF
--- a/spec/services/bulk_upload/lettings/year2022/csv_parser_spec.rb
+++ b/spec/services/bulk_upload/lettings/year2022/csv_parser_spec.rb
@@ -78,7 +78,7 @@ RSpec.describe BulkUpload::Lettings::Year2022::CsvParser do
     end
 
     it "counts the number of valid field numbers correctly" do
-      expect(service.incorrect_field_count?).to be false
+      expect(service).to be_correct_field_count
     end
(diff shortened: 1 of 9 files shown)


spec/services/bulk_upload/lettings/year2022/csv_parser_spec.rb
```diff
@@ -78,7 +78,7 @@
     end
 
     it "counts the number of valid field numbers correctly" do
-      expect(service.incorrect_field_count?).to be false
+      expect(service).to be_correct_field_count
```
natdeanlewissoftwire, line 81: cool, don't think I knew about this helper

natdeanlewissoftwire (approved):

PR #1624, opened 2023-05-15
CLDC-2350 Set uprn confirmed

kosiakkatrina:
We want UPRN confirmed to be set to yes whenever we bulk upload a log. That means not reseting it to nil when processing the UPRN changes.

### DIFF
--- a/app/services/bulk_upload/sales/year2023/row_parser.rb
+++ b/app/services/bulk_upload/sales/year2023/row_parser.rb
@@ -824,6 +824,8 @@ private
 
     attributes["uprn"] = field_19
     attributes["uprn_known"] = field_19.present? ? 1 : 0
+    attributes["uprn_confirmed"] = 1 if field_19.present?
+    attributes["skip_update_uprn_confirmed"] = true
     attributes["address_line1"] = field_20
     attributes["address_line2"] = field_21
     attributes["town_or_city"] = field_22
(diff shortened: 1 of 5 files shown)


app/services/bulk_upload/sales/year2023/row_parser.rb
```diff
@@ -816,6 +816,9 @@ def attributes_for_log
     attributes["mortgageused"] = mortgageused
 
     attributes["uprn"] = field_19
+    attributes["uprn_known"] = 1 if field_19.present?
```
natdeanlewissoftwire: this will conflict with the PR I've just put in, I set uprn_known as 0 if field_19 is not present as well.  Just a heads up in case mine gets in first :)

natdeanlewissoftwire (approved):

natdeanlewissoftwire:
lgtm!

app/services/bulk_upload/sales/year2023/row_parser.rb
natdeanlewissoftwire: this one: (https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/1625)

kosiakkatrina: Yup, sounds good, I'll rebase when you get yours merged, thanks!

PR #1631, opened 2023-05-16
CLDC-2359 Do not allow 32 as an option for 22/23 sale type

kosiakkatrina:
32 is a valid option for 23/24 shared ownership sale type, but it does not exist for 22/23. We want to prevent it from being saved with a setup error

### DIFF
--- a/spec/services/bulk_upload/sales/year2023/row_parser_spec.rb
+++ b/spec/services/bulk_upload/sales/year2023/row_parser_spec.rb
@@ -931,5 +931,16 @@ RSpec.describe BulkUpload::Sales::Year2023::RowParser do
         expect(log["mscharge"]).to be_nil
       end
     end
+
+    describe "shared ownership sale type" do
+      context "when 32 is selected for shared ownership type" do
+        let(:attributes) { valid_attributes.merge(field_8: "32") }
+
+        it "sets the value correctly" do
+          log = parser.log
+          expect(log["type"]).to eq(32)
+        end
+      end
+    end
   end
 end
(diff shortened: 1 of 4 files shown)


spec/services/bulk_upload/sales/year2023/row_parser_spec.rb
```diff
@@ -901,5 +901,16 @@ def purch_id_field
...
       end
     end
+
+    describe "shared ownership sale type" do
+      context "when 32 is selected for shared ownership type" do
+        let(:attributes) { valid_attributes.merge(field_8: "32") }
+
+        it "sets the value correctly" do
+          log = parser.log
+          expect(log["type"]).to eq(32)
```
asmega: log should be an activerecord object. can you therefore not use `log.type`?

asmega:
if its not a valid option would `validate_valid_radio_option` not clear it?

kosiakkatrina:
> if its not a valid option would `validate_valid_radio_option` not clear it?

@asmega I think it wouldn't cause it's a valid option in one of the years (?)

spec/services/bulk_upload/sales/year2023/row_parser_spec.rb
kosiakkatrina: Yup!

natdeanlewissoftwire (approved):
lgtm :) (although I also would have expected 32 to be cleared by `validate_valid_radio_option`)

update - I've just checked and it looks like it is being cleared, but because the error is in the setup section the log is created with only the setup section in progress...  I think actually there's a bug in some of the row parsers that are checking `if errors[field].present?`, I believe they should actually be `unless errors[field].present?` e.g. the sales 22 row parser, but look like it's not alone.

natdeanlewissoftwire (changes requested):
changing to request changes according to my update above, happy to be wrong but would like to work it out first!

natdeanlewissoftwire (approved):

PR #1633, opened 2023-05-16
CLDC-2305 Bulk upload setup fields validations

asmega:
# Context

- https://digital.dclg.gov.uk/jira/browse/CLDC-2305
- This is rework to add missing validations for sales setup fields for bulk upload

# Changes

- There is code that clears log fields if they are not a valid option. in order to generate errors for these we check certain fields for content prior to `log.valid?` being called. Unfortunately there does not appear to be an easy way to access to valid options therefore these values have been hard coded
- Privacy notice must be accepted other considered a setup error
- The `type` question can actually be one of three questions with the same identifier. these are excluded as part of `validate_valid_radio_option` and instead validated with a different mechanism. The rationale being that `log.valid?` will clear data before we get there
- Remove tests associated to upload threshold which are no longer required

### DIFF
--- a/app/services/bulk_upload/sales/year2022/row_parser.rb
+++ b/app/services/bulk_upload/sales/year2022/row_parser.rb
@@ -966,9 +966,9 @@ private
 
       if setup_question?(question)
         fields.each do |field|
-          if errors[field].blank?
+          if errors[field].none?
             block_log_creation!
-            errors.add(field, I18n.t("validations.invalid_option", question: QUESTIONS[field]), category: :setup)
+            errors.add(field, I18n.t("validations.invalid_option", question: QUESTIONS[field].downcase), category: :setup)
           end
         end
       else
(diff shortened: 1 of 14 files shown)


app/services/bulk_upload/sales/year2022/row_parser.rb
```diff
@@ -266,9 +266,63 @@ class BulkUpload::Sales::Year2022::RowParser
...
+              in: [8, 14, 27, 9, 29, 21, 22],
+              category: :setup,
+              question: QUESTIONS[:field_76].downcase,
+            },
+            if: proc { field_76.present? && discounted_ownership? },
+            on: :before_log
+
+  validates :field_76,
+            presence: {
+              message: I18n.t("validations.not_answered", question: "shared ownership type"),
```
kosiakkatrina: Should this be `question: "discounted ownership type"` ?

asmega: good spot, I've changed this now. there were a couple of others which I have also fixed

kosiakkatrina (approved):

PR #1634, opened 2023-05-16
CLDC-2313 add bulk upload validation buyer 1 cannot be child

arfacamble:
https://digital.dclg.gov.uk/jira/browse/CLDC-2313

9 (child under 16) is an option for all ecstat questions except ecstat 1, so rather than displaying the standard provide valid value validation we display a particular validation for buyer 1.
Further notes on particular code choices in line in diff

### DIFF
--- a/app/services/bulk_upload/sales/year2022/row_parser.rb
+++ b/app/services/bulk_upload/sales/year2022/row_parser.rb
@@ -274,6 +274,7 @@ class BulkUpload::Sales::Year2022::RowParser
   validates :field_114, presence: { message: I18n.t("validations.not_answered", question: "company buyer") }, if: :outright_sale?, on: :after_log
   validates :field_109, presence: { message: I18n.t("validations.not_answered", question: "more than 2 buyers") }, if: :joint_purchase?, on: :after_log
 
+  validate :validate_buyer1_economic_status, on: :before_log
   validate :validate_nulls, on: :after_log
   validate :validate_valid_radio_option, on: :before_log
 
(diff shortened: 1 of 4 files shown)


app/services/bulk_upload/sales/year2022/row_parser.rb
```diff
@@ -807,7 +808,7 @@ def mortgageused
   end
 
   def owning_organisation
-    Organisation.find_by_id_on_multiple_fields(field_92)
+    @owning_organisation ||= Organisation.find_by_id_on_multiple_fields(field_92)
```
arfacamble, line 816: memoize owning_organisation to remove need to for repeated identical DB calls.

app/services/bulk_upload/sales/year2022/row_parser.rb
```diff
@@ -941,7 +942,7 @@ def validate_nulls
         end
       else
         fields.each do |field|
-          unless errors.any? { |e| fields.include?(e.attribute) }
+          if errors.none? { |e| fields.include?(e.attribute) }
```
arfacamble, line 950: if preferred to else where easily possible

spec/services/bulk_upload/sales/year2022/row_parser_spec.rb
```diff
@@ -107,14 +107,6 @@
...
   end
 
-  around do |example|
-    FormHandler.instance.use_real_forms!
-
-    example.run
-
-    FormHandler.instance.use_fake_forms!
-  end
-
```
arfacamble, line 117: use_fake_forms! loads json forms from spec/fixtures rather than config/forms
There are no sales forms in either directory, so this isn't making any difference in this case I don't believe.

kosiakkatrina (approved):

PR #1635, opened 2023-05-17
Disable log updates by default for importers

jamesrose:
- Previously, when attempting to import a log that already exists in the service, the importers would update its attributes.
- In most situations we don't want this as it could override changes that the user has made to the logs.
- This change adds an option to the log import services to allow this functionality to be toggled.
- This PR allows us to re-migrate organisations where logs weren't previously created due to a validation error.

### DIFF
--- a/spec/services/imports/lettings_logs_import_service_spec.rb
+++ b/spec/services/imports/lettings_logs_import_service_spec.rb
@@ -85,12 +85,40 @@ RSpec.describe Imports::LettingsLogsImportService do
           .to change(LettingsLog, :count).by(3)
       end
 
-      it "only updates existing lettings logs" do
+      it "does not by default update existing lettings logs" do
         expect(logger).not_to receive(:error)
         expect(logger).not_to receive(:warn)
-        expect(logger).to receive(:info).with(/Updating lettings log/).exactly(3).times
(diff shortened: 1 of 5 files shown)


spec/services/imports/lettings_logs_import_service_spec.rb
```diff
@@ -85,12 +85,32 @@ def open_file(directory, filename)
...
+      context "with updates allowed" do
+        subject(:lettings_log_service) { described_class.new(storage_service, logger, allow_updates: true) }
+
+        it "only updates existing lettings logs" do
+          expect(logger).not_to receive(:error)
+          expect(logger).not_to receive(:warn)
+          expect(logger).to receive(:info).with(/Updating lettings log/).exactly(3).times
+          expect { 2.times { lettings_log_service.create_logs(remote_folder) } }
+            .to change(LettingsLog, :count).by(3)
+        end
```
kosiakkatrina: Do we expect(updated_logs).to eq(3) here?

kosiakkatrina (approved):

spec/services/imports/lettings_logs_import_service_spec.rb
jamesrose: @kosiakkatrina Ah yes.

PR #1637, opened 2023-05-17
CLDC-1731 Allow data coordinators to view parent's schemes

kosiakkatrina:
Display parent schemes for data coordinators:
Do not allow editing, deactivating schemes and locations or adding new locations to parent organisation scheme.

### DIFF
--- a/app/controllers/locations_controller.rb
+++ b/app/controllers/locations_controller.rb
@@ -227,7 +227,7 @@ private
 
   def authenticate_action!
     unless user_allowed_action?
-      render_not_found and return
+      render_not_found
     end
   end
 
--- a/app/controllers/locations_controller.rb
+++ b/app/controllers/locations_controller.rb
@@ -4,7 +4,7 @@ class LocationsController < ApplicationController
   before_action :authenticate_scope!
   before_action :find_location, except: %i[create index]
   before_action :find_scheme
-  before_action :authenticate_action!
+  before_action :authenticate_action!, only: %i[create update index new_deactivation deactivate_confirm deactivate postcode local_authority name units type_of_unit mobility_standards availability check_answers]
   before_action :scheme_and_location_present, except: %i[create index]
 
   include Modules::SearchFilter
@@ -21,6 +21,7 @@ class LocationsController < ApplicationController
   end
 
   def postcode; end
+  def update; end
 
   def update_postcode
     @location.postcode = location_params[:postcode]
@@ -225,13 +226,13 @@ private
   end
 
   def authenticate_action!
-    if %w[create update index new_deactivation deactivate_confirm deactivate postcode local_authority name units type_of_unit mobility_standards availability check_answers].include?(action_name) && !user_allowed_action?
+    unless user_allowed_action?
       render_not_found and return
     end
   end
 
   def user_allowed_action?
-    (current_user.organisation == @scheme&.owning_organisation) || (current_user.organisation.parent_organisations.any? { |org| org == @scheme&.owning_organisation }) || current_user.support?
+    current_user.support? || current_user.organisation == @scheme&.owning_organisation || current_user.organisation.parent_organisations.exists?(@scheme&.owning_organisation_id)
   end
 
   def location_params
--- a/app/controllers/locations_controller.rb
+++ b/app/controllers/locations_controller.rb
@@ -225,9 +225,13 @@ private
   end
 
   def authenticate_action!
-    if %w[create update index new_deactivation deactivate_confirm deactivate postcode local_authority name units type_of_unit mobility_standards availability check_answers].include?(action_name) && !((current_user.organisation == @scheme&.owning_organisation) || current_user.support?)
+    if %w[create update index new_deactivation deactivate_confirm deactivate postcode local_authority name units type_of_unit mobility_standards availability check_answers].include?(action_name) && !user_allowed_action?
       render_not_found and return
     end
+  end
+
+  def user_allowed_action?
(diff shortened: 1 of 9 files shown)


app/controllers/locations_controller.rb
```diff
@@ -225,11 +225,15 @@ def authenticate_scope!
...
 
   def authenticate_action!
-    if %w[create update index new_deactivation deactivate_confirm deactivate postcode local_authority name units type_of_unit mobility_standards availability check_answers].include?(action_name) && !((current_user.organisation == @scheme&.owning_organisation) || current_user.support?)
+    if %w[create update index new_deactivation deactivate_confirm deactivate postcode local_authority name units type_of_unit mobility_standards availability check_answers].include?(action_name) && !user_allowed_action?
       render_not_found and return
     end
   end
 
+  def user_allowed_action?
+    (current_user.organisation == @scheme&.owning_organisation) || (current_user.organisation.parent_organisations.any? { |org| org == @scheme&.owning_organisation }) || current_user.support?
```
ghost: you'd probably make this check a bit more efficient by moving `current_user.support?` to the front

app/controllers/schemes_controller.rb
```diff
@@ -265,11 +265,15 @@ def find_resource
...
     head :unauthorized and return unless current_user.data_coordinator? || current_user.support?
 
-    if %w[show locations primary_client_group confirm_secondary_client_group secondary_client_group support details check_answers edit_name deactivate].include?(action_name) && !((current_user.organisation == @scheme&.owning_organisation) || current_user.support?)
+    if %w[show locations primary_client_group confirm_secondary_client_group secondary_client_group support details check_answers edit_name deactivate].include?(action_name) && !user_allowed_action?
       render_not_found and return
     end
   end
 
+  def user_allowed_action?
+    (current_user.organisation == @scheme&.owning_organisation) || (current_user.organisation.parent_organisations.any? { |org| org == @scheme&.owning_organisation }) || current_user.support?
```
ghost: same comment as above

app/controllers/locations_controller.rb
ghost:
```
(current_user.organisation.parent_organisations.any? { |org| org == @scheme&.owning_organisation })
```
Could use [exists?](https://apidock.com/rails/ActiveRecord/Base/exists%3F/class) and run this check with a single query without having to store anything in memory

app/controllers/locations_controller.rb
```diff
@@ -225,11 +225,15 @@ def authenticate_scope!
   end
 
   def authenticate_action!
-    if %w[create update index new_deactivation deactivate_confirm deactivate postcode local_authority name units type_of_unit mobility_standards availability check_answers].include?(action_name) && !((current_user.organisation == @scheme&.owning_organisation) || current_user.support?)
+    if %w[create update index new_deactivation deactivate_confirm deactivate postcode local_authority name units type_of_unit mobility_standards availability check_answers].include?(action_name) && !user_allowed_action?
```
ghost: can't this be moved to a more standard `except` `only` to the top of the file?

app/helpers/locations_helper.rb
```diff
@@ -84,6 +84,10 @@ def location_creation_success_notice(location)
     end
   end
 
+  def user_can_edit_scheme?(user, scheme)
+    user.support? || user.organisation == scheme.owning_organisation
```
ghost, line 88: could move this check to a policy,

app/controllers/locations_controller.rb
ghost: and could also move this logic to a Policy

app/controllers/schemes_controller.rb
kosiakkatrina: Nice!

app/controllers/locations_controller.rb
kosiakkatrina: Yup, can do!

app/controllers/locations_controller.rb
```diff
@@ -225,13 +226,13 @@ def authenticate_scope!
   end
 
   def authenticate_action!
-    if %w[create update index new_deactivation deactivate_confirm deactivate postcode local_authority name units type_of_unit mobility_standards availability check_answers].include?(action_name) && !user_allowed_action?
+    unless user_allowed_action?
       render_not_found and return
```
ghost: you might not need the return anymore here but not 100% sure

app/controllers/locations_controller.rb
```diff
@@ -225,13 +226,13 @@ def authenticate_scope!
...
     end
   end
 
   def user_allowed_action?
-    (current_user.organisation == @scheme&.owning_organisation) || (current_user.organisation.parent_organisations.any? { |org| org == @scheme&.owning_organisation }) || current_user.support?
+    current_user.support? || current_user.organisation == @scheme&.owning_organisation || current_user.organisation.parent_organisations.exists?(@scheme&.owning_organisation_id)
```
ghost, line 235: can this happen?

`current_user.organisation.parent_organisations.exists?(nil) == true` ?

kosiakkatrina: `Organisation.exists?(nil)` and `Organisation.exists?(nil&.id)` both return false, so should be fine

app/helpers/locations_helper.rb
kosiakkatrina: Hmm, don't know if this alone merits a policy 🤔

ghost (approved):

asmega:
I feel like this will somewhat clash with my PR https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/1642

PR #1638, opened 2023-05-17
CLDC-2348 Purchaser code bug fix

natdeanlewissoftwire:
Purchaser code was being set to field_1, but in 23/24 it's field 6 so we assign it in the row_parsers instead.

Don't think this needs a test as the new row parsers will be made from the current ones, so the purchaser_code method should be present and updated in future years too.

Ticket: https://digital.dclg.gov.uk/jira/browse/CLDC-2348

### DIFF
--- a/app/services/bulk_upload/sales/validator.rb
+++ b/app/services/bulk_upload/sales/validator.rb
@@ -23,7 +23,7 @@ class BulkUpload::Sales::Validator
         bulk_upload.bulk_upload_errors.create!(
           field: error.attribute,
           error: error.message,
-          purchaser_code: row_parser.field_1,
+          purchaser_code: row_parser.purchaser_code,
           row:,
           cell: "#{col}#{row}",
           col:,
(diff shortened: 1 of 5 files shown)


asmega (approved):

ghost:
even if this is a trivial change, I think we should cover this with a test. Especially because this PR is fixing a bug

natdeanlewissoftwire:
> even if this is a trivial change, I think we should cover this with a test. Especially because this PR is fixing a bug

Yeah, I sort of half agree - have had a think about a sensible way to test this for the future without knowing what field purch_id is for future years' bulk upload templates... I guess I can check for the current row parser that purchaser_code is the field with the question "What is the purchaser code?" but that's about it, maybe slightly less vulnerable to errors but doesn't totally protect us as we don't know what fields to actually check yet.

I don't see much value in adding one for the existing row_parsers for 22/23 and 23/24 checking that purchaser_code returns the right field because that's just as hard coded as the method itself.

ghost (approved):

ghost (approved):

PR #1642, opened 2023-05-17
CLDC-1732 Provider and read only schemes

asmega:
# Context

- https://digital.dclg.gov.uk/jira/browse/CLDC-1732
- data providers are given read-only able access to schemes and locations 

# Changes

- introduce `pundit` policies to schemes and locations. the old scope mechanism has been removed
- apply policies at view level so hide write access based functionality from data providers

### DIFF
--- a/app/helpers/navigation_items_helper.rb
+++ b/app/helpers/navigation_items_helper.rb
@@ -24,6 +24,7 @@ module NavigationItemsHelper
       [
         NavigationItem.new("Lettings logs", lettings_logs_path, lettings_logs_current?(path)),
         NavigationItem.new("Sales logs", sales_logs_path, sales_logs_current?(path)),
+        (NavigationItem.new("Schemes", "/schemes", subnav_supported_housing_schemes_path?(path)) if current_user.organisation.holds_own_stock?),
         NavigationItem.new("Users", users_organisation_path(current_user.organisation), subnav_users_path?(path)),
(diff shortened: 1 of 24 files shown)


app/helpers/navigation_items_helper.rb
```diff
@@ -24,6 +24,7 @@ def primary_items(path, current_user)
       [
         NavigationItem.new("Lettings logs", lettings_logs_path, lettings_logs_current?(path)),
         NavigationItem.new("Sales logs", sales_logs_path, sales_logs_current?(path)),
+        (NavigationItem.new("Schemes", "/schemes", subnav_supported_housing_schemes_path?(path)) if current_user.organisation.holds_own_stock?),
         NavigationItem.new("Users", users_organisation_path(current_user.organisation), subnav_users_path?(path)),
         NavigationItem.new("About your organisation", "/organisations/#{current_user.organisation.id}", subnav_details_path?(path)),
         NavigationItem.new("Stock owners", stock_owners_organisation_path(current_user.organisation), stock_owners_path?(path)),
```
kosiakkatrina, line 20: Is this the same as above? Maybe we can remove the one above? I mean this condition: ` elsif current_user.data_coordinator? && current_user.organisation.holds_own_stock?`

asmega: ok, i think i see you and I've deleted the branch

kosiakkatrina (approved):

PR #1643, opened 2023-05-18
Load Form factories only in test environment

ghost:
It caused superclass mismatch errors. We can stop seeing this bug in local development by defining the classes only in test environments

### DIFF
--- a/spec/factories/form.rb
+++ b/spec/factories/form.rb
@@ -1,25 +1,27 @@
-class FormFixture < Form
-  attr_accessor :sections, :subsections, :pages, :questions
-end
-
-class FormFactory
-  def initialize(year:, type:)
-    @year = year
-    @type = type
+if Rails.env.test?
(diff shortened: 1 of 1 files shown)


arfacamble (commented):
🎉 🎉 🎉 🎉 🎉 
MASSIVE APPROVE

thanks for showing me this.

I still don't really understand the error, but this kind of makes sense

arfacamble (commented):
🎉 🎉 🎉 🎉 🎉 
MASSIVE APPROVE

thanks for showing me this.

I still don't really understand the error, but this kind of makes sense

arfacamble (approved):
🎉 🎉 🎉 🎉 🎉 
MASSIVE APPROVE

thanks for showing me this.

I still don't really understand the error, but this kind of makes sense

jamesrose (approved):

PR #1645, opened 2023-05-18
CLDC-2375 Remove soft validations from setup

kosiakkatrina:
## Remove soft validation pages from setup and make them mandatory.
We currently are marking them as optional, because any missing answer in setup section locks and clears the rest of the log.

The issue with that is that a log would appear as completed even if some validations are not confirmed. This also causes inconsistent error and log numbers on bulk uploads.

It might be enough just having the soft validations next to all the other related questions in other sections, especially as the new soft validation design now links to all the relevant questions as well.


ghost (approved):

spec/models/form_spec.rb
```diff
@@ -354,9 +354,9 @@ def answer_local_authority(lettings_log)
       expect(form.sections[0].class).to eq(Form::Sales::Sections::Setup)
       expect(form.subsections.count).to eq(1)
       expect(form.subsections.first.id).to eq("setup")
-      expect(form.pages.count).to eq(18)
+      expect(form.pages.count).to eq(12)
```
natdeanlewissoftwire, line 357: I'd support just setting these to test for positive values rather than specific counts like we have in `form_handler_spec`

```suggestion
      expect(form.pages.count).to be_positive
```

spec/models/form_spec.rb
```diff
@@ -354,9 +354,9 @@ def answer_local_authority(lettings_log)
...
       expect(form.pages.first.id).to eq("organisation")
-      expect(form.questions.count).to eq(19)
+      expect(form.questions.count).to eq(13)
```
natdeanlewissoftwire, line 359: same as above :)

natdeanlewissoftwire (approved):
one optional suggestion, but lgtm!

spec/models/form_spec.rb
kosiakkatrina: Imma keep as is, cause the setup section doesn't change much (and it shouldn't), so maybe thips test failing would make people think twice before adding/removing things from setup

PR #1648, opened 2023-05-19
Only loop through existing field mappings for errors

kosiakkatrina:
We got this error on 2022 row parser.
```
undefined method `each' for nil:NilClass
field_mapping_for_errors[interruption_screen_question_id.to_sym].each do |field|
```
This PR adds the only missing interruption screen question id from 2022 form, but I've also added a check for nil before we look through those errors.

### DIFF
--- a/app/services/bulk_upload/lettings/year2022/row_parser.rb
+++ b/app/services/bulk_upload/lettings/year2022/row_parser.rb
@@ -770,7 +770,7 @@ private
       next if question.completed?(log)
 
       question.page.interruption_screen_question_ids.each do |interruption_screen_question_id|
-        field_mapping_for_errors[interruption_screen_question_id.to_sym].each do |field|
+        field_mapping_for_errors[interruption_screen_question_id.to_sym]&.each do |field|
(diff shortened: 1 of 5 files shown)


app/services/bulk_upload/lettings/year2022/row_parser.rb
```diff
@@ -770,7 +770,7 @@ def validate_incomplete_soft_validations
       next if question.completed?(log)
 
       question.page.interruption_screen_question_ids.each do |interruption_screen_question_id|
-        field_mapping_for_errors[interruption_screen_question_id.to_sym].each do |field|
+        field_mapping_for_errors[interruption_screen_question_id.to_sym]&.each do |field|
```
jamesrose, line 773: ~Are you sure this is what you want to do? `nil#each` will still error. How about `field_mapping_for_errors.fetch(interruption_screen_question_id.to_sym, [])`?~

asmega (approved):

jamesrose (approved):

PR #1652, opened 2023-05-19
CLDC-2390 Extract town question to a separate row in CYA

kosiakkatrina:
Migrated logs don't always have all the address questions answered.
When some of the address fields are answered, but a mandatory field, such as town or city is not answered, the log is marked in progress, but there's no clear indication as to what question needs answering (because we display all the address answers on a single row in CYA).

This PR splits out town or city question in CYA, so it's clear to see when it's not answered.
image

### DIFF
--- a/app/models/form/lettings/questions/address_line1.rb
+++ b/app/models/form/lettings/questions/address_line1.rb
@@ -15,7 +15,6 @@ class Form::Lettings::Questions::AddressLine1 < ::Form::Question
       log.address_line1,
       log.address_line2,
       log.postcode_full,
-      log.town_or_city,
       log.county,
     ].select(&:present?).join("\n")
   end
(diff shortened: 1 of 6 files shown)


natdeanlewissoftwire (approved):
Lgtm, though I'm assuming there's another PR to split out address line 1 as that's mandatory too?

A couple of design points separate from the technical work - should we be numbering the town or city question?

Also, shouldn't we split out town or city ONLY when it's not been answered?

kosiakkatrina:
> Lgtm, though I'm assuming there's another PR to split out address line 1 as that's mandatory too?
> 
> A couple of design points separate from the technical work - should we be numbering the town or city question?
> 
> Also, shouldn't we split out town or city ONLY when it's not been answered?

Hmm, I'm not sure. I think the point of doing it like this in the first place was to keep it simple and adding logic to only split it up when answered seems like an overkill, but I'll leave that to PO

natdeanlewissoftwire:
Cool, works for me!

PR #1657, opened 2023-05-23
CLDC-2290 implement delete multiple logs story

arfacamble:
[ticket](https://digital.dclg.gov.uk/jira/browse/CLDC-2290)

All routes and methods for this story have been extracted to own controller.

|image
|-
|-
|image
|-
|image
|-
|image
|-


app/controllers/lettings_logs_controller.rb
```diff
@@ -111,6 +110,37 @@ def update_logs
...
+    render "logs/delete_logs"
+  end
+
+  def delete_logs_confirmation
+    log_ids = params.dig(:forms_delete_logs_form, :logs_to_delete)
+    log_type = params.dig(:forms_delete_logs_form, :log_type).to_sym
+    @delete_logs_form = Forms::DeleteLogsForm.new(log_ids:, log_type:)
+    @search_term = params.dig(:forms_delete_logs_form, :search_term)
+    if @delete_logs_form.no_logs_selected?
+      redirect_to delete_logs_lettings_logs_path(search: @search_term, no_logs_selected: true)
```
jamesrose: We shouldn't need to redirect here. Instead, we can just re-render the template with errors.

app/controllers/lettings_logs_controller.rb
```diff
@@ -111,6 +110,37 @@ def update_logs
     end
   end
 
+  def delete_logs_list
+    all_logs = current_user.lettings_logs.visible
+    @search_term = search_term
+    @delete_logs_form = Forms::DeleteLogsForm.new(logs: filtered_logs(all_logs, @search_term, @session_filters))
+    if params[:no_logs_selected] == "true"
+      @delete_logs_form.errors.add(:logs_ids, "Select at least one log to delete or press cancel to return")
```
jamesrose: This logic belongs in the form object itself.

app/controllers/lettings_logs_controller.rb
```diff
@@ -111,6 +110,37 @@ def update_logs
...
+    end
+    render "logs/delete_logs"
+  end
+
+  def delete_logs_confirmation
+    log_ids = params.dig(:forms_delete_logs_form, :logs_to_delete)
+    log_type = params.dig(:forms_delete_logs_form, :log_type).to_sym
+    @delete_logs_form = Forms::DeleteLogsForm.new(log_ids:, log_type:)
```
jamesrose:
```suggestion
    @delete_logs_form = Forms::DeleteLogsForm.new(attributes: params[:forms_delete_logs_form])
```

The extraction of these attributes should be encapsulated within the form object.

app/controllers/lettings_logs_controller.rb
```diff
@@ -111,6 +110,37 @@ def update_logs
...
+    end
+  end
+
+  def delete_logs
+    ids_to_delete = params.require("ids").split.map(&:to_i)
+    ids_to_delete.each do |id|
+      log = LettingsLog.visible.find(id)
+      authorize log, :destroy?
+      log.discard!
+    end
```
jamesrose: I'm curious why we need to split IDs here?

```suggestion
  def delete_logs
    logs = LettingsLog.visible.find(params[:log_ids])
    logs.each { |log| authorize(log, :destroy?) }
    logs.discard_all!
  end
```

app/models/forms/delete_logs_form.rb
```diff
@@ -0,0 +1,18 @@
...
+
+    def initialize(logs: nil, log_ids: nil, log_type: nil)
+      @logs = logs
+      @log_ids = log_ids
+      @log_type = log_type || (logs.first&.lettings? ? :lettings : :sales)
+    end
+
+    def no_logs_selected?
+      @log_ids.blank? || @log_ids.reject(&:blank?).blank?
+    end
```
jamesrose: We should add validations here and instead use `valid?` externally to determine whether we're save to progress the deletion.

jamesrose:
@arfacamble I think you've made a good start here, but we can better encapsulate all the data and related behavior within the form object, as opposed to the controller.

app/models/forms/delete_logs_form.rb
```diff
@@ -0,0 +1,18 @@
...
+  class DeleteLogsForm
+    include ActiveModel::Model
+    include ActiveModel::Validations
+
+    attr_reader :logs, :log_type
+
+    def initialize(logs: nil, log_ids: nil, log_type: nil)
+      @logs = logs
+      @log_ids = log_ids
+      @log_type = log_type || (logs.first&.lettings? ? :lettings : :sales)
```
jamesrose: I think this optimisation has introduced complexity. Did you consider two separate form objects and templates for now?

app/helpers/log_list_helper.rb
```diff
@@ -0,0 +1,25 @@
...
+  def display_delete_logs?(current_user, search_term)
+    if current_user.data_provider?
+      filter_selected?("user", "yours")
+    else
+      any_filter_selected? || search_term.present?
+    end
+  end
+
+  def delete_logs_path(logs, search_term)
+    controller_type = controller.class.name.delete_suffix("Controller").underscore.to_sym
```
jamesrose: We shouldn't have to do this. Can you explain what you're trying to din this method?

app/controllers/lettings_logs_controller.rb
arfacamble: this is just the sort of railsy thing I would expect to be able to do. great.

app/views/logs/delete_logs.html.erb
```diff
@@ -0,0 +1,61 @@
...
+            <% case @delete_logs_form.log_type %>
+            <% when :lettings %>
+              <% row.cell(text: log.tenancycode) %>
+              <% row.cell(text: log.propcode) %>
+            <% when :sales %>
+              <%# for sales display owning org name and creation date %>
+            <% end %>
+            <% row.cell(text: status_tag(log.status)) %>
+            <% row.cell do %>
+              <% f.check_box :bla %>
```
jamesrose: !

app/views/logs/delete_logs.html.erb
```diff
@@ -0,0 +1,61 @@
...
+              <% f.govuk_check_box :logs_to_delete, log.id,
```
jamesrose:
```suggestion
              <% f.govuk_check_box :log_ids, log.id,
```

app/models/forms/delete_logs_form.rb
arfacamble: Given that in the confirmation method, we need to create a form object to ask if it is valid and then either pass that thing back to the list of logs to delete view or on to the confirmation page, I have stuck with one class of form object but admittedly it does have some complexity in its initialize method still. Happy to discuss how that might be mitigated.

app/views/logs/delete_logs.html.erb
arfacamble: there is already a variable log_ids representing all the logs in the table.
Do you want the change so that it is expressing that this thing is a collection of ids rather than actual logs?
I think here in the context of this form and consistent with the naming in the controller methods I will change this to selected_ids unless you don't like that

app/helpers/log_list_helper.rb
arfacamble: I need to insert a Delete Logs button into the index page (_log_list partial) which is reused for all different log types and can be rendered by three different controllers.
I'm up for discussing this and further thoughts on it, but I also think the implementation in anger of this functionality can be worried about when I'm implementing for these other controllers.

app/controllers/lettings_logs_controller.rb
```diff
@@ -111,6 +111,31 @@ def update_logs
     end
   end
 
+  def delete_logs_list
+    log_ids = params.require(:log_ids).map(&:to_i)
+    selected_ids = params[:selected_ids]&.map(&:to_i)
```
jamesrose: Are you sure that you need to map these to integers? AR finders are quite content with strings.

app/views/logs/delete_lettings_logs.html.erb
```diff
@@ -0,0 +1,49 @@
...
+<% content_for :before_content do %>
+  <%= govuk_back_link(href: :back) %>
+<% end %>
+
+<h1 class="govuk-heading-l">
+  <span class="govuk-caption-l"><%= title %></span>
+  Review the logs you want to delete
+</h1>
+<% logs = @delete_logs_form.logs %>
+<p>You've selected <%= logs&.count %> <%= "log".pluralize(logs&.count) %> to delete</p>
```
asmega: the form can act a view object so i wouldn't bother with the local variable and just reference `@delete_logs_form` wholly instead

as an example if you want less code is to push stuff/logic downwards so the responsibility becomes delegated. if the form implemented `form.log_count` you could push that downwards then view does not need to deal with nil guards. the view just wants to ask for some info ie log count to render. it shouldn't need to deal with nil guards as that is someone else's responsibility, in this case the form/view.object

config/routes.rb
```diff
@@ -217,6 +222,10 @@
       post "email-csv", to: "sales_logs#email_csv"
       get "csv-confirmation", to: "sales_logs#csv_confirmation"
 
+      get "delete-logs", to: "sales_logs#delete_logs_list"
+      get "delete-logs-confirmation", to: "sales_logs#delete_logs_confirmation"
+      delete "delete-logs", to: "sales_logs#delete_logs"
```
asmega: how come sales has 3 routes added here but lettings has 4 new routes?

app/models/forms/delete_logs_form.rb
```diff
@@ -0,0 +1,29 @@
...
+    validate :at_least_one_log_selected
+
+    def initialize(attributes)
+      @log_type = attributes[:log_type]
+      @search_term = attributes[:search_term]
+      current_user = attributes[:current_user]
+      all_logs = case @log_type
+                 when :lettings then current_user.lettings_logs.visible
+                 when :sales then current_user.sales_logs.visible
+                 end
```
asmega: if you made `current_user` available with a getter ie `attr_reader :current_user` the rest of the code can be simplified. if you're concerned with exposing this then you can make the getter private

if the above is done then `all_logs` can be extracted to a private method giving shorter more self contained and documented methods

app/helpers/log_list_helper.rb
asmega: if we want fewer if/case statements if this were moved to the relevant controller or helper the switching can be removed. ie `LettingsLogsController` or `SalesLogsController` (or the helpers, not sure on those names or ordering)

arfacamble: 🤦‍♂️

app/views/organisations/logs.html.erb
```diff
@@ -36,6 +36,7 @@
                  total_count: @total_count,
                  csv_download_url: csv_download_url_by_log_type(@log_type, @organisation, search: @search_term, codes_only: false),
                  csv_codes_only_download_url: csv_download_url_by_log_type(@log_type, @organisation, search: @search_term, codes_only: true),
+                 delete_logs_path: "#",
```
ghost: what does this do?

spec/requests/lettings_logs_controller_spec.rb
```diff
@@ -347,7 +347,6 @@
               Timecop.freeze(2022, 3, 1) do
                 example.run
               end
-              Timecop.return
```
ghost, line 350: did you mean to remove this line?

app/views/organisations/logs.html.erb
arfacamble: as in slack description, I am now implementing the work you see here for lettings for organisations and sales also. This will be improved

spec/requests/lettings_logs_controller_spec.rb
arfacamble: very much so.
The point of putting `example.run` in a block is that the time freeze exists only for that block.
The reasonable alternative to what I have changed this to would be:
```
Timecop.freeze(date)
example.run
Timecop.return
```

spec/views/form/page_view_spec.rb
```diff
@@ -22,8 +22,6 @@ def assign_attributes(object, attrs)
       Singleton.__init__(FormHandler)
       example.run
     end
-    Timecop.return
-    Singleton.__init__(FormHandler)
```
arfacamble, line 26: @bibblobcode similarly here Jack, you may not be aware that a while ago @kosiakkatrina put a line in the test config so that the FormHandler singleton is reinitialised after every single test, so anywhere you see this you can also try removing it as it should be redundant

config/application.rb
arfacamble, line 1: this allows us to create a controller from the command line and create matching test files for request specs rather than controller tests

app/models/forms/delete_logs_form.rb
```diff
@@ -0,0 +1,40 @@
...
+
+  private
+
+    def at_least_one_log_selected
+      if selected_ids.blank? || selected_ids.reject(&:blank?).blank?
+        errors.add(:log_ids, "Select at least one log to delete or press cancel to return")
+      end
+    end
+
+    def all_logs
```
jamesrose: Given this is not _all_ logs, how about renaming to?

```suggestion
    def visible_logs
```

app/views/logs/delete_logs.html.erb
```diff
@@ -0,0 +1,58 @@
...
+<p>You've selected <%= @delete_logs_form.log_count %> <%= "log".pluralize(@delete_logs_form.log_count) %> to delete</p>
+
+<%= form_with model: @delete_logs_form, url: @delete_logs_form.delete_confirmation_path do |f| %>
+  <%= f.hidden_field :search_term, value: @delete_logs_form.search_term %>
+  <%= f.govuk_error_summary %>
+  <%= govuk_table do |table| %>
+    <% table.head do |head| %>
+      <% head.row do |row| %>
+        <% row.cell(header: true, text: "Log ID") %>
+        <% if @delete_logs_form.log_type == :lettings %>
```
jamesrose: There is enough variation here that you should render a different template for each table, embedded within the form.

app/views/logs/update_logs.html.erb
```diff
@@ -4,43 +4,42 @@
...
+        <% body.row do |row| %>
+          <% row.cell(text: log.id) %>
+          <% row.cell(text: log.tenancycode) %>
+          <% row.cell(text: log.propcode) %>
+          <% row.cell(text: status_tag(log.status)) %>
+          <% row.cell(html_attributes: {
+            scope: "row",
+            class: "govuk-!-text-align-right",
+          }) do %>
+            <%= govuk_link_to("Update now", send(log.form.unresolved_log_path, log)) %>
```
jamesrose: Using `send` here is superfluous:

```suggestion
            <%= govuk_link_to("Update now", log.form.unresolved_log_path(log)) %>
```

app/models/forms/delete_logs_form.rb
```diff
@@ -0,0 +1,40 @@
...
+    attr_reader :logs, :log_type, :selected_ids, :search_term, :delete_confirmation_path, :back_to_logs_path, :delete_path
+
+    validate :at_least_one_log_selected
+
+    def initialize(attributes)
+      @log_type = attributes[:log_type]
+      @search_term = attributes[:search_term]
+      @current_user = attributes[:current_user]
+      @logs = FilterService.filter_logs(all_logs, @search_term, attributes[:log_filters], nil, @current_user)
+      @selected_ids = attributes[:selected_ids] || @logs.map(&:id)
```
jamesrose, line 15: As far as I can see, `attributes[:selected_ids]` will always have values, or default to `[]`. Is that right?

If you fetch with a default then you can remove the defaulting outside of this class:

```suggestion
      @selected_ids = attributes.fetch(:selected_ids, [])
```

app/controllers/delete_logs_controller.rb
```diff
@@ -0,0 +1,196 @@
...
+    @delete_logs_form = Forms::DeleteLogsForm.new(attributes)
+    if @delete_logs_form.valid?
+      render "logs/delete_logs_confirmation"
+    else
+      render "logs/delete_logs"
+    end
+  end
+
+  def discard_sales_logs_for_organisation
+    logs = SalesLog.find(params.require(:ids))
```
jamesrose: If we know the organisation at this point, we should try and be defensive and scope our query under that organisation.

app/controllers/delete_logs_controller.rb
```diff
@@ -0,0 +1,196 @@
...
+    }.merge sales_logs_for_organisation_paths
+    attributes = form_attributes.merge(default_attributes)
+    attributes[:selected_ids] = [] unless attributes.key? :selected_ids
+    @delete_logs_form = Forms::DeleteLogsForm.new(attributes)
+    if @delete_logs_form.valid?
+      render "logs/delete_logs_confirmation"
+    else
+      render "logs/delete_logs"
+    end
+  end
```
jamesrose: As you repeat the first block of this method, would you consider extracting the params out into a separate method?

```ruby
  def delete_sales_logs_for_organisation_confirmation
    @delete_logs_form = Forms::DeleteLogsForm.new(attributes_for_sales_logs(organisation: true))
    if @delete_logs_form.valid?
      render "logs/delete_logs_confirmation"
    else
      render "logs/delete_logs"
    end
  end
  
  private
  
  # ...
  
  def attributes_for_sales_logs(organisation: false)
     attributes = {
      current_user:,
      log_filters: @session_filters,
      log_type: :sales,
    }
    attributes.merge(organisation ? sales_logs_for_organisation_paths : sales_logs_paths)
  end 
```

app/views/logs/delete_logs_confirmation.html.erb
```diff
@@ -0,0 +1,25 @@
...
+<p>You've selected <%= log_count %> <%= "log".pluralize(log_count) %> to delete</p>
+
+<%= govuk_warning_text(icon_fallback_text: "Danger") do %>
+  You will not be able to undo this action
+<% end %>
+
+<div class="govuk-button-group">
+  <%= govuk_button_to "Delete logs", @delete_logs_form.delete_path, method: "delete", params: { ids: @delete_logs_form.selected_ids } %>
+  <%= form_with url: @delete_logs_form.delete_path do |f| %>
+    <%= f.hidden_field :selected_ids, value: @delete_logs_form.selected_ids %>
```
jamesrose, line 21: You can pass these in as an array and it should just work:

```suggestion
    <% @delete_logs_form.selected_ids.each do |id|
      <%= f.hidden_field "selected_ids[]", value: id %>
    <% end %>
```

spec/requests/delete_logs_controller_spec.rb
```diff
@@ -0,0 +1,946 @@
...
+
+  describe "GET lettings-logs/delete-logs" do
+    let!(:log_1) { create(:lettings_log, :in_progress, created_by: user) }
+    let!(:log_2) { create(:lettings_log, :completed, created_by: user) }
+
+    before do
+      allow(FilterService).to receive(:filter_logs).and_return LettingsLog.all
+    end
+
+    it "calls the filter service with the filters in the session and the search term from the query params" do
```
jamesrose, line 20: You are testing the functionality of `FilterService` here which should be tested within its own spec.

jamesrose:
Do you have any tests that capture permissions for deleting logs that are only relevant to the user or organisation itself?

app/views/logs/delete_logs.html.erb
arfacamble: I just merged `delete_lettings_logs.html.erb` and `delete_sales_logs.html.erb` yesterday as after I had polished everything then they had two lines different.
I think I will need a little convincing to change this back as as far as I can tell what you're suggesting is still having an if/else inside the form to render two different table partials that only have 2 lines out of ~30 different, so it feels like much of a muchness to me.

app/views/logs/update_logs.html.erb
arfacamble: I only changed the indentation in this file, I didn't read it in much detail, happy to change this.

app/views/logs/delete_logs.html.erb
jamesrose: You could render the template based on the type, which would avoid the need for a conditional statement, for example:

```ruby
<%= render partial: "delete_#{@delete_logs.form.log_type}" %>
```

app/models/forms/delete_logs_form.rb
arfacamble: There may be a better way to organise this:
- when the user initially navigates from an index page, there are no selected_ids in attributes and we wish to create an array with all ids in it (we can only create this array after calling the FilterService, which happens in the Form object initialize)
- if the user presses cancel on the confirmation page, he is returned with an array of selected_ids which are passed into the form object initialize
- if the user unselects all checkboxes and presses delete, then in the controller confirmation methods `params[:forms_delete_logs_form]` has no selected_ids key in it at all, so if it is not present, we set it to be an empty array, eg on line 25.
I can't immediately see a way to rejig what I've done to make it cleaner and account for these situations, but I'm happy to hear suggestions.

app/views/logs/delete_logs.html.erb
arfacamble: aha ok cool. will do.
out of interest, I would expect that to be `render partial: ...`
a quick scan of docs and SO makes me think it doesn't massively matter the difference?

jamesrose: Sorry, yes, typo!

app/controllers/delete_logs_controller.rb
arfacamble: I hadn't fully looked through this suggested change, but I have refactored in a similar but different way.
I've retained the delete_logs_form private method, but made it able to be used by all controller methods wanting a delete logs form. I will push that along with a range of other changes soon

app/controllers/delete_logs_controller.rb
arfacamble: I'm happy to do this. It seems sensible good practice but is it necessary more than that?
The ids are coming from the last page, so they are either correct, or the user is doing something devious, but the user can still only delete logs for which he is authorised to on the next line anyway.

spec/requests/delete_logs_controller_spec.rb
arfacamble: I'm interested in discussing this.
The FilterService is tested independently, my intention was to make sure that the method calls the FilterService with the correct arguments and then in the next test that the page displays whatever is returned.

app/views/logs/delete_logs_confirmation.html.erb
arfacamble: I would expect both of these options to result in the same html. Is there a reason to want the more verbose one?

arfacamble:
> Do you have any tests that capture permissions for deleting logs that are only relevant to the user or organisation itself?

in the controller :discard method I call authorize log, destroy? on each log, which checks against the LettingsLogPolicy. That policy has tests written.

I hve got tests that only logs for which the user is authorized can be destroyed, for the sales and lettings, but this doesn't make sense under the organisations tab as only support users can navigate here and support users can delete any log

app/models/forms/delete_logs_form.rb
```diff
@@ -0,0 +1,44 @@
+module Forms
```
ghost, line 1: this object is missing unit tests

ghost (approved):
Just added a comment regarding a missing spec file but the rest looks great. Amazing job @arfacamble

PR #1665, opened 2023-05-24
CLDC-2386 Validate the correct template for the year

kosiakkatrina:
When a 22/23 template is uploaded for a 23/24 bulk upload (23/24 selected in the form at `/bulk-upload-logs/year`) we attempt to create the logs, but all the questions are different and out of order.
Instead we should discard that file as it is using the wrong template.
We're doing it here by trying to construct the startdate/saledate from the first row and if it is not the same collection year as we're expecting, we send a failure email and don't create the logs.

### DIFF
--- a/app/services/bulk_upload/lettings/year2023/csv_parser.rb
+++ b/app/services/bulk_upload/lettings/year2023/csv_parser.rb
@@ -1,6 +1,8 @@
 require "csv"
 
 class BulkUpload::Lettings::Year2023::CsvParser
+  include CollectionTimeHelper
+
   FIELDS = 134
   MAX_COLUMNS = 142
 
@@ -61,6 +63,12 @@ class BulkUpload::Lettings::Year2023::CsvParser
     max_columns_count > MAX_COLUMNS
   end
 
+  def correct_template_for_year?
+    collection_start_year_for_date(first_record_start_date) == 2023
+  rescue Date::Error
+    true
+  end
+
 private
 
   def default_field_numbers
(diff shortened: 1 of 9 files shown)


app/services/bulk_upload/lettings/year2023/csv_parser.rb
```diff
@@ -61,6 +63,12 @@ def too_many_columns?
     max_columns_count > MAX_COLUMNS
   end
 
+  def correct_template_for_year?
+    collection_start_year_for_date(first_record_start_date) == 2023
+  rescue Date::Error
+    true
```
jamesrose: Would the correct rescue not return false here, as in, we couldn't construct a valid start date to determine validity?

app/services/bulk_upload/lettings/year2023/csv_parser.rb
```diff
@@ -61,6 +63,12 @@ def too_many_columns?
     max_columns_count > MAX_COLUMNS
   end
 
+  def correct_template_for_year?
+    collection_start_year_for_date(first_record_start_date) == 2023
```
jamesrose: We could extract the year here to allow us to easily duplicate these files for the next year?

```suggestion
    collection_start_year_for_date(first_record_start_date) == FORM_YEAR
```

app/services/bulk_upload/lettings/year2023/csv_parser.rb
kosiakkatrina: So, the start date for 2023 would need to be constructed from different fields. This is using the field numbers for 22/23 and checking if a 22/23 date can be constructed and if yes, then we know for sure that this is a template for 22/23. Otherwise we default to saying that the template is correct, cause there could be errors, for example the user forgot to fill in a part of the date etc

app/services/bulk_upload/lettings/year2023/csv_parser.rb
kosiakkatrina: Can do!

app/services/bulk_upload/lettings/year2023/csv_parser.rb
jamesrose: Ah, this method is a bit more lenient is that it assumes the template is correct if anything goes wrong.

How about calling it `assumed_correct_template_for_year?`?

kosiakkatrina: How about now, I reversed it and changed to wrong_template_for_year?

jamesrose (approved):

PR #1670, opened 2023-05-25
CLDC-2330 Fix again

asmega:
# Context

- https://digital.dclg.gov.uk/jira/browse/CLDC-2330
- Ensure if user goes back in the how to fix journey they are forced to continue forwards in the journey

# Changes

- Add preflight check to how to fix journey and redirect if preflight fails ie they have already chosen to fix inline
- Set `Cache-Control` headers in this journey to `no-cache` so browser back button will be forced to re-request the page
- Add new `chosen` page to journey which outlines they must continue on their selected journey

### DIFF
--- a/app/models/forms/bulk_upload_lettings_soft_validations_check/chosen.rb
+++ b/app/models/forms/bulk_upload_lettings_soft_validations_check/chosen.rb
@@ -1,0 +1,31 @@
+module Forms
+  module BulkUploadLettingsSoftValidationsCheck
+    class Chosen
+      include ActiveModel::Model
+      include ActiveModel::Attributes
+      include Rails.application.routes.url_helpers
+
+      attribute :bulk_upload
+
+      def view_path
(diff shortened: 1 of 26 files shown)


kosiakkatrina (approved):

app/models/forms/bulk_upload_lettings_soft_validations_check/chosen.rb
```diff
@@ -0,0 +1,31 @@
+module Forms
+  module BulkUploadLettingsSoftValidationsCheck
+    class Chosen
```
ghost, line 3: what does Chosen mean in this context?

kosiakkatrina: So in here, it's a model for a form page. This page would get redirected to if the correction method for the bulk upload has already been chosen

ghost (approved):

PR #1671, opened 2023-05-25
CLDC-2372 Update bulk upload resources

kosiakkatrina:
Original description: Update bulk upload resources for sales and change ordering

Edit by @natdeanlewissoftwire: have updated a few files since Kat left this as specified on the ticket (https://digital.dclg.gov.uk/jira/browse/CLDC-2372)

I've also added a collection resources helper to simplify the metadata creation, didn't include automatic page count calculation as that would be best done with a gem so have just left this as a parameter for simplicity.


jamesrose (approved):

jackscotti (approved):

ghost (approved):

ghost (approved):

app/views/layouts/_collection_resources.html.erb
```diff
@@ -7,50 +7,65 @@
...
+      href: download_23_24_sales_bulk_upload_specification_path,
+      metadata: metadata_helper("bulk-upload-sales-specification-2023-24.xlsx"),
     },
   ]) %>
 
   <h3 class="govuk-heading-s">Lettings 2022/23</h3>
   <%= render DocumentListComponent.new(items: [
     {
       name: "Lettings log for tenants (2022/23)",
-      href: "https://core.communities.gov.uk/public/download/guides-and-manuals/2022-23%20Lettings%20paper%20form.pdf?download-format=pdf",
```
natdeanlewissoftwire, line 33: I think this is fine, but happy to have a second opinion in case there's a reason for this being pulled from core that I've missed

app/services/feature_toggle.rb
```diff
@@ -25,7 +25,7 @@ def self.bulk_upload_lettings_logs?
   end
 
   def self.bulk_upload_sales_logs?
-    !Rails.env.production?
+    true
```
jamesrose: This is an important change so should be in a separate PR.

app/helpers/collection_resources_helper.rb
```diff
@@ -0,0 +1,13 @@
...
+    extension = file.split(".")[-1]
+    type_string = case extension
+                  when "xlsx" then "Microsoft Excel"
+                  when "pdf" then "PDF"
+                  else extension
+                  end
+    size = number_to_human_size(File.size("public/files/#{file}"), precision: 0, significant: false)
+    pages_string = pages.present? ? pluralize(pages, "page") : nil
+    [type_string, size, pages_string].compact.join(", ")
+  end
```
jamesrose: Can you think of a name that describes what it does? (e.g., it probably shouldn't include `_helper`) 

```suggestion
  def file_type_size_and_pages(file, number_of_pages: nil)
    extension_mapping = {
      "xlsx" => "Microsoft Excel",
      "pdf"  => "PDF",
    }
    extension = File.extname(file)[1..-1]
    
    file_type = extension_mapping.fetch(extension, extension)
    file_size = number_to_human_size(File.size("public/files/#{file}"), precision: 0, significant: false)
    file_pages = number_of_pages ? pluralize(number_of_pages, "page") : ""
    
    [file_type, file_size, file_pages].join(", ")
  end
```

natdeanlewissoftwire: sure, good point

app/services/feature_toggle.rb
natdeanlewissoftwire: No problem

PR #1674, opened 2023-05-26
CLDC-2383 Sidekiq graceful restart

asmega:
# Context

- https://digital.dclg.gov.uk/jira/browse/CLDC-2383
- cloudfoundry sends shutdown signal to processes including sidekiq killing running jobs

# Changes

- lower sidekiq timeout from the default 25 seconds to 3 seconds. cloudfoundry only gives us 10 seconds before killing the process. 3 seconds should give some time to finish existing jobs
- when shutdown signal is given we tell sidekiq to shutdown quietly. this way jobs are cancelled and re-queued so they can be picked by the newly deployed worker
- for long running jobs ie bulk upload. we clear down existing errors generated before starting incase these are from a previous run. this to a degree will help with if the job if re-queued part way thru the process

# Notes

- for testing I was able to:
  - start a big upload
  - take note of the increasing count of errors
  - stop the box
  - start the box
  - take note of the error count starting again from zero then increasing again
- this show evidence on stopping, it does re-queue the job which is then picked up later by the worker again. and that the code to clear the errors to start again is also working as expected

### DIFF
--- a/app/services/bulk_upload/processor.rb
+++ b/app/services/bulk_upload/processor.rb
@@ -6,6 +6,8 @@ class BulkUpload::Processor
   end
 
   def call
+    destroy_any_existing_errors_from_prior_run
+
     download
 
     return send_failure_mail(errors: validator.errors.full_messages) if validator.invalid?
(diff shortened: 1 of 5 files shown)


jamesrose (approved):
Awesome. Nice work!

PR #1675, opened 2023-05-26
CLDC-2257 Refactor filters

natdeanlewissoftwire:
General refactor of filters, makes a general filter manager that handles filters separately for sales vs lettings.
This means sales filters are the same in the secondary nav tab or primary nav tab, same for lettings. This can be changed easily by updating the organisation controller if we want.

ticket: https://digital.dclg.gov.uk/jira/browse/CLDC-2257


app/models/filter_manager.rb
```diff
@@ -0,0 +1,33 @@
...
+  end
+
+  def session_filters(filter_type, specific_org: false)
+    @session_filters ||= deserialize_filters_from_session(filter_type, specific_org)
+  end
+
+  def deserialize_filters_from_session(filter_type, specific_org)
+    current_filters = session["#{filter_type}_filters"]
+    new_filters = current_filters.present? ? JSON.parse(current_filters) : {}
+    if filter_type.include?("logs")
```
natdeanlewissoftwire: alternative is schemes but we ignore until that functionality is here

app/models/filter_manager.rb
```diff
@@ -0,0 +1,33 @@
...
+      current_user.logs_filters(specific_org:).each do |filter|
+        new_filters[filter] = params[filter] if params[filter].present?
+      end
+    end
+    params["organisation_select"] == "all" ? new_filters.except("organisation") : new_filters
+  end
+
+  def filtered_logs(logs, search_term, filters)
+    all_orgs = params["organisation_select"] == "all"
```
arfacamble: I don't believe params[:organisation_select] is ever equal to "all"

app/controllers/lettings_logs_controller.rb
```diff
@@ -113,6 +114,18 @@ def update_logs
...
+  def set_session_filters
+    filter_manager.serialize_filters_to_session("lettings_logs")
+  end
+
+  def session_filters
+    filter_manager.session_filters("lettings_logs")
+  end
+
+  def filter_manager
+    FilterManager.new(current_user:, session:, params:)
```
jamesrose: All but one of the public methods in this class accept the filter type as the first argument. Did you consider initializing the class with this value?

app/models/filter_manager.rb
natdeanlewissoftwire: I believe it is

app/controllers/lettings_logs_controller.rb
natdeanlewissoftwire: Good point, think I'll do that

app/models/filter_manager.rb
```diff
@@ -0,0 +1,33 @@
...
+  attr_reader :current_user, :session, :params
+
+  def initialize(current_user:, session:, params:)
+    @current_user = current_user
+    @session = session
+    @params = params
+  end
+
+  def serialize_filters_to_session(filter_type, specific_org: false)
+    session["#{filter_type}_filters"] = session_filters(filter_type, specific_org:).to_json
```
jamesrose: How about? 

```suggestion
    session[session_name_for(filter_type)] = session_filters(filter_type, specific_org:).to_json
```

app/helpers/filters_helper.rb
```diff
@@ -1,8 +1,8 @@
 module FiltersHelper
-  def filter_selected?(filter, value)
-    return false unless session[:logs_filters]
+  def filter_selected?(filter, value, filter_type)
```
jamesrose, line 2: Could these methods be merged into `FilterManager?`

natdeanlewissoftwire: They could, but I think it makes more sense to have the view helpers separate rather than all mixed together - what's the benefit of having them merged in?

jamesrose (approved):

app/models/filter_manager.rb
arfacamble: I would love you to show me when. I have snooped in the params on all user actions I could think of and never found it equal to all, although it seems reasonable for it to be that sometimes

app/services/filter_manager.rb
```diff
@@ -0,0 +1,78 @@
...
+    else
+      base_collection
+    end
+  end
+
+  def self.filter_logs(logs, search_term, filters, all_orgs, user)
+    logs = filter_by_search(logs, search_term)
+
+    filters.each do |category, values|
+      next if Array(values).reject(&:empty?).blank?
```
arfacamble, line 23: Array(values).compact_blank?
I think

app/models/filter_manager.rb
natdeanlewissoftwire: So it's added to the params as part of the form in _log_filters.rb as far as I can see, and you can see it in the url if you apply filters, but this filter isn't added to session filters if it's "all" (at the end of deserialize_filters_from_session).

app/services/filter_manager.rb
natdeanlewissoftwire: good point this can be simplified, have gone for values.each(&:empty?)

natdeanlewissoftwire: actually this needs to be this to handle nil/strings/different kinds of arrays, I could fiddle more with it but am just going to leave it for now - if we change the actual structure of filters later on/change how we serialize/deserialize then this will become worth changing

PR #1676, opened 2023-05-30
password reset error bugfix

arfacamble:
avoid error if email is nil

sentry error alert from production
https://dluhc-core.sentry.io/issues/4211300771/?referrer=slack

### DIFF
--- a/app/controllers/auth/passwords_controller.rb
+++ b/app/controllers/auth/passwords_controller.rb
@@ -4,7 +4,7 @@ class Auth::PasswordsController < Devise::PasswordsController
   def reset_confirmation
     self.resource = resource_class.new
     @email = params["email"]
-    if @email.empty?
+    if @email.blank?
       resource.errors.add :email, I18n.t("validations.email.blank")
       render "devise/passwords/new", status: :unprocessable_entity
     elsif !email_valid?(@email)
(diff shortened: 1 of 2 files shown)


RachaelBooth (commented):
Looks sensible to me. 

@arfacamble Is there a ticket reference or anything that should be in the PR name before merge? (to help anyone looking back at commits). I'm happy to approve as is if there isn't.

app/controllers/auth/passwords_controller.rb
```diff
@@ -4,7 +4,7 @@ class Auth::PasswordsController < Devise::PasswordsController
   def reset_confirmation
     self.resource = resource_class.new
     @email = params["email"]
-    if @email.empty?
+    if @email.blank?
```
jamesrose, line 7: Is there a spec for this already? If not, could you please add one?

arfacamble:
> 

apologies slow reply @RachaelBooth  this slipped off my radar.
there is no ticket for this. There was a sentry alert but it was a while ago now so I have lost it.
Yes @jamesrose  I will write a test for this. It was a production bug so I chucked somethign up quick in case it recurred and then slightyl forgot about it

jamesrose (approved):

PR #1679, opened 2023-05-30
CLDC-2322 read and sign data sharing agreement

ghost:
https://digital.dclg.gov.uk/jira/browse/CLDC-2322

- Add dynamic data sharing agreement template
- Update details page to include a row for the data sharing agreement
- Update notification banner to dynamically handle a text paragraph
- persist DPO and org data on table to prevent future updates

### DIFF
--- a/app/helpers/data_sharing_agreement_helper.rb
+++ b/app/helpers/data_sharing_agreement_helper.rb
@@ -1,0 +1,22 @@
+module DataSharingAgreementHelper
+  def data_sharing_agreement_row(user:, organisation:, summary_list:)
+    summary_list.row do |row|
+      row.key { "Data Sharing Agreement" }
+      row.value { organisation.data_sharing_agreement.present? ? "Accepted" : "Not accepted" }
+      row.action(
+        href: data_sharing_agreement_organisation_path(organisation),
+        text: "View agreement",
+      )
+    end
+  end
+
+  def name_for_data_sharing_agreement(data_sharing_agreement, user)
+    if data_sharing_agreement.present?
+      data_sharing_agreement.data_protection_officer.name
+    elsif user.is_dpo?
+      user.name
+    else
+      "[DPO name]"
+    end
+  end
+end
--- a/app/helpers/data_sharing_agreement_helper.rb
+++ b/app/helpers/data_sharing_agreement_helper.rb
@@ -2,11 +2,22 @@ module DataSharingAgreementHelper
   def data_sharing_agreement_row(user:, organisation:, summary_list:)
     summary_list.row do |row|
       row.key { "Data Sharing Agreement" }
-      row.value { organisation.data_sharing_agreement.present? ? "Accepted" : "Not accepted" }
       row.action(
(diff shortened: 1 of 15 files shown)


app/helpers/data_sharing_agreement_helper.rb
```diff
@@ -0,0 +1,53 @@
...
+    end
+  end
+
+  def name_for_data_sharing_agreement(data_sharing_agreement, user)
+    if data_sharing_agreement.present?
+      data_sharing_agreement.data_protection_officer.name
+    elsif user.is_dpo?
+      user.name
+    else
+      "[DPO name]"
```
natdeanlewissoftwire, line 30: don't we want this to be the actual name of the DPO?

app/views/organisations/data_sharing_agreement.html.erb
```diff
@@ -0,0 +1,139 @@
...
+      <% if @data_sharing_agreement %>
+        This agreement is made the <%= @data_sharing_agreement.signed_at.day.ordinalize %> day of <%= @data_sharing_agreement.signed_at.strftime("%B") %> <%= @data_sharing_agreement.signed_at.year %>
+      <% elsif current_user.is_dpo? %>
+        This agreement is made the <%= Time.zone.now.day.ordinalize %> day of <%= Time.zone.now.strftime("%B") %> <%= Time.zone.now.year %>
+      <% else %>
+        This agreement is made the [XX] day of [XX] 20[XX]
+      <% end %>
+    </p>
+    <p class="govuk-body-m"><strong>between</strong></p>
+    <p class="govuk-body-m">1) [core data provider organisation] of [full address] (“CORE Data Provider”)</p>
```
natdeanlewissoftwire: shouldn't these [] be populated with actual data?

natdeanlewissoftwire (commented):
have done a first pass, looks good - have a feeling my queries are just me misunderstanding the ticket but I've left them in anyway

app/helpers/data_sharing_agreement_helper.rb
ghost: yes but but only when not DPO or when not signed!

app/views/organisations/data_sharing_agreement.html.erb
ghost: apparently not, only certain bits in the document need to be replaced

natdeanlewissoftwire (approved):

app/views/organisations/data_sharing_agreement.html.erb
ghost: I was wrong, these need updating! Will fix shortly

app/views/organisations/data_sharing_agreement.html.erb
```diff
@@ -15,7 +15,11 @@
       <% end %>
```
natdeanlewissoftwire, line 15: maybe some of this logic should go in a helper

natdeanlewissoftwire (approved):
looks good, possible case for a helper but other than that nothing from me :)

PR #1680, opened 2023-05-31
CLDC 2328: Resend invitation button

Airk0n:
### **Change**
This ticket is concerned with moving the existing 'deactivate'/'reactivate' buttons to the bottom of the page as well as adding a button to resend invitations to new users who's details have not been confirmed via email.

### **Old design :**
Old

### **New design :**
New

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -4,9 +4,9 @@ class UsersController < ApplicationController
   include Helpers::Email
   include Modules::SearchFilter
 
-  before_action :authenticate_user!
+  before_action :authenticate_user!, except: %i[resend_invite]
   before_action :find_resource, except: %i[new create]
-  before_action :authenticate_scope!, except: %i[new]
+  before_action :authenticate_scope!, except: %i[new resend_invite]
 
   def index
     redirect_to users_organisation_path(current_user.organisation) unless current_user.support?
@@ -27,6 +27,12 @@ class UsersController < ApplicationController
         end
       end
     end
+  end
+
+  def resend_invite
+    @user.send_confirmation_instructions
+    flash[:notice] = "Invitation sent"
+    render :show
   end
 
   def show; end
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -4,9 +4,9 @@ class UsersController < ApplicationController
   include Helpers::Email
   include Modules::SearchFilter
 
-  before_action :authenticate_user!, except: %i[resend_invite]
+  before_action :authenticate_user!
   before_action :find_resource, except: %i[new create]
-  before_action :authenticate_scope!, except: %i[new resend_invite]
+  before_action :authenticate_scope!, except: %i[new]
 
   def index
     redirect_to users_organisation_path(current_user.organisation) unless current_user.support?
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -31,7 +31,7 @@ class UsersController < ApplicationController
 
   def resend_invite
     @user.send_confirmation_instructions
-    flash[:notice] = "Invitation sent"
+    flash[:notice] = "Invitation sent to #{@user.email}"
     render :show
   end
 
--- a/app/views/users/show.html.erb
+++ b/app/views/users/show.html.erb
@@ -5,17 +5,6 @@
     <h1 class="govuk-heading-l">
       <%= content_for(:title) %>
     </h1>
-    <p>
-      <% if current_user.can_toggle_active?(@user) %>
-        <% if @user.active? %>
-          <%= govuk_link_to "Deactivate user", "/users/#{@user.id}/deactivate" %>
-        <% else %>
-          <span class="app-!-colour-muted govuk-!-margin-right-2">
-            This user has been deactivated. <%= govuk_link_to "Reactivate user", "/users/#{@user.id}/reactivate" %>
-          </span>
-        <% end %>
-      <% end %>
-    </p>
 
     <h2 class="govuk-heading-m">
       Personal details
@@ -103,5 +92,30 @@
             end
           end %>
     <% end %>
+
+    <div class="govuk-button-group">
+      <p>
+        <% if current_user.can_toggle_active?(@user) %>
+          <% if @user.active? %>
+            <%= govuk_button_link_to "Deactivate user", deactivate_user_path(@user), warning: true %>
+          <% else %>
+            <span class="app-!-colour-muted govuk-!-margin-right-2">
+              This user has been deactivated. <%= govuk_button_link_to "Reactivate user", reactivate_user_path(@user) %>
+            </span>
+          <% end %>
+        <% end %>
+
+        <% if current_user.can_toggle_active?(@user) %>
+          <% if @user.active? %>
+            <%= govuk_button_link_to "Resend invite link", resend_invite_user_path(@user) %>
+          <% else %>
+            <span class="app-!-colour-muted govuk-!-margin-right-2">
+              This user has been deactivated. <%= govuk_link_to "Reactivate user", "/users/#{@user.id}/reactivate" %>
+            </span>
+          <% end %>
+        <% end %>
+      </p>
+    </div>
+
   </div>
 </div>
--- a/app/views/users/show.html.erb
+++ b/app/views/users/show.html.erb
@@ -96,12 +96,14 @@
     <div class="govuk-button-group">
       <% if current_user.can_toggle_active?(@user) %>
         <% if @user.active? %>
+          <% if current_user.support? %>
+            <%= govuk_button_to "Resend invite link", resend_invite_user_path(@user) %>
+          <% end %>
           <%= govuk_button_link_to "Deactivate user", deactivate_user_path(@user), warning: true %>
-          <%= govuk_button_link_to "Resend invite link", resend_invite_user_path(@user) %>
         <% else %>
-          <span class="app-!-colour-muted govuk-!-margin-right-2">
-            This user has been deactivated. <%= govuk_button_link_to "Reactivate user", reactivate_user_path(@user) %>
-          </span>
+            <span class="app-!-colour-muted govuk-!-margin-right-2">
+              This user has been deactivated. <%= govuk_button_link_to "Reactivate user", reactivate_user_path(@user) %>
+            </span>
         <% end %>
       <% end %>
     </div>
--- a/app/views/users/show.html.erb
+++ b/app/views/users/show.html.erb
@@ -96,10 +96,10 @@
     <div class="govuk-button-group">
       <% if current_user.can_toggle_active?(@user) %>
         <% if @user.active? %>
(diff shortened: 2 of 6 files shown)


arfacamble:
James likes tickets to have full sentences and the start has hyphens so here it might be something like:
CLDC-2328-add-resend-invite-button-to-user-profile-page

app/controllers/users_controller.rb
```diff
@@ -4,9 +4,9 @@ class UsersController < ApplicationController
   include Helpers::Email
   include Modules::SearchFilter
 
-  before_action :authenticate_user!
+  before_action :authenticate_user!, except: %i[resend_invite]
```
arfacamble: I'm sceptical this should be here. It's essentially saying that before the resend_invite method runs, we don't need to check whether a user is logged in, but a user would definitely need to be logged in to do that.

app/controllers/users_controller.rb
```diff
@@ -4,9 +4,9 @@ class UsersController < ApplicationController
...
   before_action :find_resource, except: %i[new create]
-  before_action :authenticate_scope!, except: %i[new]
+  before_action :authenticate_scope!, except: %i[new resend_invite]
```
arfacamble: here, I think we should definitely not include resend_invite in the except.
I think it's possible that we will need to adjust the `authenticate_scope!` method with reference to what type of user has the necessary permissions to do the thing that is being done.

app/controllers/users_controller.rb
```diff
@@ -29,6 +29,12 @@ def index
     end
   end
 
+  def resend_invite
+    @user.send_confirmation_instructions
+    flash[:notice] = "Invitation sent"
+    render :show
+  end
```
arfacamble, line 36: looks good

app/views/users/show.html.erb
```diff
@@ -103,5 +92,19 @@
             end
           end %>
     <% end %>
+
+    <div class="govuk-button-group">
+      <% if current_user.can_toggle_active?(@user) %>
```
arfacamble, line 97: we should check what happens in `can_toggle_active?` as it might be the info we want to amend the authenticate_scope method in the controller.
If the conditionals on this page are correct, then we may just want the scope for `resend_invite` to be the same as the method that is called on the deactivate_user_path

app/views/users/show.html.erb
```diff
@@ -103,5 +92,19 @@
...
+        <% if @user.active? %>
+          <%= govuk_button_link_to "Deactivate user", deactivate_user_path(@user), warning: true %>
+          <%= govuk_button_link_to "Resend invite link", resend_invite_user_path(@user) %>
```
arfacamble: looks good

arfacamble (commented):
looks good. one or two small amendments/discussions.

while I think of it I will write it, but we may both forget by the time it gets that far!
When this is approved, it will go to PO review and Jemma will test it in a review environment. In order for her to test it, we will need to log into a rails console for that review app and create her a user with her own email, and some other users with real names (eg me and you) that she can deactivate and reactivate and check whether correct emails are sent

app/controllers/users_controller.rb
arfacamble: more specifically if you removed this change I would expect it to continue working. Does it not?

spec/requests/users_controller_spec.rb
```diff
@@ -96,6 +96,13 @@
         expect(response).to redirect_to("/account/sign-in")
       end
     end
+
+    describe "#resend_invite" do
+      it "does not allow resending activation emails" do
+        get "/users/#{user.id}/deactivate", headers: headers, params: {}
```
arfacamble: I know it's largely the done thing in this project, but it is best practice to also use _path (eg here would it be `deactivate_user_path` ? ) helpers in spec files as you are then also implicitly testing your routing.

spec/requests/users_controller_spec.rb
```diff
@@ -123,6 +130,10 @@
           expect(page).not_to have_link("Deactivate user", href: "/users/#{user.id}/deactivate")
         end
 
+        it "does not allow resending invitation emails" do
+          expect(page).not_to have_link("Resend invite link", href: "/users/#{other_user.id}/resend-invite")
```
arfacamble: this is great. compare to line 82 where someone did something 💩  instead.

however, again change to _path helper please.

spec/requests/users_controller_spec.rb
```diff
@@ -132,6 +143,10 @@
           it "does not allow reactivating the user" do
             expect(page).not_to have_link("Reactivate user", href: "/users/#{user.id}/reactivate")
           end
+
+          it "does not allow resending invitation emails" do
+            expect(page).not_to have_link("Resend invite link", href: "/users/#{other_user.id}/resend-invite")
```
arfacamble: actually for all these tests asserting we expect .not_to etc. we don't need to specify the href, since it's already asserting there is no link like that.

spec/requests/users_controller_spec.rb
```diff
@@ -530,6 +553,10 @@
             expect(page).to have_link("Deactivate user", href: "/users/#{other_user.id}/deactivate")
           end
 
+          it "allows you to resend invitation emails" do
+            expect(page).to have_link("Resend invite link", href: "/users/#{other_user.id}/resend-invite")
```
arfacamble: contrast with this where yes we definitely do want the path specified.

config/routes.rb
```diff
@@ -103,6 +103,7 @@
     member do
       get "deactivate", to: "users#deactivate"
       get "reactivate", to: "users#reactivate"
+      get "resend-invite", to: "users#resend_invite"
```
arfacamble: It's just occurred to me that if we want to be properly RESTful, we may want to consider whether these should be get paths.
This is something I'm learning about at the moment so I would be interested in what @jamesrose or @RachaelBooth have to say about this.
I feel like deactivate and reactivate are changing the state of the model in question so they should probably be patch? (that would be outside the scope of this ticket anyway)
Does resend invite update anything on the user? Like a last invited field or anything? If an action has a side effect such as in this case sending an email ... ?
My instinct is a post would be most appropriate but I'm not confident in that!

RachaelBooth: On deactivate and reactivate, you're right that we shouldn't expect a get request to be changing an object - but I think in this case those paths are actually opening an "are you sure" type page rather than being endpoints that directly do the de-/reactivation, which means that a get is expected even though it looks a little odd here.

Resend invite is a bit harder to obviously fit to an http verb, because it's doing an action (i.e. not just reading data) but not creating or updating the resource on our server that it is pointing at. I think it should be a post - the intended effect is to send an email which is not a read-only action and not one where repeating a request to the endpoint is without consequence (i.e. as long as we include emails to the user in things we're considering, this request is neither safe nor idempotent - and whatever we consider we don't want browsers treating it as something it can do without user input). 

Side node: this doesn't mean that gets can't have side effects - we could have counters or last read times or request logging or things like that being updated on a get and that would be fine, as long as there was nothing where we would care if it happened an extra time. (Also, a properly RESTful way might be to post to an email resource, but that is almost certainly not what we want to do.)

arfacamble: ok great. very interesting. I haven't got a great handle on concept of idempotence. The resend method does update something on the user model, just to set something like initial_invitation_sent to true. As I understand it that is idempotent as however many times you click that button, it just gets set to true, so the same result.

app/views/users/show.html.erb
Airk0n: I can't parse that, I don't have a memory of wanting to amend the authenticate scope, lets chat about that.

spec/requests/users_controller_spec.rb
Airk0n: All absolute paths changed to paths in upcoming commit

spec/requests/users_controller_spec.rb
Airk0n: Oh 82 is fun yeah. 
All paths changed I to path helper in upcoming commit

spec/requests/users_controller_spec.rb
Airk0n: Yup, this is one of them following the convention of the code things, as you can see reactivate user on line 132/143 also provides a path.

config/routes.rb
Airk0n: So we're agreed that I change this from `get` to `post`?
Is there anything else I need to change and are there any knock on effects or other things I should be aware of?

spec/requests/users_controller_spec.rb
jamesrose: There are a lot of unhappy path tests here, but do you have one that ensures the confirmation method is called for a given user?

app/views/users/show.html.erb
```diff
@@ -96,12 +96,14 @@
...
         <% else %>
-          <span class="app-!-colour-muted govuk-!-margin-right-2">
-            This user has been deactivated. <%= govuk_button_link_to "Reactivate user", reactivate_user_path(@user) %>
-          </span>
+            <span class="app-!-colour-muted govuk-!-margin-right-2">
+              This user has been deactivated. <%= govuk_button_link_to "Reactivate user", reactivate_user_path(@user) %>
+            </span>
         <% end %>
       <% end %>
     </div>
```
Airk0n, line 109: I'm not a fan of this level of indentation and if/else statements. 
I'm guessing a partial or something similar like a component that takes arguments is a a solution? (though perhaps a bit much for this)

spec/features/user_spec.rb
```diff
@@ -452,6 +452,41 @@
...
+          link: include("/account/confirmation?confirmation_token=#{other_user.confirmation_token}"),
+        }
+      end
+
+      before do
+        other_user.update!(confirmation_token: "abc")
+        visit(lettings_logs_path)
+        fill_in("user[email]", with: user.email)
+        fill_in("user[password]", with: "pAssword1")
+        click_button("Sign in")
```
arfacamble:
```suggestion
        sign_in user
```

spec/features/user_spec.rb
```diff
@@ -452,6 +452,41 @@
         expect(page).to have_css(".govuk-notification-banner.govuk-notification-banner--success")
       end
     end
+
+    context "when reinviting a user after initial confirmation email has been sent" do
+      let!(:other_user) { FactoryBot.create(:user, name: "new user", organisation: user.organisation, email: "new_user@example.com", initial_confirmation_sent: true) }
```
arfacamble: very minor refactor
we can remove 469 and add `confirmation_token: "abc"` in the factory
we can remove `FactoryBot.` as create is aliased to be called on FactoryBot in the spec config somewhere.
line 474 we can refactor to `user_path(other_user)` and we can remove the ! from the let here on 457 as the other_user will be created when we attempt to navigate to that path which is fine

app/views/users/show.html.erb
arfacamble: not amazing, but not worth fuss

arfacamble (approved):

PR #1682, opened 2023-05-31
CLDC-1908 Update lettings compound validation

natdeanlewissoftwire:
Just adding a validation on sales to lettings too. Also, we allow ecstat to be refused OR student to pass this validation

ticket: https://digital.dclg.gov.uk/jira/browse/CLDC-1908

### DIFF
--- a/app/models/validations/household_validations.rb
+++ b/app/models/validations/household_validations.rb
@@ -187,10 +187,20 @@ private
     relationship = record.public_send("relat#{person_num}")
     return unless age && economic_status && relationship
 
-    if age >= 16 && age <= 19 && tenant_is_child?(relationship) && (!tenant_is_fulltime_student?(economic_status) && !tenant_economic_status_refused?(economic_status))
-      record.errors.add "ecstat#{person_num}", I18n.t("validations.household.ecstat.not_student_16_19", person_num:)
-      record.errors.add "age#{person_num}", I18n.t("validations.household.age.not_student_16_19", person_num:)
-      record.errors.add "relat#{person_num}", I18n.t("validations.household.relat.not_student_16_19", person_num:)
+    age_between_16_19 = age.between?(16, 19)
+    student = tenant_is_fulltime_student?(economic_status)
+    child = tenant_is_child?(relationship)
+
+    if age_between_16_19 && !student && child
+      record.errors.add "age#{person_num}", I18n.t("validations.household.age.student_16_19.cannot_be_16_19.child_not_student")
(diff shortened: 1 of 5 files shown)


spec/models/validations/household_validations_spec.rb
```diff
@@ -363,6 +363,29 @@
       end
     end
 
+    it "does not add an error for a person aged 16-19 who is a student but not a child of the buyer" do
```
RachaelBooth: This wants re-wording to not refer to a buyer (I assume this is coming from the similar sales validation tests?)

RachaelBooth: Is it worth also having a test for "age > 19, is student, not child is fine" (with better wording), or is that covered sufficiently elsewhere?

app/models/validations/household_validations.rb
```diff
@@ -187,10 +187,21 @@ def validate_person_age_and_relationship_matches_economic_status(record, person_
...
     return unless age && economic_status && relationship
 
-    if age >= 16 && age <= 19 && tenant_is_child?(relationship) && (!tenant_is_fulltime_student?(economic_status) && !tenant_economic_status_refused?(economic_status))
-      record.errors.add "ecstat#{person_num}", I18n.t("validations.household.ecstat.not_student_16_19", person_num:)
-      record.errors.add "age#{person_num}", I18n.t("validations.household.age.not_student_16_19", person_num:)
-      record.errors.add "relat#{person_num}", I18n.t("validations.household.relat.not_student_16_19", person_num:)
+    age_between_16_19 = age.between?(16, 19)
+    student = tenant_is_fulltime_student?(economic_status)
+    economic_status_refused = tenant_economic_status_refused?(economic_status)
+    child = tenant_is_child?(relationship)
```
RachaelBooth, line 193: 😄 Much nicer to read...

RachaelBooth (commented):
Couple of small comments, otherwise looks good.

Few things that slightly urk me in general but not from this PR - there is code very much duplicated between the sales and lettings household validations, and sales things living in a subfolder when lettings things don't - but neither of those are problems to solve in this PR. I'm just trying to consciously notice these things before I get used to them, so very much not on you Nat!

spec/models/validations/household_validations_spec.rb
natdeanlewissoftwire: Yep thanks, good spot

natdeanlewissoftwire: Sure, have added :)

ghost (approved):

PR #1684, opened 2023-06-02
CLDC-2319 prevent log creation

ghost:
- add data sharing agreement (DSA) banner
- hide create log and upload buttons when DSA not signed
- add model validation to prevent log creation if DSA not signed
- Drop newly introduced redundant DSA model and use old DataProtectionConfrimation instead
- Generalise tests to that they test for the right thing and are more flexible in case factory behaviour changes like in this case


app/components/create_log_actions_component.html.erb
```diff
@@ -0,0 +1,10 @@
+<% if display_actions? %>
```
natdeanlewissoftwire, line 1: feel like I have to suggest a helper as there are nested ifs, just my personal limit (also occurs in app/views/logs/_create_for_org_actions.html.erb for example)

app/components/create_log_actions_component.rb
```diff
@@ -0,0 +1,55 @@
...
+  def create_button_href
+    case log_type
+    when "lettings"
+      lettings_logs_path
+    when "sales"
+      sales_logs_path
+    end
+  end
+
+  def create_button_copy
```
natdeanlewissoftwire, line 31: this and the below copy function are presentation-related so should probably go in a helper too

app/models/log.rb
```diff
@@ -9,6 +9,7 @@ class Log < ApplicationRecord
   belongs_to :bulk_upload, optional: true
 
   before_save :update_status!
+  before_validation :verify_dsa_signed, on: :create
```
natdeanlewissoftwire: can we use the full name in this method name for clarity to people who won't know what dsa means?

app/models/organisation.rb
```diff
@@ -89,7 +88,7 @@ def rent_period_labels
   end
 
   def data_protection_confirmed?
-    !!data_protection_confirmations.order(created_at: :desc).first&.confirmed
+    !!data_protection_confirmation&.confirmed?
```
natdeanlewissoftwire, line 92: is !! necessary here, confirmed is always a bool/nil already right?

app/models/organisation.rb
```diff
@@ -103,16 +102,21 @@ def display_provider_type
   end
 
   def display_organisation_attributes
-    [
+    attrs = [
```
natdeanlewissoftwire, line 106: i don't see a reason for this abbreviation, prefer attributes for clarity :)

app/views/logs/_log_filters.html.erb
```diff
@@ -11,7 +11,7 @@
         <% if bulk_upload_options(@bulk_upload).present? %>
           <%= render partial: "filters/checkbox_filter",
                      locals: {
-                       f: f,
+                       f:,
```
natdeanlewissoftwire, line 14: nice 👍

app/models/organisation.rb
natdeanlewissoftwire: Is this method even really needed anymore now we're not ordering?

db/seeds.rb
```diff
@@ -7,6 +7,14 @@
 #   Character.create(name: 'Luke', movie: movies.first)
 
 # rubocop:disable Rails/Output
+def create_dsa(user)
```
natdeanlewissoftwire: i prefer the full phrase for clarity rather than dsa, mentioned above but thought I'd flag here too in case you decide to change :)

natdeanlewissoftwire: I feel fine about DPO, i think that's a more common acronym and easier to google

spec/views/organisations/data_sharing_agreement.html.erb_spec.rb
```diff
@@ -86,30 +88,32 @@
...
-          signed_at: Time.zone.now - 1.day,
+          created_at: Time.zone.now - 1.day,
         )
       end
 
+      let(:dpo) { data_protection_confirmation.data_protection_officer }
+
       it "renders dynamic content" do
         render
         # sees signed_at date
```
natdeanlewissoftwire, line 104: I know this isn't you but I'm not sure these comments are necessary, what do you think?

app/helpers/data_sharing_agreement_helper.rb
```diff
@@ -21,32 +21,32 @@ def data_sharing_agreement_row(user:, organisation:, summary_list:)
...
 
   # rubocop:disable Rails/HelperInstanceVariable
-  def section_12_2(data_sharing_agreement:, user:, organisation:)
-    if data_sharing_agreement
-      @org_address = data_sharing_agreement.organisation_address
-      @org_name = data_sharing_agreement.organisation_name
-      @org_phone = data_sharing_agreement.organisation_phone_number
-      @dpo_name = data_sharing_agreement.dpo_name
-      @dpo_email = data_sharing_agreement.dpo_email
+  def section_12_2(data_protection_confirmation:, user:, organisation:)
```
natdeanlewissoftwire: I don't know what this method name means, I assume it's a clause that the DSA refers to? Is there a better name for people like me unfamiliar with what this is?

natdeanlewissoftwire (approved):
a smattering of comments, mostly minor, lgtm otherwise!

app/components/create_log_actions_component.rb
ghost: these are view components, I think they're made for this kind of logic and testing

app/models/organisation.rb
ghost: it's being used as a replacemend for having to run every time `data_protection_confirmation&.confirmed?`

db/seeds.rb
ghost: will update 👍

app/models/log.rb
ghost: will update 👍

app/components/create_log_actions_component.html.erb
ghost: this is a view component which is what we should be using over helpers as we can more easily test them

app/models/organisation.rb
ghost: attributes might be a reserved word within a model, I will update if not

ghost: it kinda is, so not renaming this for now

app/controllers/bulk_upload_lettings_logs_controller.rb
```diff
@@ -1,5 +1,6 @@
 class BulkUploadLettingsLogsController < ApplicationController
   before_action :authenticate_user!
+  before_action :validate_data_protection_agrement_signed!
```
arfacamble, line 3: small typo in agreement

PR #1685, opened 2023-06-05
CLDC-2438 Fix carriage return csv bug

natdeanlewissoftwire:
Small bug fix from an alert we saw last week. Seems to work on test files I've generated, but I'm not sure if we can get the file that caused the original error, would be nice to test using that if we can?

Pretty straightforward fix though so probably ok regardless.

Update - this has been tested on the original file, and the bug is now fixed as expected.

ticket: https://digital.dclg.gov.uk/jira/browse/CLDC-2438

### DIFF
--- a/app/services/bulk_upload/lettings/year2023/csv_parser.rb
+++ b/app/services/bulk_upload/lettings/year2023/csv_parser.rb
@@ -97,6 +97,8 @@ private
 
     @normalised_string = File.read(path, encoding: "bom|utf-8")
     @normalised_string.gsub!("\r\n", "\n")
+    @normalised_string.gsub!("\r\r", "\n")
+    @normalised_string.tr!("\r", "\n")
     @normalised_string.scrub!("")
 
     @normalised_string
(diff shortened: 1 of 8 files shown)


jamesrose:
@natdeanlewissoftwire The files are indeed stored on S3. Would you like to pair to pull it down so you can test?

app/services/bulk_upload/lettings/year2023/csv_parser.rb
```diff
@@ -97,6 +97,8 @@ def normalised_string
 
     @normalised_string = File.read(path, encoding: "bom|utf-8")
     @normalised_string.gsub!("\r\n", "\n")
+    @normalised_string.gsub!("\r\r", "\n")
+    @normalised_string.tr!("\r", "\n")
```
jamesrose: Does this do the same thing as the previous line?

natdeanlewissoftwire: Not exactly - line 100 replaces \r\r with a single newline to avoid clutter but it's not necessary imo so I think I'll remove, good point

jamesrose (approved):

PR #1687, opened 2023-06-07
CLDC-2436 Multiple checkbox field bu bug

natdeanlewissoftwire:
Problem:
When errors are added to checkbox questions with multiple answers, they are added to the question not to any individual field. This means in bulk upload we get failures because the bulk upload parser doesn't respond to the question name, only to the individual field names.

Solution:
This PR fixes the above by clearing all multiple checkbox question fields if there is an error on the question in bulk upload, and also adds errors to relevant fields for bulk upload only.

ticket: https://digital.dclg.gov.uk/jira/browse/CLDC-2436


app/services/bulk_upload/lettings/year2023/row_parser.rb
```diff
@@ -1374,15 +1383,6 @@ def housingneeds_other
     return 1 if field_86 == 1
   end
 
-  def homeless
```
natdeanlewissoftwire, line 1377: this is a bug fix, this code was erroneously copied from the 22 parser

app/models/log.rb
```diff
@@ -106,7 +106,12 @@ def blank_invalid_non_setup_fields!
       errors.each do |error|
         next if setup_ids.include?(error.attribute.to_s)
 
-        public_send("#{error.attribute}=", nil)
+        question = form.questions.find { |q| q.id == error.attribute.to_s }
+        if question.type == "checkbox"
+          question.answer_keys_without_dividers { |attribute| public_send("#{attribute}=", nil) }
```
jamesrose: Are you missing an enumerator here? If so, please could you add a spec for this change?

```suggestion
          question.answer_keys_without_dividers.each { |attribute| public_send("#{attribute}=", nil) }
```

jamesrose:
Please could you include information in the PR and/or commits about what this bug is, and why your changes resolve it? Thanks!

app/models/log.rb
jamesrose: Are these changes specific to bulk upload, or should they be applicable for single log submission as well?

natdeanlewissoftwire (commented):
This is still a draft, sorry @jamesrose - I'll let you know when it's ready for review :D

jamesrose:
@natdeanlewissoftwire Whoops! Sorry about that, let me know when you're ready.

app/models/log.rb
```diff
@@ -106,7 +106,12 @@ def blank_invalid_non_setup_fields!
       errors.each do |error|
         next if setup_ids.include?(error.attribute.to_s)
 
-        public_send("#{error.attribute}=", nil)
+        question = form.questions.find { |q| q.id == error.attribute.to_s }
+        if question&.type == "checkbox"
+          question.answer_keys_without_dividers.each { |attribute| public_send("#{attribute}=", nil) }
+        else
```
kosiakkatrina, line 112: How come we're clearing the values here? Would this impact anything beyond bulk upload?

natdeanlewissoftwire: Nope, this method is only called by the validators/log_creators for bulk upload, we were already clearing on old line 109, this just makes sure to actually clear ALL the values for multiple checkbox questions (welcome back!)

kosiakkatrina (approved):

kosiakkatrina (approved):

app/services/bulk_upload/sales/year2022/row_parser.rb
```diff
@@ -467,6 +468,15 @@ def validate_data_protection_answered
     end
   end
 
+  def validate_buyers_organisations
+    organisations_fields = %i[field_44 field_45 field_46 field_47]
+    if organisations_fields.all? { |field| attributes[field.to_s].blank? }
```
jamesrose, line 473: The error message below suggests this should be `any?` — please could you confirm?

app/services/bulk_upload/sales/year2023/row_parser.rb
```diff
@@ -562,6 +563,15 @@ def validate_data_protection_answered
     end
   end
 
+  def validate_buyers_organisations
+    organisations_fields = %i[field_67 field_68 field_69 field_70]
+    if organisations_fields.all? { |field| attributes[field.to_s].blank? }
```
jamesrose, line 568: The error message below suggests this should be any? — please could you confirm?

spec/services/bulk_upload/sales/year2022/row_parser_spec.rb
```diff
@@ -349,6 +349,19 @@ def purch_id_field
       end
     end
 
+    describe "#field_44 - 7" do # buyers organisations
+      context "when all nil" do
+        let(:attributes) { setup_section_params.merge(field_44: nil, field_45: nil, field_46: nil, field_47: nil) }
```
jamesrose, line 354: Can these keys be missing entirely? Does the validation do what you expect in that instance?

app/services/bulk_upload/sales/year2022/row_parser.rb
natdeanlewissoftwire: I think you read the `blank?` as a `present?` perhaps... we add the error message that one must be selected if none of them are

spec/services/bulk_upload/sales/year2022/row_parser_spec.rb
natdeanlewissoftwire: Yep, stll works in that case - though that's down to the row parser not anything I've built so don't think it's a test that should be added here

jamesrose (approved):

PR #1688, opened 2023-06-07
CLDC-2444 production bug logs uneditable

arfacamble:
push back end date of forms to allow them to continue to be edited for now. Adjust the copy about the date until when logs can be edited to keep it consistent with how things have been up until now

https://digital.dclg.gov.uk/jira/browse/CLDC-2444

### DIFF
--- a/app/helpers/tasklist_helper.rb
+++ b/app/helpers/tasklist_helper.rb
@@ -31,9 +31,9 @@ module TasklistHelper
 
   def review_log_text(log)
     if log.collection_period_open?
-      link = log.sales? ? review_sales_log_path(id: log, sales_log: true) : review_lettings_log_path(log)
+      path = log.sales? ? review_sales_log_path(id: log, sales_log: true) : review_lettings_log_path(log)
 
-      "You can #{govuk_link_to 'review and make changes to this log', link} until #{log.form.end_date.to_formatted_s(:govuk_date)}.".html_safe
+      "You can #{govuk_link_to 'review and make changes to this log', path} until #{log.form.display_end_date.to_formatted_s(:govuk_date)}.".html_safe
     else
       start_year = log.startdate ? collection_start_year_for_date(log.startdate) : log.form.start_date.year
 
--- a/app/models/form.rb
+++ b/app/models/form.rb
@@ -1,6 +1,6 @@
 class Form
   attr_reader :form_definition, :sections, :subsections, :pages, :questions,
-              :start_date, :end_date, :type, :name, :setup_definition,
+              :start_date, :end_date, :display_end_date, :type, :name, :setup_definition,
               :setup_sections, :form_sections, :unresolved_log_redirect_page_id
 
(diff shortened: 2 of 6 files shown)


app/models/form.rb
```diff
@@ -9,7 +9,7 @@ def initialize(form_path, start_year = "", sections_in_form = [], type = "lettin
       @end_date = if start_year && start_year.to_i > 2022
                     Time.zone.local(start_year + 1, 6, 9)
                   else
-                    Time.zone.local(start_year + 1, 6, 7)
+                    Time.zone.local(start_year + 1, 8, 7)
```
natdeanlewissoftwire, line 12: I think we agreed this should be march 31st 2024 as we're expecting this to be in september when Rachel confirms so don't want to push it forward again but instead want to bring it back

app/helpers/tasklist_helper.rb
```diff
@@ -31,9 +31,11 @@ def subsection_link(subsection, log, current_user)
 
   def review_log_text(log)
     if log.collection_period_open?
-      link = log.sales? ? review_sales_log_path(id: log, sales_log: true) : review_lettings_log_path(log)
+      path = log.sales? ? review_sales_log_path(id: log, sales_log: true) : review_lettings_log_path(log)
 
-      "You can #{govuk_link_to 'review and make changes to this log', link} until #{log.form.end_date.to_formatted_s(:govuk_date)}.".html_safe
+      date = log.collection_start_year == 2022 ? Time.zone.local(2023, 6, 9) : log.form.end_date
```
jamesrose: So my preference here is that the magic dates belong in the form definitions. If we’re defining a new date to display to users, albeit a fake one, then it should also belong there. How about display_end_date or similar, with the default that `@display_end_date = @end_date`?

app/models/form.rb
arfacamble: I wasn't clear on what date it was but I feel like within two months we need to have a clearer idea of what different end dates are and how they are implemented.
I'm slightly uneasy about putting it so far in the future as it makes it easier to just leave it working and forget all about this.
If we're leaving it like this for any length of time, tests need writing etc to cover it so if we don't have better solutions conceptually we should at least make a PR for writing tests etc and then this can be updated if we wish

natdeanlewissoftwire: Also I'm imagining we should change this for years after 2022/23 too (line 10), so we can just replace the whole if/else with a single date

app/helpers/tasklist_helper.rb
arfacamble: I'm happy with that. I think that whatever is written in this PR will hopefully be superseded pretty soon by something a bit more lasting anyway!

app/models/form.rb
natdeanlewissoftwire: your first point is fine by me 👍, we do need to update this either way

jamesrose (approved):

natdeanlewissoftwire (approved):

PR #1689, opened 2023-06-08
CLDC-2456 Sales collection year test fixes

natdeanlewissoftwire:
This fixes around 300 tests that fail when the 22/23 collection end year is in the past, as time was not being frozen in a number of tests where it should have been.

As far as I am able to predict this is all the fixes that are required (for sales at least). It may be worth investigating whether we will have a similar lettings issue when the crossover period for that ends, and also if any new failures (from code yet to be merged) will arise when the sales crossover period (currently pushed back to august) ends.


spec/features/sales_log_spec.rb
```diff
@@ -38,16 +38,16 @@
           visit("/sales-logs")
           click_button("Create a new sales log")
           click_link("Set up this sales log")
-          fill_in("sales_log[saledate(1i)]", with: "2022")
-          fill_in("sales_log[saledate(2i)]", with: "08")
-          fill_in("sales_log[saledate(3i)]", with: "10")
+          fill_in("sales_log[saledate(1i)]", with: Time.zone.today.year)
```
ghost, line 41: can't the time be frozen in this test?

natdeanlewissoftwire: it can be, I slightly prefer not freezing time where it's possible/easy as that makes it more likely we'll catch genuine bugs in the future, I don't see this behaviour as being limited to 2022.

ghost (approved):

PR #1692, opened 2023-06-13
Cldc 2445 increase offered max

natdeanlewissoftwire:
https://digital.dclg.gov.uk/jira/browse/CLDC-2445

- increase max value for offered
- combine duplicated identital questions for offered depending on if it's social let or not
- add/update tests


app/models/form/lettings/pages/property_number_of_times_relet.rb
```diff
@@ -0,0 +1,12 @@
+class Form::Lettings::Pages::PropertyNumberOfTimesRelet < ::Form::Page
+  def initialize(id, hsh, subsection)
+    super
+    @id = "property_number_of_times_relet"
+    @depends_on = [{ "first_time_property_let_as_social_housing" => 0, "is_renewal?" => false },
```
ghost, line 5: this should probably be covered by a test

ghost (approved):

app/models/form/lettings/pages/property_number_of_times_relet.rb
natdeanlewissoftwire: yes indeed

PR #1694, opened 2023-06-14
CLDC-2321 Add Data Protection Confirmation validation on log change

ghost:
- Validate when assigning a log to a *managing* organisation without a data protection confirmation
- Validate when assigning a log to a *owning* organisation without a data protection confirmation
- Validate when creating a log with *owning* organisation without a data protection confirmation


natdeanlewissoftwire (approved):

app/models/validations/setup_validations.rb
```diff
@@ -42,6 +42,14 @@ def validate_organisation(record)
     end
   end
 
+  def validate_managing_organisation_data_sharing_agremeent_signed(record)
+    return unless FeatureToggle.new_data_protection_confirmation?
+
+    if record.managing_organisation_id_changed? && record.managing_organisation.present? && !record.managing_organisation.data_protection_confirmed?
```
natdeanlewissoftwire, line 48: should we check if the org is present before we check if the id is changed to avoid a NoMethodError? Or is _changed? happy with that

app/models/validations/shared_validations.rb
```diff
@@ -117,6 +117,14 @@ def date_valid?(question, record)
     end
   end
 
+  def validate_owning_organisation_data_sharing_agremeent_signed(record)
+    return unless FeatureToggle.new_data_protection_confirmation?
+
+    if record.owning_organisation_id_changed? && record.owning_organisation.present? && !record.owning_organisation.data_protection_confirmed?
```
natdeanlewissoftwire, line 123: ditto

natdeanlewissoftwire (approved):
one thought, lgtm once that's addressed/ignored

app/models/validations/setup_validations.rb
ghost: managing_organisation_id_changed should just be checking old/new value without calling a method on it. I've written tests to cover this case anyway

PR #1695, opened 2023-06-14
CLDC-2447 Bulk upload 23-24 lettings net income known bug

natdeanlewissoftwire:
ticket: https://digital.dclg.gov.uk/jira/browse/CLDC-2447

Another bug from value assignment being different in 23/24 parser but having just been copied from 22/23. I'll check with Jemma to see if I can check whether there are any other of these that we haven't caught. I came across one other by chance while working on a separate ticket, I don't think there's been a systematic check.

### DIFF
--- a/spec/services/bulk_upload/lettings/year2023/row_parser_spec.rb
+++ b/spec/services/bulk_upload/lettings/year2023/row_parser_spec.rb
@@ -1365,10 +1365,20 @@ RSpec.describe BulkUpload::Lettings::Year2023::RowParser do
     end
 
     describe "#net_income_known" do
-      let(:attributes) { { bulk_upload:, field_120: "1" } }
+      context "when 1" do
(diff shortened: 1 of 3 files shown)


spec/services/bulk_upload/lettings/year2023/row_parser_spec.rb
```diff
@@ -1365,10 +1365,20 @@
...
+        let(:attributes) { { bulk_upload:, field_120: "1" } }
 
-      it "sets value from correct mapping" do
-        expect(parser.log.net_income_known).to eq(0)
+        it "sets value from correct mapping" do
+          expect(parser.log.net_income_known).to eq(0)
+        end
+      end
+
+      context "when 3" do
```
ghost, line 1384: let's test all the possible permutations?

ghost (approved):

PR #1696, opened 2023-06-14
CLDC-2390 CYA address redesign

natdeanlewissoftwire:
all that's left on [this ticket](https://digital.dclg.gov.uk/jira/browse/CLDC-2390) is updating the CYA designs to separate address fields, so this PR is just frontend changes.

image

also, have fixed the error messages as it's related (https://digital.dclg.gov.uk/jira/browse/CLDC-2228):

image

### DIFF
--- a/app/models/form/lettings/questions/address_line1.rb
+++ b/app/models/form/lettings/questions/address_line1.rb
@@ -6,7 +6,7 @@ class Form::Lettings::Questions::AddressLine1 < ::Form::Question
     @header = "Address line 1"
     @type = "text"
     @plain_label = true
-    @check_answer_label = "Q12 - Address"
+    @check_answer_label = "Q12 - Address lines 1 and 2"
     @disable_clearing_if_not_routed_or_dynamic_answer_options = true
   end
 
@@ -14,17 +14,6 @@ class Form::Lettings::Questions::AddressLine1 < ::Form::Question
     [
       log.address_line1,
       log.address_line2,
-      log.postcode_full,
-      log.town_or_city,
-      log.county,
     ].select(&:present?).join("\n")
-  end
-
-  def get_extra_check_answer_value(log)
-    return unless log.is_la_inferred?
-
-    la = LocalAuthority.find_by(code: log.la)&.name
-
-    la.presence
   end
 end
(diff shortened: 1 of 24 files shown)


spec/services/bulk_upload/lettings/year2023/row_parser_spec.rb
```diff
@@ -340,8 +340,8 @@
 
           it "fetches the question's check_answer_label if it exists, otherwise it gets the question's header" do
             parser.valid?
-            expect(parser.errors[:field_19]).to eql(["You must answer q12 - address"])
-            expect(parser.errors[:field_21]).to eql(["You must answer town or city"])
+            expect(parser.errors[:field_19]).to eql(["You must answer q12 - address lines 1 and 2"])
```
natdeanlewissoftwire: we shouldn't be lower casing these but it's not related to this ticket, more of a bulk upload bug, I'm checking if there's a ticket to fix this (if not I'll make one)

app/models/form/lettings/questions/address_line1.rb
```diff
@@ -4,27 +4,19 @@ def initialize(id, hsh, page)
     @id = "address_line1"
     @check_answer_label = "Address"
```
kosiakkatrina: Should this be removed?

app/models/form/question.rb
```diff
@@ -195,7 +197,7 @@ def value_is_refused?(value)
   end
 
   def display_label
-    check_answer_label || header || id.humanize
+    error_label || check_answer_label || header || id.humanize
```
kosiakkatrina, line 200: Is display label only for errors? If so, maybe worth renaming it something more specific?
Also, would it be possible to avoid adding error_label attribute in general by updating check_answers_label with correct display_label?

natdeanlewissoftwire: Yes to the first bit - I'll look at renaming
No to the second bit, this design requires it as we have a different label for the error message (address line 1) to the cya label (address lines 1 and 2)

app/models/form/lettings/questions/address_line1.rb
natdeanlewissoftwire: hah yes good spot :)

kosiakkatrina (approved):

PR #1697, opened 2023-06-15
CLDC-2437 Fix interruption screen routing bug

kosiakkatrina:
There was an `InvalidURIError` on sentry when navigating to `/sales-logs/:x/purchase-price-outright-sale?referrer=interruption_screen` it was getting hit on this line:
`@interruption_page_id = URI.parse(request.headers["HTTP_REFERER"]).path.split("/").last.underscore`

It was possible to reproduce it by directly navigating to the URL, and it was being caused by `HTTP_REFERER` header being empty.

Only setting `@interruption_page_id` if `HTTP_REFERER` is provided seems to fix this error.

### DIFF
--- a/app/controllers/form_controller.rb
+++ b/app/controllers/form_controller.rb
@@ -46,7 +46,7 @@ class FormController < ApplicationController
   end
 
   def show_page
-    if request.params["referrer"] == "interruption_screen"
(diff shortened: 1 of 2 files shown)


natdeanlewissoftwire (approved):
lgtm!

PR #1698, opened 2023-06-15
CLDC-2449 production recursion error on finding url for next incomplete section

arfacamble:
small fix in the method where the error was occurring.
amend lettings log factory to ensure that completed trait really is completed
tests for :calculate_status that also implicitly test the factory traits

We may also iterate over production data with the following (this does not need to be done at the same time as this work is deployed)

```
logs_with_incorrect_status = []
LettingsLog.find_each { |log|  logs_with_incorrect_status << log if log.status == "in_progress" && log.form.subsections.all? { |ss| ss.complete?(log) } }
logs_with_incorrect_status.each { |log| log.update!(status: "completed") }
```

### DIFF
--- a/app/models/form.rb
+++ b/app/models/form.rb
@@ -127,7 +127,7 @@ class Form
     if log.status == "completed"
       return first_question_in_last_subsection(subsection_ids)
     elsif log.calculate_status == "completed"
-      log.update(status: "completed")
+      log.update!(status: "completed")
       return first_question_in_last_subsection(subsection_ids)
     end
 
--- a/app/models/form.rb
+++ b/app/models/form.rb
@@ -126,6 +126,9 @@ class Form
 
     if log.status == "completed"
       return first_question_in_last_subsection(subsection_ids)
+    elsif log.calculate_status == "completed"
+      log.update(status: "completed")
+      return first_question_in_last_subsection(subsection_ids)
     end
 
     next_subsection = next_subsection(subsection, log, subsection_ids)
--- a/app/models/form.rb
+++ b/app/models/form.rb
@@ -124,10 +124,7 @@ class Form
   def next_incomplete_section_redirect_path(subsection, log)
     subsection_ids = subsections.map(&:id)
 
-    if log.status == "completed"
-      return first_question_in_last_subsection(subsection_ids)
-    elsif log.calculate_status == "completed"
-      log.update!(status: "completed")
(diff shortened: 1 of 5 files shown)


spec/models/form_spec.rb
```diff
@@ -190,8 +190,6 @@ def answer_local_authority(lettings_log)
         FormHandler.instance.use_real_forms!
 
         example.run
-
-        FormHandler.instance.use_fake_forms!
```
kosiakkatrina, line 194: I think we might need to set it to avoid flaky tests in other files (?)

app/models/form.rb
```diff
@@ -126,6 +126,9 @@ def next_incomplete_section_redirect_path(subsection, log)
 
     if log.status == "completed"
       return first_question_in_last_subsection(subsection_ids)
+    elsif log.calculate_status == "completed"
+      log.update(status: "completed")
+      return first_question_in_last_subsection(subsection_ids)
```
kosiakkatrina: Are there any risks that we're updating the status here but the answer turns out to be invalid etc?
If we miss `log.update(status: "completed")` would the `update_status!` in log model be enough?

kosiakkatrina (approved):

app/models/form.rb
arfacamble: I'm not sure what answer you mean?
This is a method to create a path, so will be called at the point of rendering a view.
This does have a code smell, because this shouldn't be a place where a model is updated in the db, we could just dodge the error and leave the log in an inconsistent state on the assumption that it will get updated at some point, but I think that might not work for this situation since the log really is complete, but the status is in progress, so the user has no reason to update anything.
the update_status method wouldn't be useful here for that, as it only mutates the object in memory, but I could add a (save: false) to that method and only here set it to true, that would add some flexibility for the future

spec/models/form_spec.rb
arfacamble: Since the FormHandler is reinitialized after every test runs from the spec_helper, I'm sceptical this can be doing anything.
I have run all tests several times on local and it's passed ~once~ twice on remote

app/models/form.rb
kosiakkatrina: Ah, okay, I think I understand. So is this only to address some of the bad data (wrong statuses) that we have in the database?
Would it still be a problem if all the statuses were up to date? Do we know why those statuses were incomplete in the first place?

PR #1699, opened 2023-06-15
CLDC 1507: Explain rent types details added to rent type question page

Airk0n:
This ticket is concerned with adding a details component to the 'Rent Type' question page.

**Before:**
image-2023-02-23-17-44-51-166

**After:**
image

**After expanded:**
image


arfacamble (commented):
Looks good - I don't think it needs to be marked as draft really!

Let's just add a spec to the question. I wonder if it's easy to test the contents of a partial

kosiakkatrina (approved):
Nice!

PR #1700, opened 2023-06-15
Remove feature toggle for schemes and locations

arfacamble:
the title says it all

I would like to make clear that any green bits in the diff are purely red bits moved, this PR is PURE deletion


app/views/locations/index.html.erb
arfacamble, line 1: the diff here is stupid, but essentially there were two similar tables, one which displayed if feature on the other if off. I kept the one in the if not the else

natdeanlewissoftwire (approved):

PR #1701, opened 2023-06-15
CLDC-2420 Update deactivating soon label

kosiakkatrina:
All schemes and locations with an end date have a "deactivating soon" label this appears no matter what the end date is: e.g. if the end date is way in the future, this still appears.

Now we only show "Deactivating soon" label if the deactivation is happening in 6 months or less. If the deactivation is in more than 6 months the label for location or scheme is "Active".
This change only happens on the display layer and the underlying statuses are still the same.

The user can also deactivate a scheme or location that is deactivating_soon but appears as "Active". If this happens we update the open deactivation instead of creating a new one.

All the availability information is staying the same.

### DIFF
--- a/app/controllers/locations_controller.rb
+++ b/app/controllers/locations_controller.rb
@@ -149,7 +149,7 @@ class LocationsController < ApplicationController
   def show; end
 
   def new_deactivation
-    @location_deactivation_period = if @location.deactivates_in_more_than_6_months?
+    @location_deactivation_period = if @location.deactivates_in_a_long_time?
                                       @location.open_deactivation || LocationDeactivationPeriod.new
                                     else
                                       LocationDeactivationPeriod.new
--- a/app/controllers/locations_controller.rb
+++ b/app/controllers/locations_controller.rb
@@ -149,7 +149,11 @@ class LocationsController < ApplicationController
   def show; end
 
   def new_deactivation
-    @location_deactivation_period = LocationDeactivationPeriod.new
+    @location_deactivation_period = if @location.deactivates_in_more_than_6_months?
+                                      @location.open_deactivation || LocationDeactivationPeriod.new
+                                    else
+                                      LocationDeactivationPeriod.new
+                                    end
 
     if params[:location_deactivation_period].blank?
       render "toggle_active", locals: { action: "deactivate" }
@@ -176,7 +180,7 @@ class LocationsController < ApplicationController
   end
 
   def deactivate
-    if @location.location_deactivation_periods.create!(deactivation_date: params[:deactivation_date])
+    if @location.open_deactivation&.update!(deactivation_date: params[:deactivation_date]) || @location.location_deactivation_periods.create!(deactivation_date: params[:deactivation_date])
       logs = reset_location_and_scheme_for_logs!
 
       flash[:notice] = deactivate_success_notice
--- a/app/controllers/schemes_controller.rb
+++ b/app/controllers/schemes_controller.rb
@@ -27,7 +27,7 @@ class SchemesController < ApplicationController
   end
 
   def new_deactivation
-    @scheme_deactivation_period = if @scheme.deactivates_in_more_than_6_months?
+    @scheme_deactivation_period = if @scheme.deactivates_in_a_long_time?
                                     @scheme.open_deactivation || SchemeDeactivationPeriod.new
                                   else
                                     SchemeDeactivationPeriod.new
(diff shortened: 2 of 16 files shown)


app/models/location.rb
```diff
@@ -108,6 +108,10 @@ def reactivating_soon?
     status == :reactivating_soon
   end
 
+  def deactivates_in_more_than_6_months?
+    status == :deactivating_soon && open_deactivation.present? && open_deactivation.deactivation_date > 6.months.from_now
+  end
```
jamesrose: I'd make this more generic:

```suggestion
  DEACTIVATING_SOON_CUTOFF = 6.months
  
  def deactivates_in_a_long_time?
    status == :deactivating_soon && open_deactivation.present? && open_deactivation.deactivation_date > DEACTIVATING_SOON_CUTOFF.from_now
  end
```

But, is this the same as?

```suggestion  
  def deactivates_in_a_long_time?
    status_at(6.months.from_now) == :deactivating_soon
  end
```

kosiakkatrina: Maybe, although I can imagine there being some weird situations going on when there are multiple deactivation periods in the future. I'll see if it works tho, thanks!

app/models/location.rb
```diff
@@ -108,6 +108,10 @@ def reactivating_soon?
     status == :reactivating_soon
   end
 
+  def deactivates_in_a_long_time?
```
Airk0n, line 111: For me, at first glance I may interpret "deactivates in a long" time to mean that it's not deactivating soon, instead it is deactivating way after 6 months.

spec/requests/locations_controller_spec.rb
```diff
@@ -1594,7 +1625,7 @@
         end
       end
 
-      context "when deactivation date is during a deactivated period" do
+      xcontext "when deactivation date is during a deactivated period" do
```
Airk0n: I'm guessing you either forgot to remove or unsnap this context? Unless there's a ticket that you're aware of needing this or something that I'm not aware of :)

app/models/location.rb
kosiakkatrina: I think that's exactly what it's checking! If the status in 6 months is still :deactivating_soon, that means that it's gonna be in :deactivating_soon state for at least 6 months (otherwise the status in 6 months would be :deactivated)

spec/requests/locations_controller_spec.rb
kosiakkatrina: Good spot! Thanks!

jamesrose (approved):

app/controllers/schemes_controller.rb
```diff
@@ -27,7 +27,11 @@ def show
   end
 
   def new_deactivation
-    @scheme_deactivation_period = SchemeDeactivationPeriod.new
+    @scheme_deactivation_period = if @scheme.deactivates_in_a_long_time?
+                                    @scheme.open_deactivation || SchemeDeactivationPeriod.new
+                                  else
+                                    SchemeDeactivationPeriod.new
+                                  end
```
ghost, line 34:
```
@scheme_deactivation_period = if @scheme.deactivates_in_a_long_time?
  @scheme.open_deactivation 
end || SchemeDeactivationPeriod.new
```

app/controllers/locations_controller.rb
```diff
@@ -176,7 +180,7 @@ def deactivate_confirm
   end
 
   def deactivate
-    if @location.location_deactivation_periods.create!(deactivation_date: params[:deactivation_date])
+    if @location.open_deactivation&.update!(deactivation_date: params[:deactivation_date]) || @location.location_deactivation_periods.create!(deactivation_date: params[:deactivation_date])
```
ghost, line 183: putting this logic into a method would make this line easier to read and understand

app/controllers/schemes_controller.rb
```diff
@@ -54,7 +58,7 @@ def deactivate_confirm
   end
 
   def deactivate
-    if @scheme.scheme_deactivation_periods.create!(deactivation_date: params[:deactivation_date])
+    if @scheme.open_deactivation&.update!(deactivation_date: params[:deactivation_date]) || @scheme.scheme_deactivation_periods.create!(deactivation_date: params[:deactivation_date])
```
ghost, line 61: putting this logic into a method would make this line easier to read and understand

app/models/location.rb
ghost: can we rename the method to include the 6 months rather than saying "long time"? long/short is all relative

PR #1702, opened 2023-06-15
CLDC-1741 Scheme attributes bug fix

natdeanlewissoftwire:
a small non-critical frontend tweak @arfacamble found from the original CLDC-1741 PR, this PR ensures we only display a scheme's secondary client group if they have answered yes to having one.

DO NOT MERGE until we know what to do with the schemes on prod that have second client groups but no answer to `has_other_client_group`/the importer that created this situation


spec/features/schemes_spec.rb
```diff
@@ -149,20 +149,45 @@
         context "when I click to see individual scheme" do
           let(:scheme) { schemes.first }
 
-          before do
-            click_link(scheme.service_name)
+          context "when there is a secondary client group" do
+            before do
```
Airk0n, line 153: I can see that in the context where you expect it not to have a secondary scheme that you explicitly set it not to, for me I would expect this context you've created to be responsible for making sure that it is the case. 
Looking at the file as a whole though it's one of them things that someone will fix after they break your test by changing something in a wider scope so I guess it's debatable either way

Airk0n (commented):
Looks good to me 👍

spec/features/schemes_spec.rb
natdeanlewissoftwire: I take your point but think this is fine - the only reason we explicitly test the "secondary client group" text doesn't appear in the other test if that there is no client group name for us to test non-existence of in that case, whereas here the secondary_client_group name appearing is enough to see that the correct details are being displayed

app/helpers/schemes_helper.rb
```diff
@@ -10,7 +10,6 @@ def display_scheme_attributes(scheme, user)
       { name: "Support services provided by", value: scheme.arrangement_type },
       { name: "Primary client group", value: scheme.primary_client_group },
       { name: "Has another client group", value: scheme.has_other_client_group },
-      { name: "Secondary client group", value: scheme.secondary_client_group },
```
kosiakkatrina, line 14: Would this with `.compact` at the end work? 
`scheme.has_other_client_group == "Yes" ? { name: "Level of support given", value: scheme.support_type }: nil,`

arfacamble: 🔥 

if that works you have also removed all the ifs that come after the initial array definition so you can remove saving in variable and just return directly

spec/factories/scheme.rb
```diff
@@ -19,6 +20,7 @@
       scheme_type { 7 }
       intended_stay { "M" }
       primary_client_group { "G" }
+      has_other_client_group { 1 }
```
arfacamble, line 23: I would have expected that the attributes defined outside of traits are defaults that would still be applied to all factory instances so this line would be redundant.

If you remove it and everything still works, you can probably remove line 18 as well `registered_under_care_act`

spec/features/schemes_spec.rb
```diff
@@ -149,20 +149,45 @@
...
-            expect(page).to have_content(schemes.first.registered_under_care_act)
-            expect(page).to have_content(schemes.first.primary_client_group)
-            expect(page).to have_content(schemes.first.secondary_client_group)
-            expect(page).to have_content(schemes.first.support_type)
-            expect(page).to have_content(schemes.first.intended_stay)
+          context "when there is no secondary client group" do
+            before do
+              scheme.has_other_client_group = "No"
+              scheme.secondary_client_group = nil
+              scheme.save!
```
arfacamble: Just a bit of code golf, but would this work? 
```suggestion
              scheme.update!(has_other_client_group: "No", secondary_client_group: nil)
```

spec/features/schemes_spec.rb
```diff
@@ -685,23 +710,52 @@
...
+              expect(page).to have_content(schemes.first.support_type)
+              expect(page).to have_content(schemes.first.intended_stay)
+            end
+          end
+
+          context "when there is no secondary client group" do
+            before do
+              scheme.has_other_client_group = "No"
+              scheme.secondary_client_group = nil
+              scheme.save!
```
arfacamble: code golf potential as above

app/helpers/schemes_helper.rb
natdeanlewissoftwire: tasty, will give it a go

spec/factories/scheme.rb
natdeanlewissoftwire: Yep that's correct - but I think the factory should model the expected state of an object i.e. if there's a secondary client group, `has_other_client_group` == 1

natdeanlewissoftwire: I could add a trait for a separate scheme that has a secondary client group and test separately but I don't personally see a benefit in that when I'm doing that test anyway more explicitly in the specs

spec/features/schemes_spec.rb
natdeanlewissoftwire: 🚀 as above

app/helpers/schemes_helper.rb
natdeanlewissoftwire: have done both of these 👍

app/helpers/schemes_helper.rb
```diff
@@ -1,34 +1,21 @@
...
-    if user.data_coordinator?
-      base_attributes.delete_if { |item| item[:name] == "Housing stock owned by" }
-    end
-
-    if scheme.has_other_client_group == "Yes"
-      base_attributes.append
-    end
-
-    base_attributes
+      FeatureToggle.scheme_toggle_enabled? ? { name: "Status", value: status_tag(scheme.status) } : nil,
```
arfacamble: forgot to mention that you should rebase early and resolve conflicts 😬

spec/factories/scheme.rb
arfacamble: We may be talking cross purposes but it's extremely minor so let's move on with our lives and enjoy the cricket

spec/requests/delete_logs_controller_spec.rb
natdeanlewissoftwire, line 1: unrelated linting

arfacamble (approved):

PR #1703, opened 2023-06-15
CLDC-2374 Update bulk upload guidance

kosiakkatrina:
Content update for bulk upload guidance pages.

### DIFF
--- a/app/models/forms/bulk_upload_sales/guidance.rb
+++ b/app/models/forms/bulk_upload_sales/guidance.rb
@@ -15,8 +15,8 @@ module Forms
         bulk_upload_sales_log_path(id: "prepare-your-file", form: { year: })
       end
 
-      def old_template_path
-        Forms::BulkUploadSales::PrepareYourFile.new.old_template_path
+      def legacy_template_path
+        Forms::BulkUploadSales::PrepareYourFile.new.legacy_template_path
       end
 
       def template_path
--- a/app/models/forms/bulk_upload_sales/guidance.rb
+++ b/app/models/forms/bulk_upload_sales/guidance.rb
@@ -16,7 +16,7 @@ module Forms
       end
 
       def old_template_path
-        Forms::BulkUploadLettings::PrepareYourFile.new.old_template_path
+        Forms::BulkUploadSales::PrepareYourFile.new.old_template_path
       end
 
       def template_path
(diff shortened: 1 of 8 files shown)


app/models/forms/bulk_upload_sales/guidance.rb
```diff
@@ -16,7 +16,7 @@ def back_path
       end
 
       def old_template_path
```
natdeanlewissoftwire: might be worth calling this "legacy_template_path" if that's the term we use in the service

natdeanlewissoftwire: ( I think it may have been me that called it old originally )

app/views/bulk_upload_shared/guidance.html.erb
```diff
@@ -54,7 +54,14 @@
 
     <div class="govuk-!-padding-bottom-4">
       <h2 class="govuk-heading-s">Getting help</h2>
-      <p class="govuk-body">There is no step-by-step bulk upload guide like there is with single log upload. However, you can download <%= govuk_link_to "our template", @form.template_path %>, which you can copy-paste data into from your systems column-by-column. You can also view a post-upload report showing any data errors, and our <%= govuk_link_to "data specification", @form.specification_path, target: "_blank" %> can help fix these.</p>
+      <p class="govuk-body">There is no step-by-step bulk upload guide like there is with single log upload. However, you can download <%= @form.year == 2022 ? govuk_link_to("our template", @form.template_path) : "one of our templates" %>, which you can copy-paste data into from your systems column-by-column. You can also view a post-upload report showing any data errors, and our <%= govuk_link_to "data specification", @form.specification_path, target: "_blank" %> can help fix these.</p>
```
natdeanlewissoftwire, line 57: surely they'd copy-paste row-by-row (1 log at a time) rather than column-by-column (1 attribute at a time)?

natdeanlewissoftwire (approved):

app/views/bulk_upload_shared/guidance.html.erb
kosiakkatrina: I'll ask on the ticket, ty!

app/models/forms/bulk_upload_sales/guidance.rb
kosiakkatrina: Sure! Will fix, thanks!

PR #1706, opened 2023-06-16
CLDC-1783 Add telephone number question to the user form

kosiakkatrina:
We have been storing user telephone numbers, now we also display them on the user pages:
image

Telephone numbers can be updated (same permissions as for the email updates):
image

There is a simple validation checking that the number is digits and over 11 digits long, only applicable to newly created logs:
image


app/controllers/users_controller.rb
```diff
@@ -114,6 +114,14 @@ def validate_attributes
     if user_params[:role].present? && !current_user.assignable_roles.key?(user_params[:role].to_sym)
       @resource.errors.add :role, I18n.t("validations.role.invalid")
     end
+
+    if user_params[:phone].present? && !valid_phone_number?(user_params[:phone])
```
ghost, line 118: i'd probably prefer seeing this validation on the model rather than on the controller action, do we want to validate the phone number only here?

app/helpers/user_helper.rb
```diff
@@ -34,4 +34,8 @@ def can_edit_key_contact?(user, current_user)
   def can_edit_org?(current_user)
     current_user.data_coordinator? || current_user.support?
   end
+
+  def can_edit_telephone_numbers?(user, current_user)
+    (current_user == user || current_user.data_coordinator? || current_user.support?) && user.active?
```
ghost: could make sense as a policy?

spec/features/user_spec.rb
```diff
@@ -322,16 +322,42 @@
         expect(page).to have_title("Error")
       end
 
+      it "validates telephone number is numeric" do
```
ghost, line 325: validation on the model would save us from having to write feature spec for this

app/controllers/users_controller.rb
kosiakkatrina: I think it's only here because some of the migrated logs have numbers that otherwise would not pass this validation (for example numbers with extensions etc.)
From ACs:
_If possible - for any newly created numbers (so not those that are imported) include a validation where  all characters are numbers and the number is at least 11 digits long_ 

Let me know if you can think of another way to accomplish this tho

spec/features/user_spec.rb
kosiakkatrina: same as above

app/controllers/users_controller.rb
ghost: ok fair enough, thought there might be some other reason!

app/helpers/user_helper.rb
kosiakkatrina: Is it better now?

ghost (approved):

app/controllers/users_controller.rb
```diff
@@ -114,6 +114,14 @@ def validate_attributes
...
+      @resource.errors.add :phone
+    end
+  end
+
+  def valid_phone_number?(number)
+    number.to_i.to_s == number && number.length >= 11
```
jamesrose, line 124: This should be in the model

kosiakkatrina: Some context here: https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/1706#discussion_r1233662526

We spoke about it when we were refining this ticket as well

PR #1707, opened 2023-06-16
CLDC 2075: Hint details added to 3 question pages for year 2023

Airk0n:
This ticket is concerned with adding an additional details component to three questions each with unique content.
These details components are only to appear when the year of the form is 2023/2024.

**Before :** - and still for years before that are not 2023/24
image

**After : **  
image

### DIFF
--- a/spec/models/form/sales/pages/discounted_ownership_type_spec.rb
+++ b/spec/models/form/sales/pages/discounted_ownership_type_spec.rb
@@ -20,7 +20,7 @@ RSpec.describe Form::Sales::Pages::DiscountedOwnershipType, type: :model do
   end
 
   it "has the correct header" do
-    expect(page.header).to be_nil
+    expect(page.header).to eq("Discounted ownership type")
   end
 
   it "has the correct description" do
--- a/spec/models/form/sales/pages/discounted_ownership_type_spec.rb
+++ b/spec/models/form/sales/pages/discounted_ownership_type_spec.rb
@@ -5,7 +5,26 @@ RSpec.describe Form::Sales::Pages::DiscountedOwnershipType, type: :model do
 
   let(:page_id) { nil }
   let(:page_definition) { nil }
-  let(:subsection) { instance_double(Form::Subsection) }
+  let(:subsection) { instance_double(Form::Subsection, form: instance_double(Form, start_date:)) }
+  let(:start_date) { Time.zone.today }
+
+  describe "headers" do
+    context "when form year is for 2023" do
+      let(:start_date) { Time.utc(2023, 2, 8) }
+
+      it "has the correct header" do
+        expect(page.header).to eq("Discounted ownership type")
+      end
+    end
+
+    context "when form year is for before 2023" do
+      let(:start_date) { Time.utc(2022, 2, 8) }
+
+      it "does not have a page header" do
+        expect(page.header).to eq(nil)
+      end
+    end
+  end
 
   it "has correct subsection" do
     expect(page.subsection).to eq(subsection)
@@ -17,10 +36,6 @@ RSpec.describe Form::Sales::Pages::DiscountedOwnershipType, type: :model do
 
   it "has the correct id" do
     expect(page.id).to eq("discounted_ownership_type")
-  end
-
-  it "has the correct header" do
-    expect(page.header).to eq("Discounted ownership type")
   end
 
   it "has the correct description" do
--- a/spec/models/form/sales/pages/discounted_ownership_type_spec.rb
+++ b/spec/models/form/sales/pages/discounted_ownership_type_spec.rb
@@ -13,7 +13,7 @@ RSpec.describe Form::Sales::Pages::DiscountedOwnershipType, type: :model do
       let(:start_date) { Time.utc(2023, 2, 8) }
 
       it "has the correct header" do
-        expect(page.header).to eq("Discounted ownership type")
+        expect(page.header).to eq("Type of discounted ownership sale")
       end
     end
 
--- a/spec/models/form/sales/pages/discounted_ownership_type_spec.rb
+++ b/spec/models/form/sales/pages/discounted_ownership_type_spec.rb
@@ -9,16 +9,16 @@ RSpec.describe Form::Sales::Pages::DiscountedOwnershipType, type: :model do
   let(:start_date) { Time.zone.today }
 
   describe "headers" do
-    context "when form year is for 2023" do
-      let(:start_date) { Time.utc(2023, 2, 8) }
+    context "when form year is for 2023/24" do
+      let(:start_date) { Time.zone.local(2023, 4, 8) }
 
       it "has the correct header" do
         expect(page.header).to eq("Type of discounted ownership sale")
       end
     end
 
-    context "when form year is for before 2023" do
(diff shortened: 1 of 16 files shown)


kosiakkatrina (approved):
Looks good, I think you might need to update the failing test probably with something like this: `let(:form) { instance_double(Form) }` like we have in
`spec/models/form/sales/subsections/household_characteristics_spec.rb`

Airk0n:
> 

Ahh yes, ty :)

spec/models/form/sales/pages/shared_ownership_type_spec.rb
```diff
@@ -1,13 +1,30 @@
-require "rails_helper"
+require "rails_hel\per"
```
RachaelBooth: What's happened here?

spec/models/form/sales/pages/shared_ownership_type_spec.rb
```diff
@@ -1,13 +1,30 @@
...
 
   let(:page_id) { nil }
   let(:page_definition) { nil }
+  let(:subsection) { instance_double(Form::Subsection, form: instance_double(Form, start_date:)) }
   let(:start_date) { Time.utc(2022, 4, 1) }
-  let(:form) { instance_double(Form, start_date:) }
-  let(:subsection) { instance_double(Form::Subsection, form:) }
+
+  describe "headers" do
+    context "when 2023" do
```
RachaelBooth: Can these two contexts have better descriptions please? e.g. "when the form year is 2023"

(My rule of thumb here is essentially read it as a sentence - I think you have done that elsewhere and maybe just missed these ones)

spec/models/form/sales/pages/discounted_ownership_type_spec.rb
```diff
@@ -5,7 +5,26 @@
 
   let(:page_id) { nil }
   let(:page_definition) { nil }
-  let(:subsection) { instance_double(Form::Subsection) }
+  let(:subsection) { instance_double(Form::Subsection, form: instance_double(Form, start_date:)) }
+  let(:start_date) { Time.zone.today }
+
+  describe "headers" do
+    context "when form year is for 2023" do
+      let(:start_date) { Time.utc(2023, 2, 8) }
```
RachaelBooth: I find it slightly odd that these aren't the 1st of April, since that's when form start dates are. But it doesn't actually matter at the moment, and in practice I can see that other dates do float around in our testing so :shrug:

spec/models/form/sales/questions/outright_ownership_type_spec.rb
```diff
@@ -43,4 +41,31 @@
...
+      end
+    end
+
+    context "when the form is for before year 2023" do
+      let(:start_date) { Time.utc(2022, 2, 8) }
+
+      it "does not display a guidance partial" do
+        expect(question.guidance_partial).to eq(nil)
+      end
+    end
```
arfacamble: my personal preference for these things, for complete clarity and reduction of any cognitive load parsing dates for the next reader, is to to name these contexts something like 23/24 rather than just the start year.

lines 50/51 I think show this as when there are other places around the codebase where we use one year only, this tends to refer to the start year, whereas (2023, 2, 8) is in the 22/23 year.

this therefore warrants investigation as I would expect this test to fail if the guidance is only meant to be displayed from 23/24 onwards.

spec/models/form/sales/questions/shared_ownership_type_spec.rb
```diff
@@ -50,6 +50,10 @@
         "30" => { "value" => "Shared Ownership - 2021 model lease" },
       })
     end
+
+    it "does not show a guidance_partial" do
+      expect(question.guidance_partial).to eq(nil)
```
arfacamble, line 57: nothing wrong with this, but just to give you another option which is marginally more semantic, this can be:
`expect(thing).to be_nil`

spec/models/form/sales/subsections/setup_spec.rb
```diff
@@ -5,7 +5,8 @@
 
   let(:subsection_id) { nil }
   let(:subsection_definition) { nil }
-  let(:section) { instance_double(Form::Sales::Sections::Setup) }
+  let(:section) { instance_double(Form::Sales::Sections::Setup, form: instance_double(Form, start_date:)) }
+  let(:start_date) { Time.utc(2022, 4, 1) }
```
arfacamble, line 9: why do we need to do this? We should try to have as few set dates in our testing as possible as it tends to make them brittle in the future.

spec/models/form/sales/pages/shared_ownership_type_spec.rb
arfacamble: interesting that it seems to have passed all tests anyway! maybe there is no need for rails_helper in certain test files???

arfacamble (changes requested):
looks good overall, one thing that needs looking at and a couple of other minor things to think about

spec/models/form/sales/pages/shared_ownership_type_spec.rb
Airk0n: Yeah interesting, I'm actually more concerned with how I ended up committing that in the first place. But yes I guess this particular spec doesn't the rails_helper. So I guess I'll just remove it from this spec now that I know that

spec/models/form/sales/pages/shared_ownership_type_spec.rb
Airk0n: Yes I agree, I had actually changed them and missed these ones discarded the commit by mistake.

spec/models/form/sales/pages/discounted_ownership_type_spec.rb
Airk0n: It's weird but with things like dates and times I like to test within the boundaries instead of on the first/last day. I guess it's a habit picked up from working with dates that change even though we pretty much know that doesn't happen here.

spec/models/form/sales/subsections/setup_spec.rb
Airk0n: I agree, however by introducing the condition of dates within specific sales questions, we now require a form for the set up which requires a date.

spec/models/form/sales/pages/discounted_ownership_type_spec.rb
kosiakkatrina: This would actually mean that the form start year would be 2022 (cause 2022 runs 22/04/01 till 23/04/01) so either the test name or the date should change (Ah I think Arthur mentioned it in his comment below so feel free to ignore)

spec/models/form/sales/questions/outright_ownership_type_spec.rb
```diff
@@ -43,4 +41,31 @@
...
     })
   end
+
+  it "has the correct header" do
+    expect(question.header).to eq("What is the type of outright sale?")
+  end
+
+  describe "partial guidance" do
+    context "when the form is for year 2023" do
+      let(:start_date) { Time.utc(2023, 2, 8) }
```
arfacamble:
```suggestion
      let(:start_date) { Time.zone.local(2023, 2, 8) }
```

arfacamble (approved):

PR #1709, opened 2023-06-19
CLDC-2452: Add method for importing the 'offered' field into already imported lettings logs

RachaelBooth:


RachaelBooth:
N.B. This field does not appear to be relevant to sales

RachaelBooth:
To do the import, will need to have the relevant data and run an import script calling this - I'll want to test on staging.

app/services/imports/lettings_logs_field_import_service.rb
```diff
@@ -8,13 +8,32 @@ def update_field(field, folder)
         import_from(folder, :update_major_repairs)
       when "lettings_allocation"
         import_from(folder, :update_lettings_allocation)
+      when "offered"
```
jamesrose, line 11: Please could you add a spec for this change? See: `spec/services/imports/lettings_logs_field_import_service_spec.rb`

RachaelBooth: Have done so - although I have missed out the "value we're trying to re-import is invalid" case because the set up effort was becoming a bit more than it seemed worth.

jamesrose (approved):

PR #1710, opened 2023-06-19
CLDC-2324 assign data protection officer

ghost:
- Update banner to show a message when org does not have a DPO 
- Send email to user to tell them to sign DSA after being made a DPO

### DIFF
--- a/app/components/data_protection_confirmation_banner_component.rb
+++ b/app/components/data_protection_confirmation_banner_component.rb
@@ -3,6 +3,8 @@ class DataProtectionConfirmationBannerComponent < ViewComponent::Base
 
   attr_reader :user, :organisation
 
+  HELPDESK_URL = "https://digital.dclg.gov.uk/jira/servicedesk/customer/portal/4/group/21".freeze
+
   def initialize(user:, organisation: nil)
     @user = user
     @organisation = organisation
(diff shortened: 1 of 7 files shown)


app/models/user.rb
```diff
@@ -179,4 +182,15 @@ def valid_for_authentication?
   def password_required?
     !persisted? || !password.nil? || !password_confirmation.nil?
   end
+
+private
+
+  def send_data_protection_confirmation_reminder
+    return unless FeatureToggle.new_data_protection_confirmation?
+    return unless persisted?
+    return unless is_dpo_changed?
```
kosiakkatrina: Do we need this twice? (line 24 as well)

app/components/data_protection_confirmation_banner_component.rb
```diff
@@ -13,26 +15,62 @@ def initialize(user:, organisation: nil)
...
+    else
+      "Your data protection officer must accept the Data Sharing Agreement on CORE before you can create any logs."
+    end
+  end
 
-    !DataProtectionConfirmation.exists?(
-      organisation: org_or_user_org,
-      confirmed: true,
-    )
+  def second_row
```
kosiakkatrina: Can we rename this to something else? Like banner_text/banner_detail/banner_content etc?

kosiakkatrina (approved):
Looks good!

PR #1711, opened 2023-06-19
CLDC-2451 Add submission and edit end dates

kosiakkatrina:


spec/helpers/tasklist_helper_spec.rb
```diff
@@ -165,7 +165,7 @@
 
           it "returns relevant text" do
             expect(review_log_text(lettings_log)).to eq(
-              "You can #{govuk_link_to 'review and make changes to this log', review_lettings_log_path(lettings_log)} until 9 June 2024.".html_safe,
+              "You can #{govuk_link_to 'review and make changes to this log', review_lettings_log_path(lettings_log)} until 7 June 2024.".html_safe,
```
Airk0n, line 168: Though I'm guessing there isn't something in place for this, I don't like the idea of dates being plain text like this, perhaps some rudimentary CMS or JSON for storing significant dates as well as a description of what each date represents would be good for this sort of thing.

Airk0n (commented):
Nothing stands out much to me, I could do with a short description to get some context without going through the ticket.

app/models/form.rb
```diff
@@ -38,9 +43,10 @@ def initialize(form_path, start_year = "", sections_in_form = [], type = "lettin
       @questions = pages.flat_map(&:questions)
       @start_date = Time.iso8601(form_definition["start_date"])
       @end_date = Time.iso8601(form_definition["end_date"])
+      @submission_deadline = Time.zone.local(2023, 6, 9)
```
arfacamble, line 46: Surely this needs to be varied by year? OR is this hard-coded because from now on it could only ever apply to a 22/23 log?

app/models/form.rb
```diff
@@ -38,9 +43,10 @@ def initialize(form_path, start_year = "", sections_in_form = [], type = "lettin
...
       @unresolved_log_redirect_page_id = form_definition["unresolved_log_redirect_page_id"]
     end
-    @display_end_date = start_year == 2022 ? Time.zone.local(2023, 6, 9) : @end_date
+    @edit_end_date = @end_date
```
arfacamble, line 49: what is @end_date? I thought we were removing that? Or is it meant to represetn the middle end date where we can no longer create logs? Since we have lots of different end dates I think they should all have a descriptive name.

spec/helpers/tasklist_helper_spec.rb
kosiakkatrina: I'm not sure what you mean, we're storing them as date objects on the form model. The text representation is just for the UI (?)

app/models/form.rb
kosiakkatrina: I've added an ARD as part of this, I think that might answer your question, but let me know if it doesn't!

app/models/form.rb
kosiakkatrina: I think it drops into this part of conditional statement only for forms with a start year before 2022, submission_deadline above depends on the year

app/models/form.rb
kosiakkatrina: Ah, in terms of the descriptive name. I agree! I didn't update it here because end_date is used in a lot of validations etc. There is a separate ticket (CLDC-2459) for the new logs end date, so I thought it might be better to update the name there, to keep the PRs smaller

arfacamble: I would disagree with this except that you've picked up all three of these tickets

arfacamble (approved):

PR #1712, opened 2023-06-19
CLDC-2468 Put bulk upload controller checks behind feature flag

ghost:
These new controller checks should have been placeed behind the feature toggle


kosiakkatrina (approved):

app/controllers/bulk_upload_lettings_logs_controller.rb
```diff
@@ -25,9 +25,10 @@ def update
 private
 
   def validate_data_protection_agrement_signed!
```
arfacamble, line 27: still a type in agreement FYI

PR #1713, opened 2023-06-19
CLDC-2379 Stop soft validations being applied to fields that are not routed to in bulk upload

arfacamble:
As in the screenshots in the [JIRA ticket](https://digital.dclg.gov.uk/jira/browse/CLDC-2379), we have currently been adding soft validations to every field that they could possibly affect in bulk upload.

This PR ensures that soft validations are only added to bulk upload fields that relate to questions in the form that would be routed to

### DIFF
--- a/app/services/bulk_upload/lettings/year2022/row_parser.rb
+++ b/app/services/bulk_upload/lettings/year2022/row_parser.rb
@@ -793,8 +793,10 @@ private
       next if question.completed?(log)
 
       question.page.interruption_screen_question_ids.each do |interruption_screen_question_id|
+        next if log.form.questions.none? { |q| q.id == interruption_screen_question_id && q.page.routed_to?(log, nil) }
+
         field_mapping_for_errors[interruption_screen_question_id.to_sym]&.each do |field|
-          unless errors.any? { |e| e.options[:category] == :soft_validation && field_mapping_for_errors[interruption_screen_question_id.to_sym].include?(e.attribute) }
+          if errors.none? { |e| e.options[:category] == :soft_validation && field_mapping_for_errors[interruption_screen_question_id.to_sym].include?(e.attribute) }
             error_message = [display_title_text(question.page.title_text, log), display_informative_text(question.page.informative_text, log)].reject(&:empty?).join(". ")
             errors.add(field, message: error_message, category: :soft_validation)
           end
--- a/app/services/bulk_upload/sales/year2022/row_parser.rb
+++ b/app/services/bulk_upload/sales/year2022/row_parser.rb
@@ -1110,8 +1110,10 @@ private
       next if question.completed?(log)
 
       question.page.interruption_screen_question_ids.each do |interruption_screen_question_id|
+        next if log.form.questions.none? { |q| q.id == interruption_screen_question_id && q.page.routed_to?(log, nil) }
+
         field_mapping_for_errors[interruption_screen_question_id.to_sym]&.each do |field|
-          unless errors.any? { |e| e.options[:category] == :soft_validation && field_mapping_for_errors[interruption_screen_question_id.to_sym].include?(e.attribute) }
+          if errors.none? { |e| e.options[:category] == :soft_validation && field_mapping_for_errors[interruption_screen_question_id.to_sym].include?(e.attribute) }
             error_message = [display_title_text(question.page.title_text, log), display_informative_text(question.page.informative_text, log)].reject(&:empty?).join(". ")
             errors.add(field, message: error_message, category: :soft_validation)
           end
--- a/spec/services/bulk_upload/lettings/year2022/row_parser_spec.rb
+++ b/spec/services/bulk_upload/lettings/year2022/row_parser_spec.rb
@@ -990,11 +990,23 @@ RSpec.describe BulkUpload::Lettings::Year2022::RowParser do
         end
       end
 
+      context "when a soft validation is triggered that relates both to fields that are and are not routed to" do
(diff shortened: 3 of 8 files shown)


app/services/bulk_upload/lettings/year2022/row_parser.rb
```diff
@@ -793,8 +793,10 @@ def validate_incomplete_soft_validations
       next if question.completed?(log)
 
       question.page.interruption_screen_question_ids.each do |interruption_screen_question_id|
+        next if log.form.questions.none? { |q| q.id == interruption_screen_question_id && q.page.routed_to?(log, nil) }
+
         field_mapping_for_errors[interruption_screen_question_id.to_sym]&.each do |field|
-          unless errors.any? { |e| e.options[:category] == :soft_validation && field_mapping_for_errors[interruption_screen_question_id.to_sym].include?(e.attribute) }
+          if errors.none? { |e| e.options[:category] == :soft_validation && field_mapping_for_errors[interruption_screen_question_id.to_sym].include?(e.attribute) }
```
Airk0n, line 799: I didn't think I would like 'unless' when I came to ruby, however 'unless there's any errors' makes more sense to me than 'if there are none errors' which is how I parse this, totally subjective, either way works :)

app/services/bulk_upload/sales/year2022/row_parser.rb
```diff
@@ -1110,8 +1110,10 @@ def validate_incomplete_soft_validations
       next if question.completed?(log)
 
       question.page.interruption_screen_question_ids.each do |interruption_screen_question_id|
+        next if log.form.questions.none? { |q| q.id == interruption_screen_question_id && q.page.routed_to?(log, nil) }
```
Airk0n, line 1113: I don't know much about this row parsing, however this sort of function appears to come up a lot, could this be put in a helper method and referred to by multiple questions? even from multiple years perhaps.

Airk0n (commented):
I'm not particularly familiar with this area of the code base but nothing stands out to me as unreadable or counter-intuitive

app/services/bulk_upload/lettings/year2022/row_parser.rb
jamesrose: Here's a good litmus for when to use unless or not: https://github.com/airbnb/ruby#conditional-keywords

arfacamble: The way I read that style guide @jamesrose is essentially to only use unless for the very simplest conditions.
The way I'm thinking of it is if the condition is remotely complicated then I am in the end going to parse that chunk of code as is and then have to take the negative of that after.
For example in this case, I think the only way I would like to leave it as an unless is if I put the condition into its own variable so I can compartmentalise it more easily:
```
any_errors_on_this_field_for_this_attribute = errors.any? { |e| e.options[:category] == :soft_validation && field_mapping_for_errors[interruption_screen_question_id.to_sym].include?(e.attribute) }
unless any_errors_on_this_field_for_thisAttribute
  stuff
end
```

app/services/bulk_upload/sales/year2022/row_parser.rb
arfacamble: There is definitely a wider conversation that can be had about whether all these row parser objects should inherit from a general row parser class or include a helper module.
When I suggested that Kat pointed me towards this (which is an interesting short read):
https://sandimetz.com/blog/2016/1/20/the-wrong-abstraction
which is what Phil cited when he made the design choice to have all the row parsers just be their own self-contained complete things.
I'm not sure I agree with that in this case since I think this particular method and a couple of others really are doing the same thing in each row parser and I wouldn't imagine they would start to diverge. However I didn't feel like taking on the tech debt at this stage and Phil is a much more experienced developer than me so he may have had some instinct as to why they may diverge in the future.

spec/services/bulk_upload/lettings/year2022/row_parser_spec.rb
```diff
@@ -990,11 +990,23 @@
...
+          expect(parser.errors.where(:field_23, category: :soft_validation)).not_to be_present
+        end
+      end
+
       context "when soft validation is triggered and the mappings for errors are not defined" do
         let(:attributes) { setup_section_params.merge({ field_12: 22, field_35: 5 }) }
-        # rubocop:disable RSpec/AnyInstance
-        let(:mock_interruption_ids) { allow_any_instance_of(Form::Page).to receive(:interruption_screen_question_ids).and_return(%w[fake_question_id]) }
-        # rubocop:enable RSpec/AnyInstance
+        let(:mock_interruption_ids) { allow_any_instance_of(Form::Page).to receive(:interruption_screen_question_ids).and_return(%w[fake_question_id]) } # rubocop:disable RSpec/AnyInstance
```
natdeanlewissoftwire, line 1009: I'd prefer this commented rubocop disable on a separate line, if it's actually necessary?

natdeanlewissoftwire (commented):
lgtm other than the one change suggested

natdeanlewissoftwire (approved):

spec/services/bulk_upload/lettings/year2022/row_parser_spec.rb
arfacamble: why is that? it seems to me to be the intention through design that if you want to disable the rubocop warning for one line, you disable it inline for that line only, but if you want to disable it for a chunk of a file you wrap it around the whole section you want to disable it for

PR #1714, opened 2023-06-21
CLDC-2475 Update telephone validation

kosiakkatrina:
Change number validation to allow 0 at the beginning of the string

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -121,7 +121,9 @@ private
       @resource.errors.add :role, I18n.t("validations.role.invalid")
     end
 
-    if user_params[:phone].present? && !valid_phone_number?(user_params[:phone])
+    if user_params[:phone].blank?
+      @resource.errors.add :phone, :blank
+    elsif !valid_phone_number?(user_params[:phone])
       @resource.errors.add :phone
     end
   end
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -129,7 +129,7 @@ private
   end
 
   def valid_phone_number?(number)
-    /^\d{11,}$/.match?(number)
+    /^[+\d]{11,}$/.match?(number)
   end
 
   def format_error_messages
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -127,7 +127,13 @@ private
   end
 
   def valid_phone_number?(number)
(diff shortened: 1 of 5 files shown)


app/controllers/users_controller.rb
```diff
@@ -127,7 +127,13 @@ def validate_attributes
   end
 
   def valid_phone_number?(number)
-    number.to_i.to_s == number && number.length >= 11
+    numeric?(number) && number.length >= 11
+  end
+
+  def numeric?(str)
```
ghost: couldn't this be a regex instead, could use it to do the digit and length check all at once

ghost: soft of but not for decimal digits https://stackoverflow.com/questions/5416250/regex-contains-at-least-8-decimal-digits

kosiakkatrina: I think we use this method somewhere else to check for numbers and from what I understand regex can be quite slow for failing strings and I find this to be a bit more readable. Let me know if you feel strongly about this tho

ghost: this should do `^\d{11,}$`

ghost: for something this simple I think I'd prefer a regex
 I did not know of this slowness! do you have something for me to read about it?

ghost: I agree that the string length check is easier/comparable to read in ruby, but not sure this other check is? It includes casting + negation + checking for nil + rescuing a standard error and returning false

spec/features/user_spec.rb
```diff
@@ -346,6 +346,13 @@
         expect(page).to have_title("Error")
       end
 
+      it "allows telephone number in correct format" do
```
ghost: I think we could extract this check in a helper and unit test it rather than testing it with a slow feature spec

app/controllers/users_controller.rb
kosiakkatrina: I was looking at this, but I think you're right. I thought it would be way longer. Thanks! 
https://mentalized.net/journal/2011/04/14/ruby-how-to-check-if-a-string-is-numeric/

spec/features/user_spec.rb
kosiakkatrina: Moved them to request tests now, that should be quicker, let me know if that looks okay

ghost (approved):
nice one thanks Kat!

ghost (approved):

PR #1715, opened 2023-06-21
CLDC-2471 Add org id to about your org page

natdeanlewissoftwire:
This just adds the new CORE id (prepended with "ORG") to the about your org page.

ticket: https://digital.dclg.gov.uk/jira/browse/CLDC-2471

looks like this: 
image

### DIFF
--- a/app/models/organisation.rb
+++ b/app/models/organisation.rb
@@ -107,10 +107,10 @@ class Organisation < ApplicationRecord
       { name: "Name", value: name, editable: true },
       { name: "Organisation ID", value: "ORG#{id}", editable: false },
       { name: "Address", value: address_string, editable: true },
-      { name: "Telephone_number", value: phone, editable: true },
+      { name: "Telephone number", value: phone, editable: true },
       { name: "Type of provider", value: display_provider_type, editable: false },
       { name: "Registration number", value: housing_registration_no || "", editable: false },
-      { name: "Rent_periods", value: rent_period_labels, editable: false, format: :bullet },
+      { name: "Rent periods", value: rent_period_labels, editable: false, format: :bullet },
       { name: "Owns housing stock", value: holds_own_stock ? "Yes" : "No", editable: false },
     ].compact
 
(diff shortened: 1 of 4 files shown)


app/models/organisation.rb
```diff
@@ -105,11 +105,12 @@ def display_provider_type
   def display_organisation_attributes
     attrs = [
       { name: "Name", value: name, editable: true },
+      { name: "Organisation ID", value: "ORG#{id}", editable: false },
```
Airk0n, line 108: I know this is more a point on the ticket itself, but is there any risk that users will copy and paste this value, attach it to support tickets, try searching it etc and run into issues as it's just amended on a surface level? 
- Can the actual ID be a string and have it amended at an earlier point
- Is this documented clearly somewhere
- Can ORG be appended to the end of the field title so that when highlighted only the number is selected
- Or is this not potentially an issue at all :)

natdeanlewissoftwire: So the id here is prepended with ORG because there are actually two types of org IDs - orgs from old CORE have a different numbering, so every org in new CORE is prepended with ORG so that's clear in bulk upload, as users can do bulk upload for old core OR new core org ids depending on if they just give a number or ORG then a number.
We shouldn't amend ID much earlier than here (maybe there's an argument for a helper but it's not even conditional so I think this is fine in the model) because we went to be passing IDs as ints where possible throughout the codebase for convenience.
I agree with your point about users searching for orgs - except I don't think there's anywhere in the service where they can actually do that, you can't search for orgs in logs and you search by org name as a support user, so I think that's probably fine too.

natdeanlewissoftwire: re the documentation question, I would not at all be surprised if this is not documented clearly anywhere, I was mainly referring to the codebase to check this is consistent with the bulk upload pattern - I'll see if I can find it in docs anywhere - though it is in the bulk upload template so that's something

app/models/organisation.rb
```diff
@@ -105,11 +105,12 @@ def display_provider_type
...
       { name: "Address", value: address_string, editable: true },
-      { name: "Telephone_number", value: phone, editable: true },
+      { name: "Telephone number", value: phone, editable: true },
```
kosiakkatrina, line 110: Oh, nice one

kosiakkatrina (approved):

PR #1716, opened 2023-06-21
CLDC-2459 Update end date for new logs creation

kosiakkatrina:
This builds on top of [CLDC-2451-collection-year-dates](https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/tree/CLDC-2451-collection-year-dates)

* Rename `end_date` to `new_logs_end_date`
* Update `edit_end_date` to be at the end of the year. It would have to be manually updated every year when we're ready to close editing for the collection period
* Instead of `lettings/sales_in_edit_crossover_period?` use `lettings/sales_in_edit_crossover_period?` to determine if a log can still be edited
* Use `record.saledate_was` to determine if it's a new log, if it is `nil` only allow a log to be created until `new_logs_end_date` if `record.saledate_was` has a value, allow it to be updated until `edit_logs_end_date`
* Bulk upload is still using `new_logs_end_date` to determine whether it's in crossover period

### DIFF
--- a/spec/models/validations/sales/setup_validations_spec.rb
+++ b/spec/models/validations/sales/setup_validations_spec.rb
@@ -105,6 +105,52 @@ RSpec.describe Validations::Sales::SetupValidations do
           expect(record.errors[:saledate]).to include("Enter a date within the 23/24 or 24/25 collection years, which is between 1st April 2023 and 31st March 2025")
         end
       end
+
+      context "when after the new logs end date but before edit end date for the previous period" do
+        let(:record) { build(:sales_log, saledate: Time.zone.local(2025, 4, 1)) }
+
+        before do
+          allow(Time).to receive(:now).and_return(Time.zone.local(2024, 8, 8))
+        end
+
+        it "cannot create new logs for the previous collection year" do
+          record.update!(saledate: nil)
+          record.saledate = Time.zone.local(2024, 1, 1)
+          setup_validator.validate_saledate_collection_year(record)
+          expect(record.errors["saledate"]).to include(match "Enter a date within the 24/25 collection year, which is between 1st April 2024 and 31st March 2025")
+        end
+
+        it "can edit already created logs logs for the previous collection year" do
+          record.saledate = Time.zone.local(2024, 1, 2)
+          record.save!(validate: false)
+          record.saledate = Time.zone.local(2024, 1, 1)
+          setup_validator.validate_saledate_collection_year(record)
+          expect(record.errors["saledate"]).not_to include(match "Enter a date within the 24/25 collection year, which is between 1st April 2024 and 31st March 2025")
+        end
+      end
+
+      context "when after the new logs end date and after the edit end date for the previous period" do
+        let(:record) { build(:sales_log, saledate: Time.zone.local(2025, 4, 1)) }
+
+        before do
+          allow(Time).to receive(:now).and_return(Time.zone.local(2024, 12, 8))
+        end
+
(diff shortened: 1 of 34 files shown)


spec/models/validations/sales/setup_validations_spec.rb
```diff
@@ -105,6 +105,52 @@
           expect(record.errors[:saledate]).to include("Enter a date within the 23/24 or 24/25 collection years, which is between 1st April 2023 and 31st March 2025")
         end
       end
+
+      context "when after the new logs end date but before edit end date for the previous period" do
```
Airk0n: I'm guessing the linter want you to put 'when' before 'after' even though this makes sense starting with after :)

Airk0n: Is this when a sales log's sale date is after the new logs end date? I'm not sure how to reword this but I don't think I understand it without looking at the contents.

spec/models/validations/sales/setup_validations_spec.rb
```diff
@@ -105,6 +105,52 @@
...
+        end
+
+        it "cannot create new logs for the previous collection year" do
+          record.update!(saledate: nil)
+          record.saledate = Time.zone.local(2024, 1, 1)
+          setup_validator.validate_saledate_collection_year(record)
+          expect(record.errors["saledate"]).to include(match "Enter a date within the 24/25 collection year, which is between 1st April 2024 and 31st March 2025")
+        end
+
+        it "cannot edit already created logs logs for the previous collection year" do
```
Airk0n: minor point, but 'logs' x2

spec/models/validations/setup_validations_spec.rb
Airk0n, line 1: Similar points to the other test, I feel the wording could be a bit clearer

spec/requests/lettings_logs_controller_spec.rb
```diff
@@ -1095,6 +1095,29 @@
...
+        before do
+          Timecop.freeze(2022, 7, 5)
+        end
+
+        after do
+          Timecop.return
+        end
+
+        it "allows you to change the answers for previous collection year logs" do
+          get "/lettings-logs/#{completed_lettings_log.id}/setup/check-answers", headers: { "Accept" => "text/html" }, params: {}
```
Airk0n, line 1108: could this be a path helper( `lettings_log_path` ) instead of hard coded?

Airk0n (commented):
Lots of renaming there :)
Looks good, the wording thing is a bit subjective do don't worry about that too much.
I can see that other tests already have hard coded paths nearby however they will all have to change if a path changes for some reason so if a path helper exists for that I would use that.

spec/models/validations/sales/setup_validations_spec.rb
kosiakkatrina: Ah, good point, updated now, let me know if it's at all clearer!

spec/models/validations/sales/setup_validations_spec.rb
kosiakkatrina: Nice spot!

spec/requests/lettings_logs_controller_spec.rb
kosiakkatrina: We make requests using urls in the tests across the codebase, I don't know if there are any benefits doing it one way over the other. I think if we want to change it it's probably best to do it everywhere so that it stays consistent. I'd say it's something we can discuss outside of this PR

ghost (approved):

PR #1717, opened 2023-06-21
CLDC 2454: add answer option to q90 depending on q78

Airk0n:
This ticket ( https://digital.dclg.gov.uk/jira/browse/CLDC-2454 ) is concerned with adding the option 'don't know' to Q90 - 'Was a mortgage used for the purchase of this property?' if the answer to Q78 - 'What percentage of the property does the buyer now own in total?' is 100%.

This is achieved by looking at the value of the current log and showing different answers depending on the answer to Q78.


arfacamble (approved):
no further comments given we paired on this. looks good

arfacamble:
ticket title has wrong ticket number

PR #1718, opened 2023-06-21
CLDC 2439: apply 22/23 rent range soft validation to 23/24 

Airk0n:
This ticket - https://digital.dclg.gov.uk/jira/browse/CLDC-2439
Is concerned with displaying the validation that displays for letting logs for years 22/23 for letting logs for the years 23/24.
The existing implementation used a single rent value check page to cover both min and max validation.
This implementation uses a separate `min_rent_value_check` and `max_rent_value_check` as the hint text displayed is now different for each.


app/models/form/lettings/pages/min_rent_value_check.rb
```diff
@@ -10,19 +10,11 @@ def initialize(id, hsh, subsection, check_answers_card_number: nil)
...
-        "key" => "field_formatted_as_currency",
-        "arguments_for_key" => "soft_min_for_period",
-        "i18n_template" => "soft_min_for_period",
-      }],
-    }
     @check_answers_card_number = check_answers_card_number
   end
 
   def questions
-    @questions ||= [Form::Lettings::Questions::RentValueCheck.new(nil, nil, self, check_answers_card_number: @check_answers_card_number)]
```
arfacamble, line 25: you can delete this question if it's not being used anywhere anymore

app/models/form/lettings/questions/min_rent_value_check.rb
```diff
@@ -0,0 +1,19 @@
...
+    @check_answers_card_number = check_answers_card_number
+    @answer_options = ANSWER_OPTIONS
+    @hidden_in_check_answers = { "depends_on" => [{ "rent_value_check" => 0 }, { "rent_value_check" => 1 }] }
+  end
+
+  ANSWER_OPTIONS = { "0" => { "value" => "Yes" }, "1" => { "value" => "No" } }.freeze
+
+  def hint_text
+    "This is lower than we would expect. Check:<ul class=\"govuk-body-l app-panel--interruption\"><li>the decimal point</li><li>the frequency, for example every week or every calendar month</li><li>the rent type is correct, for example affordable or social rent</li></ul>"
+  end
```
arfacamble: since this method is just serving up a string now it could just be inline on line 8

config/locales/en.yml
```diff
@@ -595,8 +595,6 @@ en:
       over_soft_max_for_la_combined: "You told us the combined income of this household is %{combined_income}. This seems high.  Are you sure this is correct?"
     rent:
       outside_range_title: "You told us the rent is %{brent}"
-      min_hint_text: "The minimum rent expected for this type of property in this local authority is %{soft_min_for_period}."
-      max_hint_text: "The maximum rent expected for this type of property in this local authority is %{soft_max_for_period}."
```
arfacamble, line 599: did you double check these aren't being used anywhere else, ie. global search soft_validations.rent.min_hint_text

app/models/form/lettings/questions/min_rent_value_check.rb
arfacamble: We are actually able to call I18n.t from in these classes, I think that now we have separate question classes for min and max, this string could be put in the yml with one variable for lower/higher and called inline with @hint_text in each class. Better to have it in one place if poss

config/locales/en.yml
Airk0n: yes, however I want to check again now :)

app/models/form/lettings/pages/max_rent_value_check.rb
```diff
@@ -12,21 +12,12 @@ def initialize(id, hsh, subsection, check_answers_card_number: nil)
...
-      "translation" => "soft_validations.rent.max_hint_text",
-      "arguments" => [
-        {
-          "key" => "field_formatted_as_currency",
-          "arguments_for_key" => "soft_max_for_period",
-          "i18n_template" => "soft_max_for_period",
-        },
-      ],
-    }
+    @informative_text = {}
```
arfacamble: I think you can delete this line if there is no informative text

app/models/form/lettings/pages/min_rent_value_check.rb
arfacamble: unresolved this conversation as it looks like this file hasn't been deleted?

arfacamble: or is it being used somewhere else?

app/models/form/lettings/pages/min_rent_value_check.rb
```diff
@@ -10,19 +10,12 @@ def initialize(id, hsh, subsection, check_answers_card_number: nil)
...
     }
-    @informative_text = {
-      "translation" => "soft_validations.rent.min_hint_text",
-      "arguments" => [{
-        "key" => "field_formatted_as_currency",
-        "arguments_for_key" => "soft_min_for_period",
-        "i18n_template" => "soft_min_for_period",
-      }],
-    }
+    @informative_text = {}
```
arfacamble: as above

app/models/form/lettings/questions/min_rent_value_check.rb
arfacamble, line 1: ignore those comments. I see that github has treated it as an altered file and filename here

app/models/form/lettings/pages/max_rent_value_check.rb
Airk0n: I did originally, but I had to add it back as it was failing tests, I looked around and there are other instances of a blank informative_text field and I presume they ran into the same problem.

app/helpers/interruption_screen_helper.rb
```diff
@@ -1,5 +1,6 @@
 module InterruptionScreenHelper
   def display_informative_text(informative_text, log)
+    return informative_text if informative_text.is_a? String
```
arfacamble, line 3: let's just add one test to the spec for this helper that it now also accepts a string

arfacamble (changes requested):

arfacamble (approved):

PR #1719, opened 2023-06-22
CLDC-2291 refactor and minor updates to lettings log csv service

arfacamble:
https://digital.dclg.gov.uk/jira/browse/CLDC-2291

I decided against extracting code from the two csv services in the end. I thought it would be possible to commonise the `:value` method and some or all of the logic to convert the questions into the list of attributes for the headings, however in the end there are enough small differences between the two use cases that I don't believe it makes sense.

Other than this the PR almost entirely follows the pattern implemented in the sales log csv export work done previously ([PR link](https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/1568))


Airk0n (commented):
Looks good to me, there's a lot going on here so I can't claim to understand it all, but I see no issues :)

spec/services/csv/sales_log_csv_service_spec.rb
```diff
@@ -3,23 +3,27 @@
...
-    end
-    Timecop.return
-    Singleton.__init__(FormHandler)
-  end
-
   it "calls the form handler to get all questions in order when initialized" do
     allow(FormHandler).to receive(:instance).and_return(form_handler_mock)
     allow(form_handler_mock).to receive(:ordered_sales_questions_for_all_years).and_return([])
-    described_class.new(export_type: "codes")
+    service
```
kosiakkatrina, line 26: How come we need this?

spec/services/csv/lettings_log_csv_service_spec.rb
```diff
@@ -1,430 +1,227 @@
...
+  let(:service) { described_class.new(user:, export_type:) }
+  let(:export_type) { "labels" }
+  let(:csv) { CSV.parse(service.prepare_csv(LettingsLog.where(id: logs.map(&:id)))) }
+  let(:logs) { [log] }
+  let(:headers) { csv.first }
+
+  it "calls the form handler to get all questions in order when initialized" do
+    allow(FormHandler).to receive(:instance).and_return(form_handler_mock)
+    allow(form_handler_mock).to receive(:ordered_lettings_questions_for_all_years).and_return([])
+    service
```
kosiakkatrina, line 34: Would it make sense to put this tests (and all the mocks) into context and add these lines into a before block, as we  don't seem to be mocking form in other tests (?)

spec/services/csv/lettings_log_csv_service_spec.rb
```diff
@@ -1,430 +1,227 @@
...
+        expect(headers).to include(*question_ids.first(3))
+      end
+
+      it "removes some log attributes related to questions from the headers and replaces them with their derived values in the correct order" do
+        expect(headers).not_to include "rent_type"
+        expect(headers).to include(*%w[wrent renttype rent_type_detail])
+      end
     end
 
-    it "sets csv attributes in correct order" do
```
kosiakkatrina, line 13: Do we no longer want to test the order?

spec/services/csv/sales_log_csv_service_spec.rb
arfacamble: just that this initialises the service, and as part of that we call the form handler

spec/models/form_handler_spec.rb
```diff
@@ -270,5 +270,56 @@
...
+      first_section = build(:section, :with_questions, question_ids: %w[1 2 3])
+      second_section = build(:section, :with_questions, question_ids: %w[2 3 4 5])
+      lettings_form = FormFactory.new(year: 2936, type: "lettings")
+                              .with_sections([first_section, second_section])
+                              .build
+      described_class.instance.use_fake_forms!({ only_lettings: lettings_form })
+      expect(result.map(&:id)).to eq %w[1 2 3 4 5]
+    end
+
+    it "returns the questions in the same order as the form" do
```
kosiakkatrina, line 297: Ah, nice

spec/services/csv/sales_log_csv_service_spec.rb
arfacamble: maybe in this case to make this test explicit I should just inline `described_class.new(export_type: "labels")` which is what it resolves to

spec/services/csv/lettings_log_csv_service_spec.rb
arfacamble: it's tested implicitly in the full csv download tests

spec/services/csv/lettings_log_csv_service_spec.rb
arfacamble: I'm not sure I see the benefit of this. The form handler needs mocking here for this test only, and then below where there is a context with multiple tests then I pulled the mocking needed there into a before for that collection of tests.

spec/services/csv/lettings_log_csv_service_spec.rb
```diff
@@ -1,430 +1,227 @@
...
+  let(:csv) { CSV.parse(service.prepare_csv(LettingsLog.where(id: logs.map(&:id)))) }
+  let(:logs) { [log] }
+  let(:headers) { csv.first }
+
+  it "calls the form handler to get all questions in order when initialized" do
+    allow(FormHandler).to receive(:instance).and_return(form_handler_mock)
+    allow(form_handler_mock).to receive(:ordered_lettings_questions_for_all_years).and_return([])
+    service
+    expect(form_handler_mock).to have_received(:ordered_lettings_questions_for_all_years)
+  end
```
arfacamble, line 36: do you mean
```suggestion
context "when initialising" do
  before do
    allow(FormHandler).to receive(:instance).and_return(form_handler_mock)
    allow(form_handler_mock).to receive(:ordered_lettings_questions_for_all_years).and_return([])
  end

    it "calls the form handler to get all questions in order when initialized" do
    service
    expect(form_handler_mock).to have_received(:ordered_lettings_questions_for_all_years)
  end
end
```

kosiakkatrina (approved):

PR #1723, opened 2023-06-23
Fix bug in DPC import script

ghost:
Replicate the bug by:

- run Imports::DataProtectionConfirmationImportService for a DPO user that can't be found
- The script will [create a DPO user with "" email ](https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/blob/main/app/services/imports/data_protection_confirmation_import_service.rb#L18)
- any future run of ` Imports::DataProtectionConfirmationImportService` for a DPO user that can't be found will:
- -  not create a DPC because it thinks it already exists when instead the error is being raised when trying to create another DPO user with "" email 
- - output a misleading message saying the DPC already exists

Issues that need fixing:
- should not create a user with "" email but instead one with a unique one
- reimport all Data Protection Confirmations for orgs that don't have it

### DIFF
--- a/app/services/imports/data_protection_confirmation_import_service.rb
+++ b/app/services/imports/data_protection_confirmation_import_service.rb
@@ -8,6 +8,10 @@ module Imports
 
     def create_data_protection_confirmation(xml_document)
       org = Organisation.find_by(old_org_id: record_field_value(xml_document, "institution"))
+
+      return log_org_must_exist if org.blank?
+      return log_dpc_already_present(org) if org.data_protection_confirmed?
+
       dp_officer = User.find_by(
         name: record_field_value(xml_document, "dp-user"),
         organisation: org,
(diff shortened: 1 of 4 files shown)


natdeanlewissoftwire (approved):

app/services/imports/data_protection_confirmation_import_service.rb
```diff
@@ -8,6 +8,10 @@ def create_data_protection_confirmations(folder)
 
     def create_data_protection_confirmation(xml_document)
       org = Organisation.find_by(old_org_id: record_field_value(xml_document, "institution"))
+
+      return log_org_must_exist if org.blank?
+      return log_dpc_already_present(org) if org.data_protection_confirmed?
```
kosiakkatrina, line 13: Could 2 people have signed dsa in the old core?

ghost: no, there's only one document per organisation in old core

kosiakkatrina (approved):

PR #1725, opened 2023-06-26
CLDC-2484 Add rake task to resend invitation emails

kosiakkatrina:
There is a number of users on the service that haven't logged in once, we want to resend the invitation links to them, this PR introduces a rake task to do that

### DIFF
--- a/lib/tasks/resend_invitation_emails.rake
+++ b/lib/tasks/resend_invitation_emails.rake
@@ -1,0 +1,9 @@
+namespace :emails do
+  desc "Resend invitation emails"
+  task :resend_invitation_emails, %i[] => :environment do |_task, _args|
+    users = User.where(sign_in_count: 0, active: true)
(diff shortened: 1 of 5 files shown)


lib/tasks/resend_invitation_emails.rake
```diff
@@ -0,0 +1,9 @@
+namespace :emails do
+  desc "Resend invitation emails"
+  task :resend_invitation_emails, %i[] => :environment do |_task, _args|
+    users = User.where(sign_in_count: 0, active: true)
+    users.each(&:send_confirmation_instructions)
```
ghost: do you know:
- how many users that match this we have in prod? 
- if these emails are sent via a job or synchronously?

Just trying to avoid hammering Notify. Maybe a sleep or running this task in batches could be enough to avoid that.

ghost (approved):

PR #1726, opened 2023-06-27
CLDC-2365 Set managing org to owning org if there are no managing agents

kosiakkatrina:
When a support user tries to create a log for an organisation without any managing orgs, managing org was not routed to but it was also not being inferred so the setup section would be impossible to complete.

It worked fine for data coordinators, because the managing organisation would get inferred in the `create` log request in `org_params`

This PR updates managing organisation when a support user updates the owning organisation, because we can't set it before we know the owning organisation.

### DIFF
--- a/app/controllers/form_controller.rb
+++ b/app/controllers/form_controller.rb
@@ -104,7 +104,7 @@ private
 
       if current_user.support? && question.id == "owning_organisation_id" && @log.managing_organisation.blank?
         owning_organisation = Organisation.find(result["owning_organisation_id"])
-        if owning_organisation.present? && !owning_organisation.has_managing_agents?
+        if owning_organisation&.managing_agents&.empty?
           result["managing_organisation_id"] = owning_organisation.id
         end
       end
--- a/app/controllers/form_controller.rb
+++ b/app/controllers/form_controller.rb
@@ -101,6 +101,14 @@ private
       else
         result[question.id] = question_params
       end
+
+      if current_user.support? && question.id == "owning_organisation_id" && @log.managing_organisation.blank?
+        owning_organisation = Organisation.find(result["owning_organisation_id"])
+        if owning_organisation.present? && !owning_organisation.has_managing_agents?
+          result["managing_organisation_id"] = owning_organisation.id
+        end
+      end
+
       result
     end
   end
--- a/spec/requests/form_controller_spec.rb
+++ b/spec/requests/form_controller_spec.rb
@@ -135,6 +135,32 @@ RSpec.describe FormController, type: :request do
       end
     end
 
+    context "when owning organisation doesn't have any managing agents" do
+      let(:params) do
+        {
+          id: lettings_log.id,
+          lettings_log: {
+            page: "stock_owner",
+            owning_organisation_id: managing_organisation.id,
+          },
+        }
+      end
+
+      before do
+        lettings_log.update!(owning_organisation: nil, created_by: nil, managing_organisation: nil)
+        lettings_log.reload
+      end
+
+      it "sets managing organisation to owning organisation" do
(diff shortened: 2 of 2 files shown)


Airk0n:
> When a support user tries to create a log for an organisation without any managing orgs, managing org was not routed to but it was also not being inferred so the setup section would be impossible to complete.
> 
> It worked fine for data coordinators, because the managing organisation would get inferred in the `create` log request in `org_params`
> 
> This PR updates managing organisation when a support user updates the owning organisation, because we can't set it before we know the owning organisation.

When you say it's inferred for data controllers, am I right in saying that a data controller is part of an organisation and that organisation is being used, a support user isn't part of an organisation and thus doesn't have that?

spec/requests/form_controller_spec.rb
```diff
@@ -135,6 +135,32 @@
...
+        }
+      end
+
+      before do
+        lettings_log.update!(owning_organisation: nil, created_by: nil, managing_organisation: nil)
+        lettings_log.reload
+      end
+
+      it "sets managing organisation to owning organisation" do
+        post "/lettings-logs/#{lettings_log.id}/stock-owner", params: params
```
Airk0n, line 155: Can a path helper be used for this?

app/controllers/form_controller.rb
```diff
@@ -101,6 +101,14 @@ def responses_for_page(page)
       else
         result[question.id] = question_params
       end
+
```
Airk0n, line 104: I don't know where this is in the path, but is there a risk of a support user creating, or trying to create a log for an org that hasn't accepted the data agreement?

Airk0n (commented):
Looks good to me, my comments are just questions for me to understand better rather than issues :)

spec/requests/form_controller_spec.rb
kosiakkatrina: Similar answer to this: https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/1716#discussion_r1238704714

app/controllers/form_controller.rb
kosiakkatrina: I think that there should be validations around it. Not sure if they're implemented yet, but I'd say that's a separate task from this

kosiakkatrina:
> > When a support user tries to create a log for an organisation without any managing orgs, managing org was not routed to but it was also not being inferred so the setup section would be impossible to complete.
> > It worked fine for data coordinators, because the managing organisation would get inferred in the `create` log request in `org_params`
> > This PR updates managing organisation when a support user updates the owning organisation, because we can't set it before we know the owning organisation.
> 
> When you say it's inferred for data controllers, am I right in saying that a data controller is part of an organisation and that organisation is being used, a support user isn't part of an organisation and thus doesn't have that?

@Airk0n Yes, because data coordinators belong to an organisation, the logs get assigned to their organisation straightaway (they can be changed in some cases, but either owning or managing organisation has to be the one that data coordinator belongs to). And in the case of support users they can create a log for any organisation, so we can't infer anything until an organisation is chosen

app/controllers/form_controller.rb
```diff
@@ -101,6 +101,14 @@ def responses_for_page(page)
...
+      if current_user.support? && question.id == "owning_organisation_id" && @log.managing_organisation.blank?
+        owning_organisation = Organisation.find(result["owning_organisation_id"])
+        if owning_organisation.present? && !owning_organisation.has_managing_agents?
```
ghost: `owning_organisation.present? && !owning_organisation.has_managing_agents?` could just be 

`owning_organisation&.managing_agents.empty?`

ghost (approved):

app/controllers/form_controller.rb
kosiakkatrina: good point

PR #1729, opened 2023-06-27
CLDC-1800 Update provider relationships permissions

kosiakkatrina:
Data providers should not be able to:
- remove housing providers from organisation relationships
- add new housing providers to organisation relationships
- remove managing agents from organisation relationships (CLDC-1799)
- add new managing agents to organisation relationships (CLDC-1799)

### DIFF
--- a/app/policies/organisation_relationship_policy.rb
+++ b/app/policies/organisation_relationship_policy.rb
@@ -15,7 +15,7 @@ class OrganisationRelationshipPolicy
     remove_managing_agent?
   ].each do |method_name|
     define_method method_name do
-      return true unless user.data_provider?
+      !user.data_provider?
     end
   end
 end
--- a/app/policies/organisation_relationship_policy.rb
+++ b/app/policies/organisation_relationship_policy.rb
@@ -13,4 +13,8 @@ class OrganisationRelationshipPolicy
   def remove_stock_owner?
     return true unless user.data_provider?
   end
+
+  def create_managing_agent?
+    return true unless user.data_provider?
+  end
 end
--- a/app/policies/organisation_relationship_policy.rb
+++ b/app/policies/organisation_relationship_policy.rb
@@ -1,0 +1,12 @@
+class OrganisationRelationshipPolicy
+  attr_reader :user, :organisation_relationship
+
+  def initialize(user, organisation_relationship)
+    @user = user
+    @organisation_relationship = organisation_relationship
+  end
+
+  def create_stock_owner?
+    return true unless user.data_provider?
+  end
+end
--- a/app/policies/organisation_relationship_policy.rb
+++ b/app/policies/organisation_relationship_policy.rb
@@ -6,27 +6,16 @@ class OrganisationRelationshipPolicy
     @organisation_relationship = organisation_relationship
   end
 
-  def add_stock_owner?
-    return true unless user.data_provider?
-  end
-
-  def create_stock_owner?
-    return true unless user.data_provider?
-  end
-
-  def remove_stock_owner?
-    return true unless user.data_provider?
-  end
-
-  def add_managing_agent?
-    return true unless user.data_provider?
-  end
-
-  def create_managing_agent?
-    return true unless user.data_provider?
-  end
-
-  def remove_managing_agent?
-    return true unless user.data_provider?
+  %w[
+    add_stock_owner?
+    create_stock_owner?
+    remove_stock_owner?
+    add_managing_agent?
+    create_managing_agent?
+    remove_managing_agent?
+  ].each do |method_name|
+    define_method method_name do
+      return true unless user.data_provider?
+    end
   end
 end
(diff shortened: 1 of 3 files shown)


ghost (approved):

app/policies/organisation_relationship_policy.rb
```diff
@@ -0,0 +1,21 @@
...
+  %w[
+    add_stock_owner?
+    create_stock_owner?
+    remove_stock_owner?
+    add_managing_agent?
+    create_managing_agent?
+    remove_managing_agent?
+  ].each do |method_name|
+    define_method method_name do
+      return true unless user.data_provider?
```
arfacamble: very minor but...

```suggestion
      !user.data_provider?
```

spec/requests/organisation_relationships_controller_spec.rb
```diff
@@ -296,6 +296,94 @@
...
+        end
+
+        it "does not create a new organisation relationship" do
+          expect { request }.not_to change(OrganisationRelationship, :count)
+        end
+      end
+
+      context "when directly removing a managing agent" do
+        let(:managing_agent) { FactoryBot.create(:organisation) }
+        let(:request) { get "/organisations/#{organisation.id}/managing-agents/remove?target_organisation_id=#{managing_agent.id}", headers: }
```
arfacamble: from what I've read one of the benefits of using request specs is that it allows you to implicitly test your routing. Would it be possible to use a path helper here?

arfacamble: I've also found that in many places where I've tried removing the `headers:` part in these tests it works fine as the headers are just set to the default html at the top and never changed

arfacamble (commented):
In a broad sense. Are we happy as a team about starting to have a mix of authenticate_scope and pundit for authorisation? Do we have a rule of thumb for when to use which? Or do we just want to use pundit as the default from now on and move things over as tech debt where we feel appropriate?

It would be helpful for this PR to have seen one or two screenshots as I'm not familiar with this part of the app. I imagine that somewhere there was a remove or add relationship button that is now no longer visible for data providers? I can see they can no longer hit the endpoints. But where is the entry point to this journey for hte user?

spec/requests/organisation_relationships_controller_spec.rb
kosiakkatrina: Happy to use helpers instead of urls, but it would make sense to keep the tests consistent. Also happy to discuss this as a tech debt if we think that we want to change all of these.

Removed headers, nice spot!

app/policies/organisation_relationship_policy.rb
kosiakkatrina: nice

kosiakkatrina:
> In a broad sense. Are we happy as a team about starting to have a mix of authenticate_scope and pundit for authorisation? Do we have a rule of thumb for when to use which? Or do we just want to use pundit as the default from now on and move things over as tech debt where we feel appropriate?
> 
> It would be helpful for this PR to have seen one or two screenshots as I'm not familiar with this part of the app. I imagine that somewhere there was a remove or add relationship button that is now no longer visible for data providers? I can see they can no longer hit the endpoints. But where is the entry point to this journey for hte user?

@arfacamble Hmm, good question. I’ve implemented it as a policy, because it’s nice to have it separated and we’re already using pundit in a few places. If it’s something we need to talk about in terms of tech debt, happy to talk with the team in a wider context of updating these.

So the button to add or remove stock owner/managing agent has already been correctly hidden from the providers, the only issue is if they navigate directly to the confirmation page using a url, they were previously able to access it. If I understood the task correctly, data providers should not be able to access those pages at all. Let me know if that’s still a bit confusing and we can talk through some examples.

arfacamble:
> > In a broad sense. Are we happy as a team about starting to have a mix of authenticate_scope and pundit for authorisation? Do we have a rule of thumb for when to use which? Or do we just want to use pundit as the default from now on and move things over as tech debt where we feel appropriate?
> > It would be helpful for this PR to have seen one or two screenshots as I'm not familiar with this part of the app. I imagine that somewhere there was a remove or add relationship button that is now no longer visible for data providers? I can see they can no longer hit the endpoints. But where is the entry point to this journey for hte user?
> 
> @arfacamble Hmm, good question. I’ve implemented it as a policy, because it’s nice to have it separated and we’re already using pundit in a few places. If it’s something we need to talk about in terms of tech debt, happy to talk with the team in a wider context of updating these.
> 
> So the button to add or remove stock owner/managing agent has already been correctly hidden from the providers, the only issue is if they navigate directly to the confirmation page using a url, they were previously able to access it. If I understood the task correctly, data providers should not be able to access those pages at all. Let me know if that’s still a bit confusing and we can talk through some examples.

arfacamble (approved):
the button hiding makes sense. as in slack, I suspected it was already done there and we were just protecting endpoints.

re using path helpers in tests I think we're already reasonably inconsistent, up to you

app/policies/organisation_relationship_policy.rb
ghost: `user.data_provider.blank?`

PR #1731, opened 2023-06-28
Reimport offered

natdeanlewissoftwire:
Slight update to previously merged PR from this branch, just updating the rake task


app/services/imports/lettings_logs_field_import_service.rb
```diff
@@ -2,7 +2,7 @@ module Imports
   class LettingsLogsFieldImportService < LogsImportService
     def update_field(field, folder)
       case field
-      when "tenancycode"
+      when "tenant_code"
```
kosiakkatrina: Should it be `tenancycode` and needs changing in the rake task instead? I think `tenancycode` is the column in lettings logs table

natdeanlewissoftwire: I don't think the field is actually used, it just selects the method in this case block - but now you mention it the error message should probably have the exact field so I'll change this

kosiakkatrina (approved):

PR #1732, opened 2023-06-28
CLDC 2405: Show more information on the sales log info component

Airk0n:
This ticket - https://digital.dclg.gov.uk/jira/browse/CLDC-2405 is concerned with exposing the purchaser id, the sale type and the sale completion date to users viewing the list of sales logs in a similar way to how the data is presented on the lettings log list.

To implement this I split the log summary component into a sales and lettings log as the single component was starting to show lots of clearly different behaviour based on if the log was sales or lettings.

It was suggested to me that an alternative approach would have been move internal logic up a level and have the component decoupled from a log type and simply receive 'heading 1' 'sub heading text' etc (as I believe the original component was intended).

If a third log type is introduced in the future then that would how I would be the way I would change this component.

### DIFF
--- a/app/models/sales_log.rb
+++ b/app/models/sales_log.rb
@@ -384,12 +384,12 @@ class SalesLog < Log
   end
 
   def ownership_scheme(uppercase: false)
-    owner_scheme = case ownershipsch
+    ownership_scheme = case ownershipsch
                    when 1 then "shared ownership"
                    when 2 then "discounted ownership"
                    when 3 then "outright or other sale"
                    end
-    uppercase ? owner_scheme.capitalize : owner_scheme
+    uppercase ? ownership_scheme.capitalize : ownership_scheme
   end
 
   def combined_income
--- a/app/models/sales_log.rb
+++ b/app/models/sales_log.rb
@@ -383,12 +383,13 @@ class SalesLog < Log
     beds.nil? ? nil : [beds, LaSaleRange::MAX_BEDS].min
   end
 
-  def ownership_scheme
-    case ownershipsch
-    when 1 then "shared ownership"
-    when 2 then "discounted ownership"
-    when 3 then "outright sale"
-    end
+  def ownership_scheme(uppercase: false)
+    owner_scheme = case ownershipsch
+                   when 1 then "shared ownership"
+                   when 2 then "discounted ownership"
+                   when 3 then "outright or other sale"
+                   end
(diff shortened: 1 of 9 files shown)


app/models/sales_log.rb
```diff
@@ -61,6 +61,10 @@ def self.editable_fields
     attribute_names
   end
```
Airk0n, line 63: This felt redundant to implement as though I should have access to the underlying log in the view however It is being done for other fields, I'm guessing the log itself is a database object and we're working with an object representation of it instead of the raw data? I'd be interested in the distinction between the two.

arfacamble: this does look redundant. I would expect that wherever you are calling sales_log.purchaser_code you would be able to call sales_log.purchid

In this case I'm personally fine with it anyway since purchid is meaningless to humans

app/models/sales_log.rb
```diff
@@ -379,12 +383,13 @@ def beds_for_la_sale_range
...
 
-  def ownership_scheme
-    case ownershipsch
-    when 1 then "shared ownership"
-    when 2 then "discounted ownership"
-    when 3 then "outright sale"
-    end
+  def ownership_scheme(uppercase: false)
+    owner_scheme = case ownershipsch
```
arfacamble: can this variable be called ownership_scheme please or just scheme or scheme_title if that causes some kind of naming conflict with the method itself

app/views/logs/_log_list.html.erb
```diff
@@ -11,12 +11,15 @@
...
   <% end %>
 </div>
 </div>
 </h2>
 <% logs.map do |log| %>
-  <%= render(LogSummaryComponent.new(current_user:, log:)) %>
+  <% if log.sales? %>
+    <%= render(SalesLogSummaryComponent.new(current_user:, log:)) %>
+    <% else %>
+    <%= render(LettingsLogSummaryComponent.new(current_user:, log:)) %>
```
arfacamble: very minor point, but render is one place where it's most common not to use parentheses around arguments in the method call in rails
```suggestion
    <%= render LettingsLogSummaryComponent.new(current_user:, log:) %>
```

app/models/sales_log.rb
Airk0n: I had to reread that 4 times to spot it on 387, yes good spot :)

arfacamble (approved):

PR #1739, opened 2023-06-29
CLDC-1770 Activating soon locations validations

natdeanlewissoftwire:
Adds validations for schemes and locations depending on status when compared with tenancy start date for supported housing logs. Some of these were already present, but the set is now complete. Note, we also now allow schemes/locations to be selected if they will activate in the next 2 weeks but the tenancy start date must be strictly after the activation date.

ticket: https://dluhcdigital.atlassian.net/browse/CLDC-1770

### DIFF
--- a/app/models/validations/shared_validations.rb
+++ b/app/models/validations/shared_validations.rb
@@ -71,6 +71,7 @@ module Validations::SharedValidations
       date, scope, deactivation_date = location_inactive_status.values_at(:date, :scope, :deactivation_date)
       record.errors.add :startdate, :not_active, message: I18n.t("validations.setup.startdate.location.#{scope}.startdate", postcode: record.location.postcode, date:, deactivation_date:)
       record.errors.add :location_id, :not_active, message: I18n.t("validations.setup.startdate.location.#{scope}.location_id", postcode: record.location.postcode, date:, deactivation_date:)
+      record.errors.add :scheme_id, :not_active, message: I18n.t("validations.setup.startdate.location.#{scope}.location_id", postcode: record.location.postcode, date:, deactivation_date:)
     end
   end
 
@@ -98,8 +99,7 @@ module Validations::SharedValidations
            when :deactivated then open_deactivation.deactivation_date
            end
 
-    scope = status
-    { scope:, date: date&.to_formatted_s(:govuk_date), deactivation_date: closest_reactivation&.deactivation_date&.to_formatted_s(:govuk_date) }
+    { scope: status, date: date&.to_formatted_s(:govuk_date), deactivation_date: closest_reactivation&.deactivation_date&.to_formatted_s(:govuk_date) }
   end
 
   def shared_validate_partner_count(record, max_people)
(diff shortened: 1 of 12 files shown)


app/models/validations/shared_validations.rb
```diff
@@ -64,20 +64,22 @@ def validate_property_postcode(record)
...
 
-  def location_during_startdate_validation(record, field)
+  def location_during_startdate_validation(record)
     location_inactive_status = inactive_status(record.startdate, record.location)
 
     if location_inactive_status.present?
       date, scope, deactivation_date = location_inactive_status.values_at(:date, :scope, :deactivation_date)
-      record.errors.add field, :not_active, message: I18n.t("validations.setup.startdate.location.#{scope}", postcode: record.location.postcode, date:, deactivation_date:)
+      record.errors.add :startdate, :not_active, message: I18n.t("validations.setup.startdate.location.#{scope}.startdate", postcode: record.location.postcode, date:, deactivation_date:)
+      record.errors.add :location_id, :not_active, message: I18n.t("validations.setup.startdate.location.#{scope}.location_id", postcode: record.location.postcode, date:, deactivation_date:)
```
kosiakkatrina, line 73: we never add this validation to :scheme_id, although we used to? I suspect we did it because location id is sometimes inferred (?)

app/models/validations/shared_validations.rb
```diff
@@ -96,7 +98,8 @@ def inactive_status(date, resource)
            when :deactivated then open_deactivation.deactivation_date
            end
 
-    { scope: status, date: date&.to_formatted_s(:govuk_date), deactivation_date: closest_reactivation&.deactivation_date&.to_formatted_s(:govuk_date) }
+    scope = status
+    { scope:, date: date&.to_formatted_s(:govuk_date), deactivation_date: closest_reactivation&.deactivation_date&.to_formatted_s(:govuk_date) }
```
kosiakkatrina: What is the benefit of this?

config/locales/en.yml
```diff
@@ -248,13 +248,22 @@ en:
         ten_years_after_mrc_date: "Enter a tenancy start date that is no more than 10 years after the major repairs completion date"
 
         location:
-          deactivated: "The location %{postcode} was deactivated on %{date} and was not available on the day you entered."
-          reactivating_soon: "The location %{postcode} is not available until %{date}. Select another location or edit the tenancy start date"
-          activating_soon: "The location %{postcode} is not available until %{date}. Enter a tenancy start date after %{date}"
+          deactivated:
+            startdate: "The location %{postcode} was deactivated on %{date} and was not available on the day you entered. Enter a tenancy start date after %{date}"
```
kosiakkatrina: Does this make sense? If the location was deactivated wouldn't we want to enter a date `before` it was deactivated instead of `after`?

app/models/validations/shared_validations.rb
natdeanlewissoftwire: ah fair point this is from a previous approach, cheers

config/locales/en.yml
natdeanlewissoftwire: good point, have updated this copy :)

kosiakkatrina (approved):

PR #1740, opened 2023-06-29
Fix interruption screen success banner

kosiakkatrina:
```
undefined method `downcase' for nil:NilClass
flash[:notice] = "You have successfully updated #{@page.questions.map(&:check_answer_label).first.downcase}" if previous_interruption_screen_page_id.present?
```

This error happens because we have some empty check_answer_label values, we should now be filtering them out.

### DIFF
--- a/app/controllers/form_controller.rb
+++ b/app/controllers/form_controller.rb
@@ -11,7 +11,7 @@ class FormController < ApplicationController
       mandatory_questions_with_no_response = mandatory_questions_with_no_response(responses_for_page)
 
       if mandatory_questions_with_no_response.empty? && @log.update(responses_for_page.merge(updated_by: current_user))
-        flash[:notice] = "You have successfully updated #{@page.questions.map(&:check_answer_label).first.downcase}" if previous_interruption_screen_page_id.present?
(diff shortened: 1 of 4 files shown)


app/controllers/form_controller.rb
```diff
@@ -11,7 +11,7 @@ def submit_form
       mandatory_questions_with_no_response = mandatory_questions_with_no_response(responses_for_page)
 
       if mandatory_questions_with_no_response.empty? && @log.update(responses_for_page.merge(updated_by: current_user))
-        flash[:notice] = "You have successfully updated #{@page.questions.map(&:check_answer_label).first.downcase}" if previous_interruption_screen_page_id.present?
+        flash[:notice] = "You have successfully updated #{@page.questions.map(&:check_answer_label).reject { |e| e.to_s.empty? }.first&.downcase}" if previous_interruption_screen_page_id.present?
```
natdeanlewissoftwire: what does e stand for here? Unless I'm being stupid I don't think it's obvious, could it be named something clearer?

natdeanlewissoftwire (approved):
just 1 suggestion :)

app/controllers/form_controller.rb
kosiakkatrina: Sure! done!

PR #1744, opened 2023-06-30
CLDC-2310 add creation method to both log import services, lettings log field import and create sales log field import

arfacamble:
https://digital.dclg.gov.uk/jira/browse/CLDC-2310

- add creation method to:
  - both log import services
  - lettings log field import

- create sales log field import

DEPLOY TASK ALERT this ticket contains a rake task to run when deployed

### DIFF
--- a/app/services/imports/lettings_logs_field_import_service.rb
+++ b/app/services/imports/lettings_logs_field_import_service.rb
@@ -110,31 +110,5 @@ module Imports
         @logger.warn("Could not find record matching legacy ID #{old_id}")
       end
     end
-
-    def compose_date(xml_doc, day_str, month_str, year_str)
-      day = Integer(field_value(xml_doc, "xmlns", day_str), exception: false)
-      month = Integer(field_value(xml_doc, "xmlns", month_str), exception: false)
-      year = Integer(field_value(xml_doc, "xmlns", year_str), exception: false)
-      if day.nil? || month.nil? || year.nil?
-        nil
-      else
-        Time.zone.local(year, month, day)
-      end
-    end
-
-    def string_or_nil(xml_doc, attribute)
-      str = field_value(xml_doc, "xmlns", attribute)
-      str.presence
-    end
-
-    # Unsafe: A string that has more than just the integer value
-    def unsafe_string_as_integer(xml_doc, attribute)
-      str = string_or_nil(xml_doc, attribute)
-      if str.nil?
-        nil
-      else
-        str.to_i
-      end
-    end
   end
 end
--- a/app/services/imports/lettings_logs_field_import_service.rb
+++ b/app/services/imports/lettings_logs_field_import_service.rb
@@ -10,6 +10,8 @@ module Imports
         import_from(folder, :update_lettings_allocation)
       when "offered"
         import_from(folder, :update_offered)
+      when "creation_method"
+        import_from(folder, :update_creation_method)
       else
         raise "Updating #{field} is not supported by the field import service"
       end
@@ -33,6 +35,29 @@ module Imports
         end
       else
         @logger.warn("lettings log with old id #{old_id} not found")
+      end
+    end
+
+    def update_creation_method(xml_doc)
+      old_id = meta_field_value(xml_doc, "document-id")
+      log = LettingsLog.find_by(old_id:)
+
+      if log
+        upload_method = meta_field_value(xml_doc, "upload-method")
+
+        case upload_method
+        when "Manual Entry"
+          @logger.info "lettings log with old id #{old_id} entered manually, no need for update"
+        when "Bulk Upload"
+          if log.creation_method == "bulk upload"
+            @logger.info "lettings log #{log.id} creation method already set to bulk upload, no need for update"
+          else
+            log.update!(creation_method: "bulk upload")
+            @logger.info "lettings log #{log.id} creation method set to bulk upload"
+          end
+        end
+      else
+        @logger.warn "lettings log with old id #{old_id} not found"
       end
     end
 
--- a/app/services/imports/lettings_logs_field_import_service.rb
+++ b/app/services/imports/lettings_logs_field_import_service.rb
@@ -43,12 +43,11 @@ module Imports
       log = LettingsLog.find_by(old_id:)
 
       if log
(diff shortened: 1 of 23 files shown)


app/services/imports/lettings_logs_field_import_service.rb
```diff
@@ -110,31 +135,5 @@ def update_tenant_code(xml_doc)
...
-
-    # Unsafe: A string that has more than just the integer value
-    def unsafe_string_as_integer(xml_doc, attribute)
-      str = string_or_nil(xml_doc, attribute)
-      if str.nil?
-        nil
-      else
-        str.to_i
-      end
-    end
```
arfacamble, line 138: these methods exist on the parent class already

app/services/imports/logs_import_service.rb
```diff
@@ -11,11 +11,7 @@ def safe_string_as_integer(xml_doc, attribute)
     # Unsafe: A string that has more than just the integer value
     def unsafe_string_as_integer(xml_doc, attribute)
       str = string_or_nil(xml_doc, attribute)
-      if str.nil?
-        nil
-      else
-        str.to_i
-      end
+      str&.to_i
```
arfacamble, line 14: someone sense check me here, but if str is nil we get nil, otherwise we get str.to_i

app/services/imports/lettings_logs_field_import_service.rb
```diff
@@ -36,6 +38,29 @@ def update_offered(xml_doc)
...
+        upload_method = meta_field_value(xml_doc, "upload-method")
+
+        case upload_method
+        when "Manual Entry"
+          @logger.info "lettings log with old id #{old_id} entered manually, no need for update"
+        when "Bulk Upload"
+          if log.creation_method == "bulk upload"
+            @logger.info "lettings log #{log.id} creation method already set to bulk upload, no need for update"
+          else
+            log.update!(creation_method: "bulk upload")
```
jamesrose: This should be an enum.

app/services/imports/lettings_logs_field_import_service.rb
```diff
@@ -36,6 +38,29 @@ def update_offered(xml_doc)
...
+      log = LettingsLog.find_by(old_id:)
+
+      if log
+        upload_method = meta_field_value(xml_doc, "upload-method")
+
+        case upload_method
+        when "Manual Entry"
+          @logger.info "lettings log with old id #{old_id} entered manually, no need for update"
+        when "Bulk Upload"
+          if log.creation_method == "bulk upload"
```
jamesrose:
```suggestion
          if log.bulk_upload_creation_method?
```

app/services/imports/lettings_logs_field_import_service.rb
```diff
@@ -36,6 +38,29 @@ def update_offered(xml_doc)
...
+        case upload_method
+        when "Manual Entry"
+          @logger.info "lettings log with old id #{old_id} entered manually, no need for update"
+        when "Bulk Upload"
+          if log.creation_method_bulk_upload?
+            @logger.info "lettings log #{log.id} creation method already set to bulk upload, no need for update"
+          else
+            log.creation_method_bulk_upload!
+            @logger.info "lettings log #{log.id} creation method set to bulk upload"
+          end
```
kosiakkatrina: Would there be a situation where the log already has the creation method set? Wouldn't just having the else part of the statement achieve the same result?

app/services/imports/lettings_logs_field_import_service.rb
```diff
@@ -36,6 +38,29 @@ def update_offered(xml_doc)
...
+        end
+      else
+        @logger.warn "lettings log with old id #{old_id} not found"
+      end
```
kosiakkatrina: Would a guard clause at the top instead of this else be more readable?

app/services/imports/lettings_logs_field_import_service.rb
kosiakkatrina: nice

app/services/imports/lettings_logs_field_import_service.rb
arfacamble: the creation method is always set, with a default of "single log"
49 - if the old log is manual entry, nothing to do
52 - if it's already set to bulk upload, we've been here before
54 - make it bulk upload

arfacamble: oh I think I see what you mean, if the old log is bulk upload, why not just set it to bulk upload in any case and have fewer lines?

for that I was following the model of the existing methods in the field import services, which seemed to check whether the field was already set correctly and log that if so

app/services/imports/logs_import_service.rb
```diff
@@ -29,6 +25,16 @@ def compose_date(xml_doc, day_str, month_str, year_str)
...
 
+    CREATION_METHODS = {
+      "Bulk Upload" => "bulk upload",
+      "Manual Entry" => "single log",
+    }.freeze
+
+    def get_creation_method(xml_doc)
+      upload_method = meta_field_value(xml_doc, "upload-method")
+      CREATION_METHODS[upload_method]
+    end
```
kosiakkatrina: `get_` is quite java like, maybe we could rename the method to just `creation_method` ?

app/services/imports/lettings_logs_field_import_service.rb
arfacamble: I considered that and then followed the model of the other methods in these classes. I kind fo think they're equal, but it does reduce one level of nesting

arfacamble: I was going to change it and then I realised that because we need to log a warning and return, you can't one line it, so you don't really reduce nesting without adding more lines, so I'll eave it as is unless you feel strongly

kosiakkatrina: Would `return @logger.warn "lettings log with old id #{old_id} not found" unless log` not work?

kosiakkatrina (approved):

app/services/imports/lettings_logs_field_import_service.rb
arfacamble: 👍

kosiakkatrina (approved):

PR #1745, opened 2023-06-30
CLDC-2492-create-rake-task-to-set-creation-method-for-existing-logs

arfacamble:
create rake task to set correct creation method values for existing logs

https://digital.dclg.gov.uk/jira/browse/CLDC-2492

Last A/C on ticket forgotten in original PR

ALERT FOR DEPLOYMENT - rake task to run


lib/tasks/creation_method.rake
```diff
@@ -0,0 +1,5 @@
+desc "set creation method to bulk upload if a log has a bulk upload id"
+task set_creation_method: :environment do
+  LettingsLog.where.not(bulk_upload_id: nil).update_all(creation_method: "bulk upload")
```
jamesrose, line 3: The value of `creation_method` should be an enum, not a free text field.

PR #1748, opened 2023-07-03
CLDC 1863: Prevent non-stock organisations appearing in search for who owns property questions.

Airk0n:
This ticket - https://dluhcdigital.atlassian.net/browse/CLDC-1863 is concerned with preventing non-stock owning organisations from appearing to support users as a valid option for who owns a property.

image

Previously we would have seen the non-stock owning organisation as an available option.

### DIFF
--- a/spec/models/form/lettings/questions/stock_owner_spec.rb
+++ b/spec/models/form/lettings/questions/stock_owner_spec.rb
@@ -107,15 +107,17 @@ RSpec.describe Form::Lettings::Questions::StockOwner, type: :model do
 
       let(:log) { create(:lettings_log) }
 
+      let(:non_stock_organisation) { create(:organisation, holds_own_stock: false) }
       let(:expected_opts) do
-        Organisation.all.each_with_object(options) do |organisation, hsh|
+        Organisation.where(holds_own_stock: true).each_with_object(options) do |organisation, hsh|
           hsh[organisation.id] = organisation.name
           hsh
         end
       end
 
-      it "shows all orgs" do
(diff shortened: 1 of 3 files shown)


spec/models/form/lettings/questions/stock_owner_spec.rb
```diff
@@ -107,15 +107,17 @@
...
-        Organisation.all.each_with_object(options) do |organisation, hsh|
+        Organisation.where(holds_own_stock: true).each_with_object(options) do |organisation, hsh|
           hsh[organisation.id] = organisation.name
           hsh
         end
       end
 
-      it "shows all orgs" do
+      it "shows orgs where organisation holds own stock" do
         expect(question.displayed_answer_options(log, user)).to eq(expected_opts)
```
arfacamble, line 119: small one, but now that you have limited the organisations here, have you put a binding in to check there is at least one?

arfacamble (approved):

spec/models/form/lettings/questions/stock_owner_spec.rb
Airk0n: I didn't use binding, but I did change the to and not_to around, to see what I got, and yes there was at least one

arfacamble: if there were no stock owning organisations then `expect({}).not_to eq({})` would fail

Airk0n: And it did fail :)

PR #1749, opened 2023-07-04
CLDC-1636 Update new logs end date tests

kosiakkatrina:
We will need to eventually move the new end date forward and there are a lot of tests that will fail when that happens. This PR adds some stubs to reduce the number of affected tests.

### DIFF
--- a/spec/services/csv/sales_log_csv_service_spec.rb
+++ b/spec/services/csv/sales_log_csv_service_spec.rb
@@ -5,7 +5,7 @@ RSpec.describe Csv::SalesLogCsvService do
   let(:organisation) { create(:organisation) }
   let(:fixed_time) { Time.zone.local(2023, 2, 8) }
   let(:user) { create(:user, email: "billyboy@eyeKLAUD.com") }
-  let!(:log) do
+  let(:log) do
     create(
       :sales_log,
       :completed,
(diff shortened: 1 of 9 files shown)


app/models/form.rb
```diff
@@ -39,7 +39,7 @@ def initialize(form_path, start_year = "", sections_in_form = [], type = "lettin
       @pages = subsections.flat_map(&:pages)
       @questions = pages.flat_map(&:questions)
       @start_date = Time.iso8601(form_definition["start_date"])
-      @new_logs_end_date = Time.zone.local(@start_date.year + 1, 12, 31)
+      @new_logs_end_date = Time.zone.local(@start_date.year + 1, 6, 10)
```
kosiakkatrina: This is for review apps only, will revert before merging

app/services/feature_toggle.rb
```diff
@@ -1,14 +1,20 @@
...
 
   def self.startdate_two_week_validation_enabled?
+    return true
+
     Rails.env.production? || Rails.env.test? || Rails.env.staging?
   end
 
   def self.saledate_two_week_validation_enabled?
+    return true
+
```
kosiakkatrina: This is for review apps only, will revert before merging

spec/services/csv/sales_log_csv_service_spec.rb
```diff
@@ -20,6 +20,12 @@
   let(:service) { described_class.new(export_type: "labels") }
   let(:csv) { CSV.parse(service.prepare_csv(SalesLog.all)) }
 
+  before do
+    allow(Time).to receive(:now).and_return(fixed_time)
+    Singleton.__init__(FormHandler)
+    log
+  end
```
arfacamble, line 27: Is there a reason we're not doing this the "normal" Timecop way?

arfacamble (approved):

spec/services/csv/sales_log_csv_service_spec.rb
kosiakkatrina: Not really, I don't think it matters, just means we don't have to unfreeze it again

arfacamble: no doesn't matter much. Of course if you do around with Timecop.freeze do and end, then time is unfrozen at the end of the block

PR #1750, opened 2023-07-05
CLDC-2318 Update duplicate logs errors

kosiakkatrina:
- Don't trigger location code error on general needs logs
- Don't trigger postcode error on supported housing logs
- Add the “tenancy code” field to the duplicate log check

### DIFF
--- a/app/services/bulk_upload/lettings/year2022/row_parser.rb
+++ b/app/services/bulk_upload/lettings/year2022/row_parser.rb
@@ -450,7 +450,7 @@ class BulkUpload::Lettings::Year2022::RowParser
       "field_96",  # startdate
       "field_97",  # startdate
       "field_98",  # startdate
-      bulk_upload.needstype != 2 ? "field_100" : nil, # propcode
+      "field_100", # propcode
       bulk_upload.needstype != 2 ? "field_108" : nil, # postcode
       bulk_upload.needstype != 2 ? "field_109" : nil, # postcode
       "field_111", # owning org
--- a/app/services/bulk_upload/lettings/year2022/row_parser.rb
+++ b/app/services/bulk_upload/lettings/year2022/row_parser.rb
@@ -442,7 +442,7 @@ class BulkUpload::Lettings::Year2022::RowParser
 
   def spreadsheet_duplicate_hash
     attributes.slice(
-      "field_5",   # location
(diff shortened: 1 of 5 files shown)


app/services/bulk_upload/lettings/year2022/row_parser.rb
```diff
@@ -862,8 +863,8 @@ def validate_if_log_already_exists
       errors.add(:field_97, error_message) # startdate
       errors.add(:field_98, error_message) # startdate
       errors.add(:field_100, error_message) # propcode
-      errors.add(:field_108, error_message) # postcode_full
-      errors.add(:field_109, error_message) # postcode_full
+      errors.add(:field_108, error_message) unless bulk_upload.needstype == 2  # postcode_full
```
Airk0n: I would say that seeing as earlier on in the code you were using 
`bulk_upload.needstype != 2` I would expect this to be an `if` and not `unless` either way works but I would want them to be consistent

app/services/bulk_upload/lettings/year2023/row_parser.rb
```diff
@@ -853,11 +855,12 @@ def validate_if_log_already_exists
...
       errors.add(:field_8, error_message) # startdate
       errors.add(:field_9, error_message) # startdate
+      errors.add(:field_13, error_message) # tenancycode
       errors.add(:field_14, error_message) # propcode
-      errors.add(:field_17, error_message) # location
-      errors.add(:field_23, error_message) # postcode_full
-      errors.add(:field_24, error_message) # postcode_full
-      errors.add(:field_25, error_message) # la
+      errors.add(:field_17, error_message) unless field_4 == 1 # location
+      errors.add(:field_23, error_message) unless field_4 == 2 # postcode_full
```
Airk0n: Yeah same again, using != and unless when talking about the same thing. 
I don't know, maybe this is done elsewhere and others are ok with it, but for me I might be tempted to double check the logic when working on this in the future as I'm seeing `==`and `!=` very close together.

Airk0n (approved):
Looks good to me :)

app/services/bulk_upload/lettings/year2022/row_parser.rb
kosiakkatrina: Yup, sure! Will change!

PR #1752, opened 2023-07-05
CLDC-2495 Add delete duplicate logs page

kosiakkatrina:
Add delete duplicates page, which allows to confirm the deletion of listed duplicate logs: 
image
List the deleted logs in the success banner:
image

- link to this page from duplicate logs page "Keep this log and delete duplicates” buttons
- When "Delete these logs" is clicked direct the user to the duplicate logs page showing the details of the log they chose to keep
image


app/models/lettings_log.rb
```diff
@@ -53,6 +53,8 @@ class LettingsLog < Log
   scope :filter_by_before_startdate, ->(date) { where("lettings_logs.startdate >= ?", date) }
   scope :unresolved, -> { where(unresolved: true) }
   scope :filter_by_organisation, ->(org, _user = nil) { where(owning_organisation: org).or(where(managing_organisation: org)) }
+  scope :duplicate_logs, ->(log) { where(tenancycode: log.tenancycode, propcode: log.propcode, startdate: log.startdate).where.not(id: log.id) } # need to update to check all the fields
+  scope :duplicate_logs_for_organisation, ->(org, log) { filter_by_organisation(org).duplicate_logs(log) }
```
arfacamble: as it stands, this scope could be lifted to `log.rb` if we wanted

app/policies/lettings_log_policy.rb
```diff
@@ -21,4 +21,8 @@ def destroy?
     # Data providers can only delete the log if it is assigned to them
     log.created_by == user
   end
+
+  def delete_duplicates?
+    user.support? || log.owning_organisation == user.organisation || log.managing_organisation == user.organisation
+  end
```
arfacamble: 👍

app/views/logs/delete_duplicates.html.erb
```diff
@@ -0,0 +1,39 @@
...
+    <%= govuk_warning_text(text: "You will not be able to undo this action.") %>
+
+    <p class="govuk-body govuk-!-margin-bottom-2">
+      <%= @duplicate_logs.count == 1 ? "This log" : "These logs" %> will be deleted:
+    </p>
+    <ul class="govuk-list">
+      <% @duplicate_logs.each do |duplicate_log| %>
+        <li>
+          <strong>
+            <%= govuk_link_to "Log #{duplicate_log.id}", duplicate_log.lettings? ? lettings_log_path(duplicate_log) : sales_log_path(duplicate_log) %>
```
arfacamble:
```suggestion
            <%= govuk_link_to "Log #{duplicate_log.id}", url_for(duplicate_log) %>
```
I think you can do this

app/views/logs/delete_duplicates.html.erb
```diff
@@ -0,0 +1,39 @@
...
+          </strong>
+        </li>
+      <% end %>
+    </ul>
+
+    <div class="govuk-button-group">
+      <%= govuk_button_to @duplicate_logs.count == 1 ? "Delete this log" : "Delete these logs",
```
arfacamble, line 29: Does this work?
```suggestion
      <%= govuk_button_to "Delete #{"this".pluralize(@duplicate_logs.count)} #{"log".pluralize(@duplicate_logs.count)}",
```
or maybe even with "this log".pluralize
I suspect the first works and the second doesn't, first is them similarly verbose so maybe not better

config/locales/en.yml
```diff
@@ -182,6 +182,9 @@ en:
     logs_deleted:
       one: "%{count} log has been deleted"
       other: "%{count} logs have been deleted"
+    duplicate_logs_deleted:
+      one: "%{log_ids} has been deleted"
+      other: "%{log_ids} have been deleted"
```
arfacamble: v nice.

spec/features/lettings_log_spec.rb
```diff
@@ -413,5 +413,24 @@
...
+
+      it "is possible to delete duplicate logs" do
+        visit lettings_log_delete_duplicates_path(lettings_log)
+        click_button "Delete this log"
+        duplicate_log.reload
+        expect(duplicate_log.status).to eq("deleted")
+        expect(page).to have_css(".govuk-notification-banner.govuk-notification-banner--success")
+        expect(page).to have_content("Log #{duplicate_log.id} has been deleted")
+      end
+    end
```
arfacamble, line 451: is the plan to add to this feature test to encompass the earlier steps that are part of Aaron's ticket once that's done?

arfacamble (approved):

app/views/logs/delete_duplicates.html.erb
kosiakkatrina: Doesn't work unfortunately

spec/features/lettings_log_spec.rb
kosiakkatrina: Hopefully!

app/controllers/delete_logs_controller.rb
```diff
@@ -193,4 +199,16 @@ def discard(logs)
...
+  def duplicate_log_ids(logs)
+    logs.map { |log| "Log #{log.id}" }.join(", ")
+  end
+
+  def original_log_id
+    CGI.parse(URI.parse(request.referer).query)["original_log_id"][0]
+  end
+
+  def remaining_log_id
+    request.referer.split("/")[-2]
```
arfacamble: Rather than getting these things out of the url of the last place we came from with fiddly methods, why don't we just pass them in the params?
since this is a delete action I assume the route is an http DELETE so it should be happy to have a body.

app/controllers/delete_logs_controller.rb
```diff
@@ -52,8 +55,11 @@ def delete_sales_logs_confirmation
   def discard_sales_logs
     logs = SalesLog.find(params.require(:ids))
     discard logs
-
-    redirect_to sales_logs_path, notice: I18n.t("notification.logs_deleted", count: logs.count)
+    if request.referer&.include?("delete-duplicates")
+      redirect_to sales_log_duplicate_logs_path(sales_log_id: remaining_log_id, original_log_id:), notice: I18n.t("notification.duplicate_logs_deleted", count: logs.count, log_ids: duplicate_log_ids(logs))
```
arfacamble: is this a placeholder or the intention? We are returned back to the page that shows the duplicates? What happens there given we have just deleted all the duplicates (other than that success banner)

app/controllers/duplicate_logs_controller.rb
```diff
@@ -47,4 +60,11 @@ def duplicate_check_question_ids
       %w[owning_organisation_id saledate purchid age1 sex1 ecstat1 postcode_full]
     end
   end
+
+  def find_original_log_id
+    return unless @log
+
+    query_params = URI.parse(request.url).query
+    @original_log_id = query_params.present? ? CGI.parse(query_params)["original_log_id"][0]&.to_i : @log.id
```
arfacamble: this seems fiddly. What is the situation where the original log id is in the query rather than in the path?

arfacamble: I also can't see where this original log id is used, so I imagine I'm missing context on what it's for

app/controllers/duplicate_logs_controller.rb
```diff
@@ -47,4 +60,11 @@ def duplicate_check_question_ids
       %w[owning_organisation_id saledate purchid age1 sex1 ecstat1 postcode_full]
     end
   end
```
arfacamble, line 62: I'm hoping that we find a way to to have the duplicate fields saved on the log in such a way that there is a single source of truth. I appreciate there is fiddlyness around certain fields only being taken into account under certain conditions, so it can't _only_ be a constant on the class.

I don't have a solution for this currently

app/controllers/form_controller.rb
```diff
@@ -146,6 +146,17 @@ def referrer_from_query
...
 
+  def original_duplicate_log_id_from_query
+    query_params = URI.parse(request.url).query
+
+    return unless query_params
+
+    parsed_params = CGI.parse(query_params)
+    return unless parsed_params["original_log_id"]
+
+    parsed_params["original_log_id"][0]
```
arfacamble: if you want to play code golf
```suggestion
    parsed_params["original_log_id"]&.first
```

app/helpers/duplicate_logs_helper.rb
```diff
@@ -0,0 +1,21 @@
+module DuplicateLogsHelper
+  include GovukLinkHelper
+
+  def duplicate_logs_continue_button(all_duplicates, duplicate_log, original_log_id)
+    if all_duplicates.count > 1
+      return govuk_button_link_to "Keep this log and delete duplicates", send("#{duplicate_log.class.name.underscore}_delete_duplicates_path", duplicate_log, original_log_id:)
+    end
```
arfacamble, line 12: I don't always find these easy to follow when we're building strings then calling them to get a url and extra arguments hanging off the end. James pointed me towards url_for before. I think here it would look something like:
```
url_for(
  controller: "duplicate_logs",
  action: "delete_duplicates",
  id: duplicate_log.id,
  original_log_id:
)
```

app/helpers/duplicate_logs_helper.rb
```diff
@@ -0,0 +1,21 @@
...
+
+    if original_log_id == duplicate_log.id
+      govuk_button_link_to "Back to Log #{duplicate_log.id}", send("#{duplicate_log.class.name.underscore}_path", duplicate_log)
+    else
+      return govuk_button_link_to "Back to lettings logs", lettings_logs_path if duplicate_log.lettings?
+
+      govuk_button_link_to "Back to sales logs", sales_logs_path
```
arfacamble:
```suggestion
  type = duplicate_log.lettings? "lettings" : "sales"

  govuk_button_link_to "Back to#{type} logs", url_for duplicate_log.class
```

spec/features/lettings_log_spec.rb
```diff
@@ -389,5 +389,70 @@
...
+      let!(:duplicate_log) do
+        duplicate = lettings_log.dup
+        duplicate.id = nil
+        duplicate.save!
+        duplicate
+      end
+
+      before do
+        lettings_log.update!(tenancycode: "different")
+        visit("/lettings-logs/#{lettings_log.id}/tenant-code")
```
arfacamble, line 399: I'm giong to keep chirping up that we could be implicitly testing oiur routing by using url helpers in feature and request specs

spec/features/sales_log_spec.rb
```diff
@@ -137,4 +137,72 @@
...
+
+  context "when a log becomes a duplicate" do
+    let(:user) { create(:user, :data_coordinator) }
+    let(:sales_log) { create(:sales_log, :completed, owning_organisation: user.organisation, created_by: user) }
+    let!(:duplicate_log) do
+      duplicate = sales_log.dup
+      duplicate.id = nil
+      duplicate.save!
+      duplicate
+    end
```
arfacamble: how come you're not just using your :duplicate traits in the factories here?

spec/features/sales_log_spec.rb
```diff
@@ -137,4 +137,72 @@
       end
     end
   end
+
+  context "when a log becomes a duplicate" do
+    let(:user) { create(:user, :data_coordinator) }
+    let(:sales_log) { create(:sales_log, :completed, owning_organisation: user.organisation, created_by: user) }
```
arfacamble: mini code golf, but owning organisation applied in factory as the org of the user, so created_by is sufficient here

spec/features/sales_log_spec.rb
```diff
@@ -137,4 +137,72 @@
...
+      sales_log.update!(purchid: "different")
+      visit("/sales-logs/#{sales_log.id}/purchaser-code")
+      fill_in("sales-log-purchid-field", with: duplicate_log.purchid)
+      click_button("Save and continue")
+    end
+
+    it "allows keeping the original log and deleting duplicates" do
+      expect(page).to have_current_path("/sales-logs/#{sales_log.id}/duplicate-logs")
+      click_link("Keep this log and delete duplicates", href: "/sales-logs/#{sales_log.id}/delete-duplicates?original_log_id=#{sales_log.id}")
+      expect(page).to have_current_path("/sales-logs/#{sales_log.id}/delete-duplicates?original_log_id=#{sales_log.id}")
```
arfacamble, line 158: I think this explains to me what the original_log_id is. Is this the id of the log we want to keep, and we want to hold onto that so we can redirect to it later? If so I'm not sure about the "original" bit, seems like "log_to_keep_id" might be more accurate?

spec/features/sales_log_spec.rb
```diff
@@ -137,4 +137,72 @@
...
+      click_button "Delete this log"
+      duplicate_log.reload
+      expect(duplicate_log.status).to eq("deleted")
```
arfacamble:
```suggestion
      expect(duplicate_log.status_deleted?).to be true
```

arfacamble: I've just been working on enums, so I didn't know about these dynamically generated methods before. We should be using them wherever possible I think.
It's also where eg the user.data_provider? method comes from

spec/features/sales_log_spec.rb
```diff
@@ -137,4 +137,72 @@
...
+      expect(page).to have_link("Back to Log #{sales_log.id}", href: "/sales-logs/#{sales_log.id}")
+    end
+
+    it "allows changing answer on remaining original log" do
+      click_link("Keep this log and delete duplicates", href: "/sales-logs/#{sales_log.id}/delete-duplicates?original_log_id=#{sales_log.id}")
+      click_button "Delete this log"
+      click_link("Change", href: "/sales-logs/#{sales_log.id}/purchaser-code?original_log_id=#{sales_log.id}&referrer=interruption_screen")
+      click_button("Save and continue")
+      expect(page).to have_current_path("/sales-logs/#{sales_log.id}/duplicate-logs?original_log_id=#{sales_log.id}")
+      expect(page).to have_link("Back to Log #{sales_log.id}", href: "/sales-logs/#{sales_log.id}")
```
arfacamble, line 176: at this point do we expect the page to display the other formerly duplicate log any more?

spec/features/sales_log_spec.rb
```diff
@@ -137,4 +137,72 @@
...
+      expect(page).to have_link("Back to sales logs", href: "/sales-logs")
+    end
+
+    it "allows changing answers on remaining duplicate log" do
+      click_link("Keep this log and delete duplicates", href: "/sales-logs/#{duplicate_log.id}/delete-duplicates?original_log_id=#{sales_log.id}")
+      click_button "Delete this log"
+      click_link("Change", href: "/sales-logs/#{duplicate_log.id}/purchaser-code?original_log_id=#{sales_log.id}&referrer=interruption_screen")
+      click_button("Save and continue")
+      expect(page).to have_current_path("/sales-logs/#{duplicate_log.id}/duplicate-logs?original_log_id=#{sales_log.id}")
+      expect(page).to have_link("Back to sales logs", href: "/sales-logs")
```
arfacamble, line 200: I'm surprised that if we have edited a value on another log, and not deleted anything, that we wouldn't want to return back into the flow of continuing to work on the log we were originally on

spec/requests/duplicate_logs_controller_spec.rb
```diff
@@ -2,111 +2,296 @@
...
-        :completed,
-        created_by: user,
-      )
+  let(:user) { create(:user, :data_coordinator) }
+  let(:lettings_log) do
+    create(
+      :lettings_log,
+      :completed,
+      created_by: user,
+      owning_organisation: user.organisation,
```
arfacamble:
```suggestion
      created_by: user,
```
as above, this is sufficient

spec/requests/duplicate_logs_controller_spec.rb
```diff
@@ -2,111 +2,296 @@
...
+    )
+  end
+  let(:sales_log) do
+    create(
+      :sales_log,
+      :completed,
+      created_by: user,
+      owning_organisation: user.organisation,
```
arfacamble: ditto

arfacamble (commented):
I'm not fully clear on what the routes are. As shown by my confusion on what original_log_id means

spec/features/sales_log_spec.rb
kosiakkatrina: That's a separate ticket. This only accounts for editing after duplicates have been deleted and we only have 1 log left

spec/features/sales_log_spec.rb
kosiakkatrina: Nope, just the one that the user chose to keep

app/controllers/delete_logs_controller.rb
kosiakkatrina: it's the intention, I've updated the last screenshot in PR description. We basically allow changing answers after a single log is kept

app/controllers/delete_logs_controller.rb
kosiakkatrina: Good point!

app/controllers/duplicate_logs_controller.rb
kosiakkatrina: So when we click on `keep this log and delete duplicates` the path would be `/lettings-logs/:id_of_the_log_we_keep/duplicate-logs?original_log_id=:id_of_the_original_log`

kosiakkatrina: And sometimes they're the same, but I don't think it makes sense to keep the original log id in the path if it's deleted, especially if we allow the log to be edited afterwards

arfacamble: what's the situation where we don't find it in the params and fall back on @log.id? I guess it's confusing me slightly that the original log identifier might not be kept in the query param as you have it above

spec/features/sales_log_spec.rb
kosiakkatrina: Ah good point! Wrote these tests before the trait

spec/features/sales_log_spec.rb
kosiakkatrina: Hm, so it's not necessarily the one we want to keep, it's the one which we started the journey from

app/controllers/duplicate_logs_controller.rb
kosiakkatrina: Ah, I see what you mean. 
So the very first time we land on the page it's on duplicate-logs page (`/lettings-logs/:original_log_id/duplicate-logs`), after we choose which log to keep we get to delete-logs page (`/lettings-logs/:id_of_the_log_we_keep/delete-duplicates?original_log_id=:id_of_the_original_log`) and after we confirm the deletion it goes back to the duplicate-logs page (` /lettings-logs/:id_of_the_log_we_keep/duplicate-logs?original_log_id=:id_of_the_original_log`). So at that point we're on duplicate logs page for the log that we kept, instead of the log we started on, because we deleted that one. We still want to know what it was tho, because we need to determine what onwards journey to choose, hence original_log_id.
Let me know if that makes it any clearer, happy to chat through it if easier

kosiakkatrina: Okay, so I've updated this slightly. We can pass that ID through from the very beginning, because we know which journey we're coming from. We don't default anymore so it shouldn't interfere with your ticket as much

app/views/logs/delete_duplicates.html.erb
baarkerlounger: You could add:

```ruby
ActiveSupport::Inflector.inflections(:en) do |inflect|
    inflect.irregular 'this', 'these'
end
```

to https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/blob/main/config/initializers/inflections.rb, then it would work

kosiakkatrina: Yes! Good point!
I'm not gonna do it tho, I don't think double interpolation is more readable in this case, but shout if you disagree

app/views/duplicate_logs/show.erb
arfacamble, line 1: I'm surprised this doesn't need to be named .html.erb
I wonder why we do that if it's not necessary

kosiakkatrina: Oh good catch 👀 I think it ends up being read the same under the hood so it wasn't failing. Renamed it now

config/routes.rb
```diff
@@ -171,6 +171,7 @@
   resources :lettings_logs, path: "/lettings-logs" do
     get "delete-confirmation", to: "lettings_logs#delete_confirmation"
     get "duplicate-logs", to: "duplicate_logs#show"
```
arfacamble, line 173: gentle preference for these to be just duplicates but it's obv not very important

app/helpers/duplicate_logs_helper.rb
arfacamble: woohooo

app/views/logs/delete_duplicates.html.erb
```diff
@@ -0,0 +1,39 @@
...
+    <%= govuk_warning_text(text: "You will not be able to undo this action.") %>
+
+    <p class="govuk-body govuk-!-margin-bottom-2">
+      <%= @duplicate_logs.count == 1 ? "This log" : "These logs" %> will be deleted:
+    </p>
+    <ul class="govuk-list">
+      <% @duplicate_logs.each do |duplicate_log| %>
+        <li>
+          <strong>
+            <%= govuk_link_to "Log #{duplicate_log.id}", url_for(duplicate_log) %>
```
arfacamble, line 22: 🎈

PR #1753, opened 2023-07-05
CLDC-2474 Allow numeric fields to accept decimal representations in bulk upload

natdeanlewissoftwire:
A user was seeing an error because they had a decimal value 3.00 in their csv for lettings type, and there's a validation that compares lettings type to regex that checks it is an integer (not a decimal). I've updated the regex to allow decimal representations of integers (optional .00000... at the end for n zeroes). This is _sort of_ user error but good to protect against this in future as the error message the user was seeing wasn't super helpful.

ticket: https://dluhcdigital.atlassian.net/browse/CLDC-2474


spec/requests/delete_logs_controller_spec.rb
natdeanlewissoftwire, line 1: unrelated linting

spec/services/bulk_upload/lettings/year2023/row_parser_spec.rb
```diff
@@ -334,6 +334,276 @@
         end
       end
 
+      context "when valid row with valid decimal (integer) field_5" do
+        before do
+          allow(FeatureToggle).to receive(:bulk_upload_duplicate_log_check_enabled?).and_return(true)
+        end
+
+        let(:attributes) do
```
kosiakkatrina: Would it make sense to extract this into :valid_attributes and only edit relevant fields in the tests? Similar to what we have in some other parser tests

kosiakkatrina (approved):

spec/services/bulk_upload/lettings/year2023/row_parser_spec.rb
natdeanlewissoftwire: yep good shout, will do :)

PR #1762, opened 2023-07-07
Add duplicate_logs scope

kosiakkatrina:
This PR is adding a scope to get all the duplicates for a given log and a scope to get all duplicates for a given log within the organisation. 
The rules for determining a duplicate log are the same as in bulk upload and require all the duplicate check fields to be completed (except for purchid and tenancy code) to be recognised as duplicates


app/models/sales_log.rb
```diff
@@ -41,6 +41,14 @@ class SalesLog < Log
       .or(filter_by_id(param))
   }
   scope :filter_by_organisation, ->(org, _user = nil) { where(owning_organisation: org) }
+  scope :duplicate_logs, lambda { |log|
+    attrs_to_check = %w[purchid saledate age1 sex1 ecstat1 postcode_full]
```
natdeanlewissoftwire: I prefer this indentation to the one in the equivalent block in lettings_log.rb, can you use that there too?

spec/models/lettings_log_spec.rb
```diff
@@ -2797,6 +2797,130 @@ def previous_postcode_fields
         expect(described_class.filter_by_user(%w[all yours], created_by_user).count).to eq(3)
       end
     end
+
```
natdeanlewissoftwire, line 2800: can you add some tests for `duplicate_logs_for_organisation` too? same in `sales_log_spec` :)

natdeanlewissoftwire (approved):
lgtm, 2 suggestions

app/models/lettings_log.rb
```diff
@@ -53,6 +53,15 @@ class LettingsLog < Log
...
   scope :unresolved, -> { where(unresolved: true) }
   scope :filter_by_organisation, ->(org, _user = nil) { where(owning_organisation: org).or(where(managing_organisation: org)) }
+  scope :duplicate_logs, lambda { |log|
+                           attrs_to_check = %w[tenancycode propcode startdate age1 sex1 ecstat1 tcharge postcode_full]
+                           where(log.slice(*attrs_to_check))
+                           .where.not(id: log.id)
+                           .where.not(status: "deleted")
+                           .where.not("startdate IS NULL OR age1 IS NULL OR sex1 IS NULL OR ecstat1 IS NULL OR tcharge IS NULL OR postcode_full IS NULL OR propcode IS NULL OR needstype IS NULL")
+                           .where("location_id = ? OR needstype = 1", log.location_id)
+                         }
```
arfacamble: since this is over many lines now I'm surprised rubocop doesn't enforce do end rather than {} wouldn't that be preferable as standard ruby for a multiline block that is not having something chained on the end of it?

I'm inclined to want to pull attrs_to_check into a constant on the LettingsLog class, hopefully there is a way then to access this list from the various places where it is used, eg bulk upload

app/models/lettings_log.rb
```diff
@@ -53,6 +53,15 @@ class LettingsLog < Log
...
+  scope :duplicate_logs_for_organisation, ->(org, log) { filter_by_organisation(org).duplicate_logs(log) }
```
arfacamble: I would be happy for you to dodge writing the tests that Nat is asking for by removing this maybe?
I think it would be perfectly readable wherever this is used to just chain the scopes that it combines anyway.

natdeanlewissoftwire: If there's a common use case for this I reckon we should keep it in, but if not happy for it to be taken out, up to you :)

app/models/lettings_log.rb
```diff
@@ -44,15 +44,24 @@ class LettingsLog < Log
...
+        .or(filter_by_id(param))
+  }
   scope :filter_by_before_startdate, ->(date) { where("lettings_logs.startdate >= ?", date) }
   scope :unresolved, -> { where(unresolved: true) }
   scope :filter_by_organisation, ->(org, _user = nil) { where(owning_organisation: org).or(where(managing_organisation: org)) }
+  scope :duplicate_logs, lambda { |log|
+    attrs_to_check = %w[tenancycode propcode startdate age1 sex1 ecstat1 tcharge postcode_full]
+    where(log.slice(*attrs_to_check))
+    .where.not(id: log.id)
+    .where.not(status: "deleted")
```
arfacamble: will we ever want to see logs that are not visible? Can we just replace this line with `.visible`?

app/models/lettings_log.rb
kosiakkatrina: You're right! We can just call `current_user.lettings_logs.duplicate_logs(@log)` that's way better! Thanks

app/models/lettings_log.rb
kosiakkatrina: This is what rubocop corrects it to, I think we need to use lambda cause it's multiline and this is the syntax for it.

Good shout, extracted the constants 👍

PR #1763, opened 2023-07-07
CLDC-2494 Add duplicate logs page

Airk0n:
Needs https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/1762 to correctly determine duplicates.

This PR adds a duplicate logs controller with the page that displays all the duplicates for a specific log.
image
image


app/controllers/duplicate_logs_controller.rb
```diff
@@ -0,0 +1,27 @@
+class DuplicateLogsController < ApplicationController
+  before_action :find_resource_by_named_id
+
+  def show
+    @duplicate_logs = @log.class.duplicate_logs_for_organisation(current_user.organisation_id, @log)
+    @all_duplicates = [@log, *@duplicate_logs]
+    @duplicate_check_questions = duplicate_check_question_ids.map { |question_id| @log.form.get_question(question_id, @log) }.compact
```
ghost: I feel all this + `duplicate_check_question_ids` could be in an helper or presenter model/module where we could unit test it

app/controllers/form_controller.rb
```diff
@@ -155,6 +155,10 @@ def previous_interruption_screen_referrer
   end
 
   def successful_redirect_path
+    if @log.class.duplicate_logs_for_organisation(current_user.organisation, @log).count.positive?
```
ghost: this change is not unit tested

kosiakkatrina: Hmm, we're testing that the redirect happens when there are duplicate logs in form_controller_spec, and the opposite condition would get tested by other successful redirect cases that expect some of the further code to get executed. I feel like that covers it (?)

app/controllers/duplicate_logs_controller.rb
kosiakkatrina: Hm, so the scopes are tested here, the one in this PR is just a placeholder https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/1762
And the rest is tested in request specs - we're checking that correct logs and correct questions get displayed

ghost (approved):
@kosiakkatrina thanks for replying to my comments, all good!

ghost (approved):

PR #1766, opened 2023-07-10
CLDC-2490 Tidy up DSA model

ghost:
- Remove new_data_protection_confirmation? flag
- remove unique index on dpo_id / org on DPC
- add new columns to persist org and DPO data on DPC
- update import to persist org and DPO data on DPC
- update org controller to persist org and DPO data when signing DPC
- add migration to persist org and DPO data on DPC

Next steps:
- run migration on staging and prod
- further code cleanup

### DIFF
--- a/app/controllers/organisations_controller.rb
+++ b/app/controllers/organisations_controller.rb
@@ -171,7 +171,7 @@ class OrganisationsController < ApplicationController
         confirmed: true,
         data_protection_officer: current_user,
         created_at: Time.zone.now,
-        signed_at_at: Time.zone.now,
+        signed_at: Time.zone.now,
         organisation_name: @organisation.name,
         organisation_address: @organisation.address_row,
         organisation_phone_number: @organisation.phone,
--- a/app/controllers/organisations_controller.rb
+++ b/app/controllers/organisations_controller.rb
@@ -159,13 +159,10 @@ class OrganisationsController < ApplicationController
   end
 
   def data_sharing_agreement
-    return render_not_found unless FeatureToggle.new_data_protection_confirmation?
-
     @data_protection_confirmation = current_user.organisation.data_protection_confirmation
   end
 
   def confirm_data_sharing_agreement
-    return render_not_found unless FeatureToggle.new_data_protection_confirmation?
     return render_not_found unless current_user.is_dpo?
     return render_not_found if @organisation.data_protection_confirmed?
 
--- a/app/controllers/organisations_controller.rb
+++ b/app/controllers/organisations_controller.rb
@@ -170,14 +170,25 @@ class OrganisationsController < ApplicationController
       @organisation.data_protection_confirmation.update!(
         confirmed: true,
         data_protection_officer: current_user,
-        # When it was signed
         created_at: Time.zone.now,
+        signed_at_at: Time.zone.now,
+        organisation_name: @organisation.name,
+        organisation_address: @organisation.address_row,
+        organisation_phone_number: @organisation.phone,
+        data_protection_officer_email: current_user.email,
+        data_protection_officer_name: current_user.name,
       )
     else
       DataProtectionConfirmation.create!(
-        organisation: current_user.organisation,
+        organisation: @organisation,
         confirmed: true,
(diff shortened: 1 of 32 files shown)


app/controllers/organisations_controller.rb
```diff
@@ -159,28 +159,36 @@ def merge_request
...
     return render_not_found unless current_user.is_dpo?
     return render_not_found if @organisation.data_protection_confirmed?
 
     if @organisation.data_protection_confirmation
       @organisation.data_protection_confirmation.update!(
         confirmed: true,
         data_protection_officer: current_user,
-        # When it was signed
         created_at: Time.zone.now,
+        signed_at_at: Time.zone.now,
```
kosiakkatrina: typo? `singed_at_at`

app/controllers/organisations_controller.rb
```diff
@@ -159,28 +159,36 @@ def merge_request
...
-    return render_not_found unless FeatureToggle.new_data_protection_confirmation?
     return render_not_found unless current_user.is_dpo?
     return render_not_found if @organisation.data_protection_confirmed?
 
     if @organisation.data_protection_confirmation
       @organisation.data_protection_confirmation.update!(
         confirmed: true,
         data_protection_officer: current_user,
-        # When it was signed
         created_at: Time.zone.now,
```
kosiakkatrina, line 173: Do we want to be updating created_at on update?

app/helpers/data_sharing_agreement_helper.rb
```diff
@@ -61,7 +61,11 @@ def present_section_12_2(data_protection_confirmation:, user:, organisation:)
       end
     end
 
-    "12.2. For #{@org_name}: Name: #{@dpo_name}, Postal Address: #{@org_address}, E-mail address: #{@dpo_email}, Telephone number: #{@org_phone}"
+    if data_protection_confirmation&.confirmed? && @dpo_email.exclude?("@") # Do not show invalid email addresses
+      "12.2. For #{@org_name}: Name: #{@dpo_name}, Postal Address: #{@org_address}, Telephone number: #{@org_phone}"
+    else
+      "12.2. For #{@org_name}: Name: #{@dpo_name}, Postal Address: #{@org_address}, E-mail address: #{@dpo_email}, Telephone number: #{@org_phone}"
+    end
```
kosiakkatrina, line 68: I'd flip this and use `.include?` instead, but that's just preference
Would it drop into `else` if `data_protection_confirmation` is not confirmed?

app/models/organisation.rb
```diff
@@ -112,13 +112,7 @@ def display_organisation_attributes
       { name: "Registration number", value: housing_registration_no || "", editable: false },
       { name: "Rent periods", value: rent_period_labels, editable: false, format: :bullet },
       { name: "Owns housing stock", value: holds_own_stock ? "Yes" : "No", editable: false },
-    ].compact
-
-    unless FeatureToggle.new_data_protection_confirmation?
-      attrs << { name: "Data protection agreement", value: data_protection_agreement_string, editable: false }
```
kosiakkatrina, line 118: Do we not want to display this at all?

app/controllers/organisations_controller.rb
ghost: I will remove this after having run the migration so keep things in sync

app/models/organisation.rb
ghost: this is the old way to display it, now this is done with a helper

app/controllers/organisations_controller.rb
ghost: yes! will fix

app/helpers/data_sharing_agreement_helper.rb
ghost: I think then i'd have to invert `data_protection_confirmation&.confirmed?` and that'll look as bad or worse than `.exclude`

kosiakkatrina: Sure!

kosiakkatrina (approved):

PR #1768, opened 2023-07-11
CLDC-2241 Add filter count and clear button

natdeanlewissoftwire:
Adds a count of applied filters and a clear link if this number is > 0. These are separated by filter type so lettings and sales can have separate sets of filters - clearing one set doesn't affect the other.
image

ticket: https://dluhcdigital.atlassian.net/browse/CLDC-2241


kosiakkatrina (approved):

kosiakkatrina (commented):
How would it work with the bulk upload filter?

natdeanlewissoftwire:
> How would it work with the bulk upload filter?

Same way - I've added the bulk_upload_id filter to the count function too, good idea!

PR #1770, opened 2023-07-12
CLDC-2483 Add rake task to remove placeholder deactivation periods

kosiakkatrina:
Some imported schemes/locations have end dates that are in distant future and likely not representative of an actual end date. We have imported them as deactivation dates in the past, but we would no longer be importing any end dates that are after 2031/1/1

This PR adds rake tasks to remove deactivation periods later than 31/1/1 that have been imported in the past.


RachaelBooth (approved):
LGTM

I've just been thinking through what happens if something goes wrong during the task, but I think it's fine (i.e. we see an error and we can safely fix + re-run).

PR #1775, opened 2023-07-13
CLDC-2242 Add assigned to filter

natdeanlewissoftwire:
Adds an assigned to filter (like the previous user filter, but with the option to select a specific user as well as all/you)

image

ticket: https://dluhcdigital.atlassian.net/browse/CLDC-2242


app/models/user.rb
```diff
@@ -144,9 +144,9 @@ def assignable_roles
 
   def logs_filters(specific_org: false)
     if (support? && !specific_org) || organisation.has_managing_agents?
-      %w[status years user organisation bulk_upload_id]
+      %w[status years assigned_to user organisation bulk_upload_id]
```
kosiakkatrina: do we still need user here?

app/helpers/filters_helper.rb
```diff
@@ -42,6 +43,11 @@ def organisations_filter_options(user)
     [OpenStruct.new(id: "", name: "Select an option")] + organisation_options.map { |org| OpenStruct.new(id: org.id, name: org.name) }
   end
 
+  def assigned_to_filter_options(user)
+    user_options = user.organisation.users
+    [OpenStruct.new(id: "", name: "Select an option")] + user_options.map { |user_option| OpenStruct.new(id: user_option.id, name: user_option.name) }
+  end
+
```
kosiakkatrina: Do we want this to show all the users for support user?

app/models/user.rb
natdeanlewissoftwire: Yep - just to distinguish between the "you" selection or the "specific user" selection with the current user selected (both would have the same value for user, but are distinguishable when we also include the assigned_to value

app/helpers/filters_helper.rb
natdeanlewissoftwire: Hm good point - I imagine so but I'll check and update the ACs to include that if so :)

natdeanlewissoftwire: have made this change, just waiting to hear back re whether we should also include non-support users' managing agents'/stock owners' users

kosiakkatrina (approved):

app/helpers/filters_helper.rb
natdeanlewissoftwire: we should, and now we do ^

PR #1776, opened 2023-07-13
CLDC-2497 Review duplicate logs from logs index page

arfacamble:
https://dluhcdigital.atlassian.net/browse/CLDC-2497
[JIRA ticket](https://dluhcdigital.atlassian.net/browse/CLDC-2497)

|image
|-
|image
|-


app/controllers/lettings_logs_controller.rb
```diff
@@ -11,20 +13,20 @@ class LettingsLogsController < LogsController
   before_action :redirect_if_bulk_upload_resolved, only: [:index]
 
   def index
-    respond_to do |format|
-      format.html do
```
arfacamble, line 15: I got rid of this format stuff for this method because we never do any other format than the default. Someone tell me if any of that is not quite true.

app/models/log.rb
```diff
@@ -85,6 +86,10 @@ def recalculate_start_year!
     collection_start_year
   end
 
+  def setup_completed?
+    form.setup_sections.all? { |sections| sections.subsections.all? { |subsection| subsection.status(self) == :completed } }
+  end
+
```
arfacamble, line 87: Is there any reason why I shouldn't have lifted this from sales log up to log?

app/helpers/duplicate_logs_helper.rb
```diff
@@ -0,0 +1,37 @@
...
+    duplicate_sets = { lettings: {}, sales: {} }
+    lettings_count = 0
+    sales_count = 0
+    duplicate_lettings_ids = Set.new
+    duplicate_sales_ids = Set.new
+
+    user.lettings_logs(created_by: true).each do |log|
+      next if duplicate_lettings_ids.include? log.id
+
+      duplicates = LettingsLog.filter_by_organisation(user.organisation).duplicate_logs(log)
```
kosiakkatrina: `user.lettings_logs` would return all the logs available for this user, might be safer if we ever change the scope of what's available. Also might be worth adding `.visible` to avoid listing all the deleted logs

app/models/log.rb
kosiakkatrina: I think it's fine in log, we only use it for sales logs so far

app/controllers/lettings_logs_controller.rb
kosiakkatrina: I guess there was also `format.csv` in the past, so it's probably fine to remove now

app/controllers/duplicate_logs_controller.rb
```diff
@@ -47,4 +59,11 @@ def duplicate_check_question_ids
       %w[owning_organisation_id saledate purchid age1 sex1 ecstat1 postcode_full]
     end
   end
+
+  def authenticate_scope!
+    case action_name
+    when :index
+      render_not_found unless current_user.data_provider?
+    end
+  end
```
kosiakkatrina: Would this make sense as a policy?

app/helpers/duplicate_logs_helper.rb
```diff
@@ -0,0 +1,37 @@
+module DuplicateLogsHelper
+  def duplicates_for_user(user)
+    duplicate_sets = { lettings: {}, sales: {} }
+    lettings_count = 0
+    sales_count = 0
+    duplicate_lettings_ids = Set.new
+    duplicate_sales_ids = Set.new
+
+    user.lettings_logs(created_by: true).each do |log|
```
kosiakkatrina: Probably makes sense to have .visible here too

app/models/user.rb
```diff
@@ -56,16 +56,20 @@ class User < ApplicationRecord
   scope :search_by, ->(param) { search_by_name(param).or(search_by_email(param)) }
   scope :sorted_by_organisation_and_role, -> { joins(:organisation).order("organisations.name", role: :desc, name: :asc) }
 
-  def lettings_logs
-    if support?
+  def lettings_logs(created_by: false)
+    if created_by
+      LettingsLog.created_by(self)
+    elsif support?
```
kosiakkatrina: This wasn't immediately clear, especially as it's called with (created_by: true) my first thought was that the logs had created_by not nil.
Would it make sense to use `.filter_by_user` (or `created_by`) scope where needed instead of changing this method?

app/helpers/duplicate_logs_helper.rb
```diff
@@ -0,0 +1,37 @@
+module DuplicateLogsHelper
+  def duplicates_for_user(user)
+    duplicate_sets = { lettings: {}, sales: {} }
+    lettings_count = 0
```
kosiakkatrina: Is this count of sets? Is it possible to count it at the end?

app/models/log.rb
```diff
@@ -29,6 +29,7 @@ class Log < ApplicationRecord
   scope :visible, -> { where(status: %w[not_started in_progress completed]) }
   scope :exportable, -> { where(status: %w[not_started in_progress completed deleted]) }
 
+  scope :created_by, ->(user) { where(created_by: user) }
```
kosiakkatrina: There's already a created_by scope

app/models/user.rb
kosiakkatrina: Also, wouldn't we still want to filter by organisation? In case we allow users to change org in the future or something

spec/requests/lettings_logs_controller_spec.rb
```diff
@@ -308,6 +306,13 @@
           expect(page).to have_link("Download (CSV, codes only)", href: "/lettings-logs/csv-download?codes_only=true")
         end
 
+        it "does not show a notification banner even if there are duplicate logs for this user" do
+          allow_any_instance_of(DuplicateLogsHelper).to receive(:duplicates_for_user).and_return({ lettings: [[1, 2]], sales: [] }) # rubocop:disable RSpec/AnyInstance
```
kosiakkatrina: We could use :duplicate trait, mock the scope instead and test a little bit more of the flow

app/helpers/duplicate_logs_helper.rb
arfacamble: have changed the first bit. duplicate logs scope has visible in it so I won't repeat that.

app/controllers/duplicate_logs_controller.rb
arfacamble: my mental model so far has been that we continue using authenticate scope for very basic conditions/plain get requests and pundit is more for authorisation to do stuff to the datas. But I feel like we don't have any strict rules for consistency.

app/models/user.rb
arfacamble: I've reverted this related to your first comment. I have asked Jemma about your second comment.

app/helpers/duplicate_logs_helper.rb
arfacamble: no. this bit is very slightyl awkward, but I am making it possible to pass the object returned by this method through in the params. rails does not handle multi-dimensional arrays, but it will happily handle nested hashes. So the object returned looks like this:
```
{
  lettings: {
    '0' => [3, 5],
    '1' => [6, 23],
  }
}
```
rather than
```
{
  lettings: [ [3, 5], [6, 23] ]
}
```

kosiakkatrina: Ahh, I see. Would maybe index instead of count make sense then?

app/helpers/duplicate_logs_helper.rb
kosiakkatrina: Hm, I think it was showing me deleted logs as well. Does it only check if the duplicates are visible, or the log that gets passed is as well?

app/models/user.rb
kosiakkatrina: This still takes in created_by, am I missing something?

spec/requests/lettings_logs_controller_spec.rb
arfacamble: I'm not sure what you're suggesting here.
This is for a support user, who never sees the banner.

kosiakkatrina: Right, well maybe in the coordinator context then? Conscious that requests specs are probably more suited for integration testing and I'm not sure we're doing any integration testing for this feature

app/models/lettings_log.rb
```diff
@@ -45,31 +45,58 @@ class LettingsLog < Log
...
+      .or(filter_by_id(param))
   }
   scope :filter_by_before_startdate, ->(date) { where("lettings_logs.startdate >= ?", date) }
   scope :unresolved, -> { where(unresolved: true) }
-
   scope :filter_by_organisation, ->(org, _user = nil) { where(owning_organisation: org).or(where(managing_organisation: org)) }
   scope :filter_by_owning_organisation, ->(owning_organisation, _user = nil) { where(owning_organisation:) }
   scope :filter_by_managing_organisation, ->(managing_organisation, _user = nil) { where(managing_organisation:) }
-
+  scope :age1_answered, -> { where.not(age1: nil).or(where(age1_known: 1)) }
```
Airk0n, line 58: Could this be done via a helper? or by looking to another class? if feels wrong having it hard coded on the model, though I can't articulate why. Is this code copied and pasted anywhere else that could also depend on a helper instead?.

app/models/sales_log.rb
```diff
@@ -42,14 +42,32 @@ class SalesLog < Log
   }
   scope :filter_by_organisation, ->(owning_organisation, _user = nil) { where(owning_organisation:) }
   scope :filter_by_owning_organisation, ->(owning_organisation, _user = nil) { where(owning_organisation:) }
+  scope :age1_answered, -> { where.not(age1: nil).or(where(age1_known: [1, 2])) }
```
Airk0n, line 45: Similar point to the lettings log, is it avoidable having hard coded specific questions answered tied to the model? I think it's about levels of abstraction and that a sales log model shouldn't be concerned with specific questions and if they're answered.

app/models/lettings_log.rb
natdeanlewissoftwire: I agree it does seem a bit hard codey, but I'm struggling to think of a way we could make it more dynamic (like a general `question_answered(id:)` method) and still be useful in other cases given there's variation between different fields and their `_known` fields' values for yes and no - if it's not something we need to generalise I think this is ok.

kosiakkatrina: I think for me this is mostly there so that the query runs faster, and it's separated out so that it's a bit more readable

Airk0n (approved):
LGTM

PR #1777, opened 2023-07-14
CLDC-2391 Update content on bulk upload pages

kosiakkatrina:
Add new content to say number of logs uploaded, number of logs with errors, and number of errors in total on fix choice page, confirmation page and the upload banner
Add new content in the details under "how to choose between fixing errors on the CORE site or in the CSV" on fix choice page
Change title on confirmation page

Store total_logs_count on bulk_upload, because bulk_upload.logs.count would return 0 before the logs get created

### DIFF
--- a/app/services/bulk_upload/lettings/validator.rb
+++ b/app/services/bulk_upload/lettings/validator.rb
@@ -73,6 +73,10 @@ class BulkUpload::Lettings::Validator
     row_parsers.any?(&:log_already_exists?)
   end
 
+  def total_logs_count
+    csv_parser.body_rows.count
+  end
+
 private
 
   # n^2 algo
--- a/app/helpers/logs_helper.rb
+++ b/app/helpers/logs_helper.rb
@@ -41,9 +41,15 @@ module LogsHelper
     end
   end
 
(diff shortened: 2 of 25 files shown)


app/services/bulk_upload/lettings/validator.rb
```diff
@@ -73,6 +73,10 @@ def any_logs_already_exist?
     row_parsers.any?(&:log_already_exists?)
   end
 
+  def total_logs_count
```
natdeanlewissoftwire, line 76: seems odd to me to have this in the validator - is there a better place for it?

natdeanlewissoftwire (approved):
Looks good, seems like a few places where we /could/ share more copy especially between lettings and sales rather than duplicating the same content, but if it's more effor than that's worth then I'm happy :)

app/services/bulk_upload/lettings/validator.rb
kosiakkatrina: Yeah, I'm not sure, I think I put it there cause we need the csv parser to get the count and it currently lives in the validator. We also want the update to the db to happen when we are processing the logs and that's why we call it where we do. If you have any thoughts on where this could be moved tho, let me know

PR #1785, opened 2023-07-18
CLDC-2247 Add locations filter

kosiakkatrina:
Add status filter for locations:
image

### DIFF
--- a/app/helpers/filters_helper.rb
+++ b/app/helpers/filters_helper.rb
@@ -56,6 +56,17 @@ module FiltersHelper
     }.freeze
   end
 
+  def location_status_filters
+    {
+      "incomplete" => "Incomplete",
+      "active" => "Active",
+      "deactivating_soon" => "Deactivating soon",
+      "activating_soon" => "Activating soon",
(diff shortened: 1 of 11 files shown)


app/helpers/filters_helper.rb
```diff
@@ -56,6 +56,17 @@ def scheme_status_filters
     }.freeze
   end
 
+  def location_status_filters
```
Airk0n, line 59: I've not seen this before, is this shorthand for a switch statement? 
If it's doing what I think it is then I would assume there's a helper function that takes a string, removes replaces the underscores with spaces and capitalises the first letter?

app/models/location.rb
Airk0n, line 1: Looks good

Airk0n (approved):
Looks good to me

app/helpers/filters_helper.rb
kosiakkatrina: Do you mean the hash? https://ruby-doc.org/core-2.7.0/Hash.html
We're just accessing values using keys, similarly like you'd access values in arrays using indexes

PR #1787, opened 2023-07-19
CLDC-2572 Add merged status

natdeanlewissoftwire:
Adds a merge_date to the orgs table that is used to define status (at the moment just active or merged - I've assumed the active label and merged labels are the same as where they appear in designs for other sections, this will get picked up at PO review if it's wrong).

Adds this status label to the about org and org list pages and stops users being added to already merged orgs via a validation.

Status label on the org list page:
image

ticket: https://dluhcdigital.atlassian.net/browse/CLDC-2572


spec/models/user_spec.rb
```diff
@@ -334,6 +334,16 @@
           .to raise_error(ActiveRecord::RecordInvalid, error_message)
       end
     end
+
+    context "when a user is added to a merged organisation" do
+      let(:merged_organisation) { create(:organisation, merge_date: Time.zone.yesterday) }
+      let(:error_message) { "Validation failed: Organisation #{I18n.t('validations.organisation.merged')}" }
+
+      it "validates email format" do
```
kosiakkatrina: does this test name need updating?

kosiakkatrina (approved):

db/schema.rb
```diff
@@ -192,14 +192,14 @@
     t.integer "hb"
     t.integer "hbrentshortfall"
     t.integer "property_relet"
-    t.datetime "mrcdate", precision: nil
+    t.datetime "mrcdate"
```
kosiakkatrina, line 195: Any reason we're removing these?

natdeanlewissoftwire: This happens automatically when I run migrations - I don't think this change actually makes a difference, I assume it's some sort of linting but I'm not actually certain why it's happening.

PR #1789, opened 2023-07-19
CLDC-2571 Add absorbed org relationship

natdeanlewissoftwire:
Adds an absorbing_organisation_id to orgs so they can be marked with the org that absorbed them during a merge. As discussed with the team in refinement, we don't need any more complex relationship details so this is enough. Adds the relevant relations to orgs, and includes these absorbed orgs in org filters and the user's log lists.

ticket: https://dluhcdigital.atlassian.net/browse/CLDC-2571


db/schema.rb
```diff
@@ -192,14 +192,14 @@
     t.integer "hb"
     t.integer "hbrentshortfall"
     t.integer "property_relet"
-    t.datetime "mrcdate", precision: nil
+    t.datetime "mrcdate"
```
natdeanlewissoftwire: not sure why these are being removed but it doesn't make a difference afaik, default precision either way

app/models/form/lettings/questions/stock_owner.rb
```diff
@@ -23,13 +23,13 @@ def answer_options(log = nil, user = nil)
...
 
-    stock_owners_answer_options = if user.support?
-                                    Organisation.where(holds_own_stock: true)
-                                  else
-                                    user.organisation.stock_owners
-                                  end.pluck(:id, :name).to_h
+    user_answer_options = if user.support?
+                            Organisation.where(holds_own_stock: true)
+                          else
+                            user.organisation.stock_owners + user.organisation.absorbed_organisations
```
kosiakkatrina: What would happen if absorbed organisation doesn't hold own stock?

app/models/organisation.rb
```diff
@@ -1,9 +1,6 @@
 class Organisation < ApplicationRecord
   has_many :users, dependent: :delete_all
   has_many :data_protection_officers, -> { where(is_dpo: true) }, class_name: "User"
-  has_many :owned_lettings_logs, class_name: "LettingsLog", foreign_key: "owning_organisation_id", dependent: :delete_all
-  has_many :managed_lettings_logs, class_name: "LettingsLog", foreign_key: "managing_organisation_id"
-  has_many :owned_sales_logs, class_name: "SalesLog", foreign_key: "owning_organisation_id", dependent: :delete_all
```
kosiakkatrina, line 6: Are we not using this anywhere?

kosiakkatrina: nevermind!

app/helpers/filters_helper.rb
```diff
@@ -45,12 +45,12 @@ def selected_option(filter, filter_type)
   end
 
   def owning_organisations_filter_options(user)
```
kosiakkatrina: Could we update/add tests for these helper methods?

app/models/form/lettings/questions/stock_owner.rb
natdeanlewissoftwire: hm, good point - I'll check for that too and flag at PO, think that's the desired behaviour

app/helpers/filters_helper.rb
natdeanlewissoftwire: good idea, have done :)

kosiakkatrina (approved):

PR #1792, opened 2023-07-20
Update confirmation link timeout to 24 hours

RachaelBooth:
Discussed needing this, particularly around the larger migrations - no ticket because it would take more time to write that than to make this PR.


RachaelBooth:
N.B. We'll need to co-ordinate this with updating the email templates.

ghost (approved):

PR #1793, opened 2023-07-20
CLDC-2496 Correct duplicate data

kosiakkatrina:
Update the routing to allow correcting duplicate logs from this page.
image
After a log is corrected and is no longer a duplicate update the routing to go to the remaining duplicates with the updated log removed. 
image

Allow returning to the initial log that was interrupted after the logs have ben deduplicated. (Back to log button)
image

### DIFF
--- a/app/controllers/form_controller.rb
+++ b/app/controllers/form_controller.rb
@@ -48,7 +48,7 @@ class FormController < ApplicationController
   def show_page
     if request.params["referrer"] == "interruption_screen" && request.headers["HTTP_REFERER"].present?
       @interruption_page_id = URI.parse(request.headers["HTTP_REFERER"]).path.split("/").last.underscore
-      @interruption_page_referrer_type = referrer_from_query
+      @interruption_page_referrer_type = from_referrer_query("referrer")
     end
 
     if @log
@@ -130,10 +130,10 @@ private
   end
 
   def is_referrer_type?(referrer_type)
-    referrer_from_query == referrer_type
+    from_referrer_query("referrer") == referrer_type
   end
 
-  def referrer_from_query
+  def from_referrer_query(query_param)
     referrer = request.headers["HTTP_REFERER"]
     return unless referrer
 
@@ -141,7 +141,7 @@ private
     return unless query_params
 
     parsed_params = CGI.parse(query_params)
-    parsed_params["referrer"]&.first
+    parsed_params[query_param]&.first
   end
 
   def original_duplicate_log_id_from_query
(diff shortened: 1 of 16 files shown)


app/controllers/form_controller.rb
```diff
@@ -163,11 +163,15 @@ def previous_interruption_screen_referrer
 
   def successful_redirect_path
     if FeatureToggle.deduplication_flow_enabled?
-      if @log.lettings?
-        if current_user.lettings_logs.duplicate_logs(@log).count.positive?
-          return send("lettings_log_duplicate_logs_path", @log, original_log_id: @log.id)
-        end
-      elsif current_user.sales_logs.duplicate_logs(@log).count.positive?
+      if is_referrer_type?("duplicate_logs")
```
natdeanlewissoftwire, line 166: small thing, these three ifs feel like they should be an if..elsif to me

spec/requests/sales_logs_controller_spec.rb
```diff
@@ -352,14 +352,14 @@
...
-              let!(:excluded_log) { create(:sales_log, :in_progress, owning_organisation: organisation) }
+              let!(:included_log) { create(:sales_log, :in_progress, purchid: "included_id", bulk_upload:, owning_organisation: organisation) }
+              let!(:excluded_log) { create(:sales_log, :in_progress, purchid: "excluded_id", owning_organisation: organisation) }
 
               it "returns logs only associated with the bulk upload" do
                 get "/sales-logs?bulk_upload_id[]=#{bulk_upload.id}"
 
-                expect(page).to have_content(included_log.id)
-                expect(page).not_to have_content(excluded_log.id)
+                expect(page).to have_content(included_log.purchid)
```
Airk0n, line 367: Can you just confirm that `purchid` is really what you want here? 
My understanding is that the context (and what the test was doing before) are looking at the log.id and the `purchid` is like an optional question answer on a log and not used as a primary key or anything like that.

Airk0n (changes requested):
Looks good, will approve after confirming that `purchase id`. or `log.id` on the test.

spec/requests/sales_logs_controller_spec.rb
kosiakkatrina: Ah, this is just a flaky test. So checking for log id is correct, but often fails because id can be any integer, and ofter it's being found somewhere else where we don't expect it. The purpose of this test is to check that a specific log appears or does not appear in the list, so checking the purchid should be enough to confirm that

app/helpers/duplicate_logs_helper.rb
```diff
@@ -22,4 +22,9 @@ def duplicate_logs_continue_button(all_duplicates, duplicate_log, original_log_i
   def duplicate_logs_action_href(log, page_id, original_log_id)
     send("#{log.model_name.param_key}_#{page_id}_path", log, referrer: "interruption_screen", original_log_id:)
   end
+
+  def change_duplicate_logs_action_href(log, page_id, all_duplicates, original_log_id)
+    remaining_duplicate_id = all_duplicates.map(&:id).reject { |id| id == log.id }.first
```
natdeanlewissoftwire: would renaming this to `first_remaining_duplicate_id` be clearer if there can be multiple remaining at this stage?

app/views/duplicate_logs/_duplicate_log_check_answers.erb
```diff
@@ -7,31 +7,38 @@
...
               <span class="govuk-!-font-weight-regular app-!-colour-muted"><%= inferred_answer %></span>
             <% end %>
           <% end %>
-
-          <% row.action(
-            text: question.action_text(@log),
-            href: duplicate_logs_action_href(@log, question.page.id, @original_log_id),
-            visually_hidden_text: question.check_answer_label.to_s.downcase,
-          ) %>
+          <% if @all_duplicates.count > 1 %>
```
natdeanlewissoftwire: we could use `many?` here I think if you like, up to you
https://apidock.com/rails/Enumerable/many%3F

natdeanlewissoftwire: there's another place in this PR we could as well though it's not a line you've changed

natdeanlewissoftwire (approved):
looks good!

app/controllers/form_controller.rb
kosiakkatrina: Hmm, maybe, I fell like it's less cluttered if they're separate tho. It would execute the same, cause they've all got returns in them

app/helpers/duplicate_logs_helper.rb
kosiakkatrina: sure, I think that's a good idea!

Airk0n (approved):

PR #1801, opened 2023-07-25
CLDC-2570 Add merge organisations service and rake task

kosiakkatrina:
Add `merge_organisations_service` which allows several merging organisations to be merged into a single absorbing organisation.

Service updates relevant absorbing organisation details and transfers or duplicates some resources from merging organisations to absorbing organisation. These include, rent periods, organisation relationships, users, schemes, locations.

Merge organisations get updated with merge_date and an absorbing_organisation_id, schemes get deactivated and some relationships get removed.

The `merge_organisation_service` is called from a rake task which can be triggered with an absorbing organisation id and a space separated list of merging organisation ids:
```
rake merge:merge_organisations[absorbing_organisation_id, merging_organisation_ids]
```

For example if running it on remote apps:
```
cf ssh <App-name> -t -c "/tmp/lifecycle/launcher /home/vcap/app 'rake merge:merge_organisations[9,\"10 11\"]' ''" 
```

or locally:
```
rake "merge:merge_organisations[2,1 3]”
```

### DIFF
--- a/app/models/lettings_log.rb
+++ b/app/models/lettings_log.rb
@@ -50,7 +50,7 @@ class LettingsLog < Log
         .or(filter_by_postcode(param))
         .or(filter_by_id(param))
   }
-  scope :filter_by_before_startdate, ->(date) { where("lettings_logs.startdate >= ?", date) }
+  scope :after_date, ->(date) { where("lettings_logs.startdate >= ?", date) }
   scope :unresolved, -> { where(unresolved: true) }
 
   scope :filter_by_organisation, ->(org, _user = nil) { where(owning_organisation: org).or(where(managing_organisation: org)) }
--- a/app/services/merge/merge_organisations_service.rb
+++ b/app/services/merge/merge_organisations_service.rb
@@ -93,13 +93,13 @@ private
   end
 
   def merge_sales_logs(merging_organisation)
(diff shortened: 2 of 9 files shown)


app/models/lettings_log.rb
```diff
@@ -50,7 +50,7 @@ class LettingsLog < Log
         .or(filter_by_postcode(param))
         .or(filter_by_id(param))
   }
-  scope :filter_by_before_startdate, ->(date) { where("lettings_logs.startdate >= ?", date) }
+  scope :after_date, ->(date) { where("lettings_logs.startdate >= ?", date) }
```
Airk0n, line 53: I'm slightly surprised that the name has changed from filter by before start date to after date but the method is the same, is it like this gives you the date after? and the previous naming meant like 'we filter out the dates before' ?

app/services/merge/merge_organisations_service.rb
Airk0n, line 1: Looks good

Airk0n (approved):
Looks good, I would say that we might want to get Nat to look as well, I'm not overly confident that I haven't missed something

app/models/lettings_log.rb
kosiakkatrina: This gets all the logs with the startdate after the given date, so it made sense to me to rename it

PR #1803, opened 2023-07-25
CLDC-2591: Rewrite import task

RachaelBooth:
Rewrites rake import task;
- Expects export format more conveniently for how we actually get it
- Moves user invite emails to end of import
- Removes warn logging from log imports to improve usability (we always ignored all the warnings)
- Pulls in a number of changes that had happened in the console scripts we were using for this


RachaelBooth:
The instructions (in confluence) will also need updating - I'll expect to do that asap.

There are more improvements that could be made but I think we're hitting the limit of what it's actually worth doing.

You'll note that I removed rather than re-writing the tests of the rake task itself. This is not ideal, but what was there was going to need to be redone from scratch to be actually useful, and it's no worse than the untested copy-and-paste scripts so I'm feeling ok with it for the moment and given that we should try this for a dry run ASAP in order to make the most of it. We may want to add some tests shortly if there's time available.

app/services/imports/user_import_service.rb
```diff
@@ -26,19 +26,20 @@ def create_user(xml_document)
...
-          name:,
-          password: Devise.friendly_token,
-          phone: user_field_value(xml_document, "telephone-no"),
-          organisation:,
-          role: role(user_field_value(xml_document, "user-type")),
-          is_dpo: is_dpo?(user_field_value(xml_document, "user-type")),
-          is_key_contact: is_key_contact?(user_field_value(xml_document, "contact-priority-id")),
-          active: user_field_value(xml_document, "active"),
-        )
+        user = User.new
```
natdeanlewissoftwire, line 30: is this an efficiency improvement?

lib/tasks/full_import.rake
```diff
@@ -1,33 +1,92 @@
-Import = Struct.new("Import", :import_class, :import_method, :folder)
-
-namespace :core do
-  desc "Import all data XMLs from legacy CORE"
-  task :full_import, %i[archive_path] => :environment do |_task, args|
-    archive_path = args[:archive_path]
-    raise "Usage: rake core:full_import['path/to/archive']" if archive_path.blank?
+namespace :import do
+  desc "Run a full import for the institutions listed in the named file on s3"
```
natdeanlewissoftwire, line 2: is an institution a group of orgs?

lib/tasks/full_import.rake
```diff
@@ -1,33 +1,92 @@
...
+        if archive_service.folder_present?(step.folder)
+          step.import_class.new(archive_service).send(step.import_method, step.folder)
+        end
+      end
+    end
+
+    logs_import_list = [
       Import.new(Imports::LettingsLogsImportService, :create_logs, "logs"),
-      # Import.new(Imports::SalesLogsImportService, :create_logs, "logs"),
+      Import.new(Imports::SalesLogsImportService, :create_logs, "logs"),
```
natdeanlewissoftwire, line 38: seems like a significant change, why weren't we doing this before?

natdeanlewissoftwire (approved):
Just a couple of questions for my own interest, lgtm otherwise

lib/tasks/full_import.rake
RachaelBooth: No, I'm just mixing up old core terminology because the exported file you get from old core has institutions in the name 🤦

lib/tasks/full_import.rake
RachaelBooth: We were in the console scripts we were actually running - as far as I can tell the rake task has actually not been used for some time.

app/services/imports/user_import_service.rb
RachaelBooth: It shouldn't be, but it lets you tell devise to skip the confirmation notification before attempting to save the user - we then trigger them at the end of the import, once all the logs are in (but we need the user objects to exist earlier).

PR #1804, opened 2023-07-25
CLDC-2531: Update application dockerfiles

cschiragb:
The previous Dockerfiles in the codebase were dated and not used very much

We've created one multi-stage Dockerfile for all env's and updated the Dockerfile definition where necessary (e.g. removing unnecessary Docker commands)

We've also updated the Docker compose to target the "development" stage and also to target a specific platform to avoid an issue around building on Mac ARM64 architecture (Mac users with M chips will see this if it not set)


Dockerfile
```diff
@@ -1,71 +1,50 @@
...
 RUN apk add --update --no-cache tzdata && \
   cp /usr/share/zoneinfo/Europe/London /etc/localtime && \
   echo "Europe/London" > /etc/timezone
 
-# build-base: complication tools for bundle
+# build-base: compilation tools for bundle
 # yarn: node package manager
 # postgresql-dev: postgres driver and libraries
-RUN apk add --no-cache build-base yarn postgresql-dev git
+RUN apk add --no-cache build-base yarn postgresql-dev git bash
```
cschiragb: Useful to have bash when you want to debug things on the image - but not essential

cschiragb: We think we need to install git due to a gem called "overcommit". Seems like it's being used for git hooks. Do you know if we still use it?

Dockerfile
```diff
@@ -1,71 +1,50 @@
...
 
-# Copy all files to /app (except what is defined in .dockerignore)
 COPY . /app/
 
-# Compile assets and run webpack. We set a dummy secret key.
-RUN RAILS_ENV=production bundle exec rails SECRET_KEY_BASE=pickasecuretoken assets:precompile
+RUN bundle exec rake assets:precompile
 
-# Cleanup to save space in the production image
-RUN rm -rf node_modules log tmp && \
```
cschiragb, line 39: We don't know why the node modules folder was being deleted in production when it was being created by the `yarn install` step. We're of the opinion that it would need it in order to run?

As for the other files, we're unsure how appropriate they are and whether it's having much benefit removing them, so opted to leave out for now, and observe image sizes / image build speed in the future

Dockerfile
```diff
@@ -1,71 +1,50 @@
...
+  && mv geckodriver /usr/local/bin/
 
-# Copy files generated in the builder image
-COPY --from=builder /app /app
-COPY --from=builder /usr/local/bundle/ /usr/local/bundle/
+CMD RAILS_ENV=${RAILS_ENV} bundle exec rake db:migrate && bundle exec rails s -e ${RAILS_ENV} -p ${PORT} --binding=0.0.0.0
 
-ARG GIT_COMMIT_SHA="UNKNOWN"
-ENV GIT_COMMIT_SHA=${GIT_COMMIT_SHA}
-RUN echo ${GIT_COMMIT_SHA} > ./GIT_COMMIT_SHA
```
cschiragb, line 65: We couldn't work out why this `GIT_COMMIT_SHA` bit is here. It seems to just create an env var and a file with the value "UNKNOWN". Removing it didn't seem to break anything.

If you know what it might be there for we'd be interested to know! 😄
But for now it's been removed

docker-compose.yml
```diff
@@ -20,7 +20,9 @@ services:
   app:
     build:
       context: .
-      dockerfile: ./Dockerfile_dev
+      dockerfile: ./Dockerfile
+      target: development
+    platform: linux/amd64
```
cschiragb, line 25: Mac users with an M1 chip will see an error when trying to build this image, so we've targeted a specific platform for now so that any devs wanting to use this container can do so (works on Windows / Mac)

Only downside is it slows down how long it takes to build the application image if your machine's on different architecture

docker-compose.yml
```diff
@@ -11,7 +11,7 @@ services:
     volumes:
       - dbdata:/var/lib/postgresql/data
     ports:
-      - 5433:5432 # Mapped to 5433 in case Postgres is already running locally on 5432
+      - 8081:5432 # Mapped to 8081 in case Postgres is already running locally on 5432
```
cschiragb, line 14: This port is free on Windows and Mac machines (5433 wasn't on Sam's Mac), so just makes life easier changing it for all

Dockerfile
```diff
@@ -1,71 +1,50 @@
...
+FROM base as staging
 
-ENV PORT=8080
+CMD RAILS_ENV=${RAILS_ENV} bundle exec rake db:migrate && bundle exec rails s -e ${RAILS_ENV} -p ${PORT} --binding=0.0.0.0
```
RachaelBooth: This may be on your list already, but just in case: I don't think these are going to want to be running the migrations on start up - once we have ecs services trying to start multiple containers at once that will be problematic. 

(It does make sense to do so for local dev, of course...)

Dockerfile
```diff
@@ -1,71 +1,50 @@
...
+# build-base: compilation tools for bundle
 # yarn: node package manager
 # postgresql-dev: postgres driver and libraries
-RUN apk add --no-cache build-base yarn postgresql-dev git
+RUN apk add --no-cache build-base yarn postgresql-dev git bash
 
-# Install bundler to run bundle exec
-# This should be the same version as the Gemfile.lock
+# Bundler version should be the same version as what the Gemfile.lock was bundled with
 RUN gem install bundler:2.3.14 --no-document
```
SamSeed-Softwire, line 16: Why is this specific version of `bundler` being used? Is it just what the devs were using at the time the `Gemfile.lock` file was originally created?

Dockerfile
ghost: In a place where I worked previously we used [strong migrations](https://github.com/ankane/strong_migrations) and this multiple container issue was not a problem. ie check if column/index is exists before adding to db

cschiragb: We've removed these commands as we'll want to run them in the pipeline rather than on the AWS ECS instead anyway (for all envs)

Dockerfile
```diff
@@ -1,71 +1,54 @@
...
 
-# Install gems defined in Gemfile
 COPY .ruby-version Gemfile Gemfile.lock /app/
+RUN bundle config set without "development"
```
cschiragb: We're installing all default and test group gems as standard

Presumably we might want the test group gems in case we want to run application tests on the containers built for any env, but wanted to check if you agree?

Dockerfile
```diff
@@ -1,71 +1,54 @@
...
-# The application runs from /app
-WORKDIR /app
+FROM base as development
 
-# Add postgres driver library
-# Add the timezone as it's not configured by default in Alpine
-RUN apk add --update --no-cache libpq tzdata && \
-  cp /usr/share/zoneinfo/Europe/London /etc/localtime && \
-  echo "Europe/London" > /etc/timezone
+RUN bundle config set without ""
```
cschiragb, line 36: Since the `without` command persists what was set before (in this case `without development` further up in the Dockerfile), we need to set it to blank so it installs the "development" and "test" group gems here

Dockerfile
RachaelBooth: Without looking in detail, there is at least one asset compilation step above - I'd expect that to create everything the app is actually using and render node_modules etc unnecessary. In general that's a fairly common pattern - you need various dependencies around to build from but then you only need your built/compiled artifacts to actually run, and making sure you have as little as possible on your production image is ideal.

It would be worth looking into that/what's going on here a bit more.

Dockerfile
RachaelBooth: I don't think we want test group gems on our production instance - they can be a vector for security issues, and also add a bit of bloat (although to be fair not masses of it).

Dockerfile
cschiragb: So it looks like these are ok files to delete in the end. We have implemented this for staging and production - env's where we'd want to keep the image size low. 

We're also deleting the `frontend` folder because we think it's redundant after compiling the assets

However we are no longer deleting the `log` and `tmp` folders, as we discovered that the app tries to make these when it starts up (and needs permissions to do as well)

Dockerfile
```diff
@@ -1,71 +1,89 @@
...
+RUN bundle install --jobs=4 --no-binstubs --no-cache
+
+# Cleanup to reduce image size
+RUN rm -rf node_modules && \
+    rm -rf /usr/local/bundle/cache && \
+    find /usr/local/bundle/gems -name "*.c" -delete && \
+    find /usr/local/bundle/gems -name "*.h" -delete && \
+    find /usr/local/bundle/gems -name "*.o" -delete && \
+    find /usr/local/bundle/gems -name "*.html" -delete && \
+    rm -rf .env  && \
```
hugh-emerson: Shouldn't `.env` not even be copied into the image in the first place? And it still exists in the previous layer, so you're leaking secrets (though only dev ones) in the image.

Dockerfile
hugh-emerson: Looks to me like the intention was to set the build ARG appropriately when it builds the image (CI/CD pipeline could easily do so), and then the sha gets saved to a file. So you can look up the git commit that resulted in the image. But I still agree with deleting it.

Dockerfile
```diff
@@ -1,71 +1,89 @@
...
-ENV PORT=8080
+USER nonroot
 
-EXPOSE ${PORT}
+CMD bundle exec rails s -e ${RAILS_ENV} -p ${PORT} --binding=0.0.0.0
+
+FROM base as staging
+
+RUN bundle config set without "development"
+RUN bundle install --jobs=4 --no-binstubs --no-cache
```
hugh-emerson: Why is this necessary for "staging" and not production?

It would be nice if you could have the same image for both staging and prod.

Dockerfile
```diff
@@ -1,71 +1,89 @@
...
+
+# Cleanup to reduce image size
+RUN rm -rf node_modules && \
+    rm -rf /usr/local/bundle/cache && \
+    find /usr/local/bundle/gems -name "*.c" -delete && \
+    find /usr/local/bundle/gems -name "*.h" -delete && \
+    find /usr/local/bundle/gems -name "*.o" -delete && \
+    find /usr/local/bundle/gems -name "*.html" -delete && \
+    rm -rf .env  && \
+    rm -rf app/frontend
```
hugh-emerson: Note that deleting files will only reduce image size if you squash the layers (docker build --squash) but that's still an experimental feature.

Because otherwise the file still exists in the previous layer.

Dockerfile
hugh-emerson: Only a year old, so not super important, but you may as well update this to a newer version if you can.

But yeah, this currently matches the `BUNDLED WITH` section of gemfile.lock. So bumping the bundler version is kinda out of scope of just updating the Dockerfile

hugh-emerson (commented):
Pipeline has a test failure, which can't be caused by these changes, but maybe just check you're up to date with latest main and that the pipeline succeeds?

```
 1) SalesLogsController GET when displaying a collection of logs when the user is a customer support user when filtering with bulk_upload_id filter with bulk upload that belongs to current user returns logs only associated with the bulk upload
     Failure/Error: expect(page).not_to have_content(excluded_log.id)
       expected `#<Capybara::Node::Simple tag="" path="/">.has_content?(331)` to be falsey, got true
     # ./spec/requests/sales_logs_controller_spec.rb:366:in `block (8 levels) in <top (required)>'
 ```

Even if you're not yet pushing the image to ECR, you should still get GitHub actions to check the image builds. As part of the "staging pipeline"? And it looks like you install firefox etc in the dev image. So could the pipeline build the dev image, run tests within the image, then check the prod image builds?

File: Dockerfile (hunk 7 above)
Comment (cschiragb) [reply Dockerfile:16]: We'll mention this to Rachael / CORE Application team, and leaving as is for now

File: Dockerfile (hunk 3 above)
Comment (cschiragb) [reply Dockerfile:65]: Thanks. It seems like in a way this could be useful. Do you agree with deleting it because there's maybe a better way of recording this info e.g. maybe when the image ends up in the ECR there is something else that can indicate the Git SHA, or it's just not that useful to record?

File: Dockerfile (hunk 11 above)
Comment (cschiragb) [reply Dockerfile]: I think the intention here was that staging _might_ require the test group gems but production definitely will not. Do you disagree? 😃

File: Dockerfile (hunk 10 above)
Comment (cschiragb) [reply Dockerfile]: Indeed this should not be copied as it is mentioned in the `.dockerignore` file so is redundant here. We've removed this command 😀

File: Dockerfile (hunk 12 above)
Comment (cschiragb) [reply Dockerfile]: We don't think we want to use `--squash` as it's an experimental feature and we agree this isn't actually reducing the image size, so have opted to remove this. Do you think there is an advantage to keeping it to reduce attack surface?

File: Dockerfile (hunk 3 above)
Comment (hugh-emerson) [reply Dockerfile:65]: Short answer: I don't think it's that important. You should be able to just see when the image was built in github actions if you really need to.

Imo, the ideal solution would be to push tags to git, e.g. if you publish an image tagged `core:release-2.0`, then automatically tag the commit that built it with "release-2.0". But I think the communitiesuk org prevents github actions from doing so.

**Comment by cschiragb** [comment]
> Pipeline has a test failure, which can't be caused by these changes, but maybe just check you're up to date with latest main and that the pipeline succeeds?
> 
> ```
>  1) SalesLogsController GET when displaying a collection of logs when the user is a customer support user when filtering with bulk_upload_id filter with bulk upload that belongs to current user returns logs only associated with the bulk upload
>      Failure/Error: expect(page).not_to have_content(excluded_log.id)
>        expected `#<Capybara::Node::Simple tag="" path="/">.has_content?(331)` to be falsey, got true
>      # ./spec/requests/sales_logs_controller_spec.rb:366:in `block (8 levels) in <top (required)>'
> ```
> 
> Even if you're not yet pushing the image to ECR, you should still get GitHub actions to check the image builds. As part of the "staging pipeline"? And it looks like you install firefox etc in the dev image. So could the pipeline build the dev image, run tests within the image, then check the prod image builds?

---

We've rebased and saw the test did fail on the first run, however we re-ran them and they passed (as sometimes happens with the CORE application tests)

We have [this ticket](https://dluhcdigital.atlassian.net/browse/CLDC-2562) for updating the pipelines to build the image etc at a later point in time so won't do this now 😃

File: Dockerfile (hunk 11 above)
Comment (hugh-emerson) [reply Dockerfile]: If you can, then using the same image on both staging and production (the only difference being env vars) would reduce a possible source of "but it worked on staging!" surprises. In this particular case, is there any chance of bundler having issues with nested dependencies conflicting for example?

I don't see why you need e.g. selenium drivers installed on staging so I'd just remove it, simplify this file a bit and you have to publish fewer docker images

File: Dockerfile (hunk 12 above)
Comment (hugh-emerson) [reply Dockerfile]: Fair enough. I don't think those files around really affects attack surface. 

If you did want to have a slimmer final image, the way to do it would be to have your final stage start from the Ruby image, not "base" and use `COPY --from` to take precompiled files from the base stage. But like I say, I don't think you need to bother if image size is not an issue.

File: Dockerfile (hunk 2 above)
Comment (SamSeed-Softwire) [reply Dockerfile:39]: We have ended up removing this entire section, i.e. not deleting any of these files, based on this conversation we had with Hugh: https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/1804#discussion_r1283396047

File: Dockerfile (hunk 1 above)
Comment (SamSeed-Softwire) [reply Dockerfile]: We will wait until the main CORE codebase has transitioned to Yarn 2 before we use it here.

File: Dockerfile (hunk 10 above)
Comment (SamSeed-Softwire) [reply Dockerfile]: In response to Hugh's suggestion "Even if you're not yet pushing the image to ECR, you should still get GitHub actions to check the image builds." - we will do this later in the project when updating/building the deployment pipelines. That's out of scope for this ticket.

File: Dockerfile (hunk 11 above)
Comment (SamSeed-Softwire) [reply Dockerfile]: We've updated this now to remove the `staging` stage. The staging and prod apps will both be built from the `prod` stage of the Dockerfile.

File: Dockerfile (hunk 12 above)
Comment (SamSeed-Softwire) [reply Dockerfile]: We have removed this section now - agree it's not worth switching to use `COPY --from`.

**Review by hugh-emerson — APPROVED** [review]
I haven't tested it, but LGTM

File: Dockerfile (hunk 1 above)
Comment (Chirag-Bhatti) [reply Dockerfile]: The versions for all these packages have been fixed now, see updated file

File: Dockerfile [hunk 13]
```diff
@@ -1,71 +1,74 @@
...
+RUN mkdir -p tmp log
+RUN chown -R nonroot tmp log
+RUN chown nonroot db/schema.rb
+
+USER nonroot
+
+CMD bundle exec rails s -e ${RAILS_ENV} -p ${PORT} --binding=0.0.0.0
 
-CMD RAILS_ENV=${RAILS_ENV} bundle exec rake db:migrate && bundle exec rails s -e ${RAILS_ENV} -p ${PORT} --binding=0.0.0.0
+FROM staging as production
```
hugh-emerson: What's the purpose of this? I think it's neater to build and publish a single image for deployment to staging/production.

You've set this up to let you `docker build` with both `--target staging` and `--target production` but they're actually the same?

Chirag-Bhatti: My thinking was it would indicate the Dockerfile can build images for `staging` and `production`, and that they're the same. But I agree it would be neater to just have a `production` stage to serve both, and maybe its easy enough to decipher anyway, so have removed the `staging` stage 😀

hugh-emerson (approved):

RachaelBooth (approved):

PR #1807, opened 2023-07-26
CLDC-2151 ensure new discarded_at field is in the correct date format on export

arfacamble:
ensure new date field converted to ISO 8601 format

I had a look at the spec file for the exporter and it does not seem to be considered necessary to test the formats of individual fields


Airk0n (approved):
Looks good to me

PR #1811, opened 2023-07-27
CLDC-2554 Update back to all logs link

kosiakkatrina:
Back to all logs button kept on redirecting back to the same page. It's now clearing all the filters instead.

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema[7.0].define(version: 2023_07_25_081029) do
+ActiveRecord::Schema[7.0].define(version: 2023_07_19_150610) do
   # These are extensions that must be enabled in order to support this database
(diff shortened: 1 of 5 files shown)


db/schema.rb
```diff
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema[7.0].define(version: 2023_07_25_081029) do
+ActiveRecord::Schema[7.0].define(version: 2023_07_19_150610) do
```
natdeanlewissoftwire, line 13: why is this included?

natdeanlewissoftwire (approved):
looks good, not sure why schema is included but approving pending that :)

db/schema.rb
kosiakkatrina: I don't think there is a 20230725081029 migration, the latest one in the files seems to be 20230719150610, so I'm not sure why 2023_07_25_081029 was committed in the first place. Lmk if I'm missing something tho

PR #1812, opened 2023-07-27
CLDC-2589 Infer housingneeds_other as no

kosiakkatrina:
Set housingneeds other to no in bulk upload if any of the housing needs are given and housing needs the is set to nil


natdeanlewissoftwire (approved):
lgtm :)

PR #1813, opened 2023-07-28
CLDC-2591: Split migration task to give options to run/re-run specific sections

RachaelBooth:
This split is potentially very useful if e.g. an import fails at the logs stage (to avoid needing to re-run the entire thing).

Instructions remain the same for the usual case, I'll add a note about using these in case of errors.


natdeanlewissoftwire (approved):
lgtm!

PR #1818, opened 2023-08-03
CLDC-2621 Flag orgs without coordinators

kosiakkatrina:
- Add import report service and move logs report generation from rake task to the new service
- Generate a separate report of organisations with missing data coordinators

### DIFF
--- a/app/services/imports/import_report_service.rb
+++ b/app/services/imports/import_report_service.rb
@@ -8,7 +8,7 @@ module Imports
 
     BYTE_ORDER_MARK = "\uFEFF".freeze # Required to ensure Excel always reads CSV as UTF-8
 
-    def create_report(report_suffix)
+    def create_reports(report_suffix)
       generate_missing_data_coordinators_report(report_suffix)
       generate_logs_report(report_suffix)
     end
--- a/app/services/imports/import_report_service.rb
+++ b/app/services/imports/import_report_service.rb
@@ -8,11 +8,11 @@ module Imports
 
     BYTE_ORDER_MARK = "\uFEFF".freeze # Required to ensure Excel always reads CSV as UTF-8
 
-    def create_report(report_directory)
-      write_missing_data_coordinators_report(report_directory)
+    def create_report(report_suffix)
+      generate_missing_data_coordinators_report(report_suffix)
     end
 
-    def write_missing_data_coordinators_report(report_directory)
+    def generate_missing_data_coordinators_report(report_suffix)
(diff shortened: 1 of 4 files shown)


lib/tasks/full_import.rake
```diff
@@ -99,31 +99,9 @@ namespace :import do
     raise "Usage: rake import:generate_report['institutions_csv_name']" if institutions_csv_name.blank?
```
natdeanlewissoftwire: nit: this should probably be called `generate_reports` now right?

app/services/imports/import_report_service.rb
```diff
@@ -0,0 +1,57 @@
...
+  class ImportReportService
+    def initialize(storage_service, institutions_csv, logger = Rails.logger)
+      @storage_service = storage_service
+      @logger = logger
+      @institutions_csv = institutions_csv
+    end
+
+    BYTE_ORDER_MARK = "\uFEFF".freeze # Required to ensure Excel always reads CSV as UTF-8
+
+    def create_report(report_suffix)
```
natdeanlewissoftwire, line 11: same for this, should be reports not report imo

natdeanlewissoftwire: same for this, should be plural imo (I like that they're grouped, nice one :) )

natdeanlewissoftwire (approved):
🥳

app/services/imports/import_report_service.rb
```diff
@@ -0,0 +1,57 @@
...
+
+          completed_sales_logs = organisation.owned_sales_logs.where(status: "completed").count
+          in_progress_sales_logs = organisation.owned_sales_logs.where(status: "in_progress").count
+          completed_lettings_logs = organisation.owned_lettings_logs.where(status: "completed").count
+          in_progress_lettings_logs = organisation.owned_lettings_logs.where(status: "in_progress").count
+          report << row.push(completed_lettings_logs, in_progress_lettings_logs, completed_sales_logs, in_progress_sales_logs)
+        end
+      end
+
+      report_name = "MigratedLogsReport_#{report_suffix}.csv"
```
RachaelBooth: I have a feeling this is going to lead to e.g. `MigratedLogsReport_suffix.csv.csv`? Unless people are uploading the institutions csv to s3 without the `.csv` in the name, which I'd hope they weren't.

kosiakkatrina: Oh I see, yeah, that makes sense. I guess we might need to change it for files that are not csv (if we ever want that in the future), but given that we're only writing to .csv atm, happy to change back. Thanks for spotting!

PR #1820, opened 2023-08-04
CLDC-1425 Send notification to old email address when changing it

ghost:
- Send confirmation email to old and previous email addresses when changing some other user email address
- Do not rend reconfirmation emails to users that have confirmed their email address


kosiakkatrina (approved):

spec/requests/users_controller_spec.rb
```diff
@@ -1538,7 +1538,7 @@
           expect(whodunnit_actor.id).to eq(user.id)
         end
 
-        context "when user changes email, dpo and key contact" do
+        context "when user changes email, dpo and key contact", :aggregate_failures do
```
kosiakkatrina, line 1541: Do we want to leave this here?

kosiakkatrina (approved):

spec/requests/users_controller_spec.rb
ghost: this should probably be present in every block where we test more than one thing at once

PR #1822, opened 2023-08-04
CLDC-2539: configure application for PaaS and AWS

Chirag-Bhatti:
https://dluhcdigital.atlassian.net/browse/CLDC-2539

- Added a helper module to determine if the platform is PaaS
- Updated codebase to use existing `EnvConfigurationService` instead of `PaasConfigurationService` if platform is _not_ PaaS
- Make the updates in such a way that we can remove all PaaS specific code easily in the future

### DIFF
--- a/app/jobs/data_export_xml_job.rb
+++ b/app/jobs/data_export_xml_job.rb
@@ -2,7 +2,7 @@ class DataExportXmlJob < ApplicationJob
   queue_as :default
 
   def perform(full_update: false)
-    storage_service = Storage::S3Service.new(Configuration::PaasConfigurationService.new, ENV["EXPORT_PAAS_INSTANCE"])
+    storage_service = Storage::S3Service.new(PlatformHelper.is_paas? ? Configuration::PaasConfigurationService.new : Configuration::EnvConfigurationService.new, ENV["EXPORT_PAAS_INSTANCE"])
     export_service = Exports::LettingsLogExportService.new(storage_service)
 
     export_service.export_xml_lettings_logs(full_update:)
--- a/config/initializers/rack_attack.rb
+++ b/config/initializers/rack_attack.rb
@@ -1,14 +1,18 @@
 require "configuration/configuration_service"
 require "configuration/paas_configuration_service"
+require "configuration/env_configuration_service"
+require Rails.root.join("app/helpers/platform_helper")
+
+configuration_service = PlatformHelper.is_paas? ? Configuration::PaasConfigurationService.new : Configuration::EnvConfigurationService.new
 
 if Rails.env.development? || Rails.env.test?
   Rack::Attack.cache.store = ActiveSupport::Cache::MemoryStore.new
   Rack::Attack.enabled = false
 elsif Rails.env.review?
-  redis_url = Configuration::PaasConfigurationService.new.redis_uris.to_a[0][1]
+  redis_url = configuration_service.redis_uris.to_a[0][1]
   Rack::Attack.cache.store = ActiveSupport::Cache::RedisCacheStore.new(url: redis_url)
 else
-  redis_url = Configuration::PaasConfigurationService.new.redis_uris[:"dluhc-core-#{Rails.env}-redis"]
+  redis_url = configuration_service.redis_uris[:"dluhc-core-#{Rails.env}-redis"]
   Rack::Attack.cache.store = ActiveSupport::Cache::RedisCacheStore.new(url: redis_url)
 end
 
--- a/spec/lib/tasks/data_import_spec.rb
+++ b/spec/lib/tasks/data_import_spec.rb
@@ -3,6 +3,7 @@ require "rake"
 
 describe "data import", type: :task do
   let(:instance_name) { "paas_import_instance" }
+  let(:env_config_service) { instance_double(Configuration::EnvConfigurationService) }
   let(:paas_config_service) { instance_double(Configuration::PaasConfigurationService) }
   let(:storage_service) { instance_double(Storage::S3Service) }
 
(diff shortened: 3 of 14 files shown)


spec/lib/tasks/data_import_spec.rb
```diff
@@ -29,13 +32,22 @@
...
 
-      it "creates an organisation from the given XML file" do
+      it "creates an organisation from the given XML file when the VCAP_SERVICES environment variable exists" do
         expect(Storage::S3Service).to receive(:new).with(paas_config_service, instance_name)
         expect(Imports::OrganisationImportService).to receive(:new).with(storage_service)
         expect(import_service).to receive(:create_organisations).with(fixture_path)
 
         task.invoke(type, fixture_path)
       end
+
```
SamSeed-Softwire, line 42: I didn't think it was worth adding a PaaS/Env test (like this one) for every context block in this file. I felt that testing it once was sufficient.

spec/lib/tasks/date_import_field_spec.rb
```diff
@@ -34,13 +37,22 @@
...
 
-      it "updates the logs from the given XML file" do
+      it "updates the logs from the given XML file when the VCAP_SERVICES environment variable exists" do
         expect(Storage::S3Service).to receive(:new).with(paas_config_service, instance_name)
         expect(storage_service).to receive(:get_file_io).with("spec/fixtures/imports/logs")
         expect(Imports::LettingsLogsFieldImportService).to receive(:new).with(archive_service)
         expect(import_service).to receive(:update_field).with(field, "logs")
         task.invoke(field, fixture_path)
       end
+
```
SamSeed-Softwire, line 47: I didn't think it was worth adding a PaaS/Env test (like this one) for every context block in this file. I felt that testing it once was sufficient.

config/initializers/rack_attack.rb
```diff
@@ -1,14 +1,18 @@
...
+require Rails.root.join("app/helpers/platform_helper")
+
+configuration_service = PlatformHelper.is_paas? ? Configuration::PaasConfigurationService.new : Configuration::EnvConfigurationService.new
 
 if Rails.env.development? || Rails.env.test?
   Rack::Attack.cache.store = ActiveSupport::Cache::MemoryStore.new
   Rack::Attack.enabled = false
 elsif Rails.env.review?
-  redis_url = Configuration::PaasConfigurationService.new.redis_uris.to_a[0][1]
+  redis_url = configuration_service.redis_uris.to_a[0][1]
```
Airk0n, line 12: I don't have a solution or know exactly why it's this way, but it's worth mentioning that I don't like to see array entries hard coded.

spec/lib/tasks/data_import_spec.rb
Airk0n: You're probably right, I can't see another context that might interfere but if you can imagine another context that might affect this test that I would have it in 2, not all of them but just more than 1.

Airk0n (approved):
Looks good to me, I approve but I would consider getting a second(more senior) opinion

app/jobs/data_export_xml_job.rb
```diff
@@ -2,7 +2,7 @@ class DataExportXmlJob < ApplicationJob
   queue_as :default
 
   def perform(full_update: false)
-    storage_service = Storage::S3Service.new(Configuration::PaasConfigurationService.new, ENV["EXPORT_PAAS_INSTANCE"])
+    storage_service = Storage::S3Service.new(PlatformHelper.is_paas? ? Configuration::PaasConfigurationService.new : Configuration::EnvConfigurationService.new, ENV["EXPORT_PAAS_INSTANCE"])
```
RachaelBooth, line 5: FYI: It may have saved some effort to pull this out into a configuration service used everywhere which handles this check internally. Would solve some of the "in how many places do we do a test" questions too.

Given that this state of affairs only exists until we do the "get rid of paas everywhere" tickets, and that we don't often add configuration-y things, I'm happy to leave it if you think the effort of changing it would be more than an hour or so.

RachaelBooth: Separately, `EXPORT_PAAS_INSTANCE` and similar are (and were already) really badly named things about to become more badly named - I'm going to raise a separate ticket for changing that, although it would be worth us discussing in a refinement session when and how to do it.

spec/lib/tasks/data_import_spec.rb
SamSeed-Softwire, line 1: @RachaelBooth there are some test files which reference `PaaSConfigurationService` but which are passing because they don't explicitly test anything to do with the configuration service being used.

We decided not to update these tests now and instead raised a ticket for this, as it felt a bit out of scope for this ticket: https://dluhcdigital.atlassian.net/browse/CLDC-2655.

Let us know if you disagree!

app/jobs/data_export_xml_job.rb
SamSeed-Softwire: As agreed, we're not going to refactor the configuration class conditional logic right now.

config/initializers/rack_attack.rb
SamSeed-Softwire: Thanks Aaron - agree that there might be a better way to do this, but we don't think it's worth updating as part of this PR.

spec/lib/tasks/data_import_spec.rb
SamSeed-Softwire: Thanks. We can't think of another context right now so are happy with just 1 test, but also we're going to change how this works later on when we stop using PaaS, so think time spent on this now might end up being wasted anyway.

Chirag-Bhatti:
Here's a list of the env vars we believe we'll need to configure for running in AWS. We've created `.tfvars` files of these for each env, and stored them in the tf-state s3 buckets for now

- `API_KEY`
- `API_USER`
- `APP_HOST`
- `CSV_DOWNLOAD_PAAS_INSTANCE`
- `DATABASE_URL`
- `EXPORT_PAAS_INSTANCE`
- `GOVUK_NOTIFY_API_KEY`
- `IMPORT_PAAS_INSTANCE`
- `OS_DATA_KEY`
- `RAILS_ENV`
- `RAILS_LOG_TO_STDOUT`
- `RAILS_MASTER_KEY`
- `RAILS_SERVE_STATIC_FILES`
- `REDIS_CONFIG`
- `S3_CONFIG`
- `SENTRY_DSN`

PR #1823, opened 2023-08-04
CLDC-2583 Infer Incref

kosiakkatrina:
Infer incref in all cases and add rake task to update existing incref values

### DIFF
--- a/lib/tasks/correct_incref_values.rake
+++ b/lib/tasks/correct_incref_values.rake
@@ -1,0 +1,6 @@
+desc "Alter incref values for non imported lettings logs in the database"
+task correct_incref_values: :environment do
+  LettingsLog.where(old_id: nil, net_income_known: 0).update!(incref: 0)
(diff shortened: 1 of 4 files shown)


lib/tasks/correct_incref_values.rake
```diff
@@ -0,0 +1,6 @@
+desc "Alter incref values for non imported lettings logs in the database"
+task correct_incref_values: :environment do
+  LettingsLog.where(old_id: nil, net_income_known: 0).update!(incref: 0)
```
ghost, line 3: does `update!` on a collection work? Thought you'd have to use `update_all`

ghost (approved):

lib/tasks/correct_incref_values.rake
kosiakkatrina: Hmm, yeah, I think update will work. I think update_all doesn't run the validations and doesn't update updated_at, cause it runs straight on the db (?)

PR #1825, opened 2023-08-08
CLDC-2505 Bulk upload email link journey

Airk0n:
This ticket is concerned with changing the behaviour of what happens when a user clicks a link they have received by email when they have already started actioning it:

Before: They would have been taken to a page that asks them how they want to fix the issues regardless of if they had started or finished fixing the upload.

After: Now they are taken to the choice page if they haven't made the choice, the page that tells them they've completed the logs if they have, or the page that reminds them they still need to make fixes if they do.

This ticket also had a content change in the form of copy on the completed logs pages(both sales and lettings):
image

### DIFF
--- a/app/views/bulk_upload_lettings_soft_validations_check/chosen.html.erb
+++ b/app/views/bulk_upload_lettings_soft_validations_check/chosen.html.erb
@@ -5,9 +5,9 @@
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds">
     <span class="govuk-caption-l">Bulk upload for lettings (<%= @bulk_upload.year_combo %>)</span>
-    <h1 class="govuk-heading-l">These logs have been created</h1>
+    <h1 class="govuk-heading-l">These logs have been completed</h1>
 
-    <p class="govuk-body">You have created logs from your bulk upload. Return to lettings logs to view them.</p>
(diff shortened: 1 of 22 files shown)


app/views/bulk_upload_lettings_soft_validations_check/chosen.html.erb
```diff
@@ -5,9 +5,9 @@
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds">
     <span class="govuk-caption-l">Bulk upload for lettings (<%= @bulk_upload.year_combo %>)</span>
-    <h1 class="govuk-heading-l">These logs have been created</h1>
+    <h1 class="govuk-heading-l">These logs have been completed</h1>
 
-    <p class="govuk-body">You have created logs from your bulk upload. Return to lettings logs to view them.</p>
+    <p class="govuk-body">You have created logs from your bulk upload, and the logs are complete. Return to lettings logs to view them.</p>
```
kosiakkatrina: If the user lands on this page it wouldn't always mean that the logs have been completed tho, right?
What would happen if they choose to fix and re upload the csv?

kosiakkatrina (approved):
Looks good! Left one comment to sense check the soft validations journey

PR #1830, opened 2023-08-09
CLDC-2504 Add org select to sales if merged

natdeanlewissoftwire:
This PR puts the merge orgs behaviour in the previous PR (https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/1815) for this branch behind a feature flag.


app/models/form/sales/pages/organisation.rb
```diff
@@ -11,25 +11,29 @@ def questions
   end
 
   def routed_to?(log, current_user)
-    return false unless current_user
-    return true if current_user.support?
+    if FeatureToggle.merge_organisations_enabled?
```
Airk0n, line 14: Is it worth using the existing returns or refactoring these conditions to use a similar phrasing. I can see in some instances the feature toggle neatly encloses the method and in one occasion you add it to a condition.
Here it feels messy when followed immediately by return unless. But I can also see the benefit of having a neat encapsulation of the feature flag for the whole method.
I guess I my problem is actually with them lines

Airk0n (approved):
Looks good to me

app/models/form/sales/pages/organisation.rb
natdeanlewissoftwire: I see your point, it'll be easier to remove later if they're all in the same format, I'll make that change to the ternary - the place where it's been added within another condition is necessary to satisfy our linter so I'll leave that one

PR #1833, opened 2023-08-11
CLDC-2656 Import location code

kosiakkatrina:
Get the location code from scheme import if we can't infer it from the postcode


app/services/imports/scheme_location_import_service.rb
```diff
@@ -26,6 +26,10 @@ def create_scheme_location(xml_document)
       4 => "(Registered nursing care home)",
     }.freeze
 
+    LA_NAME_TO_CODE = {
```
Airk0n: So this will be a list of maps when you have that data is that right?

Airk0n (approved):
LGTM

app/services/imports/scheme_location_import_service.rb
kosiakkatrina: Yup, all there now

PR #1834, opened 2023-08-11
Full import optimisation

natdeanlewissoftwire:


RachaelBooth (approved):
LGTM!

(For reference, the export of org relationships is looking at what's in the logs for the relevant years, so this won't overall change which ones are created)

PR #1837, opened 2023-08-14
Skip validations when updating incref

kosiakkatrina:


natdeanlewissoftwire (approved):
looks good, this won't update the updated_at field for these logs which seems desired to me but just wanted to be explicit about that :)

PR #1841, opened 2023-08-15
CLDC-2658 Export additional id fields

kosiakkatrina:
Add additional ID fields to the export to easier locate resources on the new core:
`log_id` - current log id
`scheme`  - current scheme id
`location_code` - current location id
`scheme_old` - old location/scheme id
`scheme_status` - scheme status during the startdate of the log
`location_status` - location status during the startdate of the log
`created_by` - created by user email
`amended_by` - updated by user email

### DIFF
--- a/app/services/exports/lettings_log_export_service.rb
+++ b/app/services/exports/lettings_log_export_service.rb
@@ -196,6 +196,10 @@ module Exports
         attribute_hash["age#{index}"] = -9 if attribute_hash["age#{index}_known"] == 1
       end
 
+      attribute_hash["log_id"] = lettings_log.id
+      attribute_hash["created_by"] = lettings_log.created_by&.email
+      attribute_hash["amended_by"] = lettings_log.updated_by&.email
+
       # Supported housing fields
       if lettings_log.is_supported_housing?
         attribute_hash["unittype_sh"] = lettings_log.unittype_sh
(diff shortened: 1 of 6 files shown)


app/services/exports/lettings_log_export_service.rb
```diff
@@ -196,6 +196,10 @@ def apply_cds_transformation(lettings_log, export_mode)
         attribute_hash["age#{index}"] = -9 if attribute_hash["age#{index}_known"] == 1
       end
 
+      attribute_hash["log_id"] = lettings_log.id
+      attribute_hash["created_by"] = lettings_log.created_by&.email
```
natdeanlewissoftwire, line 200: can we return name/ID if email is nil?

app/services/exports/lettings_log_export_service.rb
```diff
@@ -219,12 +223,16 @@ def add_scheme_fields!(scheme, attribute_hash)
...
+      attribute_hash["scheme"] = scheme.id
+      attribute_hash["scheme_status"] = scheme.status_at(attribute_hash["startdate"])
     end
 
     def add_location_fields!(location, attribute_hash)
       attribute_hash["mobstand"] = location.mobility_type_before_type_cast
-      attribute_hash["scheme"] = location.old_visible_id || (location.id + LOG_ID_OFFSET)
+      attribute_hash["scheme_old"] = location.old_visible_id
       attribute_hash["units"] = location.units
+      attribute_hash["location_code"] = location.id
```
natdeanlewissoftwire, line 234: this is the old core location code right? not the new core local authority code (which we're confusingly calling location_code in our db iirc)

natdeanlewissoftwire (approved):

app/services/exports/lettings_log_export_service.rb
kosiakkatrina: I don't think we allow user records without email, so should be fine
`t.string "email", default: "", null: false`

app/services/exports/lettings_log_export_service.rb
kosiakkatrina: I was confused by this too. This is meant to be the location ID that we have on our service

app/services/exports/lettings_log_export_service.rb
natdeanlewissoftwire: Ah, I thought so - the `&.` made me second guess myself haha

app/services/exports/lettings_log_export_service.rb
natdeanlewissoftwire: Lgtm in that case 👍

PR #1850, opened 2023-08-17
Add owning org id to field import task for sales

natdeanlewissoftwire:


app/services/imports/sales_logs_field_import_service.rb
```diff
@@ -12,6 +14,8 @@ def update_field(field, folder)
   private
 
     def update_creation_method(xml_doc)
+      return if meta_field_value(xml_doc, "form-name").include?("Lettings")
```
natdeanlewissoftwire: unrelated but I think this should be here (same for the change to lettings_logs_field_import_service)

lib/tasks/data_import_field.rake
```diff
@@ -7,13 +7,17 @@ namespace :core do
...
     case field
-    when "tenancycode", "major_repairs", "lettings_allocation", "offered"
+    when "tenancycode", "major_repairs", "lettings_allocation", "offered", "owning_organisation_id"
       s3_service = Storage::S3Service.new(PlatformHelper.is_paas? ? Configuration::PaasConfigurationService.new : Configuration::EnvConfigurationService.new, ENV["IMPORT_PAAS_INSTANCE"])
       archive_io = s3_service.get_file_io(path)
       archive_service = Storage::ArchiveService.new(archive_io)
       if archive_service.folder_present?("logs")
         Rails.logger.info("Start importing field from folder logs")
-        Imports::LettingsLogsFieldImportService.new(archive_service).update_field(field, "logs")
+        if field == "owning_organisation_id"
```
natdeanlewissoftwire: to be fair this if probably isn't necessary, we could run both lettings and sales field import always due to the "unrelated" lines I mentioned earlier but no harm leaving this in to be explicit

app/services/imports/sales_logs_field_import_service.rb
kosiakkatrina: Hmm, I'd say maybe change it to this: `return unless meta_field_value(xml_doc, "form-name").include?("Sales")`
Cause I seem to recall that not all the lettings form names include `Lettings`, that's why we're only checking for sales or not sales everywhere else we do this

app/services/imports/sales_logs_field_import_service.rb
```diff
@@ -28,5 +32,25 @@ def update_creation_method(xml_doc)
...
+
+      old_id = meta_field_value(xml_doc, "document-id")
+      record = SalesLog.find_by(old_id:)
+
+      if record.present?
+        if record.owning_organisation_id.present?
+          @logger.info("sales log #{record.id} has a value for owning_organisation_id, skipping update")
+        else
+          old_owning_organisation_id = safe_string_as_integer(xml_doc, "OWNINGORGID")
```
kosiakkatrina: `find_organisation_id(xml_doc, "OWNINGORGID")` ?

kosiakkatrina: Or `string_or_nil(xml_doc, id_field)` if you're trying to recreate the `find_organisation_id` method

spec/services/imports/sales_logs_field_import_service_spec.rb
```diff
@@ -73,4 +73,41 @@
       end
     end
   end
+
+  context "when updating owning_organisation_id" do
+    let(:field) { "owning_organisation_id" }
+    let(:sales_log_filename) { "shared_ownership_sales_log" }
+
+    context "when the sales log has no offered value" do
+      let(:sales_log) { SalesLog.find_by(old_id: sales_log_filename) }
```
kosiakkatrina, line 82: ~if this doesn't have old_visible_id, how did it find the new organisation ID?~ sorry thought this was a create

app/services/imports/sales_logs_field_import_service.rb
natdeanlewissoftwire: ahh nice ok, good point, thanks

app/services/imports/sales_logs_field_import_service.rb
natdeanlewissoftwire: 🙌

lib/tasks/data_import_field.rake
```diff
@@ -20,4 +20,26 @@ namespace :core do
       raise "Field #{field} cannot be updated by data_import_field"
```
kosiakkatrina: Field #{field} cannot be updated by _lettings_data_import_field_ ? Is there any other documentation we need to update?

natdeanlewissoftwire: are there docs for field importing?

kosiakkatrina: Probably not

kosiakkatrina (approved):

PR #1851, opened 2023-08-17
CLDC-2663 Import sales form id

kosiakkatrina:
Import old form id for sales
Add it to the export instead of old_id
Add a rake task to reimport missing old_form_ids for already migrated orgs


natdeanlewissoftwire (approved):
lgtm 👍

PR #1854, opened 2023-08-18
CLDC-2665 Switch lettings ppcodenk

ghost:
Done:

- refactor and simplify Question model (no need to handle lettings and sales logs differently)
- swap 0 and 1 `ppcodenk` values in code

Partially done:
- updated tests to reflect changes. In the importers there's something odd going because of the way StandardError is rescued. You'll end up in a recursive loop that complicates debugging, my best guess is somewhere I haven't flipped the `ppcodenk` values.  Try running to see`rspec ./spec/services/imports/lettings_logs_import_service_spec.rb:191` 
- update fixtures, some might still need updating


app/models/lettings_log.rb
```diff
@@ -260,8 +260,8 @@ def postcode_known?
   end
 
   def previous_postcode_known?
-    # 1: Yes
-    ppcodenk == 1
+    # 0: Yes
+    postcode_known&.zero?
```
kosiakkatrina: Is this meant to be previous postcode?

app/services/bulk_upload/lettings/year2022/row_parser.rb
```diff
@@ -1283,12 +1283,7 @@ def previous_la_known
   end
 
   def ppcodenk
-    case field_65
-    when 1
-      1
-    when 2
-      0
-    end
+    field_65 == 1 ? 0 : 1
```
kosiakkatrina: should this be a required field tho?

app/models/lettings_log.rb
ghost: yes good catch, fixing

app/services/bulk_upload/lettings/year2022/row_parser.rb
ghost: I think so? Just inverting 0 and 1 in this change, do you think it needs something else?

kosiakkatrina: Checked the values for this and updated!

app/models/form/question.rb
```diff
@@ -166,28 +166,17 @@ def label_from_value(value, _log = nil, _user = nil)
...
       value == 1
     when "radio"
-      is_lettings ? RADIO_YES_VALUE_LETTINGS[id.to_sym]&.include?(value) : RADIO_YES_VALUE_SALES[id.to_sym]&.include?(value)
+      RADIO_YES_VALUE[id.to_sym]&.include?(value)
     else
       %w[yes].include?(value.downcase)
     end
   end
 
-  def value_is_no?(value, is_lettings)
```
natdeanlewissoftwire, line 180: I assume we don't use this anywhere?

app/models/form/question.rb
```diff
@@ -338,6 +327,7 @@ def inferred_answer_value(log)
     is_carehome: [1],
     hbrentshortfall: [1],
     net_income_value_check: [0],
+    ppcodenk: [0],
   }.freeze
 
   RADIO_NO_VALUE = {
```
natdeanlewissoftwire, line 344: Do we need this if we don't use value_is_no?

natdeanlewissoftwire (approved):
lgtm if you're happy you've caught everywhere this change needs to happen :)

app/models/form/question.rb
kosiakkatrina: Doesn't look like it

app/models/form/question.rb
kosiakkatrina: Uhm, good point. I think I might bring back value_is_no? Cause we've got those other convenience methods

PR #1856, opened 2023-08-21
CLDC-2684 Add some logging to export

kosiakkatrina:
Currently running the full export runs out of memory, this PR adds some logging to test how far the export gets before that happens.

### DIFF
--- a/app/services/exports/lettings_log_export_service.rb
+++ b/app/services/exports/lettings_log_export_service.rb
@@ -36,7 +36,7 @@ module Exports
     end
 
     def build_export_run(collection, current_time, base_number, full_update)
-      logger.info("Building export run for #{collection}")
+      @logger.info("Building export run for #{collection}")
       previous_exports_with_data = LogsExport.where(collection:, empty_export: false)
 
       increment_number = previous_exports_with_data.where(base_number:).maximum(:increment_number) || 1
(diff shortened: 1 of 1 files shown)


jackscotti (approved):

ghost (approved):

app/services/exports/lettings_log_export_service.rb
```diff
@@ -91,6 +93,11 @@ def write_export_archive(export, lettings_logs)
 
       # Write all archives
       archive_datetimes = {}
+      @logger.info("Following archives to write:")
+      # rubocop:disable Style/CombinableLoops
+      lettings_logs_per_archive.each do |archive, lettings_logs_to_export|
```
natdeanlewissoftwire, line 98: if we define each of these loops that rubocop wants to combine as a method, I think that might avoid us having to disable the linter

natdeanlewissoftwire (approved):

app/services/exports/lettings_log_export_service.rb
natdeanlewissoftwire: actually I'm not sure that's true, but regardless do we really want to be logging before we're actually doing the thing we're logging to say we're doing?

kosiakkatrina: Yeah, I think cause it's running out of memory before completing it might not print all of these out if added to the loop below, I will likely remove this logging it's mostly for testing atm

natdeanlewissoftwire: Sure, I'm convinced this is fine then :)

PR #1859, opened 2023-08-21
CLDC-2684 Keep fewer logs in memory when running exports

kosiakkatrina:
In the past we would retrieve all the logs that we want to export and keep them in `logs_by_collection`. When testing full export on staging it wouldn't get past this step because the number of logs we'd try to keep in memory was quite large.

This PR attempts to reduce that by querying and writing logs to zip buffer in batches of `MAX_XML_RECORDS` (10_000). It's the same format of writing them as before except we now only keep up to 10_000 logs in a variable at a time.

Further this PR refactors `write_export_archive` by dropping the loop through all the archives that are to be zipped and uploaded. As we call this method for each collection year and only retrieve the logs relevant to that collection year it seems that there would ever be a single archive per call.

### DIFF
--- a/app/services/exports/lettings_log_export_service.rb
+++ b/app/services/exports/lettings_log_export_service.rb
@@ -63,53 +63,58 @@ module Exports
       @storage_service.write_file(file_path, string_io)
     end
 
-    def get_archive_name(lettings_log, base_number, increment)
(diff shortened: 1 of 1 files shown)


RachaelBooth (approved):
Looks good to me!

PR #1860, opened 2023-08-23
CLDC-2684 Retrieve recent export before creating new one

kosiakkatrina:
We look at the recent export to determine which logs to export in the partial export
https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/1859 has shifted `retrieve_lettings_logs` into the loop so a new export would be saved before we could write the second collection's logs into the export.
Now we retrieve the recent export outside of the loop and pass it down to `retrieve_lettings_logs`

[CLDC-2684]: https://dluhcdigital.atlassian.net/browse/CLDC-2684?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/services/exports/lettings_log_export_service.rb
+++ b/app/services/exports/lettings_log_export_service.rb
@@ -13,9 +13,10 @@ module Exports
       daily_run_number = get_daily_run_number
       archives_for_manifest = {}
       base_number = LogsExport.where(empty_export: false).maximum(:base_number) || 1
+      recent_export = LogsExport.order("started_at").last
       available_collection_years.each do |collection|
         export = build_export_run(collection, start_time, base_number, full_update)
-        archives = write_export_archive(export, collection, start_time, full_update)
+        archives = write_export_archive(export, collection, start_time, recent_export, full_update)
 
         archives_for_manifest.merge!(archives)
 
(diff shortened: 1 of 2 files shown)


app/services/exports/lettings_log_export_service.rb
```diff
@@ -71,13 +72,12 @@ def get_archive_name(collection, base_number, increment)
       "core_#{collection}_#{collection + 1}_apr_mar_#{base_number_str}_#{increment_str}".downcase
     end
 
-    def write_export_archive(export, collection, start_time, full_update)
-      @logger.info("Writing export archives")
+    def write_export_archive(export, collection, start_time, recent_export, full_update)
       archive = get_archive_name(collection, export.base_number, export.increment_number) # archive name would be the same for all logs because they're already filtered by year (?)
 
       # Write archive
```
natdeanlewissoftwire:
```suggestion
      # Create archive
```

natdeanlewissoftwire (approved):

app/services/exports/lettings_log_export_service.rb
kosiakkatrina: I mean I'll remove it fully, cause it's literally in the method name 😅

PR #1866, opened 2023-08-24
CLDC-2702 Prioritise addresses over UPRN on import

kosiakkatrina:
Addresses coming through the import seem to be more reliable than UPRN, so we want to prioritise them

### DIFF
--- a/app/services/imports/lettings_logs_import_service.rb
+++ b/app/services/imports/lettings_logs_import_service.rb
@@ -200,13 +200,13 @@ module Imports
       attributes["first_time_property_let_as_social_housing"] = first_time_let(attributes["rsnvac"])
       attributes["declaration"] = declaration(xml_doc)
 
-      attributes["uprn"] = string_or_nil(xml_doc, "UPRN")
-      attributes["uprn_known"] = attributes["uprn"].present? ? 1 : 0
-      attributes["uprn_confirmed"] = attributes["uprn"].present? ? 1 : 0
       attributes["address_line1"] = string_or_nil(xml_doc, "AddressLine1")
       attributes["address_line2"] = string_or_nil(xml_doc, "AddressLine2")
       attributes["town_or_city"] = string_or_nil(xml_doc, "TownCity")
       attributes["county"] = string_or_nil(xml_doc, "County")
+      attributes["uprn"] = address_given?(attributes) ? nil : string_or_nil(xml_doc, "UPRN")
+      attributes["uprn_known"] = attributes["uprn"].present? ? 1 : 0
+      attributes["uprn_confirmed"] = attributes["uprn"].present? ? 1 : 0
 
       set_partial_charges_to_zero(attributes)
 
@@ -649,6 +649,10 @@ module Imports
       return if parent_organisation_id == child_organisation_id
 
       OrganisationRelationship.find_or_create_by!(parent_organisation_id:, child_organisation_id:)
+    end
+
+    def address_given?(attributes)
+      attributes["address_line1"].present? && attributes["town_or_city"].present?
     end
   end
 end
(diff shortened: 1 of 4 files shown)


app/services/imports/lettings_logs_import_service.rb
```diff
@@ -650,5 +650,9 @@ def create_organisation_relationship!(attributes)
 
       OrganisationRelationship.find_or_create_by!(parent_organisation_id:, child_organisation_id:)
     end
+
+    def address_given?(attributes)
+      attributes["address_line1"].present? && attributes["town_or_city"].present?
+    end
```
RachaelBooth, line 656: Does this need to look at the postcode at all?

RachaelBooth: (I've lost track of how that's asked about on old core, so maybe not)

kosiakkatrina: Good point.

So, I mostly went off from the fact that on this page we're only validating address_line_1 and town_or_city, so existence of those 2 fields would be enough for the data on this page not to get cleared. 
My thoughts were that if we have any address data that can be imported, we prioritise that, but perhaps we only prioritise it when the entire address is given. I'll check on the ticket. Thanks!
image

kosiakkatrina: This seems to be fine

app/services/imports/lettings_logs_import_service.rb
```diff
@@ -200,13 +200,13 @@ def create_log(xml_doc)
...
-      attributes["uprn"] = string_or_nil(xml_doc, "UPRN")
-      attributes["uprn_known"] = attributes["uprn"].present? ? 1 : 0
-      attributes["uprn_confirmed"] = attributes["uprn"].present? ? 1 : 0
       attributes["address_line1"] = string_or_nil(xml_doc, "AddressLine1")
       attributes["address_line2"] = string_or_nil(xml_doc, "AddressLine2")
       attributes["town_or_city"] = string_or_nil(xml_doc, "TownCity")
       attributes["county"] = string_or_nil(xml_doc, "County")
+      attributes["uprn"] = address_given?(attributes) ? nil : string_or_nil(xml_doc, "UPRN")
+      attributes["uprn_known"] = attributes["uprn"].present? ? 1 : 0
+      attributes["uprn_confirmed"] = attributes["uprn"].present? ? 1 : 0
```
natdeanlewissoftwire, line 209: if there's no uprn isn't the standard behaviour for `uprn_confirmed` to be nil?

spec/services/imports/lettings_logs_import_service_spec.rb
```diff
@@ -1531,7 +1531,24 @@ def open_file(directory, filename)
         expect(lettings_log&.postcode_full).to eq("A1 1AA")
       end
 
-      it "correctly sets address and uprn if uprn is given" do
+      it "prioritises address and doesn't set UPRN if both address and UPRN is given" do
+        lettings_log_service.send(:create_log, lettings_log_xml)
+
+        lettings_log = LettingsLog.find_by(old_id: lettings_log_id)
+        expect(lettings_log&.uprn_known).to eq(0) # no
+        expect(lettings_log&.uprn).to be_nil
```
natdeanlewissoftwire, line 1539: we should check `uprn_confirmed` is whatever we expect here too I think

natdeanlewissoftwire (approved):
couple of comments but happy to approve pending those

app/services/imports/lettings_logs_import_service.rb
kosiakkatrina: Hmm, I only moved it. I assume if it worked before it would retain the same behaviour

PR #1867, opened 2023-08-24
CLDC-2662 Add incomplete locations validation

kosiakkatrina:
We are currently not validating if an incomplete location is selected for a log, because in the single log submission journey we don't display incomplete locations, therefore there is on way of it being selected.

It is causing issues during the import where some of those incomplete locations get set.

This PR adds a validation and clears the location fields if the validation is hit upon import.

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -305,6 +305,8 @@ en:
       lettype:
         general_needs_mismatch: Lettings type must be a general needs type because you selected general needs when uploading the file
         supported_housing_mismatch: Lettings type must be a supported housing type because you selected supported housing when uploading the file
+      location:
+        incomplete: "This location is incomplete"
 
     property:
       uprn:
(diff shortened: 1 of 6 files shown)


config/locales/en.yml
```diff
@@ -305,6 +305,8 @@ en:
       lettype:
         general_needs_mismatch: Lettings type must be a general needs type because you selected general needs when uploading the file
         supported_housing_mismatch: Lettings type must be a supported housing type because you selected supported housing when uploading the file
+      location:
+        incomplete: "This location is incomplete"
```
natdeanlewissoftwire: Has this copy been approved? We normally say something like "Select another location or update this one" in this case to make it clear to the user what to do

natdeanlewissoftwire (approved):
1 thought, lgtm though

config/locales/en.yml
kosiakkatrina: I'm not sure if this is actually possible to trigger in the UI, I think it's more for import logging, but I'll ask in the ticket. I'll also add your suggestion

PR #1870, opened 2023-08-25
CLDC-2630 Location fixes

kosiakkatrina:
Show `You didn't answer this question` next to local authority answer. It was blank before because the value was and empty string instead of nil

Fix the back button for local authority to navigate back to check answers.

### DIFF
--- a/spec/features/schemes_spec.rb
+++ b/spec/features/schemes_spec.rb
@@ -553,7 +553,7 @@ RSpec.describe "Schemes scheme Features" do
             expect(page).to have_content "Check your answers"
           end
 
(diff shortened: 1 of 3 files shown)


spec/features/schemes_spec.rb
```diff
@@ -552,6 +552,13 @@
             fill_in_and_save_second_location
             expect(page).to have_content "Check your answers"
           end
+
+          it "dispays correct text for uunanswered questions" do
```
natdeanlewissoftwire:
```suggestion
          it "dispays correct text for unanswered questions" do
```

natdeanlewissoftwire (approved):

PR #1872, opened 2023-08-25
Infra migration updates

RachaelBooth:
There will be some clean up to do here later (it will become a lot nicer when we don't need the app to support both paas and ecs!) - for now this covers some changes required to have a working environment on AWS. 

I am willing to let it be a little messy (in contained locations) for the moment - with a promise of cleanup within the scope of the infra migration work. Hopefully others agree!


SamSeed-Softwire (approved):
Lgtm! Agree it's a little messy, but not a big deal right now especially as we're cleaning up the PaaS/non-PaaS logic in a few months.

PR #1873, opened 2023-08-29
CLDC-2689 Export migrated logs

kosiakkatrina:
We have been skipping imported logs in the partial export, because updated_at column is set getting set to what we get from the import source instead of current time.
We would also skip any logs that had their values corrected manually (for example fields were inferred incorrectly, we changed mapping or fields were re-imported).

We use updated_at for determining which logs have been modified and need to be exported.
We want to keep updated_at as it currently is for imported or corrected logs, so this PR adds an value_updated_at column to ensure we export all the logs that have changed.

### DIFF
--- a/db/migrate/20230828145454_add_migrated_on_fields.rb
+++ b/db/migrate/20230828145454_add_migrated_on_fields.rb
@@ -1,0 +1,6 @@
+class AddMigratedOnFields < ActiveRecord::Migration[7.0]
+  def change
+    add_column :lettings_logs, :imported_at, :datetime
+    add_column :sales_logs, :imported_at, :datetime
+  end
+end
--- a/db/migrate/20230828145454_add_migrated_on_fields.rb
+++ b/db/migrate/20230828145454_add_migrated_on_fields.rb
@@ -1,6 +1,6 @@
 class AddMigratedOnFields < ActiveRecord::Migration[7.0]
   def change
-    add_column :lettings_logs, :imported_at, :datetime
-    add_column :sales_logs, :imported_at, :datetime
+    add_column :lettings_logs, :values_updated_at, :datetime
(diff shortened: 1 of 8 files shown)


spec/services/imports/sales_logs_import_service_spec.rb
```diff
@@ -139,17 +147,6 @@ def open_file(directory, filename)
       end
     end
 
-    context "when the log is valid" do
```
natdeanlewissoftwire: why do we remove this?

db/migrate/20230828145454_add_migrated_on_fields.rb
```diff
@@ -0,0 +1,6 @@
+class AddMigratedOnFields < ActiveRecord::Migration[7.0]
+  def change
+    add_column :lettings_logs, :values_updated_at, :datetime
```
natdeanlewissoftwire, line 3: if this is just a migrated at date, why don't we call it that?

natdeanlewissoftwire (approved):
Are there logs (e.g. the ones we were looking at yesterday afternoon) that have an unexpected updated_at that will affect exports in an unaccounted for way?

Also, I assume this field also needs to be updated for previous migration batches depending when they were migrated?

db/migrate/20230828145454_add_migrated_on_fields.rb
kosiakkatrina: Cause we decided to make this more generic, so that if in the future we need to adjust values we can use the same column. Added some info in description

spec/services/imports/sales_logs_import_service_spec.rb
kosiakkatrina: Whoops, fixed now

PR #1875, opened 2023-08-30
CLDC-2684 Allow full export per year

kosiakkatrina:
Full exports are quite large, so processing them after they have been exported takes a lot of time.
As the processor runs once a day, the intention is to export logs for different years on separate days and this PR would enable us to do that.

### DIFF
--- a/app/services/exports/lettings_log_export_service.rb
+++ b/app/services/exports/lettings_log_export_service.rb
@@ -14,7 +14,7 @@ module Exports
       archives_for_manifest = {}
       base_number = LogsExport.where(empty_export: false).maximum(:base_number) || 1
       recent_export = LogsExport.order("started_at").last
-      available_collection_years(collection_year).each do |collection|
+      collection_years_to_export(collection_year).each do |collection|
         export = build_export_run(collection, start_time, base_number, full_update)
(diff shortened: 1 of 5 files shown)


natdeanlewissoftwire (approved):

app/services/exports/lettings_log_export_service.rb
```diff
@@ -8,13 +8,13 @@ def initialize(storage_service, logger = Rails.logger)
...
 
-    def export_xml_lettings_logs(full_update: false)
+    def export_xml_lettings_logs(full_update: false, collection_year: nil)
       start_time = Time.zone.now
       daily_run_number = get_daily_run_number
       archives_for_manifest = {}
       base_number = LogsExport.where(empty_export: false).maximum(:base_number) || 1
       recent_export = LogsExport.order("started_at").last
-      available_collection_years.each do |collection|
+      available_collection_years(collection_year).each do |collection|
```
natdeanlewissoftwire: not a major issue, I think this naming is slightly confusing - maybe a method called `filtered_collection_years` that returns `[collection_year] if collection_year.present`, and `available_collection_years` otherwise would be clearer?

kosiakkatrina: Renamed it, thanks!

PR #1876, opened 2023-08-30
CLDC-2699 Add extra options for renewal reason

kosiakkatrina:
image


natdeanlewissoftwire (approved):
does this need any tests?

PR #1881, opened 2023-09-01
Migration fixes

natdeanlewissoftwire:
2 fixes in advance of the migration: 

1. accept decimals as values for offered as some orgs list 0 as 0.00
2. add in additional options for sex and relat questions as the "refused" values being exported from old core are different to what the code previously expected

### DIFF
--- a/app/services/imports/lettings_logs_import_service.rb
+++ b/app/services/imports/lettings_logs_import_service.rb
@@ -174,7 +174,7 @@ module Imports
                                      0
                                    end
 
-      attributes["offered"] = safe_string_as_integer(xml_doc, "Q20")
+      attributes["offered"] = unsafe_string_as_integer(xml_doc, "Q20")
       attributes["propcode"] = string_or_nil(xml_doc, "Q21a")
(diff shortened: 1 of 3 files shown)


app/services/imports/lettings_logs_import_service.rb
```diff
@@ -174,7 +174,7 @@ def create_log(xml_doc)
                                      0
                                    end
 
-      attributes["offered"] = safe_string_as_integer(xml_doc, "Q20")
+      attributes["offered"] = unsafe_string_as_integer(xml_doc, "Q20")
```
kosiakkatrina: safe_string_as_decimal?

kosiakkatrina (approved):

app/services/imports/lettings_logs_import_service.rb
natdeanlewissoftwire: ahh I didn't know that existed, thanks

kosiakkatrina: Not sure if this was a good suggestion, cause we probably wanted the integer to get saved, but it looks like the tests pass

PR #1882, opened 2023-09-01
Deploy to aws in staging pipeline

RachaelBooth:


.github/workflows/aws_deploy.yml
```diff
@@ -0,0 +1,87 @@
...
+        id: ecr-login
+        uses: aws-actions/amazon-ecr-login@v1
+        with:
+          mask-password: 'true'
+
+      - name: Get timestamp
+        id: timestamp
+        run: echo "timestamp=$(date +%Y%m%d%H%M%S)" >> $GITHUB_ENV
+
+      - name: Add environment tag to existing image
```
RachaelBooth, line 80: This separation will make it much easier later to sort out production deployments (where the image with the commit tag will already exist), plus allow us to just re-run deploy if image push succeeded but deploy failed for some reason (without running into problems with tag immutability).

.github/workflows/aws_deploy.yml
```diff
@@ -0,0 +1,87 @@
...
+        env:
+          registry: ${{ steps.ecr-login.outputs.registry }}
+          commit_tag: ${{ github.sha }}
+          readable_tag: ${{ inputs.environment }}-${{ env.timestamp }}
+        run: |
+          manifest=$(aws ecr batch-get-image --repository-name $repository --image-ids imageTag=$commit_tag --output text --query images[].imageManifest)
+          aws ecr put-image --repository-name $repository --image-tag $readable_tag --image-manifest "$manifest"
```
SamSeed-Softwire, line 87: Is this not removing the commit SHA and if so, is that right? I thought was always wanted to keep that?

.github/workflows/aws_deploy.yml
```diff
@@ -0,0 +1,87 @@
...
+          echo "image=$registry/$repository:$readable_tag" >> $GITHUB_ENV
+
+      - name: TODO
```
SamSeed-Softwire: Is this TODO because it's going to be done in another ticket? Or is there anything else you want to add as part of https://dluhcdigital.atlassian.net/browse/CLDC-2562?

SamSeed-Softwire:
I'm unsure how you were able to test this pipeline - let's discuss on Monday.

.github/workflows/aws_deploy.yml
RachaelBooth: No, this command adds the given tag as an additional tag.

.github/workflows/aws_deploy.yml
RachaelBooth: It's todo because this step only exists until the next ticket adds in an actual step using the image, I just wanted one to check the image env variable was getting set right and ended up with the tags I'd expected - probably should have called it something that encompassed that but I couldn't immediately think of one so it was more "todo - the rest of this pipeline".

SamSeed-Softwire (approved):
Lgtm 🙂

.github/workflows/staging_pipeline.yml
```diff
@@ -230,3 +230,15 @@ jobs:
           cf set-env $APP_NAME CSV_DOWNLOAD_PAAS_INSTANCE $CSV_DOWNLOAD_PAAS_INSTANCE
           cf set-env $APP_NAME SENTRY_DSN $SENTRY_DSN
           cf push $APP_NAME --strategy rolling -t 180
+
+  aws_deploy:
+    name: AWS Deploy
+    if: github.ref == 'refs/heads/main'
+    needs: [lint, test, feature_test, audit]
```
SamSeed-Softwire, line 237: Do we need lint, test etc for this job? Is that not only necessary for the (non-AWS) deploy job?

SamSeed-Softwire (approved):

.github/workflows/staging_pipeline.yml
RachaelBooth: Yes - we want this deployment to happen if and only if the paas one does (ignoring failures during deployment, but the same triggers/prerequisites). If nothing else, it would be a very easy thing to miss when removing the paas deployment later.

SamSeed-Softwire: Gotcha, thanks!

PR #1883, opened 2023-09-01
CLDC-2753 Add unassigned user for imported logs

kosiakkatrina:
Assign imported logs that we cannot find a user for to "Unassigned" user and add logging and reporting on the logs without user.

### DIFF
--- a/app/services/imports/lettings_logs_import_service.rb
+++ b/app/services/imports/lettings_logs_import_service.rb
@@ -251,6 +251,23 @@ module Imports
       owner_id = meta_field_value(xml_doc, "owner-user-id").strip
       if owner_id.present?
         user = LegacyUser.find_by(old_user_id: owner_id)&.user
+        if user.blank?
+          @logger.error("Lettings log '#{attributes['old_id']}' belongs to legacy user with owner-user-id: '#{owner_id}' which cannot be found. Assigning log to 'Unassigned' user.")
+          if User.find_by(name: "Unassigned")
+            user = User.find_by(name: "Unassigned")
+          else
+            user = User.new(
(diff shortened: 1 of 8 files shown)


app/services/imports/lettings_logs_import_service.rb
```diff
@@ -251,6 +251,23 @@ def create_log(xml_doc)
...
+          @logger.error("Lettings log '#{attributes['old_id']}' belongs to legacy user with owner-user-id: '#{owner_id}' which cannot be found. Assigning log to 'Unassigned' user.")
+          if User.find_by(name: "Unassigned")
+            user = User.find_by(name: "Unassigned")
+          else
+            user = User.new(
+              name: "Unassigned",
+              organisation_id: attributes["managing_organisation_id"],
+              is_dpo: false,
+              encrypted_password: SecureRandom.hex(10),
+              email: SecureRandom.uuid,
```
natdeanlewissoftwire, line 264: think this makes sense, good call

app/services/imports/user_import_service.rb
```diff
@@ -39,9 +39,15 @@ def create_user(xml_document)
...
-        user.legacy_users.create!(old_user_id:)
-        user
+
+        begin
+          user.save!
+          user.legacy_users.create!(old_user_id:)
+          user
+        rescue ActiveRecord::RecordInvalid => e
+          @logger.error(e.message)
+          @logger.error("Could not save user with email: #{email}")
```
natdeanlewissoftwire, line 49: 🎉

natdeanlewissoftwire (approved):
nice one!

PR #1890, opened 2023-09-05
CLDC-2713 Add a rake task to reimport addresses from csv

kosiakkatrina:
Some organisations have wrong addresses imported as the UPRN from the import were not correct.

Those organisation have now provided us with a csv file with the correct addresses that we would want to reimport.


RachaelBooth (approved):
Looks good to me. I was wondering about error handling but I'm assuming given the numbers that we're fine to just look through for anything logged and sort it.

kosiakkatrina:
> Looks good to me. I was wondering about error handling but I'm assuming given the numbers that we're fine to just look through for anything logged and sort it.

Good point, not expecting large numbers for this one for sure. I can only think about the postcode format validation for this one, so should be able to catch those in the logs

PR #1891, opened 2023-09-05
CLDC-2696 Add reimport addresses task

kosiakkatrina:
Some organisations have wrong addresses imported as the UPRN from the import were not correct.
We have since started prioritising manually entered addresses over UPRN during the import, and the addresses for impacted organisations should be reimported

### DIFF
--- a/app/services/imports/lettings_logs_field_import_service.rb
+++ b/app/services/imports/lettings_logs_field_import_service.rb
@@ -170,10 +170,8 @@ module Imports
     end
 
     def postcode_known(record)
-      if record.postcode_full.nil? && record.la.nil?
-        nil
-      elsif record.postcode_full.nil?
-        0 # Assumes we selected No in the form since the LA is present
+      if record.postcode_full.nil?
+        record.la.nil? ? nil : 0 # Assumes we selected No in the form since the LA is present
       else
         1
       end
--- a/app/services/imports/lettings_logs_field_import_service.rb
+++ b/app/services/imports/lettings_logs_field_import_service.rb
@@ -12,6 +12,8 @@ module Imports
         import_from(folder, :update_offered)
       when "creation_method"
         import_from(folder, :update_creation_method)
+      when "address"
+        import_from(folder, :update_address)
       else
         raise "Updating #{field} is not supported by the field import service"
       end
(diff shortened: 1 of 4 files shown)


app/services/imports/lettings_logs_field_import_service.rb
```diff
@@ -132,5 +134,49 @@ def update_tenant_code(xml_doc)
         @logger.warn("Could not find record matching legacy ID #{old_id}")
       end
     end
+
+    def update_address(xml_doc)
+      return if meta_field_value(xml_doc, "form-name").include?("Sales")
+
+      old_id = meta_field_value(xml_doc, "document-id")
+      record = LettingsLog.find_by(old_id:)
+      return @logger.info("lettings log #{record.id} is from previous collection year, skipping") if record.collection_start_year < 2023
```
natdeanlewissoftwire, line 143: is this right? does this error not occur in imported 22/23 logs too?

app/services/imports/lettings_logs_field_import_service.rb
```diff
@@ -132,5 +134,49 @@ def update_tenant_code(xml_doc)
...
+          record.uprn_confirmed = 0
+          record.values_updated_at = Time.zone.now
+          record.save!
+          @logger.info("lettings log #{record.id} address_line1 value has been set to #{record.address_line1}")
+          @logger.info("lettings log #{record.id} address_line2 value has been set to #{record.address_line2}")
+          @logger.info("lettings log #{record.id} town_or_city value has been set to #{record.town_or_city}")
+          @logger.info("lettings log #{record.id} county value has been set to #{record.county}")
+          @logger.info("lettings log #{record.id} postcode_full value has been set to #{record.postcode_full}")
+        else
+          @logger.info("lettings log #{record.id} does not include address_line1 or town or city, skipping")
```
natdeanlewissoftwire:
```suggestion
          @logger.info("lettings log #{record.id} is missing either or both of address_line1 and town or city, skipping")
```

app/services/imports/lettings_logs_field_import_service.rb
```diff
@@ -132,5 +134,49 @@ def update_tenant_code(xml_doc)
...
+        end
+      else
+        @logger.warn("Could not find record matching legacy ID #{old_id}")
+      end
+    end
+
+    def postcode_known(record)
+      if record.postcode_full.nil? && record.la.nil?
```
natdeanlewissoftwire: I think this would be more readable as a nested if but that's just personal preference

natdeanlewissoftwire (approved):
lgtm!

app/services/imports/lettings_logs_field_import_service.rb
kosiakkatrina: I think we've not been collecting UPRNs before 2023

app/services/imports/lettings_logs_field_import_service.rb
kosiakkatrina: better?

PR #1892, opened 2023-09-06
CLDC-2724 Add missing data reporting for imported lettings logs

kosiakkatrina:
Add a task that creates 2 files. 
- A report of all imported, in progress logs with missing answers count, for example if there are 30 logs that are only missing an answer to `beds` question, the report will have a `beds,30` row. If there are 10 logs that are only missing answers to age1_known and beds, there would be a row with `"age1_known,beds",10` etc
- Example log IDs for all the combinations of missing answers

This is to identify if there are fields that are getting cleared in large quantities during the imports (for example due to validations). However this does not identify whether the fields have been cleared during the import or they haven't existed on the old service either.

### DIFF
--- a/app/services/imports/import_report_service.rb
+++ b/app/services/imports/import_report_service.rb
@@ -96,10 +96,10 @@ module Imports
 
         if unanswered_question_counts[unanswered_questions].present?
           unanswered_question_counts[unanswered_questions] += 1
-          missing_answers_example_sets[unanswered_questions] << { id: lettings_log.id, old_id: lettings_log.old_id, owning_organisation_id: lettings_log.owning_organisation_id } unless unanswered_question_counts[unanswered_questions] > 10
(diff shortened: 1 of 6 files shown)


app/services/imports/import_report_service.rb
```diff
@@ -84,5 +84,52 @@ def generate_unassigned_logs_report(report_suffix)
...
     end
+
+    def generate_missing_answers_report(report_suffix)
+      Rails.logger.info("Generating missing imported lettings logs answers report")
+      unanswered_question_counts = {}
+      missing_answers_example_sets = {}
+
+      LettingsLog.where.not(old_id: nil).where(status: "in_progress").each do |lettings_log|
+        applicable_questions = lettings_log.form.subsections.map { |s| s.applicable_questions(lettings_log).select { |q| q.enabled?(lettings_log) } }.flatten
+        unanswered_questions = applicable_questions.filter { |q| q.unanswered?(lettings_log) }.map(&:id) - lettings_log.optional_fields
```
natdeanlewissoftwire: if we're always joining this by ", " let's just do that here instead of every time we refer to it?

natdeanlewissoftwire (approved):
nice one!

PR #1894, opened 2023-09-07
Set unassigned user if legacy user belongs to a different organisation

kosiakkatrina:
During the logs import, some of the users coming through do not belong to the owning or managing organisation of the log.
This is due to duplicate emails over several accounts and the ability to create logs for users in a different org via bulk upload (on old CORE).
We would be assigning these logs to a fake "Unassigned" user in order to import as much data as possible.
We don't want to keep these logs without a created_by set, as that would clear most of the data from other sections


natdeanlewissoftwire (approved):
nice one!

PR #1896, opened 2023-09-07
CLCD-2699 Add reimport reason case

kosiakkatrina:
We have recently added more reason options for renewals (https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/1876) and want to reimport any logs that had those values selected

### DIFF
--- a/app/services/imports/lettings_logs_field_import_service.rb
+++ b/app/services/imports/lettings_logs_field_import_service.rb
@@ -14,6 +14,8 @@ module Imports
         import_from(folder, :update_creation_method)
       when "address"
         import_from(folder, :update_address)
+      when "reason"
+        import_from(folder, :update_reason)
       else
         raise "Updating #{field} is not supported by the field import service"
       end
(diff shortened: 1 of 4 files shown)


app/services/imports/lettings_logs_field_import_service.rb
```diff
@@ -176,5 +178,30 @@ def postcode_known(record)
...
+      record = LettingsLog.find_by(old_id:)
+
+      if record.present?
+        if record.reason.present?
+          @logger.info("lettings log #{record.id} has a value for reason, skipping update")
+        else
+          reason = unsafe_string_as_integer(xml_doc, "Q9a")
+          reasonother = string_or_nil(xml_doc, "Q9aa")
+          if reason == 20 && reasonother.blank?
+            @logger.info("lettings log #{record.id}'s reason is other but other reason is not provided, skipping update")
```
natdeanlewissoftwire, line 195: why don't we want to import in this case?

natdeanlewissoftwire (approved):

app/services/imports/lettings_logs_field_import_service.rb
kosiakkatrina: It'd trow an error if we selected other reason but don't provide it

PR #1899, opened 2023-09-08
CLDC-2768: Sidekiq (AWS)

Chirag-Bhatti:
- Update AWS staging deploy pipeline to deploy sidekiq
- Once approved, I'll raise a duplicate PR against `main` after the `AwsStagingDeploy` has been merged into it

### DIFF
--- a/.github/workflows/aws_deploy.yml
+++ b/.github/workflows/aws_deploy.yml
@@ -135,26 +135,12 @@ jobs:
         run: |
           aws ecs describe-task-definition --task-definition $app_task_definition --query taskDefinition > app-task-definition.json
 
-      - name: Download sidekiq service task definition
-        env:
-          app_task_definition: ${{ inputs.aws_resource_prefix }}-sidekiq
-        run: |
-          aws ecs describe-task-definition --task-definition $app_task_definition --query taskDefinition > sidekiq-task-definition.json
-
       - name: Update app image ID
         id: app-task-def
         uses: aws-actions/amazon-ecs-render-task-definition@v1
         with:
           task-definition: app-task-definition.json
           container-name: app
-          image: ${{ env.image }}
-
-      - name: Update sidekiq image ID
-        id: sidekiq-task-def
-        uses: aws-actions/amazon-ecs-render-task-definition@v1
-        with:
-          task-definition: sidekiq-task-definition.json
-          container-name: sidekiq
           image: ${{ env.image }}
 
       - name: Deploy updated application
@@ -164,6 +150,20 @@ jobs:
           service: ${{ inputs.aws_resource_prefix }}-app
           task-definition: ${{ steps.app-task-def.outputs.task-definition }}
           wait-for-service-stability: true
+
+      - name: Download sidekiq service task definition
+        env:
+          sidekiq_task_definition: ${{ inputs.aws_resource_prefix }}-sidekiq
+        run: |
+          aws ecs describe-task-definition --task-definition $sidekiq_task_definition --query taskDefinition > sidekiq-task-definition.json
+
+      - name: Update sidekiq image ID
+        id: sidekiq-task-def
+        uses: aws-actions/amazon-ecs-render-task-definition@v1
+        with:
+          task-definition: sidekiq-task-definition.json
+          container-name: sidekiq
+          image: ${{ env.image }}
 
       - name: Deploy updated sidekiq
         uses: aws-actions/amazon-ecs-deploy-task-definition@v1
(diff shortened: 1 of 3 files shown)


.github/workflows/aws_deploy.yml
Chirag-Bhatti, line 1: Generally I think whenever the application code changes, we have to deploy a new application image and then a new sidekiq image, at least for now (since Sidekiq is in the application code)

Chirag-Bhatti: I've used the same application image for running the app and sidekiq in ECS, but with different task definitions (e.g. the entrypoint, need for port mapping and set of env vars required are different)

Chirag-Bhatti: See [this PR](https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data-infrastructure/pull/29) for more info about the AWS infrastructure changes

.github/workflows/aws_deploy.yml
```diff
@@ -129,24 +129,46 @@ jobs:
...
 
-      - name: Download service task definition
+      - name: Download app service task definition
         env:
           app_task_definition: ${{ inputs.aws_resource_prefix }}-app
         run: |
           aws ecs describe-task-definition --task-definition $app_task_definition --query taskDefinition > app-task-definition.json
 
-      - name: Update image ID
+      - name: Download sidekiq service task definition
```
RachaelBooth: I think this pipeline would make more sense conceptually if it did all the app stuff and then all the sidekiq stuff -and neither of these steps take long enough to need to do all the prep then all the deploy.

.github/workflows/aws_deploy.yml
```diff
@@ -129,24 +129,46 @@ jobs:
...
           task-definition: ${{ steps.app-task-def.outputs.task-definition }}
           wait-for-service-stability: true
+
+      - name: Deploy updated sidekiq
+        uses: aws-actions/amazon-ecs-deploy-task-definition@v1
+        with:
+          cluster: ${{ inputs.aws_resource_prefix }}
+          service: ${{ inputs.aws_resource_prefix }}-sidekiq
+          task-definition: ${{ steps.sidekiq-task-def.outputs.task-definition }}
+          wait-for-service-stability: true
```
RachaelBooth, line 174: If this step fails, we then need manual intervention of some sort (whether that's re-running or rolling back) to ensure that sidekiq and the app are running the same version. This is fine, but when it comes to writing the runbook we should ensure that it's a documented thing to handle.

RachaelBooth (commented):
Small ordering comment, otherwise looks good to me.

.github/workflows/aws_deploy.yml
```diff
@@ -129,24 +129,46 @@ jobs:
...
+      - name: Download app service task definition
         env:
           app_task_definition: ${{ inputs.aws_resource_prefix }}-app
         run: |
           aws ecs describe-task-definition --task-definition $app_task_definition --query taskDefinition > app-task-definition.json
 
-      - name: Update image ID
+      - name: Download sidekiq service task definition
+        env:
+          app_task_definition: ${{ inputs.aws_resource_prefix }}-sidekiq
```
SamSeed-Softwire: Do you think we should maybe rename this to sidekiq_task_definition to avoid confusion?

.github/workflows/aws_deploy.yml
SamSeed-Softwire: @Chirag-Bhatti maybe add a placeholder in the runbook now to ensure it doesn't get forgotten?

.github/workflows/aws_deploy.yml
Chirag-Bhatti: Agree! I'll update the order 👍

.github/workflows/aws_deploy.yml
Chirag-Bhatti: Yes good spot! I'll update this 👍

.github/workflows/aws_deploy.yml
SamSeed-Softwire: Sorry that sounded quite micromanaging...

Chirag-Bhatti: No problem! I've added a placeholder in the runbook [here](https://dluhcdigital.atlassian.net/wiki/spaces/MC/pages/32441141/PLACEHOLDER+Handling+incomplete+deployments+from+the+app+pipeline) with some very pre-emptive text :-)

RachaelBooth (approved):
LGTM!

May as well merge this to the AwsStagingDeploy branch - we'll merge the lot to main once we've got https sorted.

PR #1902, opened 2023-09-11
Only blank tenancy length when it is invalid for tenancy type

kosiakkatrina:
When importing logs only blank tenancy length when it is invalid for tenancy type
We want to keep tenancy type to make it clearer what the import issues are

### DIFF
--- a/spec/services/imports/lettings_logs_import_service_spec.rb
+++ b/spec/services/imports/lettings_logs_import_service_spec.rb
@@ -325,7 +325,7 @@ RSpec.describe Imports::LettingsLogsImportService do
             .not_to raise_error
         end
 
-        it "clears out the invalid tenancy lenght" do
+        it "clears out the invalid tenancy length" do
           allow(logger).to receive(:warn)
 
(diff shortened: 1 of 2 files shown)


spec/services/imports/lettings_logs_import_service_spec.rb
```diff
@@ -325,15 +325,15 @@ def open_file(directory, filename)
             .not_to raise_error
         end
 
-        it "clears out the invalid answers" do
+        it "clears out the invalid tenancy lenght" do
```
natdeanlewissoftwire:
```suggestion
        it "clears out the invalid tenancy length" do
```

natdeanlewissoftwire: :)

natdeanlewissoftwire (approved):

spec/services/imports/lettings_logs_import_service_spec.rb
kosiakkatrina: (:

PR #1903, opened 2023-09-11
Move person details if some persons are skipped

kosiakkatrina:
There is no validations around the order in which person details can be entered on old core.
So if the household members count is set to 2 but details only for person 1 and 3 are given in the import, we currently set the details for person 1 but ignore the data for person 3 (as there are only 2 household members)

Instead we want to take the person 3 data and add it to person 2 on our service (but only if person 2 details are not given)

The way we do that is by getting the indexes of all the persons with data provided in old service and mapping them to the lowest available person index on the new service. 

We consider person with either of these fields given as having person data: age, gender, working situation, relationship to tenant 1

We always expect tenant 1 to have person data

### DIFF
--- a/app/services/imports/lettings_logs_import_service.rb
+++ b/app/services/imports/lettings_logs_import_service.rb
@@ -77,13 +77,15 @@ module Imports
       attributes["rent_type"] = rent_type(xml_doc, attributes["lar"], attributes["irproduct"])
       attributes["hhmemb"] = household_members(xml_doc, previous_status)
       (1..8).each do |index|
-        attributes["age#{index}"] = safe_string_as_integer(xml_doc, "P#{index}Age")
-        attributes["age#{index}_known"] = age_known(xml_doc, index, attributes["hhmemb"])
-        attributes["sex#{index}"] = sex(xml_doc, index)
-        attributes["ecstat#{index}"] = unsafe_string_as_integer(xml_doc, "P#{index}Eco")
+        person_index = people_with_details_ids(xml_doc)[index - 1] || index
+        attributes["age#{index}"] = safe_string_as_integer(xml_doc, "P#{person_index}Age")
+        attributes["age#{index}_known"] = age_known(xml_doc, index, person_index, attributes["hhmemb"])
+        attributes["sex#{index}"] = sex(xml_doc, person_index)
+        attributes["ecstat#{index}"] = unsafe_string_as_integer(xml_doc, "P#{person_index}Eco")
       end
       (2..8).each do |index|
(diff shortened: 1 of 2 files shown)


spec/services/imports/lettings_logs_import_service_spec.rb
```diff
@@ -1282,6 +1282,143 @@ def open_file(directory, filename)
...
+            lettings_log_service.send(:create_log, lettings_log_xml)
+
+            lettings_log = LettingsLog.where(old_id: lettings_log_id).first
+            expect(lettings_log&.hhmemb).to eq(2)
+            expect(lettings_log&.details_known_2).to eq(0)
+            expect(lettings_log&.age2_known).to eq(0)
+            expect(lettings_log&.age2).to eq(7)
+            expect(lettings_log&.sex2).to eq("M")
+            expect(lettings_log&.relat2).to eq("C")
+          end
```
natdeanlewissoftwire, line 1326: can we add checks that the values for later people are nil in these tests (especially P3 in this case but later ones too maybe) for future proofing?

app/services/imports/lettings_logs_import_service.rb
```diff
@@ -77,13 +77,15 @@ def create_log(xml_doc)
       attributes["rent_type"] = rent_type(xml_doc, attributes["lar"], attributes["irproduct"])
       attributes["hhmemb"] = household_members(xml_doc, previous_status)
       (1..8).each do |index|
-        attributes["age#{index}"] = safe_string_as_integer(xml_doc, "P#{index}Age")
-        attributes["age#{index}_known"] = age_known(xml_doc, index, attributes["hhmemb"])
-        attributes["sex#{index}"] = sex(xml_doc, index)
-        attributes["ecstat#{index}"] = unsafe_string_as_integer(xml_doc, "P#{index}Eco")
+        person_index = people_with_details_ids(xml_doc)[index - 1] || index
```
natdeanlewissoftwire: won't this `|| index` condition mean e.g. if we have details for people 1 and 3 given, then we return person 3's xml details for both the second AND third pass of this loop?

kosiakkatrina: This is a very good point, updated it now and changed the test a bit as well. Thanks!

spec/services/imports/lettings_logs_import_service_spec.rb
kosiakkatrina: Sure, added a couple of cases/tests

natdeanlewissoftwire (approved):

PR #1904, opened 2023-09-11
Make saledate nil safe(r)

natdeanlewissoftwire:
We need to allow missing day/month/year values if CompletionDate exists as we don't always receive these from old core.

### DIFF
--- a/app/services/imports/logs_import_service.rb
+++ b/app/services/imports/logs_import_service.rb
@@ -72,7 +72,7 @@ module Imports
       if str.nil?
         nil
       else
-        BigDecimal(str, exception: false).round(2)
+        BigDecimal(str, exception: false)&.round(2)
       end
     end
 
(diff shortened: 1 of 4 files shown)


app/services/imports/logs_import_service.rb
```diff
@@ -72,7 +72,7 @@ def safe_string_as_decimal(xml_doc, attribute)
       if str.nil?
         nil
       else
-        BigDecimal(str, exception: false).round(2)
+        BigDecimal(str, exception: false)&.round(2)
```
natdeanlewissoftwire, line 75: this is an unrelated change but as it's on theme with nil safeness I thought I'd include it, it's for another migration log import failure I saw

kosiakkatrina: Can we add a failing test for this?

kosiakkatrina (approved):

app/services/imports/logs_import_service.rb
natdeanlewissoftwire: We sure can

PR #1906, opened 2023-09-11
Mark hodate value check confirmed

kosiakkatrina:
We mark all soft validations as confirmed upon import, this one was missed


natdeanlewissoftwire (approved):
Have you done a check that no other soft vals were misseed too?

kosiakkatrina:
> Have you done a check that no other soft vals were misseed too?

@natdeanlewissoftwire very good point, there were a few, added them in now. Lettings looks correct. Thanks!

PR #1908, opened 2023-09-12
Only set soc tenant fields for 2022 or if they're given

kosiakkatrina:
In old CORE, the question "Was the buyer a private registered provider, housing association or local authority tenant immediately before this sale?" doesn't exist. The questions that follow that are conditional on this question (`frombeds`, `fromprop`, `socprevten`) are optional. We didn't want to make them answer it in 22/23 if they hadn't already, therefore for 22/23, we have a secret "don't know" option that we mark as the answer to this, as well as `fromprop` and `socprevten`, and then we make `frombeds` optional (if they're not answered in old CORE)

For 23/24, we do want to get them to answer this question

We remove this inferring for 23/24, so that if none of the subsequent questions are answered (`frombeds`, `fromprop`, `socpreveten`) then "Was the buyer a private registered provider, housing association or local authority tenant immediately before this sale?" is blank and needs to be answered, and we don't have the "don't know" options available.

### DIFF
--- a/spec/services/imports/sales_logs_import_service_spec.rb
+++ b/spec/services/imports/sales_logs_import_service_spec.rb
@@ -384,6 +384,42 @@ RSpec.describe Imports::SalesLogsImportService do
           sales_log = SalesLog.find_by(old_id: sales_log_id)
           expect(sales_log.proplen_asked).to eq(1)
         end
+
+        context "when setting soctenant fields" do
+          it "does not set soctenant value if none of the soctenant questions are answered" do
+            sales_log_xml.at_xpath("//xmlns:Q20Bedrooms").content = nil
+            sales_log_xml.at_xpath("//xmlns:PrevRentType").content = nil
+            sales_log_xml.at_xpath("//xmlns:Q21PropertyType").content = nil
+
+            expect(logger).not_to receive(:error)
+            expect(logger).not_to receive(:warn)
+            expect(logger).not_to receive(:info)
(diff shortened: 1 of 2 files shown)


spec/services/imports/sales_logs_import_service_spec.rb
```diff
@@ -384,6 +384,42 @@ def open_file(directory, filename)
           sales_log = SalesLog.find_by(old_id: sales_log_id)
           expect(sales_log.proplen_asked).to eq(1)
         end
+
+        context "when setting soctenant fields" do
```
natdeanlewissoftwire, line 388: I think we should test the different behaviour in the different years

kosiakkatrina: There are tests for 2022, the behaviour for 2022 stayed the same, so I kept those tests and added some for 2023

kosiakkatrina: But let me know if there are cases that I missed

natdeanlewissoftwire: ahh ok gotcha, nice!

natdeanlewissoftwire (approved):
I assume the "we don't have the don't know options available" in 2023 from the description was already included in another PR?

kosiakkatrina:
> I assume the "we don't have the don't know options available" in 2023 from the description was already included in another PR?

@natdeanlewissoftwire Yup! It was never available to select from the UI, as it's not on the displayed answer options list! (only on the answer options)

PR #1910, opened 2023-09-12
CLDC-2776 Add referral option to general needs logs

kosiakkatrina:
Referral option 4 should only be applicable to supported housing logs, but there are legitimate scenarios where it is applicable to general needs. We therefore want this option to display for general needs (PRP only) but have a soft validation when it is chosen:
image


config/locales/en.yml
```diff
@@ -706,6 +706,9 @@ Make sure these answers are correct."
     mortgage:
       title_text: "You told us that the mortgage amount is %{mortgage}"
       hint_text: "This is more than 5 times the income, which is higher than we would expect."
+    referral:
+      title_text: "Are you sure?"
+      hint_text: "This is a general needs log, and this referral type is for supported housing."
```
Airk0n, line 711: I'm guessing this comes from content, but I can imagine users reading this and changing their answer when it was actually applicable.

Airk0n (approved):
Looks good to me

config/locales/en.yml
kosiakkatrina: They have an option to confirm it's correct and I guess the point is to bring their attention to the fact that it looks off, but I see your point

PR #1911, opened 2023-09-13
Set income and savings not known for in progress logs

kosiakkatrina:
If P1IncKnown is not answered on old core but income value is given, we still don't set income1nk.
Same for savingsKnown and P2IncKnown - we don't set income2nk and savingsnk if those questions aren't answered and if the status for the log is saved(in progress)

If it was completed/submitted we would set a default value for income1nk and savingsnk.

Now we want to set the default value for these fields for in progress logs as well, but only if the value for savings and income is given


Airk0n (approved):
Looks good to me

PR #1912, opened 2023-09-13
CLDC-2795 Update charges validations

kosiakkatrina:
A lot of the imported logs cross the hard maximum for the charges and we lose all the charges data
This PR updates hard validations for scharge, pscharge and supcharg fields with higher maximums and adds soft validations for those fields:
image


app/models/form/lettings/questions/supcharg_value_check.rb
```diff
@@ -0,0 +1,15 @@
+class Form::Lettings::Questions::SupchargValueCheck < ::Form::Question
```
natdeanlewissoftwire, line 1: Not related to this PR but I hope there's an intelligent reason there's no e on the end of supcharg :(

config/forms/2022_2023.json
```diff
@@ -8321,6 +8321,141 @@
...
+                      {
+                        "scharge_value_check": 0
+                      },
+                      {
+                        "scharge_value_check": 1
+                      }
+                    ]
+                  },
+                  "header": "Are you sure?",
+                  "hint_text": "This is higher than we would expect. Check:<ul class=\"govuk-body-l app-panel--interruption\"><li>the decimal point</li><li>the frequency, for example every week or every calendar month</li><li>the rent type is correct, for example affordable or social rent</li></ul>",
```
natdeanlewissoftwire: has this copy been checked? seems weird to have "is correct" only in the third bullet point

natdeanlewissoftwire (approved):
is there a separate ticket already made for setting these soft vals as answered for previously migrated logs or is that part of this one?

app/models/validations/soft_validations.rb
```diff
@@ -97,6 +97,42 @@ def net_income_higher_or_lower_text
     net_income_in_soft_max_range? ? "higher" : "lower"
   end
 
+  def scharge_over_soft_max?
+    return unless scharge && period && needstype
+    return if weekly_value(scharge).blank?
+
+    max = if needstype == 1
```
natdeanlewissoftwire, line 104: are there any situations where nil values here could exist for needstype or provider type? I don't think so but if that does happen we could get some incorrect maxes

app/models/form/lettings/questions/supcharg_value_check.rb
kosiakkatrina: No idea, it's been a source of entertainment as I tried my best to spell it correctly every time

app/models/validations/soft_validations.rb
kosiakkatrina: If needstype isn't answered we wouldn't route to the soft validation anyway, but good point about provider type. Let me check that

config/forms/2022_2023.json
kosiakkatrina: copied the copy wrong... Will fix!

natdeanlewissoftwire (approved):

kosiakkatrina:
> is there a separate ticket already made for setting these soft vals as answered for previously migrated logs or is that part of this one?

@natdeanlewissoftwire That would be the same ticket but a separate PR, don't have that yet tho

PR #1913, opened 2023-09-13
Migration use kebab org fields

natdeanlewissoftwire:
We need to use the `owner-institution-id` and `managing-institution-id` fields from migration xmls instead of `OWNINGORGID` and `MANINGORGID` as they are more accurate on old Core and the discrpeancy was causing occasional issues in the 5.42 migration

### DIFF
--- a/spec/fixtures/imports/logs/00d2343e-d5fa-4c89-8400-ec3854b0f2b4.xml
+++ b/spec/fixtures/imports/logs/00d2343e-d5fa-4c89-8400-ec3854b0f2b4.xml
@@ -4,7 +4,7 @@
     <meta:document-id>00d2343e-d5fa-4c89-8400-ec3854b0f2b4</meta:document-id>
     <meta:owner-user-id>c3061a2e6ea0b702e6f6210d5c52d2a92612d2aa</meta:owner-user-id>
     <meta:owner-institution-id>7c5bd5fb549c09a2c55d7cb90d7ba84927e64618</meta:owner-institution-id>
-    <meta:managing-institution-id>7c5bd5fb549c09a2c55d7cb90d7ba84927e64618</meta:managing-institution-id>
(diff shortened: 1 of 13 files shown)


Airk0n (approved):
Looks good to me, but I would consider another opinion as I am quite inexperienced with migrations.

spec/fixtures/imports/logs/00d2343e-d5fa-4c89-8400-ec3854b0f2b4.xml
```diff
@@ -4,7 +4,7 @@
     <meta:document-id>00d2343e-d5fa-4c89-8400-ec3854b0f2b4</meta:document-id>
     <meta:owner-user-id>c3061a2e6ea0b702e6f6210d5c52d2a92612d2aa</meta:owner-user-id>
     <meta:owner-institution-id>7c5bd5fb549c09a2c55d7cb90d7ba84927e64618</meta:owner-institution-id>
-    <meta:managing-institution-id>7c5bd5fb549c09a2c55d7cb90d7ba84927e64618</meta:managing-institution-id>
+    <meta:managing-institution-id>7c5bd5fb549c09z2c55d9cb90d7ba84927e64618</meta:managing-institution-id>
```
kosiakkatrina, line 7: What is the change here? (?)

kosiakkatrina (approved):

spec/fixtures/imports/logs/00d2343e-d5fa-4c89-8400-ec3854b0f2b4.xml
natdeanlewissoftwire: a2c55d7 instead of z2c55d9, for some reason these different ids used for different orgs but it's not been picked up before as we've been using the other fields to distinguish them

PR #1915, opened 2023-09-14
CLDC-2795 Confirm scharge, pscharge and supcharge soft validations

kosiakkatrina:
We've added new soft validations, but would like to keep the previously completed logs as completed. 
This PR is to confirm those soft validations for completed logs

### DIFF
--- a/lib/tasks/confirm_charges_soft_validations.rake
+++ b/lib/tasks/confirm_charges_soft_validations.rake
@@ -1,0 +1,8 @@
+desc "Confirms scharge, pscharge and supcharge soft validations for completed logs"
+task confirm_charges_soft_validations: :environment do
+  LettingsLog.where(status: "completed").each do |log|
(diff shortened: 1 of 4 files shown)


lib/tasks/confirm_charges_soft_validations.rake
```diff
@@ -0,0 +1,8 @@
+desc "Confirms scharge, pscharge and supcharge soft validations for completed logs"
+task confirm_charges_soft_validations: :environment do
+  LettingsLog.where(status: "completed").each do |log|
+    log.update!(scharge_value_check: 0, values_updated_at: Time.zone.now) if log.scharge_value_check.blank? && log.scharge_over_soft_max?
```
natdeanlewissoftwire, line 4: I forgot 0 is our yes value 😢

natdeanlewissoftwire (approved):
lgtm :)

PR #1917, opened 2023-09-14
CLDC-2715 Reimport homeless values

kosiakkatrina:
Reimport homeless values if they failed to import in the past due to this validation that is now removed. https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/1901
https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/1877

### DIFF
--- a/app/services/imports/lettings_logs_field_import_service.rb
+++ b/app/services/imports/lettings_logs_field_import_service.rb
@@ -16,6 +16,8 @@ module Imports
         import_from(folder, :update_address)
       when "reason"
         import_from(folder, :update_reason)
+      when "homeless"
+        import_from(folder, :update_homelessness)
       else
         raise "Updating #{field} is not supported by the field import service"
       end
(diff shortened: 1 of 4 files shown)


app/services/imports/lettings_logs_field_import_service.rb
```diff
@@ -203,5 +205,30 @@ def update_reason(xml_doc)
...
+
+    def update_homelessness(xml_doc)
+      return if meta_field_value(xml_doc, "form-name").include?("Sales")
+
+      old_id = meta_field_value(xml_doc, "document-id")
+      record = LettingsLog.find_by(old_id:)
+      if record.present?
+        return @logger.info("lettings log #{record.id} has a value for homeless and rp_homeless, skipping update") if record.rp_homeless == 1 && record.homeless.present?
+        return @logger.info("lettings log #{record.id} has a value for homeless and reasonpref is not yes, skipping update") if record.reasonpref != 1 && record.homeless.present?
+        return @logger.info("lettings log #{record.id} reimport values are not homeless - 11 and rp_homeless - yes, skipping update") if unsafe_string_as_integer(xml_doc, "Q14b1").blank? || unsafe_string_as_integer(xml_doc, "Q13") != 11
```
natdeanlewissoftwire, line 217: this line says it checks if rp_homeless is yes but checks if it's blank instead - is that intentional?

natdeanlewissoftwire: I'm guessing it's a 1 or nil situation?

natdeanlewissoftwire (approved):

app/services/imports/lettings_logs_field_import_service.rb
```diff
@@ -203,5 +205,30 @@ def update_reason(xml_doc)
...
+
+        if record.rp_homeless != 1 && record.reasonpref == 1
+          record.rp_homeless = 1
+          @logger.info("updating lettings log #{record.id}'s rp_homeless value to 1")
+        end
+        if record.homeless.blank?
+          record.homeless = 11
+          @logger.info("updating lettings log #{record.id}'s homeless value to 11")
+        end
+        record.values_updated_at = Time.zone.now
```
natdeanlewissoftwire, line 227: if both of the above if conditions are false, we're not updating any values so I'm not sure we should set this in that case

app/services/imports/lettings_logs_field_import_service.rb
kosiakkatrina: Yeah, basically if this field has anything in it we set rp_homeless to true/checked

kosiakkatrina: We would not set it if it's blank, we'd return

app/services/imports/lettings_logs_field_import_service.rb
kosiakkatrina: I'm not sure it's possible to get to this point and neither of these if statements to get hit:
If record.homeless is present and rp_homeless is 1 (reasonpref would have to be 1 in order for rp_homeless to be 1 because of the routing) we'd return at the top, but let me know if I missed something

PR #1918, opened 2023-09-14
CLDC-2748 Update min value for age

kosiakkatrina:
The min value for age should be 1, but was 0 in 2022. It's correct in 2023.
This PR fixes the min value for 2022 and has a rake task to fix existing ages to 1 instead of 0


lib/tasks/correct_min_age_values.rake
```diff
@@ -0,0 +1,10 @@
+desc "Alter age values from 0 to 1"
```
natdeanlewissoftwire, line 1: I'm assuming this is because there are too many logs with age 0 to get their owners to fix them intentionally? If 0's an invalid answer now then surely the logs that have 0 have that erroneously so the correct value might not be 1?

natdeanlewissoftwire (approved):

lib/tasks/correct_min_age_values.rake
kosiakkatrina: I think we have checked that all of these records are children, and that's why we're happy to update these

PR #1919, opened 2023-09-14
CLDC-2769 Reimport unassigned logs

kosiakkatrina:
There is a number of logs that have imported with missing created_by values. 
Logs without created_by would in most cases have cleared all the data for non setup sections (provided that the before hooks have ran), due to clearing out the non routed questions.

When importing logs we have now started to assign all the logs that cannot be assigned to a real user, to a fake, deactivated user with name "Unassigned"
As most of the data (anything past the setup section) cannot be edited unless the log is assigned to someone, it is relatively safe to delete and reimport all the logs that do not have created_by value set

### DIFF
--- a/app/services/imports/lettings_logs_field_import_service.rb
+++ b/app/services/imports/lettings_logs_field_import_service.rb
@@ -18,6 +18,8 @@ module Imports
         import_from(folder, :update_reason)
       when "homeless"
         import_from(folder, :update_homelessness)
+      when "created_by"
+        import_from(folder, :update_created_by)
       else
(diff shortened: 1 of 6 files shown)


natdeanlewissoftwire:
> ...it is relatively safe to delete and reimport all the users that do not have created_by value set


this should be "all the logs" right?

natdeanlewissoftwire (approved):
lgtm

kosiakkatrina:
> > ...it is relatively safe to delete and reimport all the users that do not have created_by value set
> 
> this should be "all the logs" right?

Yikes, yes, deffo, fixed now

PR #1923, opened 2023-09-15
CLDC-2729 Reimport referral for general needs PRP

kosiakkatrina:
We were previously not able to import referral value 4 for PRP General Needs logs, because it was not a valid option.
We since have added that option and added a soft validation if it was selected for General Needs PRP.
This PR reimports referral value which would have been previously dropped.


natdeanlewissoftwire (approved):
lgtm!

PR #1925, opened 2023-09-18
CLDC-2790 Reimport person details

kosiakkatrina:
There is no validation on order in which person details are entered on old core, we changed how we're importing person details here: https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/1903
This PR is to fix any missing data for previous imports

### DIFF
--- a/app/services/imports/lettings_logs_field_import_service.rb
+++ b/app/services/imports/lettings_logs_field_import_service.rb
@@ -267,7 +267,7 @@ module Imports
 
       (2..record.hhmemb).each do |i|
         next @logger.info("lettings log #{record.id} has values for sex#{i} and relat#{i}, skipping update") if record["sex#{i}"] && record["relat#{i}"]
-        next @logger.info("lettings log #{record.id} has value 'no' for details_known_#{i}, skipping update") if record["details_known_#{i}"] == 1
+        next @logger.info("lettings log #{record.id} has value 'no' for details_known_#{i}, skipping update") if record.details_not_known_for_person?(i)
 
         if record["sex#{i}"].blank?
           record["sex#{i}"] = sex(xml_doc, i)
@@ -311,7 +311,7 @@ module Imports
       return @logger.warn("lettings log with old id #{old_id} not found") unless record
       return @logger.info("lettings log #{record.id} has no hhmemb, skipping update") unless record.hhmemb
 
-      if (2..record.hhmemb).all? { |person_index| record.has_any_person_details?(person_index) || record["details_known_#{person_index}"] == 1 }
+      if (2..record.hhmemb).all? { |person_index| record.has_any_person_details?(person_index) || record.details_not_known_for_person?(person_index) }
         return @logger.info("lettings log #{record.id} has all household member details, skipping update")
       end
       if record.hhmemb == 8 || ((record.hhmemb + 1)..8).none? { |person_index| file_contains_person_details?(xml_doc, person_index) }
@@ -323,14 +323,14 @@ module Imports
 
       while person_exists_on_the_log?(record, person_index)
         if person_exists_on_the_log?(record, next_person_index)
-          if record.has_any_person_details?(person_index) || record["details_known_#{person_index}"] == 1
+          if record.has_any_person_details?(person_index) || record.details_not_known_for_person?(person_index)
             @logger.info("lettings log #{record.id} has details for person #{person_index}, skipping person")
             person_index += 1
             next_person_index += 1
             next
           end
 
-          if !record.has_any_person_details?(next_person_index) && record["details_known_#{next_person_index}"] != 1
+          if !record.has_any_person_details?(next_person_index) && !record.details_not_known_for_person?(next_person_index)
             next_person_index += 1
             next
           end
--- a/app/services/imports/lettings_logs_field_import_service.rb
+++ b/app/services/imports/lettings_logs_field_import_service.rb
@@ -24,6 +24,8 @@ module Imports
         import_from(folder, :update_sex_and_relat)
       when "general_needs_referral"
         import_from(folder, :update_general_needs_referral)
+      when "person_details"
+        import_from(folder, :update_person_details)
       else
         raise "Updating #{field} is not supported by the field import service"
       end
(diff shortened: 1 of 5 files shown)


app/services/imports/lettings_logs_field_import_service.rb
```diff
@@ -203,5 +205,103 @@ def update_reason(xml_doc)
...
+    def update_person_details(xml_doc)
+      return if meta_field_value(xml_doc, "form-name").include?("Sales")
+
+      old_id = meta_field_value(xml_doc, "document-id")
+      record = LettingsLog.find_by(old_id:)
+
+      return @logger.warn("lettings log with old id #{old_id} not found") unless record
+      return @logger.info("lettings log #{record.id} has no hhmemb, skipping update") unless record.hhmemb
+
+      if (2..record.hhmemb).all? { |person_index| record.has_any_person_details?(person_index) || record["details_known_#{person_index}"] == 1 }
```
natdeanlewissoftwire: sort of related, do we want the people with **known** details to be the lower number people, or do we not mind as long as that question is answered?

app/services/imports/lettings_logs_field_import_service.rb
```diff
@@ -203,5 +205,103 @@ def update_reason(xml_doc)
...
+        return @logger.info("lettings log #{record.id} has all household member details, skipping update")
+      end
+      if record.hhmemb == 8 || (record.hhmemb + 1..8).none? { |person_index| file_contains_person_details?(xml_doc, person_index) }
```
natdeanlewissoftwire: can you put record.hhemb + 1 in brackets so it's clearer it's a lower bound of this array or is that not linty? I think it's quite tricky to read because it at first seems like we're adding record.hhemb to 1..8 but if this is standard notation and I've just not come across it then fine :D

app/services/imports/lettings_logs_field_import_service.rb
```diff
@@ -203,5 +205,103 @@ def update_reason(xml_doc)
...
+          record["ecstat#{person_index}"] = unsafe_string_as_integer(xml_doc, "P#{next_person_index}Eco")
+          record["relat#{person_index}"] = relat(xml_doc, next_person_index)
+          record["details_known_#{person_index}"] = details_known(person_index, record)
+          @logger.info("lettings log #{record.id}, reimported person #{person_index} details")
+          person_index += 1
+          next_person_index += 1
+        end
+      end
+
+      record.values_updated_at = Time.zone.now
```
natdeanlewissoftwire, line 347: I'd only set this if values have changed (unless we can guarantee that at this point but I'm not sure we can)

app/services/imports/lettings_logs_field_import_service.rb
```diff
@@ -203,5 +205,103 @@ def update_reason(xml_doc)
...
+        return @logger.info("lettings log #{record.id} has no additional household member details, skipping update")
+      end
+
+      person_index = 2
+      next_person_index = person_index + 1
+
+      while person_index <= record.hhmemb
+        if next_person_index <= record.hhmemb
+          if record.has_any_person_details?(person_index) || record["details_known_#{person_index}"] == 1
+            @logger.info("lettings log #{record.id} has details for person #{person_index}, skipping update")
```
natdeanlewissoftwire: maybe "skipping person" instead of "skipping update" ?

app/services/imports/lettings_logs_field_import_service.rb
```diff
@@ -203,5 +205,103 @@ def update_reason(xml_doc)
...
+        return @logger.info("lettings log #{record.id} has all household member details, skipping update")
+      end
+      if record.hhmemb == 8 || (record.hhmemb + 1..8).none? { |person_index| file_contains_person_details?(xml_doc, person_index) }
+        return @logger.info("lettings log #{record.id} has no additional household member details, skipping update")
+      end
+
+      person_index = 2
+      next_person_index = person_index + 1
+
+      while person_index <= record.hhmemb
```
natdeanlewissoftwire: GREAT loop but I feel like this might benefit from a brief comment explaining its approach, took me a while to work out what it's doing as it's a bit _clever_

natdeanlewissoftwire (approved):
nice one!

app/services/imports/lettings_logs_field_import_service.rb
kosiakkatrina: I don't think we mind. I feel like keeping them in as much order as what they came in as is probably best

app/services/imports/lettings_logs_field_import_service.rb
kosiakkatrina: Hmm, I think we can? Cause we only get it to hear if there are additional details for household members and there is at least 1 person missing details, which means the people would shuffle. LMK if I'm missing a case tho

app/services/imports/lettings_logs_field_import_service.rb
kosiakkatrina: Yeah, I like that, thanks!

app/services/imports/lettings_logs_field_import_service.rb
kosiakkatrina: I extracted some methods does it look any clearer now?

app/services/imports/lettings_logs_field_import_service.rb
natdeanlewissoftwire: Yep you're right :D

app/services/imports/lettings_logs_field_import_service.rb
natdeanlewissoftwire: Yeah much nicer, makes me think you could also extract record["details_known_#{person_index}"] == 1 as a method called details_not_known_for_person(index) so it's clearer that 1 == No

natdeanlewissoftwire (approved):

app/services/imports/lettings_logs_field_import_service.rb
kosiakkatrina: Yup, nice!

natdeanlewissoftwire (approved):

PR #1926, opened 2023-09-18
CLDC-2791 Reimport childrens social care referral values

kosiakkatrina:
Up until https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/1905 Children's social care was not an option for referral in all cases. We want to report this answer where we might have previously dropped it.


natdeanlewissoftwire (approved):
👍

PR #1927, opened 2023-09-18
Fix label content for support charge validation

kosiakkatrina:


natdeanlewissoftwire (approved):
assume this doesn't need changing in other years

kosiakkatrina:
> assume this doesn't need changing in other years

Should be correct in other years

PR #1932, opened 2023-09-20
CLDC-2770 Update lead tenants gender identity

kosiakkatrina:
Refused is also a valid option for lead tenant gender identity. Previously we were only reimporting values for tenants 2..8


natdeanlewissoftwire (approved):
nice one!

PR #1933, opened 2023-09-20
CLDC-2795 Only update 2022 and 2023 logs

kosiakkatrina:
Updating 2021 logs triggers the validation


natdeanlewissoftwire (approved):
worth adding a test for?

PR #1939, opened 2023-09-22
CLDC-2774 Update lettings materials

natdeanlewissoftwire:


kosiakkatrina (commented):
~do we not want to have 2023_34_lettings_paper_form at all?~  hold on, maybe I'm missing something. Are we just changing that file? The diff is confusing

natdeanlewissoftwire:
@kosiakkatrina yeah we are just changing all 4 files, I don't know why this new pdf is so much smaller than the previous - I also don't know why it's saying -100% or +100% as if it's losing or gaining the whole file size, it isn't.
But the files are correct from looking into this branch's public/files directory

EDIT: apparently 100% just means all the bytes have changed which I suppose is true, it's a new file with the same name

kosiakkatrina (approved):

PR #1942, opened 2023-09-26
CLDC-2576 Add duplicate log error if the chcharges are the same

kosiakkatrina:
We use total charge as one of the fields to determine duplicate logs. 
Not all households pay these charges and sometimes they pay care home charges instead, in those cases we should compare appropriate fields to determine duplicates.

This PR also sets "is_carehome" field for 2023/24 bulk upload depending on whether the care home charges are given. (this is being set correctly for 2022)
Previously this field was not being set, so if the household was paying care home charges, we would get "you need to answer rent and charges" errors.


natdeanlewissoftwire (approved):
lgtm!

PR #1943, opened 2023-09-26
CLDC-2772 Report on imported logs only

kosiakkatrina:
If we run the reports some time after the migration, there might be logs that have need created in the new service. We shouldn't report on those.


natdeanlewissoftwire (approved):
lgtm

PR #1945, opened 2023-09-26
CLDC-2772 Deduplicate logs on import

kosiakkatrina:
We have recently changed some validations to increase the number of logs migrated from old CORE.
These changes might now also allow some of the skipped logs from previous migrations to be reimported.

In order to avoid reimporting logs that might have already been recreated by the users on the new system we want to skip any of the logs that get marked as duplicates using the existing duplicate logs scope.

We only skip the log if there is a duplicate that was created on this service. If duplicates existed on old CORE, we will attempt to import both of the logs.

### DIFF
--- a/spec/services/imports/lettings_logs_import_service_spec.rb
+++ b/spec/services/imports/lettings_logs_import_service_spec.rb
@@ -103,6 +103,22 @@ RSpec.describe Imports::LettingsLogsImportService do
         expect(updated_logs).to eq(0)
       end
 
+      it "does not import the log if a duplicate log exists on the system" do
+        expect(logger).not_to receive(:error)
+        expect(logger).not_to receive(:warn)
+        expect(logger).not_to receive(:info).with(/Updating lettings log/)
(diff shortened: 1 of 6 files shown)


spec/services/imports/lettings_logs_import_service_spec.rb
```diff
@@ -103,6 +103,35 @@ def open_file(directory, filename)
         expect(updated_logs).to eq(0)
       end
 
+      it "does not import the log if a duplicate log exists on the system" do
```
natdeanlewissoftwire:
```suggestion
      it "does not import the log if a duplicate log exists on the system (that was not migrated from old CORE)" do
```

spec/services/imports/sales_logs_import_service_spec.rb
```diff
@@ -81,6 +81,36 @@ def open_file(directory, filename)
       expect(updated_logs).to eq(0)
     end
 
+    it "does not import the log if a duplicate log exists on the system" do
```
natdeanlewissoftwire: same suggestion as above :)

natdeanlewissoftwire (approved):

PR #1946, opened 2023-09-27
CLDC-2816 scheme filter bug 2

natdeanlewissoftwire:
Second PR on this branch as a separate schemes bug was found and is included in this ticket - we need to include confirmed: nil in our incomplete scopes. Have also added for locations though that's not strictly necessary for the ACs


app/models/location.rb
```diff
@@ -49,6 +49,7 @@ class Location < ApplicationRecord
 
   scope :incomplete, lambda {
     where.not(confirmed: true)
+      .or(where(confirmed: nil))
```
kosiakkatrina, line 52: Wouldn't nil not be true tho?

natdeanlewissoftwire: Yeah that's what I thought initially too, but this is how the bug didn't get caught earlier - ActiveRecord's where.not translates to a sql != which doesn't return null values - see https://thoughtbot.com/blog/activerecord-s-where-not-and-nil

kosiakkatrina: Oh yeah, okay. Good spot!

kosiakkatrina (approved):

PR #1950, opened 2023-09-29
Reimport lettings log form ID

kosiakkatrina:
Some of the packages did not contain form IDs where expected, we want to regenerate packages and reimport those form IDs.


natdeanlewissoftwire (approved):
This all looks good, but don't we also want to do it for sales?

kosiakkatrina:
> This all looks good, but don't we also want to do it for sales?

@natdeanlewissoftwire we have the same rake task for sales as we ran it recently (because we weren't importing form id until recently), so if we are missing any of the form IDs for sales we can run it as well

PR #1951, opened 2023-10-02
CLDC-2799 Lettings and Sales CSV export updates and sales bug fix

natdeanlewissoftwire:
This PR sets the correct "Not known" ore "Prefers not to say" labels and/or codes to sales and lettings logs. In particular, the age fields now contain this and then age_known fields are not included in the csvs. There's also now a higher (though more could be done here e.g. in the rejected fields) level of consistency between sales and lettings csv exports.

In this sense it incorporates ACs for [CLDC-2797](https://dluhcdigital.atlassian.net/browse/CLDC-2797) as well as [CLDC-2799](https://dluhcdigital.atlassian.net/browse/CLDC-2799)

[CLDC-2797]: https://dluhcdigital.atlassian.net/browse/CLDC-2797?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[CLDC-2799]: https://dluhcdigital.atlassian.net/browse/CLDC-2799?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


kosiakkatrina (approved):
looks good, I assume we want the lettings to be the same, but it's not explicitly stated (?), so make sure it gets tested

app/models/form/lettings/questions/person_working_situation.rb
```diff
@@ -29,6 +29,6 @@ def initialize(id, hsh, page, person_index:)
     },
     "0" => { "value" => "Other" },
     "divider" => { "value" => true },
-    "10" => { "value" => "Tenant prefers not to say" },
```
natdeanlewissoftwire, line 32: unrelated fix - this is not for the lead tenant so for consistency we should say person here as we do for the other extra person age/sex/relat questions

kosiakkatrina: They'd still be a tenant even if they're not lead tho, no?

natdeanlewissoftwire: Yeah Tenant wouldn't be wrong per se, it's just to inconsistent with other questions - it might be worth revisiting the naming to make sure they're all what we want them to be rather than just consistent with each other, I'll make a tech debt ticket for it but this is fine for now

PR #1952, opened 2023-10-02
CLDC-2810 Send missing addresses template

kosiakkatrina:
Due to several issues during migration, such as wrong uprn or missing address values, there is a lot of mandatory address data that is missing for several orgs.
In these cases the data either did not exist or was wrong on the old core, so we want to provide a way to more easily update it in bulk for identified logs.

If organisation logs fall into one of 3 issue categories we will send a spreadsheet with affected logs and address fields to be filled in and returned. 

The emails would be either sent to all active data coordinators, or to all active users if no coordinators exist


app/services/csv/missing_addresses_csv_service.rb
```diff
@@ -0,0 +1,165 @@
...
+                                .where("uprn = propcode OR uprn = tenancycode or town_or_city = 'Bristol'")
+                             end
+
+      return if logs_with_missing_addresses.empty? && logs_with_missing_town_or_city.empty? && logs_with_wrong_uprn.empty?
+
+      CSV.generate(headers: true) do |csv|
+        csv << ["Issue type", "Lettings log ID", "Tenancy start date", "Tenant code", "Property code", "Log owner", "Owning organisation name", "Managing organisation name", "UPRN", "Address line 1", "Address line 2 (optional)", "Town or City", "County (optional)", "Postcode"]
+
+        logs_with_missing_addresses.each do |log|
+          csv << ["Full address required",
```
natdeanlewissoftwire:
```suggestion
          csv << lettingslog_as_csv_row("Full address required")
```

app/services/csv/missing_addresses_csv_service.rb
```diff
@@ -0,0 +1,165 @@
...
+                  log.uprn,
+                  log.address_line1,
+                  log.address_line2,
+                  log.town_or_city,
+                  log.county,
+                  log.postcode_full]
+        end
+      end
+    end
+
```
natdeanlewissoftwire, line 46:
```suggestion
def lettings_log_as_csv_row(issue_type)
    [error_type,
                  log.id,
                  log.startdate&.to_date,
                  log.tenancycode,
                  log.propcode,
                  log.created_by&.email,
                  log.owning_organisation&.name,
                  log.managing_organisation&.name,
                  log.uprn,
                  log.address_line1,
                  log.address_line2,
                  log.town_or_city,
                  log.county,
                  log.postcode_full]
end
```

app/services/csv/missing_addresses_csv_service.rb
natdeanlewissoftwire: and similar for the other error types

app/services/csv/missing_addresses_csv_service.rb
```diff
@@ -0,0 +1,165 @@
...
+                                .where.not(uprn: nil)
+                                .where("uprn = purchid or town_or_city = 'Bristol'")
+                             end
+      return if logs_with_missing_addresses.empty? && logs_with_missing_town_or_city.empty? && logs_with_wrong_uprn.empty?
+
+      CSV.generate(headers: true) do |csv|
+        csv << ["Issue type", "Sales log ID", "Sale completion date", "Purchaser code", "Log owner", "Owning organisation name", "UPRN", "Address line 1", "Address line 2 (optional)", "Town or City", "County (optional)", "Postcode"]
+
+        logs_with_missing_addresses.each do |log|
+          csv << ["Full address required",
```
natdeanlewissoftwire: A similar sales_log_as_csv_row method could be used here :)

lib/tasks/send_missing_addresses_csv.rake
```diff
@@ -0,0 +1,78 @@
+namespace :correct_addresses do
+  # rubocop:disable Lint/ConstantDefinitionInBlock
+  MISSING_ADDRESSES_THRESHOLD = 50
```
natdeanlewissoftwire: can we just define this in lower case to avoid the lint disabling?
Or define it in the MissingAddressesCSVService/EmailMissingAddressesCsvJob instead?

lib/tasks/send_missing_addresses_csv.rake
```diff
@@ -0,0 +1,78 @@
...
+
+      logs_impacted_by_uprn_issue = if JSON.parse(ENV["SKIP_UPRN_ISSUE_ORG_IDS"]).include?(organisation.id)
+                                      []
+                                    else
+                                      organisation.managed_lettings_logs
+                                      .imported
+                                      .filter_by_year(2023)
+                                      .where(needstype: 1)
+                                      .where.not(uprn: nil)
+                                      .where("uprn = propcode OR uprn = tenancycode or town_or_city = 'Bristol'")
```
natdeanlewissoftwire: inconsistent OR casing here

spec/factories/lettings_log.rb
```diff
@@ -190,6 +190,9 @@
       status { 4 }
       discarded_at { Time.zone.now }
     end
+    trait :imported do
+      old_id { rand(1...99_999) }
```
natdeanlewissoftwire: isn't this the format of the old_form_ids?
I think the old_ids are long hex strings

app/jobs/email_missing_addresses_csv_job.rb
```diff
@@ -0,0 +1,32 @@
+class EmailMissingAddressesCsvJob < ApplicationJob
+  queue_as :default
+
+  BYTE_ORDER_MARK = "\uFEFF".freeze # Required to ensure Excel always reads CSV as UTF-8
+  EXPIRATION_TIME = 72.hours.to_i
```
natdeanlewissoftwire: Is this enough? This sets the expiration of the s3 objects right?

app/services/csv/missing_addresses_csv_service.rb
```diff
@@ -0,0 +1,165 @@
+module Csv
+  class MissingAddressesCsvService
+    def initialize(organisation)
+      @organisation = organisation
+    end
+
+    def create_missing_lettings_addresses_csv
+      logs_with_missing_addresses = @organisation.managed_lettings_logs
```
natdeanlewissoftwire, line 9: are there any logs with missing address data that we don't have a managing org for? I don't _think_ they could have been created if not but thought I'd confirm

lib/tasks/send_missing_addresses_csv.rake
```diff
@@ -0,0 +1,78 @@
+namespace :correct_addresses do
+  # rubocop:disable Lint/ConstantDefinitionInBlock
+  MISSING_ADDRESSES_THRESHOLD = 50
+  # rubocop:enable Lint/ConstantDefinitionInBlock
+
+  desc "Send missing lettings addresses csv"
+  task :send_missing_addresses_lettings_csv, %i[] => :environment do |_task, _args|
+    Organisation.all.each do |organisation|
+      logs_impacted_by_missing_address = organisation.managed_lettings_logs
```
natdeanlewissoftwire, line 7: I feel like we're using similar combinations of scopes a lot that could be combined into two or three reusable compound scopes

lib/tasks/send_missing_addresses_csv.rake
natdeanlewissoftwire: and elsewhere :)

natdeanlewissoftwire (approved):
some thoughts, lmk what you think!

app/jobs/email_missing_addresses_csv_job.rb
kosiakkatrina: I think not the object but the url. Yeah, maybe, this was my first instinct. What do you think would be appropriate time?

app/services/csv/missing_addresses_csv_service.rb
kosiakkatrina: I think they'd get filtered out in the rest of the query anyway, because you can't have uprn and address values if managing org is not answered

app/services/csv/missing_addresses_csv_service.rb
kosiakkatrina: Very nice

spec/factories/lettings_log.rb
kosiakkatrina: Sure, updated now!

lib/tasks/send_missing_addresses_csv.rake
kosiakkatrina: Yeah, okay, I moved it to the job. Not sure if it belongs there, but perhaps better than defining it globally

lib/tasks/send_missing_addresses_csv.rake
kosiakkatrina: I've combined a couple, not sure if extracting the rest would make it clearer, especially as this is quite specific to this task. Let me know what you think!

app/jobs/email_missing_addresses_csv_job.rb
natdeanlewissoftwire: I'm imagining some users might take on the scale of a week to get back to us on this, is there any disadvantage to setting the expiration time to be much much longer?

natdeanlewissoftwire: I guess these links do have user info so we don't want them to be available forever but a month would probably be alright and give users time to go back and avoid us having to resend links if users take multiple attempts to get the data right for example

app/services/csv/missing_addresses_csv_service.rb
natdeanlewissoftwire: Cool, I agree, happy with this :)

lib/tasks/send_missing_addresses_csv.rake
natdeanlewissoftwire: Yeah the ones you've done are all the ones I think are worth doing, love it 🚀

natdeanlewissoftwire (approved):
nice one!

PR #1955, opened 2023-10-04
Keep the logs in memory when exporting

kosiakkatrina:
lettings_logs_slice was ActiveRecord Relation, which has been lazy-loading this entire time, so the result of it when trying to iterate over it vs when trying to count it further down the line was always dynamic.

.to_a keeps it in memory, .load didn't work for some reason

### DIFF
--- a/spec/services/exports/lettings_log_export_service_spec.rb
+++ b/spec/services/exports/lettings_log_export_service_spec.rb
@@ -292,6 +292,44 @@ RSpec.describe Exports::LettingsLogExportService do
         end
       end
 
+      context "and underlying data changes between getting the logs and writting the manifest" do
+        before do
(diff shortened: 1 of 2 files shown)


spec/services/exports/lettings_log_export_service_spec.rb
```diff
@@ -258,20 +258,40 @@ def replace_person_details(export_file)
       end
     end
 
-    context "and multiple lettings logs are available for export on same quarter" do
```
RachaelBooth: Is this intended to be a replacement or an addition?

kosiakkatrina: Oh dear, definitely an addition

RachaelBooth (approved):

PR #1956, opened 2023-10-05
Reimport postcodes for 2022

kosiakkatrina:
Some postcodes in the imports have trailing spaces in them and we've been running postcode format validation before stripping them, we should reimport those logs to get the correct postcodes.
Addresses are different in 2023, so for for the time being this is only a fix for 2022

### DIFF
--- a/app/services/imports/lettings_logs_field_import_service.rb
+++ b/app/services/imports/lettings_logs_field_import_service.rb
@@ -30,6 +30,8 @@ module Imports
         import_from(folder, :update_childrens_care_referral)
       when "old_form_id"
         import_from(folder, :update_old_form_id)
+      when "postcode_full"
+        import_from(folder, :update_postcode_full)
       else
         raise "Updating #{field} is not supported by the field import service"
       end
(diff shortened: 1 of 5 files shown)


app/services/imports/lettings_logs_field_import_service.rb
natdeanlewissoftwire, line 1: Is there another ticket for importing 2023 postcodes?

app/services/imports/lettings_logs_field_import_service.rb
```diff
@@ -453,5 +455,28 @@ def update_old_form_id(xml_doc)
         @logger.warn("lettings log with old id #{old_id} not found")
       end
     end
+
+    def update_postcode_full(xml_doc)
+      return if meta_field_value(xml_doc, "form-name").include?("Sales")
```
natdeanlewissoftwire, line 460: Do we want an import task for sales logs too?

app/services/imports/lettings_logs_field_import_service.rb
```diff
@@ -453,5 +455,28 @@ def update_old_form_id(xml_doc)
...
+
+      old_id = meta_field_value(xml_doc, "document-id")
+      record = LettingsLog.find_by(old_id:)
+      return @logger.info("lettings log #{record.id} is not from 2022/23 collection period, skipping update") if record.collection_start_year != 2022
+      return @logger.warn("Could not find record matching legacy ID #{old_id}") if record.blank?
+      return @logger.info("lettings log #{record.id} has a value for postcode_full, skipping update") if record.postcode_full.present?
+
+      record.postcode_full = compose_postcode(xml_doc, "POSTCODE", "POSTCOD2")
```
natdeanlewissoftwire, line 468: do we also have this issue for previous postcodes (PPOSTC1 and PPOSTC2)?

app/services/imports/lettings_logs_field_import_service.rb
kosiakkatrina: This hasn't been identified as an issue for 2023 yet, so chances are it might not be. Will hold off from spending any more time on this before we know that there's an issue

app/services/imports/lettings_logs_field_import_service.rb
kosiakkatrina: same as above

app/services/imports/lettings_logs_field_import_service.rb
kosiakkatrina: same as above

natdeanlewissoftwire (approved):
lgtm!

PR #1957, opened 2023-10-05
CLDC-2810 Correct addresses from csv

kosiakkatrina:
We have in the past bulk corrected addresses from a csv. This PR:
- adjusts the task to allow different format CSVs to be uploaded (there are 2 formats, they're the same but one of them has Issue type column)
- adds a separate task for sales address updates.
- adds extra rules around required fields

### DIFF
--- a/lib/tasks/import_address_from_csv.rake
+++ b/lib/tasks/import_address_from_csv.rake
@@ -1,18 +1,34 @@
 namespace :data_import do
-  desc "Import address data from a csv file"
-  task :import_address_from_csv, %i[file_name] => :environment do |_task, args|
+  desc "Import lettings address data from a csv file"
+  task :import_lettings_addresses_from_csv, %i[file_name] => :environment do |_task, args|
     file_name = args[:file_name]
 
-    raise "Usage: rake data_import:import_address_from_csv['csv_file_name']" if file_name.blank?
+    raise "Usage: rake data_import:import_lettings_addresses_from_csv['csv_file_name']" if file_name.blank?
 
     s3_service = Storage::S3Service.new(PlatformHelper.is_paas? ? Configuration::PaasConfigurationService.new : Configuration::EnvConfigurationService.new, ENV["IMPORT_PAAS_INSTANCE"])
     addresses_csv = CSV.parse(s3_service.get_file_io(file_name), headers: true)
 
     addresses_csv.each do |row|
-      lettings_log_id = row[0]
+      lettings_log_id = row[1]
+      address_line1 = row[9]
+      address_line2 = row[10]
+      town_or_city = row[11]
+      county = row[12]
+      postcode_full = row[13]
 
       if lettings_log_id.blank?
-        Rails.logger.info("Lettings log ID not provided for address: #{[row[1], row[2], row[3], row[4]].join(', ')}")
+        Rails.logger.info("Lettings log ID not provided for address: #{[address_line1, address_line2, town_or_city, county, postcode_full].join(', ')}")
+        next
+      end
+
+      uprn = row[8]
+      if uprn.present?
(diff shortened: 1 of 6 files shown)


spec/lib/tasks/correct_address_from_csv_spec.rb
```diff
@@ -2,40 +2,60 @@
...
-    subject(:task) { Rake::Task["data_import:import_address_from_csv"] }
+  before do
+    allow(Storage::S3Service).to receive(:new).and_return(storage_service)
+    allow(Configuration::EnvConfigurationService).to receive(:new).and_return(env_config_service)
+    allow(Configuration::PaasConfigurationService).to receive(:new).and_return(paas_config_service)
+    allow(ENV).to receive(:[])
+    allow(ENV).to receive(:[]).with("IMPORT_PAAS_INSTANCE").and_return(instance_name)
+    allow(ENV).to receive(:[]).with("VCAP_SERVICES").and_return("dummy")
+
+    WebMock.stub_request(:get, /api.postcodes.io/)
```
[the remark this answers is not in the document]
natdeanlewissoftwire: can we just do `\.` to avoid this check failure in the first place rather than overriding it?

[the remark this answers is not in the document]
natdeanlewissoftwire: I accept it's only a test but it could get copied into live code by a confused developer later down the line (can't really imagine how but in principle I'd rather fix vulnerabilities than rely on manually overriding them but that's just my preference), happy if you are

lib/tasks/import_address_from_csv.rake
```diff
@@ -1,18 +1,35 @@
...
+        Rails.logger.info("Lettings log ID not provided for address: #{[address_line1, address_line2, town_or_city, county, postcode_full].join(', ')}")
+        next
+      end
+
+      if uprn.present?
+        Rails.logger.info("Lettings log with ID #{lettings_log_id} contains uprn, skipping log")
+        next
+      end
+
+      if address_line1.blank? || town_or_city.blank? || postcode_full.blank?
```
natdeanlewissoftwire, line 31: isn't postcode (secretly) optional?

lib/tasks/import_address_from_csv.rake
```diff
@@ -25,18 +42,76 @@ namespace :data_import do
...
-      lettings_log.postcode_full = row[4]
+      lettings_log.address_line1 = address_line1
+      lettings_log.address_line2 = address_line2
+      lettings_log.town_or_city = town_or_city
+      lettings_log.county = county
+      lettings_log.postcode_full = postcode_full
       lettings_log.postcode_known = lettings_log.postcode_full.present? ? 1 : nil
-      lettings_log.county = nil
       lettings_log.is_la_inferred = nil
       lettings_log.la = nil
```
natdeanlewissoftwire, line 52: when does la get inferred if postcode is now provided?

natdeanlewissoftwire (approved):
couple of things I'm not sure about but otherwise looks good

spec/lib/tasks/correct_address_from_csv_spec.rb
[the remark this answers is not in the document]
kosiakkatrina: Good point

lib/tasks/import_address_from_csv.rake
kosiakkatrina: Hmm, not sure, I don't think so (?)

lib/tasks/import_address_from_csv.rake
kosiakkatrina: It would still get inferred as part of before_validation hooks, this is just clearing whatever leftover values might have been here before

PR #1958, opened 2023-10-05
CLDC-2723 Allow data coordinators access to stock owners' schemes

natdeanlewissoftwire:
Allows data coordinators to see, edit, change activation, edit locations for schemes owned by data coordinators. The scheme behaviour could probably do with some refactoring to DRY it out, but I've not included much of that in this PR as we want to get it in quickly, tidying up the previously (and still quite) messy scheme code (especially views) should be added to a tech debt ticket. (Or I'll add it here next week if there's time before we have to get this deployed)

ticket: https://dluhcdigital.atlassian.net/browse/CLDC-2723


app/controllers/schemes_controller.rb
```diff
@@ -291,10 +291,6 @@ def find_resource
     @scheme
   end
 
-  def user_allowed_action?
```
natdeanlewissoftwire, line 294: I don't believe this is used anywhere, it's not a rails method or anything so I think we should remove it

app/views/schemes/details.html.erb
```diff
@@ -48,9 +44,14 @@
                                            :description,
                                            legend: { text: "Is this scheme registered under the Care Standards Act 2000?", size: "m" } %>
 
-      <% organisations = Organisation.all.map { |org| OpenStruct.new(id: org.id, name: org.name) } %>
+      <% all_orgs = Organisation.all.map { |org| OpenStruct.new(id: org.id, name: org.name) } %>
+      <% user_org = [OpenStruct.new(id: current_user.organisation_id, name: current_user.organisation.name)] %>
+      <% stock_owners = current_user.organisation.stock_owners.map { |org| OpenStruct.new(id: org.id, name: org.name) } %>
+      <% organisations = current_user.support? ? all_orgs : user_org + stock_owners %>
```
kosiakkatrina: Would it make sense for this to be in a helper?

app/controllers/schemes_controller.rb
kosiakkatrina: Yeah, I think we were using it to authorise in the past. Good spot

kosiakkatrina (approved):

app/views/schemes/details.html.erb
natdeanlewissoftwire: It absolutely would, yes - this ticket isn't as urgent as I thought it was initially so I'll do this now

PR #1959, opened 2023-10-06
CLDC-2804 Create illness csv

kosiakkatrina:
Lots of logs moved in to progress in some migrations because  “How is the person affected by their condition or illness?" was not answered, despite "Does anybody in the household have a physical or mental health condition (or other illness) expected to last 12 months or more?" being "yes".

We want to provide a way to bulk fix these logs, this PR generates a spreadsheet for given org with logs migrated from old CORE that are missing illness data

There's also a rake task to ingest the updated data when it comes back to us.


lib/tasks/correct_illness_from_csv.rake
```diff
@@ -0,0 +1,61 @@
...
+  task :correct_illness_from_csv, %i[file_name] => :environment do |_task, args|
+    file_name = args[:file_name]
+
+    raise "Usage: rake correct_illness:correct_illness_from_csv['csv_file_name']" if file_name.blank?
+
+    s3_service = Storage::S3Service.new(PlatformHelper.is_paas? ? Configuration::PaasConfigurationService.new : Configuration::EnvConfigurationService.new, ENV["IMPORT_PAAS_INSTANCE"])
+    illness_csv = CSV.parse(s3_service.get_file_io(file_name), headers: false)
+
+    illness_csv.each_with_index do |row, index|
+      next if index < 3
```
natdeanlewissoftwire, line 26: is this avoiding headers or something?

natdeanlewissoftwire (approved):
failing test or two, but looks good otherwise!

lib/tasks/correct_illness_from_csv.rake
kosiakkatrina: Yup! There are 3 rows and I don't think there's a pretty way to skip them without using a gem

PR #1960, opened 2023-10-06
CLDC-2563: Update prod deployment pipeline

SamSeed-Softwire:


.github/workflows/aws_deploy.yml
```diff
@@ -59,6 +63,9 @@ jobs:
     runs-on: ubuntu-latest
     environment: ${{ inputs.environment }}
     needs: push_docker_image
+    if: |
+      always() &&
```
SamSeed-Softwire: `always()` was required in order to ensure the logic in the line below worked. I think this is slightly unintuitive, and I found several threads on the internet discussing this with lots of other people being confused by it! I did test it thoroughly though and found this to be the correct setup - it will run if and only if the previous step succeeded or was skipped.

.github/workflows/aws_deploy.yml
```diff
@@ -75,17 +82,30 @@ jobs:
 
       - name: Get timestamp
         id: timestamp
-        run: echo "timestamp=$(date +%Y%m%d%H%M%S)" >> $GITHUB_ENV
+        run: echo "timestamp=$(date +%Y%m%d%H%M%S)" >> $GITHUB_OUTPUT
```
SamSeed-Softwire: AFAICT, using `GITHUB_OUTPUT` is the best way to store variables for use in the action.

.github/workflows/aws_deploy.yml
```diff
@@ -75,17 +82,30 @@ jobs:
...
+      - name: Checkout at release tag
+        uses: actions/checkout@v3
+        if: inputs.environment == 'production'
+        with:
+          ref: ${{ inputs.release_tag }}
+
+      - name: Get tags
+        id: tags
+        run: |
+          echo "commit-tag=$(if [[ ${{ inputs.environment }} == 'production' ]]; then echo $(git log -1 '--format=format:%H'); else echo ${{ github.sha }}; fi)" >> $GITHUB_OUTPUT
```
SamSeed-Softwire: `git log -1 '--format=format:%H'` is the best way I found for getting the commit hash for where HEAD is (i.e. at the release tag).

.github/workflows/aws_deploy.yml
```diff
@@ -75,17 +82,30 @@ jobs:
...
-        run: echo "timestamp=$(date +%Y%m%d%H%M%S)" >> $GITHUB_ENV
+        run: echo "timestamp=$(date +%Y%m%d%H%M%S)" >> $GITHUB_OUTPUT
+
+      - name: Checkout at release tag
+        uses: actions/checkout@v3
+        if: inputs.environment == 'production'
+        with:
+          ref: ${{ inputs.release_tag }}
+
+      - name: Get tags
```
SamSeed-Softwire: I know this step is a bit messy - it could be cleaned up, but I just wanted to get something up for PR first.

.github/workflows/production_pipeline.yml
```diff
@@ -257,3 +259,16 @@ jobs:
...
+  aws_deploy:
+    name: AWS Deploy
+    if: github.ref == 'refs/heads/main'
+    needs: [lint, test, feature_test, audit]
+    uses: ./.github/workflows/aws_deploy.yml
+    with:
+      aws_account_id: 977287343304
+      aws_resource_prefix: core-prod
+      environment: production
+      release_tag: ${{ needs.test.outputs.releasetag }}
```
SamSeed-Softwire, line 271: I felt that passing through the calculated release tag, rather than recalculating it in `aws_deploy.yml`, was nicer. It might be nicer still to start the production pipeline action with a job which calculates the release tag. All other jobs will depend on this job, so we won't have to keep recalculating the release tag everywhere. I thought better to raise a PR before changing anything more though.

.github/workflows/production_pipeline.yml
```diff
@@ -48,7 +50,7 @@ jobs:
       - name: Get latest release with tag
         id: latestrelease
         run: |
-          echo "::set-output name=releasetag::$(curl -s https://api.github.com/repos/${REPO_URL}/releases/latest | jq '.tag_name' | sed 's/\"//g')"
+          echo "releasetag=$(curl -s https://api.github.com/repos/${REPO_URL}/releases/latest | jq '.tag_name' | sed 's/\"//g')" >> $GITHUB_OUTPUT
```
SamSeed-Softwire, line 53: `set-output` has been deprecated, so I changed it to the preferred method (using `GITHUB_OUTPUT`) here and in some other places. There are other places in the pipelines where I haven't updated it - I thought better to raise this PR first, then if you agree with my thinking I can raise a ticket to clean up the rest of the pipelines later.

.github/workflows/aws_deploy.yml
```diff
@@ -75,17 +82,30 @@ jobs:
...
+        id: tags
+        run: |
+          echo "commit-tag=$(if [[ ${{ inputs.environment }} == 'production' ]]; then echo $(git log -1 '--format=format:%H'); else echo ${{ github.sha }}; fi)" >> $GITHUB_OUTPUT
+          echo "additional-tag=$(if [[ ${{ inputs.environment }} == 'production' ]]; then echo ${{ inputs.release_tag }}-${{ steps.timestamp.outputs.timestamp }}; else echo ${{ steps.timestamp.outputs.timestamp }}; fi)" >> $GITHUB_OUTPUT
```
SamSeed-Softwire: Trying to update add the same tag to an image twice leads to an error. So I added timestamp to the prod image tagging too. We could maybe add some logic to check whether a tag exists, and if it does then don't try to add it again. But maybe it's easier and nicer to just stick with a timestamp in all tags.

.github/workflows/aws_deploy.yml
```diff
@@ -24,6 +27,7 @@ env:
 
 jobs:
   push_docker_image:
+    if: inputs.environment != 'production'
```
RachaelBooth: How easy would it be to check whether an image with the expected tag exists instead of relying on the environment?

It would seem overall better to avoid having too many ifs/special casing production too much, because then it's possible to fully re-run the pipeline for staging and review apps (and indeed to re-run a production pipeline directly that's old enough that the image has been expired, although that's a really unlikely thing to want to do!), but it's only worth doing right now rather than as tech debt later if it's easy.

.github/workflows/aws_deploy.yml
RachaelBooth: I actually think the norm is usually to use environment unless there's a reason to use output, because it's easier (for a human) to parse or write a reference to `env.blah` than to `steps.what-was-the-id-again.outputs.blah`. That's why this was using environment, anyway.

You need output if passing data between jobs or between workflows (including typically out of a step which is calling a 3rd party workflow because that's the workflow outputs), or if there's some reason you need a value only to be available to certain steps (which seems unlikely). For values that just need to be accessible to future steps in the same job (and, conceptually at least, particularly if they will be accessed multiple times) environment works well unless you'd be likely to get a naming clash, and seems to be what github expects you to do.

.github/workflows/aws_deploy.yml
RachaelBooth: Is the `github.sha` not available in this case? I would expect it to be.

.github/workflows/production_pipeline.yml
RachaelBooth: Yes, please raise a ticket (probably a tech debt ticket for the app dev team).

RachaelBooth: Also, it looks like several jobs in this workflow start with this step - this is probably for said tech debt ticket, but it'd be good if it was pulled out into a separate job that was a prerequisite for all the others so it only happened once.

.github/workflows/aws_deploy.yml
SamSeed-Softwire: It is, but `github.sha` refers to the commit that triggered the workflow, not where HEAD is right now. The full list of github contexts can be seen [here](https://docs.github.com/en/actions/learn-github-actions/contexts#github-context) - I don't think any of them are what we want, and the checkout action used further up the file doesn't output the commit sha (I saw a thread about this, in which people suggested what I did above as a way to fetch the current commit sha).

.github/workflows/production_pipeline.yml
SamSeed-Softwire: I have raised 2 tickets for this:
- https://dluhcdigital.atlassian.net/browse/CLDC-2888
- https://dluhcdigital.atlassian.net/browse/CLDC-2889

.github/workflows/aws_deploy.yml
RachaelBooth: Is the commit that triggered the workflow not the one we want? This is triggered on a release, but looking [here](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#release) it seems like `github.sha` should be the latest commit in the tagged release.

SamSeed-Softwire: Fair enough - I think because I wasn't testing this with a genuine release (e.g. I was forcing the 'prod' version of aws deploy workflow to run when pushing to a branch), the commit that triggered this workflow wasn't the commit of the release tag. But in a true deployment it would be, so `github.sha` would be appropriate in that case.

.github/workflows/aws_deploy.yml
SamSeed-Softwire: I've done this now and it's made things cleaner. Thanks! ☺️

.github/workflows/production_pipeline.yml
```diff
@@ -257,3 +259,16 @@ jobs:
           cf set-env $APP_NAME CSV_DOWNLOAD_PAAS_INSTANCE $CSV_DOWNLOAD_PAAS_INSTANCE
           cf set-env $APP_NAME SENTRY_DSN $SENTRY_DSN
           cf push $APP_NAME --strategy rolling
+
+  aws_deploy:
+    name: AWS Deploy
+    if: github.ref == 'refs/heads/main'
```
RachaelBooth: Just had a thought - is this restriction wanted/needed in this workflow, or is it baked into when the workflow runs? It doesn't look like the paas deploy restricts here.

SamSeed-Softwire: Yeah good spot - I've removed. Thanks!

RachaelBooth (approved):

PR #1966, opened 2023-10-09
CLDC-2796 Display managed logs for support

kosiakkatrina:
We want to display both owned and managed logs in the organisation page for support users. Filter_by_organisation returns both managed and owned logs.

### DIFF
--- a/spec/requests/organisations_controller_spec.rb
+++ b/spec/requests/organisations_controller_spec.rb
@@ -716,13 +716,20 @@ RSpec.describe OrganisationsController, type: :request do
         end
 
         context "when viewing a specific organisation's lettings logs" do
-          let(:number_of_org1_lettings_logs) { 2 }
-          let(:number_of_org1_managed_lettings_logs) { 2 }
+          let(:parent_organisation) { create(:organisation) }
+          let(:child_organisation) { create(:organisation) }
+          let(:number_of_owned_org1_lettings_logs) { 2 }
+          let(:number_of_managed_org1_lettings_logs) { 2 }
+          let(:number_of_owned_and_managed_org1_lettings_logs) { 2 }
+          let(:total_number_of_org1_logs) { number_of_owned_org1_lettings_logs + number_of_managed_org1_lettings_logs + number_of_owned_and_managed_org1_lettings_logs }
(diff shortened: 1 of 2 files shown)


spec/requests/organisations_controller_spec.rb
```diff
@@ -717,23 +717,29 @@
 
         context "when viewing a specific organisation's lettings logs" do
           let(:number_of_org1_lettings_logs) { 2 }
+          let(:number_of_org1_managed_lettings_logs) { 2 }
           let(:number_of_org2_lettings_logs) { 4 }
 
           before do
             create_list(:lettings_log, number_of_org1_lettings_logs, created_by: user)
```
natdeanlewissoftwire: can we be more explicit that these are owned_lettings_logs to avoid confusion?

spec/requests/organisations_controller_spec.rb
```diff
@@ -717,23 +717,29 @@
...
+            create_list(:lettings_log, number_of_org1_managed_lettings_logs, managing_organisation: user.organisation)
             create(:lettings_log, created_by: user, status: "pending", skip_update_status: true)
             create_list(:lettings_log, number_of_org2_lettings_logs, created_by: nil, owning_organisation_id: unauthorised_organisation.id, managing_organisation_id: unauthorised_organisation.id)
 
             get "/organisations/#{organisation.id}/lettings-logs", headers:, params: {}
           end
 
           it "only shows logs for that organisation" do
-            expect(page).to have_content("#{number_of_org1_lettings_logs} total logs")
+            expect(page).to have_content("#{number_of_org1_lettings_logs + number_of_org1_managed_lettings_logs} total logs")
```
natdeanlewissoftwire: Can we also make a log that is owned and managed by the org and ensure it isn't included twice in the count (or the page)

natdeanlewissoftwire (approved):
couple of ideas, but looks good

spec/requests/organisations_controller_spec.rb
kosiakkatrina: Sure!

natdeanlewissoftwire (approved):
🥳

PR #1967, opened 2023-10-09
CLDC-2793 Blank relevant soctenant values

kosiakkatrina:
We've inferred some values when migrating the logs as don't know, but they are more difficult to fix because they're conditionally routed to based on soctenant question and might not be relevant for every sales log. Instead we want to blank the soctenant question if there is no useful information in subsequent questions (don't know etc)


natdeanlewissoftwire (approved):
nice!

PR #1971, opened 2023-10-11
CLDC-2871 Fix bulk upload scheme lookup bug

natdeanlewissoftwire:
In 2023 lettings bulk upload, we allow users to use old or new ids for schemes and locations, depending if field_16 begins with an "S" or not. This behaviour wasn't implemented correctly at all, and we've only noticed now users have started trying to do this as the tests were testing the incorrect behaviour. More detail of correct behaviour in the ticket.

ticket: https://dluhcdigital.atlassian.net/browse/CLDC-2871


app/services/bulk_upload/lettings/year2023/row_parser.rb
```diff
@@ -329,9 +329,26 @@ class BulkUpload::Lettings::Year2023::RowParser
             },
             on: :after_log
 
+  validates :field_15,
+            presence: {
+              if: proc { [2, 4, 6, 8, 10, 12].include?(field_5) && log_uses_old_scheme_id? },
```
kosiakkatrina: would we be able to check bulk_upload.needstype here instead? Might be a bit clearer what these numbers mean

kosiakkatrina: I like the name `log_uses_old_scheme_id`

app/services/bulk_upload/lettings/year2023/row_parser.rb
```diff
@@ -732,45 +749,45 @@ def validate_nulls
...
+      errors.add(scheme_field, "The #{scheme_or_management_group} code is not correct", category: :setup)
     end
   end
 
   def validate_scheme_data_given
-    if supported_housing? && field_16.blank?
-      errors.add(:field_16, I18n.t("validations.not_answered", question: "management group code"), category: "setup")
+    if supported_housing? && scheme_field.present? && scheme_id.blank?
+      errors.add(scheme_field, I18n.t("validations.not_answered", question: "#{scheme_or_management_group} code"), category: "setup")
     end
```
kosiakkatrina, line 791: can scheme_field/location_field be nil?

app/services/bulk_upload/lettings/year2023/row_parser.rb
natdeanlewissoftwire: Yep, sure!

app/services/bulk_upload/lettings/year2023/row_parser.rb
natdeanlewissoftwire: yep if it's a general needs log, but they're compulsory for supported housing logs

kosiakkatrina: What if they haven't entered any of the location/scheme values. Would scheme_field.present? be false?

natdeanlewissoftwire: in that case we don't know which is the scheme field or location field so they'll be nil

natdeanlewissoftwire: everywhere we're adding errors to scheme_field or location_field I've checked it exists first, and if it is nil another error is added to field_16 saying it must be answered, which there is a test for

kosiakkatrina: okay, gotcha! tyty!

kosiakkatrina (approved):

app/services/bulk_upload/lettings/year2023/row_parser.rb
natdeanlewissoftwire: nw, good question!

PR #1972, opened 2023-10-11
CLDC-2627 Update scheme table styling and labels

kosiakkatrina:
Update scheme table tables, wrap long scheme names and sort schemes by name only

### DIFF
--- a/app/frontend/styles/_table-group.scss
+++ b/app/frontend/styles/_table-group.scss
@@ -9,7 +9,6 @@
     margin-bottom: -1px;
   }
 
-  .govuk-table__header,
   .govuk-table__cell {
     white-space: nowrap;
   }
--- a/app/frontend/styles/_table-group.scss
+++ b/app/frontend/styles/_table-group.scss
@@ -23,6 +23,10 @@
   }
 }
 
+.scheme-name-cell {
+  word-break: break-all;
+}
+
 .app-table-group:focus {
   box-shadow: 0 0 0 #{$govuk-focus-width * 2} $govuk-focus-colour;
   outline: $govuk-focus-width solid govuk-colour("black");
--- a/app/frontend/styles/_table-group.scss
+++ b/app/frontend/styles/_table-group.scss
@@ -9,10 +9,6 @@
     margin-bottom: -1px;
   }
 
-  .govuk-table__cell {
-    white-space: nowrap;
-  }
-
   &::-webkit-scrollbar {
     height: govuk-spacing(1);
     width: govuk-spacing(1);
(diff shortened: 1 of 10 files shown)


app/models/scheme.rb
```diff
@@ -8,17 +8,18 @@ class Scheme < ApplicationRecord
...
+                                          .or(search_by_service_name(param))
+                                          .or(search_by_location_name(param))
+                                          .or(filter_by_id(param))
                     }
 
-  scope :order_by_completion, -> { order("schemes.confirmed ASC NULLS FIRST") }
-  scope :order_by_service_name, -> { order(service_name: :asc) }
+  scope :order_by_service_name, lambda {
+    order("lower(service_name) ASC")
+  }
```
kosiakkatrina, line 22: It was previously prioritising capitalised names, despite the alphabetical order

app/models/scheme.rb
```diff
@@ -8,17 +8,18 @@ class Scheme < ApplicationRecord
 
   scope :filter_by_id, ->(id) { where(id: (id.start_with?("S") ? id[1..] : id)) }
   scope :search_by_service_name, ->(name) { where("service_name ILIKE ?", "%#{name}%") }
-  scope :search_by_postcode, ->(postcode) { left_joins(:locations).where("REPLACE(locations.postcode, ' ', '') ILIKE ?", "%#{postcode.delete(' ')}%") }
-  scope :search_by_location_name, ->(name) { left_joins(:locations).where("locations.name ILIKE ?", "%#{name}%") }
+  scope :search_by_postcode, ->(postcode) { where("schemes.id IN (SELECT DISTINCT scheme_id FROM locations WHERE REPLACE(locations.postcode, ' ', '') ILIKE ?)", "%#{postcode.delete(' ')}%") }
+  scope :search_by_location_name, ->(name) { where("schemes.id IN (SELECT DISTINCT scheme_id FROM locations WHERE locations.name ILIKE ?)", "%#{name}%") }
```
kosiakkatrina, line 12: Refactor these into subqueries instead of joins so that we don't need distinct in search_by anymore. Distinct clashes with count and we probably want to be able to count the results of search_by

app/frontend/styles/_table-group.scss
```diff
@@ -9,11 +9,6 @@
     margin-bottom: -1px;
   }
 
-  .govuk-table__header,
```
natdeanlewissoftwire, line 12: isn't this being used in `cookies/show.html.erb`?

app/models/scheme.rb
natdeanlewissoftwire: uh oh, good spot!

natdeanlewissoftwire (approved):

app/frontend/styles/_table-group.scss
kosiakkatrina: I assume it's being used on every table, but I'm not a 100% sure why. I don't think default splits words and wrapping sounds better than having a scrollbar. But shout if disagree!

natdeanlewissoftwire: Looks like Paul Robert Lloyd added it when working on the logs summary page in march 2022, don't think it's relevant anymore so happy to remove, logs pages look fine in the review app

PR #1974, opened 2023-10-11
Delete obsolete data_export_csv task

SamSeed-Softwire:
I spotted that there was a `data_export_csv` task in `lib/tasks/data_export.rake`. It referenced a class `DataExportCsvJob` which wasn't defined anywhere else in the codebase. @kosiakkatrina confirmed it was redundant.


natdeanlewissoftwire (approved):
nice spot!

PR #1976, opened 2023-10-12
CLDC-2628 Update schemes guidance page

kosiakkatrina:
Refactor guidance partials to allow top and bottom guidance at the same time.
Add/change guidance content for schemes question and add a new scheme changes page
image

### DIFF
--- a/app/models/form/lettings/pages/scheme.rb
+++ b/app/models/form/lettings/pages/scheme.rb
@@ -6,6 +6,7 @@ class Form::Lettings::Pages::Scheme < ::Form::Page
         "needstype" => 2,
       },
     ]
+    @header = "Scheme"
     @next_unresolved_page_id = "location"
   end
 
--- a/app/views/form/guidance/_finding_scheme.erb
+++ b/app/views/form/guidance/_finding_scheme.erb
@@ -1,0 +1,5 @@
+<%= govuk_details(summary_text: "Can’t find your scheme?") do %>
+  <p class="govuk-body">If your organisation’s schemes were migrated from old CORE, they may have new names and codes. Search by postcode to find your scheme.</p>
+  <p class="govuk-body"><%= govuk_link_to("View your organisation’s schemes.", schemes_path) %></p>
+  <p class="govuk-body"><%= govuk_link_to("Read more about how schemes have changed", scheme_changes_path) %></p>
+<% end %>
--- a/app/views/form/guidance/_finding_scheme.erb
+++ b/app/views/form/guidance/_finding_scheme.erb
@@ -1,5 +1,5 @@
 <%= govuk_details(summary_text: "Can’t find your scheme?") do %>
   <p class="govuk-body">If your organisation’s schemes were migrated from old CORE, they may have new names and codes. Search by postcode to find your scheme.</p>
-  <p class="govuk-body"><%= govuk_link_to("View your organisation’s schemes.", schemes_path) %></p>
+  <p class="govuk-body"><%= govuk_link_to("View your organisation’s schemes", schemes_path) %></p>
   <p class="govuk-body"><%= govuk_link_to("Read more about how schemes have changed", scheme_changes_path) %></p>
 <% end %>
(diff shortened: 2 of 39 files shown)


app/views/form/guidance/_finding_scheme.erb
```diff
@@ -0,0 +1,5 @@
+<%= govuk_details(summary_text: "Can’t find your scheme?") do %>
+  <p class="govuk-body">If your organisation’s schemes were migrated from old CORE, they may have new names and codes. Search by postcode to find your scheme.</p>
+  <p class="govuk-body"><%= govuk_link_to("View your organisation’s schemes.", schemes_path) %></p>
+  <p class="govuk-body"><%= govuk_link_to("Read more about how schemes have changed", scheme_changes_path) %></p>
```
natdeanlewissoftwire, line 4: missing a full stop?

app/views/form/guidance/_scheme_selection.html.erb
```diff
@@ -1,5 +1,5 @@
 <% if current_user.data_provider? %>
   <p class="govuk-body">If you can’t find the supported housing service you’re looking for or not sure which to choose, contact a data coordinator at <%= current_user.organisation.name %>.</p>
 <% elsif current_user.data_coordinator? %>
-  <p class="govuk-body">or <%= govuk_link_to "create a new supported housing service", new_scheme_path %></p>
+  <p class="govuk-body"><%= govuk_link_to "Create a new supported housing scheme", new_scheme_path %></p>
```
natdeanlewissoftwire, line 4: Missing a full stop here too?

app/models/form/lettings/pages/scheme.rb
```diff
@@ -6,6 +6,7 @@ def initialize(_id, hsh, subsection)
         "needstype" => 2,
       },
     ]
+    @header = "Scheme"
```
natdeanlewissoftwire, line 9: are we intenionally not showing the question number anymore?

app/views/schemes/changes.html.erb
```diff
@@ -0,0 +1,29 @@
...
+    <p class="govuk-body">On old CORE, a scheme’s data was split into management group and scheme. On new CORE, the data is split into scheme and location.</p>
+    <p class="govuk-body">These are the main changes:</p>
+    <ul class="govuk-list govuk-list--bullet">
+      <li>Schemes now store data about the stock owner and support provider. This was previously stored in management group.</li>
+      <li>Schemes still store data about the type of support, as they did in old CORE.</li>
+      <li>Schemes no longer store data about postcodes where support is provided. These are now in locations.</li>
+    </ul>
+    <p class="govuk-body">This new structure means data coordinators only needs to fill in support details once.</p>
+    <h2 class="govuk-heading-m">How schemes are migrated from old CORE</h2>
+    <p class="govuk-body">If your organisation has migrated from old CORE to new CORE, your existing schemes are on the <%= govuk_link_to("schemes page", schemes_path) %>, with the same details.</p>
```
natdeanlewissoftwire, line 25: is old CORE and new CORE the correct terminology or just something we've been using internally? Seems a bit casual, I'd have expected something closer to "the previous version of this service" but arguably that's less clear

natdeanlewissoftwire (approved):
just a thought, might be worth adding a test for the new scheme changes page routing

app/views/form/guidance/_finding_scheme.erb
kosiakkatrina: Oh I think we don't want a full stop on either of those cause they're links

app/views/form/guidance/_scheme_selection.html.erb
kosiakkatrina: same as above

app/models/form/lettings/pages/scheme.rb
kosiakkatrina: Oh, I think we still do! Just not for 2022, which was the case already

app/views/schemes/changes.html.erb
kosiakkatrina: hmm, I think we've used it around bulk upload in the past as well, so I assume that's the language that's the most clear

PR #1978, opened 2023-10-12
CLDC-2640 Add location guidance

kosiakkatrina:
image

### DIFF
--- a/app/models/form/lettings/questions/location_id.rb
+++ b/app/models/form/lettings/questions/location_id.rb
@@ -13,6 +13,7 @@ class Form::Lettings::Questions::LocationId < ::Form::Question
     }
     @question_number = 10
     @disable_clearing_if_not_routed_or_dynamic_answer_options = true
+    @top_guidance_partial = "finding_location"
   end
 
   def answer_options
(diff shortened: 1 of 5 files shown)


app/views/form/guidance/_finding_location.erb
```diff
@@ -0,0 +1,4 @@
+<%= govuk_details(summary_text: "What is a location?") do %>
+  <p class="govuk-body">A location is a postcode where supported housing is provided under a scheme. A scheme can have multiple locations, and a location can have multiple units at the same postcode.</p>
+  <p class="govuk-body"><%= govuk_link_to("Read more about how schemes and locations", scheme_changes_path) %></p>
```
natdeanlewissoftwire: This isn't a full sentence - is there missing copy here?

kosiakkatrina: You make a very good point

app/models/form/lettings/questions/location_id.rb
```diff
@@ -13,6 +13,7 @@ def initialize(id, hsh, page)
     }
     @question_number = 10
     @disable_clearing_if_not_routed_or_dynamic_answer_options = true
+    @top_guidance_partial = "finding_location"
```
natdeanlewissoftwire, line 16: should we add a spec for this?

natdeanlewissoftwire (approved):

PR #1982, opened 2023-10-16
CLDC-2882 Changed void date routing to only use isRenewal

davidm-m:


app/models/form/lettings/pages/void_date.rb
```diff
@@ -2,8 +2,7 @@ class Form::Lettings::Pages::VoidDate < ::Form::Page
   def initialize(id, hsh, subsection)
     super
     @id = "void_date"
-    @depends_on = [{ "is_renewal?" => false, "vacancy_reason_not_renewal_or_first_let?" => true },
-                   { "is_renewal?" => false, "has_first_let_vacancy_reason?" => true }]
+    @depends_on = [{ "is_renewal?" => false }]
```
kosiakkatrina, line 5: Do we also need to do it for 2022?

davidm-m: No, per Rachel:
> it’s a hangover from the 22/23 forms when “Is it a renewal?” was an optional question so someone decided it would be a good idea to check whether reason for vacancy was renewal

2022 doesn't have option 20 for reasons for vacancy and we do still want to check reason for vacancy, so no changes needed there.

kosiakkatrina (approved):

PR #1983, opened 2023-10-16
CLDC-2671 Update duplicate logs routing

natdeanlewissoftwire:
When user uses the duplicates log banner to edit duplicate logs, change the button at the end of the journey to return them to the review duplicates page, and add a "No more duplicate logs" page if there are none left
image
image

ticket: https://dluhcdigital.atlassian.net/browse/CLDC-2671


app/services/csv/sales_log_csv_service.rb
```diff
@@ -58,12 +58,6 @@ def prepare_csv(logs)
       end
     }.freeze
 
-    AGE_KNOWN_FIELDS = {}.tap { |hash|
```
natdeanlewissoftwire, line 61: leftover unused code from a separate PR

kosiakkatrina (approved):

PR #1984, opened 2023-10-17
CLDC-2746 Add floating support type to schemes

kosiakkatrina:
Add floating support type for migrated schemes, but do not allow it to be chosen from the UI

### DIFF
--- a/app/models/scheme.rb
+++ b/app/models/scheme.rb
@@ -108,6 +108,7 @@ class Scheme < ApplicationRecord
     "Medium level": 3,
     "High level": 4,
     "Nursing care in a care home": 5,
+    "Floating support": 6,
   }.freeze
 
   enum support_type: SUPPORT_TYPE, _suffix: true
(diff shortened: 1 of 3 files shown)


app/models/scheme.rb
```diff
@@ -224,7 +225,7 @@ def support_level_options_with_hints
       "Medium level": "Staff on site daily or making frequent visits with some out-of-hours cover.",
       "High level": "Intensive level of staffing provided on a 24-hour basis.",
     }
-    Scheme.support_types.keys.excluding("Missing").map { |key, _| OpenStruct.new(id: key, name: key.to_s.humanize, description: hints[key.to_sym]) }
+    Scheme.support_types.keys.excluding("Missing").excluding("Floating support").map { |key, _| OpenStruct.new(id: key, name: key.to_s.humanize, description: hints[key.to_sym]) }
```
natdeanlewissoftwire, line 228: I think we can combine these two `.excluding` calls for conciseness

app/views/schemes/support.html.erb
```diff
@@ -15,7 +15,7 @@
 
       <% support_level_options_hints = { "Low level": "Staff visiting once a week, fortnightly or less.", "Medium level": "Staff on site daily or making frequent visits with some out-of-hours cover.", "High level": "Intensive level of staffing provided on a 24-hour basis." } %>
 
-      <% support_level_options_with_hints = Scheme.support_types.keys.excluding("Missing").map { |key, _| OpenStruct.new(id: key, name: key.to_s.humanize, description: support_level_options_hints[key.to_sym]) } %>
+      <% support_level_options_with_hints = Scheme.support_types.keys.excluding("Missing").excluding("Floating support").map { |key, _| OpenStruct.new(id: key, name: key.to_s.humanize, description: support_level_options_hints[key.to_sym]) } %>
```
natdeanlewissoftwire, line 18: likewise :)

natdeanlewissoftwire (approved):
lgtm, just a style comment

PR #1986, opened 2023-10-17
CLDC-2924 Correct location admin district

kosiakkatrina:
We need both location_admin_district and location_code to be answered for a location to be marked completed. 
If it cannot be inferred from postcode we ask location_admin_district question and infer location_code from that answer.
Upon import we import location_code which doesn’t set location_admin_district. 
Because we displayed linked local authorities based on the location_code instead of location_admin_district it’s not clear that something is missing.
This task infers location_admin_district value from location_code where it’s not present.

### DIFF
--- a/lib/tasks/correct_location_admin_district.rake
+++ b/lib/tasks/correct_location_admin_district.rake
@@ -1,0 +1,6 @@
+desc "Infers location admin district for locations from location code where it's missing"
+task correct_location_admin_district: :environment do
+  Location.where.not(location_code: nil).where(location_admin_district: nil).each do |location|
+    location.update(location_admin_district: LocalAuthority.all.active(Time.zone.today).england.find_by(code: location.location_code)&.name)
+  end
+end
--- a/spec/lib/tasks/correct_location_admin_district_spec.rb
+++ b/spec/lib/tasks/correct_location_admin_district_spec.rb
@@ -1,0 +1,52 @@
+require "rails_helper"
+require "rake"
+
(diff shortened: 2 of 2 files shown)


lib/tasks/correct_location_admin_district.rake
```diff
@@ -0,0 +1,6 @@
+desc "Infers location admin district for locations from location code where it's missing"
+task correct_location_admin_district: :environment do
+  Location.where.not(location_code: nil).where(location_admin_district: nil).each do |location|
+    location.update(location_admin_district: LocalAuthority.all.active(Time.zone.today).england.find_by(code: location.location_code)&.name)
```
natdeanlewissoftwire, line 4: I feel like you mentioned this before, but do we not want to chose the la that existed at the time of the log startdate rather than now?

natdeanlewissoftwire (approved):
Jemma mentioned this morning we're migrating a couple of new orgs, but i'm pretty certain that won't include schemes so no reason to delay this there, correct me if you think otherwise though!

lib/tasks/correct_location_admin_district.rake
kosiakkatrina: So it's a little weird, I don't think we're actually using this location_admin_district for anything other than inferring the E-code(location_code). Once we have that we display the history of location admin districts for that code using Local Authority table and we also choose the relevant local authority based on the location_code instead of the location_admin_district for each log. The only reason the location_admin_district is mandatory is because it's the field in the UI that gets us the location_code.

PR #1988, opened 2023-10-18
CLDC-2672 Update save and cancel buttons for duplicates

kosiakkatrina:
Instead of having `Save and continue` and `Skip for now` buttons when updating duplicate logs, we want it to act similar to check your answers and have `Save changes` and `Cancel`

### DIFF
--- a/app/views/form/page.html.erb
+++ b/app/views/form/page.html.erb
@@ -63,20 +63,19 @@
       <%= f.hidden_field :interruption_page_referrer_type, value: @interruption_page_referrer_type %>
 
       <div class="govuk-button-group">
-        <% if !@page.interruption_screen? && if request.query_parameters["referrer"] != "check_answers" && request.query_parameters["referrer"] != "duplicate_logs" %>
-          <%= f.govuk_submit "Save and continue" %>
+      <% if accessed_from_duplicate_logs?(request.query_parameters["referrer"]) %>
+        <%= f.govuk_submit "Save changes" %>
+        <%= govuk_link_to "Cancel", send("#{@log.class.name.underscore}_duplicate_logs_path", @log, original_log_id: request.query_parameters["original_log_id"]) %>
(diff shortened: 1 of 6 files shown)


app/views/form/page.html.erb
```diff
@@ -63,12 +63,15 @@
       <%= f.hidden_field :interruption_page_referrer_type, value: @interruption_page_referrer_type %>
 
       <div class="govuk-button-group">
-        <% if !@page.interruption_screen? && if request.query_parameters["referrer"] != "check_answers" %>
+        <% if !@page.interruption_screen? && if request.query_parameters["referrer"] != "check_answers" && request.query_parameters["referrer"] != "duplicate_logs" %>
```
natdeanlewissoftwire: if this is merged after https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/1983 then it should also exclude the `duplicate_logs_banner` referrer as it stands

app/views/form/page.html.erb
```diff
@@ -63,12 +63,15 @@
...
           <%= f.govuk_submit "Save and continue" %>
           <%= govuk_link_to(
             (@page.skip_text || "Skip for now"),
             (@page.skip_href(@log) || send(@log.form.next_page_redirect_path(@page, @log, current_user), @log)),
           ) %>
+          <% elsif request.query_parameters["referrer"] == "duplicate_logs" %>
```
natdeanlewissoftwire: similarly this would be `duplicate_logs` OR `duplicate_logs_banner`

natdeanlewissoftwire (approved):
lgtm!

app/views/form/page.html.erb
kosiakkatrina: Yup, just saw that!

natdeanlewissoftwire: I've just merged the above PR so this extra referrer check will need to be in this PR please :) happy to pair if you like/want any clarification

PR #1991, opened 2023-10-18
Fix flaky test

kosiakkatrina:


spec/requests/organisations_controller_spec.rb
```diff
@@ -129,6 +129,7 @@
           schemes[2].update!(service_name: "baa", owning_organisation: user.organisation)
```
natdeanlewissoftwire, line 129: the nittiest of nits: while we're here, line 126 has a typo, should be "alphabetical", probably worth doing a find and replace :)

natdeanlewissoftwire (approved):

PR #1992, opened 2023-10-18
CLDC-2874 Remove UPRN validation for supported housing logs

kosiakkatrina:


spec/requests/organisations_controller_spec.rb
```diff
@@ -129,6 +129,7 @@
           schemes[2].update!(service_name: "baa", owning_organisation: user.organisation)
           schemes[3].update!(service_name: "Faa", owning_organisation: user.organisation)
           schemes[4].update!(service_name: "Caa", owning_organisation: user.organisation)
+          same_org_scheme.update!(service_name: "zzz", owning_organisation: user.organisation)
```
kosiakkatrina: unrelated flaky test

natdeanlewissoftwire: this was in the PR I just approved right?
https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/1991/files

app/services/bulk_upload/lettings/year2023/row_parser.rb
```diff
@@ -402,7 +402,7 @@ class BulkUpload::Lettings::Year2023::RowParser
 
   validate :validate_nulls, on: :after_log
 
-  validate :validate_uprn_exists_if_any_key_address_fields_are_blank, on: :after_log
+  validate :validate_uprn_exists_if_any_key_address_fields_are_blank, on: :after_log, unless: -> { supported_housing? }
```
natdeanlewissoftwire, line 405: tiny thing, making this condition `if {: -> { general_needs? }`  instead seems a bit clearer to me but I accept that it doesn't make any functional difference because we couldn't get past the setup section if needstype was blank so feel free to merge as is

natdeanlewissoftwire (approved):

spec/requests/organisations_controller_spec.rb
kosiakkatrina: Yup! The test was failing here too 😅

app/services/bulk_upload/lettings/year2023/row_parser.rb
kosiakkatrina: Good point, I think we're checking the value of field 4 instead of the bulk upload needstype, so if that's not given we want to run this as well. That was my thought process for doing an unless, but I don't think it matters much cause of a similar reason you mentioned, given a setup field would be missing

PR #1997, opened 2023-10-20
CLDC-2105 Display merged orgs for owning organisation question

kosiakkatrina:
I think we were already displaying them but this also adds the dates in brackets:
image
image
For support:
image

### DIFF
--- a/app/models/form/lettings/questions/stock_owner.rb
+++ b/app/models/form/lettings/questions/stock_owner.rb
@@ -16,20 +16,38 @@ class Form::Lettings::Questions::StockOwner < ::Form::Question
     return answer_opts unless log
 
     if log.owning_organisation_id.present?
-      answer_opts = answer_opts.merge({ log.owning_organisation.id => log.owning_organisation.name })
+      answer_opts[log.owning_organisation.id] = log.owning_organisation.name
     end
 
+    recently_absorbed_organisations = user.organisation.absorbed_organisations.merged_during_open_collection_period
     if !user.support? && user.organisation.holds_own_stock?
-      answer_opts[user.organisation.id] = "#{user.organisation.name} (Your organisation)"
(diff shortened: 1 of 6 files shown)


app/models/form/lettings/questions/stock_owner.rb
```diff
@@ -16,20 +16,38 @@ def answer_options(log = nil, user = nil)
...
     if log.owning_organisation_id.present?
-      answer_opts = answer_opts.merge({ log.owning_organisation.id => log.owning_organisation.name })
+      answer_opts[log.owning_organisation.id] = log.owning_organisation.name
     end
 
+    recently_absorbed_organisations = user.organisation.absorbed_organisations.merged_during_open_collection_period
     if !user.support? && user.organisation.holds_own_stock?
-      answer_opts[user.organisation.id] = "#{user.organisation.name} (Your organisation)"
+      answer_opts[user.organisation.id] = if recently_absorbed_organisations.exists?
+                                            "#{user.organisation.name} (Your organisation, active as of #{user.organisation.created_at.to_fs(:govuk_date)})"
```
RachaelBooth: This looks like it'll be a bit weird if the absorbing organisation already existed with users before the merge - both because they'd see this and because the date will be further in the past than merged-in users would expect (although presumably that's something we could adjust when we do work around not relying on the created_at date?).

kosiakkatrina: Ah, good point! I'll get the available from change merged first and will update this.
I think we might need to update It on managing org and sales logs org pages as well.
If the org has available from date, I'll display it here, otherwise I'll just leave it as `(Your organisation)`

kosiakkatrina: Updated this now!

davidm-m (approved):

PR #2000, opened 2023-10-23
CLCD-2836 Use available from date for absorbing organisations validations

kosiakkatrina:
If an organisation has absorbed other organisations we run some validations on the start date for logs.
Up until now we only expected the merges to form a new organisation, so logs could be created for any merged organisations before the merge date and only for the new organisation after the creation of the new (absorbing) organisation date (which we intended to be the same as merge date).

These validations would also work if a set of organisations merged into an existing organisation if the created_at date for that organisation represented the date from which the organisation was available. Currently that's not the case and created_at date represents the date the organisation has been created on our system.

This PR adds available_from date to organisation which will used for validations on start date for absorbing organisation logs. We will not run date validations for an absorbing organisation unless this date is provided

### DIFF
--- a/spec/models/validations/setup_validations_spec.rb
+++ b/spec/models/validations/setup_validations_spec.rb
@@ -147,14 +147,14 @@ RSpec.describe Validations::SetupValidations do
       end
 
       context "and owning organisation is no longer active" do
-        it "does not allow startate after organisation has been merged" do
+        it "does not allow startdate after organisation has been merged" do
           record.startdate = Time.zone.local(2023, 3, 1)
           record.owning_organisation_id = merged_organisation.id
           setup_validator.validate_startdate_setup(record)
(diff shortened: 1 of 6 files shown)


spec/models/validations/setup_validations_spec.rb
```diff
@@ -163,19 +163,27 @@
       end
 
       context "and owning organisation is not yet active during the startdate" do
-        it "does not allow startate before absorbing organisation has been created" do
+        it "does not allow startate before absorbing organisation has become available" do
```
RachaelBooth: This hasn't changed, but these all have a typo in `startdate` that it seems worth fixing.

RachaelBooth (approved):

spec/models/validations/setup_validations_spec.rb
kosiakkatrina: good spot!

PR #2001, opened 2023-10-24
CLDC-2891 Allow refuge to be selected with internal transfer

kosiakkatrina:
We want it to be possible to select refuge for “Where was the household immediately before this letting?” and “Internal transfer” for “What was the source of referral for this letting”

### DIFF
--- a/app/models/validations/household_validations.rb
+++ b/app/models/validations/household_validations.rb
@@ -81,7 +81,7 @@ module Validations::HouseholdValidations
     # 27 Owner occupation (low-cost home ownership)
     # 28 Living with Friends or Family
     # 29 Prison / Approved Probation Hostel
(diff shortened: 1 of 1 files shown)


app/models/validations/household_validations.rb
```diff
@@ -81,7 +81,7 @@ def validate_previous_housing_situation(record)
     # 27 Owner occupation (low-cost home ownership)
     # 28 Living with Friends or Family
     # 29 Prison / Approved Probation Hostel
-    if record.is_internal_transfer? && [3, 4, 7, 10, 13, 14, 19, 21, 23, 24, 25, 26, 27, 28, 29].include?(record.prevten)
+    if record.is_internal_transfer? && [3, 4, 7, 10, 13, 14, 19, 23, 24, 25, 26, 27, 28, 29].include?(record.prevten)
```
davidm-m, line 83: Can you update the comment above?

kosiakkatrina: Ah, very good point

davidm-m (approved):

PR #2004, opened 2023-10-25
CLDC-2895 Add squish to the names of locations, schemes, users, and organisations

davidm-m:
Easier than I expected, we're already trimming these so we just need to change that to squish.

### DIFF
--- a/lib/tasks/squish_names.rake
+++ b/lib/tasks/squish_names.rake
@@ -1,0 +1,19 @@
+desc "Squish names of locations, schemes, users, and organisations"
+task squish_names: :environment do
+  Location.find_each do |location|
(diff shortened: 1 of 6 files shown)


lib/tasks/squish_names.rake
```diff
@@ -0,0 +1,19 @@
+desc "Squish names of locations, schemes, users, and organisations"
+task squish_names: :environment do
```
kosiakkatrina, line 2: Any chance we could get some tests?

davidm-m: Absolutely :+1:

kosiakkatrina (approved):

PR #2005, opened 2023-10-25
CLDC-2967 Infer lar and irproduct values

kosiakkatrina:
- lar and irproduct values were not being set. Now we infer them based on rent_type
- add a task to update currently wrong lar values:
  - correct any logs that did not have lar values, but were supposed to
  - correct any logs that had wrong lar values
  - lar was being set to 2 (no) for all the imported logs that didn't have it set to 1 (yes). We only want lar to be 2 if it's affordable rent, in other cases we want it to be nil. Because of this being most of the logs, we will no be setting these to export.
- add a task to update currently wrong irproduct values


lib/tasks/recalculate_irproduct_values.rake
```diff
@@ -0,0 +1,18 @@
...
+  end
+  LettingsLog.exportable.where(rent_type: 3).where.not(irproduct: 1).each do |log| # irproduct was set wrong
+    Rails.logger.info("Could not update irproduct for LettingsLog #{log.id}") unless log.update(values_updated_at: Time.zone.now)
+  end
+  LettingsLog.exportable.where(rent_type: 4).where.not(irproduct: 2).each do |log| # irproduct was set wrong
+    Rails.logger.info("Could not update irproduct for LettingsLog #{log.id}") unless log.update(values_updated_at: Time.zone.now)
+  end
+  LettingsLog.exportable.where(rent_type: 5).where.not(irproduct: 3).each do |log| # irproduct was set wrong
+    Rails.logger.info("Could not update irproduct for LettingsLog #{log.id}") unless log.update(values_updated_at: Time.zone.now)
+  end
```
kosiakkatrina, line 14: There are none of these on prod atm, but if rent_type is changed, these could happen

lib/tasks/recalculate_irproduct_values.rake
```diff
@@ -0,0 +1,18 @@
...
+  LettingsLog.exportable.where.not(rent_type: [3, 4, 5]).where.not(irproduct: nil).each do |log| # irproduct was set when it should have been nil
+    Rails.logger.info("Could not update irproduct for LettingsLog #{log.id}") unless log.update(values_updated_at: Time.zone.now)
+  end
```
kosiakkatrina, line 17: Currently only 1 occurrence of this

lib/tasks/recalculate_irproduct_values.rake
```diff
@@ -0,0 +1,18 @@
+desc "Forces to recalculate irproduct values"
+task recalculate_irproduct_values: :environment do
+  LettingsLog.exportable.where(rent_type: [3, 4, 5]).where(irproduct: nil).each do |log| # irproduct was never set
+    Rails.logger.info("Could not update irroduct for LettingsLog #{log.id}") unless log.update(values_updated_at: Time.zone.now)
```
davidm-m: Small type here in irproduct

davidm-m (approved):
One tiny typo but doesn't need a re-review, looks great!

lib/tasks/recalculate_irproduct_values.rake
kosiakkatrina: Ah, good spot, thanks!

PR #2006, opened 2023-10-25
CLDC-2862: Prepare maintenance page for PaaS app

SamSeed-Softwire:
https://dluhcdigital.atlassian.net/browse/CLDC-2862

This implementation may be a little naive, but given how experienced people on the team are with Rails I thought it made sense to raise a PR sooner rather than later.


app/controllers/application_controller.rb
```diff
@@ -3,8 +3,17 @@ class ApplicationController < ActionController::Base
 
   rescue_from Pundit::NotAuthorizedError, with: :render_not_authorized
 
+  before_action :check_maintenance
   before_action :set_paper_trail_whodunnit
 
+  def check_maintenance
+    if FeatureToggle.maintenance_mode_enabled? && request.fullpath.split("?")[0].delete("/") != "service-unavailable"
```
SamSeed-Softwire: The addition of the `request.fullpath.split("?")[0].delete("/") != "service-unavailable"` condition was to stop the app getting stuck in a loop (it wouldn't load the page without this).

app/services/feature_toggle.rb
```diff
@@ -37,4 +37,8 @@ def self.deduplication_flow_enabled?
   def self.duplicate_summary_enabled?
     !Rails.env.production?
   end
+
+  def self.maintenance_mode_enabled?
+    true
```
SamSeed-Softwire: I guess we might want to use an environment variable here instead? That can be set to true or false in the deployment pipelines. Wanted to sanity check the best approach with others before implementing that though.

app/views/maintenance/service_unavailable.html.erb
```diff
@@ -0,0 +1,11 @@
+<h1 class="govuk-heading-l govuk-!-width-two-thirds">
+  Sorry, the service is unavailable
+</h1>
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds-from-desktop">
+    <p class="govuk-body">You will be able to use the service later from TIME on DAY.</p>
+    <p class="govuk-body">We saved your answers.</p>
+    <p class="govuk-body"><%= govuk_link_to "Contact the helpdesk", "https://dluhcdigital.atlassian.net/servicedesk/customer/portal/6/group/11" %> if you need assistance in the meantime.</p>
```
SamSeed-Softwire: I've asked Susanna to confirm she's happy with this content.

app/views/maintenance/service_unavailable.html.erb
```diff
@@ -0,0 +1,11 @@
+<h1 class="govuk-heading-l govuk-!-width-two-thirds">
+  Sorry, the service is unavailable
+</h1>
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds-from-desktop">
+    <p class="govuk-body">You will be able to use the service later from TIME on DAY.</p>
+    <p class="govuk-body">We saved your answers.</p>
```
SamSeed-Softwire: Here's my current understanding of whether answers get saved or not:
- If maintenance mode gets activated whilst a user is on a page, anything they've done on that page (e.g. making a change to an answer on a log, or submitting a bulk upload) will be lost.
- Otherwise data is saved.

That seems to be the case from my testing and my understanding of what's going on, but I wanted to check here with people more experienced with Rails and this codebase in case there are any edge cases I'm missing.

Assuming I'm right, what do people think would be sensible content here? "We have not saved any changes you made immediately prior to the site undergoing maintenance"?

app/views/maintenance/service_unavailable.html.erb
```diff
@@ -0,0 +1,11 @@
+<h1 class="govuk-heading-l govuk-!-width-two-thirds">
+  Sorry, the service is unavailable
+</h1>
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds-from-desktop">
+    <p class="govuk-body">You will be able to use the service later from TIME on DAY.</p>
```
SamSeed-Softwire: This will need confirming and updating.

spec/requests/maintenance_controller_spec.rb
```diff
@@ -0,0 +1,54 @@
...
+        get "/lettings-logs"
+        expect(response).not_to redirect_to(service_unavailable_path)
+        expect(page).to have_content("Create a new lettings log")
+      end
+    end
+
+    context "when a user visits the maintenance page" do
+      it "redirects the user to the start page" do
+        get "/service-unavailable"
+        expect(response).to redirect_to(root_path)
```
SamSeed-Softwire, line 75: I wasn't able to simulate it redirecting 'back' (in `app/controllers/application_controller.rb` we have `redirect_back(fallback_location: root_path)`). I think this is fine - I'm not trying to test that the `redirect_back` method works - I'm trying to test that the redirect gets triggered correctly.

spec/requests/maintenance_controller_spec.rb
```diff
@@ -0,0 +1,54 @@
...
+    before do
+      allow(FeatureToggle).to receive(:maintenance_mode_enabled?).and_return(true)
+    end
+
+    context "when a user visits a page other than the maintenance page" do
+      it "redirects the user to the maintenance page" do
+        get "/lettings-logs"
+        expect(response).to redirect_to(service_unavailable_path)
+        follow_redirect!
+        expect(page).to have_content("Sorry, the service is unavailable")
```
SamSeed-Softwire, line 24: I haven't tested that the user is signed out here, because I tested that in `spec/controllers/maintenance_controller_spec.rb`.

app/views/maintenance/service_unavailable.html.erb
RachaelBooth: I agree with your understanding on what does and doesn't get saved. 

I think if I was in the middle of filling out a log though and was told that my answers hadn't been saved (including with any immediately prior wording), I'd expect the entire log's worth of answers not to be saved. Although I guess the buttons on each question do say "save and continue". Probably worth asking Susannah about wording.

RachaelBooth (commented):
Were we going to disable scheduled jobs as part of this, or is that for a future PR?

RachaelBooth:
Presumably testing this will require some co-ordination of people using the review app and turning the maintenance mode on.

SamSeed-Softwire:
> Presumably testing this will require some co-ordination of people using the review app and turning the maintenance mode on.

I tested it locally by doing this and was happy with everything. But yes I imagine whoever tests this will want to do the same.

SamSeed-Softwire:
> Were we going to disable scheduled jobs as part of this, or is that for a future PR?

@RachaelBooth ah, forgot about that (it's not mentioned in the ticket and I don't think it's in any other tickets btw). I will implement now as part of this PR.

RachaelBooth:
> > Were we going to disable scheduled jobs as part of this, or is that for a future PR?
> 
> @RachaelBooth ah, forgot about that (it's not mentioned in the ticket and I don't think it's in any other tickets btw). I will implement now as part of this PR.

It's vaguely mentioned in a comment on this ticket as an "either here or another ticket" thing - but we've not got another ticket for it.

app/controllers/application_controller.rb
```diff
@@ -3,8 +3,17 @@ class ApplicationController < ActionController::Base
 
   rescue_from Pundit::NotAuthorizedError, with: :render_not_authorized
 
+  before_action :check_maintenance
   before_action :set_paper_trail_whodunnit
 
+  def check_maintenance
+    if FeatureToggle.maintenance_mode_enabled? && !%w[service-unavailable accessibility-statement privacy-notice cookies].include?(request.fullpath.split("?")[0].delete("/"))
```
SamSeed-Softwire, line 10: The exclusion of the `service-unavailable` path was to stop the app getting stuck in a loop (it wouldn't load the page without this).

SamSeed-Softwire, line 10: I've also allowed the user to interact with the cookie banner and see the cookies page, as well as the accessibility statement and privacy notice, as I assumed these are bits of information that we should probably make available even in maintenance mode. If you disagree and would prefer these to be hidden I can also do that.

config/initializers/sidekiq.rb
```diff
@@ -32,10 +32,13 @@
...
   config.on(:startup) do
-    Sidekiq::Cron::Job.load_from_hash YAML.load_file("config/sidekiq_cron_schedule.yml")
+    unless FeatureToggle.maintenance_mode_enabled?
+      Sidekiq::Cron::Job.load_from_hash YAML.load_file("config/sidekiq_cron_schedule.yml")
+    end
   end
 
   config.on(:shutdown) do
     Sidekiq::CLI.instance.launcher.quiet
+    Sidekiq::Cron::Job.all.each(&:destroy)
```
SamSeed-Softwire: Without this line, jobs will not be destroyed during a deployment. So enabling maintenance mode wouldn't turn off the data export xml job. Also on staging right now there is still the data export csv job running (and failing) every day - it would be good to clean bits like that up.

We could probably also destroy the jobs in the startup block, but I don't think it matters. Also we could specify specific jobs to destroy instead of all of them, but it seems wise to me to destroy everything!

app/controllers/application_controller.rb
SamSeed-Softwire: Note I haven't written tests for this because I wanted to get confirmation first that we did indeed want to retain these links in maintenance mode.

config/initializers/sidekiq.rb
SamSeed-Softwire: Note that AFACIT there were no tests for this sidekiq config/job stuff. Seems wise for me to write some, shout if you think that's not worth the time right now though.

RachaelBooth: Does this apply every time an app instance is shut down?

app/controllers/application_controller.rb
RachaelBooth: I think that makes sense.

config/initializers/sidekiq.rb
SamSeed-Softwire: As agreed, we want to move this line into the server startup section. There's a chance that if a server starts the moment before the job is supposed to run then the job will get deleted and not recreated in time so the job won't get run. We think this is a tiny risk and is easily outweighed by the advantages of ensuring the jobs which exists are aligned with what's defined in the code.

app/services/feature_toggle.rb
SamSeed-Softwire: When we enable it, we'll do it by specifying which environments it should be enabled in (like all the other feature toggles).

app/controllers/application_controller.rb
SamSeed-Softwire: I've written tests for these now.

spec/requests/maintenance_controller_spec.rb
```diff
@@ -0,0 +1,85 @@
...
+        get "/service-unavailable"
+      end
+
+      it "redirects the user to the start page" do
+        expect(response).to redirect_to(root_path)
+      end
+
+      it "the cookie banner is visible" do
+        follow_redirect!
+        follow_redirect!
```
SamSeed-Softwire, line 80: I had to follow the redirect twice here in order to reach a page which wasn't just a "You are being redirected." page

spec/requests/maintenance_controller_spec.rb
```diff
@@ -0,0 +1,85 @@
...
+        get "/lettings-logs"
+      end
+
+      it "redirects the user to the maintenance page" do
+        expect(response).to redirect_to(service_unavailable_path)
+        follow_redirect!
+        expect(page).to have_content("Sorry, the service is unavailable")
+      end
+
+      it "the cookie banner is visible" do
```
SamSeed-Softwire, line 27: I could imagine it being sensible to move these cookie banner tests into another file (e.g. `spec/views/layouts/application_layout_spec.rb`, although I'm not convinced that is the right place) but I thought they were probably fine here. Happy to be persuaded otherwise though.

spec/requests/cookies_controller_spec.rb
```diff
@@ -0,0 +1,42 @@
+require "rails_helper"
+
+RSpec.describe CookiesController, type: :request do
```
SamSeed-Softwire, line 3: There are other aspects of the cookies controller which I'm not testing here (some of the actual behaviour of the banner), but I thought that was out of scope for this ticket.

config/initializers/sidekiq.rb
```diff
@@ -32,7 +32,10 @@
 
 Sidekiq.configure_server do |config|
```
SamSeed-Softwire, line 33: There were no existing tests for this initializer file (nor for most initialisers). I'm not convinced writing a test here to check that Sidekiq can start up a job successfully is that worth the time. There's obviously the question of whether the DataExportXMLJob class works correctly (and there's already a test for that) so I guess the final piece would be to check that `config/sidekiq_cron_schedule.yml` is defined correctly, however I don't think that's worth the time investment now (if we do want to do that, I think we should raise a separate ticket for that). Do you agree?

RachaelBooth: That's definitely out of scope for this (given there are currently no tests and it's not really a thing where there's an obviously worthwhile test to do) - it's an app team problem to deal with later (or to decide not to).

SamSeed-Softwire: I raised a ticket for this then: https://dluhcdigital.atlassian.net/browse/CLDC-2994.

In it I do mention that we might not want tests at all - it's more a placeholder for making the decision, so it doesn't get forgotten.

RachaelBooth (approved):

RachaelBooth:
Should make sure someone explicitly does some testing of this before merging (I've looked at the code, but not the review app with the maintenance mode on)

SamSeed-Softwire:
> Should make sure someone explicitly does some testing of this before merging (I've looked at the code, but not the review app with the maintenance mode on)

@RachaelBooth I've tested this and happy with it, but it should probably have another pair of eyes on it - will ask the team.

SamSeed-Softwire:
> > Should make sure someone explicitly does some testing of this before merging (I've looked at the code, but not the review app with the maintenance mode on)
> 
> @RachaelBooth I've tested this and happy with it, but it should probably have another pair of eyes on it - will ask the team.

@RachaelBooth Nat has tested this and added his comments to the [ticket](https://dluhcdigital.atlassian.net/browse/CLDC-2862). As agreed he has not tested the sidekiq job enqueueing functionality.

PR #2009, opened 2023-10-26
CLDC- 2268 Validate using correct deactivation periods

kosiakkatrina:
We validate whether the scheme/location is active during the tenancy startdate using status_at(date) method.
This method has sometimes been incorrectly returning reactivating_soon status instead of active, therefore preventing some schemes/locations from being selected.
This would happen because we always looked at the most recent deactivation to determine reactivating_soon status, but depending on the status, the most recent deactivation might not have been relevant.

For example, if the tenancy startdate is 18th November 2022 and the last deactivation that has been reactivated happened 20th November 2022 - 22nd August 2023, we would look at this deactivation, see that the scheme/location reactivates in August 2023 and mark it as reactivating_soon.

Now we only look at deactivations that happened before the given date and find the most recent one of those to determine reactivating soon status.

### DIFF
--- a/spec/models/location_spec.rb
+++ b/spec/models/location_spec.rb
@@ -930,6 +930,29 @@ RSpec.describe Location, type: :model do
     end
   end
 
+  describe "status_at" do
+    let(:location) { FactoryBot.build(:location, startdate: Time.zone.local(2022, 4, 1)) }
+
+    before do
+      Timecop.freeze(2022, 6, 7)
+    end
+
+    after do
+      Timecop.unfreeze
+    end
+
+    context "when there have been previous deactivations" do
+      before do
+        FactoryBot.create(:location_deactivation_period, deactivation_date: Time.zone.local(2022, 5, 4), reactivation_date: Time.zone.local(2022, 6, 5), location:)
+        location.save!
+      end
+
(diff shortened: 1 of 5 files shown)


spec/models/location_spec.rb
```diff
@@ -930,6 +930,29 @@
...
+    context "when there have been previous deactivations" do
+      before do
+        FactoryBot.create(:location_deactivation_period, deactivation_date: Time.zone.local(2022, 5, 4), reactivation_date: Time.zone.local(2022, 6, 5), location:)
+        location.save!
+      end
+
+      it "returns active if the location has no relevant deactivation records" do
+        expect(location.status_at(Time.zone.local(2022, 4, 4))).to eq(:active)
+      end
+    end
```
davidm-m, line 953: Is it worth adding in tests for when this should return other statuses too?

kosiakkatrina: I've added them initially, but they're all sort of implicitly being tested by
```
  def status
    @status ||= status_at(Time.zone.now)
  end
```

davidm-m: Hmm, yeah good point. All looks good then!

davidm-m (approved):

PR #2012, opened 2023-10-30
CLDC-2864: Allow PaaS-only deployments

SamSeed-Softwire:
Basically all I've done is duplicate the staging pipeline and tinker with it a bit. To be precise, here are the non-trivial changes that I've made:

- Only allowed the pipeline to be manually triggered
- Changed all 'staging' parameters/strings to 'production' 
- Stop enforcing that the deploy job can only happen on main
- Removed the `API_USER` and `API_KEY` env vars (to match prod)
- Removed the 180s timeout from the `cf push` command (to match prod)

Here's how I think the GitHub actions stuff should go on switchover day:

1. Disable the normal prod pipeline via the GitHub UI.
2. Create a new branch in which we turn on maintenance mode for the production environment.
3. Enable the PaaS-only pipeline via the GitHub UI. (Does it also need to be updated to only run on the new branch created? Or can we choose which branch to trigger it on via the UI?)
4. Trigger the PaaS-only pipeline via the GitHub UI. This will turn on maintenance mode for the PaaS production app.
5. Later on (I'm ignoring various non-GitHub actions steps here) we push a commit to the branch which updates some more feature flag logic to turn on redirect mode (i.e. instead of the 'service unavailable' page we will now show the 'redirect' page) for the production environment. This hasn't been built yet, but that doesn't matter for the sake of this plan.
6. Trigger the PaaS-only pipeline via the GitHub UI again. This will turn on redirect mode for the PaaS production app.
7. On another new branch, update the regular prod pipeline so that it only deploys to the AWS infra (and then merge it to main via a PR). Then re-enable it via the GitHub UI.

### DIFF
--- a/.github/workflows/paas_only_production_pipeline.yml
+++ b/.github/workflows/paas_only_production_pipeline.yml
@@ -1,4 +1,4 @@
-name: Staging CI/CD Pipeline
+name: PaaS-only Production CI/CD Pipeline
 
 on:
   workflow_dispatch:
--- a/.github/workflows/paas_only_production_pipeline.yml
+++ b/.github/workflows/paas_only_production_pipeline.yml
@@ -259,15 +259,3 @@ jobs:
           cf set-env $APP_NAME CSV_DOWNLOAD_PAAS_INSTANCE $CSV_DOWNLOAD_PAAS_INSTANCE
           cf set-env $APP_NAME SENTRY_DSN $SENTRY_DSN
           cf push $APP_NAME --strategy rolling
-
-  aws_deploy:
-    name: AWS Deploy
-    needs: [lint, test, feature_test, audit]
-    uses: ./.github/workflows/aws_deploy.yml
-    with:
-      aws_account_id: 977287343304
-      aws_resource_prefix: core-prod
-      environment: production
-      release_tag: ${{ needs.test.outputs.releasetag }}
-    permissions:
-      id-token: write
--- a/.github/workflows/paas_only_production_pipeline.yml
+++ b/.github/workflows/paas_only_production_pipeline.yml
@@ -174,7 +174,6 @@ jobs:
     concurrency: staging
     runs-on: ubuntu-latest
     environment: staging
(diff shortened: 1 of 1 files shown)


RachaelBooth:
Much as keeping this as similar as possible to the existing prod deployment pipeline makes sense, I'm not sure it really makes sense to be deploying from a github release - or in fact from the main branch. What do you think about the possibility of these deployments being something that we trigger from a different branch?

Re: how github actions things go on the day, it is possible to disable and re-enable workflows through the github UI - I was imagining doing that rather than PRs to avoid unnecessary code changes (this is what's been done when turning off the pipeline during org migrations). Although we can't avoid needing to push changes to toggles to somewhere.

SamSeed-Softwire:
> Much as keeping this as similar as possible to the existing prod deployment pipeline makes sense, I'm not sure it really makes sense to be deploying from a github release - or in fact from the main branch. What do you think about the possibility of these deployments being something that we trigger from a different branch?
> 
> Re: how github actions things go on the day, it is possible to disable and re-enable workflows through the github UI - I was imagining doing that rather than PRs to avoid unnecessary code changes (this is what's been done when turning off the pipeline during org migrations). Although we can't avoid needing to push changes to toggles to somewhere.

Sounds sensible, thanks for the steer! Will update.

SamSeed-Softwire:
> > Much as keeping this as similar as possible to the existing prod deployment pipeline makes sense, I'm not sure it really makes sense to be deploying from a github release - or in fact from the main branch. What do you think about the possibility of these deployments being something that we trigger from a different branch?
> > Re: how github actions things go on the day, it is possible to disable and re-enable workflows through the github UI - I was imagining doing that rather than PRs to avoid unnecessary code changes (this is what's been done when turning off the pipeline during org migrations). Although we can't avoid needing to push changes to toggles to somewhere.
> 
> Sounds sensible, thanks for the steer! Will update.

I've done this now, and have updated my plan in the PR description accordingly.

RachaelBooth (approved):

PR #2014, opened 2023-11-01
CLDC-2636 Fix enter address instead bug

natdeanlewissoftwire:
The link on the "enter address instead" button was correct, but we weren't routing the address page if the `uprn_known` hadn't been answered, which was causing it to redirect to the log tasklist. This PR fixes this routing.

ticket: https://dluhcdigital.atlassian.net/browse/CLDC-2636

### DIFF
--- a/spec/factories/lettings_log.rb
+++ b/spec/factories/lettings_log.rb
@@ -3,7 +3,10 @@ FactoryBot.define do
     created_by { FactoryBot.create(:user) }
     owning_organisation { created_by.organisation }
     managing_organisation { created_by.organisation }
+    created_at { Time.zone.today }
+    updated_at { Time.zone.today }
     trait :setup_completed do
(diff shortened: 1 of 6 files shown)


spec/factories/lettings_log.rb
```diff
@@ -3,7 +3,10 @@
     created_by { FactoryBot.create(:user) }
     owning_organisation { created_by.organisation }
     managing_organisation { created_by.organisation }
+    created_at { Time.zone.today }
+    updated_at { Time.zone.today }
     trait :setup_completed do
+      startdate_today
```
natdeanlewissoftwire, line 9: unrelated, noticed this while I was testing this ticket - startdate is part of a setup-completed lettings log so should be here

RachaelBooth (approved):

PR #2016, opened 2023-11-03
CLDC-2966 Fix bulk upload scheme lookup bug

natdeanlewissoftwire:
We now lookup schemes and locations by looking for a matching combination of: scheme id, location id, owning org id (either the owning or managing org for the log). This ensures we return the correct combination of scheme and location - before we were returning any scheme and location that had the correct `old_visible_id` when users used old CORE ids which gave erroneous results.

This PR also attempts to return a scheme even when location id is not present (or invalid) so that the error message only applies to location in this case, even though logs cannot be created in this situation anyway.

Comes from this support ticket: https://dluhcdigital.atlassian.net/browse/CLDC-2966

### DIFF
--- a/app/services/bulk_upload/lettings/year2022/row_parser.rb
+++ b/app/services/bulk_upload/lettings/year2022/row_parser.rb
@@ -354,12 +354,10 @@ class BulkUpload::Lettings::Year2022::RowParser
   validate :validate_managing_org_exists, on: :after_log
   validate :validate_managing_org_related, on: :after_log
 
-  validate :validate_scheme_related, on: :after_log
-  validate :validate_scheme_exists, on: :after_log
+  validate :validate_related_scheme_exists, on: :after_log
   validate :validate_scheme_data_given, on: :after_log
 
-  validate :validate_location_related, on: :after_log
-  validate :validate_location_exists, on: :after_log
+  validate :validate_related_location_exists, on: :after_log
   validate :validate_location_data_given, on: :after_log
 
   validate :validate_created_by_exists, on: :after_log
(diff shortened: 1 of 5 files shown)


app/services/bulk_upload/lettings/year2022/row_parser.rb
```diff
@@ -530,53 +528,36 @@ def duplicate_check_fields
...
 
     @location ||= scheme.locations.find_by_id_on_multiple_fields(field_5)
   end
 
-  def validate_location_exists
+  def validate_related_location_exists
     if scheme && field_5.present? && location.nil?
-      errors.add(:field_5, "Location could not be found with the provided scheme code", category: :setup)
+      block_log_creation!
+      errors.add(:field_5, "Scheme code must relate to a scheme that is owned by the owning organisation or managing organisation", category: :setup)
```
RachaelBooth, line 540: Should this be talking about the location?

RachaelBooth (commented):
This looks like it adds more ways for log creation to be blocked, but in ways that make sense to me so as long as they are deliberate that's fine.

app/services/bulk_upload/lettings/year2022/row_parser.rb
natdeanlewissoftwire: I'm pretty sure this was an error in copy before - this is for the 2022 form that only refers to management groups and old-core schemes (i.e. locations) so the error message should match that I think.

natdeanlewissoftwire:
And yep, it does add more log creation blockers, I ran them by Jemma this morning and she agrees they should have been there to begin with. In practice I think another error would have blocked log creation regardless in the case we were missing scheme/location so not sure it'll make any difference but good to be explicit

RachaelBooth (approved):

PR #2019, opened 2023-11-06
Fix flaky tests

natdeanlewissoftwire:


spec/services/merge/merge_organisations_service_spec.rb
```diff
@@ -141,7 +141,6 @@
         let!(:owned_lettings_log_no_location) { create(:lettings_log, :sh, scheme:, startdate: Time.zone.tomorrow, owning_organisation: merging_organisation) }
 
         before do
-          create(:location, scheme:, name: "fake location", postcode: "A1 1AA")
           create(:location, scheme: deactivated_scheme)
```
SamSeed-Softwire, line 144: What's this line for?

SamSeed-Softwire (approved):
As discussed, no need to use `Timecop.travel`, and happy for this to be merged once the relevant `create(:location, scheme: deactivated_scheme)` lines have been removed from the `before do` blocks.

natdeanlewissoftwire:
My mistake on those lines in our discussion earlier - they don’t affect the tests I was fixing but that lcoation object IS referenced in later tests, where it’s used to check that the deactivated scheme has one active and one inactive location, which does provide value so I’m leaving them in after all

PR #2020, opened 2023-11-06
Make linting consistent in staging and prod pipelines

SamSeed-Softwire:
The prod pipeline was using `bundle exec rubocop`, but the staging pipeline was using `bundle exec rake lint` (which includes `bundle exec rubocop` along with some other linting commands). Discussed with @natdeanlewissoftwire and agreed it would make sense for prod to be updated to match staging.

### DIFF
--- a/.github/workflows/production_pipeline.yml
+++ b/.github/workflows/production_pipeline.yml
@@ -167,9 +167,19 @@ jobs:
         with:
           bundler-cache: true
 
-      - name: Rubocop
+      - name: Set up Node.js
+        uses: actions/setup-node@v3
+        with:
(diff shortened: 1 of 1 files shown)


natdeanlewissoftwire (approved):
Looks good to me! (I'm assuming the yarn install is necessary for the rake lint)

natdeanlewissoftwire:
(I'm assuming the yarn install is necessary for the rake lint)

SamSeed-Softwire:
> Looks good to me! (I'm assuming the yarn install is necessary for the rake lint)

Yep, it is. Thanks for checking!

PR #2022, opened 2023-11-07
CLDC-2863: Create 'redirect to new service' page

SamSeed-Softwire:
Ticket: https://dluhcdigital.atlassian.net/browse/CLDC-2863

Previously I created a "Sorry, the service is unavailable" page for when we are doing the switchover. Once the switchover is complete and we are ready for users to use the app in the new AWS infra, we will present a new "The URL for this service has changed." page (i.e. a 'redirect' page) to users when they try to visit the app in the PaaS infra. This PR creates that new page.

It might be worth looking at the PR in two stages - the first of which is the first 3 commits in which I refactored the "Sorry, the service is unavailable" page code. Basically I just renamed `maintenance_mode` to `service_unavailable` as I thought that was more clear, and it also was cleaner when adding code for the new redirect page (the second stage of this work - described next).

The code for the new redirect page is almost identical to the `service_unavailable` code, but instead it's called `service_moved`. (Hence hiding the initial refactor commits when reviewing this part might make things cleaner.) The only important difference is that if both feature toggles are turned on at the same time then the `service_moved` feature has priority. This is because on switchover day we will be turning on `service_unavailable` first and then `service_moved` later, so I wanted to prepare for the case where we forget to turn off `service_unavailable`.

Of course I added tests for when the `service_moved` feature is on, but because it's possible for both feature toggles to be on at the same time I added tests for this possibility as well.

Thanks in advance!

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -3,10 +3,10 @@ class ApplicationController < ActionController::Base
 
   rescue_from Pundit::NotAuthorizedError, with: :render_not_authorized
 
-  before_action :check_maintenance
+  before_action :check_maintenance_status
   before_action :set_paper_trail_whodunnit
 
-  def check_maintenance
+  def check_maintenance_status
     if FeatureToggle.service_unavailable? && !%w[service-unavailable accessibility-statement privacy-notice cookies].include?(request.fullpath.split("?")[0].delete("/"))
       redirect_to service_unavailable_path
     elsif !FeatureToggle.service_unavailable? && request.fullpath.split("?")[0].delete("/") == "service-unavailable"
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -7,9 +7,15 @@ class ApplicationController < ActionController::Base
   before_action :set_paper_trail_whodunnit
 
   def check_maintenance_status
-    if FeatureToggle.service_unavailable? && !%w[service-unavailable accessibility-statement privacy-notice cookies].include?(request.fullpath.split("?")[0].delete("/"))
-      redirect_to service_unavailable_path
-    elsif !FeatureToggle.service_unavailable? && request.fullpath.split("?")[0].delete("/") == "service-unavailable"
+    if FeatureToggle.service_moved?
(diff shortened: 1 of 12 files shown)


app/controllers/application_controller.rb
```diff
@@ -7,9 +7,15 @@ class ApplicationController < ActionController::Base
   before_action :set_paper_trail_whodunnit
 
   def check_maintenance_status
-    if FeatureToggle.service_unavailable? && !%w[service-unavailable accessibility-statement privacy-notice cookies].include?(request.fullpath.split("?")[0].delete("/"))
-      redirect_to service_unavailable_path
-    elsif !FeatureToggle.service_unavailable? && request.fullpath.split("?")[0].delete("/") == "service-unavailable"
```
SamSeed-Softwire: I could have written the below logic without nesting single unless statements inside if/elsif statements, but I actually thought it was the clearest way to present the logic.

I wrote out a table of what behaviour we'd expect in each case and used that to determine the logic here.

image

app/controllers/application_controller.rb
```diff
@@ -3,13 +3,19 @@ class ApplicationController < ActionController::Base
...
-  before_action :check_maintenance
+  before_action :check_maintenance_status
   before_action :set_paper_trail_whodunnit
 
-  def check_maintenance
-    if FeatureToggle.maintenance_mode_enabled? && !%w[service-unavailable accessibility-statement privacy-notice cookies].include?(request.fullpath.split("?")[0].delete("/"))
-      redirect_to service_unavailable_path
-    elsif !FeatureToggle.maintenance_mode_enabled? && request.fullpath.split("?")[0].delete("/") == "service-unavailable"
+  def check_maintenance_status
+    if FeatureToggle.service_moved?
```
RachaelBooth, line 10: I've been wondering about some way to make this bit a little nicer, just to avoid having so many conditions, but I don't think it'd be worth the effort.

RachaelBooth (approved):
Cool, looks fine to me.

app/controllers/application_controller.rb
SamSeed-Softwire: I agree 👍

PR #2024, opened 2023-11-08
CLDC-3025 Fix leading zeroes bug in bulk upload

natdeanlewissoftwire:
Some schemes and locaitons have leading 0s in their `old_visible_id`s  because they did on Old CORE. This means when users type e.g. 1 in a bulk upload form scheme field, it won’t find a scheme with id “001” when it should.

This PR removes leading zeroes when we match by `old_visible_id` in bulk upload to address this, this is not necessary for organisation `old_visible_id`s as they do not have leading zeroes in any cases (I've checked)

ticket: https://dluhcdigital.atlassian.net/browse/CLDC-3025

### DIFF
--- a/app/models/location.rb
+++ b/app/models/location.rb
@@ -110,7 +110,7 @@ class Location < ApplicationRecord
   def self.find_by_id_on_multiple_fields(id)
     return if id.nil?
 
-    where(id:).or(where("ltrim(old_visible_id, '0') = ?", id.to_s)).first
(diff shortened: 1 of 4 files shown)


RachaelBooth (commented):
Just to check, when we parse the id that users have entered do we do so without any leading zeros they have put in?

natdeanlewissoftwire:
Good point - probably worth removing those zeroes too as well to be safe

RachaelBooth (approved):

PR #2026, opened 2023-11-08
Cldc 2812 run app from relative url

RachaelBooth:
N.B. This doesn't make tests work with a relative root - a number of our tests annoyingly hardcode a path, so that seemed to be more effort than it was worth.


RachaelBooth:
N.B. We won't be able to test this on a normal PaaS review app (although checking that app will test that this doesn't break the app when a relative root is not provided, which is also important). Ensuring everything works with a relative root set will have to be done in conjunction with testing deploying review apps to AWS.

PR #2027, opened 2023-11-08
CLDC-2962 hide uneditable duplicate sets

natdeanlewissoftwire:
Only show duplicate logs banner (and count) for sets that are from editable collection years.

ticket: https://dluhcdigital.atlassian.net/browse/CLDC-2962

### DIFF
--- a/app/models/organisation.rb
+++ b/app/models/organisation.rb
@@ -133,14 +133,23 @@ class Organisation < ApplicationRecord
     :active
   end
 
-  def duplicate_lettings_logs_sets
-    lettings_logs.duplicate_sets.map { |array_str| array_str ? array_str.map(&:to_i) : [] }
+  def editable_duplicate_lettings_logs_sets
+    all_sets = lettings_logs.duplicate_sets.map { |array_str| array_str ? array_str.map(&:to_i) : [] }
+    editable_sets = []
+    all_sets.each do |set|
+      editable_sets << set if LettingsLog.find(set.first).collection_period_open_for_editing?
+    end
+    editable_sets
   end
 
-  def duplicate_sales_logs_sets
-    sales_logs.duplicate_sets.map { |array_str| array_str ? array_str.map(&:to_i) : [] }
+  def editable_duplicate_sales_logs_sets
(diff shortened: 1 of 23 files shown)


RachaelBooth (approved):

app/models/organisation.rb
```diff
@@ -133,12 +133,14 @@ def status_at(date)
     :active
   end
 
-  def duplicate_lettings_logs_sets
+  def editable_duplicate_lettings_logs_sets
     lettings_logs.duplicate_sets.map { |array_str| array_str ? array_str.map(&:to_i) : [] }
+                 .select { |set| LettingsLog.find(set.first).collection_period_open_for_editing? }
   end
```
kosiakkatrina, line 138: I feel like instead of looping through all of the logs twice (map and select), maybe we could get the dates range first and chain it as a scope after `lettings_logs`? Cause this is probably quite slow already

natdeanlewissoftwire: Yeah I tried to add it as a scope, couldn't find a way of getting this done though as the collection_period_open_for_editing? method depends on the form's edit_end_date - Is there a way we can access that in a log scope?

kosiakkatrina: ~Would this `FormHandler.instance.start_date_of_earliest_open_for_editing_collection_period` or passing it in as a param to a scope work?~ Ah, sorry, it's the end date we want 🤔

kosiakkatrina: Could potentially add a method which is something like this to FormHandler?
``` 
def earliest_open_for_editing_collection_end_date(now: Time.zone.now)
    if in_edit_crossover_period?(now:)
      previous_lettings_form.edit_end_date
    else
      current_lettings_form.edit_end_date
    end
  end
```

Would need to define previous_lettings_form method, but the current is there and it would be log type specific as well

natdeanlewissoftwire: Have made this change, think I'll add some tests in the morning but seems to be working as desired, thanks! I think I'm right that we can use the startdate after all as the crossover period condition takes into account changes to the `edit_end_date`

kosiakkatrina (approved):

PR #2028, opened 2023-11-09
CLDC-3029 Clear not routed answers twice

natdeanlewissoftwire:
We clear not routed answers a second time after recalculating routing to catch newly not-routed questions.
We found this bug because when the household was reduced in size, the details known question was cleared but not the details as they were newly not-routed to only after the details known question had been cleared.

ticket: https://dluhcdigital.atlassian.net/browse/CLDC-3029

### DIFF
--- a/app/models/form.rb
+++ b/app/models/form.rb
@@ -192,54 +192,64 @@ class Form
 
   def reset_checkbox_questions_if_not_routed(log)
     checkbox_questions = routed_and_not_routed_questions_by_type(log, type: "checkbox")
-    checkbox_questions[:not_routed].each do |not_routed_question|
-      valid_options = checkbox_questions[:routed]
-                                        .select { |q| q.id == not_routed_question.id }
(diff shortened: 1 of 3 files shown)


RachaelBooth (commented):
Do we need to deal with interactions between questions of different types here? Or is that not actually an issue.

RachaelBooth (approved):
Am approving on the grounds of what this does do - happy for any interactions between types (if that is indeed a thing) to be thought about separately.

natdeanlewissoftwire:
Ah forgot to reply to the above before Rachael went off - but yes that's a good point - I'll refactor to do the first pass of all question types before the second pass of any

natdeanlewissoftwire:
On second thought as this fixes the issue at hand, and I'm not certain we NEED to consider interactions between different types (and changing it now would be a significant refactor) - I'm going to merge as is and focus on other higher priority work at the moment, will come back to this if it causes an issue down the line.

PR #2031, opened 2023-11-14
CLDC-2691 Miscellaneous navigation bug fixes

natdeanlewissoftwire:
This PR includes a number of changes. 

Primarily, it adds breadcrumbs back to an individual org's lettings/sales/scheme/user page when a support user is viewing objects in an individual org. For logs in particular, this routes to the owning org or managing org, if one exists, and if not goes to the all logs page for a given type.

Non support users just see back links as normal.

Also changed:

- Tab titles that used to say "Logs" now have the log type
- "Logs" breadcrumb now specifies log type
- The secondary nav bar appears above flash notices for an individual org
- The support user sales search bar hint text is now correct (previously it was hard coded to display lettings text for both)
- Org details links now explicitly route to details page rather than going to `organisations/id` and being redirected to `organisations/id/details`
- refactoring of redundant parameter in `scheme_items`


app/helpers/logs_helper.rb
```diff
@@ -20,6 +29,13 @@ def search_label_for_controller(controller)
     end
   end
 
+  def search_label_for_action(action_name)
+    case log_type_for_action(action_name)
```
kosiakkatrina: Could we not have case action_name directly here?

app/views/organisations/logs.html.erb
```diff
@@ -1,10 +1,17 @@
 <% item_label = format_label(@pagy.count, "logs") %>
-<% title = format_title(@searched, "Logs", current_user, item_label, @pagy.count, @organisation.name) %>
+<% title = format_title(@searched, "#{log_type_for_action(action_name).capitalize} logs", current_user, item_label, @pagy.count, @organisation.name) %>
```
kosiakkatrina: Could this be `action_type.humanize`?

app/views/schemes/show.html.erb
```diff
@@ -1,17 +1,22 @@
...
-  ) %>
+<% if current_user.support? %>
+  <% content_for :breadcrumbs, govuk_breadcrumbs(breadcrumbs: {
+    "Schemes (#{@scheme.owning_organisation.name})" => schemes_organisation_path(@scheme.owning_organisation),
+    content_for(:title) => "",
+  }) %>
+<% else %>
+  <% content_for :before_content do %>
+    <%= govuk_back_link(href: schemes_path) %>
+  <% end %>
```
kosiakkatrina, line 12: On locations we've got `back` button and it feels odd that something other (breadcrumbs) than the tab content changes when switching between scheme and locations tabs on a specific scheme

app/views/users/show.html.erb
```diff
@@ -1,5 +1,16 @@
 <% content_for :title, current_user == @user ? "Your account" : "#{@user.name.presence || @user.email}’s account" %>
 
+<% if current_user.support? %>
+  <% content_for :breadcrumbs, govuk_breadcrumbs(breadcrumbs: {
+    "Users (#{@user.organisation.name})" => users_organisation_path(@user.organisation),
```
kosiakkatrina, line 5: `DLUHC Users` sounds a bit better than `Users (DLUHC)` to me but that's preference so no need to action

app/helpers/tasklist_helper.rb
```diff
@@ -43,6 +60,10 @@ def review_log_text(log)
 
 private
 
+  def breadcrumb_organisation(log)
+    log.owning_organisation || (log.managing_organisation if log.respond_to?(:managing_organisation))
+  end
+
```
kosiakkatrina, line 66: So if we navigate from all logs we still want to show the owning org in the breadcrumbs?

natdeanlewissoftwire: Yep, this is the compromise to avoid spending a load of time on routing now - to see all logs support users can just click the tab headings in the primary nav bar so this way all options are avaliable to them at all times. If support say they'd like a more specified behaviour I'll add an improvements ticket to add this in but I'm not expecting them to.

app/helpers/logs_helper.rb
natdeanlewissoftwire: Ah good shout

app/views/users/show.html.erb
natdeanlewissoftwire: Will run this by Olivia/Susanna :)

app/views/organisations/logs.html.erb
natdeanlewissoftwire: Yep, nice! Have done.

app/views/schemes/show.html.erb
natdeanlewissoftwire: Yep, I agree this is weird - have changed it now to this for support users: 
image

natdeanlewissoftwire: (the Abdullahview Point link redirects back to the  Scheme tab)

kosiakkatrina (approved):

PR #2033, opened 2023-11-15
CLDC-2865: Add an AWS production only pipeline

Chirag-Bhatti:
- A pipeline we need to use in order to bump up the Sidekiq task CPU / Memory temporarily as part of the AWS migration
- This will only deploy to AWS Prod


natdeanlewissoftwire (approved):

SamSeed-Softwire:
@natdeanlewissoftwire closing as we realised we don't actually need this.

PR #2037, opened 2023-11-16
CLDC-2899 Make scheme search by code case insensitive

kosiakkatrina:


app/models/scheme.rb
```diff
@@ -6,7 +6,7 @@ class Scheme < ApplicationRecord
 
   has_paper_trail
 
-  scope :filter_by_id, ->(id) { where(id: (id.start_with?("S") ? id[1..] : id)) }
+  scope :filter_by_id, ->(id) { where(id: (id.start_with?("S", "s") ? id[1..] : id)) }
```
natdeanlewissoftwire, line 9: FWIW you can regex as a param with start_with? so [sS] would work here if you wanted

natdeanlewissoftwire (approved):

PR #2039, opened 2023-11-16
Cldc 2636.2 Set uprn_known to no if address answered without uprn

natdeanlewissoftwire:


app/models/derived_variables/lettings_log_variables.rb
```diff
@@ -310,6 +312,10 @@ def reset_address_fields!
     self.county = nil
   end
 
+  def address_answered_without_uprn?
+    [address_line1, town_or_city].all?(&:present?) && uprn.nil?
+  end
+
```
kosiakkatrina: What would happen to 2022 logs? (I guess won't be relevant for long)
Also, what about the sales logs?

natdeanlewissoftwire: Yep sales incoming, and good point re 2022 - I've added a condition to only apply this for 2023 logs

kosiakkatrina (approved):

natdeanlewissoftwire:
replacing with https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/2054 as branch name is incorrect

PR #2042, opened 2023-11-17
CLDC-2830 Update devise error messages

kosiakkatrina:


natdeanlewissoftwire (changes requested):
All this looks good, but I think there's a second AC in the ticket this doesn't address - "make the page robust so triggering validation errors doesn’t make the whole page expire".

kosiakkatrina:
> All this looks good, but I think there's a second AC in the ticket this doesn't address - "make the page robust so triggering validation errors doesn’t make the whole page expire".

@natdeanlewissoftwire I don't think the page actually expired. The error for that AC is for when the token in the url is wrong, so I updated that error message to make it a bit clearer. If the token is correct and it's the password that has an error on it, the page should be robust

natdeanlewissoftwire:
Ah nice, sounds good in that case @kosiakkatrina I'll approve

natdeanlewissoftwire (approved):

PR #2044, opened 2023-11-20
CLDC-2986 Close 2022/23 collection period

kosiakkatrina:


app/models/form.rb
```diff
@@ -6,7 +6,11 @@ class Form
   def initialize(form_path, start_year = "", sections_in_form = [], type = "lettings")
     if sales_or_start_year_after_2022?(type, start_year)
       @start_date = Time.zone.local(start_year, 4, 1)
-      @new_logs_end_date = Time.zone.local(start_year + 1, 12, 31)  # this is to be manually updated each year when we want to stop users from creating new logs
+      @new_logs_end_date = if start_year && start_year.to_i == 2022
```
natdeanlewissoftwire, line 9: I think if we're splitting this up anyway for sales vs lettings (which is what this essentially is) then it doesn't make sense to have a `sales_or_start_year_after_2022` condition.

Can we make it a bit clearer by removing the sales condition and just do if `start_year_after_2022?`

Or if we do want this separated for sales vs lettings, then make it more obvious with
```
if lettings
  if after 2022
  else
  end
else
  if after 2022
  else
  end
end
```

(but I don't think that's necessary)

natdeanlewissoftwire (approved):
One suggestion that might simplify this for future readers of this code, but approving as is regardless

app/models/form.rb
kosiakkatrina: I think the `sales_or_start_year_after_2022` is to determine if the form definition is coming from code of json and not just the year. And if we split everything up then we'd need to define all of the attributes 4 times (?) or am I missing something?

natdeanlewissoftwire: Ahh good point, yeah I guess it's ok as is then 😢 Maybe we could name sales_or_start_year_after_2022 something that makes it clear it's actually checking if we're using json or code rather than just an arbitrary grouping for simplifying the conditions as I had assumed

PR #2045, opened 2023-11-20
CLDC-2556 Update infrastructure documentation

kosiakkatrina:


natdeanlewissoftwire (approved):
Looks good, though maybe we should think about having a single source of truth for this sort of documentation rather than updating in 2 places concurrently

natdeanlewissoftwire:
(I'm anticipating future changes to this and the runbook so would save time if we can agree one now :))

kosiakkatrina:
> Looks good, though maybe we should think about having a single source of truth for this sort of documentation rather than updating in 2 places concurrently

@natdeanlewissoftwire Yeah, I know what you mean. My opinion is that if we can make if public, we probably should, given we're open source. I think we've got more background in the runbook that maybe isn't too relevant here, but we can always link to this service from confluence for any technical documentation. Let me know if you've got other thoughts tho

PR #2048, opened 2023-11-22
CLDC-2938 Add sales managing org

kosiakkatrina:
We want to start collecting managing organisation IDs for Sales logs as sometimes they are actually reported by organisations different from the owning organisation.
This PR adds the column and the task that copies owning organisation ID over to managing organisation ID.
This needs to be released and executed before we add the new managing organisation question to the set up section.

### DIFF
--- a/app/models/sales_log.rb
+++ b/app/models/sales_log.rb
@@ -33,6 +33,8 @@ class SalesLog < Log
   before_validation :set_derived_fields!
   before_validation :process_uprn_change!, if: :should_process_uprn_change?
 
+  belongs_to :managing_organisation, class_name: "Organisation", optional: true
+
   scope :filter_by_year, ->(year) { where(saledate: Time.zone.local(year.to_i, 4, 1)...Time.zone.local(year.to_i + 1, 4, 1)) }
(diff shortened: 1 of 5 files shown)


app/models/sales_log.rb
```diff
@@ -33,6 +33,8 @@ class SalesLog < Log
   before_validation :set_derived_fields!
   before_validation :process_uprn_change!, if: :should_process_uprn_change?
 
+  belongs_to :managing_organisation, class_name: "Organisation", optional: true
+
   scope :filter_by_year, ->(year) { where(saledate: Time.zone.local(year.to_i, 4, 1)...Time.zone.local(year.to_i + 1, 4, 1)) }
   scope :filter_by_purchaser_code, ->(purchid) { where("purchid ILIKE ?", "%#{purchid}%") }
   scope :search_by, lambda { |param|
```
natdeanlewissoftwire, line 40: Just flagging that we'll need to update the filters below as part of this, assume that's in the ticket but if not please add it!

natdeanlewissoftwire (approved):

app/models/sales_log.rb
kosiakkatrina: Yup, it's in the ticket! Added it in the follow up PR! Thanks!

PR #2049, opened 2023-11-22
CLDC-2938 Add sales managing organisation question

kosiakkatrina:
We want to start collecting managing (reporting) organisation for sales logs as well.
We would probably want to add the column and update the managing organisation for records to be the same as the owning organisation first, to avoid any data loss due to routing, so this should be merged and ran first separately: https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/2048

In this PR:
- Add managing organisation question which we only display to support users, otherwise it gets inferred as current users organisation. 
- Route to owning organisation question if the org has more than 1 stock owner (including itself).
- Created by has to be someone from the owning organisation.
- Add reported by and owned by filters where appropriate
- Add reported by and owned by tags to the logs list view


app/controllers/lettings_logs_controller.rb
```diff
@@ -123,12 +123,6 @@ def filter_manager
     FilterManager.new(current_user:, session:, params:, filter_type: "lettings_logs")
   end
 
-  def org_params
```
natdeanlewissoftwire, line 126: nice!

spec/factories/sales_log.rb
```diff
@@ -2,6 +2,7 @@
   factory :sales_log do
     created_by { FactoryBot.create(:user) }
     owning_organisation { created_by.organisation }
+    managing_organisation { owning_organisation }
```
natdeanlewissoftwire, line 5: would have assumed this doesn't work, but nice :)

app/models/form/sales/pages/organisation.rb
```diff
@@ -10,24 +10,40 @@ def questions
...
-    if FeatureToggle.merge_organisations_enabled?
+  def routed_to?(log, current_user)
+    return false unless current_user
+    return true if current_user.support?
+    return false unless FeatureToggle.sales_managing_organisation_enabled?
 
-      return false unless current_user
-      return true if current_user.support?
+    return true if stock_owners_with_own_stock_count(current_user) > 1
+    return true if current_user.organisation.holds_own_stock? && stock_owners_with_own_stock_count(current_user) >= 1
```
natdeanlewissoftwire: can we combine this and the previous line with a new method like `manages_logs_for_multiple_stock_owners?(current_user)` for clarity?

app/models/form/sales/pages/organisation.rb
```diff
@@ -10,24 +10,40 @@ def questions
...
+
+      log.update!(owning_organisation: stock_owners.first)
     end
+
+    false
+  end
+
+private
+
+  def stock_owners_with_own_stock_count(user)
```
natdeanlewissoftwire: Can we move this out of the page to a shared location for sales and lettings? Think I might have been the one to put the equivalent method on the lettings page, sorry if that's true!

app/models/form/sales/pages/organisation.rb
natdeanlewissoftwire, line 1: think we should rename this file to be `owning_organisation`

natdeanlewissoftwire: and the id etc etc.

natdeanlewissoftwire (commented):
Looks good, mostly refactoring thoughts that I've commented - one larger thought is:

We've got a lot of validations that reference managing_organisation in the lettings log setup/financial validations files. Do we want to add similar validations for any/all of these to sales?

Also, there are some validations in bulk upload that check e.g. user is in either the owning or managing org, I think we should be adding these to the sales row parser too now.

app/models/form/sales/pages/organisation.rb
kosiakkatrina: Sure!

app/models/form/sales/pages/organisation.rb
kosiakkatrina: Ended up removing this altogether

app/models/form/sales/pages/organisation.rb
kosiakkatrina: Done!

kosiakkatrina:
> Looks good, mostly refactoring thoughts that I've commented - one larger thought is:
> 
> We've got a lot of validations that reference managing_organisation in the lettings log setup/financial validations files. Do we want to add similar validations for any/all of these to sales?
> 
> Also, there are some validations in bulk upload that check e.g. user is in either the owning or managing org, I think we should be adding these to the sales row parser too now.

I think financial validations are around rent periods so probably not relevant to sales. The setup validations were around the merged orgs stuff which should be there and I've now updated the bulk upload logic! Thanks!

natdeanlewissoftwire (approved):

PR #2051, opened 2023-11-22
Upgrade node version from ^18.0.0 to ^20.0.0

kosiakkatrina:


natdeanlewissoftwire (approved):

RachaelBooth (commented):
This all looks fine (and happy to update to node 20), but I think there'll need to be an update to the dockerfile as well so that the app on AWS uses it too.

RachaelBooth (approved):

PR #2053, opened 2023-11-24
CLDC-2071 Add intermediate rent error to BU

kosiakkatrina:
For intermediate rent logs, we also need answers to field 11 and conditionally field 12, but if field 11 is not given we currently only add an error on field 5 with `You must answer rent type` which doesn't really make sense. This PR adds an error to field 11 in this situation.

### DIFF
--- a/app/services/bulk_upload/lettings/year2023/row_parser.rb
+++ b/app/services/bulk_upload/lettings/year2023/row_parser.rb
@@ -329,6 +329,14 @@ class BulkUpload::Lettings::Year2023::RowParser
             },
             on: :after_log
 
+  validates :field_11,
+            presence: {
+              if: proc { renttype == :intermediate },
+              message: I18n.t("validations.not_answered", question: "intermediate rent type"),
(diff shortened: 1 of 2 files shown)


natdeanlewissoftwire (approved):
looks good, should we not add this to field 12 too when it's answered?

natdeanlewissoftwire:
Also should we remove/change the error on field 5 in this case if it's a confusing error message?

kosiakkatrina:
> looks good, should we not add this to field 12 too when it's answered?

@natdeanlewissoftwire that's already happening, added a test!

kosiakkatrina:
> Also should we remove/change the error on field 5 in this case if it's a confusing error message?

Yeah, it's an odd one, cause the error message is on the rent_type question, which doesn't actually get set cause we don't have enough info and all the rent type errors map back field 5. I'm gonna leave it in given it's one of the setup fields and hope it won't be too confusing in conjunction with field 11 error

PR #2056, opened 2023-11-27
CLDC-2969 Move dpo user when merging

kosiakkatrina:
When merging organisations, we move users to the absorbing organisation. We still want to keep DPO in the merging organisation so that the DSA stays signed.
This PR:
- replaces dpo users with fake users in merging organisation if they have signed the data sharing agreement
- does not move dpo users who have signed data sharing agreement if they have a fake email address

This PR also adds the validation for sales and lettings logs back, but we set a flag to skip dpo validation so that we can move the logs over.

### DIFF
--- a/app/services/merge/merge_organisations_service.rb
+++ b/app/services/merge/merge_organisations_service.rb
@@ -88,18 +88,21 @@ private
         lettings_log.location = location_to_set if location_to_set.present?
       end
       lettings_log.owning_organisation = @absorbing_organisation
-      lettings_log.save!(validate: false)
+      lettings_log.skip_dpo_validation = true
+      lettings_log.save!
     end
     merging_organisation.managed_lettings_logs.after_date(@merge_date.to_time).each do |lettings_log|
       lettings_log.managing_organisation = @absorbing_organisation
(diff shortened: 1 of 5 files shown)


app/services/merge/merge_organisations_service.rb
```diff
@@ -87,18 +88,21 @@ def merge_lettings_logs(merging_organisation)
         lettings_log.location = location_to_set if location_to_set.present?
       end
       lettings_log.owning_organisation = @absorbing_organisation
-      lettings_log.save!(validate: false)
+      lettings_log.skip_dpo_validation = true
```
RachaelBooth, line 91: I feel like I'm missing something obvious, but does this (/should this) get cleared at any point? (i.e. so that any future saves would not skip this validation)

kosiakkatrina: This is just in memory tho, so it should disappear as soon as this lettings_log no longer exists. Or am I missing something?
image

RachaelBooth: Ah, of course

RachaelBooth (approved):
Makes sense to me!

PR #2057, opened 2023-11-27
CLDC-2997 Validate organisation before schemes

kosiakkatrina:
When we're adding errors to startdate on lettings, if both organisation (organisation is not active on startdate) and scheme/location (scheme/location is not active on startdate) get error messages, we previously displayed scheme error message first. Now we validate the organisation first.

This PR also removes duplicate scheme/location validations from date_validations file


app/models/validations/setup_validations.rb
```diff
@@ -18,34 +18,6 @@ def validate_startdate_setup(record)
...
-      record.errors.add :scheme_id, :no_completed_locations, message: I18n.t("validations.scheme.no_completed_locations")
-    end
-  end
-
-  def validate_scheme(record)
-    location_during_startdate_validation(record)
-    scheme_during_startdate_validation(record)
-  end
-
   def validate_organisation(record)
```
kosiakkatrina, line 21: I've only moved this method above validate_irproduct_other in this file. It makes sense to get orgs validated as one of the first things

app/models/validations/date_validations.rb
```diff
@@ -53,9 +53,6 @@ def validate_startdate(record)
     if record["mrcdate"].present? && record["startdate"].to_date - record["mrcdate"].to_date > 3650
       record.errors.add :startdate, I18n.t("validations.setup.startdate.ten_years_after_mrc_date")
     end
-
-    location_during_startdate_validation(record)
-    scheme_during_startdate_validation(record)
```
natdeanlewissoftwire, line 58: on this note, do we need `location_during_startdate_validation` to be duplicated in `validate_scheme` in setup_validations.rb?

natdeanlewissoftwire (approved):

app/models/validations/date_validations.rb
kosiakkatrina: Nope, removed it too!

PR #2059, opened 2023-11-27
CLDC-2998 Don't display available from if it doesn't exist

kosiakkatrina:
In organisation question dropdown, if organisations have merged we display the availability for each org. If organisation has existed before the merge (does not have an available_from value) we don't want to display availability for it. This behaviour was missed out from support users but should now be there.


natdeanlewissoftwire (approved):
lgtm!

PR #2060, opened 2023-11-27
CLDC-2999 Deactivate locations on merged orgs

kosiakkatrina:
Deactivate the locations on merged organisations after they get copied over to the absorbing organisation
This PR also includes copying deactivated schemes and locations over to the absorbing organisation (CLDC-3003). We're also now copying the state by duplicating all of the scheme/location deactivation periods

### DIFF
--- a/app/services/merge/merge_organisations_service.rb
+++ b/app/services/merge/merge_organisations_service.rb
@@ -66,17 +66,19 @@ private
   def merge_schemes_and_locations(merging_organisation)
     @merged_schemes[merging_organisation.name] = []
     merging_organisation.owned_schemes.each do |scheme|
-      next if scheme.deactivated?
-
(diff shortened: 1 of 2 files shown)


natdeanlewissoftwire (commented):
Hm, I've not reviewed this fully but I think this affects the behaviour in https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/2061 - if we're now copying all the period history across, we (I) might ned to reconsider how we're displaying merge date in the availability text

kosiakkatrina:
> Hm, I've not reviewed this fully but I think this affects the behaviour in #2061 - if we're now copying all the period history across, we (I) might ned to reconsider how we're displaying merge date in the availability text

@natdeanlewissoftwire hmm, good point, is it any different from when a user just deactivates a scheme on the absorbing organisation tho? I feel like availability in terms of when it became accessible to the new org and deactivations might be slightly different things

PR #2061, opened 2023-11-27
CLDC-2996 Add correct availability details to merged schemes and locations

natdeanlewissoftwire:
Adds a merge date to schemes and locations that are merged, so that we can display the correct Availability text.
ticket: https://dluhcdigital.atlassian.net/browse/CLDC-2996


app/models/location.rb
```diff
@@ -122,9 +122,7 @@ def postcode=(postcode)
   end
 
   def available_from
-    return startdate if startdate.present?
-
-    FormHandler.instance.earliest_open_collection_start_date(now: created_at)
+    [merge_date, startdate].compact.max || FormHandler.instance.earliest_open_collection_start_date(now: created_at)
```
natdeanlewissoftwire: selects the latest merge or start date, if either exists, else just returns the earliest open collection start date for the created_at value

db/migrate/20231127125729_add_merge_date_to_locations.rb
```diff
@@ -0,0 +1,5 @@
+class AddMergeDateToLocations < ActiveRecord::Migration[7.0]
+  def change
+    add_column :locations, :merge_date, :datetime
+  end
+end
```
kosiakkatrina: Would there be any use in also saving the ID of the previous scheme/location? So that we know which one it is a duplicate of?

app/models/scheme.rb
```diff
@@ -243,7 +243,7 @@ def intended_length_of_stay_options_with_hints
   end
 
   def validate_confirmed
-    required_attributes = attribute_names - %w[id created_at updated_at old_id old_visible_id confirmed end_date sensitive secondary_client_group total_units deactivation_date deactivation_date_type]
+    required_attributes = attribute_names - %w[id created_at updated_at old_id old_visible_id confirmed end_date sensitive secondary_client_group total_units deactivation_date deactivation_date_type merge_date]
```
kosiakkatrina: ~Should the merge_date be optional?~ Sorry, this is what I'd expect, I misread!

kosiakkatrina (approved):

db/migrate/20231127125729_add_merge_date_to_locations.rb
natdeanlewissoftwire: Hm, we've got that information already through `old_visible_id` I think?

natdeanlewissoftwire: Though having a direct id would make querying this simpler, is  it something you think we'll have to do often? It seems like bloat to me, I can't imagine needing to get that info much

natdeanlewissoftwire: In most cases it should be obvious anyway through the location name

app/models/location.rb
```diff
@@ -122,9 +122,7 @@ def postcode=(postcode)
   end
 
   def available_from
-    return startdate if startdate.present?
-
-    FormHandler.instance.earliest_open_collection_start_date(now: created_at)
+    startdate || FormHandler.instance.earliest_open_collection_start_date(now: created_at)
```
kosiakkatrina, line 125: nice

app/services/merge/merge_organisations_service.rb
```diff
@@ -69,9 +69,9 @@ def merge_schemes_and_locations(merging_organisation)
     merging_organisation.owned_schemes.each do |scheme|
       next if scheme.deactivated?
 
-      new_scheme = Scheme.create!(scheme.attributes.except("id", "owning_organisation_id", "old_id", "old_visible_id").merge(owning_organisation: @absorbing_organisation))
+      new_scheme = Scheme.create!(scheme.attributes.except("id", "owning_organisation_id", "old_id", "old_visible_id").merge(owning_organisation: @absorbing_organisation, startdate: @merge_date))
       scheme.locations.each do |location|
-        new_scheme.locations << Location.new(location.attributes.except("id", "scheme_id", "old_id", "old_visible_id")) unless location.deactivated?
+        new_scheme.locations << Location.new(location.attributes.except("id", "scheme_id", "old_id", "old_visible_id").merge(startdate: [location&.startdate, @merge_date].compact.max)) unless location.deactivated?
```
kosiakkatrina, line 74: nice

app/models/scheme.rb
```diff
@@ -282,6 +282,7 @@ def status_at(date)
     return :deactivated if open_deactivation&.deactivation_date.present? && date >= open_deactivation.deactivation_date
     return :deactivating_soon if open_deactivation&.deactivation_date.present? && date < open_deactivation.deactivation_date
     return :reactivating_soon if last_deactivation_before(date)&.reactivation_date.present? && date < last_deactivation_before(date).reactivation_date
+    return :activating_soon if startdate.present? && date < startdate
```
kosiakkatrina, line 290: Can we have a test for this? Also should there be a validation that prevents adding deactivation periods before startdate? I assume we already do that for locations

natdeanlewissoftwire: ah good point - I've added a test. The validation's already there - it works through available_from on both schemes and locations, which has been updated in this PR

app/models/location.rb
```diff
@@ -21,9 +21,9 @@ class Location < ApplicationRecord
   scope :search_by_postcode, ->(postcode) { where("REPLACE(postcode, ' ', '') ILIKE ?", "%#{postcode.delete(' ')}%") }
   scope :search_by_name, ->(name) { where("name ILIKE ?", "%#{name}%") }
   scope :search_by, ->(param) { search_by_name(param).or(search_by_postcode(param)) }
-  scope :started, -> { where("startdate <= ?", Time.zone.today).or(where(startdate: nil)) }
+  scope :started, -> { where("locations.startdate <= ?", Time.zone.today).or(where(startdate: nil)) }
   scope :active, -> { where(confirmed: true).and(started) }
-  scope :started_in_2_weeks, -> { where("startdate <= ?", Time.zone.today + 2.weeks).or(where(startdate: nil)) }
+  scope :started_in_2_weeks, -> { where("locations.startdate <= ?", Time.zone.today + 2.weeks).or(where(startdate: nil)) }
```
kosiakkatrina, line 26: Do we need to do this?

kosiakkatrina (approved):

app/models/location.rb
natdeanlewissoftwire: Yep, when we join schemes and locations there's ambiguity now as they both have a startdate attribute, so this is needed to resolve that

PR #2064, opened 2023-11-29
CLDC-3004 Do not reset created by for merged orgs

kosiakkatrina:
When support user selects a log owner for merged organisation it would get reset, because that user no longer belongs to the merged organisation.
This PR does not reset create by user if they belong to the absorbing organisation of owning or managing org.

### DIFF
--- a/app/models/sales_log.rb
+++ b/app/models/sales_log.rb
@@ -318,6 +318,7 @@ class SalesLog < Log
     return unless updated_by&.support?
     return if owning_organisation.blank? || created_by.blank?
     return if created_by&.organisation == owning_organisation
+    return if created_by&.organisation == owning_organisation.absorbing_organisation
 
     update!(created_by: nil)
   end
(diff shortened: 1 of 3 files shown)


app/models/sales_log.rb
```diff
@@ -318,6 +318,7 @@ def reset_created_by!
     return unless updated_by&.support?
     return if owning_organisation.blank? || created_by.blank?
     return if created_by&.organisation == owning_organisation
+    return if created_by&.organisation == owning_organisation.absorbing_organisation
```
natdeanlewissoftwire, line 321: we'll need to add managing to this too once the sales managing org behaviour is in, right?

natdeanlewissoftwire (approved):

natdeanlewissoftwire (approved):

app/models/sales_log.rb
kosiakkatrina: Yup, likely!

PR #2065, opened 2023-11-29
CLDC-3005 Display labels and filters for merged orgs

kosiakkatrina:
We've previously only displayed owned by and managed by labels and filters for logs if they have managing agents, but we also want to do it if they have absorbed organisations.

### DIFF
--- a/app/models/organisation.rb
+++ b/app/models/organisation.rb
@@ -146,4 +146,8 @@ class Organisation < ApplicationRecord
 
     absorbed_organisations.merged_during_open_collection_period.group_by(&:merge_date)
   end
+
+  def has_recent_absorbed_organisations?
+    absorbed_organisations&.merged_during_open_collection_period.present?
+  end
 end
--- a/spec/requests/sales_logs_controller_spec.rb
+++ b/spec/requests/sales_logs_controller_spec.rb
@@ -559,7 +559,7 @@ RSpec.describe SalesLogsController, type: :request do
           sign_in user
         end
 
-        it "does not have organisation values" do
+        it "does not show organisation labels" do
           get "/sales-logs", headers: headers, params: {}
           expect(page).not_to have_content("Owned by")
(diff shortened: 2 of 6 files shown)


spec/requests/sales_logs_controller_spec.rb
```diff
@@ -559,6 +559,26 @@
...
+        end
+
+        context "and organisation has absorbed organisations" do
+          let(:merged_organisation) { FactoryBot.create(:organisation) }
+
+          before do
+            merged_organisation.update!(absorbing_organisation: organisation, merge_date: Time.zone.yesterday)
+          end
+
+          it "has organisation values" do
```
natdeanlewissoftwire: can we name these tests consistently between sales and lettings if they're testing the same thing so they're easier to find in future?

app/models/organisation.rb
```diff
@@ -146,4 +146,8 @@ def recently_absorbed_organisations_grouped_by_merge_date
 
     absorbed_organisations.merged_during_open_collection_period.group_by(&:merge_date)
   end
+
+  def has_recent_absorbed_organisations?
+    absorbed_organisations&.merged_during_open_collection_period.present?
```
natdeanlewissoftwire, line 151: do we not want users to see these labels/filters for no longer editable logs?

natdeanlewissoftwire (approved):
couple of thoughts, lgtm otherwise :)

spec/requests/sales_logs_controller_spec.rb
kosiakkatrina: Sure!

app/models/organisation.rb
kosiakkatrina: not based on the ticket, I guess it would no longer make sense cause all of the relevant logs would belong to the same organisation and the labels would clutter the page (?)

PR #2066, opened 2023-11-30
CLDC-3011 Display scheme owner question for absorbing orgs

kosiakkatrina:
When creating new schemes if user's organisation has stock owners we ask "Which organisation owns the housing stock for this scheme?". We also want to ask this question if user's organisation has recently absorbed organisations

### DIFF
--- a/spec/requests/schemes_controller_spec.rb
+++ b/spec/requests/schemes_controller_spec.rb
@@ -2229,12 +2229,14 @@ RSpec.describe SchemesController, type: :request do
           get "/schemes/#{scheme.id}/edit-name"
         end
 
-        it "does not include the owning organisation question" do
-          expect(response).to have_http_status(:ok)
-          expect(page).not_to have_content("Which organisation owns the housing stock for this scheme?")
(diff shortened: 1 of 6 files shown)


spec/requests/schemes_controller_spec.rb
```diff
@@ -2224,6 +2224,31 @@
         end
       end
 
+      context "when there are no stock owners" do
+        before do
+          get "/schemes/#{scheme.id}/edit-name"
+        end
+
+        it "does not include the owning organisation question" do
```
RachaelBooth: Does this want a context of "and there are no recently absorbed organisations"? Purely for reading it purposes.

RachaelBooth: Actually, does there want to be a test where there are older absorbed organisations?

kosiakkatrina: Sure, how does it look now?

RachaelBooth: LGTM! Thanks.

RachaelBooth (approved):

PR #2067, opened 2023-11-30
CLDC-3028 Allow bulk uploading for merged organisations

kosiakkatrina:


app/models/organisation.rb
```diff
@@ -25,6 +25,11 @@ def affiliated_stock_owners
       ids << id
     end
 
+    absorbed_organisations.each do |organisation|
+      ids.concat(organisation.stock_owners.pluck(:id))
```
natdeanlewissoftwire, line 29: aren't these stock owner relationships applied to the absorbing organisation during a merge?

app/models/organisation.rb
```diff
@@ -25,6 +25,11 @@ def affiliated_stock_owners
...
+      ids << organisation.id if organisation.holds_own_stock?
```
natdeanlewissoftwire, line 30: isn't this condition always true?

app/models/organisation.rb
kosiakkatrina: Hmm, I guess if they're removed on absorbed orgs but they might still be relevant for merged 🤔

app/models/organisation.rb
kosiakkatrina: What makes you think that? I think merged organisation might not hold stock (?) Am I missing something?

app/models/organisation.rb
natdeanlewissoftwire: I'm not sure what you mean - I'm saying that this line is already covered by line 33 if that helps

app/models/organisation.rb
natdeanlewissoftwire: Ah no you're right, I was still thinking we were looking at stock owners of this org, ignore me

natdeanlewissoftwire (approved):

PR #2069, opened 2023-12-01
CLDC-2999 Deactivate locations on merged orgs

kosiakkatrina:
Copy of https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/2060

### DIFF
--- a/app/services/merge/merge_organisations_service.rb
+++ b/app/services/merge/merge_organisations_service.rb
@@ -67,17 +67,19 @@ private
   def merge_schemes_and_locations(merging_organisation)
     @merged_schemes[merging_organisation.name] = []
     merging_organisation.owned_schemes.each do |scheme|
-      next if scheme.deactivated?
-
-      new_scheme = Scheme.create!(scheme.attributes.except("id", "owning_organisation_id", "old_id", "old_visible_id").merge(owning_organisation: @absorbing_organisation, startdate: @merge_date))
+      new_scheme = Scheme.create!(scheme.attributes.except("id", "owning_organisation_id", "old_id", "old_visible_id").merge(owning_organisation: @absorbing_organisation, startdate: [scheme&.startdate, @merge_date].compact.max))
+      scheme.scheme_deactivation_periods.each do |deactivation_period|
+        split_scheme_deactivation_period_between_organisations(deactivation_period, new_scheme)
+      end
       scheme.locations.each do |location|
-        next if location.deactivated?
-
-        new_scheme.locations << Location.new(location.attributes.except("id", "scheme_id", "old_id", "old_visible_id").merge(startdate: [location&.startdate, @merge_date].compact.max))
-        LocationDeactivationPeriod.create!(location:, deactivation_date: @merge_date) unless (location.startdate.present? && location.startdate >= @merge_date) || (scheme.startdate.present? && scheme.startdate >= @merge_date)
+        new_location = Location.create!(location.attributes.except("id", "scheme_id", "old_id", "old_visible_id").merge(scheme: new_scheme, startdate: [location&.startdate, @merge_date].compact.max))
+        location.location_deactivation_periods.each do |deactivation_period|
+          split_location_deactivation_period_between_organisations(deactivation_period, new_location)
+        end
+        LocationDeactivationPeriod.create!(location:, deactivation_date: [location&.startdate, scheme&.startdate, @merge_date].compact.max) unless location.status_at(@merge_date) == :deactivated
       end
       @merged_schemes[merging_organisation.name] << { name: new_scheme.service_name, code: new_scheme.id }
-      SchemeDeactivationPeriod.create!(scheme:, deactivation_date: @merge_date)
+      SchemeDeactivationPeriod.create!(scheme:, deactivation_date: [scheme&.startdate, @merge_date].compact.max) unless scheme.status_at(@merge_date) == :deactivated
     end
   end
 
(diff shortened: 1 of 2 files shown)


natdeanlewissoftwire:
> > Hm, I've not reviewed this fully but I think this affects the behaviour in https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/2061 - if we're now copying all the period history across, we (I) might ned to reconsider how we're displaying merge date in the availability text
> 
> @natdeanlewissoftwire hmm, good point, is it any different from when a user just deactivates a scheme on the absorbing organisation tho? I feel like availability in terms of when it became accessible to the new org and deactivations might be slightly different things

Just continuing this from the previous PR - I've done some more digging and I think this is alright, could be some edge cases I've not thought of though. All the scheme/log/location merge/start/inactive date validations would benefit from some pretty thorough testing I think as it's a lot of moving parts

app/services/merge/merge_organisations_service.rb
```diff
@@ -67,14 +67,19 @@ def merge_users(merging_organisation)
...
+      scheme.scheme_deactivation_periods.each do |deactivation_period|
+        SchemeDeactivationPeriod.create!(deactivation_period.attributes.except("id", "scheme_id").merge(scheme: new_scheme))
+      end
       scheme.locations.each do |location|
-        new_scheme.locations << Location.new(location.attributes.except("id", "scheme_id", "old_id", "old_visible_id")) unless location.deactivated?
+        new_location = Location.create!(location.attributes.except("id", "scheme_id", "old_id", "old_visible_id").merge(scheme: new_scheme))
+        location.location_deactivation_periods.each do |deactivation_period|
+          LocationDeactivationPeriod.create!(deactivation_period.attributes.except("id", "location_id").merge(location: new_location))
+        end
+        LocationDeactivationPeriod.create!(location:, deactivation_date: @merge_date) unless location.deactivated?
```
natdeanlewissoftwire:
```suggestion
        LocationDeactivationPeriod.create!(location:, deactivation_date: @merge_date) unless location.status_at(@merge_date) == :deactivated
```

app/services/merge/merge_organisations_service.rb
```diff
@@ -67,14 +67,19 @@ def merge_users(merging_organisation)
...
       end
       @merged_schemes[merging_organisation.name] << { name: new_scheme.service_name, code: new_scheme.id }
-      SchemeDeactivationPeriod.create!(scheme:, deactivation_date: @merge_date)
+      SchemeDeactivationPeriod.create!(scheme:, deactivation_date: @merge_date) unless scheme.deactivated?
```
natdeanlewissoftwire:
```suggestion
      SchemeDeactivationPeriod.create!(scheme:, deactivation_date: @merge_date) unless scheme.status_at(@merge_date) == :deactivated
```

app/services/merge/merge_organisations_service.rb
natdeanlewissoftwire: I think we might also want to delete any scheme/location deactivations on the original scheme or location that are after the merge date to stop it reactivating in the future

kosiakkatrina: Yeah, you've got a point! So:
1. Move over all of the scheme deactivations after the merge_date to the absorbing org
2. Set available_from on scheme to merge_date (If merge_date is during a deactivated period, make the start_date of that deactivation == merge_date?)
3. Remove all the deactivations after merge_date from old orgs. After that:
     - If scheme is deactivated during the merge date and has no reactivation date, leave it as is
     - If scheme is deactivated during the merge date but has a reactivation date, remove the reactivation date
     - If scheme is active during the merge_date, create a deactivation starting on the merge_date
     
And same with locations? Does this make sense?

natdeanlewissoftwire: Yep, exactly! Sounds good to me :)

kosiakkatrina: Will add after https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/2061, is there much left on that? Need any eyes on it?

natdeanlewissoftwire: it's in PO now :)

kosiakkatrina: Okay, it kinda dramatically changed now. Would love another review of this!

app/services/merge/merge_organisations_service.rb
```diff
@@ -67,14 +67,19 @@ def merge_users(merging_organisation)
   def merge_schemes_and_locations(merging_organisation)
     @merged_schemes[merging_organisation.name] = []
     merging_organisation.owned_schemes.each do |scheme|
-      next if scheme.deactivated?
-
-      new_scheme = Scheme.create!(scheme.attributes.except("id", "owning_organisation_id", "old_id", "old_visible_id").merge(owning_organisation: @absorbing_organisation, startdate: @merge_date))
+      new_scheme = Scheme.create!(scheme.attributes.except("id", "owning_organisation_id", "old_id", "old_visible_id").merge(owning_organisation: @absorbing_organisation, startdate: [scheme&.startdate, @merge_date].compact.max))
```
natdeanlewissoftwire, line 70: so a scheme could only possibly have startdate present and after the merge date if it had already been merged this collection year (right?), but good to be safe here just in case, nice one!

app/services/merge/merge_organisations_service.rb
```diff
@@ -157,4 +162,36 @@ def users_to_merge(merging_organisation)
 
     merging_organisation.users.where.not(id: new_dpo.id)
   end
+
+  def deactivation_happenned_before_merge?(deactivation_period)
+    deactivation_period.deactivation_date <= @merge_date && deactivation_period.reactivation_date.present? && deactivation_period.reactivation_date <= @merge_date
```
natdeanlewissoftwire, line 180: can we guarantee all deactivation periods don't have a nil deactivation date? I _think_ it's always present but thought I'd flag that we assume that here

app/services/merge/merge_organisations_service.rb
```diff
@@ -157,4 +162,36 @@ def users_to_merge(merging_organisation)
...
+      SchemeDeactivationPeriod.create!(deactivation_period.attributes.except("id", "scheme_id", "deactivation_date").merge(scheme: new_scheme, deactivation_date: @merge_date))
+      deactivation_period.update!(reactivation_date: nil) if deactivation_period.reactivation_date.present?
+    else
+      SchemeDeactivationPeriod.create!(deactivation_period.attributes.except("id", "scheme_id").merge(scheme: new_scheme))
+      deactivation_period.destroy!
+    end
+  end
+
+  def split_location_deactivation_period_between_organisations(deactivation_period, new_location)
+    return if deactivation_happenned_before_merge?(deactivation_period)
```
natdeanlewissoftwire, line 203: these two split_ methods are super readable, lovely stuff 🥳

spec/services/merge/merge_organisations_service_spec.rb
```diff
@@ -133,68 +133,351 @@
...
+              expect(absorbed_scheme.scheme_deactivation_periods.count).to eq(1)
+              expect(absorbed_scheme.scheme_deactivation_periods.first.deactivation_date).to eq(Time.zone.today)
+              expect(absorbed_scheme.scheme_deactivation_periods.first.reactivation_date).to eq(Time.zone.today + 1.month)
+
+              expect(absorbed_location.startdate).to eq(Time.zone.today)
+              expect(absorbed_location.location_deactivation_periods.count).to eq(1)
+              expect(absorbed_location.location_deactivation_periods.first.deactivation_date).to eq(Time.zone.today)
+              expect(absorbed_location.location_deactivation_periods.first.reactivation_date).to eq(Time.zone.today + 1.month)
+
+              # deactivates schemes and locations on the merged organisation
```
natdeanlewissoftwire: this comment could be added to the `it "" do` title instead but honestly I'm not sure if that would improve this so up to you

natdeanlewissoftwire: and the later ones definitely not as they're so many - maybe this is a smell these tests should be structured as multiple `it "" do` blocks with a before action calling the merge service?

natdeanlewissoftwire (approved):
nice one! couple of thoughts but happy to approve regardless of what you decide on those :)

app/services/merge/merge_organisations_service.rb
kosiakkatrina: Hmm, good point. I think they shouldn't, but I'll check our data to see if there are any occurrences of that!

spec/services/merge/merge_organisations_service_spec.rb
kosiakkatrina: Nat, you're absolutely right! Refactored them now

app/services/merge/merge_organisations_service.rb
kosiakkatrina: hmm, yeah, good point. Would it be any collection year tho? Also, if we ever want to allow setting that date in any other way, like we do with locations for example

PR #2070, opened 2023-12-01
Do not validate records with closed collection periods

kosiakkatrina:
If merge happens in the previous collection period, the logs will fail saving as they will trigger open collection window validation. This PR prevents that.

### DIFF
--- a/app/services/merge/merge_organisations_service.rb
+++ b/app/services/merge/merge_organisations_service.rb
@@ -88,21 +88,34 @@ private
         lettings_log.location = location_to_set if location_to_set.present?
       end
       lettings_log.owning_organisation = @absorbing_organisation
-      lettings_log.skip_dpo_validation = true
-      lettings_log.save!
+      lettings_log.managing_organisation = @absorbing_organisation if lettings_log.managing_organisation == merging_organisation
+      if lettings_log.collection_period_open?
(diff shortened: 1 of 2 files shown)


app/services/merge/merge_organisations_service.rb
```diff
@@ -88,21 +88,34 @@ def merge_lettings_logs(merging_organisation)
         lettings_log.location = location_to_set if location_to_set.present?
       end
       lettings_log.owning_organisation = @absorbing_organisation
-      lettings_log.skip_dpo_validation = true
-      lettings_log.save!
+      lettings_log.managing_organisation = @absorbing_organisation if lettings_log.managing_organisation == merging_organisation
```
natdeanlewissoftwire, line 91: do we want to add this to sales logs too? or is the sales log managing org behaviour not merged yet?

natdeanlewissoftwire (approved):
lgtm

app/services/merge/merge_organisations_service.rb
kosiakkatrina: Yeah, we do, I'll add it as part of that PR cause it's still in PO

kosiakkatrina: Oh wait, hold on, we're already doing that. All good 😂

PR #2072, opened 2023-12-04
CLDC-3026 Only display 1 set of save buttons

kosiakkatrina:
Interruption screen has it's own set of Confirm and continue buttons, so we shouldn't be displaying a duplicate set

### DIFF
--- a/app/views/form/page.html.erb
+++ b/app/views/form/page.html.erb
@@ -63,18 +63,20 @@
       <%= f.hidden_field :interruption_page_referrer_type, value: @interruption_page_referrer_type %>
 
       <div class="govuk-button-group">
-      <% if accessed_from_duplicate_logs?(request.query_parameters["referrer"]) %>
-        <%= f.govuk_submit "Save changes" %>
-        <%= govuk_link_to "Cancel", send("#{@log.class.name.underscore}_duplicate_logs_path", @log, original_log_id: request.query_parameters["original_log_id"]) %>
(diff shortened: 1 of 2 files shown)


app/views/form/page.html.erb
```diff
@@ -63,18 +63,20 @@
...
-      <% elsif returning_to_question_page?(@page, request.query_parameters["referrer"]) %>
-        <%= f.govuk_submit "Save changes" %>
-        <%= govuk_link_to "Cancel", send(@log.form.cancel_path(@page, @log), @log) %>
-      <% else %>
-        <%= f.govuk_submit "Save and continue" %>
-          <%= govuk_link_to(
-            (@page.skip_text || "Skip for now"),
-            (@page.skip_href(@log) || send(@log.form.next_page_redirect_path(@page, @log, current_user), @log)),
-          ) %>
+      <% if !@page.interruption_screen? %>
```
kosiakkatrina, line 66: only added this line

RachaelBooth (approved):
Happy with this, although looking at the ticket there might also be a separate part to do on the skip for now thing?

kosiakkatrina:
> Happy with this, although looking at the ticket there might also be a separate part to do on the skip for now thing?

Hmm, I think this incorporates both of them 🤔

RachaelBooth:
> > Happy with this, although looking at the ticket there might also be a separate part to do on the skip for now thing?
> 
> Hmm, I think this incorporates both of them 🤔

Huh, I'm not sure why this would do that - although I can't replicate what Olivia mentions anywhere so 🤷‍♀️

PR #2073, opened 2023-12-04
CLDC-2952 Add missing soft validation message

kosiakkatrina:


RachaelBooth (approved):
Looks alright to me, pending someone confirming the error message content is still right.

PR #2074, opened 2023-12-04
CLDC-2953 Clear conditional values with errors

kosiakkatrina:
on conditional questions we use js to clear the values if an option that hides a question is selected. We do that by getting the input field by id and clearing it's value, however if there is an error on this field from the previous input, the field id is different and the values don't get cleared. 
This PR clears both input value versions if there are any values in them.

### DIFF
--- a/app/frontend/controllers/conditional_question_controller.js
+++ b/app/frontend/controllers/conditional_question_controller.js
@@ -14,7 +14,10 @@ export default class extends Controller {
 
       Object.entries(conditionalFor).forEach(([targetQuestion, conditions]) => {
         if (!conditions.map(String).includes(String(selectedValue))) {
-          const textNumericInput = document.getElementById(`${type}-${targetQuestion.replaceAll('_', '-')}-field`)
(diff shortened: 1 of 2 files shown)


app/frontend/controllers/conditional_question_controller.js
```diff
@@ -15,24 +15,27 @@ export default class extends Controller {
...
           const textNumericInput = document.getElementById(`${type}-${targetQuestion.replaceAll('_', '-')}-field`)
-          if (textNumericInput == null) {
-            const dateInputs = [1, 2, 3].map((idx) => {
-              return document.getElementById(`${type.replaceAll('-', '_')}_${targetQuestion}_${idx}i`)
-            })
-            this.clearDateInputs(dateInputs)
-          } else {
-            this.clearTextNumericInput(textNumericInput)
-          }
+          const errorInput = document.getElementById(`${type}-${targetQuestion.replaceAll('_', '-')}-field-error`)
```
kosiakkatrina, line 18: this works for both numeric/text and date inputs as the date error weirdly only gets added to the first date input field

RachaelBooth (approved):

PR #2079, opened 2023-12-06
Ensure aws_deploy has different concurrency groups for different revi…

RachaelBooth:
…ew apps

### DIFF
--- a/.github/workflows/aws_deploy.yml
+++ b/.github/workflows/aws_deploy.yml
@@ -12,6 +12,10 @@ on:
       aws_task_prefix:
         required: true
         type: string
+      concurrency_tag:
+        required: false
+        type: string
+        default: ""
       environment:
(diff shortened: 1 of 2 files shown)


.github/workflows/aws_deploy.yml
```diff
@@ -20,7 +24,7 @@ on:
         type: string
 
 concurrency:
-  group: deploy-${{ inputs.environment }}
+  group: deploy-${{ inputs.environment }}${{ inputs.concurrency_tag }}
```
RachaelBooth, line 27: The review pipeline itself had a concurrency group that used the PR number, but this bit was just using the environment so would sometimes end up with deployments cancelled that shouldn't have been.

natdeanlewissoftwire (approved):
Lgtm!

PR #2083, opened 2023-12-08
CLDC-3014 Add schemes and locations csv download functionality

natdeanlewissoftwire:
Adds links to the schemes page to download schemes/locations/schemes and locations data as a csv emailed to the user.

image

image

image

image

image


app/components/search_result_caption_component.html.erb
```diff
@@ -1,10 +1,10 @@
 <span class="govuk-!-margin-right-4">
   <% if searched.present? && filters_count&.positive? %>
-    <strong><%= count %></strong> <%= item_label.pluralize(count) %> matching search and filters<br>
+    <strong><%= count %></strong> <%= item_label.pluralize(count) %> matching search and filters
```
natdeanlewissoftwire: these were messing with the formatting, also for log downloads so I'm pretty sure they should have been removed already

kosiakkatrina: Hmm, I mean if you search and or filter logs it looks better with these new lines and I think it was in the design, cause otherwise the download links are split over several lines

app/controllers/organisations_controller.rb
```diff
@@ -29,6 +29,18 @@ def schemes
     @filter_type = "schemes"
   end
 
+  def download_schemes_csv
+    organisation_schemes = Scheme.where(owning_organisation_id: @organisation.id)
```
kosiakkatrina: Do we also want to get parent schemes given we display them in the list? Similar to line 24?

app/controllers/schemes_controller.rb
```diff
@@ -3,11 +3,11 @@ class SchemesController < ApplicationController
   include Modules::SearchFilter
 
   before_action :authenticate_user!
-  before_action :find_resource, except: %i[index create new changes]
+  before_action :find_resource, except: %i[index create new changes email_csv download_csv csv_confirmation]
   before_action :redirect_if_scheme_confirmed, only: %i[primary_client_group confirm_secondary_client_group secondary_client_group support details]
-  before_action :authorize_user
-  before_action :session_filters, if: :current_user, only: %i[index]
-  before_action -> { filter_manager.serialize_filters_to_session }, if: :current_user, only: %i[index]
+  before_action :authorize_user, except: %i[email_csv download_csv csv_confirmation]
```
kosiakkatrina, line 8: How come we don't want to authorise for download_csv and csv_confirmation?

kosiakkatrina (approved):

app/components/search_result_caption_component.html.erb
natdeanlewissoftwire: Ah I'd missed that - I've addressed this for schemes by adding nowrap to the download links, will do the same for the log link I think that's a better solution, rather than having an indented download link on a new line when there's loads of room above it. Sounds like the designs for log and scheme downloads contradict each other so I'll update both to the latest one.

natdeanlewissoftwire: Have updated to the user log download link too for consistency (even though I couldn't actually make it wrap incorrectly there anyway)

app/controllers/organisations_controller.rb
natdeanlewissoftwire: Ah great spot, thanks!

app/controllers/schemes_controller.rb
natdeanlewissoftwire: There isn't really anything to authorize - all user types can download their visible schemes so if I added a policy it would just check the user is of any type... or should the authorization be doing something further? I think we don't have an authorization for the labels version of log downloads which are available to non-support users.

app/views/logs/_log_list.html.erb
```diff
@@ -3,9 +3,9 @@
   <div class="govuk-grid-column-three-quarters">
     <%= render(SearchResultCaptionComponent.new(searched:, count: pagy.count, item_label:, total_count:, item: "logs", filters_count: applied_filters_count(@filter_type))) %>
     <% if logs&.any? %>
-      <%= govuk_link_to "Download (CSV)", csv_download_url, type: "text/csv", class: "govuk-!-margin-right-4" %>
+      <%= govuk_link_to "Download (CSV)", csv_download_url, type: "text/csv", class: "govuk-!-margin-right-4", style: "white-space: nowrap" %>
       <% if @current_user.support? %>
-        <%= govuk_link_to "Download (CSV, codes only)", csv_codes_only_download_url, type: "text/csv", class: "govuk-!-margin-right-4" %>
+        <%= govuk_link_to "Download (CSV, codes only)", csv_codes_only_download_url, type: "text/csv", class: "govuk-!-margin-right-4", style: "white-space: nowrap" %>
```
kosiakkatrina, line 8: Ah, sorry, so check with design, but I think we deliberately wanted all the links to be on a new line if the filter is applied https://dluhcdigital.atlassian.net/browse/CLDC-2743

natdeanlewissoftwire: Sure, will do thanks :)

natdeanlewissoftwire: you're absolutely right - I've fixed the spacing issue so now we don't have that indentation I mentioned above that makes it look weird when it's on a separate line: 
image
image
image
image

natdeanlewissoftwire: these are all the new behaviours ^ (this is the indentation issue that used to be present:)
image

kosiakkatrina (approved):

PR #2084, opened 2023-12-11
CLDC-3019 Allow scheme editing for support users

kosiakkatrina:
Allow support users edit more schemes and locations details after the resources have been created.


app/views/schemes/check_answers.html.erb
```diff
@@ -9,7 +9,7 @@
       <h2 class="govuk-visually-hidden">Scheme</h2>
       <dl class="govuk-summary-list">
         <% @scheme.check_details_attributes.each do |attr| %>
-          <%= render partial: "scheme_summary_list_row", locals: { scheme: @scheme, attribute: attr, change_link: @scheme.confirmed? ? scheme_edit_name_path(@scheme) : scheme_details_path(@scheme, check_answers: true) } %>
+          <%= render partial: "scheme_summary_list_row", locals: { scheme: @scheme, attribute: attr, change_link: current_user.support? ? scheme_details_path(@scheme, check_answers: true) : scheme_edit_name_path(@scheme) } unless attr[:name] == "Status" %>
```
natdeanlewissoftwire: won't this redirect to the edit name path for an unconfirmed (incomplete) scheme viewed by a data coordinator? I think we should still reirect to the details path in that case

app/views/schemes/show.html.erb
```diff
@@ -14,24 +14,48 @@
...
-          <% row.value do %>
-            <%= details_html(attr) %>
-            <% if attr[:name] == "Status" && @scheme.confirmed? && @scheme.locations.confirmed.none? && LocationPolicy.new(current_user, @scheme.locations.new).create? %>
-              <span class="app-!-colour-muted">Complete this scheme by adding a location using the <%= govuk_link_to("‘locations’ tab", scheme_locations_path(@scheme)) %>.</span>
-            <% end %>
-          <% end %>
-          <% if SchemePolicy.new(current_user, @scheme).update? %>
-            <% row.action(text: "Change", href: scheme_edit_name_path(scheme_id: @scheme.id)) if attr[:edit] %>
-          <% end %>
+    <dl class="govuk-summary-list">
```
natdeanlewissoftwire, line 22: nice :D

app/views/schemes/show.html.erb
```diff
@@ -14,24 +14,48 @@
...
-      <% display_scheme_attributes(@scheme).each do |attr| %>
-        <%= summary_list.row do |row| %>
-          <% row.key { attr[:name] } %>
-          <% row.value do %>
-            <%= details_html(attr) %>
-            <% if attr[:name] == "Status" && @scheme.confirmed? && @scheme.locations.confirmed.none? && LocationPolicy.new(current_user, @scheme.locations.new).create? %>
-              <span class="app-!-colour-muted">Complete this scheme by adding a location using the <%= govuk_link_to("‘locations’ tab", scheme_locations_path(@scheme)) %>.</span>
-            <% end %>
-          <% end %>
-          <% if SchemePolicy.new(current_user, @scheme).update? %>
```
natdeanlewissoftwire, line 34: is the scheme update policy still needed? or should it be changed at all in the light of this PR?

app/models/scheme.rb
```diff
@@ -187,37 +187,38 @@ def id_to_display
   def check_details_attributes
```
natdeanlewissoftwire, line 187: do we always use any of these `check_..._attributes` methods together? can we not combine any of them any further?

natdeanlewissoftwire (commented):
looks good! one general thing - I can't see the inset text in the designs anywhere here, is that being included in a different PR?
image

app/views/schemes/check_answers.html.erb
kosiakkatrina: Good spot!

app/views/schemes/show.html.erb
kosiakkatrina: I think at this point the policy is more around whether the user is someone who should be able to update the scheme (whether they're in the right org/role) as opposed to the state of the scheme, so I think it's still worth keeping around

app/models/scheme.rb
kosiakkatrina: Yup! Refactored it

kosiakkatrina:
> looks good! one general thing - I can't see the inset text in the designs anywhere here, is that being included in a different PR? 

@natdeanlewissoftwire ah, nope I just missed it, thanks! It's there now!

natdeanlewissoftwire (approved):
looks good!

PR #2085, opened 2023-12-11
Helper script for executing commands on ECS app containers

RachaelBooth:


RachaelBooth:
Was doing a support ticket involving the rails console and I got annoyed so wrote the script... Will do confluence updates on merge.

I can only confirm that it works on my machine right now.

exec.sh
```diff
@@ -0,0 +1,10 @@
+if [ $# -ne 2 ];
+  then echo "Expected 2 arguments: exec.sh env command"
+fi
+
+env=$1
+command=$2
+cluster="core-$env-app"
+
+taskId=$(aws ecs list-tasks --cluster $cluster --service-name "core-$env-app" --query "taskArns[0]" | grep -o "/$cluster/\w*" | sed "s@/$cluster/@@g")
+aws ecs execute-command --cluster "core-$env-app" --task $taskId --interactive --command "$command"
```
RachaelBooth, line 10: This should be able to use the task arn directly but that errors, and the id works.

kosiakkatrina (approved):
Nice, I assume you checked on review apps?

PR #2086, opened 2023-12-11
CLDC-2960 Display missing stock owners banner

kosiakkatrina:
Add a banner if the organisation shouldn't be able to create logs
image

Do not display create logs buttons in organisation shouldn't be able to create logs

### DIFF
--- a/app/components/missing_stock_owners_banner_component.rb
+++ b/app/components/missing_stock_owners_banner_component.rb
@@ -28,7 +28,7 @@ class MissingStockOwnersBannerComponent < ViewComponent::Base
     if user.data_coordinator? || user.support?
       "If your organisation does own stock, #{contact_helpdesk_link} to update your details.".html_safe
     else
-      "Ask a data coordinator to add a stock owner. Find your data coordinators on the #{users_link}.</br>
+      "Ask a data coordinator to add a stock owner. Find your data coordinators on the #{users_link}.</br></br>
       If your organisation does own stock, #{contact_helpdesk_link} to update your details.".html_safe
     end
   end
--- a/app/components/missing_stock_owners_banner_component.rb
+++ b/app/components/missing_stock_owners_banner_component.rb
@@ -1,0 +1,64 @@
+class MissingStockOwnersBannerComponent < ViewComponent::Base
+  include Rails.application.routes.url_helpers
+
+  attr_reader :user, :organisation
+
+  HELPDESK_URL = "https://dluhcdigital.atlassian.net/servicedesk/customer/portal/6/group/11".freeze
+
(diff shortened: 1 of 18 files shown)


app/components/missing_stock_owners_banner_component.rb
```diff
@@ -0,0 +1,64 @@
+class MissingStockOwnersBannerComponent < ViewComponent::Base
+  include Rails.application.routes.url_helpers
+
+  attr_reader :user, :organisation
+
+  HELPDESK_URL = "https://dluhcdigital.atlassian.net/servicedesk/customer/portal/6/group/11".freeze
```
natdeanlewissoftwire: not an issue that started here, but we've got this url repeated loads of times in the codebase, can we store it in one place instead?

app/components/missing_stock_owners_banner_component.rb
```diff
@@ -0,0 +1,64 @@
...
+      "Your organisation does not own stock. You must add a stock owner before you can create logs.".html_safe
+    end
+  end
+
+  def banner_text
+    if user.data_coordinator?
+      "If your organisation does own stock, #{contact_helpdesk_link} to update your details.".html_safe
+    else
+      "If your organisation does own stock, #{contact_helpdesk_link} to update your details.</br>
+      Ask a data coordinator to add a stock owner. Find your data coordinators on the #{users_link}.".html_safe
```
natdeanlewissoftwire: are these lines of copy in the wrong order?

spec/requests/lettings_logs_controller_spec.rb
```diff
@@ -291,6 +291,41 @@
             expect(page).to have_content("Managed by")
           end
         end
+
+        context "and organisation does not own stock or have valid stock owners" do
```
natdeanlewissoftwire, line 295: should we check the banner isn't being displayed when we don't want it to be as well?

natdeanlewissoftwire (commented):
nice one! approving now but a couple of thoughts below:

natdeanlewissoftwire (approved):

app/components/missing_stock_owners_banner_component.rb
kosiakkatrina: Sure!

app/components/missing_stock_owners_banner_component.rb
kosiakkatrina: Oh yes they are

spec/requests/lettings_logs_controller_spec.rb
kosiakkatrina: Can do!

PR #2087, opened 2023-12-11
Don't validate locations/schemes and deactivations

kosiakkatrina:
Can't confidently validate schemes, locations and deactivations because there are going to be genuine situations where we're moving invalid resources.
Those would be if a merge happens in a closed collection period all of the dates validations would fail.


natdeanlewissoftwire (approved):
:(

PR #2089, opened 2023-12-13
CLDC-3037 Add bespoke code to label conversion for log CSVs

natdeanlewissoftwire:
Adds a load of code to label conversion for lettings log exports - we had lots of attributes that were being output as codes in "labels" exports because we hadn't added any mapping for them. This PR addresses that.


app/services/csv/lettings_log_csv_service.rb
```diff
@@ -193,6 +201,57 @@ def label_if_boolean_value(value)
...
+    def conventional_yes_no_label(value)
+      return "Yes" if value == 1
+      return "No" if value&.zero?
+    end
+
+    def yes_or_blank_label(value)
+      value == 1 ? "Yes" : nil
+    end
+
+    LETTYPE_LABELS = {
```
kosiakkatrina, line 209: ~Can we get these from the form question instead?~ Oh sorry, these are not questions, nvm

app/services/csv/lettings_log_csv_service.rb
```diff
@@ -177,12 +177,20 @@ def value(attribute, log)
...
 
     def get_label(value, attribute, log)
+      return LETTYPE_LABELS[value] if attribute == "lettype"
+      return IRPRODUCT_LABELS[value] if attribute == "irproduct"
+      return LAR_LABELS[value] if attribute == "lar"
+      return NEWPROP_LABELS[value] if attribute == "newprop"
+      return INCREF_LABELS[value] if attribute == "incref"
+      return conventional_yes_no_label(value) if CONVENTIONAL_YES_NO_ATTRIBUTES.include?(attribute)
+      return "Yes" if YES_OR_BLANK_ATTRIBUTES.include?(attribute) && value == 1
+
```
kosiakkatrina: maybe a switch statement, or nested labels like `LABELS = {lettype: LETTYPE_LABELS, irproduct: IRPRODUCT_LABELS ... }` and `return LABELS[attribute][value] if LABELS[attribute].present?` or something like that. But only if it reads easier, otherwise this is fine

kosiakkatrina (approved):

app/services/csv/lettings_log_csv_service.rb
natdeanlewissoftwire: I like the nested labels, have done, good idea!

PR #2091, opened 2023-12-14
CLDC-3014 scheme csv download improvements

natdeanlewissoftwire:
Optimisations from testing on staging

### DIFF
--- a/app/helpers/schemes_helper.rb
+++ b/app/helpers/schemes_helper.rb
@@ -60,10 +60,10 @@ module SchemesHelper
     when "schemes"
       "You've selected #{pluralize(scheme_count, 'scheme')}."
     when "locations"
-      location_count = schemes.map(&:locations).flatten.count
+      location_count = Location.where(scheme: schemes).count
       "You've selected #{pluralize(location_count, 'location')} from #{pluralize(scheme_count, 'scheme')}."
     when "combined"
-      location_count = schemes.map(&:locations).flatten.count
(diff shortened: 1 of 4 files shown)


app/helpers/schemes_helper.rb
```diff
@@ -60,10 +60,10 @@ def selected_schemes_and_locations_text(download_type, schemes)
     when "schemes"
       "You've selected #{pluralize(scheme_count, 'scheme')}."
     when "locations"
-      location_count = schemes.map(&:locations).flatten.count
+      location_count = Location.where(scheme: schemes).count
```
natdeanlewissoftwire, line 63: Previously was very inefficient for large numbers of schemes, so the "download csv" was taking ages to load

app/services/csv/scheme_csv_service.rb
```diff
@@ -10,17 +10,22 @@ def initialize(download_type:)
     def prepare_csv(schemes)
       CSV.generate(headers: true) do |csv|
         csv << attributes
-        schemes.find_each do |scheme|
-          if @download_type == "schemes"
+
+        case @download_type
```
natdeanlewissoftwire, line 14: I've refactored this to only check download type once, rather than on every loop, and also am using each rather than find_each (even though it's less efficient) because we wanted to preserve ordering. If it's too slow I'll revert this or consider ordering the csv after we've generated it.

kosiakkatrina (approved):

PR #2093, opened 2023-12-14
CLDC-2537 Remove GOV.UK PaaS code

kosiakkatrina:


config/initializers/rack_attack.rb
```diff
@@ -12,7 +10,7 @@
   redis_url = configuration_service.redis_uris.to_a[0][1]
   Rack::Attack.cache.store = ActiveSupport::Cache::RedisCacheStore.new(url: redis_url)
 else
```
RachaelBooth, line 9: Does the `elsif Rails.env.review?` clause still need to be separate?

RachaelBooth (approved):

config/initializers/rack_attack.rb
kosiakkatrina: Ah, good spot, no it doesn't!

PR #2102, opened 2023-12-19
CLDC-2988 Update routing for address pages

kosiakkatrina:
We had defined custom `routed_to?` methods for address fields, which meant that it was possible to navigate to those questions without completing the setup section.
This PR translates the conditions from those `routed_to?` methods to `depends_on` and executes parent `routed_to?` method instead.

Given that we clear the questions we're not routing to a lot of the tests had to be updated too. This was mostly done by filling out the setup section for test logs where appropriate.

This PR also extracts some web stubs for UPRNs to `request_helper.rb`. There are some changes to the UPRN numbers in the tests, these are so that we could reuse the mocks.

### DIFF
--- a/spec/lib/tasks/send_missing_addresses_csv_spec.rb
+++ b/spec/lib/tasks/send_missing_addresses_csv_spec.rb
@@ -17,36 +17,6 @@ RSpec.describe "correct_addresses" do
       Rake.application.rake_require("tasks/send_missing_addresses_csv")
       Rake::Task.define_task(:environment)
       task.reenable
-
-      body_1 = {
-        results: [
-          {
-            DPA: {
-              "POSTCODE": "BS1 1AD",
-              "POST_TOWN": "Bristol",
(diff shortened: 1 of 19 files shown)


spec/lib/tasks/send_missing_addresses_csv_spec.rb
```diff
@@ -62,7 +32,7 @@
 
         before do
           create(:user, :data_provider, organisation:, email: "data_provider1@example.com")
-          create_list(:lettings_log, 7, :imported, startdate: Time.zone.local(2023, 9, 9), address_line1: nil, town_or_city: nil, needstype: 1, old_form_id: "form_1", owning_organisation: organisation, managing_organisation: organisation, created_by: organisation.users.first)
```
kosiakkatrina, line 65: I think this user is inactive, therefore was never getting set

RachaelBooth (approved):

PR #2113, opened 2024-01-02
CLDC-3116 Create homepage

natdeanlewissoftwire:
Adds a homepage for other homepage tickets to add elements to, and adds a home tab to the primary navbar. I've added the welcome message already to allow me to test the routing more comprehensively, all the other sections are accounted for in the other homepage tickets.

The homepage uses the root path ("/"), as this seemed cleaner than a separate home route. As such, the start controller now renders the home page rather than redirecting to the lettings page.

EDIT: Now included in this are the notifications, task, resources and deadlines sections.
image

image


spec/requests/users_controller_spec.rb
```diff
@@ -73,11 +73,11 @@
     end
 
     describe "title link" do
-      it "routes user to the /logs page" do
+      it "routes user to the home page" do
         sign_in user
         get "/", headers:, params: {}
-        follow_redirect!
-        expect(path).to include("/lettings-logs")
+        expect(path).to include("/")
```
kosiakkatrina: could this be to_eq? Or `expect(path).not_to include("/lettings-logs")`, cause otherwise we might not be testing very much here

kosiakkatrina (approved):

app/helpers/home_helper.rb
```diff
@@ -33,15 +33,15 @@ def data_subheading(user, type)
   def data_path(user, type)
     if user.data_provider?
       case type
-      when "lettings" then lettings_logs_path(status: [:in_progress], assigned_to: "you")
-      when "sales" then sales_logs_path(status: [:in_progress], assigned_to: "you")
-      when "misc" then lettings_logs_path(status: [:completed], assigned_to: "you")
+      when "lettings" then lettings_logs_path(status: [:in_progress], assigned_to: "you", years: [""], owning_organisation_select: "all", managing_organisation_select: "all")
+      when "sales" then sales_logs_path(status: [:in_progress], assigned_to: "you", years: [""], owning_organisation_select: "all", managing_organisation_select: "all")
+      when "misc" then lettings_logs_path(status: [:completed], assigned_to: "you", years: [""], owning_organisation_select: "all", managing_organisation_select: "all")
       end
```
kosiakkatrina: Can we call clear filters in any way and then route them to the logs path with relevant filters only? Dw if now

kosiakkatrina (approved):

app/helpers/home_helper.rb
natdeanlewissoftwire: Not without refactoring the filter behaviour (as far as I can think of), which I didn't really want to do as part of this ticket, I think this is the cleanest solution.

PR #2115, opened 2024-01-02
CLDC-2252 Add homepage task section

natdeanlewissoftwire:
Adds task boxes to homepage, different numbers and contents according to the designs:

image

### DIFF
--- a/spec/features/home_page_spec.rb
+++ b/spec/features/home_page_spec.rb
@@ -1,0 +1,147 @@
+require "rails_helper"
+require_relative "form/helpers"
+
+RSpec.describe "Home Page Features" do
+  include Helpers
+
+  context "when the user is a data provider" do
+    let(:user) { FactoryBot.create(:user, name: "Provider") }
+
+    before do
+      create_list(:lettings_log, 6, :in_progress, owning_organisation: user.organisation, created_by: user)
+      create_list(:lettings_log, 2, :in_progress, owning_organisation: user.organisation)
+      create_list(:lettings_log, 4, :completed, owning_organisation: user.organisation, created_by: user)
+      create_list(:lettings_log, 2, :completed)
+      sign_in user
+      visit("/")
+    end
+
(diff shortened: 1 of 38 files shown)


spec/features/home_page_spec.rb
```diff
@@ -0,0 +1,158 @@
...
+
+      it "displays correct data boxes, counts and links" do
+        data_boxes = page.find_all(class: "app-data-box-one-half")
+        expect(data_boxes.count).to eq(2)
+        expect(data_boxes[0].all("a").map(&:text)).to eq(["6", "Your lettings in progress", "View all lettings"])
+        expect(data_boxes[0].all("a").map { |line| line["href"] }).to eq([lettings_logs_path(status: [:in_progress], assigned_to: "you"), lettings_logs_path(status: [:in_progress], assigned_to: "you"), clear_filters_path(filter_type: "lettings_logs")])
+        expect(data_boxes[1].all("a").map(&:text)).to eq(["5", "Your sales in progress", "View all sales"])
+        expect(data_boxes[1].all("a").map { |line| line["href"] }).to eq([sales_logs_path(status: [:in_progress], assigned_to: "you"), sales_logs_path(status: [:in_progress], assigned_to: "you"), clear_filters_path(filter_type: "sales_logs")])
+      end
+    end
```
kosiakkatrina, line 40: I wonder if there's a different way to test this so that we're not locked into the dom as much, but if we're not expecting it to change soon or much, I'm happy with this

kosiakkatrina (approved):
This looks good! I'm concerned about how it will look with the rest of the components on smaller screens, given they're always 2/3 but we can check and adjust that further down the line

spec/features/home_page_spec.rb
natdeanlewissoftwire: yeah agreed, this was the best I could think of that ensure the right roughting was mapped to the right links, otherwise it wouldn't be quite as well tested

PR #2117, opened 2024-01-03
Fix lettings log accessor in date question

kosiakkatrina:


natdeanlewissoftwire (approved):
can we add a failing test too?

kosiakkatrina:
> can we add a failing test too?

@natdeanlewissoftwire yeah, removing `lettings_log` in `app/views/form/page.html.erb` caused some failures in form request spec

PR #2119, opened 2024-01-03
CLDC-2593 Add upcoming deadlines section

kosiakkatrina:
image

### DIFF
--- a/app/views/home/_upcoming_deadlines.html.erb
+++ b/app/views/home/_upcoming_deadlines.html.erb
@@ -3,31 +3,31 @@
 <% current_lettings_form = FormHandler.instance.in_crossover_period? ? FormHandler.instance.previous_lettings_form : FormHandler.instance.current_lettings_form %>
 <% formatted_deadline = "#{current_lettings_form.submission_deadline.strftime('%A')} #{current_lettings_form.submission_deadline.to_formatted_s(:govuk_date)}" %>
 <% if FormHandler.instance.in_crossover_period? %>
-  <p class="govuk-body govuk-body-m"><strong>End of year deadline - <%= formatted_deadline %></strong>: Deadline to submit logs for tenancies starting between <%= collection_start_date(Time.zone.now).to_formatted_s(:govuk_date) %> to <%= collection_end_date(Time.zone.now).to_formatted_s(:govuk_date) %></p>
+  <p class="govuk-body govuk-body-m"><strong>End of year deadline - <%= formatted_deadline %>:</strong> Deadline to submit logs for tenancies starting between <%= collection_start_date(Time.zone.now).to_formatted_s(:govuk_date) %> to <%= collection_end_date(Time.zone.now).to_formatted_s(:govuk_date) %></p>
 <% end %>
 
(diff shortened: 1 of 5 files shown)


app/views/home/_upcoming_deadlines.html.erb
```diff
@@ -0,0 +1,18 @@
...
+
+  <% current_lettings_form = FormHandler.instance.current_lettings_form %>
+  <p class="govuk-body govuk-body-m"><strong><%= current_lettings_form.submission_deadline.to_formatted_s(:govuk_date) %></strong>: Deadline to submit logs for tenancies starting between <%= collection_start_date(Time.zone.now).to_formatted_s(:govuk_date) %> to <%= collection_end_date(Time.zone.now).to_formatted_s(:govuk_date) %></p>
+  <%= govuk_details(summary_text: "Quarterly cut-off dates for 2024 to 2025") do %>
+    <ul class="govuk-list govuk-list--bullet">
+      <li><strong>12 July 2024</strong>: Quarterly cut off date for tenancies and sales starting between 1 April 2024 and 30 June 2024</li>
+      <li><strong>11 Oct 2024</strong>: Quarterly cut off date for tenancies and sales starting between 1 July 2024 and 30 Sept 2024</li>
+      <li><strong>10 Jan 2025</strong>: Quarterly cut off date for tenancies and sales starting between 1 Oct 2024 and 31 Dec 2024</li>
+      <li><strong>6 June 2025</strong>: Deadline to submit logs for tenancies starting between 1 April 2024 to 31 March 2025 </li>
+    </ul>
```
kosiakkatrina: I tried making this a little dynamic, but it got more confusing in the end, so leaving as is given it's only here and would only get updated yearly

app/views/home/_upcoming_deadlines.html.erb
```diff
@@ -0,0 +1,18 @@
+<div class="govuk-grid-row">
+  <h1 class="govuk-heading-l">Upcoming deadlines</h1>
+  <% current_quarter = quarter_for_date(date: Time.zone.now) %>
+  <% if current_quarter.present? %>
+    <p class="govuk-body govuk-body-m"><strong>12 January 2024</strong>: Quarterly cut off date for tenancies and sales starting between 1 October 2023 and 31 December 2023.</p>
```
RachaelBooth: Is this one meant to be using current_quarter?

app/views/home/_upcoming_deadlines.html.erb
RachaelBooth: Are there docs anywhere for things we have to update on a regular cycle? (Be good to avoid the case where we forget until someone points it out)

kosiakkatrina: If there are, I'm not immediately aware of them, but I just added one with the things I could remember and shared it with the team

app/views/home/_upcoming_deadlines.html.erb
kosiakkatrina: Sure is! Thanks!

RachaelBooth (approved):

PR #2120, opened 2024-01-04
CLDC-2253 Add collection resources

kosiakkatrina:
- Update collection resources
- Add collection resources to the homepage, link to guidance page (https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/2121)
- Remove collection resource from org page
- Replace collection resources with the new component on the start page (CLDC-1401)
image

### DIFF
--- a/app/views/layouts/_collection_resources.html.erb
+++ b/app/views/layouts/_collection_resources.html.erb
@@ -5,7 +5,7 @@
   <h2 class="govuk-heading-m">Collection resources</h2>
 <% end %>
 <p class="govuk-body">Use the 2023 to 2024 forms for lettings that start and sales that complete between 1 April 2023 and 31 March 2024.</p>
-<%= govuk_tabs(title: "Monday’s child nursery rhyme") do |c| %>
+<%= govuk_tabs(title: "Collection resources") do |c| %>
   <% if FormHandler.instance.lettings_form_for_start_year(2023) && FormHandler.instance.lettings_form_for_start_year(2023).edit_end_date > Time.zone.today %>
     <% c.with_tab(label: "Lettings 2023/24") do %>
       <%= render DocumentListComponent.new(items: [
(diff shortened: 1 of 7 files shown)


app/views/layouts/_collection_resources.html.erb
```diff
@@ -1,94 +1,59 @@
...
+<% if current_user %>
+  <h1 class="govuk-heading-l">Collection resources</h1>
+  <p class="govuk-body"><strong><%= govuk_link_to "Guidance for submitting social housing lettings and sales data (CORE)", guidance_path %></strong></p>
+<% else %>
   <h2 class="govuk-heading-m">Collection resources</h2>
-  <p class="govuk-body-s">For lettings starting during 1 April 2023 to 31 March 2024 and sales completing during the same period, use the 2023/24 forms.</p>
-
+<% end %>
+<p class="govuk-body">Use the 2023 to 2024 forms for lettings that start and sales that complete between 1 April 2023 and 31 March 2024.</p>
+<%= govuk_tabs(title: "Monday’s child nursery rhyme") do |c| %>
```
natdeanlewissoftwire: what's this?

natdeanlewissoftwire: haha found it https://govuk-components.netlify.app/components/tabs/

natdeanlewissoftwire: i think the "c" then stands for "child", so we should probably change that as well

natdeanlewissoftwire (approved):
I think we should also add a test that this is appearing correctly on the home page - though I appreciate there isn't a home page spec yet

app/views/layouts/_collection_resources.html.erb
kosiakkatrina: 🙃 I should stop copying things

kosiakkatrina: c is collection resources now. Not entirely sure what it's meant to be tho, cause we call `.with_tabs` on it, so `collection.with_tabs` kinda makes sense, but `child.with_tabs` doesn't really

kosiakkatrina:
> I think we should also add a test that this is appearing correctly on the home page - though I appreciate there isn't a home page spec yet

@natdeanlewissoftwire I think start request spec tests for content on homepage, are you thinking we should add something in addition to that?

PR #2124, opened 2024-01-05
CLDC-2590 Add about this service section

kosiakkatrina:


app/views/layouts/_about_core.html.erb
```diff
@@ -0,0 +1,7 @@
+<% if current_user.present? %>
```
RachaelBooth: Is there a better way to decide which formatting is needed here? It feels more like a coincidence that it relates to whether there's a current user than that being actually what we care about - although I guess that is the same thing that says whether we are on the home page or the start page.

kosiakkatrina: Yeah, I see what you mean. I couldn't think of a different way, given it's the same path. 
Could duplicate the content in the page views themselves as opposed to having it in a separate component, but as you said we're determining which page to show based on the user anyway. Let me know if you have a preference

natdeanlewissoftwire: yeah we could alternatively set the home page to be "/home" rather than "/", but that wouldn't be my preference

natdeanlewissoftwire (approved):
lgtm

PR #2131, opened 2024-01-09
CLDC-2255 Add homepage notifications

natdeanlewissoftwire:
Adds notifications to the homepage using the wonderful [unread gem](https://github.com/ledermann/unread) to handle user-notification management. 

image


app/controllers/notifications_controller.rb
```diff
@@ -0,0 +1,19 @@
+class NotificationsController < ApplicationController
+  def dismiss
```
natdeanlewissoftwire, line 2: no auth here as users can see notifications even if they're unauthenticated in some cases

db/migrate/20240108145545_create_notification.rb
```diff
@@ -0,0 +1,14 @@
+class CreateNotification < ActiveRecord::Migration[7.0]
+  def change
+    create_table :notifications do |t|
```
natdeanlewissoftwire, line 3: we could add more restrictions on this table, at the moment it doesn't restrict things like start_date < end_date and all data fields are null-allowed - the requirements for future notifications are quite vague so I'ev kept it as open as possible for now but we might revisit this later

natdeanlewissoftwire: also, not thrilled about saving html in the database (in page_content), but the thinking (as I understand it) is that if we can't deploy we can still add a notification through the database if required, in which case this is necessary

RachaelBooth: Hmm, yeah also not thrilled by html in the db but I can't really see a way round it if we want to be able to add notifications without a code release and we want to be able to put styling / headers / links in the pages.

app/models/notification.rb
```diff
@@ -0,0 +1,14 @@
+class Notification < ApplicationRecord
+  acts_as_readable
+
+  scope :active, -> { where("start_date <= ?", Time.zone.now).where("end_date >= ?", Time.zone.now) }
```
kosiakkatrina: would something like this work? `where("start_date <= ? AND end_date >= ?", Time.zone.now, Time.zone.now)`

app/views/layouts/application.html.erb
```diff
@@ -99,6 +99,14 @@
...
     end %>
 
+    <% unless current_page?(notifications_path) %>
+      <% if current_user&.active_unread_notifications.present? %>
+        <%= render "notifications/notification_banner", notification_count: current_user.active_unread_notifications.count, notification: current_user.newest_active_unread_notification %>
+      <% elsif current_user.blank? && Notification.active_unauthenticated_notifications.present? %>
+        <%= render "notifications/notification_banner", notification_count: Notification.active_unauthenticated_notifications.count, notification: Notification.newest_active_unauthenticated_notification %>
+      <% end %>
+    <% end %>
+
```
kosiakkatrina: Would it make sense to move this to notification_banner and just have the render here?

kosiakkatrina (approved):

app/models/notification.rb
natdeanlewissoftwire: yeah, like it :D

app/views/layouts/application.html.erb
natdeanlewissoftwire: I think it would be a bit more confusing to do that, as the notification_banner only renders when there are notifications so I'd like to keep at least a condition representing that here. I'll look at assigning the local variables intelligently so I can just have one render statement which would be clearer

PR #2132, opened 2024-01-09
CLDC-2995 Update renttype_detail export csv

kosiakkatrina:
We want to start exporting `rent_type` column as `renttype_detail`.
We also want to be change the mapping for this column and will be introducing it in the 2024/25 bulk upload forms.

I have considered migrating to a new column internally and updating all of the values for the records but landed on a less drastic solution of mapping to a new `renttype_detail` attribute upon exports instead.

Given that we have already been exporting `rent_type` in the csv_exports as `rent_type_detail` we probably want to replace that with `renttype_detail` with the new values.

This will mean that any CSVs exported before this change will have different `rent_type_detail`(what `rent_type` is currently getting exported as) value to the `renttype_detail` values after this change.

2023/24 logs will need to get reexported to get the new `renttype_detail` column out


natdeanlewissoftwire:
Do we need to make anyone aware that `rent_type_detail` now has a different meaning on new exports than it used to?

app/models/derived_variables/lettings_log_variables.rb
```diff
@@ -29,6 +29,15 @@ module DerivedVariables::LettingsLogVariables
     5 => 8, # "Other intermediate rent product"  => "Another Intermediate Rent basis"
   }.freeze
 
+  RENTTYPE_DETAIL_MAPPING = {
```
natdeanlewissoftwire, line 32: part of me hates this but it's definitely clearer than a +1

spec/services/csv/lettings_log_csv_service_spec.rb
```diff
@@ -89,7 +89,7 @@
 
       it "removes some log attributes related to questions from the headers and replaces them with their derived values in the correct order" do
         expect(headers).not_to include "rent_type"
-        expect(headers).to include(*%w[wrent renttype rent_type_detail])
+        expect(headers).to include(*%w[wrent renttype renttype_detail])
```
natdeanlewissoftwire, line 92: can we also check the headers don't include `rent_type_detail` just for sanity's sake? Would make it clearer to anyone looking at this in the future that it's a previous column that's now been removed

natdeanlewissoftwire (approved):

app/models/derived_variables/lettings_log_variables.rb
kosiakkatrina: Yeah, I did +1 at some point, but went against it cause it looked as if there's some sort of logic behind the mapping

PR #2135, opened 2024-01-10
CLDC-3034 Fix bulk upload error on empty field bug

natdeanlewissoftwire:
Bug: If a user bulk uploaded a log that failed the rent range validation, we were adding an error to the scheme_id field. However, for general needs logs we were assigning this error to the field [nil].

Fix: Don't include scheme_id (or location_id etc.) in bulk upload fields for error assignment when this data has not been provided in a general needs log.


config/locales/en.yml
```diff
@@ -353,7 +353,7 @@ en:
     financial:
       tshortfall:
         outstanding_amount_not_expected: "You cannot answer the outstanding amount question if you don’t have outstanding rent or charges"
-        more_than_rent: "Enter a value less less than the basic rent amount"
+        more_than_rent: "Enter a value less than the basic rent amount"
```
kosiakkatrina, line 356: 🙃

kosiakkatrina (approved):

PR #2136, opened 2024-01-11
CLDC-3110: Base income validations on all tenants for lettings logs

RachaelBooth:


kosiakkatrina (approved):

app/models/validations/financial_validations.rb
```diff
@@ -24,29 +24,61 @@ def validate_net_income_uc_proportion(record)
...
           message: I18n.t("validations.financial.earnings.over_hard_max", hard_max:),
         )
         record.errors.add(
-          :ecstat1,
+          :hhmemb,
           :over_hard_max,
-          message: I18n.t("validations.financial.ecstat.over_hard_max", earnings: format_as_currency(record.earnings), frequency:),
+          message: I18n.t("validations.financial.hhmemb.earnings.over_hard_max", earnings: format_as_currency(record.earnings), frequency:),
         )
+        ecstat_fields = %i[ecstat1 ecstat2 ecstat3 ecstat4 ecstat5 ecstat6 ecstat7 ecstat8]
```
kosiakkatrina: Would it make sense to add these errors up to hhmemb? I wonder if this would make sense on BU error summary

config/locales/en.yml
```diff
@@ -361,8 +361,8 @@ en:
       benefits:
         part_or_full_time: "Answer cannot be ‘all’ for income from Universal Credit, state pensions or benefits if the tenant or their partner works part-time or full-time"
       earnings:
-        over_hard_max: "Net income cannot be greater than %{hard_max} per week given the household’s working situation"
-        under_hard_min: "Net income cannot be less than %{hard_min} per week given the household’s working situation"
+        over_hard_max: "The household's income cannot be greater than %{hard_max} per week given the household’s working situation"
+        under_hard_min: "The household's income cannot be less than %{hard_min} per week given the household’s working situation"
```
kosiakkatrina: Would this need to be household’s instead of household's ?

app/models/form/lettings/pages/net_income_value_check.rb
```diff
@@ -36,6 +36,6 @@ def questions
   end
 
   def interruption_screen_question_ids
-    %w[incfreq earnings ecstat1]
+    %w[incfreq earnings hhmemb ecstat1 age2 ecstat2 age3 ecstat3 age4 ecstat4 age5 ecstat5 age6 ecstat6 age7 ecstat7 age8 ecstat8]
```
kosiakkatrina, line 39: Do we hide not routed to questions from the list somewhere?

RachaelBooth: I think when we display the interruption screen we filter for questions that are routed to - so we've got similar things where all the person fields are listed for the soft validation checks around pregnancy.

config/locales/en.yml
RachaelBooth: Drat, yes

kosiakkatrina (approved):

PR #2141, opened 2024-01-12
CLDC-2964 Save and export duplicate logs 

kosiakkatrina:
We want to start exporting some data about duplicate logs (sales and lettings, csv and xml)
`duplicate_set_id` field is now exported marking all logs with the same ID as duplicates of each other.

Every time a log becomes a duplicate or gets deduplicated, we add or remove a `duplicate_log_id` on a log as appropriate.

This PR also contains a task to update logs with `duplicate_log_id` for both sales and lettings for already existing duplicates. We only do it for 2023 logs as 2022 collection is already closed and logs can't be edited.


app/controllers/form_controller.rb
```diff
@@ -171,8 +171,18 @@ def successful_redirect_path
         return correcting_duplicate_logs_redirect_path
       end
 
-      if @log.lettings? && current_user.lettings_logs.duplicate_logs(@log).count.positive?
-        return send("lettings_log_duplicate_logs_path", @log, original_log_id: @log.id)
+      if @log.lettings?
```
natdeanlewissoftwire: why don't we do this for sales?

lib/tasks/set_duplicate_references.rake
```diff
@@ -0,0 +1,27 @@
...
+
+    duplicate_set.each do |log_id|
+      DuplicateLogReference.create(log_id:, log_type: "LettingsLog", duplicate_set_id:)
+    end
+  end
+end
+
+def generate_new_duplicate_set_id
+  loop do
+    duplicate_set_id = SecureRandom.random_number(1_000_000)
```
natdeanlewissoftwire: why are we doing this? seems theoretically problematic, can't we just use a number greater than the largest existing duplicate set id?

natdeanlewissoftwire (commented):
looks good, couple of questions - can revisit this in more detail tomorrow but I'll leave it for now :)

app/controllers/form_controller.rb
natdeanlewissoftwire: (and similarly there are a couple of other bits in this PR that seem to be lettings only, just checking that's right for my understanding)

lib/tasks/set_duplicate_references.rake
kosiakkatrina: Hmm, I see what you mean. I think I'm kinda worried about the race condition with incremental values especially when we create there records for the existing logs. And I can't put a uniqueness condition on the db layer cause there would be several records with this ID. Any thoughts?

app/controllers/form_controller.rb
kosiakkatrina: I'm not sure how I missed these. I've added the sales part here and to the delete logs controller. Let me know if there's anywhere else you can see that I missed! Thanks!

lib/tasks/set_duplicate_references.rake
natdeanlewissoftwire: can we set the duplicate set id just to be the lowest id of all the logs in the set?

natdeanlewissoftwire: that assumes logs can't be in multiple separate sets which they _shouldn't_ be able to be right?

kosiakkatrina: Yeah, but it kinda means that we'd need to recreate the IDs on all of the logs if a log with a lower ID joins the duplicates set. I feel like it might get kinda fiddly

natdeanlewissoftwire: I don't think we'd need to recreate the ids (doesn't matter if it's not actually the lowest), I'm just thinking it would be an easy way to get a unique id for each set

natdeanlewissoftwire: Or we could make a duplicate set table and just use ids in that table? To be clear I think this is probably fine as is but just seems like it's reinventing unique indexing that we can use db indexes to handle for us

kosiakkatrina: I see what you mean with not having to changes the IDs. But what would be the benefit of using log IDs over a random number? Cause for me it feels that if we start using log IDs as an identifier that ties duplicate logs together for anyone who reads this code in the future it might seem like there was a meaning behind it, whereas really we just want a unique number and it shouldn't matter what it is.

Maybe I'm missing something obvious, but given these wouldn't be unique I'm not sure if we can use indexes for that

And in terms of making a duplicate set table, do you mean in addition to the one added in this PR  or just modelling this one differently?

natdeanlewissoftwire (approved):
nice work!

app/controllers/delete_logs_controller.rb
```diff
@@ -213,4 +215,30 @@ def discard(logs)
...
+  def remove_lettings_duplicate_set_ids(logs)
+    duplicate_set_ids = []
+    logs.each do |log|
+      if log.duplicate_set_id.present?
+        duplicate_set_ids << log.duplicate_set_id
+        log.update!(duplicate_set_id: nil)
+      end
+    end
+    duplicate_set_ids.uniq.each do |duplicate_set_id|
+      LettingsLog.where(duplicate_set_id:).update!(duplicate_set_id: nil) if LettingsLog.where(duplicate_set_id:).count == 1
```
natdeanlewissoftwire, line 228: if this is the only Lettings/Sales specific line, can we just make this bit call a separate function and the rest of the function it's in be shared?

app/controllers/form_controller.rb
```diff
@@ -293,10 +304,22 @@ def remove_fixed_duplicate_set_ids(log)
     SalesLog.find_by(duplicate_set_id:)&.update!(duplicate_set_id: nil) if log.sales? && SalesLog.where(duplicate_set_id:).count == 1
   end
 
-  def new_duplicate_set_id
-    loop do
-      duplicate_set_id = SecureRandom.random_number(1_000_000)
-      return duplicate_set_id unless LettingsLog.exists?(duplicate_set_id:)
+  def new_duplicate_set_id(log)
+    if log.lettings?
+      LettingsLog.maximum(:duplicate_set_id).to_i + 1
```
natdeanlewissoftwire, line 308: is it possible all lettings logs will have nil `duplicate_set_id` at some point? If so we should set a default value I think

Also, just checking stats are aware that a duplicate set id won't be for a unique set but can be reassigned to a different set later on?

natdeanlewissoftwire: ah ignore the first bit, forgot `to_i` returns 0 on nil

lib/tasks/set_duplicate_references.rake
```diff
@@ -1,7 +1,7 @@
 desc "Set duplicate references for sales and lettings logs"
 task set_duplicate_references: :environment do
   SalesLog.filter_by_year(2023).duplicate_sets.each do |duplicate_set|
-    duplicate_set_id = generate_new_duplicate_set_id
+    duplicate_set_id = SalesLog.maximum(:duplicate_set_id).to_i + 1
```
natdeanlewissoftwire, line 4: do you think it's worth making this a function so we can describe what it does (and as we're doing it in 4 places)

natdeanlewissoftwire (approved):
lgtm, just some refactoring thoughts :)

app/controllers/form_controller.rb
kosiakkatrina: Yup! I thought this would break when I ran the tests, but all good!

lib/tasks/set_duplicate_references.rake
kosiakkatrina: hmm, so I think the it's only here, in the controller we've got it in the `new_duplicate_set_id` method.
I'm not sure if there's a good place to put this for a rake task and given we should only have to run it once, I assume this would get deleted relatively quickly anyways

PR #2142, opened 2024-01-12
CLDC-2972 Warn users about answers that will be cleared in bulk upload

natdeanlewissoftwire:
Adds an error report page that just shows answers to be deleted:

image

also adds some links and warning text to the confirmation page:

image

and links the error report on the fix choice page:

image


app/controllers/form_controller.rb
```diff
@@ -15,7 +15,7 @@ def submit_form
         redirect_to(successful_redirect_path)
       else
         mandatory_questions_with_no_response.map do |question|
-          @log.errors.add question.id.to_sym, question.unanswered_error_message
+          @log.errors.add question.id.to_sym, question.unanswered_error_message, category: :not_answered
```
natdeanlewissoftwire, line 18: this category has been added to all non-setup errors that are triggered when a compulsory field is blank, so that later on we can only show the user answers that will be deleted **and are not already blank** in the error count

Gemfile.lock
```diff
@@ -436,6 +436,7 @@ GEM
 PLATFORMS
   arm64-darwin-21
   arm64-darwin-22
+  arm64-darwin-23
```
natdeanlewissoftwire, line 439: discussed with @kosiakkatrina, including this as my machine wants this to be in the Gemfile and I am having to remove it from commits all the time

app/controllers/form_controller.rb
kosiakkatrina: would :blank make sense here instead?

app/helpers/logs_helper.rb
```diff
@@ -63,4 +64,16 @@ def logs_and_soft_validations_warning(bulk_upload)
...
     "You have tried to upload #{bulk_upload.total_logs_count ? pluralize(bulk_upload.total_logs_count, 'log') : 'logs'}. There are errors in #{pluralize(bulk_upload.logs_with_errors_count, 'log')}, and #{pluralize(bulk_upload.bulk_upload_errors.count, 'error')} in total."
   end
+
+  def deleted_errors_warning_text(bulk_upload)
+    unique_answers_to_be_cleared_count = unique_answers_to_be_cleared(bulk_upload).count
+    this_or_these = unique_answers_to_be_cleared_count == 1 ? "this" : "these"
+    it_is_or_they_are = unique_answers_to_be_cleared_count == 1 ? "it is" : "they are"
+
+    "#{pluralize(unique_answers_to_be_cleared_count, 'answer')} will be deleted because #{it_is_or_they_are} invalid. You will have to answer #{this_or_these} #{'question'.pluralize(unique_answers_to_be_cleared_count)} again on the site."
+  end
```
kosiakkatrina, line 74: Is there a situation where `unique_answers_to_be_cleared_count` is 0?

app/controllers/form_controller.rb
kosiakkatrina: Oh I see that the ones in row parsers kinda follow the error message path, in which case :not_answered is fine if it feels more consistent

app/helpers/logs_helper.rb
natdeanlewissoftwire: yep, there's an edge case where there are errors but they're all "not_answered" errors so won't really be cleared because they're already blank:
 
image

natdeanlewissoftwire: I think not all the copy makes sense though, I'll send this sc to olivia and susanna and see what they think

app/views/bulk_upload_lettings_resume/confirm.html.erb
```diff
@@ -5,14 +5,23 @@
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds">
     <span class="govuk-caption-l">Bulk upload for lettings (<%= @bulk_upload.year_combo %>)</span>
-    <h1 class="govuk-heading-l">You have chosen to upload all logs from this bulk upload.</h1>
+    <h1 class="govuk-heading-l">Are you sure you want to upload all logs from this bulk upload?</h1>
 
-    <p class="govuk-body"><%= logs_and_errors_warning(@bulk_upload) %></p>
+    <p class="govuk-body">
+      <%= logs_and_errors_warning(@bulk_upload) %>
+      <%= govuk_link_to "View the error report", deletion_report_bulk_upload_lettings_result_path %>
```
kosiakkatrina: I would have expected this to be the full error report instead of the deletion one, are we meant to link to it twice?

app/views/bulk_upload_lettings_resume/fix_choice.html.erb
```diff
@@ -18,6 +18,10 @@
         <%= @form.recommendation %>
       </div>
 
+      <div class="govuk-body">
+        <%= govuk_link_to "View the error report", deletion_report_bulk_upload_lettings_result_path %>
```
kosiakkatrina: Same as above, don't we want to link to the full report here?

app/helpers/logs_helper.rb
kosiakkatrina: Ah yeah, just saw that, make sense! How come in this situation the table isn't empty?

natdeanlewissoftwire: because the two errors are the :not_answered type and so not included in the count, but still errors - olivia and I agreed it would be misleading not to show them as the user might think everything was fine and then be surprised that they had missing data in the service (when they had just failed to fill it in at all)

app/views/bulk_upload_lettings_resume/confirm.html.erb
natdeanlewissoftwire: me too, the designs have been clarified, you're right :D

app/helpers/logs_helper.rb
natdeanlewissoftwire: we now don't display the link to this page in this case 👍

kosiakkatrina (approved):

PR #2144, opened 2024-01-16
CLDC-3087 Bulk update schemes and locations

kosiakkatrina:
We want to allow bulk updating scheme and location data in some cases (genuine errors - not changing schemes completely).
 
This would be done by comparing the downloaded and edited scheme and location csvs to the original ones and updating the values that have changed.

- We want to avoid putting active schemes and locations back to incomplete states so empty answers will be ignored.
- The rake tasks allow changing the organisation id on schemes, in which case those schemes would be unassigned from logs where they're used. Organisation IDs will not be changed if there are logs using this scheme in closed collection years.
- We will not update/re-export any logs with start dates earlier than the beginning of the previous collection period, because there is no way for us to export those logs and they might no longer be relevant. We will keep a record of the old logs that would have been affected.


spec/lib/tasks/update_schemes_and_locations_from_csv_spec.rb
```diff
@@ -0,0 +1,454 @@
...
+        expect(schemes[0].primary_client_group).to eq("People with drug problems")
+        expect(schemes[0].has_other_client_group).to eq("No")
+        expect(schemes[0].owning_organisation).to eq(different_organisation)
+        expect(schemes[0].created_at).to eq(Time.zone.local(2021, 4, 1))
+        expect(schemes[0].total_units).to eq(2)
+      end
+
+      it "updates the lettings log if scheme has changed owning organisation" do
+        create(:organisation_relationship, parent_organisation: schemes[0].owning_organisation, child_organisation: different_organisation)
+        closed_collection_lettings_log.update!(startdate: Time.zone.now)
```
RachaelBooth: Seems a bit annoying to have to do this - overall I think it makes sense but I had to think about why. If there's an obvious way to avoid this I'd be tempted but I don't think it's that important.

RachaelBooth (approved):
Nice!

spec/lib/tasks/update_schemes_and_locations_from_csv_spec.rb
kosiakkatrina: Yeah, you're right, it's confusing. I renamed and flipped the dates we set so it's only set to old dates in relevant tests

PR #2147, opened 2024-01-17
CLDC-3127 Amend renewal question wording for 2024

kosiakkatrina:
Amend renewal question header and hint text for 2024 and disable collection year validations on review apps only so that these changes could be tested

### DIFF
--- a/app/models/form/lettings/questions/renewal.rb
+++ b/app/models/form/lettings/questions/renewal.rb
@@ -3,15 +3,32 @@ class Form::Lettings::Questions::Renewal < ::Form::Question
     super
     @id = "renewal"
     @check_answer_label = "Property renewal"
-    @header = "Is this letting a renewal?"
+    @header = header_text
     @type = "radio"
     @answer_options = ANSWER_OPTIONS
-    @hint_text = "A renewal is a letting to the same tenant in the same property. If the property was previously being used as temporary accommodation, then answer 'no'"
+    @hint_text = hint_text
     @question_number = 4
   end
 
   ANSWER_OPTIONS = {
     "1" => { "value" => "Yes" },
     "0" => { "value" => "No" },
-  }.freeze
+  }
+  .freeze
+
+  def header_text
+    if form.start_date && form.start_date.year >= 2024
+      "Is this letting a renewal of social housing to the same tenant in the same property?"
+    else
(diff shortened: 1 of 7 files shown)


app/models/form/lettings/questions/renewal.rb
```diff
@@ -3,15 +3,32 @@ def initialize(id, hsh, page)
...
 
   ANSWER_OPTIONS = {
     "1" => { "value" => "Yes" },
     "0" => { "value" => "No" },
-  }.freeze
+  }
+  .freeze
+
+  def header_text
+    if form.start_date && form.start_date.year >= 2024
```
natdeanlewissoftwire: think it's worth saving this as a method as we're going to use this exact thing loads?

natdeanlewissoftwire: or can we use `&.`?

app/services/feature_toggle.rb
```diff
@@ -1,7 +1,11 @@
 class FeatureToggle
   # Disable check on preview apps to allow for testing of future forms
   def self.saledate_collection_window_validation_enabled?
-    true
+    Rails.env.production? || Rails.env.test? || Rails.env.staging?
```
natdeanlewissoftwire, line 4: have you checked this won't have any unexpected changes on local envs?

natdeanlewissoftwire (approved):
lgtm! just thrown a few ideas out but have approved

app/services/feature_toggle.rb
kosiakkatrina: Good point! It deffo removes the validation locally as well. It doesn't matter too much cause we can always toggle things on/off locally ourselves anyway, so any preference on this? I think it aligns with some other toggles we have atm

app/models/form/lettings/questions/renewal.rb
kosiakkatrina: `&.` will still execute `>=` and it doesn't know what to do in `nil` situations. I've extracted it to form tho, let me know if that's better

natdeanlewissoftwire (approved):

app/services/feature_toggle.rb
natdeanlewissoftwire: yeah fine as is by me :D

PR #2148, opened 2024-01-17
CLDC-3131 Amend previous rent basis answer options and hint for 2024

kosiakkatrina:


natdeanlewissoftwire (approved):
lgtm - do we need to change the mapping in the row parser too?

app/models/form/lettings/questions/previous_let_type.rb
```diff
@@ -21,4 +21,15 @@ def initialize(id, hsh, page)
     "divider" => { "value" => true },
     "3" => { "value" => "Don’t know" },
   }.freeze
+
+  ANSWER_OPTIONS_2024 = {
```
natdeanlewissoftwire: `ANSWER_OPTIONS_FROM_2024` perhaps? just to make clear it's not for this year only and then reverting

kosiakkatrina: Is `ANSWER_OPTIONS_AFTER_2024` ok? It would align with the form method

PR #2149, opened 2024-01-17
CLDC-2442 Add archived forms

kosiakkatrina:
We already display the logs that we have in the logs lists

But up until now we were not keeping any forms from before previous_form so we were putting the logs into the current form which would often have different questions.

So with the changes  introduced here, we will keep an extra form so that the logs for period before previous could be displayed properly. Named this form `archived_lettings`/`archived_sales`

We only keep this archived form around until the previous_collection editing window is closed.

We will also stop displaying all the logs for all collection years and instead we will only show the logs that have forms available for them.

For the CSV downloads there would be no changes at the moment, because currently we only have previous and current collection forms (cause the previous form is fully closed now), but in April we would have current (2024/25), previous (2023/24) and “archived” (2022/23) forms, so the questions from all 3 forms would be present in the CSV so we have until then to figure out what we want to do with it.

Why did csv files changed so much in this PR?
Because I wanted to test that adding extra forms won't change anything in the csv downloads for the time being, but to do that I had to change a date to one which would reflect the forms we have now (no archived form as the editing is closed for previous form). So the things that csv responds with don't really change, only the tests


app/models/form_handler.rb
```diff
@@ -41,7 +41,12 @@ def sales_forms
       "current_sales" => Form.new(nil, current_collection_start_year, SALES_SECTIONS, "sales"),
       "previous_sales" => Form.new(nil, previous_collection_start_year, SALES_SECTIONS, "sales"),
       "next_sales" => Form.new(nil, next_collection_start_year, SALES_SECTIONS, "sales"),
+      "archived_sales" => Form.new(nil, previous_collection_start_year - 1, SALES_SECTIONS, "sales"),
```
natdeanlewissoftwire, line 44: don't we need an equivalent line for lettings?

app/models/form_handler.rb
```diff
@@ -198,6 +207,22 @@ def sales_earliest_open_for_editing_collection_start_date(now: Time.zone.now)
     end
   end
 
+  def years_of_available_lettings_forms
+    years = []
+    lettings_forms.each_value do |form|
```
natdeanlewissoftwire, line 215: can we combine these two functions into one that takes the forms as input?

app/models/lettings_log.rb
```diff
@@ -40,6 +40,13 @@ class LettingsLog < Log
   belongs_to :managing_organisation, class_name: "Organisation", optional: true
 
   scope :filter_by_year, ->(year) { where(startdate: Time.zone.local(year.to_i, 4, 1)...Time.zone.local(year.to_i + 1, 4, 1)) }
+  scope :filter_by_years_or_nil, lambda { |years, _user = nil|
```
natdeanlewissoftwire, line 43: can we put this on the log model instead?

natdeanlewissoftwire: (given we can use startdate = saledate for sales logs)

natdeanlewissoftwire (approved):

app/models/form_handler.rb
kosiakkatrina: Do you mean something like this?
`FormHandler.instance.years_of_available_forms(FormHandler.instance.lettings_forms)` 

I feel like it might be more confusing than splitting these methods up, let me know if you meant something different

app/models/lettings_log.rb
kosiakkatrina: I'm not sure I fully understand, it wouldn't work for sales if I move this method as it is right now, cause startdate is not a column in sales logs. Is there a change to it that you had in mind that would make it work for both?

app/models/form_handler.rb
kosiakkatrina: Yup! Added it now! It worked for this and next year because 21/22 and 22/23 are defined in JSON, but needed to be updated for newer forms. Thanks!

PR #2152, opened 2024-01-19
CLDC-3210: Handle "R" properly for income/savings in sales bulk upload

RachaelBooth:


app/services/bulk_upload/sales/year2023/row_parser.rb
```diff
@@ -841,17 +841,17 @@ def attributes_for_log
     attributes["ethnic"] = field_32
     attributes["national"] = field_33
 
-    attributes["income1nk"] = field_78.present? ? 0 : 1
-    attributes["income1"] = field_78
+    attributes["income1nk"] = field_78 == "R" ? 1 : 0
+    attributes["income1"] = field_78.to_i if attributes["income1nk"]&.zero? && field_78&.match(/\A\d+\z/)
```
RachaelBooth, line 845: This allows more digits than are in the specification, but means if you enter e.g. 200,000 instead of 20,000 you'll get validation messages about being too high rather than about not having answered it.

RachaelBooth:
N.B. I've not looked at 2022, because people shouldn't be uploading those any more.

app/services/bulk_upload/sales/year2023/row_parser.rb
kosiakkatrina: Would this have been the case before this PR too?

kosiakkatrina (approved):

app/services/bulk_upload/sales/year2023/row_parser.rb
RachaelBooth: Yes, it would.

PR #2153, opened 2024-01-22
CLDC-2442 Remove 2022/23 BU code

kosiakkatrina:


app/models/forms/bulk_upload_lettings/prepare_your_file.rb
```diff
@@ -10,9 +10,7 @@ class PrepareYourFile
 
       def view_path
         case year
-        when 2022
-          "bulk_upload_lettings_logs/forms/prepare_your_file_2022"
-        else
+        when 2023
```
kosiakkatrina, line 13: This and a couple of other case statements were defaulting to 2023 previously, but they wouldn't anymore, not sure if there'd be a situation where it would need to.
Could change it to always return 2023 without the case statement, but thought I'd keep the case here so that it's clear where to add 2024 stuff. Not sure if that's the better approach tho

natdeanlewissoftwire: I think this is fine, might be worth raising in PO review  - I know we're not doing e.g. legacy templates in 24/25 so that's somewhere it definitely doesn't make sense to have a default value, so I think this approach is better

public/files/bulk-upload-lettings-specification-2022-23.xlsx
natdeanlewissoftwire, line 1: should we remove the paper forms too?

spec/features/bulk_upload_lettings_logs_spec.rb
```diff
@@ -31,37 +35,27 @@
...
-
-        expect(page).to have_content("Upload lettings logs in bulk (2022/23)")
+        expect(page).to have_content("Upload lettings logs in bulk (2023/24)")
         expect(page).to have_content("Upload your file")
         click_button("Upload")
 
         allow_any_instance_of(Forms::BulkUploadLettings::UploadYourFile).to receive(:`).and_return("not a csv")
 
         expect(page).to have_content("Select which file to upload")
-        attach_file "file", file_fixture("2021_22_lettings_bulk_upload.xlsx")
```
natdeanlewissoftwire, line 64: 😮

spec/services/bulk_upload/sales/validator_spec.rb
```diff
@@ -31,7 +31,7 @@
 
     context "when file has too many columns" do
       before do
-        file.write((%w[a] * 127).join(","))
+        file.write((%w[a] * 143).join(","))
```
natdeanlewissoftwire: can we make this less hard coded?

natdeanlewissoftwire (approved):
seems like this was a lot of fiddly work, nice job

spec/services/bulk_upload/sales/validator_spec.rb
kosiakkatrina: Does it look better now?

PR #2158, opened 2024-01-24
CLDC-3190 Add lettings BU for 24/25

kosiakkatrina:
Duplicate 2023 lettings bulk upload into 2024 lettings bulk upload and map questions to the new field numbers
Update duplicate log fields to sum individual charges and rent instead of using tcharge, because we no longer ask for tcharge in the template (field_132)

### DIFF
--- a/app/services/bulk_upload/lettings/year2024/csv_parser.rb
+++ b/app/services/bulk_upload/lettings/year2024/csv_parser.rb
@@ -3,7 +3,7 @@ require "csv"
 class BulkUpload::Lettings::Year2024::CsvParser
   include CollectionTimeHelper
 
-  FIELDS = 134
+  FIELDS = 130
   MAX_COLUMNS = 142
   FORM_YEAR = 2024
 
@@ -73,7 +73,7 @@ class BulkUpload::Lettings::Year2024::CsvParser
 private
 
   def default_field_numbers
-    [5, nil, nil, 15, 16, nil, 13, 40, 41, 42, 43, 46, 52, 56, 60, 64, 68, 72, 76, 47, 53, 57, 61, 65, 69, 73, 77, 51, 55, 59, 63, 67, 71, 75, 50, 54, 58, 62, 66, 70, 74, 78, 48, 49, 79, 81, 82, 123, 124, 122, 120, 102, 103, nil, 83, 84, 85, 86, 87, 88, 104, 109, 107, 108, 106, 100, 101, 105, 110, 111, 112, 113, 114, 115, 116, 117, 118, 119, 126, 128, 129, 130, 131, 132, 127, 125, 133, 134, 33, 34, 35, 36, 37, 38, nil, 7, 8, 9, 28, 14, 32, 29, 30, 31, 26, 27, 25, 23, 24, nil, 1, 3, 2, 80, nil, 121, 44, 89, 98, 92, 95, 90, 91, 93, 94, 97, 96, 99, 10, 11, 12, 45, 39, 6, 4, 17, 18, 19, 20, 21, 22].map { |h| h.present? && h.to_s.match?(/^[0-9]+$/) ? "field_#{h}" : "field_blank" }
+    (1..130).map { |h| h.present? && h.to_s.match?(/^[0-9]+$/) ? "field_#{h}" : "field_blank" }
   end
 
   def field_numbers
--- a/app/services/bulk_upload/lettings/year2024/csv_parser.rb
+++ b/app/services/bulk_upload/lettings/year2024/csv_parser.rb
@@ -73,7 +73,7 @@ class BulkUpload::Lettings::Year2024::CsvParser
 private
 
   def default_field_numbers
-    (1..130).map { |h| h.present? && h.to_s.match?(/^[0-9]+$/) ? "field_#{h}" : "field_blank" }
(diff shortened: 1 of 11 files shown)


app/services/bulk_upload/lettings/year2024/csv_parser.rb
```diff
@@ -0,0 +1,113 @@
+require "csv"
+
+class BulkUpload::Lettings::Year2024::CsvParser
+  include CollectionTimeHelper
+
+  FIELDS = 130
+  MAX_COLUMNS = 142
```
natdeanlewissoftwire: do we not want to reduce this by 4 as there are 4 fewer fields?

app/services/bulk_upload/lettings/year2024/csv_parser.rb
```diff
@@ -0,0 +1,113 @@
...
+      0
+    end
+  end
+
+  def col_offset
+    with_headers? ? 1 : 0
+  end
+
+  def cols
+    @cols ||= ("A".."EL").to_a
```
natdeanlewissoftwire: similarly, if there are fewer fields we shouldn't go all the way to EL right?

app/services/bulk_upload/lettings/year2024/csv_parser.rb
```diff
@@ -0,0 +1,113 @@
...
+  def wrong_template_for_year?
+    collection_start_year_for_date(first_record_start_date) != FORM_YEAR
+  rescue Date::Error
+    false
+  end
+
+private
+
+  def default_field_numbers
+    (1..130).map { |h| h.present? && h.to_s.match?(/^[0-9]+$/) ? "field_#{h}" : "field_blank" }
```
natdeanlewissoftwire: use FIELDS here?

app/services/bulk_upload/lettings/year2024/row_parser.rb
```diff
@@ -0,0 +1,1468 @@
+class BulkUpload::Lettings::Year2024::RowParser
+  include ActiveModel::Model
+  include ActiveModel::Attributes
+  include InterruptionScreenHelper
+
+  QUESTIONS = {
+    field_1: "Which organisation owns this property?",
```
natdeanlewissoftwire, line 7: I haven't checked these, happy to trust you if you're happy they're right 😅

app/services/bulk_upload/lettings/year2024/row_parser.rb
```diff
@@ -0,0 +1,1468 @@
...
+
+  validates :field_5,
+            presence: {
+              if: proc { supported_housing? },
+              message: I18n.t("validations.not_answered", question: "scheme code"),
+              category: :setup,
+            },
+            on: :after_log
+
+  validates :field_6,
```
natdeanlewissoftwire, line 332: why are we making validation changes here? I thought all that would be left to separate tickets...

app/services/bulk_upload/lettings/year2024/row_parser.rb
```diff
@@ -0,0 +1,1468 @@
...
+      errors.add(:field_122, error_message) unless general_needs? # household_charge
+      errors.add(:field_125, error_message) # brent
+      errors.add(:field_126, error_message) # scharge
+      errors.add(:field_127, error_message) # pscharge
+      errors.add(:field_128, error_message) # chcharge
+    end
+  end
+
+  def field_mapping_for_errors
+    {
```
natdeanlewissoftwire, line 851: why have we removed tcharge and offered from this?

natdeanlewissoftwire (approved):
looks good! Just a couple things I wasn't sure about but I'm sure they're intentional

app/services/bulk_upload/lettings/year2024/csv_parser.rb
kosiakkatrina: You're right! It looks like it only goes up to EA!

app/services/bulk_upload/lettings/year2024/csv_parser.rb
kosiakkatrina: So EL is 142, EA is 131 so will update it to that. Thank you!

app/services/bulk_upload/lettings/year2024/row_parser.rb
kosiakkatrina: Yeah, there were a lot of mappings, I hope they're right and I don't expect you to check every single one of them 😅

app/services/bulk_upload/lettings/year2024/row_parser.rb
kosiakkatrina: I think if you're referring to the location fields, it's cause we now only have use the new scheme and location fields, so we don't need the validation which determines which scheme/location/management group fields get answered, but we still want to validate the location. So this is just the simplified version of that. If you have other validations in mind that shouldn't be a part of this let me know!

app/services/bulk_upload/lettings/year2024/row_parser.rb
kosiakkatrina: We're no longer asking offered question and tcharge I'll add back and will map to the other charges fields (we're no longer ask tcharge either)

PR #2159, opened 2024-01-24
CLDC-3194: Update form end dates in the form model

RachaelBooth:


app/models/form.rb
```diff
@@ -3,19 +3,30 @@ class Form
...
+      edit_end_date: Time.zone.local(2023, 11, 20),
+    },
+    2023 => {
+      submission_deadline: Time.zone.local(2024, 6, 7),
+    },
+    2024 => {
+      submission_deadline: Time.zone.local(2025, 6, 6),
+    },
+    :default => {
+      submission_deadline: ->(start_year) { Time.zone.local(start_year + 1, 6, 1) },
```
RachaelBooth, line 19: I figured for the displayed submission deadline it was better to have a slightly-early default than anything else.

kosiakkatrina: Love this change! Was hoping we will ~outgrow~ outlast all the if statements, but now that we're keeping an extra form this is a great refactor!

kosiakkatrina (approved):

app/models/form.rb
natdeanlewissoftwire: 🥳

natdeanlewissoftwire (approved):

PR #2160, opened 2024-01-24
CLDC-3130 Update first let question for 24/25

natdeanlewissoftwire:


spec/models/form/lettings/questions/first_time_property_let_as_social_housing_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+  end
+
+  context "with 2024/25 form" do
+    before do
+      allow(form).to receive(:start_year_after_2024?).and_return(true)
+    end
+
+    it "has the correct answer_options" do
+      expect(question.answer_options).to eq({
+                                              "1" => { "value" => "Yes", "hint" => "This property was previously used for other purposes or is a new-build." },
```
natdeanlewissoftwire, line 41: fixing this...

app/models/form/lettings/questions/first_time_property_let_as_social_housing.rb
```diff
@@ -7,12 +7,22 @@ def initialize(id, hsh, page)
     @type = "radio"
     @check_answers_card_number = 0
     @hint_text = ""
-    @answer_options = ANSWER_OPTIONS
+    @answer_options = answer_options
```
kosiakkatrina: I think you should be able to remove this line

kosiakkatrina (approved):

app/models/form/lettings/questions/first_time_property_let_as_social_housing.rb
natdeanlewissoftwire: nice, you're right

PR #2162, opened 2024-01-25
CLDC-3133 Remove times offered from 24/25

natdeanlewissoftwire:


kosiakkatrina (approved):

kosiakkatrina (commented):
Would this need to be removed from BU too?

PR #2163, opened 2024-01-25
CLDC-3137 Add hint text for joint question for 2024

kosiakkatrina:


app/models/form/lettings/questions/joint.rb
```diff
@@ -6,7 +6,7 @@ def initialize(id, hsh, page)
     @header = "Is this a joint tenancy?"
     @type = "radio"
     @check_answers_card_number = 0
-    @hint_text = ""
+    @hint_text = form.start_year_after_2024? ? "This is where two or more people are named on the tenancy agreement" : ""
```
natdeanlewissoftwire, line 9: I think for simple changes like this I'll also just do them inline rather than separate, doesn't really matter but good to be consistent

natdeanlewissoftwire (approved):

PR #2164, opened 2024-01-25
CLDC-3136 Move void date guidance to the top

kosiakkatrina:
image


natdeanlewissoftwire (approved):
lgtm, is that definitely how we want the question number displaying though? rather than at the very top?

kosiakkatrina:
> lgtm, is that definitely how we want the question number displaying though? rather than at the very top?

I think that's what we do for other pages, cause there might be more than 1 question on the page

PR #2167, opened 2024-01-25
CLDC-3135 Infer beds for 24/25 bedsits

natdeanlewissoftwire:


app/models/derived_variables/lettings_log_variables.rb
```diff
@@ -82,6 +82,9 @@ def set_derived_fields!
         self.prevten = 30 if owning_organisation&.provider_type == "LA"
       end
     end
+    if is_bedsit? && form.start_year_after_2024?
+      self.beds = 1
+    end
```
kosiakkatrina: Would it make sense to group all the 2024 things together under a single if statement?

app/models/derived_variables/lettings_log_variables.rb
```diff
@@ -168,6 +171,9 @@ def clear_inapplicable_derived_values!
       self.wchair = nil
       self.location_id = nil
     end
+    if unittype_gn_changed? && unittype_gn_was == 2 && form.start_year_after_2024?
+      self.beds = nil
+    end
```
kosiakkatrina: Would this make more sense with the reset fields methods? Like `clear_inapplicable_derived_values!`

natdeanlewissoftwire: This _is_ in `clear_inapplicable_derived_values`... Am I missing something?

app/models/derived_variables/lettings_log_variables.rb
natdeanlewissoftwire: Sure, I'll nest this for future use

app/models/derived_variables/lettings_log_variables.rb
kosiakkatrina: Oh, haha, I thought it was in `set_derived_fields!` together with the change above 🤦 My other comment now doesn't make sense either, ignore me

kosiakkatrina (approved):

PR #2172, opened 2024-01-25
CLDC-3161 Update shared ownership type hint and details

kosiakkatrina:


app/models/form/sales/questions/shared_ownership_type.rb
```diff
@@ -4,13 +4,20 @@ def initialize(id, hsh, page)
...
-    @hint_text = "A shared ownership sale is when the purchaser buys up to 75% of the property value and pays rent to the Private Registered Provider (PRP) on the remaining portion"
     @top_guidance_partial = guidance_partial
     @type = "radio"
     @answer_options = answer_options
     @question_number = 4
   end
 
+  def hint_text
+    if form.start_year_after_2024?
+      "When the purchaser buys an initial share of up to 75% of the property value and pays rent to the Private Registered Provider (PRP) on the remaining portion, or a subsequent staircasing transaction"
```
natdeanlewissoftwire, line 15: nit: I think we've been adding full stops to the end of hint texts (though I see this wasn't there in the previous version either)

app/models/form/sales/questions/shared_ownership_type.rb
```diff
@@ -37,6 +44,10 @@ def answer_options
   end
 
   def guidance_partial
-    "shared_ownership_type_definitions" if form.start_date.year >= 2023
+    if form.start_year_after_2024?
+      "shared_ownership_type_definitions_2024"
+    elsif form.start_date.year >= 2023
```
natdeanlewissoftwire, line 49: would == 2023 be clearer here?

natdeanlewissoftwire (approved):

app/models/form/sales/questions/shared_ownership_type.rb
kosiakkatrina: I kept it as this cause that's what it already was, I don't think it makes much of a difference 🤷

PR #2176, opened 2024-01-26
CLDC-3164 Change discounted ownership value to hard validation

kosiakkatrina:
The hard validation already existed, updated the error message and tests


spec/models/validations/sales/financial_validations_spec.rb
```diff
@@ -339,4 +339,244 @@
...
+          expect(record.errors["value"]).to be_empty
+          expect(record.errors["deposit"]).to be_empty
+          expect(record.errors["ownershipsch"]).to be_empty
+          expect(record.errors["discount"]).to be_empty
+          expect(record.errors["grant"]).to be_empty
+        end
+      end
+
+      context "when ownership is not discounted" do
+        let(:record) { FactoryBot.build(:sales_log, mortgage: 10_000, deposit: 5_000, grant: 3_000, value: 20_000, discount: 10, ownershipsch: 1, saledate: Time.zone.local(2023, 4, 1)) }
```
RachaelBooth: This one should be in 2024, shouldn't it?

spec/models/validations/sales/soft_validations_spec.rb
```diff
@@ -912,6 +912,15 @@
         expect(record).not_to be_discounted_ownership_value_invalid
       end
     end
+
+    context "when it is a 2024 log" do
```
RachaelBooth, line 928: Do we want to adjust the soft validation to not run for 2024? I guess it doesn't really matter because it'll be caught by the hard one before anything from the soft comes into play, but it seems weird to still have it.

spec/models/validations/sales/financial_validations_spec.rb
kosiakkatrina: Ah, yes, good spot!

spec/models/validations/sales/soft_validations_spec.rb
kosiakkatrina: I think it's already only running for 2023! Added this test because it wasn't covered

RachaelBooth (approved):

PR #2177, opened 2024-01-26
CLDC-3159: Move buyer interview question to setup section

RachaelBooth:


app/models/form/sales/subsections/household_characteristics.rb
```diff
@@ -8,7 +8,7 @@ def initialize(id, hsh, section)
 
   def pages
     @pages ||= [
-      Form::Sales::Pages::BuyerInterview.new(nil, nil, self),
+      (Form::Sales::Pages::BuyerInterview.new(nil, nil, self) unless form.start_year_after_2024?),
```
RachaelBooth, line 11: There may want to be some refactoring when PrivacyNotice is also moved, but it makes sense to leave it until then.

natdeanlewissoftwire (approved):
lgtm - 

nit - no CLDC code in the branch name, not necessary to change it but thought I'd flag in case you hadn't noticed

Also, this will change the order of export for sales 24/25 logs, I assume that's desired.

PR #2180, opened 2024-01-29
CLDC-3166 Only trigger savings validation if mortgage used

kosiakkatrina:


app/models/validations/sales/soft_validations.rb
```diff
@@ -60,7 +60,7 @@ def savings_over_soft_max?
   end
 
   def deposit_over_soft_max?
-    return unless savings && deposit
+    return unless savings && deposit && mortgage_used?
```
natdeanlewissoftwire, line 63: do we want to apply this change to the soft min, too? (and/or the hard max/min?)

natdeanlewissoftwire (approved):

app/models/validations/sales/soft_validations.rb
kosiakkatrina: I don't think there is a soft min or hard validations and the ticket specifically points of this calculation `deposit > savings * 4 / 3` so I think it's just this one

PR #2187, opened 2024-01-29
CLDC-3155 Add accessible_register option and hint text for 24/25

natdeanlewissoftwire:


db/schema.rb
```diff
@@ -464,8 +465,6 @@
     t.string "reader_type", null: false
     t.bigint "reader_id"
     t.datetime "timestamp", precision: nil, null: false
-    t.datetime "created_at", null: false
```
natdeanlewissoftwire: ?

kosiakkatrina (commented):
There's `validate_lettings_allocation` and `letting_allocation_unknown`  in BU row parser, I suspect that would need to get updated?
Also, in the export we've got `attribute_hash["chr"] = 2 if attribute_hash["chr"]&.zero?` etc, I suspect that we'd need to add it for this as well?

natdeanlewissoftwire:
Thanks - good spot, I'll wait til the lettings row parser is merged to main and add that update too

kosiakkatrina (approved):

PR #2192, opened 2024-01-30
CLDC-3151 Update reason for leaving for 24/25

natdeanlewissoftwire:


app/models/form/lettings/questions/reason.rb
```diff
@@ -16,114 +15,217 @@ def initialize(id, hsh, page)
...
+        },
+      }.freeze
+    else
+      {
+        "40" => {
+          "value" => "End of assured shorthold tenancy (no fault)",
+        },
+        "41" => {
+          "value" => "End of assured shorthold tenancy (eviction or tenant at fault)",
+        },
```
kosiakkatrina: Picky, but could we format this as 1 option per line?
```
...
"40" => { "value" => "End of assured shorthold tenancy (no fault)" },
"41" => { "value" => "End of assured shorthold tenancy (eviction or tenant at fault)" },
...
```

kosiakkatrina (approved):

app/models/form/lettings/questions/reason.rb
natdeanlewissoftwire: you bet, that'll be much nicer

PR #2193, opened 2024-01-30
CLDC-3092 Update nationality question for 24/25

kosiakkatrina:
Update nationality questions in both sales and lettings. I considered keeping the same field, but ended up adding new ones because the codes we're saving have completely changed.

This adds changes for sales BU, lettings BU is to come when https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/2158 gets merged.

Ended up not using the [GitHub - alphagov/govuk-country-and-territory-autocomplete: An autocomplete widget that uses data from Country and Territory Register.](https://github.com/alphagov/govuk-country-and-territory-autocomplete?tab=readme-ov-file) cause we already have the accessible autocomplete and ability to add synonyms, plus we're storing different values so would need to change those anyways


app/helpers/question_view_helper.rb
```diff
@@ -14,9 +14,13 @@ def legend(question, page_header, conditional)
   end
 
   def answer_option_synonyms(resource)
-    return unless resource.instance_of?(Scheme)
+    return unless resource.instance_of?(Scheme) || resource.is_a?(Hash)
```
natdeanlewissoftwire, line 17: any reason you didn't go for instance_of? here?

app/models/form/lettings/questions/nationality_all.rb
```diff
@@ -0,0 +1,222 @@
...
+    "776" => { "name" => "Tonga", "synonyms" => "TO" },
+    "780" => { "name" => "Trinidad and Tobago", "synonyms" => "TT" },
+    "788" => { "name" => "Tunisia", "synonyms" => "TN" },
+    "792" => { "name" => "Turkey", "synonyms" => "TR" },
+    "795" => { "name" => "Turkmenistan", "synonyms" => "TM" },
+    "798" => { "name" => "Tuvalu", "synonyms" => "TV" },
+    "800" => { "name" => "Uganda", "synonyms" => "UG" },
+    "804" => { "name" => "Ukraine", "synonyms" => "UA" },
+    "784" => { "name" => "United Arab Emirates", "synonyms" => "AE" },
+    "826" => { "name" => "United Kingdom", "synonyms" => "GB,UK,England,Wales,Scotland,Northern Ireland" },
```
natdeanlewissoftwire: do we want to allow great britain/britain here too?

app/models/form/sales/questions/nationality_all.rb
```diff
@@ -0,0 +1,221 @@
...
+    @check_answer_label = "Buyer #{buyer_index}’s nationality"
+    @header = "Enter a nationality"
+    @type = "select"
+    @answer_options = ANSWER_OPTIONS
+    @check_answers_card_number = buyer_index
+    @question_number = buyer_index == 1 ? 24 : 32
+  end
+
+  ANSWER_OPTIONS = {
+    "" => "Select an option",
```
natdeanlewissoftwire: can we store this list somewhere central?

app/models/form/sales/pages/nationality1.rb
```diff
@@ -13,6 +13,13 @@ def initialize(id, hsh, subsection)
   end
 
   def questions
-    @questions ||= [Form::Sales::Questions::Nationality1.new(nil, nil, self)]
+    @questions ||= if form.start_year_after_2024?
```
natdeanlewissoftwire: can we take this opportunity to rename this page (and associated questions etc) buyer1_nationality for clarity?

app/models/form/lettings/questions/nationality_all.rb
```diff
@@ -0,0 +1,222 @@
...
+    "634" => { "name" => "Qatar", "synonyms" => "QA" },
+    "642" => { "name" => "Romania", "synonyms" => "RO" },
+    "643" => { "name" => "Russia", "synonyms" => "RU" },
+    "646" => { "name" => "Rwanda", "synonyms" => "RW" },
+    "659" => { "name" => "Saint Kitts and Nevis", "synonyms" => "" },
+    "662" => { "name" => "Saint Lucia", "synonyms" => "" },
+    "670" => { "name" => "Saint Vincent and the Grenadines", "synonyms" => "" },
+    "882" => { "name" => "Samoa", "synonyms" => "WS" },
+    "674" => { "name" => "San Marino", "synonyms" => "SM" },
+    "678" => { "name" => "Sao Tome and Principe", "synonyms" => "ST" },
```
natdeanlewissoftwire: we should have accents on this (I saw we had on cote d'ivoire) so I don't think we're being consistent accent-wise

natdeanlewissoftwire (approved):
lgtm!

app/helpers/question_view_helper.rb
kosiakkatrina: No reason, I don't think it makes a difference here, so just preference

app/models/form/sales/questions/nationality_all.rb
kosiakkatrina: I thought about that, but I feel like it's very question specific so putting it into a constant might be inviting for it to get reused by other things where it shouldn't
let me know if you feel strongly about this and I'm missing any non obvious benefits

RachaelBooth: I was wondering about this - the main thing I can think of is not having to update it in two places (lettings and sales) if it changes. But I imagine it doesn't change very often, save maybe adding some synonyms. 

That said, are there likely to be things that would want to reuse it but shouldn't? I might be missing something here, but it does seem like anything using a nationality list should want the same list (and also that there wouldn't be other places using a nationality list).

kosiakkatrina: Yeah, it's unlikely we'd want it somewhere else. I think I was thinking about the local authorities and not countries and was worried about the structure changing as opposed to the content, cause it's quite question specific. I've extracted it now tho, let me know if it looks better

RachaelBooth: Yeah, I think that works.

PR #2198, opened 2024-01-31
CLDC-3174 Validate stairowned is not equal stairbought

kosiakkatrina:


natdeanlewissoftwire (approved):
looks good, though I don't understand at all why in the real world these figures couldn't be the same!?

PR #2200, opened 2024-01-31
CLDC-3169: Infer soctenant from prevten / prevtenbuy2

RachaelBooth:


app/models/derived_variables/sales_log_variables.rb
```diff
@@ -24,6 +24,10 @@ def set_derived_fields!
     self.hhmemb = number_of_household_members
     self.hhtype = household_type
 
+    if form.start_year_after_2024?
+      self.soctenant = prevten_was_social_housing? ? 1 : 2
+    end
+
```
kosiakkatrina: We might want to check that `saledate.present?`, cause at some point 2024 form will become the default form and we've also got a default soctenant value here

kosiakkatrina: Also, do we always have a value for soctenant? Isn't there a situation where this is irrelevant? Maybe if it's an outright sale

RachaelBooth: Good point - I think we should only be doing this when it's shared ownership.

kosiakkatrina (approved):

PR #2203, opened 2024-02-02
CLDC-3172 Update routing for staircase transactions

kosiakkatrina:
If staircase is yes, we don't want to route to questions 83-87. I've only added the routing changes to questions 83 and 84, because the rest of them already have a condition that depends on 84 being answered

### DIFF
--- a/app/models/form/sales/pages/buyer_previous.rb
+++ b/app/models/form/sales/pages/buyer_previous.rb
@@ -12,7 +12,7 @@ class Form::Sales::Pages::BuyerPrevious < ::Form::Page
   end
 
   def routed_to?(log, _current_user)
-    return false if log.staircase == 1 && log.form.start_year_after_2024?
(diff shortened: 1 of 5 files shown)


app/models/form/sales/pages/buyer_previous.rb
```diff
@@ -10,4 +10,10 @@ def questions
       Form::Sales::Questions::BuyerPrevious.new(nil, nil, self, joint_purchase: @joint_purchase),
     ]
   end
+
+  def routed_to?(log, _current_user)
+    return false if log.staircase == 1 && log.form.start_year_after_2024?
```
natdeanlewissoftwire: nit: can we make staircase == 1 a method on the log model so it's clearer what value 1 means?

natdeanlewissoftwire (approved):

PR #2204, opened 2024-02-02
CLDC-3175 Add max stairbought validation

kosiakkatrina:


spec/models/validations/sales/sale_information_validations_spec.rb
```diff
@@ -467,4 +467,293 @@
...
+      Timecop.freeze(now)
+      Singleton.__init__(FormHandler)
+    end
+
+    after do
+      Timecop.return
+      Singleton.__init__(FormHandler)
+    end
+
+    context "when ownership type is Shared Ownership (new model lease)" do
```
RachaelBooth: Can these tests be parameterised at all?

spec/models/validations/sales/sale_information_validations_spec.rb
```diff
@@ -467,4 +467,293 @@
       end
     end
   end
+
+  describe "#validate_stairbought" do
+    let(:now) { Time.zone.local(2024, 4, 4) }
```
RachaelBooth, line 544: I'm assuming that this means the record ends up being a 2024 log, but that definitely feels like I'm making an assumption - would it make sense to fix the start year of the log directly (as the relevant point for these tests)?

kosiakkatrina: We're setting it on the record `let(:record) { build(:sales_log, ownershipsch: 1, type: 30, saledate: now) }` and it's also used in Timecop.freeze, otherwise we get date validations. Maybe it's a bit clearer now that it's refactored

spec/models/validations/sales/sale_information_validations_spec.rb
kosiakkatrina: For sure!

RachaelBooth (approved):

PR #2207, opened 2024-02-02
CLDC-3179 Update routing for previous postcode

kosiakkatrina:
There's a validation for this, but for 2024 we want to infer previous postcode to be the same as current for discounted sales only

### DIFF
--- a/app/models/derived_variables/sales_log_variables.rb
+++ b/app/models/derived_variables/sales_log_variables.rb
@@ -38,6 +38,10 @@ module DerivedVariables::SalesLogVariables
     self.nationality_all = nationality_all_group if nationality_uk_or_prefers_not_to_say?
     self.nationality_all_buyer2 = nationality_all_buyer2_group if nationality2_uk_or_prefers_not_to_say?
 
+    if saledate && form.start_year_after_2024? && discounted_ownership_sale?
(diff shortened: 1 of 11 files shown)


app/models/derived_variables/sales_log_variables.rb
```diff
@@ -38,6 +38,10 @@ def set_derived_fields!
     self.nationality_all = nationality_all_group if nationality_uk_or_prefers_not_to_say?
     self.nationality_all_buyer2 = nationality_all_buyer2_group if nationality2_uk_or_prefers_not_to_say?
 
+    if saledate && form.start_year_after_2024? && discounted_ownership_sale?
+      self.ppostcode_full = postcode_full
```
natdeanlewissoftwire: why are we checking saledate here? we already check `startdate` in `start_year_after_2024?`

natdeanlewissoftwire (approved):

RachaelBooth (approved):

app/models/derived_variables/sales_log_variables.rb
kosiakkatrina: We check the `start_date` of the form not the log and the log might have the default form assigned before it has the date filled in

PR #2210, opened 2024-02-05
CLDC-3176 Make deposit dynamically optional

kosiakkatrina:
Make deposit dynamically optional for 2024 onwards only and only if stair owned is 100%. 
When deposit is optional we also want to display a different hint text.

To achieve that, we construct 2 versions of deposit pages, one with optional being false, and one with optional being true (this one only get's constructed for 2024 onwards).


app/models/form/sales/subsections/shared_ownership_scheme.rb
```diff
@@ -37,14 +37,16 @@ def pages
       Form::Sales::Pages::MortgageLenderOther.new("mortgage_lender_other_shared_ownership", nil, self, ownershipsch: 1),
       Form::Sales::Pages::MortgageLength.new("mortgage_length_shared_ownership", nil, self, ownershipsch: 1),
       Form::Sales::Pages::ExtraBorrowing.new("extra_borrowing_shared_ownership", nil, self, ownershipsch: 1),
-      Form::Sales::Pages::AboutDepositWithDiscount.new(nil, nil, self),
-      Form::Sales::Pages::AboutDepositWithoutDiscount.new("about_deposit_shared_ownership", nil, self, ownershipsch: 1),
+      Form::Sales::Pages::AboutDepositWithDiscount.new("about_deposit_with_discount", nil, self, optional: false),
+      form.start_year_after_2024? ? Form::Sales::Pages::AboutDepositWithDiscount.new("about_deposit_with_discount_optional", nil, self, optional: true) : nil,
```
RachaelBooth: May want to consider ```(Form::Sales::Pages::... if form_start_year_after_2024?)``` instead of the ternary - I think that's what we've used elsewhere, not sure what ends up being neater in this case.

RachaelBooth (approved):

app/models/form/sales/subsections/shared_ownership_scheme.rb
kosiakkatrina: Ah yes, good idea!

PR #2216, opened 2024-02-05
CLDC-3202 Update year filters for 24/25

natdeanlewissoftwire:
Makes the filters automatically display the current and two previous collection years

### DIFF
--- a/app/helpers/filters_helper.rb
+++ b/app/helpers/filters_helper.rb
@@ -93,11 +93,11 @@ module FiltersHelper
   end
 
   def collection_year_options
-    if FeatureToggle.collection_2024_2025_year_enabled?
-      { "2024": "2024/25", "2023": "2023/24", "2022": "2022/23" }
-    else
-      { "2023": "2023/24", "2022": "2022/23", "2021": "2021/22" }
-    end
+    {
+      current_collection_start_year.to_s => year_combo(current_collection_start_year),
+      previous_collection_start_year.to_s => year_combo(previous_collection_start_year),
+      twice_previous_collection_start_year.to_s => year_combo(twice_previous_collection_start_year),
+    }
   end
 
   def filters_applied_text(filter_type)
--- a/app/helpers/filters_helper.rb
+++ b/app/helpers/filters_helper.rb
@@ -1,6 +1,5 @@
 module FiltersHelper
   include CollectionTimeHelper
-  require "timecop"
 
   def filter_selected?(filter, value, filter_type)
     return false unless session[session_name_for(filter_type)]
@@ -96,9 +95,6 @@ module FiltersHelper
   end
 
   def collection_year_options
-    current_collection_start_year = 2024
-    previous_collection_start_year = 2023
-    archived_collection_start_year = 2022
     {
       current_collection_start_year.to_s => year_combo(current_collection_start_year),
       previous_collection_start_year.to_s => year_combo(previous_collection_start_year),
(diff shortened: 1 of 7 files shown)


app/helpers/filters_helper.rb
```diff
@@ -93,7 +93,11 @@ def assigned_to_filter_options(user)
   end
 
   def collection_year_options
-    { "2023": "2023/24", "2022": "2022/23", "2021": "2021/22" }
+    {
+      current_collection_start_year.to_s => year_combo(current_collection_start_year),
+      previous_collection_start_year.to_s => year_combo(previous_collection_start_year),
+      twice_previous_collection_start_year.to_s => year_combo(twice_previous_collection_start_year),
+    }
```
kosiakkatrina: There's `years_of_available_lettings_forms` and `years_of_available_sales_forms` in the `FormHandler` which would only return the years for the forms we currently hold (that does include `next` form as well, so not sure if it would be useful to reuse that)

db/schema.rb
```diff
@@ -303,6 +303,7 @@
     t.integer "scharge_value_check"
     t.integer "pscharge_value_check"
     t.integer "duplicate_set_id"
+    t.integer "accessible_register"
```
kosiakkatrina: Should this be here?

spec/helpers/filters_helper_spec.rb
```diff
@@ -1,4 +1,5 @@
 require "rails_helper"
+require "timecop"
```
kosiakkatrina: I don't think you need this here, but you might need to `include CollectionTimeHelper` in `FiltersHelper`

app/helpers/filters_helper.rb
kosiakkatrina: Also, not sure if helpful, but I've been calling the form before the previous `archived` in the rest of the codebase

db/schema.rb
natdeanlewissoftwire: ah rebasing issue let me fix

spec/helpers/filters_helper_spec.rb
natdeanlewissoftwire: You're right, thanks

app/helpers/filters_helper.rb
natdeanlewissoftwire: Ah that could be useful, I'll have a look, thanks!

natdeanlewissoftwire: Think I'll leave as is to be explicit about the three years we want here (as it's not even the standard previous, current, next we used to have elsewhere) but I'll rename to `archived`

app/helpers/filters_helper.rb
```diff
@@ -93,7 +96,14 @@ def assigned_to_filter_options(user)
   end
 
   def collection_year_options
-    { "2023": "2023/24", "2022": "2022/23", "2021": "2021/22" }
+    current_collection_start_year = 2024
```
natdeanlewissoftwire: these three lines are for PO review, will be removed after

kosiakkatrina (approved):

natdeanlewissoftwire:
do not merge label added until PO review lines have been removed

spec/helpers/filters_helper_spec.rb
```diff
@@ -240,12 +240,39 @@
...
+        Timecop.freeze(Time.zone.local(2023, 5, 1)) do
+          example.run
+        end
+        Timecop.return
+      end
+
+      it "has the correct options" do
+        expect(collection_year_options).to eq(
+          {
+            # "2023" => "2023/24", "2022" => "2022/23", "2021" => "2021/22" revert to this after PO
```
natdeanlewissoftwire: so is this

PR #2220, opened 2024-02-06
CLDC-3143: Update hint text for gender questions for 2024

RachaelBooth:


app/models/form/sales/questions/gender_identity2.rb
```diff
@@ -22,4 +22,10 @@ def initialize(id, hsh, page)
     "X" => { "value" => "Non-binary" },
     "R" => { "value" => "Buyer prefers not to say" },
   }.freeze
+
+  def hint_text
+    if form.start_year_after_2024?
+      "This should be however they personally choose to identify from the options below. This may or may not be the same as their biological sex or the sex they were assigned at birth."
+    end
```
natdeanlewissoftwire: I've got a feeling we need
`else
      ""`
 here

natdeanlewissoftwire (approved):
lgtm! just one comment (that applies to 2 files)

app/models/form/sales/questions/gender_identity2.rb
RachaelBooth: Interestingly there are actually quite a few places we let hint text be nil, including previously this one. I think when rendered nil and "" will be treated as the same.

As a readability point though, maybe it's clearer to do something explicit once defining this function - will ponder.

PR #2221, opened 2024-02-06
CLDC-3196 Update prepare file and guidance pages for 24/25 bulk upload

natdeanlewissoftwire:


app/models/forms/bulk_upload_sales/prepare_your_file.rb
```diff
@@ -32,21 +32,26 @@ def legacy_template_path
         case year
         when 2023
           "/files/bulk-upload-sales-legacy-template-2023-24.xlsx"
+        when 2024
```
natdeanlewissoftwire, line 35: there's quite a lot of WET code here, I've made a tech debt ticket to combine sales/lettings bu files where possible but am not tackling it here

kosiakkatrina: I have found it easier to update separate files so far and it nicely follows the other independent BU files. If we're looking to deduplicate stuff we probably want to make sure we're introducing abstractions that help us work within the codebase, which I'm happy for us to do

kosiakkatrina:
This seems to all have been covered by [CLDC-3207](https://digital.dclg.gov.uk/jira/browse/CLDC-3207) already

[CLDC-3207]: https://dluhcdigital.atlassian.net/browse/CLDC-3207?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #2226, opened 2024-02-07
CLDC-3134 Update wheelchair hint text

kosiakkatrina:


spec/models/form/lettings/questions/wheelchair_spec.rb
```diff
@@ -39,4 +45,14 @@
   it "is not marked as derived" do
     expect(question.derived?).to be false
   end
+
```
RachaelBooth, line 48: I think the existing test around hint text ought to be in a specific context too, since what we'll expect the default to be will change at some point.

RachaelBooth (approved):

spec/models/form/lettings/questions/wheelchair_spec.rb
kosiakkatrina: hm, I've stubbed `start_year_after_2024?: false` as a default, so the tests shouldn't start failing based on when we run them, or did you mean doing it purely for visibility?

PR #2228, opened 2024-02-08
CLDC-3186 Display reported by question

kosiakkatrina:
We have been inferring reported_by for sales up until now, because it was introduced mid-year. For 24/25 onwards we want it to behave the same as lettings managing organisation

Also removed `sales_managing_organisation_enabled?` and `merge_organisations_enabled?` feature flags which have need true for a while now

### DIFF
--- a/app/models/form/sales/pages/managing_organisation.rb
+++ b/app/models/form/sales/pages/managing_organisation.rb
@@ -13,7 +13,6 @@ class Form::Sales::Pages::ManagingOrganisation < ::Form::Page
   def routed_to?(log, current_user)
     return false unless current_user
     return false unless current_user.support?
-    return false unless FeatureToggle.sales_managing_organisation_enabled?
     return false unless log.owning_organisation
 
     log.owning_organisation.managing_agents.count >= 1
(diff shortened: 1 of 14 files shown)


app/models/form/sales/pages/managing_organisation.rb
```diff
@@ -12,10 +12,20 @@ def questions
...
+      return false unless organisation
+      return false if log.owning_organisation != organisation && !organisation.holds_own_stock?
+      return true unless organisation.holds_own_stock?
+
+      organisation.managing_agents.count >= 1
+    else
+      return false unless current_user.support?
+      return false unless log.owning_organisation
+
+      log.owning_organisation.managing_agents.count >= 1
```
natdeanlewissoftwire, line 28: just double checking - we can guarantee managing_agents includes absorbed orgs right?

natdeanlewissoftwire (approved):

app/models/form/sales/pages/managing_organisation.rb
kosiakkatrina: Do you mean displaying merged orgs as managing agents options? I think we would if there was a relationship, but should we in any other case?

natdeanlewissoftwire: Yep that's all I was asking :)

PR #2229, opened 2024-02-08
CLDC-3126: Update privacy notice, including rationalising csv download ordering

RachaelBooth:
* Moves the privacy notice / declaration question to the setup section for both lettings and sales
* Minor rewording to that question
* Related update to csv downloads to ensure that deprecated questions from the start of the next section don't appear before this question in the download
* Related update to csv downloads to explicitly fix the base for the column order to the current form


app/models/form_handler.rb
```diff
@@ -50,35 +50,39 @@ def sales_forms
...
     deprecated_questions = {}
     all_questions_from_previous_forms.each_cons(2) do |preceding_question, question|
       next if current_form_question_ids.include?(question.id) || deprecated_questions.values.map(&:id).include?(question.id)
 
-      deprecated_questions[preceding_question.id] = question
+      if question.subsection.id == preceding_question.subsection.id
+        deprecated_questions[preceding_question.id] = question
+      else
+        last_in_preceding_subsection = current_form_questions.rindex { |q| q.subsection.id == preceding_question.subsection.id }
+        deprecated_questions[current_form_questions[last_in_preceding_subsection].id] = question
```
RachaelBooth, line 84: This means that a deprecated field at the start of a subsection doesn't show up before a new field added to the end of the previous section (otherwise this starts happening with postcode and the moved declaration). 

This is sufficient for the moment, there are probably other cases that could crop up in future it doesn't try and deal with.

app/services/bulk_upload/sales/year2024/row_parser.rb
```diff
@@ -556,13 +555,6 @@ def add_duplicate_found_in_spreadsheet_errors
   end
 
 private
-
-  def validate_data_protection_answered
-    unless field_18 == 1
-      errors.add(:field_18, I18n.t("validations.not_answered", question: QUESTIONS[:field_18].downcase), category: :setup)
-    end
-  end
-
```
RachaelBooth, line 565: This error was showing up twice with this left in.

kosiakkatrina: Could `validate_declaration_acceptance` also be removed from lettings?

kosiakkatrina (approved):

app/services/bulk_upload/sales/year2024/row_parser.rb
RachaelBooth: Yes - the error was only showing up once for lettings uploads (it's deduplicating somewhere), but it turns out the `validate_declaration_acceptance` check is still unnecessary now.

PR #2230, opened 2024-02-08
CLDC-3197 Export correct fields based on year

kosiakkatrina:


app/services/exports/lettings_log_export_service.rb
```diff
@@ -258,12 +259,14 @@ def filter_keys!(attributes)
       attributes.reject! { |attribute| is_omitted_field?(attribute) }
     end
 
-    def is_omitted_field?(field_name)
+    def is_omitted_field?(field_name, lettings_log = nil)
```
RachaelBooth: Does this need the default?

kosiakkatrina: Probably not, removed it now

RachaelBooth (approved):

PR #2231, opened 2024-02-08
CLDC-3168: Add Kent Reliance to mortgage lender options

RachaelBooth:


app/models/form/sales/questions/mortgage_lender.rb
```diff
@@ -54,54 +53,24 @@ def initialize(id, hsh, subsection, ownershipsch:)
     "37" =>	"Virgin Money",
     "38" =>	"West Bromwich Building Society",
     "39" =>	"Yorkshire Building Society",
+    "41" => "Kent Reliance",
```
RachaelBooth, line 56: I don't think the order actually matters for this question, but it seems more logical here to put this following the other lenders rather than after "Other".

kosiakkatrina (approved):

PR #2232, opened 2024-02-09
CLDC-3216 Validate BU headers

kosiakkatrina:


app/services/bulk_upload/lettings/validator.rb
```diff
@@ -175,6 +176,15 @@ def validate_correct_template
     errors.add(:base, :wrong_template) if csv_parser.wrong_template_for_year?
   end
 
+  def validate_missing_required_headers
+    return if halt_validations?
+
+    if csv_parser.missing_required_headers?
+      errors.add :base, I18n.t("activemodel.errors.models.bulk_upload/lettings/validator.attributes.base.no_headers", guidance_link: "https://#{ENV['APP_HOST']}/lettings-logs/bulk-upload-logs/guidance?form%5Byear%5D=#{bulk_upload.year}#using-the-bulk-upload-template")
```
RachaelBooth: We may well not care, but this will be the wrong link for review apps because it doesn't include the relative url root. There might be a rails helper that would work?

kosiakkatrina: Yeah, I'm not sure how to get the path through helper here. Any ideas?

RachaelBooth: Hmmm, might be able to just include the url_helpers from Rails.application.routes? If that doesn't work then I'm not sure.

kosiakkatrina: Right! So if I'm getting it right, `bulk_upload_lettings_log_path(id: "guidance", form: { year: bulk_upload.year })` only gives us a relative path and `bulk_upload_lettings_log_url(id: "guidance", form: { year: bulk_upload.year })` requires a `host` parameter which I can only think of getting from the ENV. I've updated it cause it looks nicer either way, but if you can think of a way to update it further let me know!

RachaelBooth: We could look at the `RAILS_RELATIVE_URL_ROOT` in the ENV and use that to build out the host? But it's much more annoying to do. How is it working for bulk upload links (or is it just not)?

I'm happy to leave it as tech debt if we need to.

kosiakkatrina: I think we set the default host in config `config.action_mailer.default_url_options = { host: ENV["APP_HOST"] }` but it doesn't work in this case, cause the link isn't constructed in the mailer

RachaelBooth: Ah, that's irritating. Ah well, lets go with this for the moment.

RachaelBooth (approved):

PR #2236, opened 2024-02-12
CLDC-3227: Add validations to reasonother content

RachaelBooth:


spec/models/validations/household_validations_spec.rb
```diff
@@ -651,24 +695,30 @@
...
+        { code: 13, label: "Children’s home or foster care" },
+        { code: 14, label: "Bed and breakfast" },
+        { code: 19, label: "Rough sleeping" },
+        { code: 23, label: "Mobile home or caravan" },
+        { code: 24, label: "Home Office Asylum Support" },
+        { code: 25, label: "Any other accommodation" },
+        { code: 26, label: "Owner occupation (private)" },
+        { code: 28, label: "Living with friends or family" },
+        { code: 29, label: "Prison or approved probation hostel" },
+      ].each do |prevten|
```
RachaelBooth, line 710: Thought since I was needing to update this (having stopped using the fake 2021 form) I may as well make it test what it says it's testing.

RachaelBooth:
CSV export tests need updating - but should wait for the open PR that touches that to go in before doing so to avoid merge conflicts.

app/models/validations/household_validations.rb
```diff
@@ -20,6 +30,12 @@ def validate_reason_for_leaving_last_settled_home(record)
       record.errors.add :referral, I18n.t("validations.household.referral.reason_permanently_decanted")
       record.errors.add :reason, I18n.t("validations.household.reason.not_internal_transfer")
     end
+
+    return unless record.form.start_year_after_2024?
+
+    if record.reason == 20 && Regexp.union(PHRASES_INDICATING_HOMELESSNESS.map { |phrase| Regexp.new("\\A[^[:alpha:]]*#{phrase}[^[:alpha:]]*\\Z", Regexp::IGNORECASE) }).match?(record.reasonother)
```
kosiakkatrina: Would this do the same thing? `PHRASES_INDICATING_HOMELESSNESS.any? { |phrase| Regexp.new("\\A[^[:alpha:]]*#{phrase}[^[:alpha:]]*\\Z", Regexp::IGNORECASE).match?(record.reason)}`

Any way we could do it without regex? Idm if not

db/schema.rb
```diff
@@ -709,29 +710,29 @@
...
     t.string "direct_otp"
-    t.datetime "direct_otp_sent_at"
+    t.datetime "direct_otp_sent_at", precision: nil
     t.datetime "totp_timestamp", precision: nil
     t.boolean "active", default: true
     t.string "confirmation_token"
-    t.datetime "confirmed_at"
-    t.datetime "confirmation_sent_at"
+    t.datetime "confirmed_at", precision: nil
+    t.datetime "confirmation_sent_at", precision: nil
```
kosiakkatrina: Should these be here?

app/models/validations/household_validations.rb
RachaelBooth: Don't think we can avoid the regex. I've moved the union so that we're only doing it once, which is definitely the less-work-to-run option and hopefully reads a bit more nicely - lmk what you think.

db/schema.rb
RachaelBooth: Hmm, probably not - must be something weird with my system since it's autogenerated.

kosiakkatrina (approved):

db/schema.rb
kosiakkatrina: Yeah, it happens sometimes, I think maybe something to do with Postgres version

PR #2238, opened 2024-02-13
CLDC-3238 Update non staircasing mortgage validation

kosiakkatrina:


app/models/validations/sales/sale_information_validations.rb
```diff
@@ -100,13 +100,23 @@ def validate_discount_and_value(record)
...
-        record.errors.add field, I18n.t("validations.sale_information.non_staircasing_mortgage", mortgage_and_deposit_total: record.field_formatted_as_currency("mortgage_and_deposit_total"), expected_shared_ownership_deposit_value: record.field_formatted_as_currency("expected_shared_ownership_deposit_value"))
+    if record.mortgage_used?
+      return unless record.mortgage
+
+      if record.mortgage_and_deposit_total != record.expected_shared_ownership_deposit_value
+        %i[mortgage value deposit equity].each do |field|
+          record.errors.add field, I18n.t("validations.sale_information.non_staircasing_mortgage.mortgage_used", mortgage_and_deposit_total: record.field_formatted_as_currency("mortgage_and_deposit_total"), expected_shared_ownership_deposit_value: record.field_formatted_as_currency("expected_shared_ownership_deposit_value"))
+        end
+      end
+    elsif record.mortgage_not_used?
```
RachaelBooth, line 115: Is this intended to apply to all years?

kosiakkatrina: I don't think so, we already had this condition `return unless record.saledate && record.form.start_year_after_2024?` before this change

RachaelBooth: Ah, I'd missed that 🤦‍♀️

RachaelBooth (approved):

PR #2242, opened 2024-02-15
CLDC-3249 Clear invalidated earnings values

kosiakkatrina:


spec/models/validations/financial_validations_spec.rb
```diff
@@ -293,6 +300,20 @@
           expect(record.errors["ecstat#{n}"]).to be_empty
         end
       end
+
+      context "when the net income is lower than the hard min for their employment status" do
+        it "adds an error" do
+          record.startdate = Time.zone.local(2022, 5, 1)
```
natdeanlewissoftwire, line 306: I think it's worth mentioning in the context/it description this is a 22/23 log given the rest in the file are 23/24

spec/lib/tasks/clear_invalidated_earnings_spec.rb
```diff
@@ -0,0 +1,48 @@
...
+
+    before do
+      Rake.application.rake_require("tasks/clear_invalidated_earnings")
+      Rake::Task.define_task(:environment)
+      task.reenable
+      FormHandler.instance.use_real_forms!
+    end
+
+    context "when the rake task is run" do
+      context "and there are 2023 logs with invalid earnings" do
```
natdeanlewissoftwire, line 16: it definitely will pass, but I wouldn't hate there being a test ensuring 2022 logs/valid 2023 logs remain unchanged and valid.

natdeanlewissoftwire (approved):

spec/models/validations/financial_validations_spec.rb
kosiakkatrina: Ah, yes, forgot to rename it, thanks!

spec/lib/tasks/clear_invalidated_earnings_spec.rb
kosiakkatrina: Yep! Will add that!

PR #2243, opened 2024-02-15
CLDC-3248: Ensure previous_la_known gets cleared if prevloc is when blanking invalid fields

RachaelBooth:


spec/models/lettings_log_spec.rb
```diff
@@ -3382,26 +3382,6 @@ def previous_postcode_fields
     end
   end
 
-  describe "#blank_invalid_non_setup_fields!" do
```
RachaelBooth, line 3385: Moved these to `log_spec` because the function is on `log`

spec/models/log_spec.rb
```diff
@@ -27,4 +27,54 @@
...
+      subject(:model) { build(:sales_log, :completed, age1: 10) }
+
+      it "blanks it" do
+        model.valid?
+        model.blank_invalid_non_setup_fields!
+        expect(model.age1).to be_nil
+      end
+    end
+
+    context "when prevloc is invalid for a lettings log" do
```
RachaelBooth, line 70: I can't see a way to make prevloc invalid on a sales log at the moment, so only have this test.

kosiakkatrina (approved):

PR #2249, opened 2024-02-16
CLDC-3218 Fix rent frequency soft validation bulk upload bug

natdeanlewissoftwire:
The soft validations on the "other charges" fields were being applied even for values outside the hard maxima/minima, which meant in bulk uploads we were displaying these soft validation errors instead of the correct hard validation ones.

TODO:
Make the same fix to other similar soft value checks that should be testing a range rather than a single inequality - e.g.  `income1_over_soft_max?` should probably be `income1_in_soft_max_range?` _not sure there is a hard max on income1 - kat_


kosiakkatrina:
This seems to all have been covered by CLDC-3207 already

kosiakkatrina:
> This seems to all have been covered by [CLDC-3207](https://digital.dclg.gov.uk/jira/browse/CLDC-3207) already

Wrong PR! Reopened!

[CLDC-3207]: https://dluhcdigital.atlassian.net/browse/CLDC-3207?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

spec/models/validations/soft_validations_spec.rb
```diff
@@ -409,13 +409,25 @@
...
+       {
+         period: { label: "monthly", value: 4 },
+         scharge: 3471,
+       },
+       {
+         period: { label: "every 2 weeks", value: 2 },
+         scharge: 1601,
        }].each do |test_case|
-        it "returns false if scharge is under soft max for general needs #{test_case[:period][:label]}(35)" do
+        it "returns false if scharge is under soft max (35) or over hard max (800) for general needs #{test_case[:period][:label]}" do
```
RachaelBooth: Can these put whether it's a below soft max or over hard max case into the test name instead of the or?

kosiakkatrina: Sure!

RachaelBooth (approved):

spec/factories/lettings_log.rb
```diff
@@ -24,13 +24,6 @@
       rent_type { 1 }
       startdate { Time.zone.today }
     end
-    trait :soft_validations_triggered do
```
natdeanlewissoftwire, line 27: this wasn't being used anywhere

app/models/bulk_upload.rb
```diff
@@ -78,37 +78,18 @@ def unpend
     end
   end
 
+  SHARED_VALUE_CHECKS = %w[retirement_value_check].freeze
```
natdeanlewissoftwire: this refactoring is a) nicer and b) allows testing that the value checks are all being confirmed in bulk_upload_spec (previously some of the value_check fields were missing from unpend_and_confirm_soft_validations)

kosiakkatrina: Hmm, I wonder if getting it from `log.form.questions.filter_the_questions_by_interruption_screen.map(&:id).uniq` would help us avoid missing some here again?

natdeanlewissoftwire: yeah okay, I'll move the explicit checking to the test so we've still got it somewhere, nice point

PR #2251, opened 2024-02-19
CLDC-3224: Fix bulk upload noint bug

RachaelBooth:


lib/tasks/correct_noint_value.rake
```diff
@@ -0,0 +1,24 @@
...
+  affected_uploads = BulkUpload.where(log_type: "sales", noint_fix_status: BulkUpload.noint_fix_statuses[:not_applied])
+  affected_uploads.each do |upload|
+    upload.logs.where(noint: 2).each do |log|
+      noint_at_upload = log.versions.length == 1 ? log.noint : log.versions.first.next.reify.noint
+      next unless noint_at_upload == 2
+
+      Rails.logger.info("Updating noint value on log #{log.id}, owning org #{log.owning_organisation_id}")
+      update_counts[log.status.to_sym] += 1
+      log.noint = 1
+      log.skip_update_status = true
```
RachaelBooth, line 18: The thing we want to avoid here is recalculating the status for anything that is pending. For anything that isn't pending, updating noint should have no impact on the status - it already had a value set, and no routing depends on which value it takes.

app/services/bulk_upload/sales/log_creator.rb
```diff
@@ -25,6 +25,7 @@ def call
         Sentry.capture_exception(e)
       end
     end
+    @bulk_upload.update!(noint_fix_status: BulkUpload.noint_fix_statuses[:not_needed])
```
kosiakkatrina: Wouldn't `@bulk_upload.update!(noint_fix_status: "not_needed")` work/ mean the same thing?

kosiakkatrina: Is this the right place to add this? Would it make sense upon bulk_upload creation?

db/migrate/20240216163519_add_no_int_fix_marker_to_bulk_uploads.rb
```diff
@@ -0,0 +1,5 @@
+class AddNoIntFixMarkerToBulkUploads < ActiveRecord::Migration[7.0]
+  def change
+    add_column :bulk_uploads, :noint_fix_status, :string, default: "not_applied"
```
kosiakkatrina, line 3: ~Would the default get added to existing records? If no, wouldn't think make more sense as `not_needed`?~ hmm, sounds like it would be on existing records, never mind

app/services/bulk_upload/sales/log_creator.rb
RachaelBooth: You can't do it on bulk upload creation because that happens from an app instance, so something could be created with updated app code but still processed with the old parser by sidekiq (if it was uploaded at exactly the wrong moment).

Although actually I think maybe it needs to happen at the start of this function rather than the end, in case of running the task at exactly the wrong moment.

RachaelBooth: > Wouldn't `@bulk_upload.update!(noint_fix_status: "not_needed")` work/ mean the same thing?

Yes - I was generally avoiding typing the string because I inevitably end up making typos, and referencing the enum itself forces it to be right.

db/migrate/20240216163519_add_no_int_fix_marker_to_bulk_uploads.rb
RachaelBooth: To confirm, yes it does get added to existing records.

app/services/bulk_upload/sales/log_creator.rb
kosiakkatrina: That's fair in terms of timing
I think a typo would cause an error, so shouldn't be too bad to use here (given there are tests) - up to you tho!
image

kosiakkatrina (approved):

PR #2252, opened 2024-02-20
CLDC-3181 Update question numbers for 2024

kosiakkatrina:


app/models/form/sales/questions/address_line1.rb
```diff
@@ -8,7 +8,7 @@ def initialize(id, hsh, page)
     @plain_label = true
     @check_answer_label = "Address lines 1 and 2"
     @disable_clearing_if_not_routed_or_dynamic_answer_options = true
-    @question_number = 15
+    @question_number = QUESION_NUMBER_FROM_YEAR[form.start_date.year]
```
RachaelBooth: Do these need a default (for when it gets to 2025 forms)?

kosiakkatrina: I mean we could always default to the latest available year, but it would be easier to spot if the number wasn't updated if it's empty. I'm not sure, it kinda depends on how likely the numbers are to change. I've updated it now, let me know if you have any thoughts on pros/cons of having the defaults

app/models/form/sales/questions/address_line1.rb
```diff
@@ -8,7 +8,7 @@ def initialize(id, hsh, page)
     @plain_label = true
     @check_answer_label = "Address lines 1 and 2"
     @disable_clearing_if_not_routed_or_dynamic_answer_options = true
-    @question_number = 15
+    @question_number = QUESION_NUMBER_FROM_YEAR.fetch(form.start_date.year, QUESION_NUMBER_FROM_YEAR.max_by { |k, _v| k }.last)
```
RachaelBooth: I've just noticed the typo here in `QUESTION` - other than that looks good to me.

kosiakkatrina: Ah good spot!

natdeanlewissoftwire (approved):
looks good, all my comments in https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/2255#pullrequestreview-1901143056 apply here too so if you decide to accept any of the suggestions there then don't forget to add them here as well :)

PR #2255, opened 2024-02-21
CLDC-3180 Update lettings question numbering

kosiakkatrina:


app/models/form/lettings/questions/irproduct_other.rb
```diff
@@ -5,6 +5,8 @@ def initialize(id, hsh, page)
     @check_answer_label = "Product name"
     @header = "Name of rent product"
     @type = "text"
-    @question_number = 6
+    @question_number = QUESION_NUMBER_FROM_YEAR.fetch(form.start_date.year, QUESION_NUMBER_FROM_YEAR.max_by { |k, _v| k }.last) if form.start_date.present?
```
kosiakkatrina: `if form.start_date.present?` is on setup section only, I think because some of the forms are still coming from JSON, so the form isn't set here (?) doesn't change things in terms of numbering

app/models/form/lettings/questions/address_line1.rb
```diff
@@ -8,7 +8,7 @@ def initialize(id, hsh, page)
     @plain_label = true
     @check_answer_label = "Address lines 1 and 2"
     @disable_clearing_if_not_routed_or_dynamic_answer_options = true
-    @question_number = 12
+    @question_number = QUESION_NUMBER_FROM_YEAR.fetch(form.start_date.year, QUESION_NUMBER_FROM_YEAR.max_by { |k, _v| k }.last)
```
natdeanlewissoftwire: QUESION typo here like in the other PR :)

app/models/form/lettings/questions/address_line1.rb
```diff
@@ -8,7 +8,7 @@ def initialize(id, hsh, page)
     @plain_label = true
     @check_answer_label = "Address lines 1 and 2"
     @disable_clearing_if_not_routed_or_dynamic_answer_options = true
-    @question_number = 12
+    @question_number = QUESTION_NUMBER_FROM_YEAR.fetch(form.start_date.year, QUESTION_NUMBER_FROM_YEAR.max_by { |k, _v| k }.last)
```
natdeanlewissoftwire: I've been trying to come up with a clearer way of getting the "default" value here - I wonder if `QUESTION_NUMBER_FROM_YEAR[QUESTION_NUMBER_FROM_YEAR.keys.max]` is a more intuitive way of getting the same thing?

natdeanlewissoftwire: `.last` seems like a confusing way to get the value of a hash pair is all

app/models/form/lettings/questions/age.rb
```diff
@@ -12,6 +12,14 @@ def initialize(id, hsh, page, person_index:, person_type:)
...
-    @question_number = 31 + (4 * person_index)
+    @person_index = person_index
+  end
+
+  def question_number
+    if form.start_date.year == 2023
+      31 + (4 * @person_index)
+    else
+      30 + (4 * @person_index)
+    end
```
natdeanlewissoftwire:
```suggestion
    base_question_number = case form.start_date.year
    when 2023
      31
    else
      30
    end
    base_question_number + (4 * @person_index)
```

natdeanlewissoftwire (approved):
looks good - did you consider doing this more programmatically at all? i.e. not hard coding question numbers but inferring an order from the order of questions in the subsection pages? We decided against it last year but we have time this time around if we wanted to make it more future-proof - no problem with it as is though, just wondering if you had looked down that route at all :)

kosiakkatrina:
> looks good - did you consider doing this more programmatically at all? i.e. not hard coding question numbers but inferring an order from the order of questions in the subsection pages? We decided against it last year but we have time this time around if we wanted to make it more future-proof - no problem with it as is though, just wondering if you had looked down that route at all :)

Yeah, I kinda considered it, I had some thoughts about this in a slack thread:
``Yeah, [the questions] kinda are in order, but there are exceptions, for example all the address questions are separate questions, but have the same number, but postcode (although it’s on the same page as the address) has it’s own number.
And some questions we don’t always number, for example owning organisation on sales doesn’t have a number this year, because we added that question mid year (so it wasn’t on the paper form and didn’t have the number), if we mapped through it, it would have shifted all the other questions mid year.
There are also different versions of the same question that depend on answers to other questions. We could only increment the number if that specific page with the question gets routed to, but without adding more info on the question itself there’s isn’t really a way to tell if the question isn’t routed to because it’s a different version of a question (and therefore shouldn’t be incremented) or if the question is not routed to because we can infer it for example (and therefore it should have it’s own question number)
It’s kinda painful, but being explicit might be less error prone as iterating might mean adding more logic around routing etc
``

PR #2256, opened 2024-02-22
CLDC-3146 Update household validations

kosiakkatrina:
Update household validations for 2024
Infer sales child relationships where relevant


app/models/validations/household_validations.rb
```diff
@@ -51,17 +51,112 @@ def validate_armed_forces(record)
...
+      next unless age && economic_status && relationship
+
+      age_between_16_19 = age.between?(16, 19)
+      student = tenant_is_fulltime_student?(economic_status)
+      economic_status_refused = tenant_economic_status_refused?(economic_status)
+      child = tenant_is_child?(relationship)
+
+      if age_between_16_19 && !(student || economic_status_refused) && child
+        if record.form.start_year_after_2024?
+          record.errors.add "ecstat#{person_num}", I18n.t("validations.household.ecstat.student_16_19.must_be_student_2024", person_num:, person: "lead tenant")
```
RachaelBooth: What happens here if you e.g answer relationship last? (and similar for the other cases below) It looks like you can get into a situation where there's a validation error but you can't see it.

kosiakkatrina: Ah, I thought there was a different validation catching it, but very good point! Will check!

kosiakkatrina: Added the errors back to all fields now!

app/models/validations/household_validations.rb
```diff
@@ -51,17 +51,114 @@ def validate_armed_forces(record)
...
+        if record.form.start_year_after_2024?
+          record.errors.add "age#{person_num}", I18n.t("validations.household.age.student_16_19.must_be_16_19_2024", person_num:, person: "lead tenant")
+        else
+          record.errors.add "age#{person_num}", I18n.t("validations.household.age.student_16_19.must_be_16_19")
+        end
+        record.errors.add "ecstat#{person_num}", I18n.t("validations.household.ecstat.student_16_19.cannot_be_student.child_not_16_19")
+        record.errors.add "relat#{person_num}", I18n.t("validations.household.relat.student_16_19.cannot_be_child.student_not_16_19")
+      end
+
+      next unless student && age_between_16_19 && !child && record.form.start_year_after_2024?
```
RachaelBooth: I know it's in the ticket, but is this actually a sensible validation? For the same reasons we allow "other" as a relationship for under 16s - could be a child of a non-lead tenant but not the lead tenant (or presumably in this age range you could e.g. move somewhere with a partner while being a student).

Not objecting to it as is necessarily (think it's up to the stats team) but seems worth querying.

spec/models/validations/household_validations_spec.rb
```diff
@@ -307,33 +309,87 @@
...
+        expect(record.errors["age2"])
+          .to include("Age must be 19 or under as you told us person 2's relationship to the lead tenant is child")
+        expect(record.errors["relat2"])
+          .not_to include("Answer cannot be child, as you told us person 2 is over 19")
       end
+    end
+  end
 
+  describe "#validate_person_age_matches_economic_status" do
+    context "when the household contains a person under 16" do
```
RachaelBooth: Does this need an explicit 2023 context?

RachaelBooth (approved):
There's a query in there but it's more a question to raise with the stats team.

app/models/validations/household_validations.rb
kosiakkatrina: Will check!

spec/models/validations/household_validations_spec.rb
kosiakkatrina: Yup, I think so! Thanks!

app/models/form/lettings/pages/multiple_partners_value_check.rb
```diff
@@ -0,0 +1,26 @@
...
+  end
+
+  def questions
+    @questions ||= [
+      Form::Lettings::Questions::MultiplePartnersValueCheck.new(nil, nil, self, person_index: @person_index),
+    ]
+  end
+
+  def interruption_screen_question_ids
+    %w[relat2 relat3 relat4 relat5 relat6 relat7 relat8]
```
natdeanlewissoftwire, line 24: I think this will add the soft validation error in bulk upload to all these fields, are we fine with that or do we want to add it to relevant fields only?

app/models/form/sales/questions/buyer2_working_situation.rb
```diff
@@ -27,7 +27,6 @@ def initialize(id, hsh, page)
     "0" => { "value" => "Other" },
     "10" => { "value" => "Buyer prefers not to say" },
     "7" => { "value" => "Full-time student" },
-    "9" => { "value" => "Child under 16" },
```
natdeanlewissoftwire, line 30: was this intentional?

app/models/validations/soft_validations.rb
```diff
@@ -188,6 +191,12 @@ def reasonother_might_be_existing_category?
     PHRASES_LIKELY_TO_INDICATE_EXISTING_REASON_CATEGORY_REGEX.match?(reasonother)
   end
 
+  def multiple_partners?
+    return unless hhmemb
+
+    (2..hhmemb).count { |n| public_send("relat#{n}") == "P" } > 1
```
natdeanlewissoftwire: suggest https://apidock.com/rails/Enumerable/many%3F here, up to you :)

natdeanlewissoftwire (approved):

app/models/form/sales/questions/buyer2_working_situation.rb
kosiakkatrina: Yup! We don't allow this combination anyway, cause buyer 2 age can't be under 16 and it would trigger a validation if this option is selected in conjunction with age over 16. This also makes it behave the same way as buyer 1 question

app/models/form/lettings/pages/multiple_partners_value_check.rb
kosiakkatrina: Hmm, I think we're checking if the page is routed to before adding the error. In row parser:
   ```
        next if log.form.questions.none? { |q| q.id == interruption_screen_question_id && q.page.routed_to?(log, nil) }
```

app/models/validations/soft_validations.rb
kosiakkatrina: Ah, of course, tyty!

PR #2257, opened 2024-02-23
CLDC-3253 Validate nationality in bulk upload

kosiakkatrina:


app/services/bulk_upload/lettings/year2024/row_parser.rb
```diff
@@ -1472,6 +1480,10 @@ def first_time_property_let_as_social_housing
     end
   end
 
+  def valid_nationality_options
+    %w[0 826] + GlobalConstants::COUNTRIES_ANSWER_OPTIONS.keys
```
natdeanlewissoftwire: can't believe I'm saying this but it might be worth adding a comment to say this is "prefers not to say" and "UK" (I'm not certain why we're specifying 826 here though, isn't it already in COUNTRIES_ANSWER_OPTIONS?)

natdeanlewissoftwire (approved):
nice :)

app/services/bulk_upload/lettings/year2024/row_parser.rb
kosiakkatrina: Good point! And will add the comment, I think I would have found it useful if I was reading it

PR #2258, opened 2024-02-23
CLDC-3246 Update homepage in progress links

kosiakkatrina:


app/helpers/home_helper.rb
```diff
@@ -1,15 +1,17 @@
 module HomeHelper
   def data_count(user, type)
+    years = FormHandler.instance.lettings_in_crossover_period? ? [current_collection_start_year, previous_collection_start_year] : [current_collection_start_year]
```
RachaelBooth: Do we ever have different periods for lettings and sales forms? It seems odd to be specifying lettings here but using it for both.

kosiakkatrina: I don't think we do, but we could so I've updated it

RachaelBooth (approved):

PR #2262, opened 2024-02-23
CLDC-2049: Use year instead of years when only one is displayed

RobertSullivanSoftwire:


app/models/form/sales/questions/living_before_purchase_years.rb
```diff
@@ -39,4 +38,8 @@ def question_number
       99
     end
   end
+
+  def suffix_label(log)
+    log[id] == 1 ? " year" : " years"
```
natdeanlewissoftwire: could do `" #{'year'.pluralize(log[id])}"`(or similar, you get the idea) to DRY this up a little
https://apidock.com/rails/ActionView/Helpers/TextHelper/pluralize

natdeanlewissoftwire (approved):

app/models/form/sales/questions/living_before_purchase_years.rb
RobertSullivanSoftwire: Nice! Speaking of repetition, I've gone ahead and made the same change for the mortgages question which also uses this format, though that seems much less likely to come up.

natdeanlewissoftwire (approved):
nice one!

PR #2263, opened 2024-02-23
CLDC-2171: Fix typo in homelessness question

RobertSullivanSoftwire:


natdeanlewissoftwire (approved):

natdeanlewissoftwire:
We could add a test for this question (and its page) - as you'll see we don't have spec files for all the questions, they've been inconsistently added thus far but we're trying to work on the process of updating/adding tests where they're missing for things we touch :)

RobertSullivanSoftwire:
> We could add a test for this question (and its page) - as you'll see we don't have spec files for all the questions, they've been inconsistently added thus far but we're trying to work on the process of updating/adding tests where they're missing for things we touch :)

I have to admit, I went looking for a test for this file, then thought to myself "ah - there's no test for this because there's no non-trivial behaviour here to test".

Given that I went through that thought process, I think it's worth asking the question - what kind of regressions would we expect a test for this file to catch?

spec/models/form/lettings/questions/homeless_spec.rb
```diff
@@ -0,0 +1,46 @@
+require "rails_helper"
```
RobertSullivanSoftwire, line 1: Right, I've added a test file here - though I am still a little concerned that all I've achieved is to make this question more difficult to change!

natdeanlewissoftwire: Yeah - I also have my doubts as to whether the way we're testing questions and pages has any real value other than avoiding accidental changes, but at least we're trying to stay consistent with the decision previously made. When we come to looking at speeding up the pipeline I think one approach might be deleting/minimising these sorts of tests, but it seems more sensible to do this unilaterally rather than changing our approach ad hoc.

natdeanlewissoftwire (approved):
🥳

spec/models/form/lettings/questions/homeless_spec.rb
natdeanlewissoftwire: Basically I agree with you that there's not much value added with our current approach to testing questions and pages, I think it's worth visiting more generally - maybe one for team time?

RobertSullivanSoftwire: Yeah, I don't disagree with any of that!

PR #2265, opened 2024-02-23
CLDC-1971: Update text for 'prefers not to say' answer to buyer 1 ethnic group question

RobertSullivanSoftwire:


app/models/form/sales/questions/buyer2_ethnic_group.rb
```diff
@@ -23,6 +23,6 @@ def initialize(id, hsh, page)
     "3" => { "value" => "Black, African, Caribbean or Black British" },
     "4" => { "value" => "Arab or other ethnic group" },
     "divider" => { "value" => true },
-    "17" => { "value" => "Buyer 2 prefers not to say" },
+    "17" => { "value" => "Buyer prefers not to say" },
```
kosiakkatrina, line 20: Would it make sense to remove `inferred_check_answers_value` here as well?

spec/models/form/question_spec.rb
```diff
@@ -357,8 +357,8 @@
       end
 
       context "when Sales form" do
-        let(:sales_log) { FactoryBot.create(:sales_log, :completed, ethnic_group: 17, saledate: Time.zone.local(2022, 1, 1)) }
-        let(:question) { sales_log.form.get_question("ethnic_group", sales_log) }
+        let(:sales_log) { FactoryBot.create(:sales_log, :completed, national: 13, saledate: Time.zone.local(2022, 1, 1)) }
+        let(:question) { sales_log.form.get_question("national", sales_log) }
```
kosiakkatrina, line 361: Why do we need this change?

RobertSullivanSoftwire: This is a test about inferred check answers values which the ethnic_group question no longer has - I just picked another question that still has one.

app/models/form/sales/questions/buyer2_ethnic_group.rb
RobertSullivanSoftwire: I absolutely intended to - should be fixed now!

kosiakkatrina (approved):

PR #2268, opened 2024-02-26
CLDC-3245: Make Q59 optional

RobertSullivanSoftwire:


app/controllers/form_controller.rb
```diff
@@ -102,6 +103,12 @@ def responses_for_page(page)
         result[question.id] = question_params
       end
 
+      if %w[checkbox].include?(question.type) &&
```
RobertSullivanSoftwire, line 106: Some context for why this is here - there's precedent back in the old system for a distinct value (`pregblank`) to represent empty answers to this question.

spec/services/bulk_upload/sales/year2023/row_parser_spec.rb
```diff
@@ -959,14 +959,24 @@ def purch_id_field
...
+
+      context "when all empty" do
+        let(:attributes) { empty_organisation_params }
+
+        it "sets pregblank field" do
+          expect(parser.log.pregblank).to be(1)
+        end
+      end
+
+      %i[field_67 field_68 field_69 field_70].each do |field_number|
```
RobertSullivanSoftwire, line 972: These tests feel a little clunky - very open to suggestions on how to make it cleaner.

app/controllers/form_controller.rb
RobertSullivanSoftwire: Currently looking into whether we can replace this with the divider system used elsewhere.

PR #2271, opened 2024-02-28
CLDC-3245: Make Q59 optional

RobertSullivanSoftwire:


app/controllers/form_controller.rb
```diff
@@ -102,6 +102,12 @@ def responses_for_page(page)
         result[question.id] = question_params
       end
 
+      if %w[checkbox].include?(question.type) &&
```
RobertSullivanSoftwire: This requires a bit of context. We can't simply leave the answer as null, since there's a field (`pregblank`) used for pre-existing "none of the above" answers to this question.

However, it's also too late to change the 24/25 form to add an exclusive "None of the above" checkbox. Instead, we need to make it so not filling out any checkboxes sets the `pregblank` field, hence the `default_answer` stuff here.

I do wonder whether it would be better to get rid of the abstraction here and just make it something we do for this specific question, as in bulk uploads.

kosiakkatrina: We set a lot of the derived values here `module DerivedVariables::LettingsLogVariables` would it be possible/make sense to do that as part of `set_derived_fields!` as well?

RobertSullivanSoftwire: Thanks for the suggestion - done!

app/models/derived_variables/sales_log_variables.rb
```diff
@@ -4,6 +4,7 @@ module DerivedVariables::SalesLogVariables
   def set_derived_fields!
     reset_invalidated_derived_values!(DEPENDENCIES)
 
+    self.pregblank = 1 if self.no_buyer_organisation?
```
kosiakkatrina: Hm, do we want this to apply to 2023 only?

app/services/bulk_upload/sales/year2024/row_parser.rb
```diff
@@ -862,6 +852,7 @@ def attributes_for_log
     attributes["pregla"] = field_68
     attributes["pregghb"] = field_69
     attributes["pregother"] = field_67
+    attributes["pregblank"] = no_buyer_organisation
```
kosiakkatrina, line 855: does this exist as an option on the BU template/ does it have it's own field? (2024 only)

kosiakkatrina (approved):

app/services/bulk_upload/sales/year2024/row_parser.rb
RobertSullivanSoftwire: Similarly, my understanding is that this isn't being introduced until 2025, per comment on https://dluhcdigital.atlassian.net/browse/CLDC-3245

app/models/derived_variables/sales_log_variables.rb
RobertSullivanSoftwire: My understanding is that "None of the above" isn't being introduced until 2025, based on comments on https://dluhcdigital.atlassian.net/browse/CLDC-3245.

PR #2272, opened 2024-02-28
CLDC-2220 Update buyer pluralisation

kosiakkatrina:


app/models/form/sales/pages/housing_benefits.rb
```diff
@@ -14,7 +14,7 @@ def depends_on
     if @joint_purchase
       [{ "joint_purchase?" => true }]
     else
-      [{ "not_joint_purchase?" => true }]
+      [{ "not_joint_purchase?" => true }, { "jointpur" => nil }]
```
kosiakkatrina, line 17: Joint purchase is routed to conditionally, so might not always be answered, added this here just to be sure we have default question version. Any reason we shouldn't do that?

natdeanlewissoftwire: Hm, good point - still displaying this question for nil jointpurs makes sense to me - will this make any previously complete logs become in progress if they hadn't satisfied the condition for jointpur to be routed to?

app/models/form/sales/questions/privacy_notice.rb
```diff
@@ -1,19 +1,20 @@
 class Form::Sales::Questions::PrivacyNotice < ::Form::Question
-  def initialize(id, hsh, page)
-    super
+  def initialize(id, hsh, page, joint_purchase:)
+    super(id, hsh, page)
     @id = "privacynotice"
-    @check_answer_label = "Buyer has seen the privacy notice?"
+    @check_answer_label = "#{joint_purchase ? 'Buyers have' : 'Buyer has'} seen the privacy notice?"
```
natdeanlewissoftwire, line 5: non-technical question - how are we expecting users to answer if one but not all buyers (in a joint purchase) have seen the privacy notice?

app/models/form/sales/questions/savings.rb
```diff
@@ -1,8 +1,8 @@
 class Form::Sales::Questions::Savings < ::Form::Question
-  def initialize(id, hsh, page)
-    super
+  def initialize(id, hsh, page, joint_purchase:)
+    super(id, hsh, page)
     @id = "savings"
-    @check_answer_label = "Buyer’s total savings before any deposit paid"
+    @check_answer_label = "#{joint_purchase ? 'Buyers’' : 'Buyer’s'} total savings before any deposit paid"
```
natdeanlewissoftwire, line 5: there's some inconsistency of `’` vs `'` in this PR (though I accept there is throughout the codebase too) - maybe a tech debt ticket to sort this out later as it's such a non-issue?

natdeanlewissoftwire: (looks like we're using the first one in "buyer's" and the second one in "weren't" for example)

app/models/validations/sales/financial_validations.rb
```diff
@@ -48,7 +48,7 @@ def validate_percentage_bought_not_greater_than_percentage_owned(record)
     return unless record.stairbought && record.stairowned
 
     if record.stairbought > record.stairowned
-      record.errors.add :stairowned, I18n.t("validations.financial.staircasing.percentage_bought_must_be_greater_than_percentage_owned")
+      record.errors.add :stairowned, I18n.t("validations.financial.staircasing.percentage_bought_must_be_greater_than_percentage_owned", persons: record.joint_purchase? ? "buyers" : "buyer")
```
natdeanlewissoftwire: can we call `persons` something like `buyers_with_apostrophe`? I think `persons` doesn't clarify whether it's about "more than one person" or "person's"

app/models/validations/sales/financial_validations.rb
```diff
@@ -125,9 +125,9 @@ def validate_equity_less_than_staircase_difference(record)
 
     if record.equity > record.stairowned - record.stairbought
       formatted_equity = sprintf("%g", record.equity)
-      record.errors.add :equity, I18n.t("validations.financial.equity.over_stairowned_minus_stairbought", equity: formatted_equity, staircase_difference: record.stairowned - record.stairbought)
-      record.errors.add :stairowned, I18n.t("validations.financial.equity.over_stairowned_minus_stairbought", equity: formatted_equity, staircase_difference: record.stairowned - record.stairbought)
-      record.errors.add :stairbought, I18n.t("validations.financial.equity.over_stairowned_minus_stairbought", equity: formatted_equity, staircase_difference: record.stairowned - record.stairbought)
+      record.errors.add :equity, I18n.t("validations.financial.equity.over_stairowned_minus_stairbought", equity: formatted_equity, staircase_difference: record.stairowned - record.stairbought, person_owns: record.joint_purchase? ? "buyers own" : "buyer owns")
```
natdeanlewissoftwire: similary can we call this buyers_own_with_apostrophe (or just buyers_own would be good)?

app/models/validations/sales/property_validations.rb
```diff
@@ -4,10 +4,10 @@ def validate_postcodes_match_if_discounted_ownership(record)
     return unless record.ppostcode_full.present? && record.postcode_full.present?
 
     if record.discounted_ownership_sale? && record.ppostcode_full != record.postcode_full
-      record.errors.add :postcode_full, I18n.t("validations.property.postcode.must_match_previous")
-      record.errors.add :ppostcode_full, I18n.t("validations.property.postcode.must_match_previous")
-      record.errors.add :ownershipsch, I18n.t("validations.property.postcode.must_match_previous")
-      record.errors.add :uprn, I18n.t("validations.property.postcode.must_match_previous")
+      record.errors.add :postcode_full, I18n.t("validations.property.postcode.must_match_previous", persons: record.joint_purchase? ? "Buyers‘" : "Buyer‘s")
```
natdeanlewissoftwire: same as above

natdeanlewissoftwire (approved):
nice!

app/models/form/sales/questions/privacy_notice.rb
kosiakkatrina: Good point, will double check

app/models/form/sales/questions/savings.rb
kosiakkatrina: Fixed it for weren't/wasn't, good spot! Don't think this PR updated buyers ones!

PR #2273, opened 2024-02-28
CLDC-3235 Update how we display create new logs actions

kosiakkatrina:
There was previously a situation where organisation doesn't own their stock but have multiple stock owners and none of those stock owners have their DSAs signed. In this case in the owning organisation question we should already be showing the errors, that the selected organisation hasn’t signed the DSA.
 
However, there is an edge case where organisation doesn’t own stock, organisation has 1 stock owner and that stock owner doesn’t have their DSA signed. In this situation we’re not showing the owning organisation question and default to the stock owner as the owning organisation and that fails (internally, without any useful message for the user) and doesn’t set the owning organisation.

This PR should prevent that by hiding the `create a new log` actions
image

### DIFF
--- a/app/components/data_protection_confirmation_banner_component.rb
+++ b/app/components/data_protection_confirmation_banner_component.rb
@@ -13,13 +13,16 @@ class DataProtectionConfirmationBannerComponent < ViewComponent::Base
   def display_banner?
     return false if user.support? && organisation.blank?
     return true if org_without_dpo?
+    return false if !org_or_user_org.holds_own_stock? && org_or_user_org.stock_owners.empty? && org_or_user_org.absorbed_organisations.empty?
 
-    !org_or_user_org.data_protection_confirmed?
+    !org_or_user_org.organisation_or_stock_owner_signed_dsa_and_holds_own_stock?
   end
 
   def header_text
     if org_without_dpo?
       "To create logs your organisation must state a data protection officer. They must sign the Data Sharing Agreement."
+    elsif !org_or_user_org.holds_own_stock?
+      "Your organisation does not own stock. To create logs your stock owner(s) must accept the Data Sharing Agreement on CORE."
     elsif user.is_dpo?
       "Your organisation must accept the Data Sharing Agreement before you can create any logs."
     else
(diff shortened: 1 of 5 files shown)


app/components/data_protection_confirmation_banner_component.rb
```diff
@@ -13,13 +13,16 @@ def initialize(user:, organisation: nil)
   def display_banner?
     return false if user.support? && organisation.blank?
     return true if org_without_dpo?
+    return false if !org_or_user_org.holds_own_stock? && org_or_user_org.stock_owners.empty? && org_or_user_org.absorbed_organisations.empty?
```
kosiakkatrina, line 16: Need this cause in this case we should be displaying missing stock owners component

app/components/data_protection_confirmation_banner_component.rb
```diff
@@ -13,13 +13,16 @@ def initialize(user:, organisation: nil)
...
 
-    !org_or_user_org.data_protection_confirmed?
+    !org_or_user_org.organisation_or_stock_owner_signed_dsa_and_holds_own_stock?
```
RachaelBooth: What should happen if there is a stock owner with dsa signed, but the org itself hasn't signed the dsa? I think this is set up so that that would be ok, but seems a bit weird to me.

kosiakkatrina: ~I don't think there should be a situation like that (?) User should only be able to sign DSA for their org~ oh sorry, I read stock owner as a user for some reason. Hmm, let me think

kosiakkatrina: Okay, updated this now to always check the current organisation's DSA as well

app/components/data_protection_confirmation_banner_component.rb
```diff
@@ -50,13 +53,21 @@ def data_protection_officers_text
...
     else
       "Read the Data Sharing Agreement"
     end
   end
 
   def link_href
-    dpo_required? ? GlobalConstants::HELPDESK_URL : data_sharing_agreement_organisation_path(org_or_user_org)
+    if dpo_required?
+      GlobalConstants::HELPDESK_URL
+    elsif !org_or_user_org.holds_own_stock?
```
RachaelBooth: Does this also need an update for if the users org also hasn't signed the agreement?

We be good to have a test for that case as well.

kosiakkatrina: Yup, makes sense, thanks! Updated again

RachaelBooth (approved):

PR #2274, opened 2024-02-28
CLDC-3187: Display org id on stock owner / managing agent pages

RachaelBooth:


app/views/organisation_relationships/_related_org_list.erb
RachaelBooth, line 1: The _managing_agent_list and _stock_owner_list partials were identical but for the remove path and item labels, and there's every reason to think they will stay that way, so I've combined them here.

RachaelBooth: If you want to see the diff between this and the previous version of each of those partials (before refactoring into one), github should be able to show the diff from only certain commits.

natdeanlewissoftwire: nice, love it!

app/views/organisation_relationships/managing_agents.html.erb
```diff
@@ -22,6 +22,15 @@
...
 <% if @total_count != 0 %>
   <%= render SearchComponent.new(current_user:, search_label: "Search for a managing agent", value: @searched) %>
-  <%= render partial: "organisation_relationships/managing_agent_list", locals: { index: @managing_agents, title: "Managing agents", pagy: @pagy, searched: @searched, item_label:, total_count: @total_count } %>
+  <%= render partial: "organisation_relationships/related_org_list", locals: {
+    related_orgs: @managing_agents,
+    title: "Managing agents",
+    pagy: @pagy,
+    searched: @searched,
+    item_label:,
+    search_item: "agents",
```
natdeanlewissoftwire: I know this wasn't your change, but shouldn't this be "managing agents"?

natdeanlewissoftwire (commented):
looks good, just one small thing :)

natdeanlewissoftwire (approved):

PR #2275, opened 2024-02-29
feat: update 24/25 lettings paper form

natdeanlewissoftwire:


kosiakkatrina (approved):

RachaelBooth (approved):
Is there a change here that we also need to reflect in a question in the service?

(Happy for that to be separate if so)

natdeanlewissoftwire:
> Is there a change here that we also need to reflect in a question in the service?
> 
> (Happy for that to be separate if so)

Yep good point - I've just checked the actual change and it's minor so I've just included it here

kosiakkatrina (approved):

PR #2278, opened 2024-02-29
CLDC-2069 Find UPRNs by address

natdeanlewissoftwire:
For lettings and sales, for 2024, adds:
- a full flow to search by first line of address and postcode in service, returning up to 10 matches over a confidence threshold from the OS Places API
- bulk upload functionality to auto-select the best result if it meets confidence criteria, and appropriate validation

Here's a basic flow through the new pages:

I don't know the uprn 😕 
image

Entering bad data... ⌨️ 
image

API can't find an address (with sufficient confidence) 0️⃣ 
image

Entering a better search 💡 
image

Selecting the right option ✅ 
image

Uprn is inferred 🥳 
image


app/models/form/lettings/pages/address_fallback.rb
```diff
@@ -0,0 +1,25 @@
+class Form::Lettings::Pages::AddressFallback < ::Form::Page
```
natdeanlewissoftwire, line 1: We should still allow address input if we can't match a UPRN as a last resort

app/models/form/lettings/pages/address_matcher.rb
```diff
@@ -0,0 +1,19 @@
+class Form::Lettings::Pages::AddressMatcher < ::Form::Page
```
natdeanlewissoftwire, line 1: This is the search page where a user inputs address line 1 and postcode for the API request

app/models/form/lettings/pages/address_selection.rb
```diff
@@ -0,0 +1,28 @@
+class Form::Lettings::Pages::AddressSelection < ::Form::Page
```
natdeanlewissoftwire: This page has the results returned by the OS Places API (if any)

app/models/form/lettings/pages/no_address_found.rb
```diff
@@ -0,0 +1,26 @@
+class Form::Lettings::Pages::NoAddressFound < ::Form::Page
```
natdeanlewissoftwire, line 1: This shows if the OS Places API doesn't return any addresses

app/models/lettings_log.rb
```diff
@@ -34,6 +34,7 @@ class LettingsLog < Log
   before_validation :reset_previous_location_fields!, unless: :previous_postcode_known?
   before_validation :set_derived_fields!
   before_validation :process_uprn_change!, if: :should_process_uprn_change?
+  before_validation :process_address_change!, if: :should_process_address_change?
```
natdeanlewissoftwire, line 37: same structure as the process_uprn_change method above

app/models/log.rb
```diff
@@ -53,7 +53,9 @@ class Log < ApplicationRecord
   scope :filter_by_owning_organisation, ->(owning_organisation, _user = nil) { where(owning_organisation:) }
   scope :filter_by_managing_organisation, ->(managing_organisation, _user = nil) { where(managing_organisation:) }
 
-  attr_accessor :skip_update_status, :skip_update_uprn_confirmed, :skip_dpo_validation
+  attr_accessor :skip_update_status, :skip_update_uprn_confirmed, :select_best_address_match, :skip_dpo_validation
```
natdeanlewissoftwire, line 56: similar to skip_update_uprn_confirmed, the new attribute select_best_address_match is only added to bulk uploads to allow us to auto-select the best match (if it's over a confidence of 0.7 which is at least a "Partial" match - this means the data the user input matches the response, but they might not have added all the data e.g. address line 2 that isn't compulsory in bulk upload)

app/models/log.rb
```diff
@@ -75,6 +78,71 @@ def process_uprn_change!
...
   end
 
+  def process_address_change!
+    if [address_line1_input, postcode_full_input].all?(&:present?) && (address_selection.present? || select_best_address_match.present?)
+      service = AddressClient.new(address_string)
+      service.call
+
+      return errors.add(:address_selection, :address_error, message: service.error) if service.error.present?
+
+      if select_best_address_match
```
natdeanlewissoftwire, line 87: this block is just relevant for bulk uploads to select the first match

app/models/log.rb
```diff
@@ -75,6 +78,71 @@ def process_uprn_change!
...
+        presenter = AddressDataPresenter.new(service.result.first)
+        os_good_match_threshold = 0.8
+        if presenter.match >= os_good_match_threshold
+          self.address_selection = 0
+        else
+          return nil
+        end
+      end
+
+      if address_selection.between?(0, 9)
```
natdeanlewissoftwire: user-selected match

app/models/log.rb
```diff
@@ -75,6 +78,71 @@ def process_uprn_change!
...
+        self.uprn_confirmed = 1
+        self.address_selection = nil
+        self.uprn = presenter.uprn
+        self.address_line1 = presenter.address_line1
+        self.address_line2 = presenter.address_line2
+        self.town_or_city = presenter.town_or_city
+        self.postcode_full = presenter.postcode
+        self.county = nil
+        process_postcode_changes!
+      elsif address_selection == 100
```
natdeanlewissoftwire: this is the "my address is not listed" option

app/models/log.rb
```diff
@@ -75,6 +78,71 @@ def process_uprn_change!
...
+        self.uprn = nil
+        self.address_line2 = nil
+        self.town_or_city = nil
+        self.county = nil
+      end
+    end
+  end
+
+  def address_string
+    "#{address_line1_input}, , , #{postcode_full_input}"
```
natdeanlewissoftwire: separating the sections by commas gives better API responses

app/models/log.rb
```diff
@@ -75,6 +78,71 @@ def process_uprn_change!
...
+        self.uprn = nil
+        self.address_line2 = nil
+        self.town_or_city = nil
+        self.county = nil
+      end
+    end
+  end
+
+  def address_string
+    "#{address_line1_input}, #{postcode_full_input}"
```
natdeanlewissoftwire, line 120: separating by commas gives a better API response

app/models/log.rb
```diff
@@ -75,6 +78,71 @@ def process_uprn_change!
...
+  end
+
+  def address_options
```
natdeanlewissoftwire, line 123: used for address_selection page's answer options

app/services/address_client.rb
```diff
@@ -0,0 +1,52 @@
...
+    client
+  end
+
+  def endpoint_uri
+    uri = URI(PATH)
+    params = {
+      query: address,
+      key: ENV["OS_DATA_KEY"],
+      maxresults: 10,
+      minmatch: 0.4,
```
natdeanlewissoftwire, line 43: from testing, this minimum match of 0.4 is the threshold below which reponse data is very noisy, and user has just given bad input (so should input it again, hence we return none of these noisy results)

natdeanlewissoftwire: the [technical specs for the API](https://osdatahub.os.uk/docs/places/technicalSpecification) say 0.6 is the "No match" threshold but as we are only giving first line of address and postcode ours is slightly lower

kosiakkatrina (approved):

kosiakkatrina (approved):

PR #2279, opened 2024-02-29
CLDC-3264: Setup to allow using future form on staging for testing

RachaelBooth:


RachaelBooth:
There is a bit of "just be sensible" here: the allow_future_form_use toggle as defined here will allow you to fill in logs on the service with dates arbitrarily far in the future, although e.g. they won't show up on the relevant logs page because even with no filters selected that does filter to the available years (current, previous, next, and when relevant archived). For bulk upload it only allows the next year - but it will show you the option of that even if we've not actually defined a row parser for that year yet.

I reckon we're probably fine with this - we're only going to set it on test environments after all.

app/services/feature_toggle.rb
```diff
@@ -1,19 +1,6 @@
...
-
-  def self.startdate_two_week_validation_enabled?
-    Rails.env.production? || Rails.env.test? || Rails.env.staging?
-  end
-
-  def self.saledate_two_week_validation_enabled?
-    Rails.env.production? || Rails.env.test? || Rails.env.staging?
+  def self.allow_future_form_use?
+    Rails.env.development? || Rails.env.review? || Rails.env.staging?
   end
```
kosiakkatrina, line 4: nice

kosiakkatrina (approved):

PR #2280, opened 2024-02-29
CLDC-1944: Replace soft date validation with hard date validation for 2024 and allow dates before 2000

RobertSullivanSoftwire:


app/controllers/form_controller.rb
```diff
@@ -86,7 +86,7 @@ def responses_for_page(page)
         year = params[@log.model_name.param_key]["#{question.id}(1i)"]
         next unless [day, month, year].any?(&:present?)
 
-        result[question.id] = if Date.valid_date?(year.to_i, month.to_i, day.to_i) && year.to_i.between?(2000, 2200)
+        result[question.id] = if Date.valid_date?(year.to_i, month.to_i, day.to_i) && year.to_i.positive?
```
RobertSullivanSoftwire, line 89: For soft validation, we want to retain the original date where a valid but suspicious date is input. In every other scenario we have stricter rules on the date range anyway.

app/models/form/sales/pages/handover_date_check.rb
```diff
@@ -6,7 +6,7 @@ def initialize(id, hsh, subsection)
                    { "saledate_check" => 1, "hodate_3_years_or_more_saledate?" => true }]
     @informative_text = {}
     @title_text = {
-      "translation" => "validations.sale_information.hodate.must_be_less_than_3_years_from_saledate",
+      "translation" => "validations.sale_information.hodate.must_be_less_than_3_years_from_saledate_soft",
```
kosiakkatrina: I think there is a separate `soft_validations` section (instead of `validations`) in our translations file, would it make sense to move this message there instead of suffixing it?

app/models/validations/sales/sale_information_validations.rb
```diff
@@ -1,14 +1,21 @@
...
+    return unless record.hodate.present? && date_valid?("hodate", record)
+    return if record.saledate.blank?
 
     if record.hodate > record.saledate
       record.errors.add :hodate, I18n.t("validations.sale_information.hodate.must_be_before_saledate")
       record.errors.add :saledate, I18n.t("validations.sale_information.saledate.must_be_after_hodate")
     end
+
+    if record.saledate - record.hodate >= 3.years && record.form.start_year_after_2024?
+      record.errors.add :hodate, :over_a_year_from_saledate, message: I18n.t("validations.sale_information.hodate.must_be_less_than_3_years_from_saledate")
```
kosiakkatrina: What does `over_a_year_from_saledate` give us in this scenario?

RobertSullivanSoftwire: I don't think it's doing anything. The only thing that has me slightly doubting myself is that it's also on line 30 below (which is unchanged), and I can't see what it's doing there either.

app/models/form/sales/pages/handover_date_check.rb
RobertSullivanSoftwire: Good spot - done!

app/models/validations/sales/sale_information_validations.rb
kosiakkatrina: Looks like we used it to clear data during migrations from the old service, we don’t need it anymore!

kosiakkatrina (approved):

PR #2282, opened 2024-03-01
Update schema since rails 7

natdeanlewissoftwire:


db/schema.rb
```diff
@@ -304,10 +304,10 @@
     t.integer "scharge_value_check"
     t.integer "pscharge_value_check"
     t.integer "duplicate_set_id"
+    t.integer "accessible_register"
```
natdeanlewissoftwire, line 307: lmk if this reordering is local to me for some reason  but I'm assuming not :)

spec/fixtures/exports/general_needs_log_24_25.xml
```diff
@@ -144,8 +144,8 @@
     <discarded_at/>
     <creation_method>1</creation_method>
     <duplicate_set_id/>
-    <nationality_all/>
```
natdeanlewissoftwire, line 147: due to reordering of fields - I _think_ this is fine but haven't worked on export much so not sure if changing this is a problem

spec/services/merge/merge_organisations_service_spec.rb
```diff
@@ -221,8 +221,8 @@
               expect(absorbed_scheme.locations.count).to eq(1)
               absorbed_location = absorbed_scheme.locations.first
 
-              expect(absorbed_scheme.startdate).to eq(Time.zone.today + 1.month)
```
natdeanlewissoftwire: be explicit here now we're defaulting to TimeWithZone - we're doing this elsewhere in this file but it wasn't relevant for these fields as they had unspecified precision before

db/schema.rb
RobertSullivanSoftwire: Pretty sure I've seen this one as well.

spec/fixtures/exports/general_needs_log_24_25.xml
RobertSullivanSoftwire: I don't have any practical experience of this either, but it would be extremely odd for this behaviour to matter.

RobertSullivanSoftwire (approved):

PR #2283, opened 2024-03-01
Fix test failures due to BST changeover

RobertSullivanSoftwire:


RachaelBooth (approved):
LGTM - is there anything in this file that might hit the same problem in the other direction when we get to April? (not for this PR, just be good to check that before we hit it)

RobertSullivanSoftwire:
I had it in my head that I'd looked and there weren't, but it looks like there are actually 2 tests with that issue. I'll put up a separate PR for them!

RobertSullivanSoftwire:
Actually, looks like I was right the first time - the tests in the other direction look like they've already been addressed at some point.

PR #2285, opened 2024-03-01
CLDC-2478 Allow deleting locations

kosiakkatrina:
Allow support users to delete incomplete or deactivated locations that are not used in editable collection periods:
image
image
or
image

Incomplete locations open on check answers page, so added a delete button there as well:
image

Will add deleting schemes separately

### DIFF
--- a/app/policies/location_policy.rb
+++ b/app/policies/location_policy.rb
@@ -89,8 +89,7 @@ private
 
   def has_any_logs_in_editable_collection_period
     editable_from_date = FormHandler.instance.earliest_open_for_editing_collection_start_date
-    editable_logs = LettingsLog.where(location_id: location.id).after_date(editable_from_date)
 
-    LettingsLog.where(location_id: location.id, startdate: nil).any? || editable_logs.any?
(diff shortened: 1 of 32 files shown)


app/policies/location_policy.rb
```diff
@@ -75,4 +86,11 @@ def scheme
   def scheme_owned_by_user_org_or_stock_owner
     scheme&.owning_organisation == user.organisation || user.organisation.stock_owners.exists?(scheme&.owning_organisation_id)
   end
+
+  def has_any_logs_in_editable_collection_period
+    editable_from_date = FormHandler.instance.earliest_open_for_editing_collection_start_date
+    editable_logs = LettingsLog.where(location_id: location.id).after_date(editable_from_date)
+
+    LettingsLog.where(location_id: location.id, startdate: nil).any? || editable_logs.any?
```
RachaelBooth: Would it be worth reordering this, or combining the queries? I'm thinking for reducing the number of queries / amount of data we're loading for performance reasons.

kosiakkatrina: Hmm, I've changed it a bit, does it make it any better?

RachaelBooth: Yeah, that does it

RachaelBooth (approved):

PR #2286, opened 2024-03-04
CLDC-2478 Allow deleting scheme

kosiakkatrina:
Merging into https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/2285 to separate tech and PO review.
Allow support users to delete incomplete or deactivated schemes that are not used in editable collection periods:

image
image
image

Incomplete schemes open on check answers page, so added a delete button there as well:
image

### DIFF
--- a/app/controllers/organisations_controller.rb
+++ b/app/controllers/organisations_controller.rb
@@ -21,7 +21,7 @@ class OrganisationsController < ApplicationController
   end
 
   def schemes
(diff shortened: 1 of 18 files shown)


RobertSullivanSoftwire:
All looks good to me - my main question is whether the following cases are worth adding tests for:
- Organisations controller does not return deleted schemes
- Schemes controller does not return deleted schemes
- Deleted scheme correctly returns deleted status

RobertSullivanSoftwire (approved):

PR #2288, opened 2024-03-05
CLDC-2479 Allow deleting users

kosiakkatrina:
Allow support users to delete deactivated users that have no logs in editable collection period and aren't DPOs who signed the DSAs.
image
image
Change how we display status information in user table and details.
image

When deactivated user can't be deleted:
image

Also refactored sign_in user into before blocks in `spec/requests/users_controller_spec.rb` so the change looks longer than it actually is!


spec/models/user_spec.rb
```diff
@@ -481,4 +481,39 @@
...
+  describe "#status" do
+    context "when the user is active" do
+      let(:user) { create(:user, last_sign_in_at: Time.zone.yesterday) }
+
+      it "returns the status of the user" do
+        expect(user.status).to eq(:active)
+      end
+    end
+
+    context "when the user is not signed in" do
```
RachaelBooth: has not?

RachaelBooth (approved):

spec/models/user_spec.rb
kosiakkatrina: Does it make much difference? I think we use is not in the rest of the codebase (?)

RachaelBooth: This test is specifically about when the user has never signed in though, rather than whether they currently are.

PR #2292, opened 2024-03-06
CLDC-3277 Set managing org to created by org on date change

kosiakkatrina:
For all 2023 sales logs we infer managing organisation as created by organisation. For 2024 we've added managing organisation question. If 2024 log is created with a manually selected managing organisation and the date then changes to 2023 we want to reset the managing organisation to created by organisation. 
We do this so that we keep consistent behaviour for all 2023 logs and there's also no way for non support users to change managing org once the date is in 2023

### DIFF
--- a/app/controllers/form_controller.rb
+++ b/app/controllers/form_controller.rb
@@ -1,4 +1,6 @@
 class FormController < ApplicationController
+  include CollectionTimeHelper
+
   before_action :authenticate_user!
   before_action :find_resource, only: %i[review]
   before_action :find_resource_by_named_id, except: %i[review]
@@ -92,6 +94,11 @@ private
                                 Date.new(0, 1, 1)
                               end
       end
+
+      if question.id == "saledate" && set_managing_organisation_to_created_by_organisation?(result["saledate"])
+        result["managing_organisation_id"] = @log.created_by.organisation_id
+      end
+
       next unless question_params
 
       if %w[checkbox validation_override].include?(question.type)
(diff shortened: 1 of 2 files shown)


app/controllers/form_controller.rb
```diff
@@ -104,13 +111,8 @@ def responses_for_page(page)
...
         owning_organisation = result["owning_organisation_id"].present? ? Organisation.find(result["owning_organisation_id"]) : nil
-        if current_user.support? && @log.managing_organisation.blank? && owning_organisation&.managing_agents&.empty?
-          result["managing_organisation_id"] = owning_organisation.id
-        elsif owning_organisation&.absorbing_organisation == current_user.organisation
-          result["managing_organisation_id"] = owning_organisation.id
-        elsif @log.managing_organisation&.absorbing_organisation == current_user.organisation && owning_organisation == current_user.organisation
-          result["managing_organisation_id"] = owning_organisation.id
-        end
+
+        result["managing_organisation_id"] = owning_organisation.id if set_managing_organisation_to_owning_organisation?(owning_organisation)
```
kosiakkatrina, line 115: This is just refactor

app/controllers/form_controller.rb
```diff
@@ -321,4 +323,19 @@ def update_logs_with_duplicate_set_id(log, dynamic_duplicates, duplicate_set_id)
...
+  def set_managing_organisation_to_owning_organisation?(owning_organisation)
+    return true if current_user.support? && @log.managing_organisation.blank? && owning_organisation&.managing_agents&.empty?
+    return true if owning_organisation&.absorbing_organisation == current_user.organisation
+    return true if @log.managing_organisation&.absorbing_organisation == current_user.organisation && owning_organisation == current_user.organisation
+
+    false
+  end
+
+  def set_managing_organisation_to_created_by_organisation?(saledate)
+    return false if current_user.support?
```
RachaelBooth, line 336: I think there's a stupid case here where the original log creator was not support, but the person making the date change is and they don't think to also change the managing org. 

That's seems sufficiently unlikely to me that I assume we don't think it will happen.

RachaelBooth (approved):

app/controllers/form_controller.rb
kosiakkatrina: Yeah, I think you're right. I think it's probably unlikely, but let me double check if we want to do anything about it!

kosiakkatrina: This is sufficiently uncommon not to address it!

PR #2293, opened 2024-03-07
CLDC-2481: Add ability to disable organisations

RobertSullivanSoftwire:


config/locales/en.yml
```diff
@@ -824,6 +826,9 @@ Make sure these answers are correct."
     offered: "Times previously offered since becoming available"
 
   warnings:
+    organisation:
+      deactivate: "All schemes and users at this organisation will be deactivated. All the organisation's relationships will be removed. It will no longer be possible to create logs for this organisation."
+      reactivate: "All schemes, users, and relationships that were active when this organisation was deactivated will be reactivated."
```
RobertSullivanSoftwire, line 867: This is copied from the designs - do we usually use the Oxford comma in situations where it's not necessary?

app/views/organisations/toggle_active.html.erb
```diff
@@ -0,0 +1,29 @@
...
+<% end %>
+
+<%= form_for(@organisation, as: :organisation, html: { method: :patch }) do |f| %>
+  <div class="govuk-grid-row">
+    <div class="govuk-grid-column-two-thirds">
+      <h1 class="govuk-heading-l">
+        <span class="govuk-caption-l"><%= @organisation.name %></span>
+        Are you sure you want to <%= action %> this organisation?
+      </h1>
+      <%= govuk_warning_text text: I18n.t("warnings.organisation.#{action}") %>
```
RobertSullivanSoftwire, line 17: This appears to display slightly differently from the designs (but consistently with https://design-system.service.gov.uk/components/warning-text/), in the the warning icon appears at the top rather than centred.

app/views/organisations/toggle_active.html.erb
```diff
@@ -0,0 +1,29 @@
...
+        href: organisation_path(@organisation),
+      ) %>
+<% end %>
+
+<%= form_for(@organisation, as: :organisation, html: { method: :patch }) do |f| %>
+  <div class="govuk-grid-row">
+    <div class="govuk-grid-column-two-thirds">
+      <h1 class="govuk-heading-l">
+        <span class="govuk-caption-l"><%= @organisation.name %></span>
+        Are you sure you want to <%= action %> this organisation?
```
RobertSullivanSoftwire, line 15: I've tried to keep the use of I18n consistent with the other similar pages, but it's worth flagging up that some text that currently isn't parameterised probably could be.

spec/models/form/lettings/questions/stock_owner_spec.rb
```diff
@@ -205,21 +208,26 @@
...
 
     context "when user is support" do
-      let(:user) { create(:user, :support) }
+      user = nil
+      log = nil
+
+      before do
+        user = create(:user, :support)
+        log = create(:lettings_log)
+      end
```
RobertSullivanSoftwire, line 217: This is a pretty messy contrivance, to achieve the following:

- user and log need to be accessed in some tests cases
- user and log (and their related organisations) need to be instantiated in all test cases, even ones where they aren't referenced.

Open to better solutions to this problem!

RobertSullivanSoftwire: Oh hey, `let!` solves exactly this problem. Let's do that.

spec/models/form/sales/questions/owning_organisation_id_spec.rb
```diff
@@ -206,20 +209,21 @@
 
     context "when user is support" do
       let(:user) { create(:user, :support, organisation: organisation_1) }
-
       let(:log) { create(:lettings_log, created_by: user) }
 
-      let(:non_stock_organisation) { create(:organisation, holds_own_stock: false) }
```
RobertSullivanSoftwire, line 210: The timing here meant that this didn't organisation didn't actually exist at the relevant point during the tests.

spec/models/form/lettings/questions/stock_owner_spec.rb
```diff
@@ -205,21 +208,26 @@
...
 
-      let(:log) { create(:lettings_log) }
+      it "shows active orgs where organisation holds own stock" do
+        non_stock_organisation = create(:organisation, name: "Non-stockholding org", holds_own_stock: false)
+        inactive_organisation = create(:organisation, name: "Inactive org", active: false)
 
-      let(:non_stock_organisation) { create(:organisation, holds_own_stock: false) }
```
RobertSullivanSoftwire, line 210: The timing here meant that this organisation didn't actually exist at the relevant point in the tests.

app/controllers/organisations_controller.rb
```diff
@@ -94,10 +94,42 @@ def edit
...
+  end
+
   def update
-    if current_user.data_coordinator? || current_user.support?
+    if (current_user.data_coordinator? && org_params[:active].nil?) || current_user.support?
       if @organisation.update(org_params)
-        flash[:notice] = I18n.t("organisation.updated")
+        case org_params[:active]
+        when "false"
+          @organisation.users.filter_by_active
```
kosiakkatrina: Would it make sense to have something like `@organisation.users.filter_by_active.each(&:deactivate!)` and have a method on the user instead?

app/controllers/organisations_controller.rb
```diff
@@ -94,10 +94,42 @@ def edit
...
+                       .update_all(
+                         active: false,
+                         confirmed_at: nil,
+                         sign_in_count: 0,
+                         initial_confirmation_sent: false,
+                         reactivate_with_organisation: true,
+                       )
```
kosiakkatrina: I assume we set all of these when we deactivate user independently of org? I like `reactivate_with_organisation`!

app/models/location.rb
```diff
@@ -139,7 +139,8 @@ def status
 
   def status_at(date)
     return :incomplete unless confirmed
-    return :deactivated if open_deactivation&.deactivation_date.present? && date >= open_deactivation.deactivation_date
+    return :deactivated if scheme.owning_organisation.status_at(date) == :deactivated ||
+      open_deactivation&.deactivation_date.present? && date >= open_deactivation.deactivation_date
```
kosiakkatrina, line 159: There's some awkward code here, cause we also have a `deactivated` scope in this file, which is independent of this and probably needs updating

app/models/organisation.rb
```diff
@@ -138,6 +139,7 @@ def status
 
   def status_at(date)
     return :merged if merge_date.present? && merge_date < date
+    return :deactivated unless active
```
kosiakkatrina, line 143: Hmm, I feel like an org could be both of these at the same time (?) but should we/ do we allow merged organisations to get deactivated?

app/models/scheme.rb
```diff
@@ -265,7 +265,8 @@ def status
 
   def status_at(date)
     return :incomplete unless confirmed && locations.confirmed.any?
-    return :deactivated if open_deactivation&.deactivation_date.present? && date >= open_deactivation.deactivation_date
+    return :deactivated if owning_organisation.status_at(date) == :deactivated ||
+      (open_deactivation&.deactivation_date.present? && date >= open_deactivation.deactivation_date)
```
kosiakkatrina, line 286: Similar to locations, I think this also has `deactivated` scope

app/views/locations/show.html.erb
```diff
@@ -42,6 +42,6 @@
   </div>
 </div>
 
-<% if LocationPolicy.new(current_user, @location).deactivate? %>
+<% if @location.scheme.owning_organisation.active? && LocationPolicy.new(current_user, @location).deactivate? %>
```
kosiakkatrina, line 50: Should this be inside the policy?

app/views/organisations/show.html.erb
```diff
@@ -40,3 +40,15 @@
     <%= render partial: "organisations/merged_organisation_details" %>
   </div>
 </div>
+
+<% if @organisation.active? %>
```
kosiakkatrina: Similar to above, should this be in the policy?

app/controllers/organisation_relationships_controller.rb
```diff
@@ -14,7 +14,7 @@ class OrganisationRelationshipsController < ApplicationController
   ]
 
   def stock_owners
-    stock_owners = organisation.stock_owners
+    stock_owners = organisation.stock_owners.filter_by_active
```
kosiakkatrina, line 17: Ooh, I was wondering how we'd _restore_ these upon reactivation, and it looks like this might be all that's needed, nice

app/controllers/organisations_controller.rb
RobertSullivanSoftwire: I think I actually structured it that way originally (since it also allows commonisation with the users controller), then switched to this on the basis that it's more performant. Happy to switch back, though (especially since reactivate is using the less performant approach anyway!), since it's otherwise nicer.

app/controllers/organisations_controller.rb
RobertSullivanSoftwire: Yes - this is otherwise the same as deactivating users. I have to admit, I kind of feel guilty about `reactivate_with_organisation`, but it was the cleanest way I could see of getting this behaviour.

app/models/organisation.rb
RobertSullivanSoftwire: Good question - I'll ask on the ticket.

RobertSullivanSoftwire: Looks like the answer is no - merged organisation should not be deactivatable.

app/views/locations/show.html.erb
RobertSullivanSoftwire: My thinking here was that I'd usually expect a policy to be restricted to authorisation logic and this feels like it's moving over into validation instead. On the other hand, we already have a very similar rule in our logs policies, so my assumptions about how things work are probably just off-base!

kosiakkatrina (approved):

PR #2294, opened 2024-03-07
CLDC-3283 Validate allocation values

kosiakkatrina:


RachaelBooth (approved):
Given that this is both years, can you make the PR title reflect that (so that once merged it's in the commit message)?

Otherwise LGTM

PR #2295, opened 2024-03-07
CLDC-3278: Restructure tenancy length validations to only error on relevant fields

RachaelBooth:


app/models/validations/tenancy_validations.rb
```diff
@@ -3,48 +3,67 @@ module Validations::TenancyValidations
   # or 'validate_' to run on submit as well
   include Validations::SharedValidations
```
RachaelBooth, line 5: These are very similar validations, but they apply to the question as asked on different pages with slightly different criteria for how you get there and what affects the minimum value - hence splitting them out.

app/models/validations/tenancy_validations.rb
```diff
@@ -3,48 +3,67 @@ module Validations::TenancyValidations
...
-        condition: (record.is_assured_shorthold_tenancy? && !is_in_range) && is_present,
-        error: I18n.t(
-          "validations.tenancy.length.shorthold",
-          min_tenancy_length: min_tenancy_length(record),
-        ),
-      },
-      {
-        condition: (record.is_secure_tenancy? && !is_in_range) && is_present,
-        error: I18n.t(
-          "validations.tenancy.length.secure",
```
RachaelBooth, line 20: This message almost matched that for shorthold but implied that you could leave the answer blank - you can't actually do that for a secure fixed term tenancy. So there's now a message for fixed and a message for periodic (where you can leave it blank instead).

spec/models/lettings_log_spec.rb
```diff
@@ -112,10 +112,17 @@
     end
 
     it "validates tenancy type" do
-      expect(validator).to receive(:validate_fixed_term_tenancy)
       expect(validator).to receive(:validate_other_tenancy_type)
     end
 
+    it "validates tenancy length" do
```
RachaelBooth, line 118: I'm not really sure this test is useful - but I think that's part of taking a wider look at our tests so I've updated it to match the current setup for the moment.

app/models/validations/tenancy_validations.rb
```diff
@@ -3,48 +3,67 @@ module Validations::TenancyValidations
...
-
-    conditions.each do |condition|
-      next unless condition[:condition]
-
-      record.errors.add :needstype, condition[:error]
-      record.errors.add :rent_type, condition[:error] if rent_type_dependent_conditions.include?(condition)
-      record.errors.add :tenancylength, :tenancylength_invalid, message: condition[:error]
-      record.errors.add :tenancy, condition[:error]
-    end
+  # N.B. To match the page split and naming, this is the supported housing case
```
natdeanlewissoftwire: Can we rename the validation method(s) to remove the need for this comment?

natdeanlewissoftwire (approved):
Looks good!

app/models/validations/tenancy_validations.rb
RachaelBooth: We could if we break the matching the page names thing - which to be fair might be fine?

PR #2296, opened 2024-03-08
CLDC-3284 Clear/error all charges if any charges are missing

kosiakkatrina:
We get an error in the full error report if any of the charges are missing, but when we say we’re gonna "clear" the missing charge answer what we’re doing instead is setting it to 0. We would now error all the charges and not set them unless all of them are answered.

### DIFF
--- a/app/services/bulk_upload/lettings/year2024/row_parser.rb
+++ b/app/services/bulk_upload/lettings/year2024/row_parser.rb
@@ -376,13 +376,13 @@ class BulkUpload::Lettings::Year2024::RowParser
 
   validate :validate_created_by_exists, on: :after_log
   validate :validate_created_by_related, on: :after_log
+  validate :validate_all_charges_given, on: :after_log, if: proc { is_carehome.zero? }
 
   validate :validate_nulls, on: :after_log
 
   validate :validate_uprn_exists_if_any_key_address_fields_are_blank, on: :after_log, unless: -> { supported_housing? }
 
   validate :validate_incomplete_soft_validations, on: :after_log
-  validate :validate_all_charges_given, on: :after_log, if: proc { is_carehome.zero? }
   validate :validate_nationality, on: :after_log
 
   def self.question_for_field(field)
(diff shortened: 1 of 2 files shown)


app/services/bulk_upload/lettings/year2023/row_parser.rb
```diff
@@ -383,6 +383,8 @@ class BulkUpload::Lettings::Year2023::RowParser
 
   validate :validate_declaration_acceptance, on: :after_log
 
+  validate :validate_all_charges_given, on: :after_log, if: proc { is_carehome.zero? }
```
kosiakkatrina: Wanna run this before validating nulls, cause validate_nulls would add an extra error otherwise

app/services/bulk_upload/lettings/year2023/row_parser.rb
```diff
@@ -863,12 +864,20 @@ def validate_all_charges_given
...
-      if public_send(field.to_sym).blank?
-        errors.add(field, I18n.t("validations.financial.charges.missing_charges", question: charge))
-      end
+      next if public_send(field.to_sym).present?
+
+      errors.add(:field_128, I18n.t("validations.financial.charges.missing_charges", question: charge)) unless field != :field_128 && field_128.blank?
+      errors.add(:field_129, I18n.t("validations.financial.charges.missing_charges", question: charge)) unless field != :field_129 && field_129.blank?
+      errors.add(:field_130, I18n.t("validations.financial.charges.missing_charges", question: charge)) unless field != :field_130 && field_130.blank?
+      errors.add(:field_131, I18n.t("validations.financial.charges.missing_charges", question: charge)) unless field != :field_131 && field_131.blank?
+      errors.add(:field_132, I18n.t("validations.financial.charges.missing_charges", question: charge)) unless field != :field_132 && field_132.blank?
```
RachaelBooth: This was slightly confusing to work out - I think it's meaning that blank fields get the error for their own missing charge and non-blank fields get the errors for all of the missing charges (which will end up with just the first missing one displaying)?

That makes sense, I'm just wondering if there's a clearer way to write it? (But there might not be - happy to leave it if nothing comes to mind)

app/services/bulk_upload/lettings/year2023/row_parser.rb
```diff
@@ -1201,11 +1210,11 @@ def attributes_for_log
     attributes["benefits"] = field_124
 
     attributes["period"] = field_126
-    attributes["brent"] = field_128
-    attributes["scharge"] = field_129
-    attributes["pscharge"] = field_130
-    attributes["supcharg"] = field_131
-    attributes["tcharge"] = field_132
+    attributes["brent"] = field_128 if all_charges_given?
```
RachaelBooth: Does clearing these at this point mean that any they have provided are not presented to the user as fields that will be cleared on upload?

app/services/bulk_upload/lettings/year2023/row_parser.rb
kosiakkatrina: Yup, was thinking this was confusing when I wrote this, so thanks for validation! I've refactored it a bit, let me know if that's at all easier to consume!

app/services/bulk_upload/lettings/year2023/row_parser.rb
kosiakkatrina: It would show all the charges in the list and the count to be deleted. We're clearing them at this point because if they get past this they're not really invalid, and the charges that were missing would get set to 0 cause we have that as a mechanism to avoid data loss
image

RachaelBooth: Cool, makes sense

RachaelBooth (approved):

app/services/bulk_upload/lettings/year2023/row_parser.rb
RachaelBooth: Yeah, I think that's great - thanks!

PR #2297, opened 2024-03-08
CLDC-2575 Allow searches by Log Id to include "log"

natdeanlewissoftwire:


app/models/lettings_log.rb
```diff
@@ -62,7 +62,7 @@ class LettingsLog < Log
       .or(filter_by_tenant_code(param))
       .or(filter_by_propcode(param))
       .or(filter_by_postcode(param))
-      .or(filter_by_id(param))
+      .or(filter_by_id(param.remove("log")))
```
RachaelBooth: Is this case sensitive? It would make sense to me not to be, or at least to remove "Log" since that's how we present them in the list.

natdeanlewissoftwire: Ah good point!

natdeanlewissoftwire: Have fixed this @RachaelBooth :)

RachaelBooth (approved):

PR #2299, opened 2024-03-08
CLDC-2741 Order log search results

kosiakkatrina:


app/models/lettings_log.rb
```diff
@@ -58,11 +58,17 @@ class LettingsLog < Log
     query.all
   }
   scope :search_by, lambda { |param|
+    by_id = Arel.sql("case when lettings_logs.id = #{param.to_i} then 0 else 1 end")
```
natdeanlewissoftwire: do we want an ILIKE for id too?

natdeanlewissoftwire: also, as these are SQL I'd personally rather they all be upper case (or at least a consistent case) - though I'm not sure this is the case thus far in the codebase so I don't have a problem if you prefer it this way :)

natdeanlewissoftwire (approved):
lgtm!

app/models/lettings_log.rb
kosiakkatrina: I think we only search IDs for exact matches, so don't think it needs to order by similarity. Good point on casing!

PR #2301, opened 2024-03-11
CLDC-3234: Prevent errors from nil hhmemb

RachaelBooth:
This fixes the "undefined method sex9" problem. 

The log in question had a validation error that applied to hhmemb, so the value was getting cleared.

### DIFF
--- a/app/models/log.rb
+++ b/app/models/log.rb
@@ -119,12 +119,13 @@ class Log < ApplicationRecord
 
   def blank_invalid_non_setup_fields!
     setup_ids = form.setup_sections.flat_map(&:subsections).flat_map(&:questions).map(&:id)
+    fields_to_keep = setup_ids + %w[hhmemb]
 
     2.times do
       next if valid?
 
       errors.each do |error|
-        next if setup_ids.include?(error.attribute.to_s)
+        next if fields_to_keep.include?(error.attribute.to_s)
 
         question = form.questions.find { |q| q.id == error.attribute.to_s }
         if question&.type == "checkbox"
--- a/app/models/validations/soft_validations.rb
+++ b/app/models/validations/soft_validations.rb
@@ -73,13 +73,17 @@ module Validations::SoftValidations
   end
 
   def all_tenants_age_and_gender_information_completed?
-    (1..hhmemb).all? do |n|
+    person_count = hhmemb || 8
+
+    (1..person_count).all? do |n|
       public_send("sex#{n}").present? && public_send("age#{n}").present? && details_known_or_lead_tenant?(n) && public_send("age#{n}_known").present? && public_send("age#{n}_known").zero?
     end
   end
 
   def all_tenants_gender_information_completed?
-    (1..hhmemb).all? do |n|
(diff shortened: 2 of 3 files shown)


app/models/log.rb
```diff
@@ -119,12 +119,13 @@ def collection_period_open_for_editing?
 
   def blank_invalid_non_setup_fields!
     setup_ids = form.setup_sections.flat_map(&:subsections).flat_map(&:questions).map(&:id)
+    fields_to_keep = setup_ids + %w[hhmemb]
```
RachaelBooth, line 122: I think this is the right call, because if hhmemb gets cleared then so does data on all the people who are no longer routed to, and it's not clear from the report that that will happen. The other fields in the relevant validation would still be cleared (since there isn't a validation that affects only hhmemb and setup fields - although there is a risk I guess that that ceases to be true at some point).

spec/models/log_spec.rb
```diff
@@ -76,5 +76,16 @@
...
+    context "when hhmemb is invalid for a lettings log" do
+      subject(:model) { build(:lettings_log, :setup_completed, hhmemb: 1, joint: 1) }
+
+      it "does not blank it" do
+        model.valid?
+        model.blank_invalid_non_setup_fields!
+        expect(model.hhmemb).to be(1)
+        expect(model.joint).to be_nil
+      end
+    end
```
RachaelBooth, line 89: N.B. I couldn't spot a hard validation for sales logs that puts an error on hhmemb.

app/models/log.rb
RachaelBooth: It is a bit weird though - please let me know if you've got thoughts / ways this could go wrong!

app/models/validations/soft_validations.rb
```diff
@@ -73,13 +73,17 @@ def female_in_pregnant_household_in_soft_validation_range?
   end
 
   def all_tenants_age_and_gender_information_completed?
-    (1..hhmemb).all? do |n|
+    person_count = hhmemb || 8
+
+    (1..person_count).all? do |n|
```
kosiakkatrina, line 78: Could we just not run these validations if hhmemb is not given?

app/models/log.rb
kosiakkatrina: What would happen if hhmemb is negative or more than 8 etc?

RachaelBooth: For bulk upload (which should be the only time we do this data clearing), we calculate hhmemb by counting how many people have data provided, so it shouldn't be possible - if it somehow did happen, then the log would still not be valid after this and we'd end up in the "correct and upload again" journey.

kosiakkatrina: What about sales? We've got hholdcount there, does that cause the same issue?

RachaelBooth: I don't think there are any validations that add errors to it - and it looks like we explicitly ask for it in the upload csv, so shouldn't treat it the same way.

kosiakkatrina (approved):

PR #2304, opened 2024-03-12
CLDC-3292 Don't validate staircase mortgage for socialhomebuy

kosiakkatrina:


app/models/validations/sales/sale_information_validations.rb
```diff
@@ -108,20 +108,20 @@ def validate_discount_and_value(record)
 
   def validate_non_staircasing_mortgage(record)
```
natdeanlewissoftwire, line 109: The PR title makes me think this is the wrong method... I think there's a different validation for staircasing staircasing logs

app/models/validations/sales/sale_information_validations.rb
```diff
@@ -108,20 +108,20 @@ def validate_discount_and_value(record)
...
-    return unless record.is_not_staircasing?
+    return unless record.is_not_staircasing? && record.type != 18
     return unless record.saledate && record.form.start_year_after_2024?
 
     if record.mortgage_used?
       return unless record.mortgage
 
       if record.mortgage_and_deposit_total != record.expected_shared_ownership_deposit_value
-        %i[mortgage value deposit equity].each do |field|
+        %i[mortgage value deposit equity type].each do |field|
```
natdeanlewissoftwire, line 118: do we need to change the validation message to make it clear this is type-dependent now?

app/models/validations/sales/sale_information_validations.rb
kosiakkatrina: Ah, good point! I think it should be added to both. It's on hold atm tho, so will see what comes out of it!

PR #2305, opened 2024-03-12
CLDC-2584 Update log summary component frontend

natdeanlewissoftwire:
Updates styling to log ID and tenant/property/purchaser Id so they're nicely centered and wrap properly:

image

### DIFF
--- a/app/components/lettings_log_summary_component.html.erb
+++ b/app/components/lettings_log_summary_component.html.erb
@@ -8,12 +8,12 @@
           <% end %>
         </h2>
         <% if log.tenancycode? or log.propcode? %>
-          <div class="app-metadata app-log-summary__details">
+          <div class="app-metadata">
             <% if log.tenancycode? %>
-                <div>Tenant <%= log.tenancycode %></div>
+              <div class = "app-log-summary__details">Tenant <%= log.tenancycode %></div>
             <% end %>
             <% if log.propcode? %>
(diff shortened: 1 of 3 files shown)


app/components/lettings_log_summary_component.html.erb
```diff
@@ -1,27 +1,21 @@
...
           <% end %>
         </h2>
         <% if log.tenancycode? or log.propcode? %>
-          <dl class="app-metadata app-metadata--inline">
+          <div class="app-metadata">
             <% if log.tenancycode? %>
-              <div class="app-metadata__item">
-                <dt class="app-metadata__term">Tenant</dt>
-                <dd class="app-metadata__definition"><%= log.tenancycode %></dd>
-              </div>
```
kosiakkatrina, line 16: Do you think description list tag might be more informative given the tenant/property codes here are metadata? (kinda?)

natdeanlewissoftwire: I see what you mean, but using these tags led to some surprirising behaviour because the tag and descriptions were being formatted as separate blocks - e.g. 
image

I could spend some time fiddling with CSS to get around this, but the issue is that we don't really want them to be separate - we want the tag and description to be associated with each other in the same string, so that it wraps nicely (and e.g. the property code can wrap under the word Property, rather than wrapping in its own container)

kosiakkatrina (approved):

PR #2307, opened 2024-03-12
CLDC-2742 Update user status display behaviour

natdeanlewissoftwire:
image
image
image

I've also updated the Deactivated text on the user list page to use the status tag:
image


RobertSullivanSoftwire (approved):
Looks good to me!

app/models/user.rb
```diff
@@ -237,6 +237,16 @@ def newest_active_unread_notification
     active_unread_notifications.last
   end
 
+  def status
```
RobertSullivanSoftwire, line 240: This might be worth a test?

PR #2310, opened 2024-03-13
CLDC-3300 Update BU mortgageused error

kosiakkatrina:
option 3 (don't know) for mortgage used it a conditional option, but we've not been validating it on BU.
It shouldn't be allowed for 23/24 logs, discounted ownership logs or shared ownership logs where stairowned is not 100 (we already have a validation for that)

### DIFF
--- a/app/services/bulk_upload/sales/year2023/row_parser.rb
+++ b/app/services/bulk_upload/sales/year2023/row_parser.rb
@@ -450,6 +450,7 @@ class BulkUpload::Sales::Year2023::RowParser
             on: :after_log
 
   validate :validate_buyer1_economic_status, on: :before_log
+  validate :validate_mortgageused, on: :before_log
   validate :validate_nulls, on: :after_log
   validate :validate_valid_radio_option, on: :before_log
 
@@ -1330,6 +1331,14 @@ private
   def validate_buyer1_economic_status
     if field_35 == 9
       errors.add(:field_35, "Buyer 1 cannot be a child under 16")
+    end
+  end
+
+  def validate_mortgageused
+    if mortgageused.present? && mortgageused.to_i != 1 && mortgageused.to_i != 2
+      errors.add(:field_105, I18n.t("validations.invalid_option", question: "Was a mortgage used for the purchase of this property? - Shared ownership")) if shared_ownership?
+      errors.add(:field_119, I18n.t("validations.invalid_option", question: "Was a mortgage used for the purchase of this property? - Discounted ownership")) if discounted_ownership?
+      errors.add(:field_128, I18n.t("validations.invalid_option", question: "Was a mortgage used for the purchase of this property? - Outright sale")) if outright_sale?
     end
   end
 end
--- a/app/services/bulk_upload/sales/year2023/row_parser.rb
+++ b/app/services/bulk_upload/sales/year2023/row_parser.rb
@@ -425,6 +425,14 @@ class BulkUpload::Sales::Year2023::RowParser
             },
             on: :before_log
 
+  validates :field_105,
+            inclusion: {
+              in: [1, 2],
+              if: proc { field_89.present? && field_89 != 100 && shared_ownership? },
+              question: QUESTIONS[:field_105],
+            },
+            on: :before_log
+
   validates :field_13,
             presence: {
               message: I18n.t("validations.not_answered", question: "buyers living in property"),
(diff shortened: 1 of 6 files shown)


app/services/bulk_upload/sales/year2024/row_parser.rb
```diff
@@ -1371,4 +1372,10 @@ def validate_buyer_2_nationality
   def valid_nationality_options
     %w[0] + GlobalConstants::COUNTRIES_ANSWER_OPTIONS.keys # 0 is "Prefers not to say"
   end
+
+  def validate_mortgageused
+    if discounted_ownership? && mortgageused.present? && mortgageused.to_i != 1 && mortgageused.to_i != 2
```
RobertSullivanSoftwire: I'm reading this as raising an error when:
- The ownership type is discounted 
- Mortgage used is "don't know"
- The year is 24/25

The rules in the ACs suggest we should be raising an error when:
- The ownership type is discounted _or_ shared
- Mortgage used is "don't know"
- _Percentage owned is not 100_
- The year is 24/25

Which suggests to me we also need a validation rule for non-100% shared ownership (similar to that in the 2023 parser).

Have I missed something here?

app/services/bulk_upload/sales/year2023/row_parser.rb
```diff
@@ -425,6 +425,14 @@ class BulkUpload::Sales::Year2023::RowParser
             },
             on: :before_log
 
+  validates :field_105,
+            inclusion: {
+              in: [1, 2],
+              if: proc { field_89.present? && field_89 != 100 && shared_ownership? },
```
RobertSullivanSoftwire: I can see this matches the ACs but I'm a bit confused because they don't seem to match the text description, which suggests "don't know" should never be allowed for 2023. Do you know what's going on there?

spec/services/bulk_upload/sales/year2023/row_parser_spec.rb
```diff
@@ -1008,6 +1008,60 @@ def purch_id_field
         end
       end
     end
+
+    describe "#field_105" do
+      let(:attributes) { valid_attributes.merge({ field_7: "1", field_105: "3" }) }
+
+      it "does not allow 3 (don't know) as an option for shared ownership" do
```
RobertSullivanSoftwire: I think this test case is current a bit misleading - I believe we currently do allow "don't know" for shared ownership under some conditions.

spec/services/bulk_upload/sales/year2024/row_parser_spec.rb
```diff
@@ -1017,14 +1017,6 @@ def purch_id_field
...
 
-      context "when it's not shared ownership" do
-        let(:attributes) { setup_section_params.merge(field_8: "2", field_103: "3", field_86: "1", field_87: "50", field_88: "99", field_109: nil) }
-
-        it "does not add errors" do
-          expect(parser.errors[:field_103]).not_to include("Enter a valid value for Was a mortgage used for the purchase of this property? - Shared ownership")
-        end
-      end
-
       context "when value is 3 and stairowned is 100" do
```
RobertSullivanSoftwire: I think we need a "when value is 3 and stairowned is  not 100" test here as well!

RobertSullivanSoftwire (changes requested):

app/services/bulk_upload/sales/year2024/row_parser.rb
RobertSullivanSoftwire: Ah, I see - this rule already exists elsewhere. Ignore this!

spec/services/bulk_upload/sales/year2024/row_parser_spec.rb
kosiakkatrina: I think there is one on line 1004 (?)

spec/services/bulk_upload/sales/year2023/row_parser_spec.rb
kosiakkatrina: Yep, I think the same thing is tested below, so removed this and tested the case where we do allow it

app/services/bulk_upload/sales/year2023/row_parser.rb
kosiakkatrina: Ah yes, so, the ACs are correct, they got changed at some point, there's more context on the last comment

kosiakkatrina: Ah yes, so, the ACs are correct, they got changed at some point there's more context on the last comment

spec/services/bulk_upload/sales/year2024/row_parser_spec.rb
```diff
@@ -1049,7 +1061,7 @@ def purch_id_field
 
         it "populates with correct error message" do
           expect(parser.errors.where(:field_31, category: :soft_validation).first.message).to eql("You told us this person is aged 22 years and retired.")
-          expect(parser.errors.where(:field_31, category: :soft_validation).first.message).to eql("You told us this person is aged 22 years and retired.")
+          expect(parser.errors.where(:field_35, category: :soft_validation).first.message).to eql("You told us this person is aged 22 years and retired.")
```
RobertSullivanSoftwire, line 1109: Good catch!

RobertSullivanSoftwire (approved):

PR #2311, opened 2024-03-13
CLDC-3302 Make mortgageused bulk upload errors only display on relevant field

natdeanlewissoftwire:


app/services/bulk_upload/sales/year2024/row_parser.rb
```diff
@@ -760,7 +760,7 @@ def field_mapping_for_errors
       stairbought: %i[field_87],
       stairowned: %i[field_88],
       socprevten: %i[field_100],
-      mortgageused: %i[field_103 field_117 field_126],
+      mortgageused: [mortgageused_field],
```
RachaelBooth, line 763: Have we got similar issues on other fields? I think I've noticed a similar bug before that we've not picked up.

Doesn't necessarily need to be fixed in this PR, but be good to flag at least (or fix, depends how much time you have!)

RachaelBooth (approved):

app/services/bulk_upload/sales/year2024/row_parser.rb
natdeanlewissoftwire: I'm not aware of any, but possibly! Can you remember what field it was?

RachaelBooth: Not off the top of my head, and it may have been more than one, but price-related and something we ask in the shared ownership / discounted ownership / outright sale sections.

natdeanlewissoftwire: I've just found some, thanks!

natdeanlewissoftwire: Yep I'll make a separate ticket for this, would be quite a lot of dev time (mainly in testing) to do it well as there's a bit of fiddligng to be done for some of the other fields

PR #2313, opened 2024-03-14
CLDC-3313 Update financial sale validations

kosiakkatrina:
- Update `validate_staircasing_mortgage` and `validate_non_staircasing_mortgage` to take into consideration whether the log is social homebuy. If it is, we are taking into account cash discount as well and use it in validation calculations and error messages
- Remove `validate_shared_ownership_deposit` because we don't want to run financial validations against equity for staircasing transactions (we want to run them against stair bought instead) and for non staircasing transactions we are already running the same validation in `validate_non_staircasing_mortgage`
- As these 2 validation trigger on `type` this gets marked as a setup error in BU and prevents log creation. To avoid that, set `skip_bu_setup_error` error type to the errors and add them as regular errors instead of setup errors. This won't block log creation and will only clear non setup fields for these validations
- Deduplicate which fields we add BU errors to depending on the sale type, if sale type is not given, add errors to all fields. Heavily inspired by https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/2311


app/models/form/sales/pages/about_deposit_with_discount.rb
```diff
@@ -14,9 +14,9 @@ def questions
 
   def depends_on
     if form.start_year_after_2024?
-      [{ "is_type_discount?" => true, "stairowned_100?" => @optional }]
+      [{ "social_homebuy?" => true, "stairowned_100?" => @optional }]
```
kosiakkatrina, line 17: Just a rename

app/services/bulk_upload/sales/year2024/row_parser.rb
```diff
@@ -514,7 +514,7 @@ def valid?
 
         question = log.form.get_question(error.attribute, log)
 
-        if question.present? && setup_question?(question)
+        if question.present? && setup_question?(question) && error.type != :skip_bu_setup_error
```
RachaelBooth: Given that we skip adding an error if there is already one on the field, I think this potentially allows us to skip adding a setup error because we've previously added a non-setup error to that field.

I'm not sure if this is getting/can get round the problem with careful ordering, but it seems risky to me to rely on that. (Mainly because I think it would be fairly easy for us to change some ordering or add this attribute to more validations which meant they weren't happening first and thus break things)

I think the way round this is going to involve either adding all the errors and thinking about what we display/how to choose which to display or doing something clever here to prioritise adding things that are setup errors (this has been in the back of my mind recently but I was figuring something for after the start of April), which might want some refinement in either case.

Maybe I'm overcomplicating things and there's a simple solution? Otherwise we might need to shelve the clever idea for a few weeks.

app/models/validations/sales/sale_information_validations.rb
```diff
@@ -107,54 +107,112 @@ def validate_discount_and_value(record)
...
+        end
+      end
+    elsif over_tolerance?(record.deposit, record.stairbought_part_of_value, 1)
+      %i[mortgageused value deposit stairbought type].each do |field|
+        record.errors.add field, :skip_bu_setup_error, message: I18n.t("validations.sale_information.staircasing_mortgage.mortgage_not_used", deposit: record.field_formatted_as_currency("deposit"), stairbought_part_of_value: record.field_formatted_as_currency("stairbought_part_of_value"))
+      end
     end
   end
+
+  def over_tolerance?(expected, actual, tolerance)
```
RachaelBooth, line 236: Is there something like this we use elsewhere already?

app/models/validations/sales/sale_information_validations.rb
```diff
@@ -107,54 +107,112 @@ def validate_discount_and_value(record)
...
         end
       end
     end
   end
 
-  def validate_staircasing_mortgage(record)
-    return unless record.mortgageused && record.value && record.deposit && record.stairbought
-    return unless record.is_staircase?
-    return unless record.saledate && record.form.start_year_after_2024?
+  def add_non_staircasing_non_socialhomebuy_mortgage_errors(record)
```
RachaelBooth: Given the naming I kinda expected this to just add errors (possibly with which ones depending on something) and not also do the check of whether it needs to add any. Not a blocker but worth pondering the name in case anything comes to mind.

app/services/bulk_upload/sales/year2024/row_parser.rb
kosiakkatrina: What if we change the type to `:skip_bu_error` and not add it to the fields at all? That way if there are other errors, they can still be added, and in this specific instance, not displaying the error on the type wouldn't be the worst, cause we wouldn't clear type anyway and it's a compound validation so it would display on all the other relevant fields

app/models/validations/sales/sale_information_validations.rb
kosiakkatrina: Yep, updated now!

app/models/validations/sales/sale_information_validations.rb
kosiakkatrina: Does it read better now?

app/services/bulk_upload/sales/year2024/row_parser.rb
RachaelBooth: Ooh, I like that - yeah, I think that solves my concern.

kosiakkatrina: Great! Updated it now!

RachaelBooth (approved):
LGTM

PR #2314, opened 2024-03-14
CLDC-3308: Enforce that the declaration has been answered before moving onto other sections in lettings logs

RachaelBooth:


spec/models/lettings_log_spec.rb
```diff
@@ -3507,5 +3507,45 @@ def previous_postcode_fields
       end
     end
   end
+
+  describe "#non_location_setup_questions_completed" do
+    before do
```
RachaelBooth, line 3512: These things undo some general setup applying to all tests in this file (which sets a date in the past and the current lettings form to be the fake 2021 one).

What we should do is remove that and update the tests accordingly - only fixing dates when the year matters. But that's a more substantial task than I can justify doing alongside this given the deadline, so I think we have to accept it as tech debt.

RachaelBooth: N.B. I've previously raised a tech debt ticket for avoiding the general problem: https://dluhcdigital.atlassian.net/browse/CLDC-3289 - we'll have to try and prioritise it after the forms stuff is done.

kosiakkatrina (approved):

PR #2315, opened 2024-03-14
CLDC-3309: Update privacy notice/declaration error message for 2024

RachaelBooth:


spec/features/form/validations_spec.rb
```diff
@@ -199,14 +187,4 @@ def fill_in_date(lettings_log_id, question, day, month, year, path)
       end
     end
   end
-
```
RachaelBooth, line 202: This is another case of year problems (because declaration is no longer a special case in question.rb), and for some reason I can't get this test setup to work for current years.

As this file seems to need looking at more generally I think this test isn't worth the time at the moment.

RobertSullivanSoftwire (approved):
Looks good to me!

PR #2316, opened 2024-03-15
CLDC-3318: Add maximum percentage discount validation for bulk uploads

RobertSullivanSoftwire:


config/locales/en.yml
```diff
@@ -211,6 +211,7 @@ en:
         already_added: "You have already added this managing agent"
       merged: "That organisation has already been merged. Select a different organisation."
     not_answered: "You must answer %{question}"
+    range: "%{question} should be between %{min} and %{max}"
```
natdeanlewissoftwire: we've got validations.numeric.within_range that might be usable instead in this case (it has slightly different phrasing though), your call :)

app/services/bulk_upload/sales/year2024/row_parser.rb
```diff
@@ -374,6 +374,15 @@ class BulkUpload::Sales::Year2024::RowParser
             },
             on: :after_log
 
+  validates :field_116,
+            numericality: {
```
natdeanlewissoftwire, line 378: nice! 🛤️

spec/services/bulk_upload/sales/year2024/row_parser_spec.rb
```diff
@@ -725,6 +725,24 @@ def purch_id_field
       end
     end
 
+    describe "#field_116" do # percentage discount
```
natdeanlewissoftwire, line 728: can we add a test for <0% too?

natdeanlewissoftwire (approved):
lgtm!

config/locales/en.yml
RobertSullivanSoftwire: Eh, I'm not going to quibble over should and must - let's get rid of this.

kosiakkatrina:
I don't think this needed to be added separately, it should be getting caught by validate_numeric_min_max (?)

PR #2322, opened 2024-03-19
CLDC-3271 Add address fields as entered in bulk upload to support csv downloads

natdeanlewissoftwire:


kosiakkatrina (approved):
I feel like `_as_entered` might be a bit confusing, because all of these values are entered. Might make more sense to name it something that indicates this value is only applicable to bulk uploaded logs

natdeanlewissoftwire:
The fields names come from the stats team so think it's easier to leave them the same in the codebase to avoid making more namesets and more confusion (due to our already large number of similarly-named-but-different-in-purpose address fields) imo :)

PR #2323, opened 2024-03-19
CLDC-3304: Add bulk upload id to xml exports from 2024

RachaelBooth:


natdeanlewissoftwire (commented):
will conflict with https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/2324 depending on which gets PO reviewed first :)

natdeanlewissoftwire (approved):

RachaelBooth:
> will conflict with #2324 depending on which gets PO reviewed first :)

True, although it should at least be easy to fix!

PR #2326, opened 2024-03-20
CLDC-3305 Correct previous renewal postcodes

kosiakkatrina:


lib/tasks/correct_renewal_postcodes.rake
```diff
@@ -0,0 +1,17 @@
+desc "Update lettings logs renewal previous postcode and la data to be the same as current postcode/la"
+task correct_renewal_postcodes: :environment do
+  LettingsLog.filter_by_year(2023).where(renewal: 1).where("(ppostcode_full != postcode_full) OR (ppostcode_full IS NULL AND postcode_full IS NOT NULL) OR (postcode_full IS NULL AND ppostcode_full IS NOT NULL)").each do |log|
```
RachaelBooth: I think this is right but just to confirm - there's no way for the la data to differ without one of these being satisfied, right?

spec/lib/tasks/correct_renewal_postcodes_spec.rb
```diff
@@ -0,0 +1,117 @@
...
+  describe ":correct_renewal_postcodes", type: :task do
+    subject(:task) { Rake::Task["correct_renewal_postcodes"] }
+
+    before do
+      Rake.application.rake_require("tasks/correct_renewal_postcodes")
+      Rake::Task.define_task(:environment)
+      task.reenable
+    end
+
+    context "when the rake task is run" do
```
RachaelBooth, line 14: This looks like it wants a test for not updating non-renewal logs - that feels like the highest value thing we want to be sure of before running this.

kosiakkatrina: For sure, thanks! Added it now!

lib/tasks/correct_renewal_postcodes.rake
kosiakkatrina: Ooh, good point, there are a few that differ on la only! Updated the query and tests now. Thanks!

RachaelBooth (approved):

PR #2328, opened 2024-03-20
CLDC-3341 Update nationality all csv labels

kosiakkatrina:
Before the CSV label was `"36" => { "name" => "Australia", "synonyms" => "AU" }` now it's just the name


spec/fixtures/files/lettings_log_csv_export_labels_24.csv
```diff
@@ -1,2 +1,2 @@
...
...ds local authority,No,2024-04-01,Affordable Rent,Affordable Rent,Rent to Buy,,No,HIJKLMN,ABCDEFG,Yes,Yes,No,,,,,,,fake address,,London,,NW9 5LL,No,Barnet,E09000003,No,Affordable rent basis,Tenant abandoned property,No,2,House,Purpose built,Yes,3,2024-03-30,,,Yes,2024-03-31,,Don’t know,Yes,Assured Shorthold Tenancy (AST) – Fixed term,,2,,4,,Yes,4,0,0,2,35,,Female,White,Irish,13,,Other,Yes,Partner,,,32,Male,Not seeking work,No,Prefers not to say,Not known,Prefers not to say,Prefers not to say,Yes,Person prefers not to say,Not known,Person prefers not to say,Person prefers not to say,,,,,,,,,,,,,,,,,,,,,Yes – the person is a current or former regular,No – they left up to and including 5 years ago,Yes,No,Yes,Fully wheelchair accessible housing,Yes,No,No,No,No,No,No,Yes,No,No,Yes,No,No,No,No,No,No,No,Less than 1 year,1 year but under 2 years,Loss of tied accommodation,,,Other supported housing,2,No,Yes,TN23 6LZ,Yes,No,Ashford,E07000105,Yes,,Yes,,,,No,No,Yes,No,,Tenant applied directly (no referral or nomination),,Yes,No,Weekly,268,,Universal Credit housing element,Yes,All,,No,Every 2 weeks,,,,,200.0,100.0,,50.0,25.0,40.0,20.0,35.0,17.5,325.0,162.5,,,,Yes,Yes,12.0,6.0,,,,,,,,,,,,,,,,,,,,
```
RachaelBooth: This looks like it's now a 2024 log but with 2023 fields filled (in particular nationality rather than nationality_all, notably as a number rather than words which makes it stand out as a bit odd). Would it be easy enough to change that?

kosiakkatrina: It would have to get the question from the previous form I guess. not sure if we're doing that already somewhere in csv service. I think it would get addressed with whatever changes we decide to do for CSV downloads tho and given it wasn't introduced here, I'd be leaning towards leaving it for now. This specific situation also shouldn't happen cause 2024 logs shouldn't really have nationality filled in

RachaelBooth: Can we make it clearer what's happening though by either having the date be in 2023 or the right field being filled in? Because currently this isn't a real scenario but that means it might be masking what happens in a real scenario (and as of 1st April we do want 2023 logs with nationality filled in to do the right thing here).

kosiakkatrina: Okay, I think I understand the purpose of this test a bit better, I've kept it as is and added a new one instead

RachaelBooth (approved):
LGTM

PR #2329, opened 2024-03-20
CLDC-3317 Add new xml columns for 24/25

natdeanlewissoftwire:


RachaelBooth (approved):
This is another where we might get not-really-merge-conflicts isn't it?

kosiakkatrina (approved):

PR #2335, opened 2024-03-20
CLDC-3343 Check relat fields for people we collect data for

kosiakkatrina:
`multiple_partners?` method was trying to access relat7 fields and above for sales, because hholdcount can be up to 15 and hhmemb = hholdcount + buyers. We only collect data of up to 6 people so if there are more household members than that we should not try to access information for them.


natdeanlewissoftwire (approved):
Ah, thought it might have been related to this, nice fix!

PR #2338, opened 2024-03-21
CLDC-3327 Deduplicate null errors in bulk upload

natdeanlewissoftwire:
Also includes related behaviour for CLDC-3328

do not merge until cldc-3327 and 3328 have both passed PO

### DIFF
--- a/spec/services/bulk_upload/sales/year2024/row_parser_spec.rb
+++ b/spec/services/bulk_upload/sales/year2024/row_parser_spec.rb
@@ -287,7 +287,7 @@ RSpec.describe BulkUpload::Sales::Year2024::RowParser do
         context "when non-setup questions are null" do
           let(:attributes) { setup_section_params.merge({ field_32: "" }) }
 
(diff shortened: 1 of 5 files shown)


spec/services/bulk_upload/sales/year2024/row_parser_spec.rb
```diff
@@ -282,6 +282,26 @@ def purch_id_field
           expect(questions.map(&:id)).to eql([])
         end
       end
+
+      describe "#validate_nulls" do
+        context "when non-setup questions are null" do
+          let(:attributes) { setup_section_params.merge({ field_32: "" }) }
+
+          it "fetches the question's check_answer_label if it exists, otherwise it gets the question's header" do
```
RobertSullivanSoftwire: This reads like it would be better as two different test cases. Is there a reason that doesn't work?

RobertSullivanSoftwire (approved):

RobertSullivanSoftwire (approved):

RobertSullivanSoftwire (approved):
Looks good to me.

PR #2344, opened 2024-03-22
CLDC-3349 Add question number to success message

kosiakkatrina:


spec/requests/form_controller_spec.rb
```diff
@@ -760,25 +760,27 @@
...
+                page: "lead_tenant_age",
                 age1: 20,
-                interruption_page_id: "retirement_value_check",
+                interruption_page_id: "age_lead_tenant_over_retirement_value_check",
               },
             }
           end
 
           before do
+            lettings_log.update!(startdate: Time.zone.local(2023, 4, 1))
```
kosiakkatrina, line 771: Had to update the year, cause we only introduced question numbers in 2023

app/controllers/form_controller.rb
```diff
@@ -13,7 +13,11 @@ def submit_form
       mandatory_questions_with_no_response = mandatory_questions_with_no_response(responses_for_page)
 
       if mandatory_questions_with_no_response.empty? && @log.update(responses_for_page.merge(updated_by: current_user))
-        flash[:notice] = "You have successfully updated #{@page.questions.map(&:check_answer_label).reject { |label| label.to_s.empty? }.first&.downcase}" if previous_interruption_screen_page_id.present?
+        if previous_interruption_screen_page_id.present?
+          updated_question = @page.questions.reject { |question| question.check_answer_label.blank? }.first
```
natdeanlewissoftwire, line 17: are there any soft validation questions that don't have question numbers? Think this'll still be fine in that case but maybe worth regression testing the copy looks right if that's relevant

natdeanlewissoftwire (approved):

app/controllers/form_controller.rb
kosiakkatrina: I think if there are no numbers, it would just have the check answers label

PR #2345, opened 2024-03-22
CLDC-3297: Add assigned to field to xml exports

RobertSullivanSoftwire:


app/services/exports/lettings_log_export_service.rb
```diff
@@ -230,6 +230,9 @@ def apply_cds_transformation(lettings_log, export_mode)
 
       attribute_hash["renttype_detail"] = LettingsLog::RENTTYPE_DETAIL_MAPPING[lettings_log.rent_type] if lettings_log.rent_type.present?
 
+      # field to be added in future
+      attribute_hash["assigned_to"] = nil
+
```
kosiakkatrina, line 235: Does it not get added without this line?

RobertSullivanSoftwire: No - it generates the file by starting with all the properties of the log, then pruning out the ones that aren't in the list. As a result anything that's not a property of the log won't get included!

kosiakkatrina (approved):

PR #2347, opened 2024-03-22
CLDC-3291: Add empty assigned_to field to support csv export

RobertSullivanSoftwire:


app/services/csv/lettings_log_csv_service.rb
```diff
@@ -174,6 +178,8 @@ def value(attribute, log)
         when "labels"
           PERSON_DETAILS.find { |key, _value| key == attribute }[1]["refused_label"]
         end
+      elsif NOT_IMPLEMENTED_FIELDS.include? attribute
+        nil
```
kosiakkatrina, line 182: Do we need this? Wouldn't else return nil (assuming we don't have these values set)

kosiakkatrina (approved):

app/services/csv/lettings_log_csv_service.rb
RobertSullivanSoftwire: At the moment the else block hits a `NoMethodError` with this field (since it's not even an attribute that exists on a log at the moment).

kosiakkatrina (approved):

PR #2349, opened 2024-03-25
CLDC-3350: Add validation for mortgage+deposit for outright sales

RachaelBooth:


app/models/validations/sales/sale_information_validations.rb
```diff
@@ -54,6 +54,20 @@ def validate_discounted_ownership_value(record)
     end
   end
 
+  def validate_outright_sale_value_matches_mortgage_plus_deposit(record)
+    return unless record.saledate && record.form.start_year_after_2024?
+    return unless record.outright_sale?
+    return unless record.mortgage_used? && record.mortgage
```
RachaelBooth, line 60: N.B. If mortgageused is "no" or "don't know", we don't ask mortgage or deposit so no validation is needed.

spec/models/validations/sales/sale_information_validations_spec.rb
```diff
@@ -505,6 +505,94 @@
     end
   end
 
+  describe "#validate_outright_sale_value_matches_mortgage_plus_deposit" do
+    context "with a 2024 outright sale log" do
+      let(:record) { FactoryBot.build(:sales_log, value: 300_000, ownershipsch: 3, saledate: Time.zone.local(2024, 5, 1)) }
+
+      context "when a mortgage is used" do
```
RachaelBooth, line 512: I've not got tests for not applying when this is false, because in that case mortgage and deposit aren't routed to.

kosiakkatrina: Would it make sense to add a test for BU? What would happen if the mortgage used is no in BU but the mortgage values are still given? I don't think it would trigger, but might be worth double checking

RachaelBooth: The mortgage value ends up being cleared in that case (and this validation doesn't trigger, which it shouldn't).

kosiakkatrina (approved):

natdeanlewissoftwire (approved):

PR #2354, opened 2024-03-26
CLDC-3351: Clear deposit in situations where it is no longer derived

RobertSullivanSoftwire:


app/models/derived_variables/sales_log_variables.rb
```diff
@@ -102,6 +115,14 @@ def set_derived_fields!
         mortgage: 0,
       },
     },
+    {
+      conditions: {
+        mortgageused: 3,
+      },
+      derived_values: {
```
RobertSullivanSoftwire: Previously the mortgage value was being left as whatever it was before when switching to "don't know". Possibly this should be nil, though? I'll check.

app/models/derived_variables/sales_log_variables.rb
```diff
@@ -17,7 +17,20 @@ def set_derived_fields!
       self.homonth = hodate.month
       self.hoyear = hodate.year
     end
-    self.deposit = value if outright_sale? && mortgage_not_used?
+
+    if outright_sale? && mortgage_not_used?
+      self.deposit = value
+    elsif outright_sale? && mortgage_use_unknown?
```
RobertSullivanSoftwire: This might not be quite right either, per https://dluhcdigital.atlassian.net/browse/CLDC-3171 - need to resolve that conversation. Either way the current behaviour (when switching to don't know, keep whatever deposit value was previously present) seems wrong!

app/models/derived_variables/sales_log_variables.rb
```diff
@@ -17,7 +17,20 @@ def set_derived_fields!
...
+      self.deposit = nil
+    elsif outright_sale? && mortgageused_changed? && !deposit_changed?
```
RobertSullivanSoftwire: This is the change that fixes the reported issue. Previously the inferred deposit was preserved during this change, and immediately became subject to validation that it might not satisfy.

app/models/derived_variables/sales_log_variables.rb
```diff
@@ -17,7 +17,20 @@ def set_derived_fields!
...
+      # Clear when switching to mortgage used
+      self.deposit = nil
+    elsif ownershipsch_changed?(from: 3) &&
```
RobertSullivanSoftwire: This change can cause the same issue.

kosiakkatrina (approved):

app/models/derived_variables/sales_log_variables.rb
RobertSullivanSoftwire: Looks like it should.

app/models/derived_variables/sales_log_variables.rb
RobertSullivanSoftwire: Confirmed at standup that this approach is preferred.

PR #2362, opened 2024-04-03
CLDC-2670: Improve navigation flow when revisiting questions from CYA page

RobertSullivanSoftwire:


app/components/check_answers_summary_list_card_component.rb
```diff
@@ -28,8 +28,9 @@ def check_answers_card_title(question)
     "Person #{question.check_answers_card_number}"
   end
 
-  def action_href(log, page_id, referrer = "check_answers")
-    send("#{log.model_name.param_key}_#{page_id}_path", log, referrer:)
+  def action_href(question, log)
+    referrer = question.displayed_as_answered?(log) ? "check_answers" : nil
```
RobertSullivanSoftwire: When somebody is selecting an unanswered question, we now send them to the normal flow rather than the CYA flow.

app/models/form/question.rb
```diff
@@ -110,12 +110,16 @@ def displayed_answer_options(log, _current_user = nil)
   end
 
   def action_text(log)
+    displayed_as_answered?(log) ? "Change" : "Answer"
+  end
+
+  def displayed_as_answered?(log)
```
RobertSullivanSoftwire, line 116: This is maybe a bit cowardly - this logic (taken from the forms helper originally) is a bit different from `completed?` below so I left it intact, but I'm not clear on why.

app/controllers/form_controller.rb
```diff
@@ -208,6 +217,24 @@ def successful_redirect_path
     send(redirect_path, @log)
   end
 
+  def redirect_path_to_question(page_to_show, unanswered_pages)
+    remaining_pages = unanswered_pages.excluding(page_to_show)
+    remaining_page_ids = remaining_pages.any? ? remaining_pages.map(&:id).join(",") : nil
+    send("#{@log.class.name.underscore}_#{page_to_show.id}_path", @log, { referrer: "check_answers", unanswered_pages: remaining_page_ids })
+  end
+
+  def pages_requiring_update(previously_visible_empty_page_ids)
```
RobertSullivanSoftwire, line 229: This is really the meat of the change - working out which pages require updates as a result of the submitted change and serving them up instead of going back to the CYA page. I've tried to keep it from intruding too much on the main flow, but it's a little tricky!

app/components/check_answers_summary_list_card_component.rb
kosiakkatrina: Sounds like we only want to route through unanswered question if we click on "Answer"? Unless that's happening already and I'm missing it, I think I'd be more inclined to change the Acs and go through normal flow as it seems is happening here

app/views/form/_check_answers_summary_list.html.erb
```diff
@@ -28,7 +28,7 @@
       <% if @log.collection_period_open_for_editing? %>
         <% row.with_action(
           text: question.action_text(@log),
-          href: action_href(@log, question.page.id, referrer),
+          href: check_answers_href(question, @log),
```
kosiakkatrina: I think "referrer" isn't always "check_answers", I think it might be "interruption_screen" or _maybe_ something duplicate log related too. Is that addressed somewhere?

app/components/check_answers_summary_list_card_component.rb
RobertSullivanSoftwire: I'm a bit confused, sorry - what would be clicking on other than "Answer" in this scenario?

app/views/form/_check_answers_summary_list.html.erb
RobertSullivanSoftwire: Good point - I think I broke this while trying to tidy my work up 🤦

app/components/check_answers_summary_list_card_component.rb
kosiakkatrina: Ah, sorry. So "change" would be the other option. When we change an answer we seem to route through any new questions that changing the answer creates.

When we click on "answer" from CYA (on unanswered question) we seem to go through the normal flow, but if I'm reading the ACs correctly, we want to only route to the unanswered questions if answering from CYA. Let me know if I'm misunderstanding or if my explanation is still confusing

RobertSullivanSoftwire: Ah, I understand the subtlety there now - good point. Before this, was there ever a scenario where the "normal" flow would display already-answered questions? If not, we could just amend the normal flow to skip unanswered questions, which would avoid creating another context a page might be viewed in to consider.

natdeanlewissoftwire (approved):

PR #2366, opened 2024-04-03
CLDC-3156: Update 2024 referral options

RobertSullivanSoftwire:


RobertSullivanSoftwire:
There's an open question here about what to do about 24/5 logs that were created in before this is released and had the values 12 or 13. I'm intending to bring this up at team time tomorrow to confirm (1) what the best way to resolve this is and (2) what the easiest technical implementation of that looks like.

RobertSullivanSoftwire:
There's no test for this file at present - I'm intending to add one shortly.

kosiakkatrina (commented):
There are 4 different versions of referral question (there's also referral_prp, referral_supported_housing and referral_supported_housing_prp), do we need to add this change there as well?

RobertSullivanSoftwire:
Good spot - I've updated those as well, though some of the discrepancies between them are giving me pause - I've raised some questions around that on Slack. I'm going to start writing tests for them as well, but since this ticket is a little urgent I've pushed everything else for review here in the mean time.

kosiakkatrina (approved):

PR #2373, opened 2024-04-08
CLDC-3345 Add and set created_by fields

kosiakkatrina:
This brings back the `created_by` column, it is not set to track the user that made the action of creating the log. We would backdate `created_by` values starting 2023/24 collection period


db/migrate/20240408102550_add_created_by.rb
```diff
@@ -0,0 +1,11 @@
+class AddCreatedBy < ActiveRecord::Migration[7.0]
+  def change
+    change_table :sales_logs do |t|
+      t.references :created_by, class_name: "User"
```
natdeanlewissoftwire, line 4: this was previously created_by_id, are we calling it just created_by intentionally? Or does rails ignore this difference anyway

natdeanlewissoftwire (approved):

db/migrate/20240408102550_add_created_by.rb
kosiakkatrina: I think cause we're using `.reference` to create this instead of `.column`, it actually ends up adding `created_by_id` to schema

PR #2381, opened 2024-04-12
CLDC-2470: Order bulk upload error rows numerically

RachaelBooth:


app/models/bulk_upload_error.rb
```diff
@@ -1,7 +1,7 @@
 class BulkUploadError < ApplicationRecord
   belongs_to :bulk_upload
 
-  scope :order_by_row, -> { order(row: :asc) }
+  scope :order_by_row, -> { order("row::integer ASC") }
```
RachaelBooth, line 4: Arguably we should just make this column be stored as an integer - but that would be quite a lot of work to do without downtime, and I think this should be ok for performance (given we always filter to be within a given upload).

natdeanlewissoftwire (approved):

PR #2382, opened 2024-04-15
Fix current date dependency in household validations spec

natdeanlewissoftwire:


spec/models/validations/household_validations_spec.rb
```diff
@@ -429,10 +429,11 @@
     end
 
     context "with 2024 logs" do
+      let(:log_date) { Time.zone.local(2024, 4, 1) }
+
       before do
-        Timecop.freeze(Time.zone.local(2024, 4, 1))
+        Timecop.freeze(log_date)
```
RachaelBooth, line 435: It sort of annoys me that we need to freeze time here at all (and in fact presumably we don't but it would break the 2024 form closes for new logs?) - that's a bit of a wider thing though.

RachaelBooth (approved):

spec/models/validations/household_validations_spec.rb
natdeanlewissoftwire: Yep, agreed - this is exactly right, rather than leave something to be fixed every year I opted for something that can be duplicated more easily for new collection years. We've got other tests elsewhere that test behaviour for today rather than a specific date in the collection year, so I think overall we should be covered.

PR #2383, opened 2024-04-16
CLDC-3406-remove-validation-on-gender-for-refuge

arfacamble:
remove validation that male cannot have been in refuge

https://dluhcdigital.atlassian.net/browse/CLDC-3406

n.b when creating log for PO review: to get the relevant questions to appear, the log needs to not be renewal on setup


natdeanlewissoftwire (approved):
I think we can we remove the male_refuge bits from en.yml too if you agree?

arfacamble:
> I think we can we remove the male_refuge bits from en.yml too if you agree?

this is a good idea

PR #2384, opened 2024-04-16
CLDC-3405-update-retirement-age-validation

arfacamble:
update retirement soft validation pages and copy in en.yml to reflect changes in english retirement age

https://dluhcdigital.atlassian.net/browse/CLDC-3405

### DIFF
--- a/app/models/log.rb
+++ b/app/models/log.rb
@@ -205,10 +205,6 @@ class Log < ApplicationRecord
   end
 
   (1..8).each do |person_num|
-    define_method("retirement_age_for_person_#{person_num}") do
-      retirement_age_for_person(person_num)
-    end
-
     define_method("plural_gender_for_person_#{person_num}") do
       plural_gender_for_person(person_num)
     end
--- a/spec/models/lettings_log_spec.rb
+++ b/spec/models/lettings_log_spec.rb
@@ -3382,61 +3382,11 @@ RSpec.describe LettingsLog do
     end
   end
 
-  describe "#retirement_age_for_person" do
-    context "when a person gender is Male" do
-      let(:lettings_log) { build(:lettings_log, sex1: "M") }
+  context "when a postcode contains unicode characters" do
+    let(:lettings_log) { build(:lettings_log, postcode_full: "SR81LS\u00A0") }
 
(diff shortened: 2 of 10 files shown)


RachaelBooth:
Is there something in the trigger for the soft validation that needs to change as well?

spec/models/lettings_log_spec.rb
```diff
@@ -3382,64 +3382,6 @@ def previous_postcode_fields
...
-      it "returns nil" do
-        expect(lettings_log.retirement_age_for_person_3).to be_nil
-      end
-
-      it "returns the expected plural" do
-        expect(lettings_log.plural_gender_for_person_3).to be_nil
-      end
-    end
-
-    context "when a postcode contains unicode characters" do
```
RachaelBooth, line 3434: This looks like an unrelated test that shouldn't have been in this context block - we should probably keep it (but in a better place) unless it's already a duplicate.

app/models/log.rb
```diff
@@ -205,10 +205,6 @@ def blank_invalid_non_setup_fields!
   end
 
   (1..8).each do |person_num|
-    define_method("retirement_age_for_person_#{person_num}") do
-      retirement_age_for_person(person_num)
-    end
-
     define_method("plural_gender_for_person_#{person_num}") do
```
RachaelBooth, line 208: Does this get used anywhere else?

spec/models/lettings_log_spec.rb
arfacamble: good spot!
what a strange place to put that test!

app/models/log.rb
arfacamble: I don't believe so, but it's always impossible to tell for sure with rails since someone _can_ do something very strange like `obj.send(%w(retirement age for person 1).join("_"))`
but I've global searched retirement_age_for which now only gets hits in the old form json files and I can't think of any other way it might be getting called that isn't far fetched.
The tests do all pass as well, so you would hope that that covers things?

RachaelBooth: Ah, rails... That implies we probably could remove it but it seems right to be wary.

RachaelBooth (approved):

PR #2389, opened 2024-04-19
Cldc 3114 enable support control of rent periods per organisation

arfacamble:
https://dluhcdigital.atlassian.net/browse/CLDC-3114


app/views/organisations/show.html.erb
arfacamble, line 1: the spec required the data agreement row to no longer be at the end

spec/requests/OrganisationsController/organisations_controller_rent_periods_spec.rb
```diff
@@ -0,0 +1,114 @@
...
+      get new_organisation_path
+    end
+
+    it "displays the rent periods question" do
+      expect(page).to have_content "What are the rent periods for the organisation?"
+    end
+
+    it "the checkboxes for each rent period are checked by default" do
+      checkboxes = page.all "input[type='checkbox'][name='organisation[rent_periods][]']"
+      expect(checkboxes.count).to be > 5
```
arfacamble, line 24: interested in thoughts on this. It kind of looks random to be greater than 5 (which is an arbitrary number) but it's annoying when this type of test fails because the number of options changes by 1 due to something unrelated.

I think it's important to have something, otherwise the next line can pass when there are no checkboxes on the page.

maybe I should just change it to > 0 ? Or maybe it's good as is.

app/controllers/organisations_controller.rb
```diff
@@ -73,21 +73,30 @@ def details
...
+    @organisation = Organisation.new
+    @rent_periods = helpers.rent_periods_with_checked_attr
     render "new", layout: "application"
   end
 
   def create
-    @resource = Organisation.new(org_params)
-    if @resource.save
-      redirect_to organisations_path
+    selected_rent_periods = params.require(:organisation).permit(rent_periods: [])[:rent_periods].compact_blank
```
kosiakkatrina: Would it make sense to extract this as a method?

app/helpers/data_sharing_agreement_helper.rb
```diff
@@ -1,7 +1,7 @@
 module DataSharingAgreementHelper
   def data_sharing_agreement_row(user:, organisation:, summary_list:)
     summary_list.with_row do |row|
-      row.with_key { "Data Sharing Agreement" }
+      row.with_key { "Data protection agreement" }
```
kosiakkatrina: Any reason for renaming it/is that the correct casing?

app/helpers/organisations_helper.rb
```diff
@@ -37,4 +38,14 @@ def organisation_name_row(user:, organisation:, summary_list:)
...
+
+  def rent_periods_with_checked_attr(checked_periods: nil)
+    all_rent_periods = RentPeriod.rent_period_mappings
+    rent_periods = all_rent_periods.map do |code, period|
+      period_copy = period.clone
+      period_copy[:checked] = true if checked_periods.nil? || checked_periods.include?(code)
+      [code, period_copy]
+    end
+    rent_periods.to_h
+  end
```
kosiakkatrina: Would this do the same thing?
```
  def rent_periods_with_checked_attr(checked_periods: nil)
    all_rent_periods = RentPeriod.rent_period_mappings
    rent_periods = all_rent_periods.each_with_object({}) do |(code, period), hash|
      hash[code] = period.merge(checked: checked_periods.nil? || checked_periods.include?(code))
    end
  end
  ```

File: app/controllers/organisations_controller.rb [hunk 5]
```diff
@@ -94,6 +94,8 @@ def create
 
   def edit
     if current_user.data_coordinator? || current_user.support?
+      checked_periods = @organisation.organisation_rent_periods.map(&:rent_period).map(&:to_s)
```
kosiakkatrina: I _think_ pluck(:rent_periods) might be more efficient

app/controllers/organisations_controller.rb
```diff
@@ -133,7 +134,17 @@ def update
         else
           flash[:notice] = I18n.t("organisation.updated")
         end
+        existing_rent_periods = @organisation.organisation_rent_periods.map(&:rent_period)
+        rent_periods_to_create = selected_rent_periods.map(&:to_i) - existing_rent_periods
```
kosiakkatrina: Could we use `find_or_create_by` when creating the new periods instead?

kosiakkatrina (commented):
So I was under the impression that if no rent periods are selected then all the rent periods apply, is that not the case?
image
I'd expect this to say "All rent periods" or something

Also, if a rent period was used but then it gets removed as part of the edit, what happens to the logs that used a that rent period?

app/helpers/data_sharing_agreement_helper.rb
arfacamble: It's just what is in the designs. I have messaged the higher ups to double check.

app/controllers/organisations_controller.rb
arfacamble: happy to.

app/helpers/organisations_helper.rb
arfacamble: I was looking for an enumerator that returned a hash. thanks

arfacamble: then I htought I would try with inject because that sounds even more like what we're doing and the linter told me to do what you did and explained why. all good learning.

app/controllers/organisations_controller.rb
arfacamble: very good yes. pretty sure you're right

app/controllers/organisations_controller.rb
arfacamble: We probably could but I'm not really concerned about that.
I put a uniqueness validation in the model anyway, so I could just call create on them all and let the duplicate inserts fail quietly.
I'm more slightly annoyed to be getting things from the db before doing this anyway, which feels like it should be unnecessary. I could just pass all values through from the form in a hidden field and and delete all `OrganisationRentPeriod.where(org: @org, rent_period: [unchecked_vals])`
What are your thoughts on that?

arfacamble:
> So I was under the impression that if no rent periods are selected then all the rent periods apply, is that not the case? image I'd expect this to say "All rent periods" or something
> 
> Also, if a rent period was used but then it gets removed as part of the edit, what happens to the logs that used a that rent period?

This has changed yes.
In organisation.rb:rent_period_labels you can see where this was the situation and is now different. Confirmed by Sam Ivko that there are such things as sales only orgs, so any that accept all rent periods should have OrganisationRentPeriod instances for all valid answers.
I need to write a script to run when this is merged to find all orgs with no OrganisationRentPeriods associated and create all of them.

The question on validations is a very good one that I had not considered. I will put this problem to the higher ups at the same time as discussing any potential edge cases around this script.

spec/models/organisation_spec.rb
```diff
@@ -242,6 +243,7 @@
 
   describe "scopes" do
     before do
+      described_class.destroy_all
```
kosiakkatrina: How come we're now doing this here?

spec/models/lettings_log_derived_fields_spec.rb
```diff
@@ -0,0 +1,1178 @@
+require "rails_helper"
+require "shared/shared_examples_for_derived_fields"
+
+# rubocop:disable RSpec/BeforeAfterAll
+# rubocop:disable RSpec/InstanceVariable
+RSpec.describe LettingsLog, type: :model do
+  before(:context) do
+    owning_organisation = build(:organisation)
+    @log = build(:lettings_log, :startdate_today, owning_organisation:, managing_organisation: owning_organisation)
```
kosiakkatrina: Any reason we're not using let to define log? Looks like defining it this way in specs is an anti-pattern (?)

spec/models/lettings_log_derived_fields_spec.rb
```diff
@@ -0,0 +1,1178 @@
...
+  it "derives shortfall_known when tshortfall is set" do
+    @log.tshortfall = 10
+
+    expect { @log.set_derived_fields! }.to change(@log, :tshortfall_known).to 0
+  end
+
+  describe "deriving has_benefits" do
+    it "correctly derives when the household receives benefits" do
+      benefits_codes = [1, 6, 8, 7]
+      @log.hb = benefits_codes.sample
```
kosiakkatrina: nice

spec/models/lettings_log_derived_fields_spec.rb
arfacamble, line 1: You will notice that the main weird thing that I am doing here is using `before(:context)` and therefore disabling a rubocop warning against doing this.
This means that one LettingsLog is instantiated and then reused throughout all tests in the file (there are some tests where this is not true below as clearing some derived values relies on calling some active record methods that check the value of a given field at the point the model was found in the DB).
using this before context then means the log has to be referred to with an instance variable to make it available in tests.

I'm doing this since I read somewhere that factories can even be quite time consuming to build (although the difference between create and build is much larger) but I intend to do some of my own testing now as to how longs tests take to run under conditions x, y, z

spec/models/lettings_log_derived_fields_spec.rb
```diff
@@ -0,0 +1,1178 @@
...
+    end
+
+    it "correctly derives when the household does not receive benefits" do
+      no_benefits_codes = [9, 3, 10, 1, 4]
+      @log.hb = no_benefits_codes.sample
+      @log.set_derived_fields!
+
+      expect(@log.has_benefits).to be 0
+    end
+  end
```
arfacamble: how do we feel about stuff like this? There's a gentle smell to having any element of randomness in tests, but I feel like this finds the right balance between being concise and comprehensive

spec/models/lettings_log_spec.rb
```diff
@@ -250,1841 +251,236 @@
...
+    end
 
-          context "when rent_type is Affordable Rent" do
-            let(:rent_type) { 1 }
+    it "correctly derives and saves day, month, year from start date" do # weird test, am I missing something?
+      record_from_db = described_class.find(lettings_log.id)
+      expect(record_from_db["startdate"].day).to eq(10)
+      expect(record_from_db["startdate"].month).to eq(10)
+      expect(record_from_db["startdate"].year).to eq(2021)
+    end
```
arfacamble: sanity check please. These tests look silly to me, we save a date in the db, we call it back out, we can do standard ruby date things to it.
Can I delete?

spec/models/lettings_log_derived_fields_spec.rb
kosiakkatrina: Would you be able to benchmark these tests and see how much time we're actually saving? Cause I think I'd prefer using let if it doesn't make much difference in time

arfacamble: It was always my intention to benchmark, annoyingly with lots of weird behaviour in tests yesterday I didn't get to do that. I have now benchmarked this file against a refactor doing it the normal way and running them both ten times then it is faster, but only by 4% which I think is ABSOLUTELY NOT WORTH IT

spec/models/lettings_log_derived_fields_spec.rb
arfacamble: this was required by doing before(:context) but I got rid of that now

spec/models/organisation_spec.rb
arfacamble: I should be able to remove this now. This was related to what I have been talking about that I keep getting organisations floating around in the DB that aren't meant to be there

spec/requests/OrganisationsController/organisations_controller_rent_periods_spec.rb
kosiakkatrina: How likely would we add new rent periods?

arfacamble: in what time frame?
low but not zero.
is this test currently doing its job and will it fail in future for good reason or no reason?

part of the reason why I've done this is the real expectation is the next line on 25, but any time you check all of something meet a condition the test can fail to be testing what it's meant to if there are none of that thing on the page because something earlier in the test setup gets broken at some point in the future so the test silently passes despite not doing what it's meant to

kosiakkatrina (approved):

PR #2394, opened 2024-04-29
CLDC-3258 Take hint into account in select input

kosiakkatrina:
## What is the issue?
If there are 2 (or more) matches in accessible auto select with the same name (text value) the same option gets displayed in the matches twice (or more times)

## Why is it happening?
Suggestion method is matching on name because that's the only thing we get. So when there are 2 matches with the same name, the first one would get picked and displayed twice in the accessible autoselect. I'm pretty sure it's due to how the accessible-autoselect package works.

## Solution
This PR updates the "name" to include the hint text as well. As the issue was discovered when matching user names in the filters, the hint text is relevant there as it's email, and would therefore be unique. This also means that we can now search by email in the filter box which didn't work in the past.

## Notes
This also however means that for schemes we would be able to search by the hint text (client group) too, which might not be the worst thing, but if it is we could make this change user specific

I don't think we have hints anywhere else.

It doesn't feel like the most elegant solution as I'd prefer sending the actual value around (which we have in the select) but I don't think we have a choice with the package we're using (or it wasn't obvious to me)

### DIFF
--- a/app/frontend/modules/search.js
+++ b/app/frontend/modules/search.js
@@ -110,7 +110,7 @@ export const sort = (query, options) => {
 export const suggestion = (value, options) => {
   const option = options.find((o) => o.name === value)
   if (option) {
-    const html = option.append ? `<span class="autocomplete__option__append">${value}</span> <span>${option.append}</span>` : `<span>${value}</span>`
+    const html = option.append ? `<span class="autocomplete__option__append">${option.text}</span> <span>${option.append}</span>` : `<span>${option.text}</span>`
     return option.hint ? `${html}<div class="autocomplete__option__hint">${option.hint}</div>` : html
   } else {
     return '<span>No results found</span>'
@@ -119,10 +119,15 @@ export const suggestion = (value, options) => {
 
 export const enhanceOption = (option) => {
   return {
-    name: option.label,
+    text: option.text,
+    name: getSearchableName(option),
     synonyms: (option.getAttribute('data-synonyms') ? option.getAttribute('data-synonyms').split('|') : []),
     append: option.getAttribute('data-append'),
     hint: option.getAttribute('data-hint'),
     boost: parseFloat(option.getAttribute('data-boost')) || 1
   }
+}
+
+export const getSearchableName = (option) => {
+  return option.getAttribute('data-hint') ? option.text + ' ' + option.getAttribute('data-hint') : option.text
 }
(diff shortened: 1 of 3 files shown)


RachaelBooth (approved):
Makes sense to me

app/frontend/modules/search.js
```diff
@@ -110,7 +110,7 @@ export const sort = (query, options) => {
 export const suggestion = (value, options) => {
   const option = options.find((o) => o.name === value)
   if (option) {
-    const html = option.append ? `<span class="autocomplete__option__append">${value}</span> <span>${option.append}</span>` : `<span>${value}</span>`
+    const html = option.append ? `<span class="autocomplete__option__append">${option.text}</span> <span>${option.append}</span>` : `<span>${option.text}</span>`
```
arfacamble, line 113: this line is appending options that meet the search to the list of suggestions in the dropdown?
I'm interested in what the value of `option.append` is anyway since I don't know how the existing code works.
What is value? I guess options an array of things that match the search in the format of the kinds of thing returned by the enhanceOption method below?
is append a hint text?

app/frontend/modules/search.js
```diff
@@ -119,10 +119,15 @@ export const suggestion = (value, options) => {
...
     synonyms: (option.getAttribute('data-synonyms') ? option.getAttribute('data-synonyms').split('|') : []),
     append: option.getAttribute('data-append'),
     hint: option.getAttribute('data-hint'),
     boost: parseFloat(option.getAttribute('data-boost')) || 1
   }
 }
+
+export const getSearchableName = (option) => {
+  return option.getAttribute('data-hint') ? option.text + ' ' + option.getAttribute('data-hint') : option.text
+}
```
arfacamble, line 133: how come we're making stuff out of html elements here (that's what we're doing right?)

arfacamble (commented):
Hi I assume I am just going to approve, but I could do with a bit more context on how this code is working first.

app/frontend/modules/search.js
kosiakkatrina: Kinda, we're using stimulus controller and that's how we pass the data through to the javascript layer and we're just sticking things together here. We could contract this in our ruby code, but I feel like it's not that complex

app/frontend/modules/search.js
kosiakkatrina: We get append values from `answer_option_append` it's currently only relevant for schemes, it just displays how many completed/incomplete locations the scheme has. Purely for displaying the info, I don't think it's functional, different from hint text

PR #2396, opened 2024-04-29
CLDC-3412 set derived? on mortgage amount

kosiakkatrina:
## What's the issue?
When pressing confirm and continue on `DepositAndMortgageValueCheck` soft validation, it does not get marked confirmed and still appears on the CYA page.

## Why did it happen?
This happens when mortgage is not used. In those cases we set the `mortgage` to 0. However when we try to set `deposit_and_mortgage_value_check` this sequence of events occurs:
1. `mortgage` gets cleared in the first part of `reset_free_user_input_questions_if_not_routed` method, because it's not routed to
2. `deposit_and_mortgage_value_check` gets reset in the second part of `reset_free_user_input_questions_if_not_routed` method, as `mortgage` is now nil and `deposit_and_mortgage_value_check` soft validation no longer triggers/is not routed to
3. `mortgage` is set to 0 again in `sales_log_variables` and `deposit_and_mortgage_value_check` is routed to again, but is not answered

This is likely to be an issue in other places where we use mortgage to surface a soft validation, even if mortgage is not used.

## Solution

This change prevents `mortgage` from being cleared in the first place if the mortgage is not used

### DIFF
--- a/app/models/form/sales/questions/mortgage_amount.rb
+++ b/app/models/form/sales/questions/mortgage_amount.rb
@@ -20,6 +20,6 @@ class Form::Sales::Questions::MortgageAmount < ::Form::Question
   }.freeze
 
   def derived?(log)
-    log && log.mortgage_not_used?
+    log&.mortgage_not_used?
   end
 end
--- a/app/models/form/sales/questions/mortgage_amount.rb
+++ b/app/models/form/sales/questions/mortgage_amount.rb
@@ -18,4 +18,8 @@ class Form::Sales::Questions::MortgageAmount < ::Form::Question
     2023 => { 1 => 91, 2 => 104, 3 => 112 },
     2024 => { 1 => 92, 2 => 105, 3 => 113 },
   }.freeze
+
+  def derived?(log)
+    log && log.mortgage_not_used?
+  end
 end
--- a/spec/models/form/sales/questions/mortgage_amount_spec.rb
+++ b/spec/models/form/sales/questions/mortgage_amount_spec.rb
@@ -48,18 +48,18 @@ RSpec.describe Form::Sales::Questions::MortgageAmount, type: :model do
   end
 
   context "when the mortgage is not used" do
-    let(:log) { create(:sales_log, :completed, mortgageused: 2, deposit: nil) }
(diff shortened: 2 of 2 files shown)


RachaelBooth (approved):

app/models/form/sales/questions/mortgage_amount.rb
```diff
@@ -18,4 +18,8 @@ def initialize(id, hsh, subsection, ownershipsch:)
     2023 => { 1 => 91, 2 => 104, 3 => 112 },
     2024 => { 1 => 92, 2 => 105, 3 => 113 },
   }.freeze
+
+  def derived?(log)
+    log && log.mortgage_not_used?
```
arfacamble: is this the same as
```suggestion
    log&.mortgage_not_used?
```

spec/models/form/sales/questions/mortgage_amount_spec.rb
```diff
@@ -46,4 +46,20 @@
...
+    end
+  end
+
+  context "when the mortgage is used" do
+    let(:log) { create(:sales_log, :completed, mortgageused: 1) }
+
+    it "is marked as derived" do
+      expect(question.derived?(log)).to be false
+    end
+  end
```
arfacamble: is there a reason these assertions don't follow the same format as the one above you changed (which looks really nicely readable), (i.e. expect(q).to be_derived)

spec/models/form/sales/questions/mortgage_amount_spec.rb
```diff
@@ -46,4 +46,20 @@
   it "has correct min" do
     expect(question.min).to be(1)
   end
+
+  context "when the mortgage is not used" do
+    let(:log) { create(:sales_log, :completed, mortgageused: 2, deposit: nil) }
```
arfacamble:
```suggestion
    let(:log) { build(:sales_log, :completed, mortgageused: 2, deposit: nil) }
```
It is my understanding that db actions are often the most expensive actions by time in rails. I believe we should avoid committing things to db in tests where it isn't necessary.

arfacamble (dismissed):
The only thing I feel strongly about is the create vs build, I endorse ignoring the other two comments if you see fit.

PR #2399, opened 2024-05-01
CLDC-3426: Ensure outright sale section can be completed with mortgage used unknown

RachaelBooth:
Previously while we didn't ask the deposit amount question when mortgage used was set to unknown we still required it to be answered for the section to be marked as complete. (This was also true when mortgage used was false, but in that case we infer the deposit as matching the value and thus we could complete the section)

This PR;
* Updates the derivations to ensure consistency around mortgage used = don't know
* Updates the deposit question to ensure it does not end up required when it shouldn't be


app/models/form/sales/questions/deposit_amount.rb
```diff
@@ -16,11 +16,7 @@ def initialize(id, hsh, subsection, ownershipsch:, optional:)
   end
 
   def derived?(log)
-    log.outright_sale? && log.mortgage_not_used?
-  end
-
-  def selected_answer_option_is_derived?(_log)
-    true
```
RachaelBooth, line 23: This was forcing the question to always be treated as applicable (and thus need an answer for the subsection to count as complete) even when it is derived and not routed to.

Which given the naming of the function is a weird thing that we should look at further - but this should be enough to sort the existing support ticket.

For this problem, as we weren't inferring a value when mortgage use was unknown this caused a log to not be completable. Sam has confirmed that we want deposit to be nil in that case, so am removing this (there doesn't seem to be a need for this override).

spec/models/sales_log_spec.rb
```diff
@@ -538,6 +538,11 @@
   describe "derived variables" do
     let(:sales_log) { create(:sales_log, :completed) }
 
+    before do
+      Timecop.return
+      Singleton.__init__(FormHandler)
+    end
```
RachaelBooth, line 544: Currently there is a time fix on the entire test file. That's too big to sort in this ticket, hence I've just taken this section.

spec/models/sales_log_derived_fields_spec.rb
```diff
@@ -0,0 +1,164 @@
+require "rails_helper"
+require "shared/shared_examples_for_derived_fields"
+
+RSpec.describe SalesLog, type: :model do
```
RachaelBooth, line 4: This should be the same set of tests as have been removed from `sales_log_spec` (with adjustments where I've actually changed the behaviour in this PR).

I had to make some adjustments in order to have them work without a date fixed in the past (which I needed for the tests relevant to this PR), and it ended up making sense to pull them out and adjust to avoid being too database heavy / having fixed time issues / relying on unspoken properties from the factory. This means they are now explicitly testing the `set_derived_fields!` function rather than it's actions when we save an update (which makes sense because this set of tests is clearly designed as unit tests).

As a bonus, this set runs significantly faster (on my machine at least).

spec/helpers/tasklist_helper_spec.rb
```diff
@@ -83,7 +83,7 @@
 
   describe "with sales" do
     let(:empty_sales_log) { create(:sales_log, owning_organisation: nil) }
-    let(:completed_sales_log) { build(:sales_log, :completed) }
+    let(:completed_sales_log) { create(:sales_log, :completed) }
```
RachaelBooth, line 86: Annoyingly this one did start needing to be create to pass.

spec/models/sales_log_derived_fields_spec.rb
arfacamble: super! yes I think I wrote those old tests. this is much nicer building and then calling a method in memory. I'm not surprised it should be much faster. Every old test was making a db call for insert and then update as well.

arfacamble (approved):

PR #2403, opened 2024-05-07
CLDC-3438 Update log csv download flow

kosiakkatrina:
We want to prevent downloading logs for multiple years, we also want to display the filters and allow users to review them before downloading the CSVs of logs.
If no year (or multiple years) is selected in the filters, we route to the collection year question when attempting to download a logs CSV:
image

We also display CYA style filter summary on the CSV download page:
image

TODO: 
- [x] Update cancel routing for years filter
- [x] Make sure this works on per org pages


app/controllers/organisations_controller.rb
```diff
@@ -188,7 +190,7 @@ def download_sales_csv
     unpaginated_filtered_logs = filter_manager.filtered_logs(organisation_logs, search_term, session_filters)
     codes_only = params.require(:codes_only) == "true"
 
-    render "logs/download_csv", locals: { search_term:, count: unpaginated_filtered_logs.size, post_path: sales_logs_email_csv_organisation_path, codes_only: }
+    render "logs/download_csv", locals: { search_term:, count: unpaginated_filtered_logs.size, post_path: sales_logs_email_csv_organisation_path, codes_only:, session_filters:, filter_type: "sales_logs" }
```
RachaelBooth: Does this one also need a redirect if not filtered to a single year first?

app/helpers/filters_helper.rb
```diff
@@ -157,6 +165,51 @@ def applied_filters_count(filter_type)
...
+
+  def check_your_answers_sales_filters_list(session_filters)
+    [
+      { id: "years", label: "Collection year", value: formatted_years_filter(session_filters) },
+      { id: "status", label: "Status", value: formatted_status_filter(session_filters) },
+      { id: "assigned_to", label: "Assigned to", value: formatted_assigned_to_filter(session_filters) },
+      { id: "owned_by", label: "Owned by", value: formatted_owned_by_filter(session_filters) },
+      { id: "managed_by", label: "Managed by", value: formatted_managed_by_filter(session_filters) },
+    ]
+  end
```
RachaelBooth, line 177: Are the lists of available filters for lettings and sales tied closely together elsewhere or defined fairly separately? 

I'm thinking that these are almost the same so it's tempting to de-duplicate a bit, but only if they are already fairly tied to being very similar.

app/helpers/filters_helper.rb
```diff
@@ -184,4 +237,48 @@ def filters_count(filters)
...
+    return "You" if session_filters["assigned_to"].include?("you")
+
+    user = User.find(session_filters["user"].first)
+    "#{user.name} (#{user.email})"
+  end
+
+  def formatted_owned_by_filter(session_filters)
+    return "All" if params["id"].blank? && (session_filters["owning_organisation"].blank? || session_filters["owning_organisation"]&.include?("all"))
+
+    session_org_id = session_filters["owning_organisation"].is_a?(Array) ? session_filters["owning_organisation"].first : session_filters["owning_organisation"]
```
RachaelBooth: What makes this sometimes an array? (I don't think I've touched filters much)

RachaelBooth (commented):
I only got about half way through this before the end of the day, sorry!

app/helpers/filters_helper.rb
kosiakkatrina: Yeah, at the moment the filter names seem to be getting set at the same place (I think the request would be coming through with the form params from _log_filters.html.erb so would likely be updated together), so I think it's probably a good shout to deduplicate here!

app/helpers/filters_helper.rb
kosiakkatrina: You ask good questions, I can't reproduce it and the tests don't seem to fail either. I'll remove it for now, it's likely I've added it cause I ran into a case that I ended up removing

app/controllers/organisations_controller.rb
kosiakkatrina: Yup! Nice catch!

RachaelBooth (approved):

PR #2404, opened 2024-05-09
CLDC-3442 Download logs csv per year

kosiakkatrina:
We want to prevent downloading logs for multiple years.
Now sales and lettings csv services take a year parameter and get relevant attributes for the csv based on it.

This builds on top of CLDC-3438.


app/models/form_handler.rb
```diff
@@ -83,6 +83,15 @@ def ordered_lettings_questions_for_all_years
     ordered_questions
   end
 
+  def ordered_questions_for_year(year, type)
```
RachaelBooth, line 64: Do we still need the `_for_all_years` functions above to exist?

spec/jobs/email_csv_job_spec.rb
```diff
@@ -55,12 +55,12 @@
     end
 
     it "creates a LettingsLogCsvService with the correct export type" do
```
RachaelBooth: Should this test name now also mention the year?

spec/services/csv/lettings_log_csv_service_spec.rb
```diff
@@ -78,15 +72,21 @@
...
-        allow(form_handler_mock).to receive(:ordered_lettings_questions_for_all_years).and_return(lettings_form.questions)
-        allow(form_handler_mock).to receive(:lettings_in_crossover_period?).and_return false
+        allow(form_handler_mock).to receive(:ordered_questions_for_year).and_return(lettings_form.questions)
+      end
+
+      it "calls the form handler to get all questions for the specified year" do
+        allow(FormHandler).to receive(:instance).and_return(form_handler_mock)
+        allow(form_handler_mock).to receive(:ordered_questions_for_year).and_return([])
+        service.prepare_csv(LettingsLog.all)
+        expect(form_handler_mock).to have_received(:ordered_questions_for_year).with(2024, "lettings")
```
RachaelBooth: Can this test either specify the year explicitly or use the year variable?

I think it's a bit confusing to rely on the earlier setup having set it to 2024 when that's fairly disconnected and not explicit in a context.

app/services/csv/lettings_log_csv_service.rb
```diff
@@ -329,5 +329,18 @@ def age_not_known?(log, attribute)
...
+      case @year
+      when 2022
+        %w[id status assigned_to is_dpo created_at updated_by updated_at creation_method old_id old_form_id collection_start_year created_by]
+      when 2023
+        %w[id status duplicate_set_id assigned_to is_dpo created_at updated_by updated_at creation_method old_id old_form_id collection_start_year created_by]
+      when 2024
+        %w[id status duplicate_set_id assigned_to is_dpo created_at updated_by updated_at creation_method collection_start_year bulk_upload_id address_line1_as_entered address_line2_as_entered town_or_city_as_entered county_as_entered postcode_full_as_entered la_as_entered created_by]
+      else
+        %w[id status duplicate_set_id assigned_to is_dpo created_at updated_by updated_at creation_method collection_start_year bulk_upload_id address_line1_as_entered address_line2_as_entered town_or_city_as_entered county_as_entered postcode_full_as_entered la_as_entered created_by]
+      end
```
RachaelBooth, line 355: Is there a nice way to avoid this being a long case statement that we (presumably) need to add to every year? (I suspect there may not be)

app/models/form_handler.rb
kosiakkatrina: nope! Thank you! Updated now!

spec/services/csv/lettings_log_csv_service_spec.rb
kosiakkatrina: Ah, yes for sure!

app/services/csv/lettings_log_csv_service.rb
kosiakkatrina: I thought about piecing it together, but it becomes less readable. It shouldn't get larger than this provided we don't decide to increase the number of forms we accommodate at any given time

RachaelBooth (approved):

PR #2406, opened 2024-05-10
CLDC-3347 Show soft validations in support CSVs only

kosiakkatrina:


RachaelBooth (approved):
I realise there's a chain here and it's not in the right order to merge this yet, but still I've looked at it so may as well have the approval for later.

PR #2408, opened 2024-05-13
CLDC 3414 update homepage

arfacamble:
https://dluhcdigital.atlassian.net/browse/CLDC-3414

let me know if it would be useful to add screenshots, but it looks a lot like the designs linked min the ticket!

question for reviewer, in PO it will be necessary for the tester to test the behaviour when not in the crossover period.
To do this, the simplest way I can think of is to just change the `in_crossover_period?` in the presenter to return false. Do you think that's acceptable or too close to my own work?


app/frontend/styles/_data_box.scss
arfacamble, line 1: there is no longer a situation where we show three boxes across

app/helpers/home_helper.rb
arfacamble, line 1: contained lots of methods calling the DB from view helpers. this is icky and is now encapsulated in the homepage presenter class so DB calls are effectively being made at the controller stage

app/views/home/_blue_box_link.html.erb
arfacamble, line 1: for this and file below, there was always previously a grey box with count and link and a blue view all link below, these are now decoupled to allow eg a grid of 4 boxes with two blue links below

app/views/home/_data_box_row_of_two.html.erb
arfacamble, line 1: it could be overkill to wrap the data box partial in further partials for arrangements of those databoxes, but I htink it makes the index view page significantly more readable, ie under this condition show a grid of 4, under this condition show a row of two, etc etc

spec/factories/sales_log.rb
arfacamble, line 1: this addition of :completed2024 trait tot he factories is the only part of this PR that I am actively unheppy with. I can imagine it being possible to do something like a transient trait completed: true which then checks its own startdate and does conditional things. Is that better? More complicated?

spec/fixtures/exports/general_needs_log.csv
arfacamble, line 1: I made the address line 1 the same in lettings and sales factory so that we need fewer address api stubs

spec/models/lettings_log_spec.rb
arfacamble, line 1: these tests started failing at some point, and reading them it looked like they would fail, so I fixed them in the way that seemed obvious but also obviously they _weren't_ failing before so it may be worth reading these changes slightly more carefully to see if it looks like I've misunderstood the purpose of these tests somehow

app/presenters/homepage_presenter.rb
```diff
@@ -0,0 +1,110 @@
+class HomepagePresenter
+  include Rails.application.routes.url_helpers
+
+  attr_reader :current_year_in_progress_lettings_data, :current_year_completed_lettings_data, :current_year_in_progress_sales_data, :current_year_completed_sales_data, :last_year_in_progress_lettings_data, :last_year_completed_lettings_data, :last_year_in_progress_sales_data, :last_year_completed_sales_data, :incomplete_schemes_data
+
+  def initialize(user)
+    @user = user
+    @display_sales = should_display_sales?
+    @in_crossover_period = FormHandler.instance.in_crossover_period?
+    @current_year = FormHandler.instance.current_lettings_form.start_date.year
```
RachaelBooth: It feels like there should be a better way of getting the current collection year than this (just because it's odd to be going via the current lettings form) - have we got a more generic method somewhere?

(N.B. only worth changing if it's very little effort to do so)

app/presenters/homepage_presenter.rb
```diff
@@ -0,0 +1,110 @@
...
+  end
+
+  def display_schemes?
+    !@user.data_provider?
+  end
+
+private
+
+  def subheading_from_year(year)
+    return "AAAAH" if year.nil?
```
RachaelBooth: Can this be a string it would be more reasonable to display if something did go weirdly wrong? (maybe empty or unknown or something like that)

spec/factories/lettings_log.rb
```diff
@@ -102,7 +102,7 @@
       layear { 2 }
       waityear { 7 }
       postcode_known { 1 }
-      postcode_full { Faker::Address.postcode }
+      postcode_full { "SW1A 1AA" }
```
RachaelBooth, line 105: Is there a reason for needing to fix this value?

spec/factories/sales_log.rb
RachaelBooth: That would be quite cool if we could do it - I was thinking that trait had a bit of an assumption about start date (and including :completed means that it might include things that were only relevant in previous years).

Probably not worth holding up getting the homepage change out for - although if you want to have a quick go as clean up for a follow up PR I'd be very up for that (just timebox it to short enough to be part of this work, otherwise we should raise it as tech debt).

spec/models/lettings_log_spec.rb
RachaelBooth: I can't really spot why these changes would have changed whether or not these tests fail, which makes me somewhat suspicious. Do they still end up consistently failing if you revert the changes to this file?

spec/requests/start_controller_spec.rb
```diff
@@ -16,29 +15,350 @@
...
         sign_in user
       end
 
       it "routes user to the home page" do
-        get "/", headers:, params: {}
+        get root_path
         expect(page).to have_content("Welcome back")
       end
 
+      context "when testing the data displayed at the top of the home page" do
```
RachaelBooth: I think there's something off about contexts that read "when testing" - it's a sensible grouping, I just think the description is almost trying to be another describe block rather than a context.

I can't really put my finger on it, and I'm not sure it's a rails thing to nest describe blocks in the same way I would in typescript, so I'm not that fussed - just something to think about in our general test structure I think.

spec/requests/start_controller_spec.rb
```diff
@@ -16,29 +15,350 @@
...
+          let(:user) { create(:user, :data_coordinator) }
+          let(:in_crossover_period) { true }
+
+          before do
+            create(:sales_log, created_by: user)
+            get root_path
+          end
+
+          it "shows the correct links for in progress lettings" do
+            type = "lettings"
```
RachaelBooth: Where these are the same tests with just the type and status variables differing (particularly as you've already pulled them out), you could parameterise the tests and loop over the combinations rather than duplicating code (see e.g. financial_validations_spec - look for `it "does allow charges inside the range when period is #{test_case[:period][:label]}"`

app/presenters/homepage_presenter.rb
arfacamble: yes this exists actually I found it and used it separately elsewhere. will update

app/presenters/homepage_presenter.rb
arfacamble: this shouldn't be here!!!
I think I just wanted to make a particular error go away while UI testing at a certain point. I think this line just doesn't need to exist at all

spec/factories/lettings_log.rb
arfacamble: this is so that a particular URL for API call can be stubbed. This way any logs marked :completed2024 will be fine

spec/requests/start_controller_spec.rb
arfacamble: I agree with this. Looking it up, it looks liek describe and context are actually completely alias.
I was finding it awkward to be writing 'when' or 'and' at the start of these blocks
I will stop doing that and chagne here

spec/requests/start_controller_spec.rb
arfacamble: affirmative

arfacamble:
@RachaelBooth in response to https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/2408#discussion_r1601458869

for some reason it considers that previous comment/thread outdated and DEAD.

I realised while going through review comments why it passed before, the postcode was all random before from faker. then I changed the factory, so now if you search for the postcode of this log, then it is the same as all the others. I feel these tests have only got better by passing for explicit reasons rather than because random things are almost always not the same

spec/features/form/tasklist_page_spec.rb
```diff
@@ -67,7 +67,7 @@
     expect(page).to have_content("This log has not been started.")
   end
 
-  context "when testing completed subsection count" do
+  describe "testing completed subsection count" do
```
RachaelBooth: This should probably just be `describe "completed subsection count" do`

RachaelBooth (approved):
Minor naming comment, otherwise LGTM!

PR #2410, opened 2024-05-13
CLDC-3352 Update sales CSV address order

kosiakkatrina:


app/services/csv/sales_log_csv_service.rb
```diff
@@ -129,31 +131,38 @@ def label_if_boolean_value(value)
...
+        "la" => %w[la la_label],
+        "prevloc" => %w[prevloc prevloc_label],
+        "assigned_to_id" => %w[assigned_to],
+        "owning_organisation_id" => %w[owning_organisation_name],
+        "managing_organisation_id" => %w[managing_organisation_name],
+        "value" => %w[value value_value_check],
+        "mscharge" => %w[mscharge mscharge_value_check],
+      }
+      if @user.support? && @year >= 2024
+        mappings["beds"] = ORDERED_ADDRESS_FIELDS + %w[beds]
```
RachaelBooth: This feels like a slightly odd use of mappings, since the others here are about mapping to related fields rather than inserting unrelated ones.

Largely I guess because it feels like it goes a bit wrong if in future "beds" is no longer the next question - is there a way to make it a mapping based on an address field? (That may be silly if it makes the later code substantially more complicated, so I'm not too bothered if it doesn't work)

arfacamble: I agree that I am surprised by this.
The ticket specifies that they should be placed after the existing address fields, but then it says put it directly after/before {field that doesn't sound addressy to me}
I think I'm missing some context here as to changes since I left the project of what info we take from the user, what we get back from the address API and what we display in the csv export.

app/services/csv/sales_log_csv_service.rb
```diff
@@ -129,31 +131,38 @@ def label_if_boolean_value(value)
       return "No" if value == false
     end
```
arfacamble, line 133: In general I am wondering if there is value in quite heavily commenting the CSV export services. Almost everything they do is quite arbitrary, so I think it may reduce developer time very significantly in the future, particularly when it's being worked on by people with little to no context, to have a comment such as:
```
After getting all the question ids in order attribute_mappings are used to change the CSV headings while keeping the ordering from the form, for example adding derived fields nex tot he field they are derived from
```

app/services/csv/sales_log_csv_service.rb
arfacamble, line 1: minor comment that is not in this PR and probably nothing to do with you but can I request the condition on line 104 is made consistent with the conditions above
`elsif PERSON_DETAILS.any? { |key, _value| key == attribute }`
`elsif PERSON_DETAILS.key? attribute`

app/services/csv/sales_log_csv_service.rb
kosiakkatrina: Good point! Updated it now, let mw know what you think

app/services/csv/sales_log_csv_service.rb
kosiakkatrina: I'm not the biggest fan of comments, unless there's no other way to make the code readable, but there are probably some nicely named methods we can extract here etc.

RachaelBooth (approved):

PR #2413, opened 2024-05-14
CLDC-3367 Rename sales support CSV fields

kosiakkatrina:


spec/services/csv/sales_log_csv_service_spec.rb
```diff
@@ -52,7 +52,7 @@
   end
 
   it "returns a csv with headers" do
-    expect(csv.first.first).to eq "id"
```
RachaelBooth, line 55: Do we want an explicit test around differing headers for support / non-support? (It will happen I guess with the check that the csv matches the expected file)

RachaelBooth (approved):

spec/services/csv/sales_log_csv_service_spec.rb
kosiakkatrina: Hmm, yeah, I feel like it's implicitly tested, so it might not be the most useful thing to check (?)

PR #2421, opened 2024-05-16
CLDC-3289 Avoid freezing time in tests (part 1)

kosiakkatrina:
There are a few more tests that can be refactored, but I'd like to get some feedback on the approach and merge these changes in smaller chunks to reduce interference with any other test updates that might be going on!

### DIFF
--- a/spec/components/check_answers_summary_list_card_component_spec.rb
+++ b/spec/components/check_answers_summary_list_card_component_spec.rb
@@ -5,103 +5,68 @@ RSpec.describe CheckAnswersSummaryListCardComponent, type: :component do
 
   let(:rendered) { render_inline(component) }
 
+  let(:user) { create(:user) }
+  let(:log) { create(:lettings_log, :completed, age2: 99, retirement_value_check: 1) }
+  let(:subsection_id) { "household_characteristics" }
+  let(:subsection) { log.form.get_subsection(subsection_id) }
+  let(:questions) { subsection.applicable_questions(log) }
+
+  it "renders a summary list card including question numbers for the answers to those questions" do
+    expect(rendered).to have_content(questions.first.answer_label(log))
+    expect(rendered).to match(/Q\d+ - Lead tenant’s age/)
+  end
+
+  it "applicable questions doesn't return questions that are hidden in check answers" do
+    expect(component.applicable_questions.map(&:id).include?("retirement_value_check")).to eq(false)
+  end
+
+  it "has the correct answer label for a question" do
+    sex1_question = questions.find { |q| q.id == "sex1" }
+    expect(component.get_answer_label(sex1_question)).to eq("Female")
+  end
+
+  context "when log was created via a bulk upload and has an unanswered question" do
+    let(:bulk_upload) { create(:bulk_upload) }
+    let(:log) { create(:lettings_log, :in_progress, creation_method: "bulk upload", age2: 99, bulk_upload:) }
+
+    it "displays tweaked copy in red" do
+      expect(rendered).to have_selector("span", class: "app-!-colour-red", text: "You still need to answer this question")
+    end
+  end
+
+  context "when log was imported with a bulk upload creation method, without bulk upload id and has an unanswered question" do
+    let(:log) { create(:lettings_log, :in_progress, creation_method: "bulk upload", age2: 99, bulk_upload_id: nil) }
+
+    it "displays tweaked copy in red" do
+      expect(rendered).not_to have_selector("span", class: "app-!-colour-red", text: "You still need to answer this question")
+    end
+  end
+
+  context "when log was not created via a bulk upload and has an unanswered question" do
+    it "displays normal copy with muted colour " do
+      expect(rendered).to have_selector("span", class: "app-!-colour-muted", text: "You didn’t answer this question")
+    end
+  end
+
+  context "when log was created via a bulk upload and has an unanswered optional question" do
+    let(:subsection_id) { "setup" }
+    let(:bulk_upload) { create(:bulk_upload) }
+    let(:log) { create(:lettings_log, :completed, creation_method: "bulk upload", tenancycode: nil, bulk_upload:) }
+
+    it "displays tweaked copy in red" do
+      expect(rendered).to have_selector("span", class: "app-!-colour-muted", text: "You didn’t answer this question")
+      expect(rendered).not_to have_selector("span", class: "app-!-colour-red", text: "You still need to answer this question")
+    end
+  end
+
   context "when before 23/24 collection" do
-    before do
-      Timecop.freeze(Time.zone.local(2023, 1, 10))
-      Singleton.__init__(FormHandler)
-    end
-
-    after do
-      Timecop.return
-    end
-
     context "when given a set of questions" do
-      let(:user) { build(:user) }
       let(:log) { build(:lettings_log, :completed, age2: 99, startdate: Time.zone.local(2021, 5, 1)) }
-      let(:subsection_id) { "household_characteristics" }
-      let(:subsection) { log.form.get_subsection(subsection_id) }
-      let(:questions) { subsection.applicable_questions(log) }
 
       it "renders a summary list card without question numbers for the answers to those questions" do
         expect(rendered).to have_content(questions.first.answer_label(log))
         expect(rendered).to have_content("Lead tenant’s age")
-        expect(rendered).not_to have_content("Q32 - Lead tenant’s age")
-      end
-
-      it "applicable questions doesn't return questions that are hidden in check answers" do
-        expect(component.applicable_questions.map(&:id).include?("retirement_value_check")).to eq(false)
-      end
-
-      it "has the correct answer label for a question" do
-        sex1_question = questions[2]
-        expect(component.get_answer_label(sex1_question)).to eq("Female")
-      end
-
-      context "when log was created via a bulk upload and has an unanswered question" do
-        subject(:component) { described_class.new(questions:, log:, user:) }
-
-        let(:bulk_upload) { create(:bulk_upload) }
-        let(:log) { build(:lettings_log, :in_progress, creation_method: "bulk upload", age2: 99, startdate: Time.zone.local(2021, 5, 1), bulk_upload:) }
-
-        it "displays tweaked copy in red" do
-          expect(rendered).to have_selector("span", class: "app-!-colour-red", text: "You still need to answer this question")
-        end
-      end
-
-      context "when log was imported with a bulk upload creation method, without bulk upload id and has an unanswered question" do
-        subject(:component) { described_class.new(questions:, log:, user:) }
-
-        let(:log) { build(:lettings_log, :in_progress, creation_method: "bulk upload", age2: 99, startdate: Time.zone.local(2021, 5, 1), bulk_upload_id: nil) }
-
-        it "displays tweaked copy in red" do
-          expect(rendered).not_to have_selector("span", class: "app-!-colour-red", text: "You still need to answer this question")
-        end
-      end
-
-      context "when log was not created via a bulk upload and has an unanswered question" do
-        subject(:component) { described_class.new(questions:, log:, user:) }
-
-        let(:log) { build(:lettings_log, :in_progress, age2: 99, startdate: Time.zone.local(2021, 5, 1)) }
-
-        it "displays normal copy with muted colour " do
-          expect(rendered).to have_selector("span", class: "app-!-colour-muted", text: "You didn’t answer this question")
-        end
-      end
-
-      context "when log was created via a bulk upload and has an unanswered optional question" do
-        subject(:component) { described_class.new(questions:, log:, user:) }
-
-        let(:subsection_id) { "setup" }
-        let(:bulk_upload) { create(:bulk_upload) }
-        let(:log) { build(:lettings_log, :completed, creation_method: "bulk upload", tenancycode: nil, startdate: Time.zone.local(2021, 5, 1), bulk_upload:) }
-
-        it "displays tweaked copy in red" do
-          expect(rendered).to have_selector("span", class: "app-!-colour-muted", text: "You didn’t answer this question")
-          expect(rendered).not_to have_selector("span", class: "app-!-colour-red", text: "You still need to answer this question")
-        end
-      end
-    end
-  end
-
-  context "when in 23/24 collection" do
-    context "when given a set of questions" do
-      let(:user) { create(:user) }
-      let(:log) { create(:lettings_log, :completed, age2: 99, startdate: Time.zone.local(2023, 5, 1)) }
-      let(:subsection_id) { "household_characteristics" }
-      let(:subsection) { log.form.get_subsection(subsection_id) }
-      let(:questions) { subsection.applicable_questions(log) }
-
-      around do |example|
-        Timecop.freeze(Time.zone.local(2023, 5, 1))
-        Singleton.__init__(FormHandler)
-        example.run
-        Timecop.return
-        Singleton.__init__(FormHandler)
-      end
-
-      it "renders a summary list card including question numbers for the answers to those questions" do
-        expect(rendered).to have_content(questions.first.answer_label(log))
-        expect(rendered).to have_content("Q32 - Lead tenant’s age")
+        expect(rendered).not_to include(" - Lead tenant’s age")
       end
     end
   end
--- a/spec/components/check_answers_summary_list_card_component_spec.rb
+++ b/spec/components/check_answers_summary_list_card_component_spec.rb
@@ -61,7 +61,7 @@ RSpec.describe CheckAnswersSummaryListCardComponent, type: :component do
 
   context "when before 23/24 collection" do
     context "when given a set of questions" do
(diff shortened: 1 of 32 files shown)


spec/helpers/check_answers_helper_spec.rb
arfacamble, line 1: There are factories for forms, questions, etc if we think using that would be better. I don't necessarily think it would, this seems pretty readable and not necessary to abstract away from it

spec/helpers/locations_helper_spec.rb
```diff
@@ -51,7 +51,7 @@
     let(:location) { FactoryBot.create(:location, startdate: nil) }
 
     before do
-      Timecop.freeze(2023, 10, 10)
+      allow(Time).to receive(:now).and_return(Time.zone.local(2023, 10, 10))
       allow(FormHandler.instance).to receive(:lettings_in_crossover_period?).and_return(true)
```
arfacamble, line 59: if we're removing the freeze before can we remove the unfreeze after?

spec/helpers/tasklist_helper_spec.rb
```diff
@@ -148,37 +143,26 @@
...
+          expect(review_log_text(lettings_log)).to eq(
+            "You can #{govuk_link_to 'review and make changes to this log', review_lettings_log_path(lettings_log)} until 9 June 2023.".html_safe,
+          )
         end
       end
 
       context "when collection_period_open? == false" do
-        let(:now) { Time.utc(2022, 6, 1) }
-        let!(:sales_log) { create(:lettings_log, :completed) }
+        let(:now) { Time.utc(2024, 6, 1) }
```
arfacamble: as a general question for the whole PR and what we're trying to do with the tests overall, I was under the impression we would try and remove things where the time is fixed in general.
for example here it looks to me that we're testing that some text is different when a collection period is closed, it seems the most robust to me to stub the call to collection_period_open? is it possible to stub methods on a model made from a factory or only an instance double?

spec/models/form/lettings/questions/address_line1_for_address_matcher_spec.rb
```diff
@@ -6,7 +6,7 @@
   let(:question_id) { nil }
   let(:question_definition) { nil }
   let(:page) { instance_double(Form::Page) }
-  let(:log) { create(:lettings_log, :in_progress, address_line1_input: "Address line 1", postcode_full_input: "AA1 1AA") }
+  let(:log) { build(:lettings_log, :in_progress, address_line1_input: "Address line 1", postcode_full_input: "AA1 1AA") }
```
arfacamble, line 9: loving these switched to build where possible

spec/helpers/tasklist_helper_spec.rb
kosiakkatrina: Yep, can do that here, nice spot! Any specific tests in this PR that you noticed a similar issue on?

spec/helpers/locations_helper_spec.rb
kosiakkatrina: Yup!

spec/helpers/check_answers_helper_spec.rb
kosiakkatrina: Ah yes, I see! I think I'll keep it in mind if I come across another use case for it. If you're comfortable with me leaving it as is for these tests, I might just do that

spec/helpers/tasklist_helper_spec.rb
arfacamble: no more a general comment. I can't see how making Time return a given date for now makes the tests more robust. Won't we still get the same type of problems when we roll over collection years?

kosiakkatrina: Yeah, that's fair. I think this might have been fine cause we're not reinstantiating the forms, but I've mocked some things now and removed the time travelling! For some suites the tests are still time dependent, so I've changed to stubbing `now` instead of freezing, cause we don't have to unfreeze it that way 🤷

spec/components/check_answers_summary_list_card_component_spec.rb
```diff
@@ -5,103 +5,68 @@
...
 
-  context "when in 23/24 collection" do
+  context "when before 23/24 collection" do
     context "when given a set of questions" do
-      let(:user) { create(:user) }
-      let(:log) { create(:lettings_log, :completed, age2: 99, startdate: Time.zone.local(2023, 5, 1)) }
-      let(:subsection_id) { "household_characteristics" }
-      let(:subsection) { log.form.get_subsection(subsection_id) }
-      let(:questions) { subsection.applicable_questions(log) }
+      let(:log) { build(:lettings_log, :completed, age2: 99, startdate: Time.zone.local(2021, 5, 1), assigned_to: create(:user)) }
```
arfacamble, line 64: I'm surprised it still makes sense to test a log with a start date so old.
But from the looks of the test it doesn't really matter.

spec/components/check_answers_summary_list_card_component_spec.rb
```diff
@@ -5,103 +5,68 @@
...
-
-        let(:bulk_upload) { create(:bulk_upload) }
-        let(:log) { build(:lettings_log, :in_progress, creation_method: "bulk upload", age2: 99, startdate: Time.zone.local(2021, 5, 1), bulk_upload:) }
-
-        it "displays tweaked copy in red" do
-          expect(rendered).to have_selector("span", class: "app-!-colour-red", text: "You still need to answer this question")
-        end
-      end
+  let(:user) { create(:user) }
+  let(:log) { create(:lettings_log, :completed, age2: 99, retirement_value_check: 1) }
```
arfacamble, line 9: if we can get away with building the log below, can we get away with it here?

spec/factories/lettings_log.rb
arfacamble, line 1: nice

spec/features/bulk_upload_lettings_logs_spec.rb
```diff
@@ -22,11 +27,7 @@
   # rubocop:disable RSpec/AnyInstance
```
arfacamble, line 27: do we need this rubocop disable? I'm assumign it's for the `allow(FormHandler.instance).to receive` but when I check for other instances of that around the codebase they don't need the disable.

spec/features/bulk_upload_lettings_logs_spec.rb
arfacamble, line 1: big fan of these changes

spec/helpers/interruption_screen_helper_spec.rb
```diff
@@ -249,7 +239,7 @@
 
     it "returns a list of questions affected by the soft validation" do
       expect(soft_validation_affected_questions(question, lettings_log).count).to eq(2)
-      expect(soft_validation_affected_questions(question, lettings_log).map(&:id)).to eq(%w[age1 ecstat1])
+      expect(soft_validation_affected_questions(question, lettings_log).map(&:id)).to eq(%w[ecstat1 age1])
```
arfacamble: is there a reason that this changed order, and if the order is not important and it's just a touch flakey, you could
```suggestion
      expect(soft_validation_affected_questions(question, lettings_log).map(&:id)).to match_array %w[ecstat1 age1]
```

spec/helpers/locations_helper_spec.rb
```diff
@@ -51,14 +51,10 @@
     let(:location) { FactoryBot.create(:location, startdate: nil) }
 
     before do
-      Timecop.freeze(2023, 10, 10)
+      allow(Time).to receive(:now).and_return(Time.zone.local(2023, 10, 10))
```
arfacamble: I'm sure you tried to remove the stubbing of now entirely, but I'm slightly surprised that we're stubbing time but there's nothing in the context or describe blocks above it about being in a given collection year

spec/helpers/schemes_helper_spec.rb
arfacamble, line 1: I'm gently icked by there being lots of dates in this and the locations helper spec.
I imagine that locations and schemes don't have such large numbers of validations so it's probably fine in this case since they will still be meaningful and robust?

I think it would probably be very tedious but I'm imagining a nice way to write this would be to define however many dates are necessary at the top of the file
```
let(today) {Time.zone.today}
let(one_year_ago) { today - 1.year }
let(six months ago) { etc }
```
and then have tests that finish with things like
```
expect(scheme_active_periods(scheme).first).to have_attributes(from: on_year_ago, to: today)
```

I completely understand if it's not worth the time to do that.

spec/helpers/tasklist_helper_spec.rb
```diff
@@ -1,41 +1,32 @@
...
+        lettings_log["uprn_known"] = 0
+        expect(get_next_incomplete_section(lettings_log).id).to eq("property_information")
       end
     end
 
     describe "get sections count" do
-      let(:real_2021_2022_form) { Form.new("config/forms/2021_2022.json") }
-
-      before do
-        allow(FormHandler.instance).to receive(:get_form).and_return(real_2021_2022_form)
```
arfacamble, line 33: 💥 💣 💥 💣 💥 💣 💥 💣 💥  
`real_2021_2022_form`
💣 💥 💣 💥 💣 💥 💣 💥 💣

spec/helpers/tasklist_helper_spec.rb
```diff
@@ -1,41 +1,32 @@
...
-      end
+      let(:completed_subsection) { instance_double("Subsection", status: :completed, displayed_in_tasklist?: true, applicable_questions: [{ id: "question" }]) }
+      let(:incomplete_subsection) { instance_double("Subsection", status: :not_started, displayed_in_tasklist?: true, applicable_questions: []) }
```
arfacamble, line 21: I've not seen that you can set the response to methods on an instance_double when defining it. that's nice

spec/helpers/tasklist_helper_spec.rb
```diff
@@ -148,44 +163,41 @@
 
     context "with lettings log" do
       context "when collection_period_open? == true" do
-        context "with 2023 deadline" do
-          let(:now) { Time.utc(2022, 6, 1) }
-          let(:lettings_log) { create(:lettings_log, :completed) }
+        let(:lettings_log) { build(:lettings_log, :completed, id: 123) }
```
arfacamble, line 166: I also hadn't thought of defining id when building. Since it wasn't in the factory I thought it might not know about it which is stooopid. This is nicer than what I've done in a few cases recently

spec/helpers/tasklist_helper_spec.rb
arfacamble, line 1: There's a lot of `Time.utc` which I guess is probably fine since it's only in spec code which is testing against the other instance of that within this file.
But if it's as easy as a filewide find and replace maybe it's worth changing them all to `.zone.local`?

spec/models/form/sales/pages/la_nominations_spec.rb
```diff
@@ -3,11 +3,13 @@
...
   subject(:page) { described_class.new(page_id, page_definition, subsection) }
 
-  let(:log) { create(:sales_log, :completed) }
+  let(:log) { build(:sales_log, :completed) }
 
   let(:page_id) { nil }
   let(:page_definition) { nil }
-  let(:subsection) { instance_double(Form::Subsection, form: instance_double(Form, start_date: Time.zone.local(2023, 4, 1))) }
+  let(:start_year_after_2024) { false }
+  let(:form) { instance_double(Form, start_date: Time.zone.local(2023, 4, 1), start_year_after_2024?: start_year_after_2024) }
```
arfacamble, line 11: if we're stubbing the response to :start_year_after_2024?` do we still need the start_date?

spec/models/form/sales/pages/last_accommodation_la_spec.rb
```diff
@@ -5,22 +5,10 @@
...
-    allow(subsection).to receive(:depends_on).and_return(nil)
-    allow(subsection).to receive(:enabled?).and_return(true)
-  end
-
-  after do
-    Timecop.return
-    Singleton.__init__(FormHandler)
-  end
+  let(:start_year_after_2024) { false }
+  let(:form) { instance_double(Form, depends_on_met: true, start_date: Time.zone.local(2023, 4, 1), start_year_after_2024?: start_year_after_2024) }
```
arfacamble, line 9: as above with stubbing v start_date

spec/models/form/sales/pages/last_accommodation_spec.rb
```diff
@@ -3,23 +3,13 @@
...
-    Singleton.__init__(FormHandler)
-    allow(subsection).to receive(:depends_on).and_return(nil)
-  end
-
-  after do
-    Timecop.return
-    Singleton.__init__(FormHandler)
-  end
+  let(:start_year_after_2024) { false }
+  let(:form) { instance_double(Form, start_date: Time.zone.local(2023, 4, 1), start_year_after_2024?: start_year_after_2024) }
```
arfacamble, line 11: as above

spec/models/form/subsection_spec.rb
```diff
@@ -69,22 +51,14 @@
...
-
-      let(:section_id) { "tenancy_and_property" }
-      let(:section_definition) { form.form_definition["sections"][section_id] }
-      let(:section) { Form::Section.new(section_id, section_definition, form) }
-      let(:subsection_id) { "property_information" }
-      let(:subsection_definition) { section_definition["subsections"][subsection_id] }
-
       it "has a started status even if only an optional field has been answered" do
-        lettings_log.postcode_known = 0
+        lettings_log.age1 = nil
```
arfacamble: I'm confused by this test. Is age1 optional? If not what is the optional field that has been answered?

spec/models/sales_log_spec.rb
```diff
@@ -1025,6 +1025,8 @@ def previous_postcode_fields
 
     describe "#search_by" do
       it "allows searching using ID" do
+        sales_log_to_search.id += 30_000
+        sales_log_to_search.save!
```
arfacamble, line 1029: woohoo. Let's hope this kills the flake

spec/components/check_answers_summary_list_card_component_spec.rb
kosiakkatrina: I think it needs creating, not sure why, but without it questions ends up being nil

spec/features/bulk_upload_lettings_logs_spec.rb
kosiakkatrina: I think it's for the `allow_any_instance_of(Forms::BulkUploadLettings::UploadYourFile).to receive(:`).and_return("text/csv")`

spec/helpers/interruption_screen_helper_spec.rb
kosiakkatrina: Yup, that's a good idea

spec/helpers/schemes_helper_spec.rb
kosiakkatrina: I think I've done something similar in other scheme/location tests in the following PRs, but if I remember it correctly for this one it's a bit more nuanced, cause it depends on whereabouts in the year we currently are. If its December, then 6 months ago is still the same collection year, if it's May, it's the previous one and there's some dependency around start dates being in the beginning of the year.
There's also some dependencies around setting the deactivation date, if we want to set deactivation date for 1 year and 5 months ago, at some times it might hist this validation `Deactivation date The date must be on or after beginning of open collection period`

So I can extract some dates so that it reads nicer, but I don't think there's a very nice way to avoid setting time here

spec/helpers/tasklist_helper_spec.rb
kosiakkatrina: Yup, if they're short it's quite nice

spec/helpers/tasklist_helper_spec.rb
kosiakkatrina: sure

spec/models/form/sales/pages/la_nominations_spec.rb
kosiakkatrina: Yeah, I think it's used in question itself for question number

spec/models/form/sales/pages/last_accommodation_la_spec.rb
kosiakkatrina: I think same

spec/models/form/sales/pages/last_accommodation_spec.rb
kosiakkatrina: same

spec/models/form/subsection_spec.rb
kosiakkatrina: Yeah, good point, I'll update it

spec/models/sales_log_spec.rb
kosiakkatrina: 🤞

spec/helpers/schemes_helper_spec.rb
arfacamble: that makes me think
1. the tests are not then documenting the functionality in an obvious way
2. it would take even longer to design the dates that go at the top
3. it would take too long to be a priority

spec/models/form/sales/pages/la_nominations_spec.rb
arfacamble: aha.

arfacamble (approved):

PR #2423, opened 2024-05-17
Remove duplicate rent value check from CSV

kosiakkatrina:


app/services/csv/lettings_log_csv_service.rb
```diff
@@ -310,7 +310,8 @@ def soft_validations_attributes(ordered_questions)
 
     def lettings_log_attributes
       ordered_questions = FormHandler.instance.ordered_questions_for_year(@year, "lettings")
-      ordered_questions.reject! { |q| q.id.match?(/age\d_known|nationality_all_group/) }
+      soft_validations_attributes = soft_validations_attributes(ordered_questions)
+      ordered_questions.reject! { |q| q.id.match?(/age\d_known|nationality_all_group|rent_value_check/) }
```
kosiakkatrina, line 314: Added rent value check here, cause it's also in attribute_mappings, as we want it to be after rent

RachaelBooth (approved):

PR #2425, opened 2024-05-17
Refactor CSV service

kosiakkatrina:
Only extracted a couple of methods and moved the methods around in order that makes more sense to me, but happy to change if the order is more confusing this way

### DIFF
--- a/app/services/csv/lettings_log_csv_service.rb
+++ b/app/services/csv/lettings_log_csv_service.rb
@@ -157,64 +157,6 @@ module Csv
       voiddate
     ].freeze
 
-    def value(attribute, log)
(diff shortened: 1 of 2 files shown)


app/services/csv/lettings_log_csv_service.rb
```diff
@@ -157,64 +157,6 @@ def prepare_csv(logs)
...
-
-    def conventional_yes_no_label(value)
-      return "Yes" if value == 1
-      return "No" if value&.zero?
-    end
-
-    def yes_or_blank_label(value)
-      value == 1 ? "Yes" : nil
-    end
-
```
kosiakkatrina, line 217: Removed this, I don't think it's being used (?)

arfacamble (approved):

PR #2427, opened 2024-05-17
CLDC-3447: Allow decimal discount percentage and increased validation tolerance

RachaelBooth:


RachaelBooth:
This does loosen the validation, which is mildly annoying. It is a bit of an edge case to have percentage discounts where the percentage isn't an integer (which is what means this is needed), but it does happen - this came from a support ticket.

spec/models/validations/sales/sale_information_validations_spec.rb
```diff
@@ -204,190 +204,71 @@
   end
 
   describe "#validate_discounted_ownership_value" do
-    let(:record) { FactoryBot.build(:sales_log, mortgage: 10_000, deposit: 5_000, value: 30_000, ownershipsch: 2, type: 8, saledate: now) }
```
RachaelBooth, line 207: These changes look bigger than they are because of a formatting change.

In essence: it's not actually possible for both discount and grant to be routed to - I've removed tests that depend on this and replaced with relevant tests around the new tolerance. I've also removed the time freeze. I've not touched the content of other tests here, just the indentation.

kosiakkatrina: Is there a test where discount is a decimal?

RachaelBooth: Good point - I wanted to tidy up the figures rather than use the ones from the ticket, but I didn't end up with a decimal one and should. Will add.

kosiakkatrina (approved):

PR #2431, opened 2024-05-21
CLDC-3460: Don't validate P2 not child in BU when not a joint purchase

RachaelBooth:


app/services/bulk_upload/sales/year2024/row_parser.rb
```diff
@@ -1450,6 +1450,8 @@ def validate_buyer1_economic_status
   end
 
   def validate_buyer2_economic_status
```
RachaelBooth, line 1452: I looked at getting rid of this entirely - it is covered by the validations on the model, but the error messages you get are then a bit confusing so I went for this.

app/services/bulk_upload/sales/year2024/row_parser.rb
```diff
@@ -635,11 +635,11 @@ def outright_sale?
   end
 
   def joint_purchase?
-    field_114 == 1
+    field_15 == 1
   end
```
kosiakkatrina, line 639: Looks like this method is wrong in 2023 as well?

RachaelBooth: Huh, you're right - I'd assumed it had changed for 2024.

kosiakkatrina (approved):

PR #2432, opened 2024-05-22
CLDC-3289 Avoid freezing time in tests (part 2)

kosiakkatrina:


spec/models/validations/household_validations_spec.rb
```diff
@@ -277,7 +269,7 @@
   end
 
   describe "#validate_partner_count" do
-    let(:log_date) { Time.zone.local(2023, 4, 1) }
+    let(:startdate) { Time.zone.local(2023, 4, 1) }
```
RachaelBooth, line 272: Is this true for any date? If so, we can remove this and have startdate now, if not can we put in a context?

spec/models/validations/sales/household_validations_spec.rb
```diff
@@ -4,8 +4,8 @@
   subject(:household_validator) { validator_class.new }
 
   let(:validator_class) { Class.new { include Validations::Sales::HouseholdValidations } }
-  let(:record) { build(:sales_log, saledate: log_date) }
-  let(:log_date) { Time.zone.local(2023, 4, 1) }
+  let(:record) { build(:sales_log, saledate:) }
+  let(:saledate) { Time.zone.local(2023, 4, 1) }
```
RachaelBooth: Is this able to be now?

spec/models/validations/sales/household_validations_spec.rb
```diff
@@ -377,21 +337,8 @@
...
-      Singleton.__init__(FormHandler)
-      record.ownershipsch = 2
-      record.saledate = now
-    end
-
-    after do
-      Timecop.return
-      Singleton.__init__(FormHandler)
-    end
+    let(:saledate) { Time.zone.local(2024, 4, 4) }
```
RachaelBooth: Same for this - does it work if this is now?

spec/models/validations/sales/setup_validations_spec.rb
```diff
@@ -7,15 +7,6 @@
 
   describe "#validate_saledate_collection_year" do
     context "with sales_in_crossover_period == false" do
-      before do
```
RachaelBooth, line 10: Do we need to force this to be true some other way?

spec/models/validations/sales/setup_validations_spec.rb
```diff
@@ -26,29 +17,41 @@
         end
       end
 
-      context "when saledate is in the 22/23 collection year" do
+      context "when saledate is in the open collection year" do
         let(:record) { build(:sales_log, saledate: Time.zone.local(2023, 1, 1)) }
 
+        before do
+          allow(Time).to receive(:now).and_return(Time.zone.local(2023, 1, 10))
+        end
```
RachaelBooth: I don't think freezing time here (essentially - I think the difference between this and timecop is that timecop may also mock some other stuff?) is helping - if the context here is a general "saledate is in the open collection year" then could that just be the saledate being (real) now?

spec/models/validations/sales/setup_validations_spec.rb
```diff
@@ -26,29 +17,41 @@
...
         end
       end
 
-      context "when saledate is before the 22/23 collection year" do
+      context "when saledate is before the open collection year" do
         let(:record) { build(:sales_log, saledate: Time.zone.local(2020, 1, 1)) }
 
+        before do
+          allow(Time).to receive(:now).and_return(Time.zone.local(2023, 1, 10))
+        end
```
RachaelBooth: Similarly to above, could these be relative to real now rather than needing to fix dates?

I think we could use a regex to avoid needing to specify specific years in the error message assertion.

spec/models/validations/sales/setup_validations_spec.rb
```diff
@@ -58,14 +61,6 @@
     end
 
     context "with sales_in_crossover_period == true" do
-      around do |example|
```
RachaelBooth, line 61: Do we need to force `with_sales_in_crossever_period` to return true for this?

RachaelBooth: Also, it looks like some of the tests in this context have some weird things (not in this PR) going on with their dates (in particular actual used dates vs mentioned in context names), and might fail once the current crossover period does finish or when the collection year rolls over - any chance you can take a look at them?

kosiakkatrina: Yeah this needed with_sales_in_crossever_period stub, the tests were also a bit wrong. Setting the new_logs_end_date and edit_end_date to before today sill passes these tests

spec/models/validations/sales/setup_validations_spec.rb
kosiakkatrina: I feel like this one is a bit trickier, cause we're testing the error message content `Enter a date within the 22/23 collection year`, I think I'd be inclined to leave this, unless we construct the error message on current open period dates, but then I feel like we're not testing that the error message gets constructed correctly cause we'd be mirroring the code that we're testing (?)

spec/models/validations/sales/setup_validations_spec.rb
kosiakkatrina: yup

spec/models/validations/sales/setup_validations_spec.rb
kosiakkatrina: yup

spec/models/validations/sales/household_validations_spec.rb
kosiakkatrina: Yeah, I guess I wanted this to be explicit that it's on/after 2024 cause there are also tests for 2023 specific behaviour, but I guess everything from now on will be on or after 2024, so will remove

spec/models/validations/sales/household_validations_spec.rb
kosiakkatrina: This fails one validation, I've set it to now and moved the date into that validation test context

spec/models/validations/household_validations_spec.rb
kosiakkatrina: I think the validation is only relevant for up to 2024

spec/models/validations/sales/setup_validations_spec.rb
```diff
@@ -26,8 +21,8 @@
         end
       end
 
-      context "when saledate is in the 22/23 collection year" do
-        let(:record) { build(:sales_log, saledate: Time.zone.local(2023, 1, 1)) }
+      context "when saledate is in the open collection year" do
+        let(:record) { build(:sales_log) }
```
arfacamble: with the description in the context block here I would expect we would still need a saledate?
`:saledate_today` trait?

spec/models/validations/sales/setup_validations_spec.rb
arfacamble: In one of the tests on my new PR I have something like this and I decided that it was sufficient to test that an error was being added and it had the right format using regex.
```
expect(record.errors[:saledate]).to include(/Enter a date within the \d{2}\/\d{2} collection year, which is between 1st April \d{4} and 31st March \d{4}/)
```

spec/models/validations/sales/setup_validations_spec.rb
```diff
@@ -36,19 +31,27 @@
...
+
         it "adds error" do
           setup_validator.validate_saledate_collection_year(record)
 
           expect(record.errors[:saledate]).to include("Enter a date within the 22/23 collection year, which is between 1st April 2022 and 31st March 2023")
         end
       end
 
-      context "when saledate is after the 22/23 collection year" do
+      context "when saledate is after the open collection year" do
```
arfacamble, line 44: similar to the one above, if you're happy testing the general format with regex you could dodge all the time based tomfoolery with saledate: today + n.years

spec/models/validations/sales/setup_validations_spec.rb
```diff
@@ -113,11 +112,11 @@
...
 
-        it "cannot create new logs for the previous collection year" do
+        it "cannot create new logs for the archived collection year" do
           record.update!(saledate: nil)
-          record.saledate = Time.zone.local(2024, 1, 1)
+          record.saledate = Time.zone.local(2023, 1, 1)
           setup_validator.validate_saledate_collection_year(record)
-          expect(record.errors["saledate"]).to include(match "Enter a date within the 24/25 collection year, which is between 1st April 2024 and 31st March 2025")
+          expect(record.errors["saledate"]).to include(match "Enter a date within the 23/24 or 24/25 collection years, which is between 1st April 2023 and 31st March 2025")
         end
```
arfacamble: I imagine these tests might run a bit quicker if you just build a log with saledate: x on line 1 and feed that straight into the validator? Is there some important state in record ebfore the test starts?

spec/models/validations/soft_validations_spec.rb
```diff
@@ -2,16 +2,7 @@
...
-
-  before do
-    Timecop.freeze(Time.zone.local(2021, 10, 10))
-    Singleton.__init__(FormHandler)
-  end
-
-  after do
-    Timecop.return
-  end
+  let(:record) { FactoryBot.build(:lettings_log, owning_organisation: organisation) }
```
arfacamble: can we also build the organisation, possibly if it is given an id?

spec/models/validations/soft_validations_spec.rb
```diff
@@ -84,7 +75,8 @@
 
   describe "retirement soft validations" do
     before do
-      record.update!(age1:, ecstat1:)
+      record.age1 = age1
+      record.ecstat1 = ecstat1
```
arfacamble, line 79: really unimportant, here, maybe worthwhile below, you could use `update_attributes(age1:, ecstat1:)`

spec/policies/location_policy_spec.rb
```diff
@@ -44,18 +44,12 @@
 
     context "with deactivated location" do
       before do
-        location.location_deactivation_periods << create(:location_deactivation_period, deactivation_date: Time.zone.local(2024, 4, 10), location:)
+        location.location_deactivation_periods << create(:location_deactivation_period, deactivation_date: Time.zone.today, location:)
         location.save!
```
arfacamble: did you try building the deactivation period on line 47? I have a feeling it would then be saved on line 48 as a thing dependent on the location, but I'm not completely sure. It might roll them into one DB transaction and be marginally quicker

spec/models/validations/sales/setup_validations_spec.rb
kosiakkatrina: I'll rebase when merged!

spec/models/validations/sales/setup_validations_spec.rb
kosiakkatrina: You're right, I thought that was the default!

spec/models/validations/sales/setup_validations_spec.rb
kosiakkatrina: For this specific test it needs to be nil before we update it to validate that it's a new record in the collection year and not one that's being updated, but you're right, I can move it to create

arfacamble (approved):

PR #2438, opened 2024-05-29
CLDC-3448 Generate validations documentation

kosiakkatrina:
Add tasks to generate documentation for validations using openAI.
Add separate tasks for:
- hard validations
- soft validations
- BU specific validations
- min/max validations (can generate these without openAI)

Add rails_admin showing only Validations table so that support users can update them.


app/services/documentation_generator.rb
```diff
@@ -0,0 +1,371 @@
...
+              },
+            },
+          ],
+          tool_choice: { type: "function", function: { name: "create_documentation_for_given_validation" } },
+        },
+      )
+    rescue StandardError => e
+      Rails.logger.error("Failed to describe #{meth} for #{form.start_date.year}")
+      Rails.logger.error("Error #{e.message}")
+      sleep(15)
```
RachaelBooth, line 202: What's the sleep for?

lib/tasks/generate_lettings_documentation.rake
```diff
@@ -0,0 +1,90 @@
...
+
+      if min && max
+        validation_name = "range"
+        error_message = I18n.t("validations.numeric.within_range", field:, min:, max:)
+        validation_description = "Field value is lower than the minimum value or higher than the maximum value"
+      end
+
+      if LogValidation.where(validation_name:, field:, log_type: "lettings").exists?
+
+        Rails.logger.info("Validation #{validation_name} already exists for #{field}")
```
RachaelBooth, line 75: What happens if we want to update this (if a value has changed)?

RachaelBooth (approved):
Looks fine to me - I think there might be some things to think about for how we keep this updated as the code changes (I don't want to be running this on every release, but it does risk getting stale), but I don't think that necessarily needs to be solved before this goes out.

lib/tasks/generate_lettings_documentation.rake
kosiakkatrina: I think there's a separate ticket for updating things, only focusing on generating and making sure there are no duplicates in this one!

app/services/documentation_generator.rb
kosiakkatrina: Oh, yeah, there's a limit on the tokens we can use per minute, for 3.5 it's kinda low and I'm expecting to run these tasks a few times anyway to populate everything, but the sleep kinda helps with that

PR #2444, opened 2024-05-30
CLDC-3463: Fix bug with vacdays calculation

RachaelBooth:


lib/tasks/recalculate_vacdays.rake
```diff
@@ -0,0 +1,5 @@
+desc "Recalculate vacdays after bugfix for daylight savings time changes"
+task recalculate_vacdays: :environment do
+  logs = LettingsLog.filter_by_years(%w[2023 2024]).where.not(vacdays: nil)
+  logs.each(&:save!)
+end
```
RachaelBooth, line 17: Mainly a task for ease of kicking it off on an ad hoc instance.

This is filtered to open years to avoid validation issues - need to check with the stats team what to do about older data. The export after running this can be expected to be relatively large, but that should be ok.

spec/models/lettings_log_spec.rb
```diff
@@ -1514,6 +1514,24 @@ def previous_postcode_fields
       expect(record_from_db["has_benefits"]).to eq(1)
     end
 
+    describe "deriving vacant days" do
```
RachaelBooth: These should be written in a way that makes it easy to solve merge fun with https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/2442 - (hence the separation from surrounding tests) but will need to do that because there wasn't a way to avoid entirely.

app/models/derived_variables/lettings_log_variables.rb
```diff
@@ -306,9 +306,9 @@ def property_vacant_days
     return unless startdate
 
     if mrcdate.present?
-      (startdate - mrcdate).to_i / 1.day
+      startdate.to_date - mrcdate.to_date
     elsif voiddate.present?
-      (startdate - voiddate).to_i / 1.day
```
RachaelBooth, line 301: FYI the bug comes from doing this calculation with a datetime. If the startdate is in summer time and the mrcdate / voiddate is not then the subtraction gives a number slightly under the expected number of days (i.e. an hour short), and the integer division gives the floor.

spec/services/csv/lettings_log_csv_service_spec.rb
```diff
@@ -245,120 +162,399 @@
...
-      let(:user) { create(:user, email: "choreographer@owtluk.com") }
+      let(:user) { create(:user, :data_coordinator, email: "choreographer@owtluk.com") }
 
       it "does not include certain attributes in the headers" do
         expect(headers).not_to include(*%w[wrent wscharge wpschrge wsupchrg wtcharge])
       end
+    end
 
-      context "and the requested form is 2024" do
+    describe "the full CSV output" do
```
RachaelBooth, line 180: I've pulled this out to reduce the number of times I need to define a full log

lib/tasks/recalculate_vacdays.rake
kosiakkatrina: I hope we don't have invalid logs on prod, but in case there are any, I think save! would fail with an error and stop processing?

Also, would it make sense to only save where the value would have changed to avoid the large export? I assume some of the values were correct?

kosiakkatrina (approved):

lib/tasks/recalculate_vacdays.rake
RachaelBooth: That's a good point - I'd expect nothing to be invalid but it could be if the validations have changed since it was touched. I guess maybe we don't actually care about whether it's valid for this purpose, although it would be nice to find that out.

We need to run `set_derived_fields!` which is in a before validation hook, but we could instead do that manually and not validate.

On the value changing, I think that rails will only actually make a database change (and thus update the updated_at time etc) if an attribute has actually been changed - will double check this.

RachaelBooth: Ok, I'm trying something that will just manually update the vacdays value only when there's a change (and not run the validations so that it can work on all logs regardless of collection year - this also means that if there are currently invalid logs it will still work, which seems not ideal but also it's not making it worse).

kosiakkatrina: You could potentially run `unless log.save Rails.logger.something_useful` as well

PR #2445, opened 2024-06-03
CLDC-3289 Avoid freezing time (part 3)

kosiakkatrina:


spec/models/validations/property_validations_spec.rb
arfacamble, line 1: I have made more comprehensive changes to this file in this PR, including the change you have made
https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/2478
suggest you revert this to avoid conflicts

kosiakkatrina: I've rebased this now, there's still a little diff, but I assume that's okay?

arfacamble: not sure. you have merge conflicts on the PR, so I imagine this will be included. This line in my PR is now just
`let(:log) {build(:lettings_log) }`

kosiakkatrina: Oh yeah, fair, not sure what happened there, but should be rebased now

spec/requests/form_controller_spec.rb
arfacamble, line 1: I've touched these tests in my PR.
similar change to you but I'm surprised that it passes from your change since I think now the saledate will be nil in yours, but they're meant to be duplicate.
Maybe really being duplicates is not necessary for the test to pass, don't know if that's a problem for the test

spec/models/bulk_upload_spec.rb
```diff
@@ -13,23 +13,24 @@
 
     context "when there are no incomplete logs" do
       it "returns true" do
-        create_list(:lettings_log, 2, :completed, bulk_upload:, startdate: Time.zone.local(2024, 3, 3), voiddate: Time.zone.local(2024, 3, 3), mrcdate: Time.zone.local(2024, 3, 3))
+        logs = build_list(:lettings_log, 2, :completed, bulk_upload:, startdate: Time.zone.local(2024, 3, 3), voiddate: Time.zone.local(2024, 3, 3), mrcdate: Time.zone.local(2024, 3, 3), assigned_to: create(:user))
+        logs.each { |log| log.save(validate: false) }
```
RachaelBooth: We've got a trait in the factory for ignoring validation errors (but still running the validations so that all the relevant hooks happen) - is this alright to use that?

spec/models/sales_log_spec.rb
```diff
@@ -6,17 +6,19 @@
   let(:owning_organisation) { create(:organisation) }
   let(:assigned_to_user) { create(:user) }
 
-  before do
-    Timecop.freeze(Time.zone.local(2024, 3, 1))
-    Singleton.__init__(FormHandler)
-  end
+  context "with shared examples" do
```
RachaelBooth: I don't think these want to be in their own context (it should be sensibly named within the shared examples file). And I'd hope we can avoid freezing time for those too - or at least if they needed it it would be done in that file.

spec/models/validations/household_validations_spec.rb
```diff
@@ -59,7 +51,7 @@
       end
 
       context "when form year is before 2024" do
-        let(:log_date) { Time.zone.local(2024, 1, 1) }
+        let(:startdate) { Time.zone.local(2024, 1, 1) }
```
RachaelBooth: Presumably this will be a problem when we hard close the collection year?

spec/requests/form_controller_spec.rb
kosiakkatrina: Yeah, I think your solutions is better, I'll leave it as is!

spec/models/bulk_upload_spec.rb
kosiakkatrina: I think I removed it altogether, cause the test was updated since. This is very useful information to have tho!

spec/models/sales_log_spec.rb
kosiakkatrina: Yep that makes sense, looks like this has also been fixed as a result of something else, so will revert, thanks!

spec/models/validations/household_validations_spec.rb
kosiakkatrina: I removed this, cause I think it got added as part of some other change, but I don't think this would be a problem, because we're building a log instead of creating it and only run the specific validation that's being tested (so it shouldn't fail on date)

spec/models/validations/sales/setup_validations_spec.rb
```diff
@@ -24,6 +24,10 @@
       context "when saledate is in the open collection year" do
         let(:record) { build(:sales_log, :saledate_today) }
 
+        before do
+          allow(Time).to receive(:now).and_return(Time.zone.local(2023, 1, 10))
```
RachaelBooth: Is this freeze needed?

spec/models/validations/sales/setup_validations_spec.rb
```diff
@@ -70,6 +74,10 @@
       context "when saledate is in the 22/23 collection year" do
```
RachaelBooth: I think this wording might want updating.

RachaelBooth (approved):

spec/models/validations/sales/setup_validations_spec.rb
kosiakkatrina: Yup!

spec/models/validations/sales/setup_validations_spec.rb
kosiakkatrina: Doesn't look like it!

PR #2447, opened 2024-06-04
CLDC-3474 Correctly construct dates for template validation

kosiakkatrina:
In some instances when we're checking if the correct template is being used for correct year, we'd use the wrong fields to construct the date

### DIFF
--- a/spec/services/bulk_upload/lettings/validator_spec.rb
+++ b/spec/services/bulk_upload/lettings/validator_spec.rb
@@ -112,7 +112,7 @@ RSpec.describe BulkUpload::Lettings::Validator do
         end
       end
 
-      context "when uploading a 2022 template for 2023 bulk upload" do
+      context "when uploading a 2022 logs for 2023 bulk upload" do
         let(:bulk_upload) { create(:bulk_upload, user:, year: 2023) }
         let(:log) { build(:lettings_log, :completed, startdate: Time.zone.local(2022, 5, 6), tenancycode: "5") }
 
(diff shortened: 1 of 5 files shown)


spec/services/bulk_upload/lettings/validator_spec.rb
```diff
@@ -144,6 +144,39 @@
...
+
+          it "is not valid" do
+            expect(validator).not_to be_valid
+          end
+        end
+
+        context "with headers" do
+          let(:seed) { rand }
+          let(:log_to_csv) { BulkUpload::LettingsLogToCsv.new(log:) }
+          let(:field_numbers) { log_to_csv.default_2023_field_numbers + %w[invalid_field_number] }
```
RachaelBooth: Why is this adding an invalid field number?

spec/services/bulk_upload/lettings/validator_spec.rb
```diff
@@ -144,6 +144,39 @@
...
+            file.write(log_to_csv.custom_field_numbers_row(seed:, field_numbers:))
+            file.write(log_to_csv.to_custom_csv_row(seed:, field_values:))
+            file.rewind
+          end
+
+          it "is not valid" do
+            expect(validator).not_to be_valid
+          end
+        end
+      end
```
RachaelBooth: I'm surprised there aren't any other tests that need adding or changing - is there anything at the moment that is covering these?

RachaelBooth (commented):
This is definitely giving a "how did this not come up in testing" vibe. Is 2024 ok?

spec/services/bulk_upload/lettings/validator_spec.rb
kosiakkatrina: Cause I copied it :sad:, updated now!

spec/services/bulk_upload/lettings/validator_spec.rb
kosiakkatrina: Added some more tests

RachaelBooth (approved):

PR #2449, opened 2024-06-05
CLDC-3482 Correct routing to property LA

kosiakkatrina:
Some depends on rules meant that we weren't always routing to LA question

### DIFF
--- a/app/models/form/lettings/pages/property_local_authority.rb
+++ b/app/models/form/lettings/pages/property_local_authority.rb
@@ -3,7 +3,8 @@ class Form::Lettings::Pages::PropertyLocalAuthority < ::Form::Page
     super
     @id = "property_local_authority"
     @depends_on = [
-      { "is_la_inferred" => false, "is_general_needs?" => true, "address_search_given?" => true },
+      { "is_la_inferred" => false, "is_general_needs?" => true, "form.start_year_after_2024?" => false },
+      { "is_la_inferred" => false, "is_general_needs?" => true, "form.start_year_after_2024?" => true, "address_search_given?" => true },
     ]
   end
 
--- a/app/models/lettings_log.rb
+++ b/app/models/lettings_log.rb
@@ -690,6 +690,16 @@ class LettingsLog < Log
     address_line1_input.present? && postcode_full_input.present?
   end
 
+  def process_postcode_changes!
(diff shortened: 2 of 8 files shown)


app/models/form/lettings/pages/property_local_authority.rb
```diff
@@ -3,7 +3,8 @@ def initialize(id, hsh, subsection)
     super
     @id = "property_local_authority"
     @depends_on = [
-      { "is_la_inferred" => false, "is_general_needs?" => true, "address_search_given?" => true },
+      { "is_la_inferred" => false, "is_general_needs?" => true, "form.start_year_after_2024?" => false },
+      { "is_la_inferred" => false, "is_general_needs?" => true, "form.start_year_after_2024?" => true, "address_search_given?" => true },
```
kosiakkatrina, line 7: I think address_search_given is only applicable to 2024 onwards, cause we're not doing address searches before then. This kinda meant that we haven't routed to LA question for 3 months, so will need to infer things where they can be inferred and set logs in progress in other places.

app/models/lettings_log.rb
```diff
@@ -687,7 +687,7 @@ def duplicates
   end
 
   def address_search_given?
-    address_line1_input.present? && postcode_full.present?
+    address_line1_input.present? && postcode_full_input.present?
```
kosiakkatrina, line 690: I'm assume this was a typo, cause postcode_full doesn't make sense for the method name and the where it's being used

app/models/form/lettings/pages/property_local_authority.rb
RachaelBooth: That's not good!

Do we need to recalculate status on some logs then?

kosiakkatrina: I'm hoping we can infer the LAs on most logs, otherwise, yeah, would need to recalculate

RachaelBooth: Cool - do we need anything to trigger that?

kosiakkatrina: I was gonna do it manually, but that was probably a bad idea. There's a rake task now

RachaelBooth (approved):

PR #2450, opened 2024-06-05
CLDC-3477: Account for ecstat=9 inferernce when calculating applicable_income_range

RachaelBooth:
This avoids the problem where net income could not be confirmed for some logs triggering the soft validation because during save;
* ecstat values for children are cleared early in the save because they aren't routed to
* this changes the soft min / soft max for earnings, meaning net_income_value_check is no longer routed to and thus cleared
* ecstat values for children are set to 9 again, meaning net_income_value_check is routed to again (but now has nil value)

### DIFF
--- a/app/models/lettings_log.rb
+++ b/app/models/lettings_log.rb
@@ -234,7 +234,13 @@ class LettingsLog < Log
 
     if hhmemb > 1
       (2..hhmemb).each do |person_index|
-        ecstat = self["ecstat#{person_index}"] || 10
+        ecstat = self["ecstat#{person_index}"]
+
+        if ecstat.nil?
+          age = self["age#{person_index}"]
+          # This should match the conditions under which ecstat is inferred as 9 (child under 16)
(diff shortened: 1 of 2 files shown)


app/models/lettings_log.rb
```diff
@@ -234,7 +234,13 @@ def applicable_income_range
...
     if hhmemb > 1
       (2..hhmemb).each do |person_index|
-        ecstat = self["ecstat#{person_index}"] || 10
+        ecstat = self["ecstat#{person_index}"]
+
+        if ecstat.nil?
+          age = self["age#{person_index}"]
+          # This should match the conditions under which ecstat is inferred as 9 (child under 16)
+          ecstat = age && age < 16 ? 9 : 10
+        end
```
RachaelBooth, line 243: This did seem to be an easier (to write and to understand) solution than fiddling with the clearing logic, although it's slightly dissatisfying in that there could be similar problems elsewhere in future.

kosiakkatrina (approved):

PR #2453, opened 2024-06-06
CLDC-3483: Prevent/fix missing letting allocation values

RachaelBooth:


lib/tasks/fix_nil_letting_allocation_values.rake
```diff
@@ -0,0 +1,26 @@
+desc "Infer nil letting allocation values as no"
+task fix_nil_letting_allocation_values: :environment do
+  logs = LettingsLog.where(cbl: nil)
+           .or(LettingsLog.where(chr: nil))
+           .or(LettingsLog.where(cap: nil))
+           .or(LettingsLog.filter_by_year(2024).where(accessible_register: nil))
```
kosiakkatrina: Is this an issue for single log submitted ones as well, or just BU? Cause could filter this down by creation_method?

kosiakkatrina (approved):
I have expected us to do this somewhere in derived variables for all checkbox fields already, which should have applied to BU. If we're not doing that, is it an issue for single log submission as well?

RachaelBooth:
> I have expected us to do this somewhere in derived variables for all checkbox fields already, which should have applied to BU. If we're not doing that, is it an issue for single log submission as well?

I can't replicate it for single log submission - as far as I can tell we're setting both the 1s and 0s when saving an answer to the question but it's not an inference that happens whenever we set any of them on a log object directly.

That said, I found a small number of imported logs with creation method single log that did have this problem, so I can't filter them out for the fix task.

PR #2456, opened 2024-06-10
CLDC-3274 Display organisation merges for a longer period

kosiakkatrina:


app/models/organisation.rb
```diff
@@ -156,16 +157,20 @@ def editable_duplicate_sales_logs_sets
     sales_logs.after_date(FormHandler.instance.sales_earliest_open_for_editing_collection_start_date).duplicate_sets.map { |array_str| array_str ? array_str.map(&:to_i) : [] }
   end
 
-  def recently_absorbed_organisations_grouped_by_merge_date
-    return unless absorbed_organisations.present? && absorbed_organisations.merged_during_open_collection_period.present?
+  def organisations_absorbed_during_displayed_collection_period_grouped_by_merge_date
+    return unless absorbed_organisations.present? && absorbed_organisations.merged_during_displayed_collection_period.present?
```
arfacamble: do we think that when this method is called we will hold the absorbed_organisations in memory already? I'm guessing not since `#merged_during_displayed...` is a scope?
if so `#present?` will convert those things to ruby objects, whereas `#exists?` would generate a DB query that returns true/false directly. That might be preferable?
Or maybe if the result of the query is cached by active record anyway it makes no difference (given we then do some more queries on line 163?

Also, given `merged_during_displayed_collection_period` is where clause a scope, I imagine that `absorbed_organisations.present?` is redundant?

app/models/organisation.rb
```diff
@@ -156,16 +157,20 @@ def editable_duplicate_sales_logs_sets
...
-    absorbed_organisations.merged_during_open_collection_period.group_by(&:merge_date)
+    absorbed_organisations.merged_during_displayed_collection_period.group_by(&:merge_date)
   end
 
   def has_recent_absorbed_organisations?
     absorbed_organisations&.merged_during_open_collection_period.present?
   end
 
+  def has_organisations_absorbed_during_displayed_collection_period?
+    absorbed_organisations&.merged_during_displayed_collection_period.present?
```
arfacamble: as above since `absorbed_organisations` is an association and `merged_during_...` then I feel like the former can't be nil, so the safe operator isn't necessary

spec/requests/organisations_controller_spec.rb
arfacamble, line 1: is there a reason why in these tests we create the absorbed orgs and then update them rather than just create them in the state we want them to be in?

arfacamble (approved):
nothing major in my review. my comments in organisation.rb are really just thoughts and don't require changes. If my question on the spec file has any worth that might be worth making a change for.

app/models/organisation.rb
kosiakkatrina: Yup, these suggestions make sense!

spec/requests/organisations_controller_spec.rb
kosiakkatrina: I think it's because there's a validation error when it's all in one statement, I think it's the one when that org tries to create a dpo but it can't cause it's merged

arfacamble: that makes it sound like there should be even more db interactions that we could/should be avoiding.

do you think it's worth adding in something like what I did for rent periods?
- a transient variables which defaults to false called something like skip_creating_dpo
- in the lifecycle hook where we create the dpo we wrap that in unless skip_creating_dpo

kosiakkatrina: Sure, yeah, that already exists, let me see if I can add it here

PR #2459, opened 2024-06-11
CLDC-2480 Allow deleting organisations

kosiakkatrina:


app/controllers/organisations_controller.rb
```diff
@@ -117,7 +117,6 @@ def reactivate
   end
 
   def update
-    selected_rent_periods = rent_period_params[:rent_periods].compact_blank
```
kosiakkatrina, line 120: Moved this to the bottom and under an if statement, because we're not using it until then and if there's any other update than rent periods, it fails (for example deactivating the org)

app/components/sales_log_summary_component.rb
```diff
@@ -10,4 +10,10 @@ def initialize(current_user:, log:)
   def log_status
     helpers.status_tag(log.status)
   end
+
+  def organisation_label(organisation)
+    return unless organisation
+
+    organisation.status == :deleted ? "#{organisation.name} (deleted)" : organisation.name
+  end
```
RachaelBooth: Could this logic go on the organisation or something? It seems like it's repeated a lot.

kosiakkatrina: Ah, good idea! Extracted that now!

app/models/form/sales/questions/owning_organisation_id.rb
```diff
@@ -16,25 +16,22 @@ def answer_options(log = nil, user = nil)
     return answer_opts unless user
     return answer_opts unless log
 
-    unless user.support?
-      if log.owning_organisation_id.present?
-        answer_opts[log.owning_organisation.id] = log.owning_organisation.name
-      end
+    if log.owning_organisation_id.present?
+      org_value = log.owning_organisation.status == :deleted ? "#{log.owning_organisation.name} (deleted)" : log.owning_organisation.name
```
RachaelBooth: I think this can use the new organisation label function as well? (Plus a few other places)

RachaelBooth (approved):

app/models/form/sales/questions/owning_organisation_id.rb
kosiakkatrina: Ah, yes, missed it, thanks!

PR #2463, opened 2024-06-11
CLDC-3504 Update period question options and error message

kosiakkatrina:


spec/models/form/lettings/questions/period_spec.rb
```diff
@@ -8,4 +8,39 @@
...
+    )
+  end
+
+  context "when managing organisation has rent periods" do
+    let(:managing_organisation) { create(:organisation, skip_rent_period_creation: true) }
+    let(:log) { create(:lettings_log, managing_organisation:) }
+
+    before do
+      create(:organisation_rent_period, organisation: managing_organisation, rent_period: 3)
+    end
```
arfacamble, line 36: Did you try just building the log?
I think a normal build, would build an associated user, and the user factory would build an organisation and the log would be assigned to the user and the user's org and the org would have all rent periods built and added to its organisation_rent_periods attribute
```suggestion
    let(:log) { build(:lettings_log) }

    before do
      log.managing_organisation.organisation_rent_periods.filter! { |period| period.rent_period == 3 }
    end
```
I feel like this would work, but I might easily be wrong

arfacamble (approved):
test might be able to be faster, but if that doesn't work it's probably not worth lots of dev time{

spec/models/form/lettings/questions/period_spec.rb
kosiakkatrina: I think it might need to be created (?) cause this doesn't return any rent periods for the managing org at all

arfacamble: oh yeah I looked at the rent_periods method and it actually uses pluck, which will force it to do database things

PR #2464, opened 2024-06-11
CLDC-3493 Update routing for major repairs question

kosiakkatrina:


arfacamble (approved):
there's not really any details or context on this ticket or the code, but it seems like a basic change so I'll just assume it makes sense?

kosiakkatrina:
> there's not really any details or context on this ticket or the code, but it seems like a basic change so I'll just assume it makes sense?

Ah yes, good point, we've added the new options to the question, but didn't add them to the routing, so this PR is doing that!

PR #2465, opened 2024-06-12
CLDC-3492-soft-validations-not-clearing

arfacamble:
https://dluhcdigital.atlassian.net/browse/CLDC-3492

ensure the correct number of beds is calculated if the log is for a bedsit and the beds attribute has been temporarily cleared

the soft validation for rent dependent on la, number of beds, etc appeared not to be clearing for the user.
This was because the log was for a bedsit, which stops routing to the beds question, so beds was cleared, the soft validation page was then not routed to and cleared, then beds was derived as 1 and the soft validation page was routed to again.

the fix is to ensure that the method returning the number of beds which is used to find the expected rent range for a given LA does return 1 if the property is a bedsit

### DIFF
--- a/app/models/lettings_log.rb
+++ b/app/models/lettings_log.rb
@@ -533,7 +533,9 @@ class LettingsLog < Log
   def beds_for_la_rent_range
     return 0 if is_supported_housing?
 
-    beds.nil? ? nil : [beds, LaRentRange::MAX_BEDS].min
+    real_beds = beds || (1 if is_bedsit?)
+
+    real_beds.nil? ? nil : [real_beds, LaRentRange::MAX_BEDS].min
   end
 
   def soft_min_for_period
--- a/spec/models/lettings_log_spec.rb
+++ b/spec/models/lettings_log_spec.rb
@@ -1834,6 +1834,14 @@ RSpec.describe LettingsLog do
         expect(lettings_log.beds_for_la_rent_range).to eq(4)
       end
     end
+
+    context "when the log is for a bedsit, beds is not routed to and is not yet re-derived" do
+      let(:lettings_log) { build(:lettings_log, unittype_gn: 2, beds: nil) }
+
+      it "returns 1" do
+        expect(lettings_log.beds_for_la_rent_range).to eq(1)
+      end
+    end
   end
 
   describe "#collection_period_open?" do


app/models/lettings_log.rb
```diff
@@ -533,7 +533,9 @@ def managing_organisation_name
   def beds_for_la_rent_range
     return 0 if is_supported_housing?
 
-    beds.nil? ? nil : [beds, LaRentRange::MAX_BEDS].min
+    real_beds = beds || (1 if is_bedsit?)
```
kosiakkatrina, line 536: This has both an if and an or, I think it's a bit confusing to read. Would it make sense to change it to `real_beds = is_bedsit? ? 1 : beds` ?
Also, real_beds kinda implies there's a fake_beds, is there a better name for it? The only thing that came to mind for me was current_beds, but it's up to you

kosiakkatrina: I'm now realising it would be a different statement, but I assume there's no reason to prioritise beds before 1 for bedsit (?)

arfacamble: yeah I can't think of a particularly spectacular name since the reason for this code is just general convoluted grossness.
beds_accounting_for_it_being_cleared_and_then_derived but I think that might be overkill

I think reordering it into a ternary is fine, but it makes no difference and is a matter of taste. I gently prefer what I've done since I feel like it reads that by default it should be beds, but then if that's nil we'll check if it should be something.

kosiakkatrina: In that case could we not contribute to the general convoluted grossness and get the value to come in as the correct value to begin with?

I think it might not get cleared if we add this to the beds question:
 ```
 def derived?(log)
    log.is_bedsit?
  end
 ```

Let me know if that works tho

spec/models/lettings_log_spec.rb
```diff
@@ -1834,6 +1834,14 @@ def previous_postcode_fields
...
       end
     end
+
+    context "when the log is for a bedsit, beds is not routed to and is not yet re-derived" do
+      let(:lettings_log) { build(:lettings_log, unittype_gn: 2, beds: nil) }
+
+      it "returns 1" do
+        expect(lettings_log.beds_for_la_rent_range).to eq(1)
+      end
+    end
```
kosiakkatrina, line 1844: So I don't think this test is actually testing anything meaningful. If beds is set to nil here, we're not actually testing the code that clears the beds value and I think if we only build and not create the log that also means that we're not running any of the before_validation callbacks, which from what I'm gathering is what causes the issue in the first place (?)

app/models/lettings_log.rb
arfacamble: this is much cleaner, thanks for pointing me towards this.

spec/models/lettings_log_spec.rb
arfacamble: given that I will now use derived? then this test would go anyway, but if I were doing it that way then I think this is a pretty standard unit test of that method. It's not testing the whole complex series of lifecycle methods that clear and then re-derive values, it's justing testing the one method that I changed and making sure that it behaves in the way that I changed it, ie returning one if the log is for a bedsit regardless of whether `beds` is nil. That method itself doesn't need to be aware of whether it is being called at any particular point in the lifecycle of a log being saved.

RachaelBooth:
Is this PR still necessary? (given https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/2473)

PR #2470, opened 2024-06-14
Cldc 3452 remove dont know option from mortgage used when invalid

arfacamble:
https://dluhcdigital.atlassian.net/browse/CLDC-3452

The work expanded from the details specified in the ticket.
There were a couple of ways to create bugs identified which needed dealing with, some validation methods that were covering similar enough content that I combined them, and some logic which was not covered in tests.

I have also had the testing strategy in mind when working on spec files, but there is of course some flex in how it could be interpreted. I would be keen for feedback of whether we find these tests to be optimally readable.


app/models/form/sales/questions/mortgageused.rb
```diff
@@ -10,21 +10,24 @@ def initialize(id, hsh, subsection, ownershipsch:)
...
 
+  def displayed_answer_options(log, _user = nil)
+    if log.outright_sale? && log.saledate && !form.start_year_after_2024?
+      answer_options_without_dont_know
+    elsif log.stairowned == 100 || log.outright_sale?
+      ANSWER_OPTIONS
+    else
+      answer_options_without_dont_know
+    end
+  end
```
arfacamble, line 21: I know this looks weird with having two branches with the same content but I ended up finding it the most readable.
It's made a bit more complicated by two branches with different outcomes both featuring `:outright_sale?` and trying to combine those two branches I always ended up with some very hard to read negated conditions.

app/models/validations/sales/sale_information_validations.rb
```diff
@@ -147,15 +147,6 @@ def validate_staircasing_mortgage(record)
...
   end
 
-  def validate_mortgage_used_and_stairbought(record)
-    return unless record.stairowned && record.mortgageused
-
-    if !record.stairowned_100? && record.mortgageused == 3
-      record.errors.add :stairowned, I18n.t("validations.sale_information.stairowned.mortgageused_dont_know")
-      record.errors.add :mortgageused, I18n.t("validations.sale_information.stairowned.mortgageused_dont_know")
-    end
-  end
```
arfacamble, line 157: we can see these validations are only triggered when mortgage used is 3, ie don't know, so I combined them into the method below

spec/models/form/sales/questions/mortgageused_spec.rb
arfacamble, line 1: removed all questions just testing eg title is eq "the title of the question"

wrote much more comprehensive tests for the parts of the question model with real logic in them.

arfacamble, line 1: comments on the readability and usability of these tests.
I feel like they score high for readability, eg a context block typically just has the value relevant for that context block below. and the method wrapping the expectation is very readably named.
on the counter I wonder on your instincts of how easy to work with these tests would be if something went wrong.

spec/models/validations/sales/sale_information_validations_spec.rb
arfacamble, line 1: similar questions on usability here.
we said we wanted the object eg log to be defined nearby. Here I define it with default values at the top of a describe block ~100 lines long and the relevant values for any context block are defined clearly.

I feel like this might be pushing it for how far you would need to scroll to find the subject of the test. Given any given context/describe block can grow over time with new edge cases tested, maybe it's better to just redefine sales_log more often.
For the testing policy doc it might be worth coming up with a guide number of lines a block should be where if it gets longer than that you should consider redefingin test subjects at higher levels of indentation?

app/models/validations/sales/sale_information_validations.rb
```diff
@@ -237,16 +228,22 @@ def check_staircasing_non_socialhomebuy_mortgage(record)
...
+    end
+    if record.shared_ownership_scheme? && record.is_not_staircasing?
+      record.errors.add(:mortgageused, I18n.t("validations.invalid_option", question: "was a mortgage used for the purchase of this property?"))
+      record.errors.add(:staircase, I18n.t("validations.financial.mortgage_used.staircasing"))
     end
-    if record.shared_ownership_scheme? && record.staircase && record.staircase != 1
-      record.errors.add(:mortgageused, I18n.t("validations.invalid_option", question: "Was a mortgage used for the purchase of this property?"))
+    if record.stairowned && !record.stairowned_100?
+      record.errors.add :stairowned, I18n.t("validations.sale_information.stairowned.mortgageused_dont_know")
+      record.errors.add :mortgageused, I18n.t("validations.sale_information.stairowned.mortgageused_dont_know")
```
RachaelBooth, line 246: Can there be consistency on whether to use brackets for `record.errors.add` please?

Not sure that I care which to go for tbh, it's just particularly odd to see both ways within one function.

RachaelBooth (approved):
Nice!

app/models/validations/sales/sale_information_validations.rb
arfacamble: oh yes. that's quite yucky, not sure how I've ended up doing that

arfacamble: ah I see I've copied them in from the other method when I combined them and not noticed they were different format

PR #2472, opened 2024-06-17
CLDC-3465 Update hard validations flow

kosiakkatrina:
This is the first part of adding a hard validations flow, I'll keep this as a feature branch, cause there are still quite a few bits that need to be added.

So the biggest challenge here was around maintaining the validation errors, especially after clicking onto the `Clear answer` page and pressing `Cancel`. I think this is what contributed to maybe some odd decisions. Very keen to hear some thoughts and suggestions on overall architecture.

As we trigger a composite error and click into “See all related questions” I've explored doing a get request to a new controller with all the info in the URL (`/lettings-log/{id}/check-errors?original_page_id=net_income&related_question_ids[]=incfreq&related_question_ids=earnings` etc). I didn’t take it far enough to get the validation message on the check answers/errors page. Some ways I thought about doing that was 
- Add the recently submitted values to the URL and revalidate on GET request - feels kinda flimsy, don’t love the idea of value being in the url
- Change this to a POST request - this works okay, but it needs to be outside of the main page form, cause we shouldn’t (can’t?) have nested forms.

I ended up posting things around until an answer is cleared, once any of the answers is cleared, the error message is no longer applicable and we don't need to show it until an error is triggered again in which case the whole process starts from the beginning. 
This uses the same POST request/form as the form submit on the page (save and continue), that grabs all the fieds, revalidates everything again, but instead of rendering the usual form page, it now renders the error summary page with all the errors at the top. I think this works well and we’re reusing a lot of code, but it gets more complicated as we navigate away from the summary page. 
For example if we click on `Clear` on any of the answers, we’re kinda of forced by `summary_list` component for that request to be a GET. I ended up rewriting this component with just html instead of using the component library.

After we have cleared answers and we try answering this just takes us to normal form pages, so there are some parameters being passed around in the url as well so that we can get back to the same set of questions and eventually to the original page where the error got triggered

Although most of the requests are going to the form controller, I've added tests to `spec/requests/check_errors_controller_spec.rb` to separate this by functionality a bit, I don't know if that's better or worse, so let me know!

----
TODO in separate PRs:

- [ ] Make it work when other error gets triggered when answering question from check errors - https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/2480
- [ ] Add clear all button functionality - https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/2483
- [ ] If we clear any answers other than the original answer that caused the error, the original answer also gets cleared, but that's cause it was never persisted. We can either persist it when another answer gets cleared or do not show it as answered to begin with - https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/2484


app/controllers/form_controller.rb
```diff
@@ -374,4 +393,28 @@ def set_managing_organisation_to_assigned_to_organisation?(saledate)
 
     true
   end
+
+  def render_check_errors_page
+    if params[@log.model_name.param_key]["clear_question_id"]
+      question_id = params[@log.model_name.param_key]["clear_question_id"]
+      @log.form.get_question(question_id, @log).page.questions.map(&:id).each { |id| @log[id] = nil }
```
RachaelBooth: Why is this needing to go through all questions on a page?

app/components/check_answers_summary_list_card_component.html.erb
```diff
@@ -35,8 +35,8 @@
 
             <% if @log.collection_period_open_for_editing? %>
               <% row.with_action(
-                text: question.action_text(log),
-                href: action_href(question, log),
+                text: question.action_text(log, correcting_hard_validation: @correcting_hard_validation),
+                href: @correcting_hard_validation ? correct_validation_action_href(question, log, applicable_questions.map(&:id)) : action_href(question, log),
```
RachaelBooth: Can this be built into action_href at all? (I guess needing to pass in applicable_questions might be a problem)

app/controllers/form_controller.rb
```diff
@@ -213,6 +227,11 @@ def successful_redirect_path(pages_to_check)
       return send("#{@log.class.name.underscore}_#{previous_interruption_screen_page_id}_path", @log, { referrer: previous_interruption_screen_referrer, original_log_id: original_duplicate_log_id_from_query }.compact)
     end
 
+    if params[@log.model_name.param_key]["check_errors"]
+      @page = form.get_page(params[@log.model_name.param_key]["page"])
+      return send("#{@log.class.name.underscore}_#{params[@log.model_name.param_key]['original_page_id']}_path", @log, { check_errors: true, related_question_ids: params[@log.model_name.param_key]["related_question_ids"].split(" ") }.compact)
```
RachaelBooth: If you can think of any way to make this a bit easier to read I'd love it, although I have failed to think of anything!

app/controllers/form_controller.rb
kosiakkatrina: I think I answered it on another PR, but it's basically this:
_I think some of our pages have additional validations checking that when there are 2 questions on the page, all of the questions have to be answered. So when we choose to clear a single question from a page with 2 questions, we get a validation error. For example, if we have uprn known - yes and a uprn value, when we choose to clear uprn to correct a composite error, the log would go into an invalid state (another example of this where the questions aren’t conditional is income frequency and earnings for lettings)_

app/components/check_answers_summary_list_card_component.html.erb
kosiakkatrina: Yes! I updated that on this branch: https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/2483

app/controllers/form_controller.rb
kosiakkatrina: Extracted some variables here as well, maybe a bit better: https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/2483

RachaelBooth (approved):
Cool, LGTM (although presumably the other PRs need merging into this one before it's merged to main).

PR #2476, opened 2024-06-20
CLDC-3522: Correct logic around BU address search and address not found errors

RachaelBooth:


spec/services/bulk_upload/lettings/year2024/row_parser_spec.rb
```diff
@@ -104,8 +104,6 @@
       "POSTCODE": "SE2 6RT"
 
          }}]}', headers: {})
-
-      parser.valid?
```
RachaelBooth, line 108: This was causing problems in some tests - the validation had already been run before anything put in a context-specific before hook, which meant some tests were not actually in the setup they looked like they were in.

RachaelBooth: I've essentially just moved the `parser.valid?` call inside all the relevant tests. I've made a few tweaks to a couple of tests that had issues, but left most things alone due to sheer volume.

spec/services/bulk_upload/sales/year2024/row_parser_spec.rb
```diff
@@ -245,8 +245,6 @@ def purch_id_field
 
       stub_request(:get, /api\.os\.uk/)
         .to_return(status: 200, body: { results: [{ DPA: { MATCH: 0.9, BUILDING_NAME: "result address line 1", POST_TOWN: "result town or city", POSTCODE: "AA1 1AA", UPRN: "12345" } }] }.to_json, headers: {})
-
-      parser.valid?
```
RachaelBooth, line 249: Same deal as lettings.

app/services/bulk_upload/lettings/year2024/row_parser.rb
```diff
@@ -584,19 +584,23 @@ def assigned_to
...
+    if log.uprn.nil? && field_16.blank? && key_address_fields_provided?
       %i[field_17 field_18 field_19 field_20 field_21 field_22].each do |field|
         errors.add(field, I18n.t("validations.no_address_found"))
       end
     end
   end
 
+  def key_address_fields_provided?
+    field_17.present? && field_19.present? && field_21.present? && field_22.present?
+  end
```
kosiakkatrina, line 602: Postcode (field_21 and field_22) are optional (cause we could use the LA instead), are we sure we want to validate it like this? What values would need to be in the BU file to upload a successful, completed log without providing a postcode or UPRN?

kosiakkatrina, line 602: Postcode (field_21 and field_22) are optional (cause we could use the LA instead), do we need to account for LA as well? What values would need to be in the BU file to upload a successful, completed log with LA but without providing a postcode or UPRN?

RachaelBooth: The BU spec says that the postcode fields can only be empty (for general needs) if both the UPRN and the LA are provided, so I think you can't (which does match what we insist on for address search elsewhere). 

(It also says that for general needs the LA must always be provided, hence all this implementation actually just looks at the UPRN/address fields combination)

kosiakkatrina: Ah, I see. It seems that in the UI, it's possible to have only the LA, address line 1, and town/city fields filled in and have that section marked as complete. In BU it looks like we can't do the same (?). Should we aim to keep this consistent with what's possible through the UI?

If not, I'd like to understand the reasoning behind allowing different criteria based on the record's source. Also, if the BU file only has LA, address line 1, and town/city provided, and the user chooses to "fix errors in service," would that mean there's actually nothing to fix? Since the UI would already consider it complete, it seems like there errors might be a bit misleading/confusing?

RachaelBooth: We do insist on more search information for BU (slightly in that we insist on town_or_city) than we do in the UI - but that makes sense because we will automatically pick the best match (assuming a good enough match) for BU but in the UI we get the user to confirm.

We do insist on the postcode in the UI in order to do an address search. The only way you can end up with different information on a completed log is by the user actively selecting to enter the address manually through the UI if the address isn't listed in search results (in which case the postcode is not required - although the page makes it look like it is and you've previously been forced to give something for search so that may be a bug). There is a something a bit weird right now if you don't give a postcode and choose to fix in the service though, in that the address selection isn't answered so some of the routing is a bit off - I'll look into that - but the log doesn't show up as complete and the user would need to go in and fix the address.

The BU spec should be the source of truth for what's possible in BU, so anything changing from that would need discussion with Sam/Rachel (and it's not immediately clear whether an inconsistency would mean changing BU or changing the UI).

kosiakkatrina: Okay, I guess that makes sense. I feel like the UI flow being so interactive makes the BU a little bit confusing, but I understand a bit better why we're asking for more info now. Thanks!

kosiakkatrina (approved):

PR #2477, opened 2024-06-21
CLDC-3523 Only count visible logs on CSV downloads page

kosiakkatrina:


spec/requests/lettings_logs_controller_spec.rb
```diff
@@ -1865,6 +1865,23 @@
...
         expect(response).to have_http_status(:unauthorized)
       end
+
+      context "when filtering by organisation and year" do
+        let(:other_organisation) { FactoryBot.create(:organisation) }
+        let(:lettings_logs) { create_list(:lettings_log, 2, :setup_completed, assigned_to: user, owning_organisation: other_organisation, managing_organisation: user.organisation) }
+
+        before do
+          create(:organisation_relationship, parent_organisation: other_organisation, child_organisation: user.organisation)
+          lettings_logs
```
arfacamble: since this is the only place `lettings_logs` is called I would have a gentle preference for removing 1871 and: 
```suggestion
          create_list(:lettings_log, 2, :setup_completed, assigned_to: user, owning_organisation: other_organisation, managing_organisation: user.organisation)
```

I would even not see nay particular need to have a before block only for one test and just put it all in the it.

I feel more strongly about the first of the above and I feel very weakly about both of them

spec/requests/lettings_logs_controller_spec.rb
```diff
@@ -1865,6 +1865,23 @@
...
+          create_list(:lettings_log, 2, :setup_completed, assigned_to: user, owning_organisation: other_organisation, managing_organisation: user.organisation, discarded_at: Time.zone.yesterday)
+        end
+
+        it "does not count deleted logs" do
+          get "/lettings-logs/csv-download?years[]=#{lettings_logs[0].form.start_date.year}&codes_only=false&owning_organisation_select=specific_org&owning_organisation=#{other_organisation.id}", headers:, params: {}
```
arfacamble: I feel marginally stronger that we should use rails route helpers in our tests
```suggestion
          params = {
              "years[]" => lettings_logs[0].form.start_date.year,
              other ones,
              "owning_org" => other_organisation.id,
          }
          get lettings_logs_csv_download_path, headers:, params:
```

arfacamble (approved):

spec/requests/lettings_logs_controller_spec.rb
kosiakkatrina: Sure!

spec/requests/lettings_logs_controller_spec.rb
kosiakkatrina: I think it was mostly to get the year in the test`lettings_logs[0].form.start_date.year`, but now that you've mentioned it, I can probably remove it from here in general, cause it lettings_logs would get created when calling the url!

PR #2478, opened 2024-06-21
CLDC-3505 refactor tests that will break when collection year hard closes

arfacamble:
refactor property validations specs to avoid breaking on collection year change

### DIFF
--- a/app/models/validations/property_validations.rb
+++ b/app/models/validations/property_validations.rb
@@ -33,20 +33,20 @@ module Validations::PropertyValidations
   end
 
   def validate_shared_housing_rooms(record)
-    unless record.unittype_gn.nil?
-      if record.is_bedsit? && record.beds != 1 && record.beds.present? && !record.form.start_year_after_2024?
-        record.errors.add :unittype_gn, I18n.t("validations.property.unittype_gn.one_bedroom_bedsit")
-        record.errors.add :beds, I18n.t("validations.property.unittype_gn.one_bedroom_bedsit")
-      end
+    return unless record.unittype_gn
 
-      if record.hhmemb == 1 && record.is_shared_housing? &&
-          !record.beds.to_i.between?(1, 3) && record.beds.present?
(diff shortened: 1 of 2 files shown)


app/models/validations/property_validations.rb
arfacamble, line 1: the only thing I have changed here is to reduce nesting by changing line 36 to a guard clause, everything else has just been unindented and github hasn't worked it out

spec/models/validations/property_validations_spec.rb
arfacamble, line 1: there's a bit of noise in this diff since I changed record to log everywhere I will highlight any meaningful changes

spec/models/validations/property_validations_spec.rb
```diff
@@ -4,194 +4,222 @@
   subject(:property_validator) { property_validator_class.new }
 
   let(:property_validator_class) { Class.new { include Validations::PropertyValidations } }
-  let(:record) { FactoryBot.create(:lettings_log, startdate: Time.zone.local(2024, 3, 3)) }
+  let(:log) { build(:lettings_log) }
```
arfacamble, line 7: build not create, remove startdate

spec/models/validations/property_validations_spec.rb
```diff
@@ -4,194 +4,222 @@
...
+
+      context "and the log is from before 24/25" do
+        it "adds an error" do
+          allow(log.form).to receive(:start_year_after_2024?).and_return false
+
+          property_validator.validate_shared_housing_rooms(log)
+
+          expect(log.errors["unittype_gn"]).to include(I18n.t("validations.property.unittype_gn.one_bedroom_bedsit"))
+          expect(log.errors["beds"]).to include(I18n.t("validations.property.unittype_gn.one_bedroom_bedsit"))
+        end
```
arfacamble, line 85: whether or not errors are added in this case was changed between years in 3135 but not tested.

kosiakkatrina (approved):

PR #2483, opened 2024-06-26
CLDC-3465 Allow clearing all error questions

kosiakkatrina:
This builds on top of https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/2472
and adds functionality to clear all button/link

### DIFF
--- a/app/controllers/check_errors_controller.rb
+++ b/app/controllers/check_errors_controller.rb
@@ -11,7 +11,13 @@ class CheckErrorsController < ApplicationController
     @page = @log.form.get_page(params[@log.model_name.param_key]["page_id"])
 
     if params["clear_all"]
-      @questions_to_clear = @related_question_ids.map { |id| @log.form.get_question(id, @log).page.questions.map(&:id) }.flatten
+      @questions_to_clear = @related_question_ids.map { |id|
+        question = @log.form.get_question(id, @log)
+        next if question.subsection.id == "setup"
+
+        question.page.questions.map(&:id)
+      }.flatten.compact
+
       render :confirm_clear_all_answers
     else
       question_id = @related_question_ids.find { |id| !params[id].nil? }
--- a/app/controllers/check_errors_controller.rb
+++ b/app/controllers/check_errors_controller.rb
@@ -8,9 +8,15 @@ class CheckErrorsController < ApplicationController
     return render_not_found unless @log
 
     @related_question_ids = params[@log.model_name.param_key].keys.reject { |id| id == "page_id" }
-    question_id = @related_question_ids.find { |id| !params[id].nil? }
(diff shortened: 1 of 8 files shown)


app/controllers/check_errors_controller.rb
```diff
@@ -8,9 +8,21 @@ def confirm_clear_answer
...
-    question_id = @related_question_ids.find { |id| !params[id].nil? }
-    @question = @log.form.get_question(question_id, @log)
     @page = @log.form.get_page(params[@log.model_name.param_key]["page_id"])
+
+    if params["clear_all"]
+      @questions_to_clear = @related_question_ids.map { |id|
+        question = @log.form.get_question(id, @log)
+        next if question.subsection.id == "setup"
+
+        question.page.questions.map(&:id)
```
RachaelBooth, line 18: Should this be just question? I know we don't put more than one on a page very often (and only if they are linked), but I'd expect others on the same page to be related in their own right if they needed to be cleared. (Maybe that expectation is wrong though)

app/views/check_errors/confirm_clear_all_answers.html.erb
```diff
@@ -0,0 +1,32 @@
+<% content_for :before_content do %>
+  <% content_for :title, "Are you sure you want to clear all?" %>
+<% end %>
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds-from-desktop">
+    <h1 class="govuk-heading-xl">
+      <%= content_for(:title) %>
+    </h1>
+    <p class="govuk-body">You've selected <%= @questions_to_clear.count %> answers to clear</p>
```
RachaelBooth, line 10: Do we need to have anything that takes into account anything that will be cleared as a result of clearing these? (That sounds really hard to identify, so maybe just a warning?)

app/controllers/check_errors_controller.rb
kosiakkatrina: So there kinda is a reason I'm doing this. I think some of our pages have additional validations checking that when there are 2 questions on the page, all of the questions have to be answered. So when we choose to clear a single question from a page with 2 questions, we get a validation error. For example, if we have uprn known - yes and a uprn value, when we choose to clear uprn to correct a composite error, the log would go into an invalid state (another example of this where the questions aren’t conditional is income frequency and earnings for lettings)

app/views/check_errors/confirm_clear_all_answers.html.erb
kosiakkatrina: Like conditional questions? Yeah, maybe a warning, let me check with design

RachaelBooth (approved):
LGTM, subject to whatever design say about the warning

PR #2485, opened 2024-06-26
Cldc 3505 fix tests that will break on rollover

arfacamble:


spec/factories/lettings_log.rb
```diff
@@ -23,15 +23,6 @@
...
     end
-    trait :conditional_section_complete do
-      tenancycode { Faker::Name.initials(number: 10) }
-      age1 { 34 }
-      sex1 { "M" }
-      ethnic { 2 }
-      national { 18 }
-      ecstat1 { 2 }
-      hhmemb { 1 }
-    end
```
arfacamble, line 34: this was only being used in one place in a test I refactored, unless someone disagrees I don't find this trait name to describe something meaningful, but I might be missing some context for what conditional section means

spec/factories/lettings_log.rb
```diff
@@ -168,14 +159,15 @@
...
-      uprn_selection { 1 }
+      after(:build) do |log, _evaluator|
+        if log.startdate >= Time.zone.local(2024, 4, 1)
+          log.address_line1_input = log.address_line1
+          log.postcode_full_input = log.postcode_full
+          log.nationality_all_group = 826
+          log.uprn = "10033558653"
+          log.uprn_selection = 1
+        end
+      end
```
arfacamble, line 170: this is within the :completed trait and makes 24/25 logs with the completed trait actually be completed.

spec/features/form/form_navigation_spec.rb
arfacamble, line 1: please check the changes in this file.
I don't immediately see that the log needs to be completed in order to test address searching. Is that correct, setup complete makes more sense to me intuitively, being complete seemed to redirect to a review this log page.

Also the extra in the query string I have just copied from what it really is since it's just another key value pair in the query it doesn't strike me as unacceptable but I'm not 100% sure why it would be one way or another

spec/models/log_spec.rb
arfacamble, line 1: @kosiakkatrina this will have conflicts with one of your PRs. I needed to remove the date which you hadn't and I've found a way to add even less to the DB, using what I saw on your last PR setting the id when building in the Factory.

spec/presenters/homepage_presenter_spec.rb
```diff
@@ -176,7 +176,8 @@
           let(:startdate) { date_last_year }
 
           it "exposes the correct data for the data box" do
-            create_list(type, expected_count, status, assigned_to: user, startdate:)
+            logs = build_list(type, expected_count, status, assigned_to: user, startdate:)
+            logs.each { |log| log.save(validate: false) }
```
arfacamble, line 180: do we think this is acceptable? it feels a little iffy to me to skip all validations when all I really want to do is skip the date validation. But it would start getting quite verbose if I do create_list, each, update date only, save(validate: false)
One reason I was kind of fine with this in this case is we create_list with everything the same other than startdate elsewhere in the file, but there's nothing to ensure that that is always so.

spec/requests/lettings_logs_controller_spec.rb
arfacamble, line 1: since there were quite a few broken tests here and they were mixed up with freezing in the fairly distant past this section has ended up being quite a fulsome refactor

spec/requests/lettings_logs_controller_spec.rb
```diff
@@ -1153,119 +1144,54 @@
...
 
-            around do |example|
-              FormHandler.instance.use_real_forms!
-              example.run
-              FormHandler.instance.use_fake_forms!
+            it "does not show property information" do
+              get lettings_log_path(lettings_log)
+              expect(page).to have_content "Tenancy information"
+              expect(page).not_to have_content "Property information"
             end
```
arfacamble, line 1191: I added this test in since it was specified in the name of the context block it's in but not tested.

spec/requests/lettings_logs_controller_spec.rb
```diff
@@ -1153,119 +1144,54 @@
...
-
-            after do
-              Timecop.unfreeze
-              Singleton.__init__(FormHandler)
-            end
-
-            it "displays a section status for a lettings log" do
-              assert_select ".govuk-tag", text: /Not started/, count: 6
-              assert_select ".govuk-tag", text: /Completed/, count: 1
-              assert_select ".govuk-tag", text: /Cannot start yet/, count: 1
```
arfacamble, line 1268: shout if you think this is testing something important, but it seems to be fairly closely repeating another test above, checking numbers of statuses of sections. again I may be missing context, but this is the only place using the :conditional_section_complete trait that I removed.

app/models/form.rb
```diff
@@ -11,6 +11,8 @@ class Form
     },
     2023 => {
       submission_deadline: Time.zone.local(2024, 6, 7),
+      # new_logs_end_date: Time.zone.local(2024, 6, 19),
+      # edit_end_date: Time.zone.local(2024, 6, 19),
```
kosiakkatrina: I assume this will go?

spec/factories/lettings_log.rb
kosiakkatrina: Nice

spec/features/form/form_navigation_spec.rb
```diff
@@ -211,15 +211,15 @@
         fill_in("lettings-log-address-line1-input-field", with: "address")
         fill_in("lettings-log-postcode-full-input-field", with: "A1 1AA")
         click_button(text: "Search")
-        expect(page).to have_current_path("/lettings-logs/#{id}/uprn-selection?referrer=check_answers")
+        expect(page).to have_current_path("/lettings-logs/#{id}/uprn-selection?referrer=check_answers&unanswered_pages=property_local_authority")
```
kosiakkatrina, line 214: How is this changing the test?

spec/presenters/homepage_presenter_spec.rb
kosiakkatrina: We used to have a feature flag for date validations, if we bring it back we could set it here

Otherwise we could create the logs and then update the dates and only skip the validations on update, but I get how that could be problematic

Another options could be adding something similar like `skip_dpo_validation` but idk if I love that

spec/factories/lettings_log.rb
RachaelBooth: Are there any pre-2024 only fields we'd want this to clear?

spec/presenters/homepage_presenter_spec.rb
RachaelBooth: We've got an `ignore_validation_errors` trait for lettings logs that might make this a bit nicer, although still has the problem of ignoring everything - would be very up for a trait that just skipped date validation.

spec/factories/lettings_log.rb
arfacamble: I'm not aware, but the test I have put this to is that if I create logs for before or after, then after being saved, validated, derived, cleared, etc, then they do come back with the correct status

spec/features/form/form_navigation_spec.rb
arfacamble: I'm not sure why my comment on this change is marked as outdated but it is here
https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/2485#discussion_r1654941111

as there, it doesn't strike me, but I would appreciate any further context, that adding something to the query string is a problem. But I will look into it a little further

spec/presenters/homepage_presenter_spec.rb
arfacamble: hmmm. I can't immediately think of a way of doing that due to deriving clearing etc
if you tried to save it with the date as nil, then everything after setup would be cleared
if you saved with today's date, then updated it without validation, then you could get different values derived on the first save which could be very confusing to debug one day.

arfacamble: actually I will have a quick go to see if you can use TimeCop in the factory

arfacamble: it is possible, and it works fine in conjunction with in progress logs on first pass, but ti doesn't play nicely with the completed trait, so I'm not sure if it isn't just going to introduce more confusion when working with tests, open to thoughts. For reference the code was:
```
    trait :skip_startdate_validation do
      to_create do |log|
        Timecop.freeze(log.startdate) do
          log.save!
        end
      end
    end

```

kosiakkatrina (approved):

PR #2491, opened 2024-06-27
Cldc 3505 fix tests that will break on rollover

arfacamble:


spec/models/bulk_upload_spec.rb
arfacamble, line 1: these soft validation pages are added for 2024 I assume they all need answering in bulk upload

spec/models/form_handler_spec.rb
arfacamble, line 1: the code in the form handler indicates that once we are no longer in the crossover period, the archived form no longer needs to be provided by the form handler.
what would we like to do with tests like this?
we can either let them break on the hard year end and it will take two seconds to delete them when that happens, then we have coverage until that point.
Or I just delete them now.
Or I make a PR with a title that says MERGE ON HARD YEAR END

spec/requests/form_controller_spec.rb
arfacamble, line 1: @kosiakkatrina I've put this TODO here that relates to the removing unnecessary time freezing in tests.

spec/models/form_handler_spec.rb
RachaelBooth: Yeah, I think ideally we have coverage for now given the test exists and get rid of it later - happy for that to be a thing on the todo list for hard year end, although we should put a reminder somewhere.

RachaelBooth (approved):

PR #2493, opened 2024-06-28
CLDC-3505 further fixes to tests for year hard end

arfacamble:


spec/services/csv/sales_log_csv_service_spec.rb
arfacamble, line 1: this one I did not have time to think about in depth
csv tests are obviously a bit fiddly since they do want to be frozen to a year in many cases, values do have to be fixed firm including time values so they can be reflected in the expectation fixtures, etc etc

I'm not confident I have made these properly robust into the future, but they will not break in this state at this year's hard end

RachaelBooth (approved):
Will merge this with the minor review fixes I've done and raise a new PR for the remaining things to do - I think that'll make it easier to keep track of / review.

PR #2494, opened 2024-07-02
CLDC-3543: Task to handle unpended logs

RachaelBooth:


RachaelBooth:
Given lack of people around, after some manual testing I'm merging a safe version of this (i.e. dry runs only) in order to get that run despite lack of 3rd-person testing/code review. I'll ensure it's all looked at properly before merging a non-safe version.

PR #2496, opened 2024-07-05
CLDC-3543: Allow live run of handle unpended logs task

RachaelBooth:


RachaelBooth:
Github will make commenting annoying for this, but the whole file does want a review (we merged the dry run version early in order to run that). Although with the context of being a one off task this is mostly a check that this isn't doing something stupid, because we'll want to get rid of the code later.

kosiakkatrina (approved):

PR #2497, opened 2024-07-05
CLDC-3520 order scheme locations

Dinssa:


app/models/form/lettings/questions/location_id.rb
```diff
@@ -30,7 +30,7 @@ def displayed_answer_options(lettings_log, _user = nil)
     return {} unless lettings_log.scheme
 
     scheme_location_ids = lettings_log.scheme.locations.visible.confirmed.pluck(:id)
-    answer_options.select { |k, _v| scheme_location_ids.include?(k.to_i) }
+    answer_options.select { |k, _v| scheme_location_ids.include?(k.to_i) }.sort_by { |_k, v| v["hint"] }
```
RachaelBooth, line 33: This looks like it might need to change in order to hit the AC about ordering when this starts with a number.

PR #2498, opened 2024-07-05
CLDC-3547: Validate against incomplete schemes being used in logs

RachaelBooth:


RachaelBooth:
This doesn't currently include a script to clear data on logs that currently use an incomplete scheme (and have been allowed to do so because weirdly it does have active locations). There are a few hundred of these, all linked to one of 3 schemes - I've raised with Rachel that we should probably start by giving users the opportunity to complete the schemes as that would fix the problem and clearing data on logs would mean they had a lot to manually re-enter. Will hold off on any scripts until that's resolved.

spec/services/bulk_upload/lettings/year2023/row_parser_spec.rb
```diff
@@ -86,8 +86,6 @@
     before do
       stub_request(:get, /api\.postcodes\.io/)
       .to_return(status: 200, body: "{\"status\":200,\"result\":{\"admin_district\":\"Manchester\", \"codes\":{\"admin_district\": \"E08000003\"}}}", headers: {})
-
-      parser.valid?
```
RachaelBooth, line 87: This was causing ordering issues, since it gets called before some setup done in nested contexts. I've moved to calling it in the tests themselves so that all setup has happened first.

app/models/validations/setup_validations.rb
```diff
@@ -90,6 +90,10 @@ def validate_scheme_has_confirmed_locations_validation(record)
   end
 
   def validate_scheme(record)
+    if record.scheme&.status == :incomplete
+      record.errors.add :scheme_id, :incomplete, message: I18n.t("validations.setup.scheme.incomplete")
+    end
+
```
kosiakkatrina, line 96: Would this invalidate those logs that have incomplete scheme assigned to them for some time?
I assume we're not showing incomplete schemes in the scheme dropdown and this happened by setting the schemes somewhere else? Any idea where?

RachaelBooth: It will invalidate logs with an incomplete scheme assigned to them - I'm waiting for an answer from Rachel around giving users the opportunity to just complete the scheme first (rather than clearing data) before I write anything to deal with that.

Logs could have got to that state via bulk upload, or possibly import from old core.

I think we don't show schemes where the scheme data isn't complete for the scheme id question, but they can also be incomplete if they have no complete locations and I think in that case we would show them (the filter is based on the 'confirmed' property) - this seems fine in terms of the user seeing why they can't select a particular thing, so I left it alone.

kosiakkatrina (approved):

PR #2499, opened 2024-07-08
CLDC-3520 order scheme locations

Dinssa:


app/models/form/lettings/questions/location_id.rb
```diff
@@ -31,6 +31,11 @@ def displayed_answer_options(lettings_log, _user = nil)
 
     scheme_location_ids = lettings_log.scheme.locations.visible.confirmed.pluck(:id)
     answer_options.select { |k, _v| scheme_location_ids.include?(k.to_i) }
+                  .sort_by { |_, v|
+                    name = v["hint"].match(/[a-zA-Z].*/).to_s
+                    number = v["hint"].match(/\d+/).to_s.to_i
+                    [name, number]
+                  }.to_h
```
RachaelBooth, line 38: Can you add some tests for this ordering?

Dinssa: Done

RachaelBooth (approved):
LGTM - if you move the ticket to PO review and copy the link to the review app (on this PR) into a comment on the ticket, then someone will test it and you can merge once they've moved it to ready for release.

app/models/lettings_log.rb
```diff
@@ -706,6 +706,20 @@ def process_postcode_changes!
     self.la = inferred_la if inferred_la.present?
   end
 
+  def scheme_has_multiple_locations?
```
RachaelBooth, line 709: There's more to do to properly fix the merge conflict here - I edited this function in the other file, while you've moved it here, and I think you've now got a different version of it in each file?

Git has just shown the conflict it knows about - i.e. lines where you removed something and main had changed it - it doesn't know that you'd moved the function to here.

With that kind of conflict, you need to look at the actual changes and ensure you've built them in to your work - so remove the version in the other file (because you've moved the function) but replicate the changes that had been made to it here so they aren't lost.

You may need similar changes to the `scheme_has_large_number_of_locations?` function below as well.

Dinssa: Amended

PR #2501, opened 2024-07-08
CLDC-3550 Fix rent_type 2024 BU mapping

kosiakkatrina:
Our BU specification for rent_type is:
  1 = Social Rent
  2 = Affordable Rent
  3 = London Affordable Rent
  4 = Rent to Buy
  5 = London Living Rent
  6 = Other intermediate rent product
Internally we save those values as:
  0 = Social Rent
  1 = Affordable Rent
  2 = London Affordable Rent
  3 = Rent to Buy
  4 = London Living Rent
  5 = Other intermediate rent product
And there isn’t any mapping going on. So the rent_type on basically all of the 2024 BU logs (unless the rent_type on them has been edited since) would be wrong.

This should fix the issue on BU and exiting logs. 

Separately, might be worth updating all the rent_type values to be 1-6 internally, as we export them out as 1-6 anyway


lib/tasks/correct_rent_type_value.rake
```diff
@@ -0,0 +1,20 @@
+desc "Alter rent_type values for bulk uploaded lettings logs for 2024 where they were not mapped correctly"
+task correct_rent_type_value: :environment do
+  affected_uploads = BulkUpload.where(log_type: "lettings", year: 2024)
```
RachaelBooth: Is there a way to ensure that there aren't any uploads processed between the fix being deployed and running this task?

lib/tasks/correct_rent_type_value.rake
```diff
@@ -0,0 +1,20 @@
...
+  affected_uploads.each do |upload|
+    upload.logs.where.not(rent_type: nil).each do |log|
+      current_rent_type = log.rent_type
+      rent_type_at_upload = log.versions.length == 1 ? log.rent_type : log.versions.first.next.reify.rent_type
+      next unless rent_type_at_upload == current_rent_type
+
+      new_rent_type_value = BulkUpload::Lettings::Year2024::RowParser::RENT_TYPE_BU_MAPPING[rent_type_at_upload]
+      log.rent_type = new_rent_type_value
```
RachaelBooth, line 11: Is there any risk of this change causing other data to be cleared? (I think probably not - not sure anything depends on rent type, just want to be sure)

kosiakkatrina: I'm like 95% sure it should be fine. I couldn't find things dependent on the `rent_type` or `lettype`, we also get `lar` and `irproduct` from `rent_type`, which don't look like should be an issue. There's a conditional on `irproduct_other` but only if it's a 5. I don't think we'll be updating any `rent_type` to a 5 cause there was no way to set it to 6 in the first place.

We do also derive `renttype` which is used for tenancy length validation (and routing, but routing is only a different wording, still same question), which from what I checked before writing this, none of the logs should have failed.

The tests also check that the logs don't go from completed to in progress, but obviously, that's with a very specific combination of answers.

lib/tasks/correct_rent_type_value.rake
kosiakkatrina: Hmm, good point, would it make sense to add something similar to what you did with noint?

RachaelBooth: Probably - I've been pondering if there's a better way, but can't think of anything short of turning off sidekiq for a bit which seems like a bad plan.

kosiakkatrina: Agree! 
I've updated it now to have the rent_type_fix_status. I've also removed noint_fix_status, and the job to fix noint just as a cleanup, but let me know if we still need that info and I'll revert this!

db/migrate/20240715072421_add_rent_fix_status.rb
```diff
@@ -0,0 +1,8 @@
+class AddRentFixStatus < ActiveRecord::Migration[7.0]
+  def change
+    change_table :bulk_uploads, bulk: true do |t|
+      t.column :rent_type_fix_status, :string, default: "not_applied"
+    end
+    remove_column :bulk_uploads, :noint_fix_status, :string, default: "not_applied"
+  end
```
kosiakkatrina, line 7: This is kinda intentionally weird, not updating the table in bulk triggers rubocop and adding t.remove into the change_table creates an irreversible migration (even when the type is given). I feel like maybe there's some obvious piece of syntax I'm missing, otherwise this does what I'd like it to do

lib/tasks/correct_rent_type_value.rake
kosiakkatrina: Ooh, I just realised there's a ticket for that and it would need 2 separate releases. I'll do that instead.
I only did the first part of removing noint here, and this is removing it from the db: https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/2508

Will merge/deploy separately after this one is out

RachaelBooth (approved):

PR #2502, opened 2024-07-09
CLDC-3549: Ensure location page routing is updated when log is changed

RachaelBooth:


spec/models/lettings_log_spec.rb
```diff
@@ -486,18 +486,56 @@ def previous_postcode_fields
       end
 
       describe "when changing a log's scheme and hence calling reset_scheme_location!" do
-        let(:scheme) { FactoryBot.create(:scheme) }
-        let(:invalid_location_1) { FactoryBot.create(:location, scheme:, startdate: Time.zone.today + 3.weeks) }
-        let(:valid_location) { FactoryBot.create(:location, scheme:, startdate: Time.zone.yesterday) }
-        let(:invalid_location_2) { FactoryBot.create(:location, scheme:, startdate: Time.zone.today + 3.weeks) }
+        before do
+          Timecop.return
```
RachaelBooth, line 490: This file still has some wide ranging dates-in-the-far-past stuff going on - I've left it alone to not interfere with anything currently going on elsewhere to fix that, but we should make sure it gets sorted.

spec/models/lettings_log_spec.rb
```diff
@@ -486,18 +486,56 @@ def previous_postcode_fields
...
+        context "when there are many valid locations in the new scheme" do
+          let(:old_scheme) { create(:scheme, owning_organisation:) }
+          let(:old_location) { create(:location, scheme: old_scheme) }
+          let(:new_scheme) { create(:scheme, owning_organisation:) }
+
+          before do
+            create_list(:location, 2, scheme: new_scheme)
+          end
+
+          context "with a 2023 log" do
```
RachaelBooth, line 515: It was 2023 logs that we've seen this bug on, so it seemed worth having a test that explicitly covers that.

app/models/derived_variables/lettings_log_variables.rb
```diff
@@ -31,8 +31,8 @@ module DerivedVariables::LettingsLogVariables
   def scheme_has_multiple_locations?
     return false unless scheme
 
-    @scheme_locations_count ||= scheme.locations.active_in_2_weeks.size
-    @scheme_locations_count > 1
+    scheme_locations_count = scheme.locations.active_in_2_weeks.size
+    scheme_locations_count > 1
```
kosiakkatrina, line 35: Would it make sense to recalculate it only if `scheme_was != scheme` ?

RachaelBooth: Potentially, although I was thinking that if someone added a location to the scheme that could also change this (that'll become more relevant in Manny's current work).

kosiakkatrina: Hmm, I think I understand, I'm curious whether this would cause much of a difference in terms of performance. Happy with this if that's the correct thing to do tho

kosiakkatrina (approved):

PR #2505, opened 2024-07-11
Split tests into more jobs

kosiakkatrina:


.github/workflows/staging_pipeline.yml
```diff
@@ -72,7 +72,7 @@ jobs:
 
       - name: Run tests
         run: |
-          bundle exec rake parallel:spec['spec\/(?!features)']
+          bundle exec rake parallel:spec['spec\/(?!features|models|requests)']
```
RachaelBooth, line 75: Does this want to use `--fail-fast` as well? Or actually do we only want that for slower tests (in which case I think models might not be in that category)

RachaelBooth (approved):
I like it - vaguely tempted to pull tests out into a reusable workflow to keep this file a bit more under control, but it's probably not worth the effort.

.github/workflows/staging_pipeline.yml
kosiakkatrina: Yeah, good point. So this was the original motivation for adding --fail-fast:
_Adds the --fail-fast flag to the pipelines to free up the runners if a test fails. This will be particularly useful if the commit is already known to have failing tests or for stopping right away on any flaky tests_

So if we agree with this logic, it probably makes sense to add everywhere. I'll do that, but lmk if you can think of a reason against it

RachaelBooth: The only reason I can think of is that it means you don't necessarily see all the failures at once - but tbh I think failing fast is more useful.

kosiakkatrina: Okay, so I think the reason this specific job didn't have it is because we're running the tests in parallel, so I'm not sure how the flag would work in this case. Saying that, updating request tests (cause they take the longest atm) to also run in parallel instead of failing fast further reduces staging pipeline 13min ~> 10min in the case where it's successful.

Should I leave it like that instead?

RachaelBooth: Yeah, that sounds good to me!

PR #2509, opened 2024-07-15
CLDC-3505: Fix tests for hard year end part 4

RachaelBooth:


spec/factories/sales_log.rb
```diff
@@ -128,8 +128,7 @@
       pregla { 1 }
       pregother { 1 }
       pregghb { 1 }
-      hhregres { 1 }
-      hhregresstill { 4 }
+      hhregres { 7 }
```
RachaelBooth, line 131: This made some debugging easier, and exposed a few places that shouldn't have been relying on the old values so I fixed those and left the change in.

spec/models/sales_log_spec.rb
```diff
@@ -564,37 +564,6 @@ def check_property_postcode_fields
       expect(record_from_db["la"]).to eq("E08000003")
     end
 
-    context "with 23/24 logs" do
```
RachaelBooth, line 567: This became remarkably painful to try and fix, and becomes irrelevant so didn't seem worth too much effort.

spec/services/bulk_upload/lettings/log_creator_spec.rb
```diff
@@ -6,37 +6,21 @@
   let(:owning_org) { create(:organisation, old_visible_id: 123, rent_periods: [2]) }
   let(:user) { create(:user, organisation: owning_org) }
 
-  let(:bulk_upload) { create(:bulk_upload, :lettings, user:) }
-  let(:csv_parser) { instance_double(BulkUpload::Lettings::Year2023::CsvParser) }
-  let(:row_parser) { instance_double(BulkUpload::Lettings::Year2023::RowParser) }
+  let(:bulk_upload) { create(:bulk_upload, :lettings, user:, year: 2024) }
+  let(:csv_parser) { instance_double(BulkUpload::Lettings::Year2024::CsvParser) }
+  let(:row_parser) { instance_double(BulkUpload::Lettings::Year2024::RowParser) }
```
RachaelBooth, line 11: This can't avoid being slightly tied to year in this sense, annoyingly.

spec/services/bulk_upload/lettings/year2023/row_parser_spec.rb
```diff
@@ -30,7 +30,8 @@
   end
 
   before do
-    allow(FormHandler.instance).to receive(:lettings_in_crossover_period?).and_return(true)
+    Timecop.freeze(now)
```
RachaelBooth, line 33: I don't think there's a way round time travel in this case - and given it's a 2023 specific parser that seems fine.

spec/support/bulk_upload/lettings_log_to_csv.rb
```diff
@@ -12,6 +12,62 @@ def row_prefix
     [nil] * col_offset
   end
 
+  def to_csv_row(seed: nil)
+    year = log.collection_start_year
+    case year
+    when 2022
```
RachaelBooth, line 18: Not sure the 2022 case is really necessary for these, but some of the 2023 functions rely on the 2022 ones so I didn't want to get rid of them - at which point this case may as well be here.

spec/support/bulk_upload/sales_log_to_csv.rb
```diff
@@ -357,7 +413,7 @@ def to_2024_row
       log.proplen,
       log.value,
       log.grant,
-      log.discount,
+      log.discount || 0,
```
RachaelBooth, line 416: I think there may be a bug around this being required when we might expect grant instead, but I'll raise that separately and leave this here for now.

app/models/form.rb
```diff
@@ -11,6 +11,8 @@ class Form
     },
     2023 => {
       submission_deadline: Time.zone.local(2024, 6, 7),
+      new_logs_end_date: Time.zone.local(2024, 7, 14),
+      edit_end_date: Time.zone.local(2024, 7, 14),
```
RachaelBooth: I'll take these out before merge but seems easier to leave them in until everything else is sorted.

spec/components/bulk_upload_error_row_component_spec.rb
```diff
@@ -6,9 +6,9 @@
     let(:tenant_code) { SecureRandom.hex(4) }
     let(:property_ref) { SecureRandom.hex(4) }
     let(:purchaser_code) { nil }
-    let(:field) { :field_134 }
+    let(:field) { :field_130 }
     let(:error) { "some error" }
-    let(:bulk_upload) { create(:bulk_upload, :lettings) }
+    let(:bulk_upload) { create(:bulk_upload, :lettings, year: 2024) }
```
kosiakkatrina: Would it make sense to set default bulk_upload to current year in the factory?
edit: Ah, I see you did that already, can the year be removed here then?

spec/mailers/bulk_upload_mailer_spec.rb
```diff
@@ -46,7 +46,7 @@
           filename: bulk_upload.filename,
           log_type: "lettings",
           upload_timestamp: bulk_upload.created_at.to_fs(:govuk_date_and_time),
-          success_description: "The lettings 2023/24 data you uploaded has been checked. The 0 logs you uploaded are now complete.",
+          success_description: "The lettings #{bulk_upload.year_combo} data you uploaded has been checked. The 0 logs you uploaded are now complete.",
```
kosiakkatrina, line 49: Do we test anywhere that year_combo returns what we expect?

spec/models/validations/sales/financial_validations_spec.rb
```diff
@@ -311,42 +309,32 @@
...
-        expect(record.errors["income2"]).to include(match I18n.t("validations.financial.income.child_has_income"))
-      end
-
-      it "does not add an error if the saledate is before the 23/24 collection window" do
-        record.saledate = Time.zone.local(2022, 4, 3)
-        record.ecstat2 = 9
-        record.income2 = 40_000
-        financial_validator.validate_child_income(record)
-        expect(record.errors).to be_empty
+      context "and saledate is current" do
```
kosiakkatrina: does this context need to be here?

spec/services/bulk_upload/lettings/year2023/row_parser_spec.rb
kosiakkatrina: I'm assuming we'll remove 2023 specific things shortly after the collection close

spec/support/bulk_upload/lettings_log_to_csv.rb
```diff
@@ -422,6 +476,23 @@ def intermediate_rent_type
...
+      3
+    when LettingsLog::RENT_TYPE[:rent_to_buy]
+      4
+    when LettingsLog::RENT_TYPE[:london_living_rent]
+      5
+    when LettingsLog::RENT_TYPE[:other_intermediate_rent_product]
+      6
+    end
+  end
+
```
kosiakkatrina: Would `LettingsLog::RENT_TYPE_MAPPING[log.rent_type]` work here?

spec/services/bulk_upload/lettings/year2023/row_parser_spec.rb
RachaelBooth: Yeah, I expect so - figured it was worth this level of keeping it in until we do, just in case.

spec/components/bulk_upload_error_row_component_spec.rb
RachaelBooth: Some of these tests do rely on things that change per year - I've split these out now to specify the year if and only if the test is actually year dependant.

spec/mailers/bulk_upload_mailer_spec.rb
RachaelBooth: Good spot - have added some.

spec/models/validations/sales/financial_validations_spec.rb
RachaelBooth: Ah, not any more because I removed the 2022 test.

kosiakkatrina (approved):

PR #2510, opened 2024-07-15
CLDC 2954 handle schemes with only deactivated locations

Dinssa:


Dinssa:
There are some early commits in #2499 that have been included at the beginning of this branch and there also are some later commits not included which this branch can/will be updated with.

app/models/location.rb
```diff
@@ -157,7 +157,7 @@ def status_at(date)
     return :incomplete unless confirmed
     return :deactivated if scheme.owning_organisation.status_at(date) == :deactivated ||
       open_deactivation&.deactivation_date.present? && date >= open_deactivation.deactivation_date
-    return :deactivating_soon if open_deactivation&.deactivation_date.present? && date < open_deactivation.deactivation_date
+    return :deactivating_soon if open_deactivation&.deactivation_date.present? && date < open_deactivation.deactivation_date && (startdate.blank? || date >= startdate)
```
kosiakkatrina: What if there is no startdate and there is a deactivation in the future tho? I think that should still be deactivating soon

We also have some scopes, if we're changing what deactivating_soon means here, we should probably be changing it in the scope as well (line 70)

RachaelBooth: Would it work to move the `:activating_soon` line above this rather than essentially duplicating the check?

app/models/scheme.rb
```diff
@@ -295,6 +295,16 @@ def active?
     status == :active
   end
 
+  def all_locations_inactive?
+    active? && locations.all? { |location| !location.active? }
+  end
```
kosiakkatrina: I'd probably flip this and call it something like `has_active_locations?` cause it reads more clearly I'd also use scopes here cause that's faster in theory 
so something like `locations.active.exists?`

app/models/validations/shared_validations.rb
```diff
@@ -84,6 +84,17 @@ def scheme_during_startdate_validation(record)
...
   end
 
+  def tenancy_during_startdate_validation(record)
+    return if record.startdate.blank?
+
+    if record.scheme.present? && !record.scheme.any_location_active_on_date?(record.startdate)
+      record.errors.add :startdate, I18n.t("validations.setup.startdate.scheme.locations_inactive.startdate", name: record.scheme.service_name)
+    end
+    if record.location.present? && !record.location.active_on_date?(record.startdate)
+      record.errors.add :startdate, I18n.t("validations.setup.startdate.location.deactivated.startdate", postcode: record.location.postcode)
```
RachaelBooth: Is this error not already added by the `location_during_startdate_validation` function?

app/models/validations/shared_validations.rb
```diff
@@ -84,6 +84,17 @@ def scheme_during_startdate_validation(record)
     end
   end
 
+  def tenancy_during_startdate_validation(record)
+    return if record.startdate.blank?
+
+    if record.scheme.present? && !record.scheme.any_location_active_on_date?(record.startdate)
+      record.errors.add :startdate, I18n.t("validations.setup.startdate.scheme.locations_inactive.startdate", name: record.scheme.service_name)
```
RachaelBooth: This looks like it depends on the scheme as well as the startdate, so should also add an error to the scheme_id field. 

I also think that maybe this check belongs in the `scheme_during_startdate_validation` function, since it's logically doing the same check (i.e. "is this scheme ok to use on this date"). Alternatively renaming both of these functions so that it's obvious which bit is done by what (in which case I think they should both be called by `validate_scheme`).

app/models/scheme.rb
Dinssa: Okay good idea

app/models/location.rb
Dinssa: That works.

app/models/validations/shared_validations.rb
Dinssa: Adjusted. The commit (7d4a806) includes the startdate/scheme validation in a separate method and called by `validate_scheme`. Just before that I had included the logic in `scheme_during_startdate_validation` and removed it for clarity. Can change it back if you prefer.

app/models/scheme.rb
```diff
@@ -295,6 +295,16 @@ def active?
     status == :active
   end
 
+  def has_active_locations?
+    active? && locations.active.exists?
```
kosiakkatrina: I'm just having a look at `active` scope, and I think it's doing something different from what I assumed at first. I think we should actually use `active_status` scope here.

If we look at `filter_by_status` scope, we're actually replacing `active` with `active_status`


This might need refactoring, but maybe separate from this PR (unless you're keen to look into where/how `active` scope is being used).

app/models/scheme.rb
```diff
@@ -295,6 +295,16 @@ def active?
...
+  end
+
+  def has_active_locations_on_date?(date)
+    return false unless date
+
+    locations.any? { |location| date && location.active_on_date?(date) }
```
kosiakkatrina: I'd maybe like to use scopes here too. Especially if an org has many locations.
Maybe worth refactoring `active` scope (and all the relevant statuses scopes) to take in a conditional date and have `now` by default (I assume that's possible)


You could then call it here something like this: `locations.active_status(date).any?`

app/models/validations/shared_validations.rb
```diff
@@ -102,6 +103,15 @@ def inactive_status(date, resource)
     { scope: status, date: date&.to_formatted_s(:govuk_date), deactivation_date: closest_reactivation&.deactivation_date&.to_formatted_s(:govuk_date) }
   end
 
+  def tenancy_startdate_with_scheme_locations(record)
+    return unless record.respond_to?(:scheme)
```
kosiakkatrina: What would be a situation where record doesn't respond to scheme?

app/views/schemes/_scheme_list.html.erb
```diff
@@ -26,7 +26,12 @@
           <% row.with_cell(text: scheme.owning_organisation&.name) %>
           <% row.with_cell(text: scheme.id_to_display) %>
           <% row.with_cell(text: scheme.locations.visible&.count) %>
-          <% row.with_cell(text: status_tag_from_resource(scheme)) %>
+          <% row.with_cell do %>
+            <%= status_tag_from_resource(scheme) %>
+            <% unless scheme.has_active_locations? %>
+              <%= content_tag(:div, "No currently active locations", class: "app-!-colour-muted", style: "margin-top: 10px") %>
+            <% end %>
```
kosiakkatrina: `has_active_locations?` also checks if scheme itself is active. I feel like the name for the method is a bit misleading.
Is is a conscious decision to only show this tag on active schemes?

config/locales/en.yml
```diff
@@ -285,8 +285,8 @@ en:
 
         location:
           deactivated:
-            startdate: "The location %{postcode} was deactivated on %{date} and was not available on the day you entered. Select another location or edit the tenancy start date"
-            location_id: "The location %{postcode} was deactivated on %{date} and was not available on the day you entered. Select another location or edit the tenancy start date"
+            startdate: "The location %{postcode} is inactive on this date. Enter another date or choose another location."
+            location_id: "This location is not active on the tenancy start date. Choose another location or edit the tenancy start date."
```
kosiakkatrina, line 290: Would we also want to remove `date` arguments from wherever we're using these error messages?

spec/models/validations/setup_validations_spec.rb
```diff
@@ -511,6 +511,59 @@
...
+        location_deactivation_period = build(:location_deactivation_period, deactivation_date: Time.zone.local(2022, 6, 4), location:)
+        location_deactivation_period.save!(validate: false)
+        location.reload
+      end
+
+      it "produces error when scheme does not have any active locations on the tenancy start date" do
+        record.startdate = Time.zone.local(2022, 7, 5)
+        record.scheme = scheme
+        setup_validator.validate_scheme(record)
+        expect(record.errors["startdate"]).to include(match I18n.t("validations.setup.startdate.scheme.locations_inactive.startdate", name: scheme.service_name))
```
kosiakkatrina, line 529: Would it make sense to also test that error gets set on `scheme_id` ?

app/models/validations/shared_validations.rb
Dinssa: Sales log tests brought up some errors, line 107 was the fix.

app/views/schemes/_scheme_list.html.erb
Dinssa: Yes its a conscious decision to only show the tag on active schemes, otherwise this tag would be shown on any scheme incomplete or deactivated which is not as designed in the ticket.

spec/models/validations/setup_validations_spec.rb
Dinssa: Yes definitely will add that, now that I've added the error message to scheme_id also.

app/views/schemes/_scheme_list.html.erb
Dinssa: I can rename it to `active_and_has_active_locations?`

app/models/scheme.rb
Dinssa: Ok will adjust it

app/models/validations/shared_validations.rb
RachaelBooth: I'm slightly worried by the idea that this is being called at all for sales logs - do you know why it is?

app/views/schemes/_scheme_list.html.erb
kosiakkatrina: That could work

config/locales/en.yml
Dinssa: Although the error message defined by `location.deactivated.location_id` no longer needs postcode or date, the other scopes do e.g. `location.activating_soon.location_id`, so must still pass date. Also for `...startdate`.

app/models/location.rb
```diff
@@ -92,6 +92,14 @@ class Location < ApplicationRecord
     .where.not(id: activating_soon.pluck(:id))
   }
 
+  scope :active_on_date, lambda { |date = Time.zone.now|
```
Dinssa: Had I not wanted to also include locations 'deactivating_soon' in the check (as well as 'active' locations) this additional scope would not have been needed and I could have only modified the active_status scope with the optional date.

app/models/validations/shared_validations.rb
Dinssa: Seems like this function failing those tests is not repeatable and this line 107 is not actually needed, so I've removed it.

app/models/location.rb
kosiakkatrina: This looks good! You can leave it like this or have something like 
`locations.active_on_date(date).exists? || locations.deactivating_soon(date).exist?`

I think I prefer it the way you did it here, cause it feels that a distinct scope for what is actually active is more reusable than `active_status` which I think we're only using for tags/filtering

app/models/location.rb
```diff
@@ -180,6 +188,11 @@ def deactivates_in_a_long_time?
     status_at(6.months.from_now) == :deactivating_soon
   end
 
+  def active_on_date?(date)
+    status = status_at(date)
+    %i[active deactivating_soon].include?(status)
+  end
+
```
kosiakkatrina: Is this being used anywhere anymore?

app/models/scheme.rb
```diff
@@ -295,6 +295,16 @@ def active?
     status == :active
   end
 
+  def has_active_locations?
+    locations.active_status.exists?
```
kosiakkatrina: Should this be `active_on_date` instead?

app/models/location.rb
Dinssa: Nope, will remove

kosiakkatrina (approved):

PR #2511, opened 2024-07-15
CLDC-3558 Display correct orgs data protection confirmation

kosiakkatrina:
This would have only been an issue for support users, cause all the other users can only access their own org anyway

### DIFF
--- a/spec/requests/organisations_controller_spec.rb
+++ b/spec/requests/organisations_controller_spec.rb
@@ -2114,7 +2114,6 @@ RSpec.describe OrganisationsController, type: :request do
           get "/organisations/#{organisation.id}/data-sharing-agreement", headers: headers
           expect(response).to have_http_status(:ok)
           expect(page).to have_content("This agreement is made the 2nd day of March 2001")
-          # binding.pry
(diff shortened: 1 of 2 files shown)


spec/requests/organisations_controller_spec.rb
```diff
@@ -2099,6 +2099,26 @@
...
+        allow(support_user).to receive(:need_two_factor_authentication?).and_return(false)
+        sign_in support_user
+      end
+
+      context "and viewing other org dsa" do
+        it "shows correct org data and dates" do
+          get "/organisations/#{organisation.id}/data-sharing-agreement", headers: headers
+          expect(response).to have_http_status(:ok)
+          expect(page).to have_content("This agreement is made the 2nd day of March 2001")
+          # binding.pry
```
RachaelBooth: Does this need to stay here?

RachaelBooth (approved):

spec/requests/organisations_controller_spec.rb
kosiakkatrina: Oh dear, no it does not!

PR #2527, opened 2024-07-23
CLDC-1718 Update numeric questions to text fields

kosiakkatrina:
Our current implementation for numeric inputs (govuk_number_field) is to use input type="number", there are a few issues with that which are outlined here: https://technology.blog.gov.uk/2020/02/24/why-the-gov-uk-design-system-team-changed-the-input-type-for-numbers/

In the past we didn't allow any characters other that numbers to be entered into the numeric input field, but there's an issue with it not showing any feedback to the user when a different character is being entered.

Now instead of preventing non numeric characters we validate the input format.

### DIFF
--- a/spec/models/validations/shared_validations_spec.rb
+++ b/spec/models/validations/shared_validations_spec.rb
@@ -240,5 +240,11 @@ RSpec.describe Validations::SharedValidations do
       shared_validator.validate_numeric_input(sales_log)
       expect(sales_log.errors[:income1]).to be_empty
     end
+
+    it "does not allow decimal point in a wrong format" do
+      sales_log.income1 = "300.09.78"
+      shared_validator.validate_numeric_input(sales_log)
+      expect(sales_log.errors[:income1]).to include I18n.t("validations.numeric.format", field: "Buyer 1’s gross annual income")
+    end
   end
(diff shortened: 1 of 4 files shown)


spec/models/validations/shared_validations_spec.rb
```diff
@@ -215,4 +215,30 @@
...
+      sales_log.income1 = "300"
+      shared_validator.validate_numeric_input(sales_log)
+      expect(sales_log.errors[:income1]).to be_empty
+    end
+
+    it "allows a decimal point" do
+      sales_log.income1 = "300.78"
+      shared_validator.validate_numeric_input(sales_log)
+      expect(sales_log.errors[:income1]).to be_empty
+    end
```
RachaelBooth, line 242: Does this want a test around having the right characters but not in the right format (so e.g. not "300." or "3.0.0")?

(Might be getting a bit fussy to have too many cases though)

RachaelBooth (approved):

spec/models/validations/shared_validations_spec.rb
kosiakkatrina: Yup, added now!

PR #2528, opened 2024-07-23
CLDC-3357-Previously-applied-filters-not-clearing

Dinssa:


app/views/form/guidance/_finding_scheme.erb
```diff
@@ -1,6 +1,6 @@
 <%= govuk_details(summary_text: "Can’t find your scheme?") do %>
   <p class="govuk-body">Schemes are attached to the organisation that owns the property. Check you have correctly answered question 1 "Which organisation owns this property?"</p>
   <p class="govuk-body">If your organisation’s schemes were migrated from old CORE, they may have new names and codes. Search by postcode to find your scheme.</p>
-  <p class="govuk-body"><%= govuk_link_to("View your organisation’s schemes", schemes_path) %></p>
+  <p class="govuk-body"><%= govuk_link_to("View your organisation’s schemes", schemes_path(reset: true)) %></p>
```
kosiakkatrina: When we're on schemes page and have applied some filters, at the top of the filter component we get `Clear` link. I think it points to something like `/clear-filters?filter_type=schemes`. Would it be possible to reuse that instead of adding new params?

Dinssa: Good idea, using it directly instead of in the controller works just as well.

kosiakkatrina (commented):
Do we want to add some tests for this?

kosiakkatrina (approved):

PR #2529, opened 2024-07-24
CLDC-759 Add accessibility testing

kosiakkatrina:


config/routes.rb
```diff
@@ -72,7 +72,6 @@
     get "details", to: "schemes#details"
     get "check-answers", to: "schemes#check_answers"
     get "edit-name", to: "schemes#edit_name"
-    get "support-services-provider", to: "schemes#support_services_provider"
```
kosiakkatrina, line 75: I don't think we serve this path anymore

spec/features/accessibility_spec.rb
```diff
@@ -0,0 +1,180 @@
...
+
+    it "is has accessible pages" do
+      scheme_paths.each do |path|
+        visit(path)
+        expect(page).to have_current_path(path)
+        expect(page).to be_axe_clean.according_to :wcag2aa
+      end
+    end
+  end
+end
```
RachaelBooth, line 194: Do we need to check the homepage / guidance page?

spec/features/accessibility_spec.rb
```diff
@@ -0,0 +1,180 @@
...
+        visit(path)
+        expect(page).to have_current_path(path)
+        expect(page).to be_axe_clean.according_to :wcag2aa
+      end
+    end
+  end
+
+  context "when viewing lettings log pages" do
+    let(:bulk_upload) { create(:bulk_upload, user:) }
+    let(:lettings_log) { create(:lettings_log, :completed, assigned_to: other_user, bulk_upload_id: bulk_upload.id) }
```
RachaelBooth, line 79: Presumably this will only end up checking pages relevant to this log - are there any types of question that would be missed because of that? (I think it's fine as long as we try each question type at least once)

kosiakkatrina: Yeah, that was my thinking too. I think we're hitting all of the types, I was mostly worried about select, but we've got org and user questions

spec/features/accessibility_spec.rb
kosiakkatrina: Yup! Added some

RachaelBooth (approved):

PR #2534, opened 2024-07-25
CLDC-3566: Clear duplicate set ids when a log ceases to be a duplicate through normal update flow

RachaelBooth:


app/controllers/form_controller.rb
```diff
@@ -315,35 +334,6 @@ def check_collection_period
...
-      remove_fixed_duplicate_set_ids(@log)
-    end
-
-    if original_log.present? && current_user.send(class_name.pluralize).duplicate_logs(original_log).any?
-      if dynamic_duplicates.none?
-        flash[:notice] = deduplication_success_banner
-      end
-      send("#{class_name}_duplicate_logs_path", original_log, original_log_id: original_log.id, referrer: params[:referrer], organisation_id: params[:organisation_id])
-    else
-      remove_fixed_duplicate_set_ids(original_log)
```
RachaelBooth, line 341: This line is not replicated in the functions above.

I think it has become unnecessary since either it still has duplicates or it was a duplicate of the log that has been edited and no longer is, and so removing the duplicate set ids from former duplicates of that log will have cleared it - but I feel like I might have missed a case, so please yell if you can spot one!

kosiakkatrina (approved):
I _think_ it covers all the cases!

PR #2535, opened 2024-07-26
CLDC-3564 Update filter search

kosiakkatrina:
When we load any page with filters we pre-load all of the filter options.
In the case of users (assigned_to filter) it might be a very large number, especially for support users or if organisation has a lot of users. 
 
This updates our search function to send the search request which we should deal with server side. We also limit results to 20 most relevant ones.
 
accessible_autocompete_controller is currently used not only for filtering but also, in questions. For example the scheme question in the lettings form. This creates a separate search controller for filtering.

 ----

From what I understand, the nature of accessible_autocomplete component requires us to preload all of the data so that the options can be selected when JS is disabled.
For the situations where JS is disabled, we now instead show a free text field and treat it as a search field in the query. 
image

This should reduce the request times on lettings log page for support users.

I've left filters for CSV downloads the way they were, because the way it works atm is that a specific user/org need to be filtered by. This means we'd have to still pre-load that data, but csv download pages are likely visited more rarely, especially the filters.


app/helpers/filters_helper.rb
```diff
@@ -84,16 +84,54 @@ def selected_option(filter, filter_type)
...
+
+      return [OpenStruct.new(id: selected_user.id, name: selected_user.name, hint: selected_user.email)] if selected_user.present?
+    end
+    [OpenStruct.new(id: "", name: "Select an option", hint: "")]
+  end
+
+  def filter_search_url(category)
+    case category
+    when :user
+      ENV["RAILS_RELATIVE_URL_ROOT"].present? ? "#{ENV['RAILS_RELATIVE_URL_ROOT']}/users/search" : "/users/search"
```
RachaelBooth: I take it we can't use url helpers here?

RachaelBooth (approved):

app/helpers/filters_helper.rb
kosiakkatrina: Ooh, I think we can!

PR #2538, opened 2024-07-26
CLDC-3546: Add error to managing org when it does not have relevant rent period

RachaelBooth:


app/services/bulk_upload/lettings/year2024/row_parser.rb
```diff
@@ -467,6 +467,7 @@ def valid?
 
       fields.each do |field|
         next if errors.include?(field)
+        next if error.type == :skip_bu_error
```
RachaelBooth, line 470: N.B. Copied this from sales since it makes sense not to turn this into a setup error for bulk upload.

config/locales/en.yml
```diff
@@ -436,7 +436,9 @@ en:
         under_10: "Enter a total charge that is at least £10.00 per week"
         less_than_shortfall: "The total charge must be more than the outstanding amount"
       rent_period:
-        invalid_for_org: "%{org_name} does not use %{rent_period} as a rent period. Choose another rent period, or a data coordinator can add rent periods to your organisation"
+        invalid_for_org:
+          period: "%{org_name} does not use %{rent_period} as a rent period. Choose another rent period, or a data coordinator can add rent periods to your organisation"
+          managing_org: "%{org_name} does not use %{rent_period} as a rent period. Set another rent period on this log, or a data coordinator can add rent periods to this organisation"
```
RachaelBooth, line 441: Felt like it needed slightly different wording to the one on the period question but there may be better options than this.

kosiakkatrina (approved):

PR #2539, opened 2024-07-30
CLDC-3241 Explanation of variables on CSVs

Dinssa:


app/services/csv/lettings_log_csv_service.rb
```diff
@@ -258,6 +268,27 @@ def lettings_log_attributes
       @user.support? ? final_attributes : final_attributes - SUPPORT_ONLY_ATTRIBUTES - soft_validations_attributes
     end
 
+    def lettings_log_definitions
+      definitions = if @user.support?
+                      CsvVariableDefinition.where("log_type = 'lettings'")
+                    else
+                      CsvVariableDefinition.where("user_type = 'user' AND log_type = 'lettings'")
+                    end
```
kosiakkatrina: could this be:
`definitions = @user.csv_variable_definitions.group_by(&:variable)` and `csv_variable_definitions` is a method on the user model?

I _think_ using hash conditions in active record calls where possible is a bit more railsy as well (for example `CsvVariableDefinition.where(log_type: "lettings")`

app/services/csv/lettings_log_csv_service.rb
```diff
@@ -5,10 +5,20 @@ def initialize(user:, export_type:, year:)
...
       @year = year
       @attributes = lettings_log_attributes
+      @definitions = lettings_log_definitions
     end
 
     def prepare_csv(logs)
       CSV.generate(headers: true) do |csv|
+        if @year >= 2023
+          csv << @attributes.map do |attribute|
+            definition = @definitions.find { |record| record.variable == attribute.downcase }
```
kosiakkatrina: would it make sense for @definitions to be something like `{created_by: "definition", id: "definition"}` ?
Then you could access them as `@definitions[attribute.downcase]` here

Dinssa: I still want access to the record itself to do the next step and update!, otherwise yes that would work too.

app/services/csv/lettings_log_csv_service.rb
Dinssa: I've amended to return the definitions from the user model and added scopes to the CsvVariableDefinition model to grab lettings or sales records.

app/services/imports/variable_definitions_service.rb
```diff
@@ -0,0 +1,58 @@
...
+      user_type = parsed_file_name[1]
+      year = "20#{parsed_file_name[3]}".to_i
+
+      records_added = 0
+
+      CSV.foreach(file) do |row|
+        next if row.empty?
+
+        variable = row[0].downcase
+        definition = row[1..].join(",")
```
kosiakkatrina, line 33: Are there situations where definition is across multiple fields?

app/services/imports/variable_definitions_service.rb
```diff
@@ -0,0 +1,58 @@
...
+          CsvVariableDefinition.create!(
+            variable: variable.strip,
+            definition: definition.strip,
+            log_type:,
+            user_type:,
+            year:,
+          )
+          records_added += 1
+        end
+      end
```
kosiakkatrina: Is this possible/feasible to achieve using `upsert`?

config/initializers/rails_admin.rb
```diff
@@ -25,8 +25,29 @@
...
+      end
+      field :year do
+        help "Required. Specify the year this definition should be available from. This definition will be used in future years unless superseded by a newer definition."
+      end
+    end
+  end
+
+  config.model "LogValidation" do
+    label "Log Validation"
+    hide
```
kosiakkatrina: What does this do?

app/services/csv/sales_log_csv_service.rb
```diff
@@ -161,6 +169,21 @@ def sales_log_attributes
...
     end
 
+    def sales_log_definitions
+      definitions = @user.variable_definitions.sales
+
+      definitions.group_by { |record| [record.variable, record.definition] }
+                 .map do |_, options|
+        exact_match = options.find { |definition| definition.year == @year && definition.user_type == @user.user_type }
+        next exact_match if exact_match
+
```
kosiakkatrina: Did we end up checking if the definitions are/should be different across different user types?

app/services/imports/variable_definitions_service.rb
Dinssa: There are situations where the definition is not wrapped in quotes " " and includes commas. The variable is taken as all the content up to the first comma, the definition is built from the rest (if there are multiple parts).

config/initializers/rails_admin.rb
Dinssa: This was the pre-existing example model on in rails admin, I just hid it from view. Can be removed instead until implemented.

spec/services/csv/lettings_log_csv_service_spec.rb
```diff
@@ -1,6 +1,16 @@
...
+require "rake"
 
 RSpec.describe Csv::LettingsLogCsvService do
+  subject(:task) { Rake::Task["data_import:add_variable_definitions"] }
+
+  before do
+    Rake.application.rake_require("tasks/log_variable_definitions")
+    Rake::Task.define_task(:environment)
+    task.reenable
+    task.invoke("spec/fixtures/variable_definitions")
```
kosiakkatrina, line 11: This bit requires us to have a lot of definition fixture files. Is there a way to test the functionality without adding all those files? Maybe stubbing some CsvVariableDefinition records instead of running the rake task?

I feel like doing it this way is probably comprehensive enough, but upon reflection it feels a bit large.

We could stub some definitions that have different user types, to test that the correct user's definition is selected, maybe some situation where there's only a definition for a different year, so that we still test all of the paths

lib/tasks/log_variable_definitions.rake
```diff
@@ -0,0 +1,10 @@
+namespace :data_import do
+  desc "Add CsvVariableDefinition records for each file in the specified directory"
+  task :add_variable_definitions, [:path] => :environment do |_task, args|
+    CsvVariableDefinition.delete_all
+    path = Rails.root.join(args[:path])
+    service = Imports::VariableDefinitionsService.new(path:)
```
kosiakkatrina, line 5: Could we add some tests for this task and the service?

app/services/imports/variable_definitions_service.rb
Dinssa: Will look into that.

app/services/csv/sales_log_csv_service.rb
Dinssa: We will check this. In any case though, should there be two different definitions one for 'user' and one for 'support' user_type the best match will be taken, if it exists, if not it should default to the next best match which will be the 'user' definition if there is no 'support' version. Should there be two and later a decision is made to just have one, all that needs to be done is to delete the 'support' record and support users will get the 'user' definition.

spec/services/csv/lettings_log_csv_service_spec.rb
Dinssa: - I could merge the definitions into one file. 
- The tests that check the rows of the CSVs against those in the spec/fixtures/files directory need all definitions, we have all the attributes so need all the definitions. Unless, we remove definitions from these files and leave them blank for ones we know will not be in the test db. It also does test against definitions of different user types because of those files, some being 'support' some being 'user'.

app/services/csv/sales_log_csv_service.rb
kosiakkatrina: Yeah, that makes sense, I'm a bit concerned tho that we're over-engineering to accommodate use case that might not happen. And if we're optimising for the best definition for a specific user, how would we extend this if we decide to have different definitions for providers vs coordinators?

Dinssa: For sure, without these duplicate definitions the code would be simpler and cleaner. As it is though the source files include for the same attribute different definitions based on user type. Should that change and be confirmed not needed, we can adjust this.

lib/tasks/log_variable_definitions.rake
Dinssa: Added

PR #2541, opened 2024-07-31
CLDC-3569 Update social homebuy errors

kosiakkatrina:
Split questions across individual pages
Update error message for social homebuy financial validation
Add guidance on financial calculations
image


app/models/form/sales/pages/deposit.rb
```diff
@@ -16,11 +16,13 @@ def depends_on
...
        { "ownershipsch" => 2 },
-       { "ownershipsch" => 3, "mortgageused" => 1 }]
+       { "ownershipsch" => 3, "mortgageused" => 1 },
+       { "social_homebuy?" => true, "stairowned_100?" => @optional }]
     else
       [{ "social_homebuy?" => false, "ownershipsch" => 1 },
        { "ownershipsch" => 2 },
-       { "ownershipsch" => 3, "mortgageused" => 1 }]
+       { "ownershipsch" => 3, "mortgageused" => 1 },
+       { "social_homebuy?" => true }]
```
RachaelBooth: Is there a way to simplify these conditions at all given the addition?

kosiakkatrina: Yes, I _think_ it still does the same thing!

app/models/form/sales/pages/deposit.rb
```diff
@@ -0,0 +1,21 @@
...
+    @questions ||= [
+      Form::Sales::Questions::DepositAmount.new(nil, nil, self, ownershipsch: @ownershipsch, optional: @optional),
+    ]
+  end
+
+  def routed_to?(log, _user)
+    return true if log.ownershipsch == 2 || (log.ownershipsch == 3 && log.mortgageused == 1)
+    return false if log.stairowned_100? != @optional && form.start_year_after_2024?
+
+    log.social_homebuy? == false && log.ownershipsch == 1 || log.social_homebuy? == true
```
RachaelBooth: Is social_homebuy relevant if the log is not shared ownership? Looks like the options are based on type, and I think the type options are fairly different based on ownership scheme.

RachaelBooth (approved):

app/models/form/sales/pages/deposit.rb
kosiakkatrina: Yeah, you're right 👀 removed that and it still passes the tests

PR #2543, opened 2024-08-02
Update xml exports

kosiakkatrina:
We want to start exporting user data to CDS
This adds a new user export service and refactors creation of the manifest outside of the logs export so that it includes both, the logs archive names and the user archives.
I've separated out the user export job, there's a lot of duplication, but also a few things that got tweaked, so I decided to make them independent, but let me know if it makes sense to make it more dry


app/services/exports/user_export_service.rb
```diff
@@ -0,0 +1,156 @@
...
+
+      Export.new(collection:, started_at: @start_time, base_number:, increment_number:)
+    end
+
+    def get_archive_name(collection, base_number, increment)
+      return unless collection
+
+      base_number_str = "f#{base_number.to_s.rjust(4, '0')}"
+      increment_str = "inc#{increment.to_s.rjust(4, '0')}"
+      "core_#{collection}_#{current_collection_start_year}_#{current_collection_start_year + 1}_apr_mar_#{base_number_str}_#{increment_str}".downcase
```
RachaelBooth: Since users don't vary by year, can we avoid the dates in this?

spec/services/exports/user_export_service_spec.rb
```diff
@@ -0,0 +1,219 @@
...
+        expect(Export.last.started_at).to be_within(2.seconds).of(start_time)
+      end
+
+      context "when this is the first export (full)" do
+        it "returns a ZIP archive for the master manifest (existing lettings logs)" do
+          expect(export_service.export_xml_users).to eq({ expected_zip_filename.gsub(".zip", "").gsub(".zip", "") => start_time })
+        end
+      end
+
+      context "and underlying data changes between getting the logs and writting the manifest" do
```
RachaelBooth: Should change this name to reference users rather than logs

app/services/exports/export_service.rb
```diff
@@ -0,0 +1,51 @@
+module Exports
+  class ExportService
+    include CollectionTimeHelper
+
+    def initialize(storage_service, logger = Rails.logger)
+      @storage_service = storage_service
+      @logger = logger
+    end
+
+    def export_xml(full_update: false, collection_year: nil)
```
RachaelBooth: Would it be worth adjusting this to allow us to trigger e.g. just a users export? (Maybe pass in a "collection" rather than "collection_year" and only run the relevant bits if provided?)

RachaelBooth:
> I've separated out the user export job, there's a lot of duplication, but also a few things that got tweaked, so I decided to make them independent, but let me know if it makes sense to make it more dry

I'm not sure on this tbh - it would be nice to have a little less duplication, on the other hand it feels like it might make the whole thing a bit more complicated to follow. Would be tempted to spend a little bit of time seeing how nicely it worked out.

app/services/exports/export_service.rb
kosiakkatrina: Yes, I think so!

kosiakkatrina:
> > I've separated out the user export job, there's a lot of duplication, but also a few things that got tweaked, so I decided to make them independent, but let me know if it makes sense to make it more dry
> 
> I'm not sure on this tbh - it would be nice to have a little less duplication, on the other hand it feels like it might make the whole thing a bit more complicated to follow. Would be tempted to spend a little bit of time seeing how nicely it worked out.

I extracted some shared logic into a parent class, let me know if that looks better

app/services/exports/xml_export_service.rb
```diff
@@ -0,0 +1,103 @@
...
+    def build_manifest_xml(record_number)
+      doc = Nokogiri::XML("<report/>")
+      doc.at("report") << doc.create_element("form-data-summary")
+      doc.at("form-data-summary") << doc.create_element("records")
+      doc.at("records") << doc.create_element("count-of-records", record_number)
+
+      xml_doc_to_temp_file(doc)
+    end
+
+    def collection_years_to_export(collection_year)
```
RachaelBooth: Does this belong in the lettings log export service?

lib/tasks/data_export.rake
```diff
@@ -8,10 +8,10 @@ namespace :core do
 
   desc "Export all data XMLs for import into Central Data System (CDS)"
   task :full_data_export_xml, %i[year] => :environment do |_task, args|
-    collection_year = args[:year].present? ? args[:year].to_i : nil
+    collection = args[:year].present? ? args[:year].to_i : nil
```
RachaelBooth: If this might be "users" now, presumably we should not cast it to a number? (Also maybe rename year?)

RachaelBooth (commented):
I like how the structure works here - just a couple of small comments.

app/services/exports/xml_export_service.rb
kosiakkatrina: I did a rebase, but this is the only change I made [Move method and update task argument](https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/2543/commits/969e65ad82760decaecfacaa4a2f8769344db168)

lib/tasks/data_export.rake
kosiakkatrina: Good shout! should also be updated here [Move method and update task argument](https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/2543/commits/969e65ad82760decaecfacaa4a2f8769344db168)

app/services/exports/user_export_service.rb
kosiakkatrina: Ended up adding a static period, cause it's required by CDS

app/services/exports/lettings_log_export_service.rb
```diff
@@ -1,69 +1,27 @@
...
-      @storage_service = storage_service
-      @logger = logger
-    end
-
     def export_xml_lettings_logs(full_update: false, collection_year: nil)
-      start_time = Time.zone.now
-      daily_run_number = get_daily_run_number
       archives_for_manifest = {}
-      recent_export = LogsExport.order("started_at").last
+      recent_export = Export.order("started_at").last
```
RachaelBooth: This might also need to depend on collection now?

RachaelBooth: (Otherwise LGTM!)

kosiakkatrina: Yes, I was a bit reluctant to add it into the loop, cause I was worried it would always grab the newly created one, but if we're filtering by collection that should be fine

RachaelBooth (approved):

PR #2545, opened 2024-08-02
CLDC-2588-Separate soft and hard validation errors in bulk upload

Dinssa:


app/components/bulk_upload_error_row_component.html.erb
```diff
@@ -18,16 +20,50 @@
...
+                <% row.with_cell(text: question_for_field(error.field)) %>
+                <% row.with_cell(header: true, text: error.error) %>
+                <% row.with_cell(text: error.field.humanize) %>
             <% end %>
           <% end %>
         <% end %>
       <% end %>
     <% end %>
+
+    <% potential_errors = bulk_upload_errors.select { |error| error.category == "soft_validation" } %>
```
kosiakkatrina: You could potentially do something like `potential_errors, errors = bulk_upload_errors.partition { |error| error.category == "soft_validation" }` to avoid iterating over bulk_upload_errors twice

app/components/bulk_upload_error_row_component.html.erb
```diff
@@ -18,16 +20,50 @@
...
-          <% bulk_upload_errors.each do |error| %>
+          <% bulk_upload_errors.reject { |error| error.category == "soft_validation" }.each do |error| %>
             <% body.with_row do |row| %>
-              <% row.with_cell(header: true, text: error.cell) %>
-              <% row.with_cell(text: question_for_field(error.field)) %>
-              <% row.with_cell(text: error.error) %>
-              <% row.with_cell(text: error.field.humanize) %>
+                <% row.with_cell(text: error.cell) %>
+                <% row.with_cell(text: question_for_field(error.field)) %>
+                <% row.with_cell(header: true, text: error.error) %>
```
kosiakkatrina: I don't think this semantically makes sense as a header, we probably want to keep the header as is and instead apply different styling to this row (probably strong tag)

app/components/bulk_upload_error_row_component.html.erb
Dinssa: That's a good idea, will adjust.

app/components/bulk_upload_error_row_component.html.erb
```diff
@@ -18,16 +21,49 @@
...
+            <% end %>
+          <% end %>
+
+          <%= table.with_body do |body| %>
+            <% potential_errors.group_by(&:error).each do |error_message, errors| %>
+              <% errors.each_with_index do |error, index| %>
+                <% row_class = "grouped-rows" %>
+                <% row_class += " first-row" if index.zero? %>
+                <% row_class += " last-row" if index == errors.size - 1 %>
+                <% body.with_row(html_attributes: { class: row_class }) do |row| %>
```
kosiakkatrina, line 55: I'm not entirely sure why we need this, as in, I can understand why we want to remove the border underneath the last row, but I'm not sure why we're removing it in other places. Looking at this it also doesn't seem like it's doing what I'd expect https://review.submit-social-housing-data.levellingup.gov.uk/2545/lettings-logs/bulk-upload-results/1

app/components/bulk_upload_error_row_component.rb
```diff
@@ -18,7 +18,7 @@ def tenant_code
   def tenant_code_html
     return if tenant_code.blank?
 
-    content_tag :span, class: "govuk-!-margin-left-3" do
+    content_tag :span, class: "govuk-!-margin-left-3", style: "font-weight: 400;" do
```
kosiakkatrina: Isn't that the default font weight?

app/components/bulk_upload_error_row_component.html.erb
Dinssa: It's not to remove the border on the last row (that can be done but I did not to keep things consistent). 

For a given error message there might be multiple rows responsible, each of those are grouped together and the error message is only displayed once with it spread across the rows using `rowspan`. These lines of code are to outline the first row and last row of the group, remove the borders of each of the interior rows and only keep the top and bottom border of the whole group. Now though since that the error message is no longer a header `<th>` and now `<td>` I will need to adjust the styling.

app/components/bulk_upload_error_row_component.rb
Dinssa: This tag is wrapped around in a `h3` so although the text before it is wrapped in `<strong>` it is redundant as the whole line was having the font-weight = 700 applied as a result of the `h3`.

`<h3 class="govuk-summary-card__title"><strong>Row <%= row %></strong> <%= tenant_code_html %> <%= property_ref_html %></h3>`

I should probably remove the `<strong>` tags or change it from a `<h3>`

Dinssa: I'll add `govuk-!-font-weight-regular` class to the `h3`, and remove the inline styling from all 3 tags I added it to here.

app/components/bulk_upload_error_row_component.html.erb
kosiakkatrina: Ahh, okay! I see now that you're grouping them! Any thoughts on the styling in the url I linked? Not sure if you can access it, so here's a screenshot:
image

Dinssa: That is the error message now being `<td>` causing this, fixing it now.

kosiakkatrina (approved):

PR #2559, opened 2024-08-08
CLDC-3597 Update equity validation for resale

kosiakkatrina:
If it is a resale it should actually be possible for equity to be larger than 75% for shared ownerships

### DIFF
--- a/app/models/validations/sales/financial_validations.rb
+++ b/app/models/validations/sales/financial_validations.rb
@@ -96,9 +96,10 @@ module Validations::Sales::FinancialValidations
     if record.equity < range.min
       record.errors.add :type, I18n.t("validations.financial.equity.under_min", min_equity: range.min)
       record.errors.add :equity, :under_min, message: I18n.t("validations.financial.equity.under_min", min_equity: range.min)
-    elsif record.equity > range.max
+    elsif !record.is_resale? && record.equity > range.max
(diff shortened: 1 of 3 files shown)


app/models/validations/sales/financial_validations.rb
```diff
@@ -96,9 +96,10 @@ def validate_equity_in_range_for_year_and_type(record)
     if record.equity < range.min
       record.errors.add :type, I18n.t("validations.financial.equity.under_min", min_equity: range.min)
       record.errors.add :equity, :under_min, message: I18n.t("validations.financial.equity.under_min", min_equity: range.min)
-    elsif record.equity > range.max
+    elsif !record.is_resale? && record.equity > range.max
       record.errors.add :type, I18n.t("validations.financial.equity.over_max", max_equity: range.max)
       record.errors.add :equity, :over_max, message: I18n.t("validations.financial.equity.over_max", max_equity: range.max)
+      record.errors.add :resale, I18n.t("validations.financial.equity.over_max", max_equity: range.max)
```
Dinssa, line 102: Why add the error to resale if this elsif block is for the scenario when the record is not a resale?

kosiakkatrina: I think the validation might trigger if resale changes from "yes" to "no", so if would make sense to display it on the resale page

Dinssa (approved):

PR #2560, opened 2024-08-08
CLDC-3595 Remove duplicate errors for check errors page

kosiakkatrina:
Example where all the errors are the same and get deduplicated:
image

Example of different errors that would all get shown:
image

### DIFF
--- a/app/helpers/form_page_error_helper.rb
+++ b/app/helpers/form_page_error_helper.rb
@@ -5,7 +5,6 @@ module FormPageErrorHelper
   end
 
   def remove_duplicate_page_errors(lettings_log)
-    lettings_log.errors.map(&:message)
     lettings_log.errors.group_by(&:message).each do |_, errors|
       next if errors.size == 1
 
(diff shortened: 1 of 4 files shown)


app/helpers/form_page_error_helper.rb
```diff
@@ -4,6 +4,16 @@ def remove_other_page_errors(lettings_log, page)
     other_page_error_ids.each { |id| lettings_log.errors.delete(id) }
   end
 
+  def remove_duplicate_page_errors(lettings_log)
+    lettings_log.errors.map(&:message)
```
Dinssa: Is the map on line 8 needed? Doesn't seem to contribute to the functionality of the method. Group_by handles removing the duplicates. Tested both examples given in your description and it works the same without line 8.

kosiakkatrina: Oh, yeah, it doesn't look like it's doing anything, must have accidentally left it! Thanks, it should be removed now!

Dinssa (approved):

PR #2562, opened 2024-08-09
CLDC-3569 Update guidance for shared ownership without mortgage

kosiakkatrina:
- Mortgage amount question is not routed to when mortgage used is "no", in this case in the guidance we would link to mortgage used question instead of mortgage amount.
- We want to format BU errors when there's html in the error message
- This also adds a header to the mortgage used question, to align it with other questions that have top guidance (because the guidance content is placed below the header)

### DIFF
--- a/app/views/form/guidance/_financial_calculations_shared_ownership.html.erb
+++ b/app/views/form/guidance/_financial_calculations_shared_ownership.html.erb
@@ -1,12 +1,20 @@
 <%= govuk_details(summary_text: "How the financial values are calculated") do %>
   <p class="govuk-body">
-    <% if log.type == 18 %>
-      The mortgage amount <%= question_link("mortgage", log, current_user) %>,
-      cash deposit <%= question_link("deposit", log, current_user) %>,
-      and cash discount <%= question_link("cashdis", log, current_user) %>
-    <% else %>
-      The mortgage amount <%= question_link("mortgage", log, current_user) %>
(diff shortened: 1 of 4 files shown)


app/views/form/guidance/_financial_calculations_shared_ownership.html.erb
```diff
@@ -1,12 +1,20 @@
...
+      <% end %>
+    <% elsif log.mortgage_not_used? || log.mortgage_use_unknown? %>
+        <% if log.type == 18 %>
+          The cash deposit <%= question_link("deposit", log, current_user) %>,
+          and cash discount <%= question_link("cashdis", log, current_user) %>
+        <% else %>
+          Cash deposit <%= question_link("deposit", log, current_user) %>
+        <% end %>
     <% end %>
     added together must equal
```
Dinssa: If we go down the second branch (`elsif log.mortgage_not_used? || log.mortgage_use_unknown?`) then if log type is not 18 and we go down the second branch again (`Cash deposit <%= question_link("deposit", log, current_user) %>`) ... the full sentence looks like it'll be: **Cash deposit added together must equal.....**, this looks like the one case where the phrase "added together" won't work.

kosiakkatrina: Does it look better now?

Dinssa (approved):

PR #2565, opened 2024-08-12
CLDC 2094: View a merge request

Dinssa:
This branch builds the page to view details of a specific merge request. When first created they have a status of incomplete and the user is expected to answer questions to identify the help desk ticket, absorbing organisation, merging organisations and merge date. All other details are calculated.

**Viewing an in progress (incomplete) request should look something like this:**
Screenshot 2024-08-12 at 13 50 29
Note:

- You are unable to click to begin a request until details are complete and "ready to merge"
- You can delete the request at any time, except when the merge is completed.
- Should the request not have an approved/signed Date Protection Agreement you will be prompted to contact a Data Protection Officer (DPO) from the absorbing organisation. Should there be multiple, the first found is shown.
- You will be able to change the details on this page, not implemented in this branch.

**If the absorbing organisation does not yet have a DPO assigned you'll be prompted to do so:**
Screenshot 2024-08-12 at 13 44 34

**An additional merge outcomes section becomes visible once all questions of the request are complete:**
Screenshot 2024-08-12 at 13 52 01
Note:

- In the absence of values in all questions the placeholder text is shown.
- However, the values for this section should be calculated once the previous sections are complete. While the request is "Ready to merge"/"Merge Issues"/"Processing", you can preview the values in this section.

**A completed merge request should look something like this:**
Screenshot 2024-08-12 at 13 50 40
Note:

- You do not have the ability to change any options when complete.
- The merge outcomes section and will now give you a snapshot of the changes on the merge date.


app/models/merge_request.rb
```diff
@@ -31,4 +31,12 @@ def organisation_name_uniqueness
   def absorbing_organisation_name
     absorbing_organisation&.name || ""
   end
+
+  def requesting_user
+    User.find(requester_id) if requester_id
+  end
+
+  def dpo_user
+    absorbing_organisation.users.find_by(is_dpo: true) if absorbing_organisation
```
Dinssa: `requesting_user `assumes in later implementation we'll set this value to equal to user id of the support user who creates the request

`dpo_user` finds the first valid DPO user

app/views/merge_requests/_details_list.html.erb
```diff
@@ -0,0 +1,27 @@
...
+      <% if detail[:action].present? %>
+        <% row.with_action(
+             text: detail[:action][:text],
+             href: detail[:action][:href],
+             visually_hidden_text: detail[:action][:visually_hidden_text],
+             ) %>
+      <% end %>
+    <% end %>
+  <% end %>
+<% end %>
```
Dinssa: Similar implementation to the form check answers list. With some changes to handle html & date values.

app/views/merge_requests/details.html.erb
```diff
@@ -0,0 +1,64 @@
...
+      <strong>The absorbing organisation must accept the Data Sharing Agreement before merging.</strong>
+      <br>
+      <% if @merge_request.dpo_user %>
+        Contact the Data Protection Officer: <%= link_to @merge_request.dpo_user.name, user_path(@merge_request.dpo_user.id) %>
+      <% else %>
+        <%= @merge_request.absorbing_organisation_name %> does not have a Data Protection Officer. You can assign one on the <%= link_to 'users page', "#{organisation_path(@merge_request.absorbing_organisation_id)}/users" %>.
+      <% end %>
+    </div>
+  </div>
+<% end %>
```
Dinssa: I added the boolean `signed_dsa` to the merge request model and use this to check if the DSA has been signed. This notification is also not shown in the case that the absorbing organisation has not yet been set. Determining who the DPO user is relies on there being an absorbing organisation first.

app/views/merge_requests/details.html.erb
```diff
@@ -0,0 +1,64 @@
...
+    Delete merge request
+  </button>
+</div>
+<% end %>
+
+<% request_details = [
+  { label: "Requester", value: display_value_or_placeholder(@merge_request.requesting_user&.name) },
+  { label: "Helpdesk ticket", value: @merge_request.helpdesk_ticket.present? ? link_to("#{@merge_request.helpdesk_ticket} (opens in a new tab)", "https://dluhcdigital.atlassian.net/browse/#{@merge_request.helpdesk_ticket}", target: "_blank", rel: "noopener noreferrer") : display_value_or_placeholder(nil), action: @merge_request.status == "request_merged" ? nil : { text: "Change", href: "#", visually_hidden_text: "helpdesk ticket" } },
+  { label: "Status", value: status_tag(@merge_request.status) },
+] %>
```
Dinssa: Each question/detail table is built according to this spec which determines the name of the row (label); the value and the accessibility hidden text for the action buttons/links.

app/controllers/merge_requests_controller.rb
```diff
@@ -32,6 +32,11 @@ def update
     end
   end
 
+  def details
+    @merge_request = MergeRequest.find(params[:id])
+    render :details
+  end
+
```
kosiakkatrina: Would it make sense for this to be `show` instead of `details` and the url to be `/merge-requests/{id}` ?

app/models/merge_request.rb
```diff
@@ -31,4 +31,12 @@ def organisation_name_uniqueness
   def absorbing_organisation_name
     absorbing_organisation&.name || ""
   end
+
+  def requesting_user
+    User.find(requester_id) if requester_id
```
kosiakkatrina: Would it make sense to have requester as an association? 

Similar to absorbing organisation (`belongs_to :absorbing_organisation, class_name: "Organisation", optional: true` at the top of this file)

That way we should be able to access it as @merge_request.requester without having to create a method for it ourselves

app/models/merge_request.rb
```diff
@@ -31,4 +31,12 @@ def organisation_name_uniqueness
...
+  end
+
+  def dpo_user
+    absorbing_organisation.users.filter_by_active.find_by(is_dpo: true) if absorbing_organisation
```
kosiakkatrina: `absorbing_organisation.data_protection_officers.filter_by_active.first` ? Also, what would happen if organisation doesn't have any DPOs?

app/views/merge_requests/details.html.erb
kosiakkatrina: Would it make sense to move this to a helper or the model? Same with other sections

spec/factories/merge_requests.rb
```diff
@@ -1,7 +1,8 @@
 FactoryBot.define do
   factory :merge_request do
-    status { "incomplete" }
+    status { "Incomplete" }
```
kosiakkatrina: Why is this change needed? I feel like we should save the status as lowercase

app/views/merge_requests/_merge_request_list.html.erb
```diff
@@ -48,7 +48,7 @@
             <% row.with_cell(html_attributes: {
               scope: "row",
             }) do %>
-              <%= govuk_link_to("View details", "/merge-request/#{merge_request.id}/organisations") %>
+              <%= govuk_link_to("View details", "/merge-request/#{merge_request.id}/details") %>
```
kosiakkatrina: Can this be route helper?

app/helpers/merge_requests_helper.rb
```diff
@@ -0,0 +1,5 @@
+module MergeRequestsHelper
+  def display_value_or_placeholder(value, placeholder = "You didn't answer this question")
+    value.presence || content_tag(:span, placeholder, class: "app-!-colour-muted")
+  end
```
kosiakkatrina, line 4: Is there anywhere else we do this already that could  be reused?

app/models/merge_request.rb
Dinssa: Sure, will try that

app/models/merge_request.rb
Dinssa: That works.

If the org doesn't have any then the notification at the top will ask them to assign one. Screenshot in the description.

spec/factories/merge_requests.rb
Dinssa: No reason really, will revert it

app/helpers/merge_requests_helper.rb
Dinssa: There are places with similar methods for questions and/or logs. Each not useful here. This method could be reusable though, doesn't have to be just for merge requests.

**check_answers_helper.rb**
```
def unanswered_value(log:, question:)
    if log.creation_method_bulk_upload? && log.bulk_upload.present? && !log.optional_fields.include?(question.id)
      "<span class=\"app-!-colour-red\">You still need to answer this question</span>".html_safe
    else
      "<span class=\"app-!-colour-muted\">You didn’t answer this question</span>".html_safe
    end
  end
```

**check_answers_summary_list_card_component.rb**
```
  def unanswered_value(question)
    if log.creation_method_bulk_upload? && log.bulk_upload.present? && !log.optional_fields.include?(question.id)
      "<span class=\"app-!-colour-red\">You still need to answer this question</span>".html_safe
    else
      "<span class=\"app-!-colour-muted\">You didn’t answer this question</span>".html_safe
    end
  end
```

app/views/merge_requests/details.html.erb
Dinssa: Moved it to the helper

app/controllers/merge_requests_controller.rb
```diff
@@ -32,9 +32,8 @@ def update
     end
   end
 
-  def details
+  def show
     @merge_request = MergeRequest.find(params[:id])
```
kosiakkatrina: I think `before_action :find_resource` might do this anyway, so we might not need to define this method at all (?)

kosiakkatrina (approved):

PR #2573, opened 2024-08-14
CLDC-3566: Clear duplicate set ids when a log ceases to be a duplicate in normal update flow - with better performance

RachaelBooth:


RachaelBooth:
Probably the most relevant bit here is to look at changes in the second commit (i.e. not the one that reverts the revert). This should vastly reduce the number of duplicate log queries we're making.

I've put a fair amount of data on the review app to help with judging performance, although not as much as staging or live has. It's hard to get an actually reliable measurement (and one we did on a review app would not be valid for prod because of differences in underlying infra), but my unscientific comparison attempts do have this being better. We may want to confirm on staging, which will be much more representative.

app/controllers/form_controller.rb
```diff
@@ -217,7 +223,7 @@ def successful_redirect_path(pages_to_check)
       original_log = current_user.send(class_name.pluralize).find_by(id: from_referrer_query("original_log_id"))
 
       if original_log.present? && current_user.send(class_name.pluralize).duplicate_logs(original_log).any?
-        if @log.duplicates.none?
+        if @log.duplicate_set_id.nil?
```
RachaelBooth, line 226: Assuming we trust the code that removes this if there are no longer any duplicates (which has been called before this), since we only need to know if there are any duplicates here I think we can just check if this attribute is nil - it should be nil if and only if there are no saved duplicates.

kosiakkatrina (approved):

PR #2578, opened 2024-08-14
Merge request fixes

kosiakkatrina:
Update merging organisations page to only save added merging organisations when save and continue is selected. Update back buttons and some copy.

### DIFF
--- a/app/controllers/merge_requests_controller.rb
+++ b/app/controllers/merge_requests_controller.rb
@@ -1,6 +1,6 @@
 class MergeRequestsController < ApplicationController
   before_action :find_resource, exclude: %i[create new]
-  # before_action :authenticate_user!
+  before_action :authenticate_user!
   before_action :authenticate_scope!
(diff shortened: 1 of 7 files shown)


app/controllers/merge_requests_controller.rb
```diff
@@ -22,22 +21,33 @@ def update
...
       render previous_template, status: :unprocessable_entity
     end
   end
 
   def update_merging_organisations
+    @new_merging_org_ids = params["merge_request"]["new_merging_org_ids"].split(" ")
     merge_request_organisation = MergeRequestOrganisation.new(merge_request_organisation_params)
-    if merge_request_organisation.save
+    if merge_request_organisation.valid?
+      @new_merging_org_ids.push(merge_request_organisation_params[:merging_organisation_id])
```
kosiakkatrina, line 40: We don't want to immediately save added orgs in case user navigates away from the page without saving

Dinssa (approved):

PR #2581, opened 2024-08-15
CLDC-3575: Remove reference to dluhc on GitHub

Dinssa:
Changes the image on the readme


kosiakkatrina (approved):
would it make sense to remove the old file?

PR #2589, opened 2024-08-16
CLDC-2100++ Add success notification after delete

Dinssa:
This branch includes a new JS frontend controller to change the behaviour of the govuk tabs and the anchors. 

The goal is to not scroll to the anchor and for visitors to the page to land at the top, with the success notification (primarily but also all other header elements) in view.

### DIFF
--- a/app/frontend/controllers/tabs_controller.js
+++ b/app/frontend/controllers/tabs_controller.js
@@ -1,0 +1,30 @@
+document.addEventListener("DOMContentLoaded", function() {
+  const urlParams = new URLSearchParams(window.location.search);
+  const tab = urlParams.get("tab");
+
+  function activateTab(tabId) {
+    const tabElement = document.getElementById(tabId);
+    if (tabElement) {
+      tabElement.click();
+    }
+    history.replaceState(null, null, `${window.location.pathname}?${urlParams.toString()}`);
+  }
+
+  function handleTabClick(event) {
+    event.preventDefault();
+    const targetId = this.getAttribute('href').substring(1);
+    activateTab(targetId);
+    urlParams.set("tab", targetId);
+    history.replaceState(null, null, `${window.location.pathname}?${urlParams.toString()}`);
+  }
+
+  if (tab) {
+    activateTab(`tab_${tab}`);
+  }
+
+  window.scrollTo(0, 0);
+
+  document.querySelectorAll('.govuk-tabs__tab').forEach(tabElement => {
+    tabElement.addEventListener('click', handleTabClick);
+  });
+});
--- a/app/frontend/controllers/tabs_controller.js
+++ b/app/frontend/controllers/tabs_controller.js
@@ -1,7 +1,12 @@
-document.addEventListener("DOMContentLoaded", function() {
+document.addEventListener('DOMContentLoaded', function() {
   const urlParams = new URLSearchParams(window.location.search);
-  const tab = urlParams.get("tab");
+  let tab = urlParams.get('tab');
 
+  if (!tab && window.location.hash) {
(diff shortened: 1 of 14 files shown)


app/frontend/controllers/tabs_controller.js
```diff
@@ -0,0 +1,30 @@
+document.addEventListener("DOMContentLoaded", function() {
+  const urlParams = new URLSearchParams(window.location.search);
+  const tab = urlParams.get("tab");
+
+  function activateTab(tabId) {
+    const tabElement = document.getElementById(tabId);
+    if (tabElement) {
+      tabElement.click();
+    }
+    history.replaceState(null, null, `${window.location.pathname}?${urlParams.toString()}`);
```
Dinssa: This line:

- Removes the anchor: The fragment identifier (anchor) is not included in the new URL, effectively removing it.
- Keeps the tab URL parameter: The urlParams object, which contains the tab parameter, is converted to a query string and appended to the path, ensuring the tab parameter remains in the URL.

app/helpers/merge_requests_helper.rb
```diff
@@ -59,7 +59,7 @@ def merge_request_back_link(merge_request, page, referrer)
 
     case page
     when "absorbing_organisation"
-      organisations_path(anchor: "merge-requests")
+      organisations_path(tab: "merge-requests")
```
Dinssa, line 62: ~~All places where previously we'd access the right tab on the organisations index with the anchor must now use the tab parameter.~~

The _Merge requests_ tab is `tab: "merge-requests"`, the _All organisations_ tab is `tab: "all-organisations"` (not currently used anywhere).

app/views/layouts/application.html.erb
```diff
@@ -126,7 +126,8 @@
           <%= govuk_notification_banner(
             title_text: "Success",
             success: true, title_heading_level: 3,
-            title_id: "flash-notice"
+            title_id: "flash-notice",
+            role: "alert",
```
Dinssa: Assuming this behaviour is desired for all notifications - the added line role: "alert" makes sure to shift focus to the notification banner on page load. Not quite works this way with the tabs and anchors on this page but should do on other pages in the app without. Saw this attribute in the method documentation so thought to add it in.

app/helpers/merge_requests_helper.rb
Dinssa: I've changed this to now accept an anchor parameter being set within the route helper. 
For example, `organisations_path(anchor: "merge-requests")` will be treated the same `organisations_path(tab: "merge-requests")`. The anchor will still be removed and replaced with a tab parameter in the url. Why change this? In the absence of the parameter tab, if anchor is set instead, then we'll use that value to determine which tab to show.

Two possible use cases:

- You want to add in this data controller with another tab element without making changes in all places the path is used.
- A developer makes a mistake and uses the anchor key word instead of tab. Anchor may be the default choice someone would use for this functionality.

app/frontend/controllers/tabs_controller.js
```diff
@@ -0,0 +1,35 @@
+document.addEventListener('DOMContentLoaded', function () {
+  const urlParams = new URLSearchParams(window.location.search)
+  let tab = urlParams.get('tab')
+
+  if (!tab && window.location.hash) {
+    tab = window.location.hash.substring(1)
+    urlParams.set('tab', tab)
+    window.history.replaceState(null, null, `${window.location.pathname}?${urlParams.toString()}`)
+  }
```
Dinssa, line 9: This code treats an anchor the same way as a tab url param.

kosiakkatrina (approved):

PR #2590, opened 2024-08-16
Yarn package upgrades

RachaelBooth:
This updates various yarn packages to clear some security warnings. 

In most cases it stays within those we specify in package.json - the exception being our linting/formatting packages. This upgrades stylelint-config-gds and stylelint to avoid conflicts, and (as recommended as part of upgrading) adds prettier to format scss (and some other) files.


.prettierignore
```diff
@@ -0,0 +1,19 @@
...
+!*.scss
+!*.md
+!*.yml
+!*.json
+
+# Ensures subdirectories are checked for files matching the above
+!*/
+
+config/locales/
+app/views/content/data_sharing_agreement.md
```
RachaelBooth, line 13: Prettier tried to make some dubious changes to this file, possibly because it's currently actually a bit off. Given it's fairly important that it stays static I think it makes sense to ignore it.

package.json
```diff
@@ -36,9 +36,10 @@
   "version": "0.1.0",
   "devDependencies": {
     "are-you-es5": "^2.1.2",
+    "prettier": "3.3.3",
     "standard": "^17.0.0",
```
RachaelBooth, line 40: Prettier is recommended in the upgrade of stylelint-config-gds to cover stylesheet files, but it can also cover various other file types - including js, where we are currently using standard. We could replace standard with prettier for js, although there are a few things that they disagree on - I've not done so here to reduce the size of the PR, but could raise another if people are keen.

RachaelBooth:
Might be worth looking at commits individually here to separate out the one where I run the new formatter.

kosiakkatrina (approved):

PR #2596, opened 2024-08-20
CLDC-3404 Allow support to update user organisation

kosiakkatrina:
Allow support users to change user organisations
image
image
Validate that relevant relationships are there to move the logs over
image
Choosing to unassign logs assigns them to a user in the current user organisation called "Unassigned", some of the organisations might already have this user as it would get created during migration from the old service in cases where no relevant user was created to own that log. If  "Unassigned" user doesn't exist for organisation, we'd create one
image

This PR also allows editing deactivated users by support user as we want to be able to move deactivated users between organisations. This does also mean that other user fields can be edited, I'm yet to check if that's okay, or whether we should do some work to split that up.

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -153,6 +153,16 @@ class UsersController < ApplicationController
     redirect_to users_organisation_path(@user.organisation), notice: I18n.t("notification.user_deleted", name: @user.name)
   end
 
+  def log_reassignment
+    authorize @user
+
+    if params[:organisation_id].present? && Organisation.where(id: params[:organisation_id]).exists?
+      @new_organisation = Organisation.find(params[:organisation_id])
+    else
+      redirect_to user_path(@user)
+    end
+  end
+
 private
 
   def validate_attributes
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -155,7 +155,7 @@ class UsersController < ApplicationController
 
   def log_reassignment
     authorize @user
-    assigned_to_logs_count = @user.assigned_to_lettings_logs.count + @user.assigned_to_sales_logs.count
+    assigned_to_logs_count = @user.assigned_to_lettings_logs.visible.count + @user.assigned_to_sales_logs.visible.count
     return redirect_to user_organisation_change_confirmation_path(@user, organisation_id: params[:organisation_id]) if assigned_to_logs_count.zero?
 
     if params[:organisation_id].present? && Organisation.where(id: params[:organisation_id]).exists?
@@ -182,7 +182,7 @@ class UsersController < ApplicationController
 
   def organisation_change_confirmation
     authorize @user
-    assigned_to_logs_count = @user.assigned_to_lettings_logs.count + @user.assigned_to_sales_logs.count
+    assigned_to_logs_count = @user.assigned_to_lettings_logs.visible.count + @user.assigned_to_sales_logs.visible.count
 
     return redirect_to user_path(@user) if params[:organisation_id].blank? || !Organisation.where(id: params[:organisation_id]).exists?
     return redirect_to user_path(@user) if params[:log_reassignment].blank? && assigned_to_logs_count.positive?
@@ -193,7 +193,7 @@ class UsersController < ApplicationController
 
   def confirm_organisation_change
     authorize @user
-    assigned_to_logs_count = @user.assigned_to_lettings_logs.count + @user.assigned_to_sales_logs.count
+    assigned_to_logs_count = @user.assigned_to_lettings_logs.visible.count + @user.assigned_to_sales_logs.visible.count
 
     return redirect_to user_path(@user) if log_reassignment_params[:organisation_id].blank? || !Organisation.where(id: log_reassignment_params[:organisation_id]).exists?
     return redirect_to user_path(@user) if log_reassignment_params[:log_reassignment].blank? && assigned_to_logs_count.positive?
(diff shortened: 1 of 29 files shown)


app/controllers/users_controller.rb
```diff
@@ -56,20 +56,24 @@ def dpo; end
...
       if @user == current_user
         bypass_sign_in @user
         flash[:notice] = I18n.t("devise.passwords.updated") if user_params.key?("password")
-        if user_params.key?("email")
+        if user_params.key?("email") && user_params[:email] != @user.email
           flash[:notice] = I18n.t("devise.email.updated", email: @user.unconfirmed_email)
         end
 
-        redirect_to account_path
+        if updating_organisation?
```
RachaelBooth, line 72: Can a user update their own organisation? I guess yes if they are support.

app/controllers/users_controller.rb
```diff
@@ -144,6 +153,58 @@ def delete
     redirect_to users_organisation_path(@user.organisation), notice: I18n.t("notification.user_deleted", name: @user.name)
   end
 
+  def log_reassignment
+    authorize @user
```
RachaelBooth, line 157: Is this not covered by one of the `before_action` things on the controller?

app/models/user.rb
```diff
@@ -270,6 +280,47 @@ def phone_with_extension
...
+      when "reassign_stock_owner"
+        reassign_stock_owners(new_organisation, lettings_logs_to_reassign, sales_logs_to_reassign)
+      when "reassign_managing_agent"
+        reassign_managing_agents(new_organisation, lettings_logs_to_reassign, sales_logs_to_reassign)
+      when "unassign"
+        unassign_organisations(lettings_logs_to_reassign, sales_logs_to_reassign, current_organisation)
+      end
+
+      send_organisation_change_email(current_organisation, new_organisation, log_reassignment, logs_count)
+    rescue ActiveRecord::RecordInvalid => e
```
RachaelBooth: Are there any other plausible exceptions that we might need to handle?

app/policies/user_policy.rb
```diff
@@ -10,17 +10,15 @@ def edit_password?
...
-  end
-
   %w[
     edit_roles?
     edit_dpo?
     edit_key_contact?
   ].each do |method_name|
     define_method method_name do
-      (@current_user.data_coordinator? || @current_user.support?) && @user.active?
+      return true if @current_user.support?
```
RachaelBooth, line 19: Does active not matter for support? (Actually, does it matter for anyone since you can't log in while not active?)

app/controllers/users_controller.rb
kosiakkatrina: Hmm, I don't think it was explicit whether they should or shouldn't be able to move themselves in the ACs, but it might be useful on staging if people can move themselves around as we had to manually move people for testing purposes in the past. Let me know if there are any concerns with this tho

app/controllers/users_controller.rb
kosiakkatrina: We do run `authenticate_scope!` but for some user paths we use policies instead. I feel like it'a a bit cleaner than adding another if into `authenticate_scope!`

app/models/user.rb
kosiakkatrina: I changed it to StandardError, I think it should be most errors (?) would that work?

app/policies/user_policy.rb
kosiakkatrina: Oh, so I don't think it matters for current user, cause as you say, they should be able to login before they even get to this point. But for the user that is getting edited we have an AC to allow deactivated users to be moved. So I'm allowing other fields for deactivated user to be editable too (cause otherwise we'd only show organisation question on the edit page, which is a bit odd that support can change the org but nothing else)

So basically support users can edit deactivated and active users, whereas coordinators can only edit active as it was before.

Hope this makes sense

RachaelBooth (approved):

app/models/user.rb
```diff
@@ -292,8 +292,8 @@ def reassign_logs_and_update_organisation(new_organisation, log_reassignment)
     return unless new_organisation
 
     ActiveRecord::Base.transaction do
-      lettings_logs_to_reassign = assigned_to_lettings_logs
-      sales_logs_to_reassign = assigned_to_sales_logs
+      lettings_logs_to_reassign = assigned_to_lettings_logs.visible
+      sales_logs_to_reassign = assigned_to_sales_logs.visible
```
RachaelBooth, line 296: Does this cause problems if a user has pending logs from a bulk upload that they may later choose to load?

kosiakkatrina: Yes, that's a very good point! I'll follow up what we want to do with those on the ticket

app/models/forms/bulk_upload_lettings_resume/fix_choice.rb
```diff
@@ -56,7 +56,7 @@ def save!
       end
 
       def preflight_valid?
-        bulk_upload.choice != "create-fix-inline" && bulk_upload.choice != "bulk-confirm-soft-validations"
+        bulk_upload.choice.blank?
```
kosiakkatrina, line 59: This should basically prevent users from accessing the fix choice page if they made any kind of choice (either manually or by user being moved). I think it makes sense to discredit as many bulk uploads so that they don't linger with a possibility of the user stumbling upon the link and accidentally creating them. 

We might also separately want to update routing here if a log has setup errors, cause I don't think anything is stopping us from navigating to fix-choice page directly (?)
On top of that any bulk uploads for closed collection years should probably also not be accessible, or at least not uploadable

PR #2597, opened 2024-08-20
CLDC-3577: Add optional phone extension to users

RachaelBooth:


app/models/user.rb
```diff
@@ -19,6 +19,7 @@ class User < ApplicationRecord
   validates :password, presence: { if: :password_required? }
   validates :password, length: { within: Devise.password_length, allow_blank: true }
   validates :password, confirmation: { if: :password_required? }
+  validates :phone_extension, format: { with: /\A\d+\z/, allow_blank: true, message: I18n.t("validations.numeric.format", field: "") }
```
kosiakkatrina, line 22: Would it make sense to allow `#` as well?

kosiakkatrina (approved):

PR #2598, opened 2024-08-21
Set merge request as non processing if it fails

kosiakkatrina:


app/services/merge/merge_organisations_service.rb
```diff
@@ -22,6 +22,7 @@ def call
       end
       @absorbing_organisation.available_from = @merge_date if @absorbing_organisation_active_from_merge_date
       @absorbing_organisation.save!
+      throw ActiveRecord::RecordInvalid
```
kosiakkatrina: Will remove this before merging, but need it to test the rollback

Dinssa (approved):

PR #2599, opened 2024-08-21
CLDC-3534 Export organisation data

kosiakkatrina:
This builds on top of the refactoring here https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/2543 and adds organisation xml export service


app/services/exports/organisation_export_service.rb
```diff
@@ -0,0 +1,72 @@
+module Exports
+  class OrganisationExportService < Exports::XmlExportService
+    include Exports::OrganisationExportConstants
+    include CollectionTimeHelper
+
+    def export_xml_organisations(full_update: false)
+      recent_export = Export.order("started_at").last
```
RachaelBooth: Should this be filtering by collection?

kosiakkatrina: Yes, that would make sense!

RachaelBooth (approved):

PR #2600, opened 2024-08-22
CLDC-3609 Add existing absorbing organisation merge request question

kosiakkatrina:
Depending on whether existing or a new organisation is absorbing other orgs we want to call different merge organisation service flow. This PR adds the question to determine which flow gets called


app/views/merge_requests/existing_absorbing_organisation.html.erb
```diff
@@ -0,0 +1,27 @@
+<% content_for :before_content do %>
+  <% title = "Tell us if your organisation is merging" %>
```
Dinssa, line 2: Is this the right title for this page?

Dinssa (approved):

PR #2602, opened 2024-08-22
CLDC-3603 View merged users outcomes

kosiakkatrina:
Update total user count to be displayed on the merge request page
image

Display a breakdown of different users from merging organisations
image

### DIFF
--- a/app/views/merge_requests/user_outcomes.html.erb
+++ b/app/views/merge_requests/user_outcomes.html.erb
@@ -10,7 +10,7 @@
 </h1>
 
 <% unless @merge_request.status == "request_merged" || @merge_request.status == "processing" %>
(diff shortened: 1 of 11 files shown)


app/views/merge_requests/user_outcomes.html.erb
```diff
@@ -10,7 +10,7 @@
 </h1>
 
 <% unless @merge_request.status == "request_merged" || @merge_request.status == "processing" %>
-  <h2 class="govuk-heading-m">19 users after merge</h2>
+  <h2 class="govuk-heading-m"><%= @merge_request.total_visible_users_after_merge %> users after merge</h2>
```
Dinssa: Should we take into account the case of having a singular user and displaying "users" or "user"?

kosiakkatrina: For sure, good spot! Should be updated now!

Dinssa (approved):

PR #2605, opened 2024-08-23
CLDC-3001  Handle locations and logs on deactivate schemes

Dinssa:
This branch is to handle locations and update their status when the scheme they belong to has a deactivation period. Logs have already been handled.

The idea is that when a scheme is deactivated the locations are also deactivated on the same date. When a scheme is reactivated the locations are returned to their previous state. Locations already deactivated before the scheme deactivation date will not be affected. Whilst the scheme is deactivating soon, the locations will have the deactivating soon status. Similarly, whilst the scheme is reactivating soon, the locations will have the reactivating soon status. This PR updates the delete confirmation page to provide information about how many locations are being affected in addition to the count for affected logs.

# Before deactivation:
## Scheme (active)
Screenshot 2024-09-03 at 15 59 20

## Locations
Screenshot 2024-09-03 at 15 58 58

# Deactivation confirmation:
Screenshot 2024-09-03 at 15 58 34

# After deactivation:
## Scheme (deactivated)
Screenshot 2024-09-03 at 16 03 32

## Locations
Screenshot 2024-09-03 at 16 03 54

## Example location
The scheme deactivation periods are included in the location's availability text. Only those between the date the location is first available and the location's last deactivation date (if present).
Screenshot 2024-09-03 at 16 04 07


spec/requests/schemes_controller_spec.rb
Dinssa, line 1: I have changed this to show the option to reactivate a scheme, whilst it has a status of deactivating soon. So a user does not have to wait until the day it is deactivated to add a reactivation period.

app/controllers/schemes_controller.rb
```diff
@@ -51,17 +51,34 @@ def new_deactivation
...
+      location.status_at(@deactivation_date) == :active
+    end
+
+    locations_deactivating_after_deactivation_date, remaining_locations = remaining_locations.partition do |location|
+      location.status_at(@deactivation_date) == :deactivating_soon || location.status_at(@deactivation_date) == :reactivating_soon
+    end
+
+    locations_startdate_after_deactivation_date, = remaining_locations.partition do |location|
+      location.status_at(@deactivation_date) == :activating_soon
+    end
```
Dinssa: Using partition, grabs the relevant locations for the three scenarios that could mean the location is affected. The last partition drops locations that are not relevant.

app/helpers/locations_helper.rb
```diff
@@ -100,14 +100,37 @@ def edit_location_text(scheme, user)
...
+
+  def scheme_deactivation_periods(location, location_deactivation_periods)
+    return [] unless location.scheme.scheme_deactivation_periods.any?
+
+    location_deactivation_date = location_deactivation_periods.last&.deactivation_date
+    periods = remove_nested_periods(location.scheme.scheme_deactivation_periods.sort_by(&:deactivation_date))
+    periods.select do |period|
+      period.deactivation_date >= location.available_from && (location_deactivation_date.nil? || period.deactivation_date <= location_deactivation_date)
+    end
+  end
```
Dinssa, line 127: Returns scheme deactivation periods between the location's available from date and the location's own last deactivation date (if present).

app/models/location.rb
```diff
@@ -40,6 +40,7 @@ class Location < ApplicationRecord
       filtered_records = filtered_records
       .left_outer_joins(:location_deactivation_periods)
       .joins(scheme: [:owning_organisation])
+      .left_outer_joins(scheme: :scheme_deactivation_periods)
```
Dinssa, line 43: Location status now also relies on the status of the scheme and it's deactivation periods.

app/controllers/schemes_controller.rb
```diff
@@ -51,17 +51,34 @@ def new_deactivation
...
-    @affected_logs = @scheme.lettings_logs.visible.after_date(params[:deactivation_date])
-    if @affected_logs.count.zero?
+    @deactivation_date = Time.zone.parse(params[:deactivation_date])
+    @affected_logs = @scheme.lettings_logs.visible.after_date(@deactivation_date)
+    @deactivation_date_type = params[:deactivation_date_type]
+
+    scheme_locations = @scheme.locations.confirmed
+
+    locations_active_on_deactivation_date, remaining_locations = scheme_locations.partition do |location|
+      location.status_at(@deactivation_date) == :active
```
kosiakkatrina: What if the location is deactivating soon on deactivation date? Wouldn't that mean that it's technically still active?
edit: never mind, read the next line 🤦‍♀️

app/controllers/schemes_controller.rb
kosiakkatrina: Would select do the same thing without the empty variable instead of the last partition?

app/models/location.rb
```diff
@@ -84,19 +102,23 @@ class Location < ApplicationRecord
 
   scope :active_status, lambda {
     where.not(id: joins(:location_deactivation_periods).reactivating_soon.pluck(:id))
-    .where.not(id: joins(scheme: [:owning_organisation]).deactivated_by_organisation.pluck(:id))
-    .where.not(id: joins(:location_deactivation_periods).deactivated_directly.pluck(:id))
-    .where.not(id: incomplete.pluck(:id))
-    .where.not(id: joins(:location_deactivation_periods).deactivating_soon.pluck(:id))
-    .where.not(id: activating_soon.pluck(:id))
+         .where.not(id: joins(scheme: [:scheme_deactivation_periods]).reactivating_soon_by_scheme.pluck(:id))
```
kosiakkatrina, line 105: Should this also have `where.not(reactivating_soon_directly?)` ?

app/models/location.rb
```diff
@@ -84,19 +102,23 @@ class Location < ApplicationRecord
...
+         .where.not(id: joins(:location_deactivation_periods).merge(Location.deactivating_soon_directly).pluck(:id))
+         .where.not(id: joins(scheme: %i[owning_organisation scheme_deactivation_periods]).merge(Location.deactivating_soon_by_scheme).pluck(:id))
+         .where.not(id: activating_soon.pluck(:id))
   }
 
   scope :active, lambda { |date = Time.zone.now|
-    where.not(id: joins(:location_deactivation_periods).reactivating_soon(date).pluck(:id))
-         .where.not(id: joins(scheme: [:owning_organisation]).deactivated_by_organisation.pluck(:id))
-         .where.not(id: joins(:location_deactivation_periods).deactivated_directly(date).pluck(:id))
+    where.not(id: joins(scheme: [:scheme_deactivation_periods]).reactivating_soon_by_scheme.pluck(:id))
```
kosiakkatrina: Would `.where(scheme: Scheme.active(date))` already remove anything that's `reactivating_soon_by_scheme` and `deactivated_by_organisation` ?

app/views/schemes/deactivate_confirm.html.erb
```diff
@@ -2,16 +2,40 @@
...
+      <h1 class="govuk-heading-l">
+        <span class="govuk-caption-l"><%= @scheme.service_name %></span>
+        <% sentence_parts = [] %>
+        <% if @affected_logs.count > 0 %>
+          <% sentence_parts << pluralize(@affected_logs.count, "log") %>
+        <% end %>
+        <% if @affected_locations.count > 0 %>
+          <% sentence_parts << pluralize(@affected_locations.count, "location") %>
+        <% end %>
+        This change will affect <%= sentence_parts.join(" and ") %>.
```
kosiakkatrina: Would it make sense to extract this into a helper?

app/controllers/schemes_controller.rb
Dinssa: Actually it would make more sense for the deactivating soon to be in this partition. However, since we no longer need to display each of these three scenarios on the next page, I will put them all together into a single select.

app/models/location.rb
Dinssa: Yes I think so, will remove them.

app/models/location.rb
kosiakkatrina: Sorry, missed the line above, ignore me, this looks fine!

kosiakkatrina (approved):

app/models/validations/shared_validations.rb
```diff
@@ -102,7 +102,11 @@ def inactive_status(date, resource)
     return unless %i[reactivating_soon activating_soon deactivated].include?(status)
 
     closest_reactivation = resource.last_deactivation_before(date)
-    open_deactivation = resource.open_deactivation
+    open_deactivation = if resource.is_a?(Location)
+                          resource.open_deactivation || resource.scheme.open_deactivation
+                        else
+                          resource.open_deactivation
+                        end
```
Dinssa, line 109: Accounts for the deactivation coming from the scheme and not the location.

PR #2606, opened 2024-08-23
CLDC-3605 View relationships page

Dinssa:
This branch adds as page to view all stock owners and managing agents of the absorbing org and merging orgs. It shows the total stock owners and managing agents that will be present on the new org after the merge.

In the cases where any merging org or absorbing org is also a stock owner/managing agent they'll be excluded from the count (as seen below with the managing agent count). That relationship will no longer exist.

The relationship outcomes page:
Screenshot 2024-08-29 at 10 26 24

The details page:
Screenshot 2024-08-29 at 10 28 11

### DIFF
--- a/app/helpers/merge_requests_helper.rb
+++ b/app/helpers/merge_requests_helper.rb
@@ -27,7 +27,7 @@ module MergeRequestsHelper
       { label: "Total users after merge", value: display_value_or_placeholder(merge_request.total_users_label), action: merge_outcome_action(merge_request, "user_outcomes") },
       { label: "Total schemes after merge", value: display_value_or_placeholder(merge_request.total_schemes), action: { text: "View", href: "#", visually_hidden_text: "total schemes after merge" } },
       { label: "Total logs after merge", value: merge_request.total_lettings_logs.present? || merge_request.total_sales_logs.present? ? "#{merge_request.total_lettings_logs} lettings logs<br>#{merge_request.total_sales_logs} sales logs".html_safe : display_value_or_placeholder(nil), action: { text: "View", href: "#", visually_hidden_text: "total logs after merge" } },
-      { label: "Total stock owners & managing agents after merge", value: merge_request.total_stock_owners.present? || merge_request.total_managing_agents.present? ? "#{merge_request.total_stock_owners} stock owners<br>#{merge_request.total_managing_agents} managing agents".html_safe : display_value_or_placeholder(nil), action: { text: "View", href: "#", visually_hidden_text: "total stock owners & managing agents after merge" } },
+      { label: "Total stock owners & managing agents after merge", value: display_value_or_placeholder(merge_request.total_stock_owners_managing_agents_label), action: merge_outcome_action(merge_request, "relationship_outcomes") },
     ]
   end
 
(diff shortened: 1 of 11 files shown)


app/helpers/merge_requests_helper.rb
```diff
@@ -104,10 +104,74 @@ def link_to_merging_organisation_users(organisation)
   end
 
   def total_users_after_merge_text(merge_request)
-    count = merge_request.total_visible_users_after_merge
+    count = merge_request.total_visible_users_after_merge || 0
```
kosiakkatrina: Wouldn't total_visible_users_after_merge return 0 if there aren't any anyway?

Dinssa: For the stock_owners and managing_agents methods below, in the view nothing/blank would be shown instead of 0. Added it there, added it here too to keep things the similar:
```
def total_stock_owners_after_merge_text(merge_request)
    count = merge_request.total_stock_owners_after_merge || 0
    "#{"#{count} stock owner".pluralize(count)} after merge"
  end

  def total_managing_agents_after_merge_text(merge_request)
    count = merge_request.total_managing_agents_after_merge || 0
    "#{"#{count} managing agent".pluralize(count)} after merge"
  end
```

app/helpers/merge_requests_helper.rb
```diff
@@ -104,10 +104,74 @@ def link_to_merging_organisation_users(organisation)
...
+                    "View all #{relationship_count} #{org.name} #{relationship_type.humanize(capitalize: false).pluralize} (opens in a new tab)"
+                  end
+      html_content += "#{govuk_link_to(link_text, send(relationship_path_helper, org), target: '_blank')}<br><br>"
+    end
+
+    html_content.html_safe
+  end
+
+  def duplicate_relationships(merge_request, relationship_type)
+    relationships = merge_request.absorbing_organisation.send(relationship_type.pluralize) + merge_request.merging_organisations.flat_map { |org| org.send(relationship_type.pluralize) }
```
kosiakkatrina, line 160: Maybe call this `related_organisations` instead of `relationships`? I'm assuming this returns orgs?

app/helpers/merge_requests_helper.rb
```diff
@@ -104,10 +104,74 @@ def link_to_merging_organisation_users(organisation)
...
+    relationships.each do |relationship|
+      if seen.include?(relationship)
+        duplicates << relationship
+      else
+        seen.add(relationship)
+      end
+    end
+
+    duplicates
+  end
```
kosiakkatrina: Would something like this do a similar thing?
`relationships.group_by { |org| org }.select { |_, occurrences| occurrences.size > 1 }.keys`

I feel like that's easier to read, we can see that we group the orgs and select the ones where they occur more than once. But if it's in fact harder to read I'm happy for it to stay as is

app/helpers/merge_requests_helper.rb
```diff
@@ -124,7 +188,7 @@ def link_to_merging_organisation_schemes(organisation)
   end
 
   def total_schemes_after_merge_text(merge_request)
-    count = merge_request.total_visible_schemes_after_merge
+    count = merge_request.total_visible_schemes_after_merge || 0
```
kosiakkatrina: Same here, does total_visible_schemes_after_merge ever return nothing?

app/models/merge_request.rb
```diff
@@ -61,8 +61,29 @@ def total_visible_users_after_merge
     absorbing_organisation.users.visible.count + merging_organisations.sum { |org| org.users.visible.count }
   end
 
+  def total_stock_owners_after_merge
+    return total_stock_owners if status == STATUS[:request_merged] || status == STATUS[:processing]
+
+    stock_owners = absorbing_organisation.stock_owners.visible + merging_organisations.flat_map { |org| org.stock_owners.visible }
+    stock_owners.uniq.count
+  end
```
kosiakkatrina: If absorbing organisation was a stock owner of any of the merging organisations (or vice versa), after the merge happens I think that relationship would no longer be relevant, so we might want to filter out absorbing org and merging orgs from stock owners list before counting (?)

I think same with managing agents

app/models/merge_request.rb
```diff
@@ -84,7 +105,7 @@ def total_visible_schemes_after_merge
   end
 
   def total_schemes_label
-    "#{total_visible_schemes_after_merge} #{'Scheme'.pluralize(total_visible_schemes_after_merge)}"
+    "#{total_visible_schemes_after_merge} #{'scheme'.pluralize(total_visible_schemes_after_merge)}"
```
kosiakkatrina, line 87: Nice catch, I thought it was uppercase 🤦‍♀️

kosiakkatrina (commented):
Coul we add some tests for model and helper methods? And maybe a request test for the page?

app/helpers/merge_requests_helper.rb
Dinssa: Good idea that explains better

app/helpers/merge_requests_helper.rb
Dinssa: That would equally work, the idea behind using Set was to be slightly more efficient, time-wise, without overheads (e.g. no hash with keys). It might be insignificant with the expected number of orgs, so happy to change it.

app/helpers/merge_requests_helper.rb
Dinssa: This method is supposed to be reused for both stock owners and managing agents

app/helpers/merge_requests_helper.rb
```diff
@@ -108,6 +108,72 @@ def total_users_after_merge_text(merge_request)
...
+  end
+
+  def total_managing_agents_after_merge_text(merge_request)
+    count = merge_request.total_managing_agents_after_merge
+    "#{"#{count} managing agent".pluralize(count)} after merge"
+  end
+
+  def related_organisations(merge_request, relationship_type)
+    organisations =  merge_request.absorbing_organisation.send(relationship_type.pluralize).visible + merge_request.merging_organisations.flat_map { |org| org.send(relationship_type.pluralize).visible }
+    organisations += [merge_request.absorbing_organisation] + merge_request.merging_organisations
```
Dinssa, line 123: The related organisations returns all organisations including the absorbing org itself and the merging orgs themselves. To account for one or the other being a stock owner or managing agents of the other.

app/models/merge_request.rb
```diff
@@ -98,4 +98,43 @@ def organisations_without_schemes
...
+    filtered_absorbing_relationships = absorbing_relationships.reject do |relationship|
+      merging_relationships.include?(relationship) || merging_organisations.include?(relationship)
+    end
+
+    filtered_merging_relationships = merging_relationships.reject do |relationship|
+      absorbing_relationships.include?(relationship) || relationship == absorbing_organisation || merging_organisations.include?(relationship)
+    end
+
+    (filtered_absorbing_relationships + filtered_merging_relationships).uniq
+  end
```
Dinssa, line 112: This method filters out relationships:

- from absorbing orgs (absorbing_relationships) that are also present in merging orgs (merging_relationships)
- from merging orgs (merging_relationships) that are also present in absorbing orgs (absorbing_relationships)
- combines those filtered lists and removes duplicates

This method is used for both stock_owners and managing_agents

kosiakkatrina (approved):

PR #2613, opened 2024-08-30
CLDC-3387: Allow support users to create notifications

RachaelBooth:


app/views/notifications/_form.html.erb
```diff
@@ -0,0 +1,29 @@
...
+
+    <%= f.govuk_text_field :title,
+                           label: { text: "Title", size: "m" },
+                           hint: { text: "Use markdown for links to existing pages" } %>
+
+    <%= f.govuk_check_boxes_fieldset :show_on_unauthenticated_pages, multiple: false, legend: { text: "Display Options" } do %>
+      <%= f.govuk_check_box :show_on_unauthenticated_pages, 1, 0, multiple: false, label: { text: "Show this notification on unauthenticated pages, for example the start page" } %>
+    <% end %>
+
+    <span class="govuk-caption-m govuk-!-margin-bottom-6">View <%= govuk_link_to "markdown syntax guide", "https://www.markdownguide.org/basic-syntax/" %></span>
```
kosiakkatrina: On gouvk design system the guidance for external links is:

> If it’s an external link to a non-government website, make that clear in the link text. For example, ‘read advice on writing link text from [name of organisation]’. 

Would it make sense to make it a bit more obvious that this is an external link?

kosiakkatrina (approved):

app/views/notifications/_form.html.erb
RachaelBooth: I've tried putting the domain in, since there isn't really an org name - will see what Callum or others think.

PR #2622, opened 2024-09-05
CLDC-3614: Allow markdown page content when creating notifications

RachaelBooth:


db/migrate/20240904135306_add_show_additional_page_column_to_notifications.rb
```diff
@@ -0,0 +1,5 @@
+class AddShowAdditionalPageColumnToNotifications < ActiveRecord::Migration[7.0]
+  def change
+    add_column :notifications, :show_additional_page, :boolean
```
RachaelBooth, line 3: I tried to continue just inferring it, but this worked out nicer all round.

I've not written a task to update existing notifications, because we're likely not to have any active ones at the time of release (and if we do it would be fairly easy to just sort them manually - it doesn't seem worth the effort of a script).

kosiakkatrina (approved):

PR #2628, opened 2024-09-09
CLDC-3633 Full stops in error messages

Dinssa:


config/locales/en.yml
```diff
@@ -728,12 +732,12 @@ Make sure these answers are correct."
       title_text: "You told us that the monthly charges were %{mscharge}."
       hint_text: "This is higher than we would expect."
     student_not_child:
-      title_text: "You told us this person is a student aged beween 16 and 19"
```
kosiakkatrina, line 748: nice

spec/models/validations/sales/setup_validations_spec.rb
```diff
@@ -119,15 +119,15 @@
...
+          expect(record.errors["saledate"]).to include(match "Enter a date within the 23/24 or 24/25 collection years, which is between 1st April 2023 and 31st March 2025.")
         end
 
         it "can edit already created logs for the previous collection year" do
           record.saledate = Time.zone.local(2024, 1, 2)
           record.save!(validate: false)
           record.saledate = Time.zone.local(2024, 1, 1)
           setup_validator.validate_saledate_collection_year(record)
-          expect(record.errors["saledate"]).not_to include(match "Enter a date within the 24/25 collection year, which is between 1st April 2024 and 31st March 2025")
+          expect(record.errors["saledate"]).not_to include(match "Enter a date within the 24/25 collection year, which is between 1st April 2024 and 31st March 2025.")
```
kosiakkatrina, line 130: There are a couple of tests like this which wouldn't have failed with the removal of `.` . This one has `match` so would probably still be testing correctly, but I checked for others and updated the few that I found. 
We mostly seem to test absence with `to be_empty` etc

config/locales/en.yml
```diff
@@ -50,8 +50,8 @@ en:
     errors:
       models:
         bulk_upload/row_parser: &bulk_upload__row_parser__base
-          inclusion: Enter a valid value for %{question}
-          spreadsheet_dupe: This is a duplicate of a log in your file
+          inclusion: "Enter a valid value for %{question}"
```
Dinssa, line 53: This stays without a full stop because the question itself will have one/a question mark.

config/locales/en.yml
```diff
@@ -301,270 +301,269 @@ en:
...
       age:
         earnings_over_hard_max: "The household’s income of %{earnings} %{frequency} is too high for the number of adults. Change either the household income or the age of the tenants."
       hhmemb:
         earnings:
           over_hard_max: "The household’s income of %{earnings} %{frequency} is too high for this number of tenants. Change either the household income or number of tenants."
           under_hard_min: "The household’s income of %{earnings} %{frequency} is too low for this number of tenants. Change either the household income or number of tenants."
       brent:
-        below_hard_min: "Rent is below the absolute minimum expected for a property of this type. Please check the rent, rent period, local authority and (if general needs) number of bedrooms"
-        above_hard_max: "Rent is higher than the absolute maximum expected for a property of this type. Please check the rent, rent period, local authority and (if general needs) number of bedrooms"
+        below_hard_min: "Rent is below the absolute minimum expected for a property of this type. Please check the rent, rent period, local authority and (if general needs) number of bedroom."
```
Dinssa, line 415: This should change back to "bedrooms".

spec/models/validations/sales/setup_validations_spec.rb
RachaelBooth: Yeah, this seems the biggest risk so glad to know you've checked! I would expect us mostly to be using `to be_empty` though 🤞.

RachaelBooth (approved):

PR #2629, opened 2024-09-09
CLDC 3623 Support user bulk uploading

Dinssa:
This branch adds the ability for support users to bulk upload on behalf of external users.

## New button
The "Upload (letting/sales) logs in bulk" button link is visible on the logs view pages for specific organisations.

Screenshot 2024-09-11 at 14 34 09

### Allowing support users to upload on behalf of organisations
This branch now allows support users to upload logs for all organisations affiliated with the organisation from which they first clicked the bulk upload button.

### Additional validation on log assigned to, when uploaded by a support user
In the past if the user a log should be assigned to was left unanswered we'd set it to the user submitting the bulk upload. As the user now could be a support user we wouldn't want to assign the logs to them so we add an extra validation to make sure the user is set.

Screenshot 2024-09-11 at 15 19 17

### Different error message for support users when validating permission to upload for an organisation
When the owning organisation of a log is not affiliated with the organisation it is being uploaded for the error message states "You do not have permission to add logs for this owning organisation". Since all support users should be able to upload logs for any organisation receiving this error message would be misleading. I have added a different error message support users would see that explains better why a log creation is being prevented.

Screenshot 2024-09-13 at 17 31 27


app/controllers/bulk_upload_lettings_logs_controller.rb
```diff
@@ -51,7 +53,7 @@ def form
               when "needstype"
                 Forms::BulkUploadLettings::Needstype.new(form_params)
               when "upload-your-file"
-                Forms::BulkUploadLettings::UploadYourFile.new(form_params.merge(current_user:))
+                Forms::BulkUploadLettings::UploadYourFile.new(form_params.merge(current_user:, request:))
```
Dinssa, line 56: Pass the request to this page so that I can set the back link based off the organisation id.

app/models/forms/bulk_upload_lettings/upload_your_file.rb
```diff
@@ -39,6 +46,7 @@ def save!
           year:,
           needstype:,
           filename: file.original_filename,
+          organisation_id: organisation_id || current_user.organisation_id,
```
Dinssa: If uploaded by a support user, they'll be uploading on behalf of an organisation hence we need to store the id of that organisation. Otherwise, this is being uploaded by a non-support user and is defaulted to the organisation of the current user, ~~but also not used anywhere else. We only care about the organisation_id when the user is a support user.~~

app/services/bulk_upload/lettings/year2024/row_parser.rb
```diff
@@ -889,12 +896,15 @@ def validate_owning_org_data_given
...
   def validate_owning_org_permitted
-    if owning_organisation && !bulk_upload.user.organisation.affiliated_stock_owners.include?(owning_organisation)
-      block_log_creation!
+    return unless owning_organisation
 
-      if errors[:field_1].blank?
-        errors.add(:field_1, "You do not have permission to add logs for this owning organisation", category: :setup)
-      end
+    bulk_upload_organisation = bulk_upload.user.support? ? Organisation.find(bulk_upload.organisation_id) : bulk_upload.user.organisation
+    return if bulk_upload_organisation.affiliated_stock_owners.include?(owning_organisation)
```
Dinssa: As a support user you are uploading on behalf of another organisation so we need to find and check the affiliated stock owners against that organisation not the support user's organisation.

app/components/create_log_actions_component.html.erb
```diff
@@ -3,7 +3,7 @@
     <% if create_button_href.present? %>
       <%= govuk_button_to create_button_copy, create_button_href, class: "govuk-!-margin-right-6" %>
     <% end %>
-    <% if upload_button_href.present? %>
+    <% if upload_button_href.present? && !user.support? %>
```
RachaelBooth, line 6: It seems odd to me that this has changed - were we previously just not defining upload_button_href for support users?

app/controllers/bulk_upload_lettings_logs_controller.rb
```diff
@@ -3,7 +3,9 @@ class BulkUploadLettingsLogsController < ApplicationController
   before_action :validate_data_protection_agrement_signed!
 
   def start
-    if have_choice_of_year?
+    if current_user.support? && params[:organisation_id].present?
+      redirect_to bulk_upload_lettings_log_path(id: "upload-your-file", form: { year: current_year }, organisation_id: params[:organisation_id])
```
RachaelBooth: I think this might need to go to the year choice as well when there are multiple years open - does this actually need a separate case for support?

app/models/forms/bulk_upload_lettings/upload_your_file.rb
```diff
@@ -20,8 +22,13 @@ def view_path
       end
 
       def back_path
-        page_id = year == 2022 ? "needstype" : "prepare-your-file"
-        bulk_upload_lettings_log_path(id: page_id, form: { year:, needstype: })
+        organisation_id = request.query_parameters[:organisation_id]
+        if organisation_id.present?
+          lettings_logs_organisation_path(organisation_id)
+        else
+          page_id = year == 2022 ? "needstype" : "prepare-your-file"
```
RachaelBooth: We could probably just get rid of this line and always go to "prepare-your-file" now.

app/models/forms/bulk_upload_lettings/upload_your_file.rb
```diff
@@ -20,8 +22,13 @@ def view_path
       end
 
       def back_path
-        page_id = year == 2022 ? "needstype" : "prepare-your-file"
-        bulk_upload_lettings_log_path(id: page_id, form: { year:, needstype: })
+        organisation_id = request.query_parameters[:organisation_id]
+        if organisation_id.present?
+          lettings_logs_organisation_path(organisation_id)
```
RachaelBooth: Again, does this need to be different? A support user will be going through the same flow by this page (might need to pass the organisation_id through) so I'd expect back to be to the same place as other users (it'll be different on the "prepare-your-file" page I guess).

app/services/bulk_upload/lettings/year2024/row_parser.rb
```diff
@@ -580,6 +581,12 @@ def validate_assigned_to_exists
     end
   end
 
+  def validate_assigned_to_when_support
+    if field_3.blank? && bulk_upload.user.support?
+      errors.add(:field_3, category: :setup, message: I18n.t("validations.not_answered", question: "what is the CORE username of the account this letting log should be assigned to?"))
+    end
```
RachaelBooth, line 589: It might be worth this validating the output of `assigned_to` (although possibly as well as whether the field was blank), to deal with cases where there was data in the field but we couldn't find a user - although maybe that would already error?

app/services/bulk_upload/lettings/year2024/row_parser.rb
RachaelBooth: Given that this is checking whether the user has permission to upload logs that have this owning organisation, I'm not sure we need to check this for support (as support users always should be able to). We will need to check that the owning and managing orgs are related to each other and that the assigned user is at a related org - I think at least some of that should already be being checked though.

app/services/bulk_upload/lettings/year2024/row_parser.rb
```diff
@@ -1135,7 +1145,7 @@ def attributes_for_log
     attributes["renewal"] = renewal
     attributes["scheme"] = scheme
     attributes["location"] = location
-    attributes["assigned_to"] = assigned_to || bulk_upload.user
+    attributes["assigned_to"] = bulk_upload.user.support? ? nil : (assigned_to || bulk_upload.user)
```
RachaelBooth: Should this be `assigned_to || (bulk_upload.user.support? ? nil : bulk_upload.user)`?

app/components/create_log_actions_component.html.erb
Dinssa: This button is shown for all users no distinction made between support and non-support.

```
  def upload_button_href
    case log_type
    when "lettings"
      bulk_upload_lettings_log_path(id: "start")
    when "sales"
      bulk_upload_sales_log_path(id: "start")
    end
  end
```

I could change the method to return if support instead of adding `!user.support?`

```
def upload_button_href
    return nil if user.support?

    case log_type
    when "lettings"
      bulk_upload_lettings_log_path(id: "start")
    when "sales"
      bulk_upload_sales_log_path(id: "start")
    end
 end
```
Would be the same result, no button shown.

app/services/bulk_upload/lettings/year2024/row_parser.rb
Dinssa: I believe the validate_assigned_to_exists method runs first, so will already produce an error if field_3 is not blank and a user if not found.

```
def validate_assigned_to_exists
    return if field_3.blank?

    unless assigned_to
      errors.add(:field_3, "User with the specified email could not be found")
    end
end

...

def assigned_to
    @assigned_to ||= User.where("lower(email) = ?", field_3&.downcase).first
end
```

app/models/forms/bulk_upload_lettings/upload_your_file.rb
Dinssa: I tried to skip the prepare-your-file page for support users, but perhaps the work to do that is overkill and could be simpler, I'll look into it. The back link from this upload-your-file page would need to change, if skipping the prepare-your-file page and the rest could be the same whilst passing organisation_id.

app/services/bulk_upload/lettings/year2024/row_parser.rb
Dinssa: This lines goal is not check if the user is a support user, but that if they are, the owning organisation in the row is affiliated with the organisation this bulk upload was started from (by using organisation id stored on the bulk upload to get the organisation).

If support user's can upload any log, they still shouldn't be able to be given a file by a user which includes logs for organisations that the user is not affiliated with. This check prevents a log in the BU that is not affiliated with the organisation the bulk upload was started from, from also being uploaded at the same time.

app/controllers/bulk_upload_lettings_logs_controller.rb
Dinssa: I've reworked this to also go back to the year page should it be needed, the support users will not skip to the upload your file page, and will begin on /start. If a year needs picking then they'll go to the year page otherwise the prepare your file page.

app/models/forms/bulk_upload_lettings/upload_your_file.rb
Dinssa: Now users go to the /start route instead of /upload-your-file directly.

app/services/bulk_upload/lettings/year2024/row_parser.rb
Dinssa: Changed

app/models/forms/bulk_upload_lettings/upload_your_file.rb
Dinssa: Removed

app/services/bulk_upload/lettings/year2024/row_parser.rb
RachaelBooth: Ok, so we are restricting support user uploads to files that would be valid for a specific org that they have specified on upload? That does make sense, as you say to prevent a support user uploading something that another user had been prevented from for good reason.

That said, the error the support user will see here is wrong, which seems confusing - they do in fact have permission to add logs for this owning organisation, whichever one it is, the reason we aren't letting them is that the org is not affiliated with whatever org this bulk upload was set up against.

Might be more of a design question on whether on not we should have a different message - I can see the argument for showing what another user would see.

On the same basis, it would be nice not to special case support users where they aren't actually a special case - I think we're always setting organisation id on the upload, could this check just always work based on that id?

app/models/forms/bulk_upload_lettings/checking_file.rb
```diff
@@ -6,13 +6,18 @@ class CheckingFile
...
+      attribute :organisation_id, :integer
 
       def view_path
         "bulk_upload_lettings_logs/forms/checking_file"
       end
 
       def back_path
-        bulk_upload_lettings_log_path(id: "start")
+        if organisation_id.present?
+          lettings_logs_organisation_path(organisation_id)
```
RachaelBooth, line 17: Is there a way to reach this with an organisation_id present if you're not a support user? If so, where does this end up taking you?

spec/services/bulk_upload/lettings/year2024/row_parser_spec.rb
```diff
@@ -1439,6 +1439,41 @@
...
+          create(:organisation_relationship, parent_organisation: owning_org, child_organisation: affiliated_org)
+          bulk_upload.update!(organisation_id: affiliated_org.id, user: unaffiliated_support_user)
+        end
+
+        it "does not block log creation and does not add an error to field_1" do
+          parser = described_class.new(attributes)
+          parser.valid?
+          expect(parser.errors[:field_1]).not_to include("You do not have permission to add logs for this owning organisation")
+        end
+      end
```
RachaelBooth, line 1495: Should there also be a test for support users having to provide a value for assigned_to?

app/components/create_log_actions_component.html.erb
RachaelBooth: Oh, I see, I had forgotten that we were currently seeing the button as support in the general logs list context (just not able to use it unless we were uploading for our own org) - I was confused that a ticket about adding this functionality was doing something that looked more like removing it.

I don't mind where the support user check is in that case, here is fine.

app/models/forms/bulk_upload_lettings/checking_file.rb
Dinssa: Ah yes that is possible, since I'm now setting a default organisation id as the current user's organisation, so will change this to check if the user is support instead. Organisation id will always be present.

app/services/bulk_upload/lettings/year2024/row_parser.rb
Dinssa: I'll adjust this to use organisation id for all users and try adding a separate error message to state that the reason is because the owning organisation is not affiliated with the specified organisation.

app/services/bulk_upload/sales/year2024/row_parser.rb
```diff
@@ -1300,12 +1301,17 @@ def validate_owning_org_owns_stock
...
+    block_log_creation!
+
+    return if errors[:field_1].present?
+
+    if bulk_upload.user.support?
+      errors.add(:field_1, "This owning organisation is not affiliated with #{bulk_upload_organisation.name}", category: :setup)
+    else
+      errors.add(:field_1, "You do not have permission to add logs for this owning organisation", category: :setup)
     end
   end
```
Dinssa, line 1318: I now use the bulk upload organisation id attribute to check if the owning organisation is affiliated, since for each bulk upload organisation_id is either the specified org (support) or the user's own org (non-support). Should this fail the appropriate error message is given. I've added a screenshot of the message in the PR description.

app/models/forms/bulk_upload_lettings/upload_your_file.rb
Dinssa: We now use organisation id to determine the organisation the bulk upload is for when validating.

RachaelBooth (approved):

PR #2630, opened 2024-09-09
CLDC-3607: Allow support users to delete notifications (by setting their end date)

RachaelBooth:


app/controllers/notifications_controller.rb
```diff
@@ -62,6 +62,22 @@ def update
     end
   end
 
+  def delete_confirmation
+    @notification = Notification.find_by(id: params[:notification_id])
```
kosiakkatrina: Would this make sense to extract into a before action?

kosiakkatrina (approved):

app/controllers/notifications_controller.rb
RachaelBooth: Yeah, probably at this point

PR #2631, opened 2024-09-11
CLDC-3622 Update bulk upload guidance text

Dinssa:


app/views/bulk_upload_shared/guidance.html.erb
```diff
@@ -81,8 +81,8 @@
...
-        <p class="govuk-body">If none of your errors are in fields 1 through 17, you can choose how to fix the errors. You can either fix them in the CSV and reupload, or create partially complete logs and answer the remaining questions on the CORE site. Any errors that affect a significant number of logs will be featured in the error report’s summary tab to help you decide.</p>
+        <p class="govuk-body">If your file has errors on fields 1 through 15 for lettings, or 1 through 18 for sales, you must fix these in the CSV. This is because we need to know these answers to validate the rest of the data. Any errors in these fields will be featured in the error report’s summary tab.</p>
+        <p class="govuk-body">If none of your errors are in fields 1 through 15 for lettings, or 1 through 18 for sales, you can choose how to fix the errors. You can either fix them in the CSV and reupload, or create partially complete logs and answer the remaining questions on the CORE site. Any errors that affect a significant number of logs will be featured in the error report’s summary tab to help you decide.</p>
```
RachaelBooth, line 85: Might be worth checking if there's a useful description we can use that won't vary by year, depending on whether we think the number of setup fields is likely to change again?

RachaelBooth (approved):

PR #2637, opened 2024-09-13
CLDC-3611 Add performance tests

kosiakkatrina:
This adds a job to staging pipeline that runs a performance script against staging after it's deployed. 
The script uses ApacheBenchmark to make some concurrent requests, there a test that runs concurrent get on sales and lettings with some filters.
There’s also a test for concurrent updates on a log - we get the newest completed lettings log and make concurrent updates to tenancy code.

### DIFF
--- a/.github/workflows/staging_pipeline.yml
+++ b/.github/workflows/staging_pipeline.yml
@@ -404,7 +404,7 @@ jobs:
               "command": ["bash", "-c", "export email=$STAGING_PERFORMANCE_TEST_EMAIL && export password=$STAGING_PERFORMANCE_TEST_PASSWORD && sh ./lib/tasks/performance_test.sh"]
             }]
           }'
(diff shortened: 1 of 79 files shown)


.github/workflows/staging_pipeline.yml
```diff
@@ -364,3 +369,48 @@ jobs:
...
+        run: |
+          echo $cluster
+          network=$(aws ecs describe-services --cluster $cluster --services $service --query services[0].networkConfiguration)
+          overrides='{
+            "containerOverrides": [{
+              "name": "app",
+              "command": ["bash", "-c", "export email=$STAGING_PERFORMANCE_TEST_EMAIL && export password=$STAGING_PERFORMANCE_TEST_PASSWORD && sh ./lib/tasks/performance_test.sh"]
+            }]
+          }'
+          arn=$(aws ecs run-task --cluster $cluster --task-definition $ad_hoc_task_definition --network-configuration "$network" --overrides "$overrides" --group migrations --launch-type FARGATE --query tasks[0].taskArn)
```
RachaelBooth: This might want a different group?

RachaelBooth (approved):
I don't think I have a good sense of how much performance degradation we'd see before these failed - but it makes sense to me to go for it.

PR #2640, opened 2024-09-16
CLDC-3576 Update soft pregnancy validation

kosiakkatrina:
We only want this soft validation to trigger if all the tenants identify as Male


RachaelBooth (approved):
I see you've raised the question of the other soft validation, which was my next thought too!

PR #2642, opened 2024-09-16
CLDC-3615 Add pending email change banner

kosiakkatrina:
image

### DIFF
--- a/app/helpers/user_helper.rb
+++ b/app/helpers/user_helper.rb
@@ -56,4 +56,23 @@ module UserHelper
       user.errors.add(attribute, message)
     end
   end
+
+  def display_pending_email_change_banner?(user)
+    user.unconfirmed_email.present? && user.email != user.unconfirmed_email
+  end
+
+  def pending_email_change_title_text(current_user, user)
+    if current_user == user
+      "You have requested to change your email address to #{user.unconfirmed_email}."
+    else
+      "There has been a request to change this user’s email address to #{user.unconfirmed_email}."
+    end
+  end
+
+  def pending_email_change_banner_text(current_user)
+    text = "A confirmation link has been sent to the new email address. The current email will continue to work until the change is confirmed."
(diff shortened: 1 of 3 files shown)


app/helpers/user_helper.rb
```diff
@@ -14,4 +14,23 @@ def can_edit_org?(current_user)
...
+    if current_user == user
+      "You have requested to change your email address to #{user.unconfirmed_email}."
+    else
+      "There has been a request to change this user’s email address to #{user.unconfirmed_email}."
+    end
+  end
+
+  def pending_email_change_banner_text(current_user)
+    text = "A confirmation link has been sent to the new email address. The current email will continue to work until the change is confirmed."
+    text += " Deactivating this user will cancel the email change request." if current_user.support?
```
RachaelBooth: Can coordinators deactivate users in their org? If so, they should probably see this too.

spec/helpers/user_helper_spec.rb
```diff
@@ -61,4 +61,65 @@
...
+  describe "display_pending_email_change_banner?" do
+    context "when the user doesn't have an unconfirmed email" do
+      let(:user) { FactoryBot.create(:user, :data_provider, unconfirmed_email: nil) }
+
+      it "does not display pending email change banner" do
+        expect(display_pending_email_change_banner?(user)).to be false
+      end
+    end
+
+    context "when the user doesn't has the same unconfirmed email as current email" do
```
RachaelBooth: You've got a stray "doesn't" in this context name, and in the context below.

app/helpers/user_helper.rb
kosiakkatrina: Yes they can, and I agree, makes sense to show it to coordinators as well!

spec/helpers/user_helper_spec.rb
kosiakkatrina: That's what I get for copying my test cases.. Thank you!

Dinssa (approved):

PR #2643, opened 2024-09-16
CLDC-3627 Problem displaying apostrophes in browser tab title

Dinssa:
There is a problem displaying apostrophes and ampersands in the browser tabs of some pages. This is due to issues with HTML escaping in browser tab titles. We'd like to sanitise these titles and display them correctly without using `html_safe` on user submitted data.

## The problem
Screenshot 2024-09-16 at 16 42 00

## A solution
Screenshot 2024-09-16 at 16 49 37

~~In any string passed through a new `sanitise_characters` method~~

- ~~the apostrophe is replaced with a similar apostrophe~~
- ~~the ampersand is replaced with a similar ampersand character - however, this symbol is slightly different with extra padding compared to the normal ampersand.~~

## Final solution
Editing the browser title method and sanitising title handled most symbols (including apostrophes), ampersands needed handling separately.

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -31,6 +31,12 @@ module ApplicationHelper
     !current_page?(notifications_path) && (authenticated_user_has_notifications? || unauthenticated_user_has_notifications?)
   end
 
+  def sanitise_characters(string)
+    return string unless string
+
+    string.gsub("'", "’").gsub("&", "and")
+  end
+
 private
 
   def paginated_title(title, pagy)
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -2,7 +2,7 @@ module ApplicationHelper
   include Pagy::Frontend
 
   def browser_title(title, pagy, *resources)
-    title = CGI.unescapeHTML(title)
+    title = sanitize(title).gsub("&amp;", "&")
     if resources.any? { |r| r.present? && r.errors.present? }
       "Error: #{[title, t('service_name'), 'GOV.UK'].select(&:present?).join(' - ')}"
     else
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -31,6 +31,7 @@ module ApplicationHelper
   def notifications_to_display?
     !current_page?(notifications_path) && (authenticated_user_has_notifications? || unauthenticated_user_has_notifications?)
   end
+
 private
 
   def paginated_title(title, pagy)
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -34,7 +34,7 @@ module ApplicationHelper
   def sanitise_characters(string)
     return string unless string
 
-    string.gsub("'", "’").gsub("&", "and")
+    string.gsub("'", "’").gsub("&", "＆")
   end
 
 private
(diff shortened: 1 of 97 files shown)


app/helpers/application_helper.rb
```diff
@@ -31,6 +31,12 @@ def notifications_to_display?
     !current_page?(notifications_path) && (authenticated_user_has_notifications? || unauthenticated_user_has_notifications?)
   end
 
+  def sanitise_characters(string)
+    return string unless string
+
+    string.gsub(/[^a-zA-Z0-9\s\-\_\'\&]/, '').tr("'", "’").tr("&", "＆")
+  end
+
```
Dinssa: Helper method that:

- Replace apostrophes with the typographic character `’`.
- Replace ampersands with the full-width character `＆`.
- Remove any characters that are not alphanumeric, spaces, hyphens, underscores, apostrophes, or ampersands.

app/views/organisations/edit.html.erb
```diff
@@ -8,7 +8,7 @@
   <div class="govuk-grid-row">
     <div class="govuk-grid-column-two-thirds">
       <h1 class="govuk-heading-l">
-        <%= content_for(:title) %>
+        <%= "Change #{@organisation.name}’s details" %>
```
Dinssa: We would not want to use text returned by the sanitise_characters method to display on the page as the ampersand is slightly off (with extra padding). It's a compromise for the title but not needed on the page itself.

app/helpers/title_helper.rb
```diff
@@ -4,6 +4,7 @@ def format_label(count, item)
   end
 
   def format_title(searched, page_title, current_user, item_label, count, organisation_name)
+    organisation_name = sanitise_characters(organisation_name)
     if searched.present?
       actual_title = support_sab_nav?(current_user, organisation_name) ? organisation_name : page_title
       "#{actual_title} (#{count} #{item_label} matching ‘#{searched}’)"
```
Dinssa: This `format_title` method is used in a few places, just using the `sanitise_characters` method here is enough to propagate the changes in those other places.

kosiakkatrina: Would using something like this also work? https://apidock.com/rails/ActionView/Helpers/SanitizeHelper/sanitize

Dinssa: No, tried that. Unfortunately not in the title

app/views/organisations/edit.html.erb
kosiakkatrina: Would it make sense to apply the method to all of the tab titles the title tag on `application.html.erb`?
`<title><%= browser_title(yield(:title), @pagy, @admin_user, @user, @organisation, @log, @resource) %></title>`

kosiakkatrina: That way we might avoid having to change the value of the title itself (?)

app/helpers/title_helper.rb
kosiakkatrina: Okay, that is frustrating, I had a bit of a dig and I see what you mean in terms of the "&"
Would having something like this instead of writing our own gsub cover all of the cases?
`sanitize(title.tr("&", "＆"))`

Dinssa: The gsub is not actually needed, I append the .tr("&", "＆") after it which does all the work. I will remove the gsub part in the `sanitise_characters` method, just included it as extra.

`string.gsub(/[^a-zA-Z0-9\s\-_'&]/, "").tr("'", "’").tr("&", "＆")`
to
`string.tr("'", "’").tr("&", "＆")`

kosiakkatrina (approved):

PR #2644, opened 2024-09-17
CLDC-3423: Add sales soft validation for person over retirement age not retired

RachaelBooth:


app/models/form/sales/subsections/household_characteristics.rb
```diff
@@ -14,10 +14,10 @@ def pages
       (Form::Sales::Pages::PrivacyNotice.new("privacy_notice", nil, self, joint_purchase: false) unless form.start_year_after_2024?),
       Form::Sales::Pages::Age1.new(nil, nil, self),
       Form::Sales::Pages::RetirementValueCheck.new("age_1_retirement_value_check", nil, self, person_index: 1),
+      (Form::Sales::Pages::NotRetiredValueCheck.new("age_1_not_retired_value_check", nil, self, person_index: 1) if form.start_year_after_2024?),
```
RachaelBooth, line 17: I thought it would be better not to change the form for older years (people can't submit them but they can still be viewed and I don't want to potentially start routing to a new question)

app/models/form/sales/subsections/household_characteristics.rb
```diff
@@ -14,10 +14,10 @@ def pages
...
       Form::Sales::Pages::OldPersonsSharedOwnershipValueCheck.new("age_1_old_persons_shared_ownership_joint_purchase_value_check", nil, self, joint_purchase: true),
       Form::Sales::Pages::OldPersonsSharedOwnershipValueCheck.new("age_1_old_persons_shared_ownership_value_check", nil, self, joint_purchase: false),
       Form::Sales::Pages::GenderIdentity1.new(nil, nil, self),
-      Form::Sales::Pages::RetirementValueCheck.new("gender_1_retirement_value_check", nil, self, person_index: 1),
```
RachaelBooth, line 20: Retirement age, and thus this validation, no longer depends on gender (and that's set up to be true regardless of form year) so there's no point in this page being here.

app/models/form/sales/pages/not_retired_value_check.rb
```diff
@@ -0,0 +1,25 @@
+class Form::Sales::Pages::NotRetiredValueCheck < Form::Sales::Pages::Person
```
RachaelBooth, line 1: This works exactly like it does for lettings (with slight naming differences because there already were those and they'd be awkward to fix)

kosiakkatrina (approved):
Could any logs go from completed to incomplete?

lib/tasks/recalculate_status_after_sales_over_retirement_age_validation.rake
```diff
@@ -0,0 +1,20 @@
+desc "Recalculates status for 2024 logs that will trigger new sales over retirement age soft validation"
+task recalculate_status_over_retirement: :environment do
+  validation_trigger_condition = "(ecstat1 != 5 AND age1 > 66) OR (ecstat2 != 5 AND age2 > 66) OR (ecstat3 != 5 AND age3 > 66) OR (ecstat4 != 5 AND age4 > 66) OR (ecstat5 != 5 AND age5 > 66) OR (ecstat6 != 5 AND age6 > 66)"
+  SalesLog.filter_by_year(2024).where(status: "pending", status_cache: "completed").where(validation_trigger_condition).find_each do |log|
```
RachaelBooth, line 4: I think we do want to do this for pending logs, because we don't recalculate the status when unpending (we just use the cache). Of course this might also catch a load of logs that will never be unpended, but there's no easy way to avoid that.

kosiakkatrina (approved):

PR #2645, opened 2024-09-17
CLDC-3620 Count duplicate schemes and locations

kosiakkatrina:
We want to get an idea of how many duplicate schemes/locations there are in the service before we add any duplicate schemes/locations features.

### DIFF
--- a/app/models/location.rb
+++ b/app/models/location.rb
@@ -133,7 +133,19 @@ class Location < ApplicationRecord
     scope.pluck("ARRAY_AGG(id)")
   }
 
-  DUPLICATE_LOCATION_ATTRIBUTES = %w[scheme_id postcode mobility_type].freeze
+  scope :duplicate_sets_within_duplicate_schemes, lambda {
+    scope = visible
+    .group(*DUPLICATE_LOCATION_ATTRIBUTES - %w[scheme_id])
+    .where(scheme_id: nil)
+    .where.not(postcode: nil)
+    .where.not(mobility_type: nil)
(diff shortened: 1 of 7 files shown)


app/models/location.rb
```diff
@@ -121,6 +121,19 @@ class Location < ApplicationRecord
...
+    .where.not(scheme_id: nil)
+    .where.not(postcode: nil)
+    .where.not(mobility_type: nil)
+    .having(
+      "COUNT(*) > 1",
+    )
+    scope.pluck("ARRAY_AGG(id)")
+  }
+
+  DUPLICATE_LOCATION_ATTRIBUTES = %w[scheme_id postcode mobility_type].freeze
```
kosiakkatrina, line 147: This assumes that a location can only be a duplicate within the same scheme

Dinssa (approved):

app/models/location.rb
RachaelBooth: If there are duplicate schemes, presumably a location in one could be a duplicate of a location in the other?

kosiakkatrina: Yes! I thought that, but then if a user merges duplicate schemes (which at the moment they'd have to recreate all the locations from 2 duplicate schemes into 1) then those duplicate locations would get flagged.
ACs have this, which I wouldn't expect, cause it suggests we only compare locations of duplicate schemes. I'll double check that!
`Only compare the locations of duplicate schemes to other locations of the same scheme.`

PR #2646, opened 2024-09-17
CLDC-3626 Changes to filtering by specific users

Dinssa:
When filtering by a specific user, I've added modified the visible scope on the user model which returns all users for support and own org and managing org users for non-support (as it did before in the `own_and_managing_org_users` scope).

This branch also adds this scope to one method it was missing from.


app/helpers/filters_helper.rb
```diff
@@ -309,13 +309,21 @@ def formatted_assigned_to_filter(session_filters)
...
   end
 
+  def find_selected_specific_user(session_filters)
+    user_id = session_filters["user"].to_i
+    if current_user.support?
+      User.find(user_id)
+    else
+      User.own_and_managing_org_users(current_user.organisation).find(user_id)
+    end
+  end
```
RachaelBooth: I'm wondering a couple of things (not really introduced by this PR, but relevant to the bug);

1. What set of users do we show people to select from for this filter? I would have expected any related org, not just own or managing orgs, but I might be wrong there. Should make sure that matches what we'll allow here.
2. Would it be sensible to set up a scope on the User model to use for this kind of restriction? (Where that scope was aware of whether the current user was support or not). That might actually be adjusting the current "visible" scope of whether a user a has been deleted, if we thought it made sense to tie that in - which it probably does.

Dinssa: I'll check (1) and (2) is a good idea, will look into it, maybe keeping the existing visible scope and creating a new one "visible_to_user".

Dinssa: (1) We are choosing the users to show people to select from in the same way already, so if support all users, if not, from their own or managing orgs. Just was missing in one place.
(2) I'll create the "visible_to_user" scope and amend all places, while keeping the logic of own or managing orgs. We can adjust if needed.

app/models/user.rb
```diff
@@ -85,7 +85,7 @@ class User < ApplicationRecord
   scope :deactivated, -> { where(active: false) }
   scope :active_status, -> { where(active: true).where.not(last_sign_in_at: nil) }
   scope :visible, -> { where(discarded_at: nil) }
-  scope :own_and_managing_org_users, ->(organisation) { where(organisation: organisation.child_organisations + [organisation]) }
+  scope :visible_to_user, ->(user) { user.support? ? visible : visible.where(organisation: user.organisation.child_organisations + [user.organisation]) }
```
Dinssa: I've used the `visible` scope to filter out discarded users, then applied further filters based on the user's role and organisation.

RachaelBooth: Does this need to be separate from `visible`? (I can't remember exactly where we're using what, but can't think of anywhere we'd need just "not discarded" in a non-support context)

Dinssa: I've added another commit which aims to create a single visible scope. It requires access to the current user logged in to determine if they're a support user or not. In order to do so within the ApplicationController I set the current user in a thread local variable (`current_user`) and added a before action to set the user for each request. Within the User model we add a method `self.current` to grab the `:current_user` from the thread and use it within the scope to determine if the user is a support user or not. Could you have a look and comment on whether this approach is okay?

app/models/user.rb
```diff
@@ -84,8 +84,13 @@ class User < ApplicationRecord
...
   scope :active_status, -> { where(active: true).where.not(last_sign_in_at: nil) }
-  scope :visible, -> { where(discarded_at: nil) }
-  scope :own_and_managing_org_users, ->(organisation) { where(organisation: organisation.child_organisations + [organisation]) }
+  scope :visible, lambda { |user = nil|
+    if user && !user.support?
+      where(discarded_at: nil, organisation: user.organisation.child_organisations + [user.organisation])
+    else
+      where(discarded_at: nil)
+    end
+  }
```
Dinssa, line 93: I've made passing a user parameter optional for the visible scope. Should a user be passed in and they are not a support user their scope will be filtered by organisation otherwise if they're a support user or the visible scope is used without a user passed all non-discarded records will be returned.

app/models/lettings_log.rb
```diff
@@ -132,7 +132,7 @@ class LettingsLog < Log
           illness_type_10: false)
   }
 
-  scope :filter_by_user_text_search, ->(param, user) { where(assigned_to: user.support? ? User.search_by(param) : User.own_and_managing_org_users(user.organisation).search_by(param)) }
+  scope :filter_by_user_text_search, ->(param, user) { where(assigned_to: User.visible(user).search_by(param)) }
```
kosiakkatrina, line 143: I think this would also work:
`scope :filter_by_user_text_search, ->(param, user) { where(assigned_to: visible(user).search_by(param)) }`

kosiakkatrina (approved):

app/models/lettings_log.rb
Dinssa: `visible` is the scope on the user model we want to determine which user's the current user has visibility of. Removing `User.` here would result in it using the `visible` scope within the Log model not the User model. I'm going to leave it unchanged.

PR #2651, opened 2024-09-20
Update postcode stubs

kosiakkatrina:


spec/request_helper.rb
```diff
@@ -16,6 +16,10 @@ def self.stub_http_requests
            .to_return(status: 200, body: "{\"status\":200,\"result\":{\"postcode\":\"ZZ1 1ZZ\",\"admin_district\":\"Westminster\",\"codes\":{\"admin_district\":\"E09000033\"}}}", headers: {})
     WebMock.stub_request(:get, "https://api.postcodes.io/postcodes/SW1A1AA")
            .to_return(status: 200, body: "{\"status\":200,\"result\":{\"postcode\":\"ZZ1 1ZZ\",\"admin_district\":\"Westminster\",\"codes\":{\"admin_district\":\"E09000033\"}}}", headers: {})
+    WebMock.stub_request(:get, "https://api.postcodes.io/postcodes/AB12CD")
```
RachaelBooth: Does it work to have these defined with the postcode service tests, but using webmock and the specific postcodes rather than excon and generic mocks as they were before?

(I know there are other places we define responses needed for specific tests and they don't seem to have caused the same kind of problems)

kosiakkatrina: Definitely, update it now!

RachaelBooth (approved):

PR #2656, opened 2024-09-23
CLDC-3639 Allow some users to update to accounts all roles on staging

kosiakkatrina:
When testing on staging we sometimes need to change our account type to make sure certain changes are only visible to Support users, and not to Data Coordinators/Providers. However, when you change to one of these roles in the user account page it doesn’t let you change it back to Support. 
This change allows users in staging_role_update_email_allowlist to update accounts to any roles.


spec/models/user_spec.rb
```diff
@@ -243,6 +243,87 @@
...
+            expect(user.assignable_roles).to eq({
+              data_provider: 1,
+              data_coordinator: 2,
+              support: 99,
+            })
+          end
+        end
+      end
+
+      context "and the user is in the staging role update email allowlist" do
```
RachaelBooth, line 326: Might be worth testing it doesn't work for prod even when in the allowlist?

RachaelBooth (approved):

PR #2663, opened 2024-09-27
Change user update success banner

kosiakkatrina:
As we have the blue information banner to notify us about email changes, we no longer want an explicit email update success banner, instead we have a generic one for any updates. If no information is update the banner wouldn't show up:
image


spec/requests/users_controller_spec.rb
```diff
@@ -1783,18 +1783,18 @@
           end
 
           it "shows flash notice" do
-            patch("/users/#{other_user.id}", headers:, params:)
+            request
 
-            expect(flash[:notice]).to eq("An email has been sent to #{new_email} to confirm this change.")
+            expect(flash[:notice]).to eq("Your account details have been updated.")
```
RachaelBooth, line 1791: If this is a data provider changing their own email, would they still see something that indicated there'd be a confirmation email, or is that just for support?

kosiakkatrina: All users (including providers) looking at their own accounts should see a version with blue banner and the bit that says a confirmation email has been sent if they have a pending email change

Dinssa (approved):

PR #2664, opened 2024-09-27
Change max width for notification banner

kosiakkatrina:


app/frontend/styles/application.scss
```diff
@@ -95,3 +95,7 @@ $govuk-breakpoints: (
 .grouped-rows .grouped-multirow-cell {
   border-bottom: 1px solid #b1b4b6;
 }
+
+.govuk-notification-banner__content > * {
+  max-width: fit-content;
```
Dinssa, line 100: Do you think it's worth also applying fit-content to one more level .i.e `.govuk-notification-banner__content > * > *`
In some places there is a heading (`govuk-notification-banner__heading`) within `govuk-notification-banner__content` which does not fit-content.

Dinssa (approved):

PR #2666, opened 2024-10-01
CLDC-3618: Support user - View bulk uploads page

Dinssa:
This branch aims to give support users the ability to view bulk uploads, access the bulk upload error report and download the associated file to help better diagnose any issues.

## Accessing the view bulk uploads pages
This page is only available to support users and can be accessed in two places.

We add a new button "View {lettings/sales} bulk uploads" to the main lettings and sales logs index page.
<br>
Screenshot 2024-10-02 at 08 42 52  
<br>

We also add the new button to the organisation page. When this is clicked it will apply a filter showing the specific organisation's bulk uploads.
<br>
Screenshot 2024-10-02 at 08 43 05
<br>

## View bulk uploads page
Screenshot 2024-10-02 at 08 48 46

### Features:

- Bulk Upload Details: Displays each bulk upload with status and failure reasons.
- Bulk Upload File: Allows downloading files and viewing error reports.
- Filtering Options: Filter by collection year, uploader, and organisation.
- Search Functionality: Search by file name, user’s name, email, or organisation.

## Error report 
Within the error report a support user will also see a new link to download the bulk upload file. <br>
Screenshot 2024-10-02 at 09 32 05


app/components/bulk_upload_summary_component.html.erb
```diff
@@ -0,0 +1,37 @@
+<article class="app-log-summary">
+  <div class="govuk-grid-row">
+    <div class="govuk-grid-column-two-thirds">
+      <header class="app-log-summary__header">
+        <h2 class="app-log-summary__title govuk-heading-m govuk-!-font-weight-regular" style="white-space: normal; word-break: break-all">
```
kosiakkatrina: Would it make sense to extract the styling into a class?

app/components/bulk_upload_summary_component.rb
```diff
@@ -0,0 +1,72 @@
...
+  def upload_status
+    helpers.status_tag(bulk_upload.status, ["app-tag--small govuk-!-font-weight-regular no-max-width"])
+  end
+
+  def setup_errors_count
+    @bulk_upload_errors.where(category: "setup").count
+  end
+
+  def critical_errors_count
+    @bulk_upload_errors.where(category: [nil, "", "not_answered"]).count
```
kosiakkatrina, line 19: would setup errors be critical too?

kosiakkatrina: Also, what if we decide to add more categories?

app/components/bulk_upload_summary_component.rb
```diff
@@ -0,0 +1,72 @@
...
+      formatted_count_text(count, singular_text, plural_text) if count.present?
+    }.compact.join("").html_safe
+  end
+
+  def download_file_link(bulk_upload)
+    send("download_#{bulk_upload.log_type}_file_link", bulk_upload)
+  end
+
+  def download_lettings_file_link(bulk_upload)
+    link_to "Download file", download_lettings_bulk_upload_path(bulk_upload), class: "govuk-link govuk-!-margin-right-2"
```
kosiakkatrina: any reason not to use `govuk_link_to` here?

app/components/create_log_actions_component.rb
```diff
@@ -53,4 +53,22 @@ def upload_button_href
       bulk_upload_sales_log_path(id: "start")
     end
   end
+
+  def view_uploads_button_copy
+    case log_type
+    when "lettings"
+      "View lettings bulk uploads"
```
kosiakkatrina: Would it make sense to interpolate log_type here?

app/components/create_log_actions_component.html.erb
```diff
@@ -6,5 +6,8 @@
     <% if upload_button_href.present? && !user.support? %>
       <%= govuk_button_link_to upload_button_copy, upload_button_href, secondary: true %>
     <% end %>
+    <% if view_uploads_button_href.present? && user.support? %>
```
kosiakkatrina: What would be scenario where `view_uploads_button_href` doesn't exist?

app/components/search_component.rb
```diff
@@ -19,6 +19,10 @@ def path(current_user)
       request.path
     elsif request.path.include?("organisations") && request.path.include?("managing-agents")
       request.path
+    elsif request.path.include?("bulk-uploads") && request.path.include?("sales-logs")
+      request.path
+    elsif request.path.include?("bulk-uploads") && request.path.include?("lettings-logs")
+      request.path
     elsif request.path.include?("users")
       user_path(current_user)
     elsif request.path.include?("organisations")
```
kosiakkatrina, line 16: Can we refactor this method?
Maybe something like
```
def path(current_user)
  return request.path if matching_path_conditions?
  ....
end
```

app/components/bulk_upload_summary_component.rb
Dinssa: Aren't setup errors treated as important errors? If we'd add more categories, we'd have to add them here. Unless we can say that "setup" and "soft_validation" are the only error types that wouldn't be critical errors so this count should count everything else?

app/components/bulk_upload_summary_component.rb
Dinssa: I'll change it

app/components/bulk_upload_summary_component.rb
kosiakkatrina: I'd treat everything other than soft-validation as critical. Do setup errors not get displayed in the critical errors section in the report?

app/controllers/lettings_logs_controller.rb
```diff
@@ -115,14 +115,50 @@ def update_logs
...
+    downloader = BulkUpload::Downloader.new(bulk_upload:)
+
+    if Rails.env.development?
+      downloader.call
+      send_file downloader.path, filename: bulk_upload.filename, type: "text/csv"
+    else
+      presigned_url = downloader.presigned_url
+      redirect_to presigned_url, allow_other_host: true
+    end
+  end
```
kosiakkatrina: What if I go to `https://review.submit-social-housing-data.communities.gov.uk/2666/lettings-logs/bulk-uploads/{id}/download` while signed in as non-support user?

app/components/create_log_actions_component.html.erb
Dinssa: None really, this component is only used on logs/index and the log_type is set from the controller - which it should always be one of lettings or sales. The methods now build a route based of that log_type, which should also always exist. I think I can refactor this component and remove the .present? checks

app/views/bulk_upload_shared/uploads.html.erb
```diff
@@ -0,0 +1,28 @@
+<% item_label = format_label(@pagy.count, "uploads") %>
+<% title = format_title(@searched, "Lettings bulk uploads", current_user, item_label, @pagy.count, nil) %>
```
kosiakkatrina: Should this sometimes be sales?

app/helpers/bulk_upload_helper.rb
```diff
@@ -0,0 +1,12 @@
+module BulkUploadHelper
+  def bulk_upload_title(controller)
```
kosiakkatrina: Would it make sense to only pass controller_name here instead of the entire class?

app/helpers/filters_helper.rb
```diff
@@ -14,6 +14,12 @@ def filter_selected?(filter, value, filter_type)
     return true if (selected_filters["owning_organisation"].present? || selected_filters["owning_organisation_text_search"].present?) && filter == "owning_organisation_select" && value == :specific_org
     return true if (selected_filters["managing_organisation"].present? || selected_filters["managing_organisation_text_search"].present?) && filter == "managing_organisation_select" && value == :specific_org
 
+    return true if !selected_filters.key?("user") && filter == "uploaded_by" && value == :all
+    return true if selected_filters["uploaded_by"] == "specific_user" && filter == "uploaded_by" && value == :specific_user
```
kosiakkatrina: Would it be at all nicer to refactor this method a bit?
How would it read if we were to extract `filter == "uploaded_by"? Should this entire method be a case statement based on filter?

app/controllers/lettings_logs_controller.rb
Dinssa: Nice catch, that is possible, I will prevent it.

app/helpers/filters_helper.rb
```diff
@@ -274,7 +285,7 @@ def filters_count(filters)
     filters.each.sum do |category, category_filters|
       if %w[years status needstypes bulk_upload_id].include?(category)
         category_filters.count(&:present?)
-      elsif %w[user owning_organisation managing_organisation user_text_search owning_organisation_text_search managing_organisation_text_search].include?(category)
+      elsif %w[user owning_organisation managing_organisation user_text_search owning_organisation_text_search managing_organisation_text_search uploading_organisation].include?(category)
```
kosiakkatrina, line 281: Do we need uploading_organisation_text_search here?

app/views/bulk_upload_shared/uploads.html.erb
Dinssa: Yes nice, will change this

app/models/bulk_upload.rb
```diff
@@ -11,11 +12,56 @@ class BulkUpload < ApplicationRecord
 
   after_initialize :generate_identifier, unless: :identifier
 
+  scope :search_by_filename, ->(filename) { where("filename ILIKE ?", "%#{filename}%") }
+  scope :search_by_user_name, ->(name) { where(user_id: User.where("name ILIKE ?", "%#{name}%").select(:id)) }
+  scope :search_by_user_email, ->(email) { where(user_id: User.where("email ILIKE ?", "%#{email}%").select(:id)) }
+  scope :search_by_organisation_name, ->(name) { where(organisation_id: Organisation.where("name ILIKE ?", "%#{name}%").select(:id)) }
```
kosiakkatrina, line 19: I think joins(:organisation)/(:user) might be more performant than a subquery

app/models/bulk_upload.rb
```diff
@@ -11,11 +12,56 @@ class BulkUpload < ApplicationRecord
...
+      .or(search_by_organisation_name(param))
+  }
+
+  scope :filter_by_id, ->(id) { where(id:) }
+  scope :filter_by_years, lambda { |years, _user = nil|
+    first_year = years.shift
+    query = where(year: first_year)
+    years.each { |year| query = query.or(where(year:)) }
+    query.all
+  }
```
kosiakkatrina: would `where(year: years)` not be enough?

kosiakkatrina: assuming years is an array

app/models/bulk_upload.rb
```diff
@@ -11,11 +12,56 @@ class BulkUpload < ApplicationRecord
...
+    return :processing if processing
+    return :blank_template if failed == "blank_template"
+    return :wrong_template if failed == "wrong_template"
+
+    if logs.visible.exists?
+      return :errors_fixed_in_service if completed? && bulk_upload_errors.any?
+      return :logs_uploaded_with_errors if bulk_upload_errors.any?
+    end
+
+    if bulk_upload_errors.any? { |error| error.category == "setup" }
```
kosiakkatrina: would it make sense for these to be scopes? `bulk_upload_errors.setup.exist?`

app/models/bulk_upload.rb
Dinssa: If I remember right, joining organisation brought up errors wanting to join organisation also on the other search scopes. This was a way to avoid doing that.

app/components/bulk_upload_summary_component.rb
kosiakkatrina: If you create this as scope if might be a bit more reusable, I left another comment about scopes here https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/2666/files#r1784320513

app/models/bulk_upload.rb
```diff
@@ -105,6 +151,10 @@ def moved_user_name
     User.find_by(id: moved_user_id)&.name
   end
 
+  def organisation
+    Organisation.find_by(id: organisation_id)
+  end
+
```
kosiakkatrina, line 153: would `belongs_to :organisation` work?

app/models/user.rb
```diff
@@ -219,7 +219,7 @@ def assignable_roles
 
   def logs_filters(specific_org: false)
```
kosiakkatrina, line 225: I'd probably have a separate `bulk_uploads_filters` method, cause the naming of this one is a bit misleading. Also it feels like the list of filters for bu wouldn't necessarily depend on the user type so I'm not sure if it's the best place for this to be. 
I think this is wrong for schemes as well, so wasn't introduced by you. 
It's up to you if you want to update it

app/services/bulk_upload/processor.rb
```diff
@@ -144,4 +146,13 @@ def validator_class
       raise "Validator not found for #{bulk_upload.log_type}"
     end
   end
+
+  def handle_invalid_validator
+    if validator.errors.full_messages.include?("Template is blank - The template must be filled in for us to create the logs and check if data is correct.")
```
kosiakkatrina: Would it make sense to extract the error messages into a translation file? In case we change them!

app/models/bulk_upload.rb
```diff
@@ -1,6 +1,7 @@
 class BulkUpload < ApplicationRecord
   enum log_type: { lettings: "lettings", sales: "sales" }
   enum rent_type_fix_status: { not_applied: "not_applied", applied: "applied", not_needed: "not_needed" }
+  enum failed: { blank_template: 1, wrong_template: 2 }
```
kosiakkatrina: Is there any benefit to having integers? Might be a bit more readable if we have something similar to log_type instead

app/services/filter_manager.rb
```diff
@@ -107,7 +121,7 @@ def deserialize_filters_from_session(specific_org)
 
       new_filters = new_filters.except("user") if params["assigned_to"] == "all"
       new_filters["user"] = current_user.id.to_s if params["assigned_to"] == "you"
-      new_filters = new_filters.except("user_text_search") if params["assigned_to"] == "all" || params["assigned_to"] == "you"
+      new_filters = new_filters.except("user_text_search") if params["assigned_to"] == "all" || params["assigned_to"] == "you" # add later
```
kosiakkatrina: Do we still need this comment?

app/services/filter_manager.rb
```diff
@@ -124,6 +138,16 @@ def deserialize_filters_from_session(specific_org)
...
+    if filter_type.include?("bulk_uploads")
+      current_user.logs_filters(specific_org:).each do |filter|
+        new_filters[filter] = params[filter] if params[filter].present?
+      end
+      # binding.pry
+      new_filters = new_filters.except("uploading_organisation") if params["uploading_organisation_select"] == "all"
+      new_filters = new_filters.except("user") if params["uploaded_by"] == "all"
+      new_filters["user"] = current_user.id.to_s if params["uploaded_by"] == "you"
+    end
+    # binding.pry
```
kosiakkatrina: a couple of bindings in this method

app/services/storage/s3_service.rb
```diff
@@ -20,10 +20,10 @@ def folder_present?(folder)
       response.key_count == 1
     end
 
-    def get_presigned_url(file_name, duration)
+    def get_presigned_url(file_name, duration, response_content_disposition: nil)
       Aws::S3::Presigner
         .new({ client: @client })
-        .presigned_url(:get_object, bucket: @configuration.bucket_name, key: file_name, expires_in: duration)
+        .presigned_url(:get_object, bucket: @configuration.bucket_name, key: file_name, expires_in: duration, response_content_disposition:)
```
kosiakkatrina, line 26: if response_content_disposition is nil do we override a default?

app/services/bulk_upload/processor.rb
```diff
@@ -6,12 +6,13 @@ def initialize(bulk_upload:)
   end
 
   def call
+    bulk_upload.update!(processing: true)
```
kosiakkatrina: I think we either want to set processing to true as soon as we create the bulk upload, or we want another status, something like `scheduled` etc., because depending on how available the queue is it won't immediately start processing the file, and we'd get `Logs uploaded with no errors` status for those bulk uploads. 

You can check this if you try uploading multiple files right after each other

db/migrate/20240927143916_add_processing_to_bulk_uploads.rb
```diff
@@ -0,0 +1,5 @@
+class AddProcessingToBulkUploads < ActiveRecord::Migration[7.0]
```
kosiakkatrina: Could we put all migrations into a single migration file?

db/migrate/20240925095041_add_failed_to_bulk_uploads.rb
```diff
@@ -0,0 +1,5 @@
+class AddFailedToBulkUploads < ActiveRecord::Migration[7.0]
+  def change
+    add_column :bulk_uploads, :failed, :integer
```
kosiakkatrina: failed implies bool to me, would failure_reason (or similar) make more sense?

app/services/filter_manager.rb
Dinssa: Will remove

app/services/filter_manager.rb
Dinssa: Will remove

app/services/storage/s3_service.rb
Dinssa: The issue stemmed from when downloading a bulk upload it would've come up as an blank file (with the right name, just empty contents) without specifying response_content_disposition, I did set the method to initialise with nil unless specified. This did not affect downloading other CSVs elsewhere in the app. This could be done better though, so will look at it again.

app/services/bulk_upload/processor.rb
Dinssa: Ah okay, that's a good idea, I was unaware of the queue, will set it to true when the bulk upload is created.

db/migrate/20240925095041_add_failed_to_bulk_uploads.rb
Dinssa: Sure I can see that, failure_reason sounds fine

app/services/bulk_upload/processor.rb
Dinssa: Sure, I can find out where they're set in the validator and use a translation here and there.

app/models/user.rb
Dinssa: It would be a good idea, I'll look into it.

app/models/bulk_upload.rb
Dinssa: I do not want to make a bulk upload belong to an organisation, that caused a few issues when setting the organisation_id during another ticket to allow support users to upload on behalf of others. I could set it correctly the first time the bulk upload was created, but once it was associated with an organisation all subsequent calls for organisation_id would return that of the uploader (support user) not the organisation it was target at. It would save the uploader's organisation_id instead of the value passed. Maybe that is something that could/should be resolved. 

In essence we need to make the bulk upload belong to an organisation different to that of the user the bulk upload belongs to.

app/models/bulk_upload.rb
Dinssa: Ye good idea

app/components/bulk_upload_summary_component.rb
Dinssa: We count setup errors as "errors on important questions" separately from the count for "critical errors". But they do appear within the critical errors section of the error report, but on their own in the summary. It would be double counting to show the setup errors in the counts for "errors on important questions" and "critical errors".

app/helpers/filters_helper.rb
```diff
@@ -346,4 +343,39 @@ def formatted_managed_by_filter(session_filters, filter_type)
   def unanswered_filter_value
     "<span class=\"app-!-colour-muted\">You didn’t answer this filter</span>".html_safe
   end
+
+  def assigned_to_filter_selected?(selected_filters, value)
+    return true if !selected_filters.key?("user") && value == :all
+    return true if selected_filters["assigned_to"] == value.to_s
+
+    false
+  end
```
kosiakkatrina: ?
```suggestion
  def assigned_to_filter_selected?(selected_filters, value)
    return true if !selected_filters.key?("user") && value == :all
    selected_filters["assigned_to"] == value.to_s
  end
```

kosiakkatrina: And similarly for the remaining methods

app/services/bulk_upload/processor.rb
```diff
@@ -1,12 +1,27 @@
...
+  WRONG_TEMPLATE_ERRORS = [
+    I18n.t("activemodel.errors.models.bulk_upload/lettings/validator.attributes.base.wrong_field_numbers_count"),
+    I18n.t("activemodel.errors.models.bulk_upload/lettings/validator.attributes.base.over_max_column_count"),
+    I18n.t("activemodel.errors.models.bulk_upload/lettings/validator.attributes.base.wrong_template"),
+    I18n.t("activemodel.errors.models.bulk_upload/lettings/validator.attributes.base.no_headers"),
+    I18n.t("activemodel.errors.models.bulk_upload/sales/validator.attributes.base.wrong_field_numbers_count"),
+    I18n.t("activemodel.errors.models.bulk_upload/sales/validator.attributes.base.over_max_column_count"),
+    I18n.t("activemodel.errors.models.bulk_upload/sales/validator.attributes.base.wrong_template"),
+    I18n.t("activemodel.errors.models.bulk_upload/sales/validator.attributes.base.no_headers"),
+  ].freeze
```
kosiakkatrina: Would it make sense to put all the wrong template errors under the same key and then get them all instead of listing them out?
From quick google something like this would let you access them all, but maybe there's a nicer solution:
`I18n.backend.send(:translations)[:en][:activemodel][:errors][:models]...`

spec/components/create_log_actions_component_spec.rb
```diff
@@ -37,13 +37,23 @@
         expect(component.create_button_href).to eq("/lettings-logs")
       end
 
-      it "returns upload button copy" do
-        expect(component.upload_button_copy).to eq("Upload lettings logs in bulk")
+      it "does not show the upload button" do
```
kosiakkatrina, line 40: Why did this change?

spec/features/organisation_spec.rb
```diff
@@ -136,7 +136,7 @@
       end
 
       it "shows a create button for that organisation" do
-        expect(page).to have_button("Create a new lettings log for this organisation")
```
kosiakkatrina, line 142: How come this is changing?

Dinssa: We introduced a new button to "View {lettings/sales} bulk uploads" within the organisation page. This button's text was shortened in the designs to comfortably fit in the new button.

spec/components/create_log_actions_component_spec.rb
Dinssa: We don't show the upload lettings logs in bulk button on that page for support users. I changed it within a previous ticket #2629, just added a test for that. We don't need to return the upload button copy or upload button href on this page, for support users. We do for other users.

app/helpers/filters_helper.rb
Dinssa: Sure that works

kosiakkatrina (approved):

PR #2671, opened 2024-10-03
CLDC-3517 Download collection resources from S3

kosiakkatrina:


Dinssa (approved):

config/environments/production.rb
```diff
@@ -147,4 +147,5 @@
       end
     }
   end
+  config.collection_resources_s3_bucket_name = "core-prod-collection-resources"
```
RachaelBooth: This (for each env) seems like it should be using an env variable, so that the terraform can just set that up without us needing to keep things in sync if they change for any reason/one of the bucket names can't follow the pattern.

kosiakkatrina: Of course!

PR #2672, opened 2024-10-03
CLDC-3650: Move copy for sales setup section to translation files

RachaelBooth:


app/models/form/page.rb
```diff
@@ -24,6 +24,12 @@ def initialize(id, hsh, subsection)
 
   delegate :form, to: :subsection
 
+  def header
+    return @header if @copy_key.nil?
+
+    I18n.t("forms.#{form.start_date.year}.#{@copy_key}.page_header")
+  end
```
kosiakkatrina: Would it make sense to set the `@header` to this to begin with? Would that make it a bit more performant?

RachaelBooth: I think I can do that - there's an ordering issue then in the definitions that will give an annoying error if someone gets it wrong, but we can probably cope with that.

kosiakkatrina (approved):

PR #2674, opened 2024-10-04
CLDC-3633++ Full stops in error messages & capitalisation within messages

Dinssa:
This branch aims to add full stops to a few additional error messages hardcoded in the row parser. It introduced three new helper methods.

1. `ensure_sentence_ending`
This method will check for a given value if there already is punctuation using the regular expression character class [[:punct:]] which includes ! " # $ % & ' ( ) * + , - . / : ; < = > ? @ [ \ ] ^ _ ` { | } ~ . It does however, exclude brackets and percentages:  ( )  { }  [ ]  % 
The idea is that we should still add a full stop after the brackets and percentages but not other sentence ending punctuation (at the moment only question marks).

2. `downcase_first_letter`
This method will downcase only the first character of a sentence and not the entire sentence as the default .downcase method does.

3. `format_ending`
This method uses both of the above.

### Example of changes:
Before: You must answer scheme name
Now: You must answer scheme name.
Using only `format_ending`

Before we had error messages where they were compounded together and the second half was not downcased such as:
"Enter a valid value for **W**as a mortgage used for the purchase of this property? - Shared ownership”

The `ensure_sentence_ending` method will add a full stop here resulting in:
"Enter a valid value for Was a mortgage used for the purchase of this property? - Shared ownership.”
This is not needed for this these types of questions where the main question is punctuated but is not a problem either. 

What the `downcase_first_letter` method does is take what the default downcase would have done:
"Enter a valid value for was a mortgage used for the purchase of this property? - **s**hared ownership.” 
(also downcasing "Shared" later in the message) and result in:
"Enter a valid value for was a mortgage used for the purchase of this property? - **S**hared ownership.”
Maintaining the capitalisation of all other words within the sentence.


config/locales/en.yml
```diff
@@ -443,7 +443,7 @@ en:
           above_hard_max: "Rent is higher than the absolute maximum expected for a property of this type based on this period."
       charges:
         complete_1_of_3: "Answer either the ‘household rent and charges’ question or ‘is this accommodation a care home‘, or select ‘no’ for ‘does the household pay rent or charges for the accommodation?’"
-        missing_charges: "Please enter the %{question}. If there is no %{question}, please enter '0'."
+        missing_charges: "Please enter the %{sentence_fragment}. If there is no %{sentence_fragment}, please enter '0'."
```
Dinssa, line 449: The idea behind this is to always assume if the expected variable is "question" it will include it's own punctuation. And for any other inputs like this were they are put in a pre formatted sentence to call them "sentence_fragment".

Dinssa:
There are examples of where we used the QUESTIONS constant at the top of a row_parser like below where I was unable to apply the `downcase_first_letter` method and left the default `.downcase` method. Only within validates methods.

 ```
 validates :field_9,
            inclusion: {
              in: [2, 30, 18, 16, 24, 28, 31, 32],
              if: proc { field_9.present? },
              category: :setup,
              question: QUESTIONS[:field_9].downcase,
            },
            on: :before_log
```

File: app/services/bulk_upload/sales/year2023/row_parser.rb [hunk 2]
```diff
@@ -1334,4 +1335,8 @@ def validate_buyer1_economic_status
       errors.add(:field_35, "Buyer 1 cannot be a child under 16.")
     end
   end
+
+  def downcase(str)
+    downcase_first_letter(str)
+  end
```
Dinssa, line 1341: Add a more concisely named private method to each parser, using the downcase_first_letter method from the helper.

app/services/bulk_upload/lettings/year2023/row_parser.rb
```diff
@@ -1606,4 +1607,8 @@ def first_time_property_let_as_social_housing
   def is_carehome
     field_127.present? ? 1 : 0
   end
+
+  def downcase(sentence)
+    downcase_first_letter(sentence)
+  end
```
Dinssa: Add a more concisely named private method to each row parser, using the downcase_first_letter method from the helper.

Dinssa: Renamed the parameter to text from sentence

app/helpers/formatting_helper.rb
```diff
@@ -0,0 +1,16 @@
+module FormattingHelper
+  def format_ending(value)
+    return value if value.blank?
+
+    ends_with_any_punctuation = value.match?(/[[:punct:]]\z/)
+    ends_with_special_char = value.match?(/[%(){}\[\]]\z/)
+
+    ends_with_any_punctuation && !ends_with_special_char ? value : "#{value}."
+  end
```
Dinssa: Change this method from this `value.match?(/[[:punct:]]\z/) && !value.match?(/[%(){}\[\]]\z/) ? value : "#{value}."`  to try improve readability. Keep changed or revert back? Do the variable names here best explain what is going on? (especially `ends_with_special_char`)

app/services/bulk_upload/sales/year2023/row_parser.rb
```diff
@@ -1270,13 +1271,13 @@ def validate_valid_radio_option
...
             block_log_creation!
-            errors.add(field, I18n.t("validations.invalid_option", question: QUESTIONS[field]), category: :setup)
+            errors.add(field, I18n.t("validations.invalid_option", question: format_ending(downcase(QUESTIONS[field]))), category: :setup)
           end
         end
       else
         fields.each do |field|
           unless errors.any? { |e| fields.include?(e.attribute) }
-            errors.add(field, I18n.t("validations.invalid_option", question: QUESTIONS[field]))
+            errors.add(field, I18n.t("validations.invalid_option", question: format_ending(downcase(QUESTIONS[field]))))
```
kosiakkatrina: Do we ever use format_ending without downcase?
Would it make sense to have 1 method that does both if not?

app/models/form/question.rb
```diff
@@ -204,11 +206,12 @@ def value_is_refused?(value)
   end
 
   def error_display_label
-    error_label || check_answer_label || header || id.humanize
+    label = error_label || check_answer_label || header || id.humanize
+    format_ending(label)
```
kosiakkatrina, line 210: Do we ever want this to not be downcased?

Dinssa: I've changed it to now use the downcase_first_letter method (renamed to lowercase_first_letter) within the format_ending method.

kosiakkatrina (approved):

PR #2677, opened 2024-10-07
CLDC-3653 Copy changes - sales setup validations

Dinssa:


kosiakkatrina (approved):

config/locales/validations/sales/setup/en.yml
```diff
@@ -0,0 +1,25 @@
...
+                        absorbing_organisations:
+                            "Enter a date when the owning organisation was active. %{owning_organisation} became active on %{available_from}."
+
+                owning_organisation:
+                    inactive:
+                        merged_organisation:
+                            "The owning organisation must be active on the sale completion date. %{owning_organisation} became inactive on %{merge_date} and was replaced by %{absorbing_organisation}."
+                        absorbing_organisation:
+                            "The owning organisation must be active on the sale completion date. %{owning_organisation} became active on %{available_from}."
+
```
kosiakkatrina: would it make sense for this file to be setup.en.yml instead of being in a setup folder?

kosiakkatrina: Also, the formatting looks a bit off (?) is it 2 tabs for each nesting? Could you check other yml files to make sure the spacing is correct?

Dinssa: Yes will change that

PR #2678, opened 2024-10-07
CLDC-3518 Add file validation to collection resources

kosiakkatrina:


RachaelBooth (approved):

app/controllers/collection_resources_controller.rb
```diff
@@ -95,4 +99,23 @@ def resource_for_year_can_be_downloaded?(year)
   def resource_for_year_can_be_updated?(year)
     editable_collection_resource_years.include?(year)
   end
+
+  def validate_file(file)
+    return @collection_resource.errors.add(:file, "Select which file to upload") unless file
+    return @collection_resource.errors.add(:file, "The file is above 100MB") if file.size > 100.megabytes
```
Dinssa: Should we also add full stops to these error messages?

Dinssa (approved):

app/controllers/collection_resources_controller.rb
Dinssa: Also, should we be putting them in the translation file?

kosiakkatrina: Yep!

PR #2680, opened 2024-10-08
CLDC-2740: Sort rent periods in organisation details page

Dinssa:


spec/models/organisation_spec.rb
```diff
@@ -111,45 +111,52 @@
...
+        [4, 2, 1].each do |rent_period|
+          create(:organisation_rent_period, organisation:, rent_period:)
+        end
         allow(RentPeriod).to receive(:rent_period_mappings).and_return(fake_rent_periods)
       end
 
       context "when the org does not use all rent periods" do
         it "#rent_periods returns the correct ids" do
-          expect(organisation.rent_periods).to eq [1]
+          expect(organisation.rent_periods).to eq [4, 2, 1]
```
RachaelBooth: Is the order guaranteed when returning the ids? It seems like this could fail with re-ordering but we wouldn't actually care for this test, so could maybe test in a way that doesn't care about the order?

Dinssa: I'd imagine the order is, as they're created in that order. However, as this test is not to check the order but to check the presence of the correct ids only I've updated two tests to make sure the array contains the values 4,2,1 not for the array to match [4, 2, 1].

RachaelBooth (approved):

PR #2681, opened 2024-10-08
CLDC-3662: Recalculate status for 2024 logs missing la

RachaelBooth:


lib/tasks/recalculate_status_when_la_missing.rake
```diff
@@ -17,3 +17,22 @@ task recalculate_status_missing_la: :environment do
     end
   end
 end
+
+desc "Recalculates status for 2024 completed logs with missing LA"
+task recalculate_status_missing_la_2024: :environment do
+  LettingsLog.filter_by_year(2024).where(needstype: 1, la: nil, status: "completed").find_each do |log|
```
RachaelBooth, line 23: This can happen if a log was marked completed before the fix here: https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/2449 with la not routed to, and the result of `address_search_given?` changed because `postcode_full` was not present but `postcode_full_input` was. 

I'm checking, but I don't think we want to infer postcodes from the address search postcodes - they don't have the same validation, and we don't necessarily know why users went to manual entry.

kosiakkatrina (approved):

PR #2685, opened 2024-10-09
CLDC-3015: Remove duplicate rent periods

Dinssa:
This branch adds a rake task to check and view duplicate rake tasks. Once confirmed, you can use the task to delete the duplicates.


lib/tasks/duplicate_rent_periods.rake
```diff
@@ -0,0 +1,52 @@
...
+  result = process_duplicate_rent_periods(log_groups: true)
+
+  Rails.logger.info "Total number of records: #{OrganisationRentPeriod.count}"
+  Rails.logger.info "Number of duplicate records: #{result[:duplicate_records].size}"
+  Rails.logger.info "Number of records to delete: #{result[:redundant_ids].size}"
+  Rails.logger.info "Number of records to keep: #{result[:ids_to_keep].size}"
+end
+
+desc "Delete redundant rent periods"
+task delete_duplicate_rent_periods: :environment do
```
RachaelBooth, line 46: Is it worth some unit tests for this? I realise we'll only run it once, but the logic looks sufficiently complicated to want something.

Dinssa: Sure

spec/lib/tasks/dupicate_rent_periods_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+
+  describe "delete_duplicate_rent_periods" do
+    let(:task) { Rake::Task["delete_duplicate_rent_periods"] }
+
+    before do
+      task.reenable
+    end
+
+    it "deletes redundant rent periods" do
+      expect { task.invoke }.to change(OrganisationRentPeriod, :count).by(-4)
```
RachaelBooth: I'd suggest checking that the remaining records are the expected ones as well, rather than just the number.

spec/lib/tasks/dupicate_rent_periods_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+require "rake"
+
+RSpec.describe "duplicate_rent_periods" do
+  before do
+    Rake.application.rake_require("tasks/duplicate_rent_periods")
+    Rake::Task.define_task(:environment)
+    allow(Rails.logger).to receive(:info)
+    organisation = create(:organisation, rent_periods: (1..11).to_a)
+
+    [2, 4, 6, 11].each do |rent_period|
```
RachaelBooth: I don't know if it shows up on live, but to me it seems worth testing with at least one group of more than 2 duplicates.

spec/lib/tasks/dupicate_rent_periods_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+      org_rent_period = build(:organisation_rent_period, organisation:)
+      org_rent_period.save!(validate: false)
+      org_rent_period.update_column(:rent_period, rent_period)
```
RachaelBooth: Could you set the rent period in the initial construction (or otherwise before the save) to avoid needing two db updates? (That will make a difference to how long the tests take to run - not huge for just these but as a general principle we want to keep them as fast as we can)

Dinssa: Setting the rent_periods directly in the constructor for the organisation variable won't work because duplicates won't be saved that way, even with validate:false. However, I can set the rent_periods before saving and will change that.

The following approaches won't work:
`organisation = create(:organisation, rent_periods: (1..11).to_a.concat([2, 4, 6, 11, 2, 11]))`
or
`organisation = create(:organisation, rent_periods: [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 2, 4, 6, 11, 2, 11])`

RachaelBooth (approved):

PR #2687, opened 2024-10-10
CLDC-3063: Sort alphabetically stock owners and managing agents lists

Dinssa:


app/controllers/organisation_relationships_controller.rb
```diff
@@ -14,7 +14,7 @@ class OrganisationRelationshipsController < ApplicationController
   ]
 
   def stock_owners
-    stock_owners = organisation.stock_owners.filter_by_active
+    stock_owners = organisation.stock_owners.filter_by_active.order('LOWER(name)')
```
Dinssa: With just .order the sort was case-sensitive and we get incorrect ordering e.g. MHCLG before Managing Agent 1. Using LOWER(name) ensures the sorting of names is case-insensitive.

app/controllers/organisation_relationships_controller.rb
```diff
@@ -14,7 +14,7 @@ class OrganisationRelationshipsController < ApplicationController
   ]
 
   def stock_owners
-    stock_owners = organisation.stock_owners.filter_by_active
+    stock_owners = organisation.stock_owners.filter_by_active.order("LOWER(name)")
```
RachaelBooth, line 17: Can there be some tests for this sorting?

Dinssa: Sure will add some

RachaelBooth (approved):

PR #2690, opened 2024-10-11
CLDC-3638 Allow adding custom resources

kosiakkatrina:
Allow adding custom resources, run the same file validations for size and presence, add a resource type and validate that it's present. Also allow updating the resource, the flow for which would work the same as updating mandatory resources
image


app/controllers/collection_resources_controller.rb
```diff
@@ -23,7 +25,16 @@ def download_mandatory_collection_resource
     download_resource(resource.download_filename)
   end
 
-  def edit
+  def download_additional_collection_resource
```
RachaelBooth, line 28: Does this also need an exception from the `before_action :authenticate_user!`?

app/views/collection_resources/new.html.erb
```diff
@@ -0,0 +1,34 @@
...
+      </p>
+
+      <%= f.govuk_file_field :file,
+        label: { text: "Upload file", size: "m" } %>
+
+      <%= f.govuk_text_field :short_display_name,
+                      label: nil,
+                      hint: { text: safe_join(["This will be used in the download link on the homepage. Do not include the log type or collection year.",
+                                               content_tag(:br),
+                                               "For example, if you enter “bulk upload change log”, the download link will say “Download the #{@collection_resource.log_type} bulk upload change log (#{text_year_range_format(@collection_resource.year)})”."]) } %>
```
Dinssa, line 28: Could you add one more `<br>`? On the edit page too. Separate out the sentences more clearly, similar to the designs.

RachaelBooth (approved):

app/views/collection_resources/new.html.erb
kosiakkatrina: This is what it looks like atm, the description screenshot it a bit outdated. What do you think?
image

app/controllers/collection_resources_controller.rb
kosiakkatrina: Yup! Good spot!

app/views/collection_resources/new.html.erb
Dinssa: Okay that's all good

PR #2691, opened 2024-10-11
Re-add get_file_io method

kosiakkatrina:
Accidentally removed it in https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/2689

### DIFF
--- a/app/services/storage/local_disk_service.rb
+++ b/app/services/storage/local_disk_service.rb
@@ -10,7 +10,7 @@ module Storage
     def get_file(filename)
       path = Rails.root.join("tmp/storage", filename)
 
-      File.open(path, "r")
+      File.open(path, "r").read
     end
 
     def get_file_io(filename)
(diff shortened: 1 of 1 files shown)


Dinssa (approved):

app/services/storage/local_disk_service.rb
```diff
@@ -13,6 +13,12 @@ def get_file(filename)
       File.open(path, "r")
     end
 
+    def get_file_io(filename)
```
RachaelBooth, line 16: Does this mean we've got `get_file` and `get_file_io` that are identical?

RachaelBooth: Ah, I see, they're not quite identical in the s3 version. Cool.

kosiakkatrina: Ah, not anymore...

kosiakkatrina: It needed updating here too!

PR #2692, opened 2024-10-11
CLDC-3330: Use partial postcodes in bulk upload

RachaelBooth:


app/services/bulk_upload/lettings/year2024/row_parser.rb
```diff
@@ -1355,7 +1355,7 @@ def postcode_known
   end
 
   def postcode_full
-    "#{field_21} #{field_22}" if field_21 && field_22
+    "#{field_21} #{field_22}" if field_21 || field_22
```
kosiakkatrina: Would it make sense joining with a space here instead? I assume we trim the spaces somewhere down the line tho, so if it works feel free to leave as is

kosiakkatrina (approved):

app/services/bulk_upload/lettings/year2024/row_parser.rb
RachaelBooth: It should get trimmed elsewhere, but you're right I think it would be a bit nicer.

PR #2693, opened 2024-10-11
CLDC-3649 Delete custom resource

kosiakkatrina:
This only marks the collection resource model as discarded but deletes the actual object from s3

Also added paper trail to collection resources

### DIFF
--- a/app/controllers/collection_resources_controller.rb
+++ b/app/controllers/collection_resources_controller.rb
@@ -35,7 +35,7 @@ class CollectionResourcesController < ApplicationController
   end
 
   def edit_mandatory_collection_resource
-    return render_not_found unless current_user.support?
+    return render_not_authorized unless current_user.support?
 
     year = params[:year].to_i
     resource_type = params[:resource_type]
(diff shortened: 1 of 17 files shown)


app/controllers/collection_resources_controller.rb
```diff
@@ -178,6 +178,29 @@ def create
     end
   end
 
+  def delete_confirmation
+    return render_not_found unless current_user.support?
```
Dinssa: Would it make sense to use the `render_not_authorized` method here instead? I know they eventually do the same thing but for readability. Keeping it `render_not_found` below when there is no collection resource.

app/controllers/collection_resources_controller.rb
```diff
@@ -178,6 +178,29 @@ def create
...
+
+    @collection_resource = CollectionResource.find_by(id: params[:collection_resource_id])
+
+    return render_not_found unless @collection_resource
+
+    render "collection_resources/delete_confirmation"
+  end
+
+  def delete
+    return render_not_found unless current_user.support?
```
Dinssa: Here too

app/controllers/collection_resources_controller.rb
kosiakkatrina: Yup, updated now!

Dinssa (approved):

PR #2694, opened 2024-10-11
CLDC-3654: Copy changes - Sales property information questions

Dinssa:


app/models/form/sales/pages/property_local_authority.rb
```diff
@@ -15,7 +15,7 @@ def questions
     ].compact
   end
 
-  def la_known_question
+  def la_known_question #TODO: Pre 2023, could be removed? Only used here.
     if form.start_date.year < 2023
       Form::Sales::Questions::PropertyLocalAuthorityKnown.new(nil, nil, self)
     end
```
Dinssa: A few places where the logic suggests its for forms pre 2023, which can be removed now or leave it?

app/models/form/sales/subsections/property_information.rb
```diff
@@ -16,7 +16,7 @@ def pages
       Form::Sales::Pages::PercentageDiscountValueCheck.new("percentage_discount_proptype_value_check", nil, self),
       Form::Sales::Pages::PropertyBuildingType.new(nil, nil, self),
       (uprn_questions if form.start_date.year == 2023),
-      (postcode_and_la_questions if form.start_date.year < 2023),
+      (postcode_and_la_questions if form.start_date.year < 2023), #TODO: Pre 2023, could be removed?
```
Dinssa: Pre2023

app/models/form/sales/subsections/property_information.rb
```diff
@@ -50,7 +50,7 @@ def uprn_questions
     end
   end
 
-  def postcode_and_la_questions
+  def postcode_and_la_questions #TODO: Pre 2023, could be removed?
```
Dinssa: Pre2023

app/models/form/sales/pages/uprn_confirmation.rb
```diff
@@ -2,7 +2,7 @@ class Form::Sales::Pages::UprnConfirmation < ::Form::Page
   def initialize(id, hsh, subsection)
     super
     @id = "uprn_confirmation"
-    @header = "We found an address that might be this property"
+    @copy_key = "sales.property.uprn_confirmed"
```
RachaelBooth: I think this will be the default, so doesn't need to be specified - that's probably true of the uprn selection page as well, along with many of the questions.

app/models/form/sales/questions/about_price_value_check.rb
```diff
@@ -2,8 +2,7 @@ class Form::Sales::Questions::AboutPriceValueCheck < ::Form::Question
   def initialize(id, hsh, page)
     super
     @id = "value_value_check"
-    @check_answer_label = "Purchase price confirmation"
-    @header = "Are you sure?"
+    @copy_key = "sales.property.value_value_check"
```
RachaelBooth: Presumably this question shows up in another section as well, since it's a soft validation - I think we should have a separate translation file for those rather than being grouped under form sections when they can span multiple.

Dinssa: I couldn’t find another section where this question is used. It seems to be relevant only to the property information.

RachaelBooth: It references value (i.e. the sale price?), so I assume it appears in the sale information section (for each sale type) as well?

Dinssa: Ah yes it does appear in some other subsections (outright_sale, discounted_ownership_scheme and shared_ownership_scheme). I'll remove this translation.

config/locales/forms/2023/sales/property.en.yml
```diff
@@ -0,0 +1,105 @@
...
+                        page_header: ""
+                        check_answer_label: "Buyer 1 income confirmation"
+                        hint_text: ""
+                        question_text: "Are you sure this is correct?"
+
+                    income2_value_check":
+                        page_header: ""
+                        check_answer_label: "Buyer 2 income confirmation"
+                        hint_text: ""
+                        question_text: "Are you sure this is correct?"
```
Dinssa: `income1_value_check` and `income2_value_check` are also used in the **income, benefits and savings** subsection, will remove from this translation file.

Dinssa: and `combined_income_value_check`

Dinssa: Also, `percentage_discount_value_check` which is also used in the discounted_ownership_scheme subsection.
And, `monthly_charges_value_check` used in: outright_sale, discounted_ownership_scheme and shared_ownership_scheme subsections.

app/models/form/sales/questions/about_price_value_check.rb
Dinssa: I've undone some further translations noted in another PR comment.

RachaelBooth: I think they want to be in a separate file rather than removed - the relevant text on soft validation questions/pages do want to be editable in the same way as other things, so we want to pull them into translations, just structurally it doesn't make sense for them to be inside section files.

app/models/form/sales/pages/uprn_confirmation.rb
```diff
@@ -2,7 +2,7 @@ class Form::Sales::Pages::UprnConfirmation < ::Form::Page
   def initialize(id, hsh, subsection)
     super
     @id = "uprn_confirmation"
-    @copy_key = "sales.property.uprn_confirmed"
+    @header = "We found an address that might be this property"
```
RachaelBooth: This header should still be in the translations file rather than here. You might have accidentally undone some changes or something?

app/models/form/sales/questions/about_price_value_check.rb
Dinssa: Within this branch?

RachaelBooth: Yes - This ticket includes soft validation questions used in this section (in large part so that we should then have examples of everything for overall structure, making future tickets much simpler).

app/models/form/sales/questions/about_price_value_check.rb
```diff
@@ -2,8 +2,7 @@ class Form::Sales::Questions::AboutPriceValueCheck < ::Form::Question
   def initialize(id, hsh, page)
     super
     @id = "value_value_check"
-    @check_answer_label = "Purchase price confirmation"
-    @header = "Are you sure?"
+    @copy_key = "sales.soft_validations.value_value_check"
```
RachaelBooth, line 5: Are there pages for these soft validations as well?

Dinssa: Yes there are

app/models/form/sales/pages/about_price_value_check.rb
```diff
@@ -7,7 +7,7 @@ def initialize(id, hsh, subsection)
       },
     ]
     @title_text = {
-      "translation" => "soft_validations.purchase_price.title_text",
+      "translation" => "forms.#{form.start_date.year}.#{@copy_key}.title_text",
```
RachaelBooth, line 11: Does this (and other soft validation pages) not need a copy key specifying? (Because the translations are under soft_validations rather than the subsection)

RachaelBooth (commented):
One question about copy keys for soft validation pages, otherwise looks good.

spec/models/form/sales/questions/uprn_confirmation_spec.rb
```diff
@@ -23,10 +23,6 @@
     expect(question.derived?(nil)).to be false
   end
 
-  it "has the correct unanswered_error_message" do
-    expect(question.unanswered_error_message).to eq("You must answer is this the right address?")
-  end
```
kosiakkatrina: Why is this no longer passing?

spec/models/form/sales/pages/about_price_value_check_spec.rb
```diff
@@ -27,27 +28,6 @@
...
-          "arguments_for_key" => "purchase_price_soft_min_or_soft_max",
-          "i18n_template" => "soft_min_or_soft_max",
-        },
-        {
-          "key" => "purchase_price_higher_or_lower_text",
-          "i18n_template" => "higher_or_lower",
-        },
-      ],
-    })
-  end
```
kosiakkatrina: did this need updating instead of removing?

Dinssa: Yes i've updated it

spec/models/form/sales/questions/uprn_confirmation_spec.rb
Dinssa: Updated this test to work

spec/models/form/sales/questions/uprn_confirmation_spec.rb
```diff
@@ -23,6 +23,10 @@
     expect(question.derived?(nil)).to be false
   end
 
+  it "has the correct unanswered_error_message" do
+    expect(question.unanswered_error_message).to eq("You must answer is this the right address?")
```
RachaelBooth: Isn't this dependant on a translation now? If so, we shouldn't test like this because it stops people making changes just by changing the translations file.

Dinssa: It is testing this in 2023 which I assume no one will change going forward, but I can update this to use the translation.

Dinssa: Updated uprn_confirmation and uprn_known

app/helpers/formatting_helper.rb
```diff
@@ -2,7 +2,8 @@ module FormattingHelper
   def format_ending(text)
     return text if text.blank?
 
-    modified_text  = lowercase_first_letter(text)
+    first_word = text.split.first
+    modified_text = first_word == first_word.upcase ? text : lowercase_first_letter(text)
     ensure_sentence_ending(modified_text)
   end
```
Dinssa, line 8: I changed this to not lowercase the first letter, if the whole first word is capitalised. For example, "You must answer UPRN known?" where "UPRN known?" is added after "You must answer" would have been changed to "uPRN known?".

config/locales/en.yml
```diff
@@ -764,7 +756,6 @@ Make sure these answers are correct."
       must_be_less_than_3_years_from_saledate: "You told us practical completion or handover date is more than 3 years before sale completion date."
     saledate:
       must_be_less_than_3_years_from_hodate: "You told us sale completion date is more than 3 years after practical completion or handover date."
-      informative_text: "We could not find an address that matches your search. You can search again or continue to enter the address manually."
```
kosiakkatrina, line 727: This is odd, do you know why this was under saledate? Is it no longer used?

kosiakkatrina (approved):

config/locales/en.yml
Dinssa:
```
    no_address_found:
      title_text: "No address found."
      informative_text: "We could not find an address that matches your search. You can search again or continue to enter the address manually."
```
It was under no_address_found, I think this was missed out when removing the title_text.

PR #2695, opened 2024-10-11
CLDC-3257: Default to current collection year for BU guidance pages

RachaelBooth:


app/models/forms/bulk_upload_lettings/guidance.rb
```diff
@@ -4,11 +4,18 @@ class Guidance
...
+      include CollectionTimeHelper
 
       attribute :year, :integer
       attribute :referrer
       attribute :organisation_id, :integer
 
+      def initialize(params)
+        super(params)
+
+        self.year = current_collection_start_year if year.nil?
```
RachaelBooth, line 16: Annoyingly you can't use the helper method if trying to set a default when defining the attribute (i.e. `attribute :year, :integer, default: current_collection_start_year`), so this is the alternative.

kosiakkatrina (approved):

PR #2699, opened 2024-10-15
CLDC-3667 Extract sale information validations into a separate translations file

kosiakkatrina:


config/locales/en.yml
```diff
@@ -660,46 +657,6 @@ en:
         before_deactivation: "This location was deactivated on %{date}. The reactivation date must be on or after deactivation date."
       deactivation:
         during_deactivated_period: "The location is already deactivated during this date, please enter a different date."
-    sale_information:
-      proplen:
-        social_homebuy: "Social HomeBuy buyers should not have lived here before."
-        rent_to_buy: "Rent to Buy buyers should not have lived here before."
```
kosiakkatrina, line 643: I don't think these were being used anywhere

config/locales/validations/sales/sale_information.en.yml
```diff
@@ -0,0 +1,123 @@
...
+            mortgage_not_used: "The cash deposit is %{deposit}.</br></br>The full purchase price (%{value}) multiplied by the percentage bought is %{stairbought_part_of_value}.</br></br>These two amounts should be the same."
+            mortgage_used_socialhomebuy: "The mortgage amount (%{mortgage}), cash deposit (%{deposit}), and cash discount (%{cashdis}) added together is %{mortgage_deposit_and_discount_total}.</br></br>The full purchase price (%{value}) multiplied by the percentage bought (%{stairbought}) is %{stairbought_part_of_value}.</br></br>These two amounts should be the same."
+            mortgage_not_used_socialhomebuy: "The cash deposit (%{deposit}) and cash discount (%{cashdis}) added together is %{deposit_and_discount_total}.</br></br>The full purchase price (%{value}) multiplied by the percentage bought (%{stairbought}) is %{stairbought_part_of_value}.</br></br>These two amounts should be the same."
```
kosiakkatrina: I duplicated all of these per field. I'm really not a 100% if that was the right choice, but only because these validations are long

config/locales/validations/sales/sale_information.en.yml
```diff
@@ -0,0 +1,125 @@
+en:
+  validations:
+    sales:
+      sale_information:
+        ownershipsch:
+          discounted_ownership_value: "The mortgage%{mortgage}%{deposit_and_grant_sentence} added together is %{mortgage_deposit_and_grant_total}.</br></br>The full purchase price%{discount_sentence} is %{value_with_discount}.</br></br>These two amounts should be the same."
+          outright_sale_value: "The mortgage%{mortgage} and cash deposit (%{deposit}) when added together is %{mortgage_and_deposit_total}.</br></br>The full purchase price is %{value}.</br></br>These two amounts should be the same."
```
RachaelBooth, line 7: Do we want to adjust how we're passing mortgage in here to be more like we're doing with deposit? It seems a bit inconsistent for editing.

kosiakkatrina: I think in some cases mortgage is not used or 0, but the copy for error messages was the same, if we did it the same as deposit, in some cases it would be empty brackets

RachaelBooth (approved):

PR #2700, opened 2024-10-15
CLDC-3655 Extract property information validations

kosiakkatrina:


app/models/validations/sales/property_validations.rb
```diff
@@ -4,19 +4,19 @@ def validate_postcodes_match_if_discounted_ownership(record)
     return unless record.ppostcode_full.present? && record.postcode_full.present?
 
     if record.discounted_ownership_sale? && record.ppostcode_full != record.postcode_full
-      record.errors.add :postcode_full, I18n.t("validations.property.postcode.must_match_previous", buyer_possessive: record.joint_purchase? ? "Buyers’" : "Buyer’s")
-      record.errors.add :ppostcode_full, I18n.t("validations.property.postcode.must_match_previous", buyer_possessive: record.joint_purchase? ? "Buyers’" : "Buyer’s")
-      record.errors.add :ownershipsch, I18n.t("validations.property.postcode.must_match_previous", buyer_possessive: record.joint_purchase? ? "Buyers’" : "Buyer’s")
-      record.errors.add :uprn, I18n.t("validations.property.postcode.must_match_previous", buyer_possessive: record.joint_purchase? ? "Buyers’" : "Buyer’s")
+      record.errors.add :postcode_full, I18n.t("validations.sales.property_information.postcode_full.postcode_must_match_previous", buyer_possessive: record.joint_purchase? ? "Buyers’" : "Buyer’s")
```
RachaelBooth: It would be more duplication, but do we want to separate the joint and non joint versions so they can be edited in ways that would change this possessive? (e.g. to remove the capital letter if it's elsewhere in the sentence)

Possibly using count / the pluralisation thing rather than completely separated, but it's duplication either way. (I could be persuaded not to if we don't think it's needed - it is something that could be done down the line if it happened to ever come up, just would be annoying if it did)

kosiakkatrina: I think it's good idea, buyer_possessive would probably be an annoying variable if someone wants to completely reword the message

RachaelBooth (approved):

PR #2701, opened 2024-10-15
CLDC-3659: Pull sales household situation questions copy into translation file

RachaelBooth:


app/models/form/sales/questions/buyers_organisations.rb
```diff
@@ -27,9 +24,5 @@ def displayed_answer_options(_log, _user = nil)
     }
   end
 
-  def unanswered_error_message
```
RachaelBooth, line 30: This is not actually used - it is perfectly fine to leave all these options blank.

kosiakkatrina: Would it not be used in row parser? `errors.add(field, question.unanswered_error_message)`

RachaelBooth: I don't think so, because we ask this as four separate columns in BU (and none is also acceptable, but not an explicit column, so there shouldn't be an error for not selecting any of them)

kosiakkatrina: Ah, okay. I see!

config/locales/forms/2023/sales/household_situation.en.yml
```diff
@@ -0,0 +1,50 @@
...
+          last_accommodation_la:
+            page_header: ""
+            previous_la_known:
+              check_answer_label: "Local authority of buyer 1’s last settled accommodation"
+              hint_text: "This is also known as the household’s 'last settled home'"
+              question_text: "Do you know the local authority of buyer 1’s last settled accommodation?"
+            prevloc:
+              check_answer_label: "Local authority of buyer 1’s last settled accommodation"
+              hint_text: ""
+              question_text: "Select a local authority"
```
kosiakkatrina, line 32: When there are 2 questions on a page, I've just been defining them the same as 1 question per page, to get the default copy_key behaviour. Is there pros/cons to doing it one way over the other?

RachaelBooth: Assuming that we don't change what's on the page / the order of things on the page later, I guess it doesn't make much difference.

I think I find this slightly more logical, in that it wouldn't be possible to set page_header from the second question so it either has a useless setting or looks weird without explanation by missing that setting. (It fits with how type / othtype ended up in the setup section, although that one has the additional complication of there being three pages/question sets)

kosiakkatrina: Okay, I think it makes sense. I'll adjust my PR to be consistent with this. Thank you!

kosiakkatrina (approved):

PR #2702, opened 2024-10-15
CLDC-3656 Extract household characteristics question and page copy to translation files

kosiakkatrina:


app/models/form/sales/pages/buyer_interview.rb
```diff
@@ -2,7 +2,11 @@ class Form::Sales::Pages::BuyerInterview < ::Form::Page
   def initialize(id, hsh, subsection, joint_purchase:)
     super(id, hsh, subsection)
     @joint_purchase = joint_purchase
-    @copy_key = "sales.setup.noint.#{joint_purchase ? 'joint_purchase' : 'not_joint_purchase'}"
+    @copy_key = if form.start_year_after_2024?
+                  "sales.setup.noint.#{joint_purchase ? 'joint_purchase' : 'not_joint_purchase'}"
+                else
+                  "sales.household_characteristics.noint.#{joint_purchase ? 'joint_purchase' : 'not_joint_purchase'}"
```
RachaelBooth: I wondered on seeing this about using `subsection.id` to avoid the if-else here, but that may be more confusing - will leave that thought with you.

app/models/form/sales/pages/buyer1_ethnic_background_arab.rb
```diff
@@ -2,6 +2,7 @@ class Form::Sales::Pages::Buyer1EthnicBackgroundArab < ::Form::Page
   def initialize(id, hsh, subsection)
     super
     @id = "buyer_1_ethnic_background_arab"
+    @copy_key = "sales.household_characteristics.ethnic.ethnic_background_asian"
```
RachaelBooth: This should be `_arab` rather than `_asian`.

app/models/form/sales/pages/buyer_interview.rb
kosiakkatrina: Good idea!

app/models/form/sales/pages/buyer1_ethnic_background_arab.rb
kosiakkatrina: Nice spot!

RachaelBooth (approved):

PR #2711, opened 2024-10-18
CLDC-3665 Extract sale information copy

kosiakkatrina:


app/models/form/sales/pages/handover_date.rb
```diff
@@ -2,6 +2,7 @@ class Form::Sales::Pages::HandoverDate < ::Form::Page
   def initialize(id, hsh, subsection)
     super
     @id = "handover_date"
+    @copy_key = "sales.sale_information.handover_date"
```
Dinssa, line 5: I don't think you need to specific the copy key within the pages in most cases where there is just one question per page. The page model grabs the copy key and header from the copy key specified in the first question of the page. 

**app/models/form/page.rb**
```
  def copy_key
    @copy_key ||= "#{form.type}.#{subsection.id}.#{questions[0].id}"
  end

  def header
    @header ||= I18n.t("forms.#{form.start_date.year}.#{copy_key}.page_header", default: "")
  end
```

I think we only need to specify the copy key on a page where the page has multiple questions and we have a key within the translation file shared between multiple questions like:
```
          address:
            page_header: "Q15 - What is the property's address?"
            address_line1:
              check_answer_label: "Address lines 1 and 2"
              hint_text: ""
              question_text: "Address line 1"
            address_line2:
              check_answer_label: ""
              hint_text: ""
              question_text: "Address line 2 (optional)"
            town_or_city:
              check_answer_label: "Town or city"
              hint_text: ""
              question_text: "Town or city"
            county:
              check_answer_label: "County"
              hint_text: ""
              question_text: "County (optional)"
            postcode_full:
              check_answer_label: "Postcode"
              hint_text: ""
              question_text: "Postcode"

```
In this case there is one page header for all the questions, so we need to specify it'll be at sales.address.page_header, had it not been specified we may have looked for the page header like this: sales.address.address_line1.page_header and come up empty.

This is an example of a question where we don't need to add the copy key to the page because it'll grab it from the one and only question on the page:
```
          uprn_confirmed:
            page_header: "We found an address that might be this property"
            check_answer_label: "Is this the right address?"
            hint_text: ""
            question_text: "Is this the property address?"
```

kosiakkatrina: Hmm, yeah I see what you mean. I think for this one I ended up explicitly adding copy_key cause `sale_information` isn't the subsection id. It's actually `shared_ownership_scheme`/`discounted_ownership_scheme`/`outright_sale` but a lot of questions are repeated across these sections, so I set a different id instead of duplicating them. Maybe duplicating was the correct choice, I'm not sure, any thoughts?

Dinssa (approved):

PR #2714, opened 2024-10-21
CLDC-2399: Bulk upload table styling

Dinssa:


app/components/bulk_upload_error_row_component.html.erb
```diff
@@ -49,24 +49,24 @@
...
+            <% errors.each_with_index do |error, index| %>
+              <% row_class = row_classes(index, errors.size) %>
+              <% body.with_row(html_attributes: { class: row_class }) do |row| %>
+                <% row.with_cell(text: error.cell) %>
+                <% row.with_cell(text: question_for_field(error.field), html_attributes: { class: "govuk-!-width-one-half" }) %>
+                <% if index == 0 %>
+                  <% cell_class = cell_classes(group_index, total_groups) %>
+                  <% row.with_cell(text: error_message.html_safe, rowspan: errors.size, html_attributes: { class: cell_class }) %>
                 <% end %>
+                <% row.with_cell(text: error.field.humanize) %>
```
Dinssa, line 64: We group errors together to merge the error message cell. The existing logic removes the row borders and then reinstates them for the first and last rows, leaving all rows in between without borders. To exclude the bottom border from the last group, I needed to determine the total number of groups. The logic to exclude the last row is implemented within the `cell_classes` method by not adding the `.grouped-multirow-cell` class, which would otherwise add the border back to the merged cell.

app/frontend/styles/_bulk-uploads.scss
```diff
@@ -12,6 +12,11 @@
   border-bottom: 1px solid #b1b4b6;
 }
 
+.no-bottom-border,
+.no-bottom-border > tbody > tr:last-of-type td {
+  border-bottom: none;
+}
+
```
Dinssa, line 19: What's important here is the use of CSS selectors to target the `td` element. While the `.no-bottom-border` class isn't directly used, it could be useful for future reuse.

kosiakkatrina (approved):

PR #2721, opened 2024-10-24
CLDC-2639 Display correct back to logs button

kosiakkatrina:


app/services/storage/local_disk_service.rb
```diff
@@ -19,7 +19,9 @@ def get_file_io(filename)
       File.open(path, "r")
     end
 
-    def write_file(filename, data, _content_type: nil)
+    # rubocop:disable Lint/UnusedMethodArgument
+    def write_file(filename, data, content_type: nil)
```
kosiakkatrina, line 23: This is irrelevant but fixes local write_file calls

RachaelBooth (approved):

PR #2723, opened 2024-10-24
CLDC-3679: Copy changes lettings setup questions

Dinssa:


app/models/form/lettings/questions/location_id_search.rb
```diff
@@ -2,10 +2,8 @@ class Form::Lettings::Questions::LocationIdSearch < ::Form::Question
   def initialize(id, hsh, page)
     super
     @id = "location_id"
-    @check_answer_label = "Location"
-    @header = header_text
-    @hint_text = '<div class="govuk-inset-text">This scheme has 20 or more locations.</div>Enter postcode or address.'
     @type = "select"
+    @hint = I18n.t("forms.#{form.start_date.year}.#{copy_key}.search_hint_text", default: "") if form.start_date.present?
```
kosiakkatrina: Should this be @hint_text?, also, wouldn't it just get pulled as long as the copy key is correct and the key is hint_text instead of search_hint_text?

Dinssa: There are two questions with the same id "location_id", there's the default which displays the answer options as radio buttons and then a few months ago I add an alternative which displayed the options in a search box (when there were more than 20 locations). The default hint_text is for the radio buttons version and this search_hint_text is for the alternative option. I added this to override the default hint_text it would grab because we do have a different hint message when there are more than 20 locations on the scheme.

Dinssa: Back when adding the search version, I kept the id's of the two different questions the same so they displayed the same way in the url.

spec/features/form/check_answers_page_lettings_logs_spec.rb
```diff
@@ -159,26 +159,6 @@
...
-      end
-
-      it "displays inferred postcode with the location_admin_district" do
-        lettings_log.update!(location:)
-        visit("/lettings-logs/#{id}/setup/check-answers")
-        expect(page).to have_content("Location")
-        expect(page).to have_content(location.location_admin_district)
-      end
-    end
-
```
kosiakkatrina: Do we just want to update this test with translation instead of removing it? It doesn't look like it's only testing content

spec/features/lettings_log_spec.rb
```diff
@@ -262,7 +262,6 @@
         click_button("Save and continue")
         log_id = page.current_path.scan(/\d/).join
         visit("lettings-logs/#{log_id}/setup/check-answers")
-        expect(page).to have_content("Stock owner User org", normalize_ws: true)
```
kosiakkatrina: Same here, should this be updated with translation?

spec/requests/form_controller_spec.rb
```diff
@@ -746,7 +746,6 @@
           it "re-renders the same page with errors if validation fails" do
             post "/lettings-logs/#{lettings_log.id}/managing-organisation", params: params
             expect(page).to have_content("There is a problem")
-            expect(page).to have_content("Error: Which organisation manages this letting?")
```
kosiakkatrina: And here? Are these failing? cause I don't think they should

app/models/form/lettings/questions/location_id_search.rb
kosiakkatrina: Ah I see, I've been duplicating and setting relevant copy keys when there are 2 versions of the question, in case we want to update the wording for other attributes in the future, so I'd have something like
```
          location_id:
            over_20_results:
              page_header: "Location"
              check_answer_label: "Location"
              hint_text:  '<div class="govuk-inset-text">This scheme has 20 or more locations.</div>Enter postcode or address.'
              question_text: "Which location is this log for?"
            under_20_results:
              page_header: "Location"
              check_answer_label: "Location"
              hint_text:  '<div class="govuk-inset-text">This scheme has 20 or more locations.</div>Enter postcode or address.'
              question_text: "Which location is this log for?"
 ```
 
This also would meant that we don't have to define a new/custom key, but idm whether you change it or keep it as is

Comment (Dinssa) [reply app/models/form/lettings/questions/location_id_search.rb]: Ah, this is a good idea I'll change it to something like this to keep this consistent.

File: spec/features/form/check_answers_page_lettings_logs_spec.rb (hunk 2 above)
Comment (Dinssa) [reply spec/features/form/check_answers_page_lettings_logs_spec.rb]: These tests are for 2021, which won't have a translation file. We could update the tests to use 2023 going forward?

File: spec/requests/form_controller_spec.rb (hunk 4 above)
Comment (Dinssa) [reply spec/requests/form_controller_spec.rb]: Ye it was failing, I think because they're testing pre-2023/2024 so don't use the translation files.

Comment (Dinssa) [reply spec/requests/form_controller_spec.rb]: Updated most of these tests to 2023, so the labels now can be tested.

File: spec/features/form/check_answers_page_lettings_logs_spec.rb [hunk 5]
```diff
@@ -167,14 +167,12 @@
...
         expect(page).to have_content(location.name)
       end
 
       it "displays inferred postcode with the location_admin_district" do
         lettings_log.update!(location:)
         visit("/lettings-logs/#{id}/setup/check-answers")
-        expect(page).to have_content("Location")
         expect(page).to have_content(location.location_admin_district)
       end
     end
```
Dinssa, line 178: I wasn't able to update this test to 2023 easily without major changes to the rest of the tests. Just deleted the test for the check answer label. As it doesn't appear anymore pre 2023.

spec/requests/form_controller_spec.rb
kosiakkatrina: Nice, thanks!

spec/features/form/check_answers_page_lettings_logs_spec.rb
kosiakkatrina: I guess we could where it makes sense to

kosiakkatrina (approved):

PR #2725, opened 2024-10-25
CLDC-3365 Validate BU file size

kosiakkatrina:
At the moment we don't have a limit on BU file size, as long as it's a csv we try to load it into memory and process it but in some instances when file is too large, the service can't handle it.
We wouldn't expect more than 13k rows of data to be logged at any given time (and that is unlikely)

13 000 rows with data ~ 4MB (4.2MB is the file is saved as .xlsx)
13 000 rows with data + empty rows up to ~1mil  ~ 140MB

This PR limits the size of the file for BU to 10MB


RachaelBooth (commented):
Do we know how well our system does at handling a 10 megabyte file of data? (e.g. how long it takes)

kosiakkatrina:
> Do we know how well our system does at handling a 10 megabyte file of data? (e.g. how long it takes)

@RachaelBooth I don't think we had a file this large, I think 10 is slightly arbitrary. The largest files we have processed would have been between 4000 and 5000 rows, not sure about size, because of retention, but I suspect it would have been under 2MB. Anything larger than those files was way larger, and in most cases just a badly saved .xlsx files with lots of empty rows, so I think the main purpose for this limit is to prevent those super large files, cause the medium large files are unlikely to occur in the first place

RachaelBooth (approved):

PR #2728, opened 2024-10-28
CLDC-3681 Extract lettings household characteristics copy

kosiakkatrina:


app/models/form/lettings/questions/age.rb
```diff
@@ -1,9 +1,8 @@
 class Form::Lettings::Questions::Age < ::Form::Question
-  def initialize(id, hsh, page, person_index:, person_type:)
```
kosiakkatrina, line 2: Removed person_type and all the duplicate pages, because it was only being used to determine pre 2024 hint. Individual questions for 2023 should no longer be accessible, so this doesn't really matter + removing this makes copy file simpler

Dinssa (approved):

PR #2731, opened 2024-10-28
CLDC-3684 Extract lettings tenancy information copy

kosiakkatrina:


config/locales/forms/2023/lettings/tenancy_information.en.yml
```diff
@@ -0,0 +1,66 @@
...
+              page_header: ""
+              check_answer_label: "Length of fixed-term tenancy"
+              hint_text: "Do not include the starter or introductory period.</br>The minimum period is 1 year for intermediate rent general needs logs and you do not need a log for shorter tenancies."
+              question_text: "What is the length of the fixed-term tenancy to the nearest year?"
+            tenancy_length_periodic:
+              page_header: ""
+              check_answer_label: "Length of periodic tenancy"
+              hint_text: "As this is a periodic tenancy, this question is optional. If you do not have the information available click save and continue"
+              question_text: "What is the length of the periodic tenancy to the nearest year?"
+
```
Dinssa, line 61: Why group these together? They're different pages with different questions.

config/locales/forms/2023/lettings/tenancy_information.en.yml
```diff
@@ -0,0 +1,66 @@
...
+            starter_tenancy_type:
+              page_header: ""
+              tenancy:
+                check_answer_label: "Type of main tenancy after the starter period has ended"
+                hint_text: "This is also known as an ‘introductory period’."
+                question_text: "What is the type of tenancy after the starter period has ended?"
+              tenancyother:
+                check_answer_label: ""
+                hint_text: ""
+                question_text: "Please state the tenancy type"
```
Dinssa, line 38: Same here, tenancy type and starter tenancy type are two separate pages, might be better as separate keys without "tenancy" as a parent? Does it make sense to keep the page as the parent key (directly after tenancy_information) and any variations underneath that?

config/locales/forms/2023/lettings/tenancy_information.en.yml
kosiakkatrina: Hmm, good question! I grouped them cause the question id is the same in all the questions, so it's basically collecting the same data, but we're asking the question differently. And we would have to set copy_key manually anyway

config/locales/forms/2023/lettings/tenancy_information.en.yml
kosiakkatrina: I think similarly to above I grouped by the question id, cause underlying data is more likely to stay the same as opposed to how we split it into different question versions

Dinssa (approved):

PR #2732, opened 2024-10-28
CLDC-3685: Copy changes lettings income and benefits

Dinssa:


config/locales/forms/2023/lettings/income_and_benefits.en.yml
```diff
@@ -0,0 +1,183 @@
...
+                hint_text: "For example, for heating or hot water. This doesn’t include housing benefit or Universal Credit."
+                question_text: "What is the personal service charge?"
+              supcharg:
+                check_answer_label: "Support charge"
+                hint_text: "Any charges made to fund support services included in tenancy agreement."
+                question_text: "What is the support charge?"
+              tcharge:
+                check_answer_label: "Household rent and charges"
+                hint_text: "This is the total for rent and all charges."
+                question_text: "Total charge"
```
Dinssa: This is all repeated for rent_weekly, rent_bi_weekly, rent_4_weekly and rent_monthly. Is there any scenario where these questions will ever be different? Based of the rent period. Could they be merged into just a "rent" key?

config/locales/forms/2024/lettings/income_and_benefits.en.yml
```diff
@@ -0,0 +1,177 @@
...
+            chcharge_4_weekly:
+              check_answer_label: "Care home charges"
+              hint_text: ""
+              question_text: "How much does the household pay every 4 weeks?"
+            chcharge_monthly:
+              check_answer_label: "Care home charges"
+              hint_text: ""
+              question_text: "How much does the household pay every month?"
+
+          rent_weekly:
```
RachaelBooth: I know it's not how we've structured pages with multiple questions elsewhere, but I wonder if this lot might be easier to use if it was grouped by question and then by frequency rather than having frequency as the outer "thing"?

RachaelBooth: Actually, it looks like these don't differ at all between different frequencies. Is it worth asking the stats team whether that's a fundamental "these will always be the same" (so we could avoid the duplication) or not? It looks like the questions themselves have suffixes for the fields which mention the frequency, and that's the only difference.

Dinssa: Ye I think we could merge these. The definitions and meaning (check answer label, etc) of any of the rent or charges questions shouldn't be different just because of the rent frequency. But I'll check.

Dinssa: I've updated this and merged the information into one key "rent_and_charges", assuming that it's all okay with the stats team. Can be reverted if not.

RachaelBooth (approved):

PR #2739, opened 2024-10-30
CLDC-3691: Extract copy - lettings setup validations

Dinssa:


config/locales/validations/lettings/setup.en.yml
```diff
@@ -0,0 +1,60 @@
...
+
+        location:
+          incomplete: "This location is incomplete. Select another location or update this one."
+
+
+
+
+
+
+
```
kosiakkatrina: That's a lot of empty lines!

kosiakkatrina (approved):

PR #2741, opened 2024-10-30
 CLDC-3689: Extract lettings household validations

Dinssa:


app/models/validations/household_validations.rb
```diff
@@ -25,29 +25,29 @@ def validate_reasonable_preference(record)
 
   def validate_reason_for_leaving_last_settled_home(record)
     if record.reason == 32 && record.underoccupation_benefitcap != 4
-      record.errors.add :underoccupation_benefitcap, I18n.t("validations.household.underoccupation_benefitcap.dont_know_required")
-      record.errors.add :reason, I18n.t("validations.household.underoccupation_benefitcap.dont_know_required")
+      record.errors.add :underoccupation_benefitcap, I18n.t("validations.lettings.household.underoccupation_benefitcap.dont_know_required")
+      record.errors.add :reason, I18n.t("validations.lettings.household.underoccupation_benefitcap.dont_know_required")
```
kosiakkatrina: Can we split this per question?
 `I18n.t("validations.lettings.household.reason.dont_know_required")`

app/models/validations/household_validations.rb
```diff
@@ -161,53 +161,53 @@ def validate_previous_housing_situation(record)
...
-      record.errors.add :referral, :internal_transfer_fixed_or_lifetime, message: I18n.t("validations.household.referral.la_general_needs.internal_transfer")
+      record.errors.add :prevten, :internal_transfer_fixed_or_lifetime, message: I18n.t("validations.lettings.household.prevten.la_general_needs.internal_transfer")
+      record.errors.add :referral, :internal_transfer_fixed_or_lifetime, message: I18n.t("validations.lettings.household.referral.la_general_needs.internal_transfer")
     end
   end
 
   def validate_prevloc(record)
     if record.previous_la_known? && record.prevloc.blank?
-      record.errors.add :prevloc, I18n.t("validations.household.previous_la_known")
+      record.errors.add :prevloc, I18n.t("validations.lettings.household.previous_la_known")
```
kosiakkatrina: ` I18n.t("validations.lettings.household.prevloc.previous_la_known")` ?

app/models/validations/household_validations.rb
```diff
@@ -161,53 +161,53 @@ def validate_previous_housing_situation(record)
...
   end
 
   def validate_layear(record)
     return unless record.layear && record.renewal
 
     if record.is_renewal? && record.layear == 1
-      record.errors.add :layear, :renewal_just_moved, message: I18n.t("validations.household.renewal_just_moved_to_area.layear")
-      record.errors.add :renewal, I18n.t("validations.household.renewal_just_moved_to_area.renewal")
+      record.errors.add :layear, :renewal_just_moved, message: I18n.t("validations.lettings.household.renewal_just_moved_to_area.layear")
+      record.errors.add :renewal, I18n.t("validations.lettings.household.renewal_just_moved_to_area.renewal")
```
kosiakkatrina: Can we split this by question too?

kosiakkatrina: Oh, sorry, it already is, but can we flip it?

app/models/validations/household_validations.rb
```diff
@@ -161,53 +161,53 @@ def validate_previous_housing_situation(record)
...
-      record.errors.add :ppostcode_full, :renewal_just_moved, message: I18n.t("validations.household.same_la_just_moved_to_area.previous_la")
-      record.errors.add :prevloc, :renewal_just_moved, message: I18n.t("validations.household.same_la_just_moved_to_area.previous_la")
+      record.errors.add :layear, :renewal_just_moved, message: I18n.t("validations.lettings.household.same_la_just_moved_to_area.layear")
+      record.errors.add :la, :renewal_just_moved, message: I18n.t("validations.lettings.household.same_la_just_moved_to_area.current_la")
+      record.errors.add :postcode_full, :renewal_just_moved, message: I18n.t("validations.lettings.household.same_la_just_moved_to_area.current_la")
+      record.errors.add :uprn, :renewal_just_moved, message: I18n.t("validations.lettings.household.same_la_just_moved_to_area.current_la")
+      record.errors.add :ppostcode_full, :renewal_just_moved, message: I18n.t("validations.lettings.household.same_la_just_moved_to_area.previous_la")
+      record.errors.add :prevloc, :renewal_just_moved, message: I18n.t("validations.lettings.household.same_la_just_moved_to_area.previous_la")
```
kosiakkatrina: Can we have all of these by question id? `...household.question_id...` etc

app/models/validations/household_validations.rb
```diff
@@ -161,53 +161,53 @@ def validate_previous_housing_situation(record)
...
   end
 
   def validate_combination_of_housing_needs_responses(record)
     if record.has_housingneeds? && record.housingneeds_type_not_listed? && record.no_or_unknown_other_housing_needs?
-      record.errors.add :housingneeds, I18n.t("validations.household.housingneeds.invalid")
-      record.errors.add :housingneeds_type, I18n.t("validations.household.housingneeds.invalid")
-      record.errors.add :housingneeds_other, I18n.t("validations.household.housingneeds.invalid")
+      record.errors.add :housingneeds, I18n.t("validations.lettings.household.housingneeds.invalid")
+      record.errors.add :housingneeds_type, I18n.t("validations.lettings.household.housingneeds.invalid")
+      record.errors.add :housingneeds_other, I18n.t("validations.lettings.household.housingneeds.invalid")
```
kosiakkatrina: Can these have keys `housingneeds_type` and `housingneeds_other` ?

app/models/validations/household_validations.rb
```diff
@@ -161,53 +161,53 @@ def validate_previous_housing_situation(record)
...
     return unless record.layear && record.la && record.prevloc && record.collection_start_year
 
     if record.la == record.prevloc && record.layear == 1 && record.collection_start_year >= 2023
-      record.errors.add :layear, :renewal_just_moved, message: I18n.t("validations.household.same_la_just_moved_to_area.layear")
-      record.errors.add :la, :renewal_just_moved, message: I18n.t("validations.household.same_la_just_moved_to_area.current_la")
-      record.errors.add :postcode_full, :renewal_just_moved, message: I18n.t("validations.household.same_la_just_moved_to_area.current_la")
-      record.errors.add :uprn, :renewal_just_moved, message: I18n.t("validations.household.same_la_just_moved_to_area.current_la")
-      record.errors.add :ppostcode_full, :renewal_just_moved, message: I18n.t("validations.household.same_la_just_moved_to_area.previous_la")
-      record.errors.add :prevloc, :renewal_just_moved, message: I18n.t("validations.household.same_la_just_moved_to_area.previous_la")
+      record.errors.add :layear, :renewal_just_moved, message: I18n.t("validations.lettings.household.layear.same_la_just_moved_to_area.current_la")
```
RachaelBooth: It looks like the `.current_la` or `.previous_la` at the end of each of these might be performing a slightly more generic version of what we're doing with putting the field in (i.e. allowing different messages depending on where you're seeing them). Do we actually need both, or could these all be just e.g. `validations.letttings.household.layear.same_la_just_moved_to_area`?

Dinssa: Would that also be for the rest too (`la`, `uprn`, `prevloc`)? Only `postcode_full` has both `current_la` and `previous_la`. `ppostcode_full` also only has `previous_la`

RachaelBooth: Does `postcode_full` actually use both? Presumably it would be in this list if it did, and there's only current used here (also it would be surprising I think to have a previous la type thing attached to `postcode_full` rather than `ppostcode_full`).

I think this applies to all 6 fields/error messages in this validation, unless `postcode_full` is using a previous la version somewhere I can't see.

Dinssa: Ah yes it doesn't actually use the previous_la, don't know why I included it. Will remove it and put these all directly under `same_la_just_moved_to_area`

Dinssa: Updated

RachaelBooth (approved):

PR #2743, opened 2024-10-30
CLDC-3692: Extract lettings tenancy validations

Dinssa:


app/models/validations/tenancy_validations.rb
```diff
@@ -10,7 +10,7 @@ def validate_supported_housing_fixed_tenancy_length(record)
     min_tenancy_length = 1
     return if record.tenancylength.to_i.between?(min_tenancy_length, 99)
 
-    message = I18n.t("validations.tenancy.length.invalid_fixed", min_tenancy_length:)
+    message = I18n.t("validations.lettings.tenancy.length.invalid_fixed", min_tenancy_length:)
```
kosiakkatrina: Can we have separate messages depending on field they're getting added to?

app/models/validations/tenancy_validations.rb
```diff
@@ -23,7 +23,7 @@ def validate_general_needs_fixed_tenancy_length_affordable_social_rent(record)
     min_tenancy_length = 2
     return if record.tenancylength.to_i.between?(min_tenancy_length, 99)
 
-    message = I18n.t("validations.tenancy.length.invalid_fixed", min_tenancy_length:)
+    message = I18n.t("validations.lettings.tenancy.length.invalid_fixed", min_tenancy_length:)
```
kosiakkatrina: Same here?
`I18n.t("validations.lettings.tenancy.needstype.invalid_fixed_tenancylength", min_tenancy_length:)` etc

kosiakkatrina: And for the rest of the validations too

Dinssa: Sure

Dinssa: Done. Is this so that we can have different messages for each field?

RachaelBooth (approved):

PR #2744, opened 2024-10-30
CLDC-3693: Extract shared validations

Dinssa:
I've separated into two commits changes that should be included in the lettings household and setup validations and perhaps not in the shared validations.


app/models/validations/shared_validations.rb
```diff
@@ -79,9 +79,9 @@ def location_during_startdate_validation(record)
 
     if location_inactive_status.present?
```
RachaelBooth: I think this validation wants to live in a different file to this one - for that same reason that you've split out where the translations are. Given there's other scheme stuff in `setup_validations.rb` it probably goes in there?

The postcode one above feels like it probably also doesn't live in this file, even though I think that will be applied to both sales and lettings - the best way of handling that might need a little more thinking (e.g. at the moment I think we just duplicate the UPRN validation in property validations for sales/lettings, which may be the way to go but is also not necessarily ideal).

RachaelBooth: The same applies to the other scheme/location related validations.

I've also got question marks over `shared_validate_partner_count` and `validate_owning_organisation_data_sharing_agreement_signed` similar to the postcode one - I think it's essentially the difference between generic validations that apply to any question fitting a certain type and specific validations applied to specified questions, even though they are shared between lettings and sales.

Dinssa: I've moved the location, scheme and tenancy start date validation methods to lettings setup validations. What's the downsides to leaving the postcode method in the shared validations file? I'm hesitant to duplicate the method. Since it's in the shared validations file and starts with "validate" it'll run everytime right? But that would also be the same situation even if moved to the property_information validations file, right?

RachaelBooth: It'll run here, it's entirely about what makes sense as code structure. 

The majority of this file is generic, i.e. validations that will run on any question that matches a certain condition (e.g. any numeric question with a range set). That's not true of the postcode validation, the partner count validation, or the owning org data sharing agreement validation, which are all specific validations on specific questions. 

I think the key is where would you look for them - I think I'd look for postcode validations in `property_validations` where the uprn validation is (and that is duplicated for lettings and sales), I'd look for partner count validation in `household_validations` (and in fact in that case the validation is there and uses this as a shared function it calls with different max_people parameters), and I'd look for an owning org dsa validation in `setup_validations`. 

I think the nice way to set it up would be to have a file of generic validations and separately have e.g. shared household validations used by both lettings and sales, as well as individual files for lettings-only and sales-only things. But that would be a bit much to do right now. I think duplication between sales and lettings is annoying but not as huge a problem as not knowing where to look for things, particularly given they do typically have slight differences (e.g. which error message). (That's why I said it might need more thinking about the best solution)

Dinssa: Moved those methods around as you've described.

RachaelBooth (approved):

app/models/validations/setup_validations.rb
```diff
@@ -72,19 +72,15 @@ def validate_irproduct_other(record)
...
       record.errors.add :scheme_id, :incomplete, message: I18n.t("validations.lettings.setup.scheme.incomplete")
     end
 
+    if record.scheme&.locations.present? && record.scheme.locations.confirmed.none?
+      record.errors.add :scheme_id, :no_completed_locations, message: I18n.t("validations.lettings.setup.scheme.no_completed_locations")
+    end
+
     scheme_during_startdate_validation(record)
     tenancy_startdate_with_scheme_locations(record)
   end
```
Dinssa: When a scheme is incomplete, the only error message displayed was “This scheme cannot be chosen as it has no completed locations.” While it is true that a scheme with incomplete details also lacks completed locations (as you cannot begin creating any until the details are complete), this is not the most relevant error message when the user has started but not finished creating a scheme. This change aims to display the “This scheme is incomplete. Select another scheme or update this one.” message and only show the “This scheme cannot be chosen as it has no completed locations.” message when the scheme details are complete but there are no completed locations.

app/models/validations/setup_validations.rb
```diff
@@ -72,19 +72,15 @@ def validate_irproduct_other(record)
...
-      record.errors.add :scheme_id, :no_completed_locations, message: I18n.t("validations.lettings.setup.scheme.no_completed_locations")
-    end
-  end
-
   def validate_scheme(record)
     if record.scheme&.status == :incomplete
       record.errors.add :scheme_id, :incomplete, message: I18n.t("validations.lettings.setup.scheme.incomplete")
     end
 
+    if record.scheme&.locations.present? && record.scheme.locations.confirmed.none?
```
RachaelBooth: This makes sense for enforcing order, but does it not also want to show if `record.scheme.locations.present?` is false?

app/models/validations/setup_validations.rb
```diff
@@ -77,7 +77,7 @@ def validate_scheme(record)
       record.errors.add :scheme_id, :incomplete, message: I18n.t("validations.lettings.setup.scheme.incomplete")
     end
 
-    unless record.scheme&.locations&.confirmed&.any?
+    if record.scheme&.locations.present? && !record.scheme.locations.confirmed.any?
       record.errors.add :scheme_id, :no_completed_locations, message: I18n.t("validations.lettings.setup.scheme.no_completed_locations")
     end
```
Dinssa: Alternative: This triggered a lot of tests to fail as they tested schemes without creating completed locations. Perhaps we could update the factory for a scheme? But that might affect tests that purposely do not include a location on the scheme.

PR #2748, opened 2024-11-04
CLDC-3066: Show filters button bug

Dinssa:
Adds the a show filters toggle to the user and scheme pages on smaller screens

## Additional change
This PR also adjusts the spacing between buttons to match that of the show filter button.

### Previously
The show filters button is added but does not fit with the existing buttons. It is possible to add more margin to the right of the button - adding a property to the `.app-filter-toggle__button` class using CSS `!important` which we shouldn't use (Rubycop).
Screenshot 2024-11-04 at 18 54 09

### Alternative - Not compliant with RubyCop Lint tests.
Screenshot 2024-11-04 at 19 10 39

### Solution
Screenshot 2024-11-04 at 18 59 40

### DIFF
--- a/app/views/organisations/users.html.erb
+++ b/app/views/organisations/users.html.erb
@@ -12,10 +12,12 @@
   <h2 class="govuk-visually-hidden">Users</h2>
 <% end %>
 
-<% if current_user.data_coordinator? || current_user.support? %>
-  <%= govuk_button_link_to "Invite user", new_user_path(organisation_id: @organisation.id), html: { method: :get } %>
-<% end %>
 <div class="app-filter-layout" data-controller="filter-layout">
+  <div class="govuk-button-group app-filter-toggle govuk-!-margin-bottom-6">
+    <% if current_user.data_coordinator? || current_user.support? %>
+      <%= govuk_button_link_to "Invite user", new_user_path(organisation_id: @organisation.id), html: { method: :get } %>
+    <% end %>
+  </div>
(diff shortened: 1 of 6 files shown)


app/views/organisations/users.html.erb
```diff
@@ -12,10 +12,12 @@
...
 
-<% if current_user.data_coordinator? || current_user.support? %>
-  <%= govuk_button_link_to "Invite user", new_user_path(organisation_id: @organisation.id), html: { method: :get } %>
-<% end %>
 <div class="app-filter-layout" data-controller="filter-layout">
+  <div class="govuk-button-group app-filter-toggle govuk-!-margin-bottom-6">
+    <% if current_user.data_coordinator? || current_user.support? %>
+      <%= govuk_button_link_to "Invite user", new_user_path(organisation_id: @organisation.id), html: { method: :get } %>
+    <% end %>
+  </div>
```
Dinssa, line 20: The key change was to include the class `.app-filter-toggle`. It could be added to the button group like this or separately with `<div class="app-filter-toggle"><div>` (Although this will not expand on even smaller screens and fill the width like the other buttons without further work). Either way, it must be contained within a `div` with the class `.app-filter-layout`.

RachaelBooth (approved):
Worth explicitly checking with design, I think, on the button spacing - we should be able to get round Rubocop's objection to !important if we want the wider spaced version.

PR #2750, opened 2024-11-05
CLDC-3719: Move sale completion date question to start of setup section

RachaelBooth:


app/models/form/sales/subsections/setup.rb
```diff
@@ -7,10 +7,10 @@ def initialize(id, hsh, section)
 
   def pages
     @pages ||= [
+      Form::Sales::Pages::SaleDate.new(nil, nil, self),
```
RachaelBooth, line 10: We can't make this depend on year, since we can't distinguish years before this is answered.

spec/models/form_spec.rb
```diff
@@ -387,10 +387,6 @@ def answer_local_authority(lettings_log)
       expect(form.sections[0].class).to eq(Form::Sales::Sections::Setup)
       expect(form.subsections.count).to eq(1)
       expect(form.subsections.first.id).to eq("setup")
-      expect(form.pages.count).to eq(13)
-      expect(form.pages.first.id).to eq("owning_organisation")
-      expect(form.questions.count).to eq(14)
-      expect(form.questions.first.id).to eq("owning_organisation_id")
```
RachaelBooth, line 393: These are just testing properties of the sales setup section, which doesn't seem useful here.

Dinssa (approved):

PR #2751, opened 2024-11-05
CLDC-3721: Sales - Move Q78 (Is this a staircasing transaction?) to setup section (25/26)

Dinssa:
The following commits change the name of a method and could be cherry-picked before making changes to questions:
[Rename method 'start_year_after_2024?' to better reflect what it does…](https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/2751/commits/3463fd6a6f1996f6e4264bfd7133fe9b8edbd5d9)
[Remove redundant method](https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/2751/commits/be6f6fb95923c87b3293588d4416dab8a718a99a)

The commits starting from [Move question for 2025/26](https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/2751/commits/daa937fda4e4007d6a541130186a44ac4d3dc4b2) are relevant to this ticket.


app/models/form/sales/pages/staircase.rb
```diff
@@ -2,7 +2,7 @@ class Form::Sales::Pages::Staircase < ::Form::Page
   def initialize(id, hsh, subsection)
     super
     @id = "staircasing"
-    @copy_key = "sales.sale_information.staircasing"
```
RachaelBooth, line 5: I'm surprised by this disappearing?

app/models/form/sales/questions/staircase.rb
```diff
@@ -2,9 +2,9 @@ class Form::Sales::Questions::Staircase < ::Form::Question
   def initialize(id, hsh, page)
     super
     @id = "staircase"
-    @copy_key = "sales.sale_information.staircasing"
+    @copy_key = form.start_year_2025_or_later? ? "sales.setup.staircasing" : "sales.sale_information.staircasing"
```
RachaelBooth: Can this just depend on the subsection id (as it would if it was using the default format) rather than the year? i.e. `"sales.#{subsection.id}.staircasing"`?

spec/models/form/sales/subsections/setup_spec.rb
```diff
@@ -49,6 +50,7 @@
   context "when start year is >= 2024" do
     before do
       allow(section.form).to receive(:start_year_2024_or_later?).and_return(true)
+      allow(section.form).to receive(:start_year_2025_or_later?).and_return(false)
```
RachaelBooth: It would be worth adjusting the context name, since this is now insisting that the start year is 2024.

Is there anything in these blocks it's worth separately testing for 2025 onwards? (There may not be, I've not checked)

spec/models/form/sales/questions/staircase_spec.rb
```diff
@@ -7,6 +7,10 @@
   let(:question_definition) { nil }
   let(:page) { instance_double(Form::Page, subsection: instance_double(Form::Subsection, form: instance_double(Form, start_date: Time.zone.local(2023, 4, 1)))) }
 
+  before do
+    allow(page.subsection.form).to receive(:start_year_2025_or_later?).and_return(false)
+  end
```
RachaelBooth: Given there's now logic on this question around the answer options, I think we should test that the answer options are correct for "before 2025" and "2025 onwards".

app/models/form/sales/questions/staircase.rb
```diff
@@ -2,9 +2,9 @@ class Form::Sales::Questions::Staircase < ::Form::Question
...
     @type = "radio"
-    @answer_options = ANSWER_OPTIONS
+    @answer_options = form.start_year_2025_or_later? ? ANSWER_OPTIONS.except("3") : ANSWER_OPTIONS
```
RachaelBooth: I am wondering if writing this differently would make it easier to tell at a glance which option was no longer available (rather than needing to work out which one 3 is) - I'd be tempted to pull it out into a function and spell things out as we do elsewhere. Not sure how worth it it is given this is relatively simple though.

Dinssa: Ye that makes sense, put it in an answer options function similar to elsewhere

RachaelBooth (approved):

PR #2756, opened 2024-11-05
CLDC-3712: Sales Form Updates

Dinssa:


app/models/form.rb
```diff
@@ -334,7 +334,11 @@ def sales_or_start_year_after_2022?(type, start_year)
...
   end
 
-  def start_year_after_2024?
+  def start_year_2024_or_later?
     start_date && start_date.year >= 2024
   end
+
+  def start_year_2025_or_later?
+    start_date && start_date.year >= 2025
+  end
```
Dinssa, line 343: We can use the `start_year_2025_or_later?` method to makes changes conditional for 2025/26.

RachaelBooth (approved):

PR #2757, opened 2024-11-05
CLDC-3718: Sales - Remove Q10 (Will any buyers live in the property?) (25/26)

Dinssa:


RachaelBooth (commented):
Does anything depend on this at the moment?

Dinssa:
There is the constant below in _app/models/derived_variables/sales_log_variables.rb_ that derives `buy1livein` and `buy2livein` based off this question. But I assume those values would not be needed either. And the first 3 items (everything except `mortgageused`) can be removed? Or just leave it, because it'll still be needed in 2024 and shouldn't matter in 2025? There is nothing else that depends on the value from this question.

```
  DEPENDENCIES = [
    {
      conditions: {
        buylivein: 2,
      },
      derived_values: {
        buy1livein: 2,
      },
    },
    {
      conditions: {
        buylivein: 2,
        jointpur: 1,
      },
      derived_values: {
        buy1livein: 2,
        buy2livein: 2,
      },
    },
    {
      conditions: {
        buylivein: 1,
        jointpur: 2,
      },
      derived_values: {
        buy1livein: 1,
      },
    },
    {
      conditions: {
        mortgageused: 2,
      },
      derived_values: {
        mortgage: 0,
      },
    },
    {
      conditions: {
        mortgageused: 3,
      },
      derived_values: {
        mortgage: nil,
      },
    },
  ].freeze
```

RachaelBooth:
> There is the constant below in _app/models/derived_variables/sales_log_variables.rb_ that derives `buy1livein` and `buy2livein` based off this question. But I assume those values would not be needed either. And the first 3 items (everything except `mortgageused`) can be removed? Or just leave it, because it'll still be needed in 2024 and shouldn't matter in 2025? There is nothing else that depends on the value from this question.

It'll need to stay because it's needed for 2024, just as long as there's nothing that assumes this has been answered but shouldn't anymore - which it doesn't look like there is.

RachaelBooth (approved):

PR #2758, opened 2024-11-05
Fix flaky tests and use a Faker class Lorem

Dinssa:
Using `Faker::Name.name` can result in names with special characters, such as “O’Riley” which we’ve encountered in the scheme factory bot.

**Solution:** Switched to using `Faker::Lorem.characters`, which should return a random combination of characters.

Other options caused test failures due to duplicate elements. For example, `Space.constellation` produced “Leo” twice on a page, and `Lorem.word` resulted in “sed” appearing twice. These duplicates caused tests that search for a term to fail.

To minimise the risk of further clashes, we also now generate the service_name with 16 characters, making it highly improbable to encounter duplicates.

### DIFF
--- a/spec/factories/scheme.rb
+++ b/spec/factories/scheme.rb
@@ -1,6 +1,6 @@
 FactoryBot.define do
   factory :scheme do
-    service_name { Faker::Space.constellation }
+    service_name { Faker::Lorem.word }
     sensitive { Faker::Number.within(range: 0..1) }
     registered_under_care_act { 1 }
     support_type { [0, 2, 3, 4, 5].sample }
--- a/spec/factories/scheme.rb
+++ b/spec/factories/scheme.rb
@@ -1,6 +1,6 @@
 FactoryBot.define do
   factory :scheme do
-    service_name { Faker::Lorem.word }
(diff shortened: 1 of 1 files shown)


RachaelBooth:
I see the problem with duplicates when our tests are looking for things by name, but I'd expect special characters to be fine - what's the problem with them?

Dinssa:
> I see the problem with duplicates when our tests are looking for things by name, but I'd expect special characters to be fine - what's the problem with them?

I think the tests which navigate the to the page and test the title is correct fail because some of the titles would be displayed as `O&#39;Riley` instead of the expected `O'Riley`. You can rerun the tests and it'll likely pass the next time because the name generated does not have an apostrophe.

Example test failing:
Screenshot 2024-11-07 at 16 39 34

RachaelBooth:
> I think the tests which navigate the to the page and test the title is correct fail because some of the titles would be displayed as `O&#39;Riley` instead of the expected `O'Riley`.

Ah, html escaping and all that I take it?

I'd much rather make the tests work with special characters, since they might get used in the system, than avoid them in tests - although I'm willing to put that off to a tech debt ticket if it's not easy to do. What do you reckon?

PR #2759, opened 2024-11-06
CLDC-3066++: Show filters button bug

Dinssa:
This is an add on to this [PR](https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/2748), having noticed a bug.

New changes are from this commit onwards:
[Ensure the app-filter-toggle class is always present regardless of th…](https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/commit/d1f1f0e64d68f22b77606d8f47e763d70e53aba9)

**Change:** Rearranged the hierarchy of the `div` container to ensure the filter toggle is always visible, regardless of user permissions.

**Details:** The class responsible for adding the filter toggle button is now outside the condition that determines whether action buttons (e.g. Invite user) are shown. This ensures that all users, regardless of their permissions, can see the filter toggle button.

**Potential issue to watch:** I’ve conditionally added the `.govuk-!-margin-bottom-6` class to each button group to add extra spacing only when there are action buttons to show. This avoids adding unnecessary spacing for Data Provider users who may not have any actions, such as on the Users page. However, this class won’t be applied when there are no action buttons and only the filter toggle button is shown. In this case, the default spacing is sufficient, especially on smaller screens where the filter toggle only appears.


app/views/organisations/schemes.html.erb
```diff
@@ -22,11 +22,12 @@
...
-    <div class="govuk-button-group app-filter-toggle govuk-!-margin-bottom-6">
-      <%= govuk_button_link_to "Create a new supported housing scheme", new_scheme_path, html: { method: :post } %>
-    </div>
-  <% end %>
+  <% display_actions = SchemePolicy.new(current_user, nil).create? %>
+  <div class="govuk-button-group app-filter-toggle <%= "govuk-!-margin-bottom-6" if display_actions %>">
+    <% if display_actions %>
+        <%= govuk_button_link_to "Create a new supported housing scheme", new_scheme_path, html: { method: :post } %>
+    <% end %>
+  </div>
```
Dinssa, line 30: The variable `display_actions` here and elsewhere allows us to use the same logic to conditionally show the action buttons and add the bottom margin.

RachaelBooth (approved):

PR #2760, opened 2024-11-07
CLDC-3735: Sales - Move Q19 (What type of unit is the property?) (25/26)

Dinssa:
This branch moves the question _What type of unit is the property?_ to before _How many bedrooms does the property have?_. If a service user answers "bedsit" then we'll infer the number of bedrooms is 1 and skip the question.

As its no longer possible for someone to incorrectly choose type as bedsit and number of bedrooms as 1 I've also removed the hint message "A bedsit has 1 bedroom." from the number of bedrooms question.


spec/models/form/sales/subsections/property_information_spec.rb
```diff
@@ -48,6 +49,7 @@
 
       before do
         allow(form).to receive(:start_year_2024_or_later?).and_return(true)
+        allow(form).to receive(:start_year_2025_or_later?).and_return(false)
```
RachaelBooth, line 52: Does this need a 2025 version of the test that changes?

Dinssa: Ye, added one

app/models/sales_log.rb
```diff
@@ -396,6 +396,10 @@ def is_bedsit?
     proptype == 2
   end
 
+  def is_beds_inferred?
+    form.start_year_2025_or_later? && is_bedsit?
+  end
```
RachaelBooth, line 401: Ah, just noticed - I'm expected some code that actively sets beds to 1 if this is true (probably in derived variables, I think that's where it is for lettings?)

Dinssa: Ah yes, that's the missing link, good spot. I'll add that in.

RachaelBooth (approved):

PR #2761, opened 2024-11-08
CLDC-3736: Don't route to property building type or wheelchair accessibility for staircasing sales from 2025

RachaelBooth:


app/models/form/sales/pages/property_wheelchair_accessible.rb
```diff
@@ -2,6 +2,10 @@ class Form::Sales::Pages::PropertyWheelchairAccessible < ::Form::Page
   def initialize(id, hsh, subsection)
     super
     @id = "property_wheelchair_accessible"
+    @depends_on = [
+      { "form.start_year_2025_or_later?" => false },
+      { "is_staircase?" => false },
+    ]
```
RachaelBooth, line 8: There doesn't seem to be a good way to test this other than just checking that depends_on is this object, which in my opinion is fairly pointless...can add them if people would prefer.

kosiakkatrina: Would testing `routed_to?` method work? Or even extract it into `routed_to?` method which runs super and then these conditions?

RachaelBooth: It would only work if I extracted it into a method here (otherwise because form is an instance double in the tests it all falls apart). My initial instinct was that that wasn't worth it, but I don't really have reasoning for that beyond it seems silly not to make use of the depends_on property - what do you reckon?

RachaelBooth: Actually, have nicked how you did it for the buyer questions ticket which works nicely!

kosiakkatrina (approved):

PR #2762, opened 2024-11-08
CLDC-3733: Adjust ordering for working situation question

RachaelBooth:


app/models/form/sales/questions/buyer1_working_situation.rb
```diff
@@ -14,18 +14,35 @@ def initialize(id, hsh, page)
...
-    "3" => { "value" => "In government training into work" },
-    "4" => { "value" => "Jobseeker" },
-    "6" => { "value" => "Not seeking work" },
-    "8" => { "value" => "Unable to work due to long term sick or disability" },
-    "5" => { "value" => "Retired" },
-    "0" => { "value" => "Other" },
-    "10" => { "value" => "Buyer prefers not to say" },
-    "7" => { "value" => "Full-time student" },
-  }.freeze
+  def answer_options
```
RachaelBooth, line 17: I considered doing something to just define the order of keys and grab the values from a single hash, but in the end I thought this was actually clearer.

kosiakkatrina: Definitely cleared like this!

kosiakkatrina (approved):

PR #2763, opened 2024-11-11
CLDC-3734 Add LA in England validation for 2025

kosiakkatrina:


app/models/validations/property_validations.rb
```diff
@@ -57,4 +57,27 @@ def validate_uprn(record)
...
+
+      record.errors.add :la, I18n.t("validations.lettings.property.la.not_in_england")
+      record.errors.add :postcode_full, I18n.t("validations.lettings.property.postcode_full.not_in_england")
+      record.errors.add :uprn, I18n.t("validations.lettings.property.uprn.not_in_england")
+      record.errors.add :uprn_confirmation, I18n.t("validations.lettings.property.uprn_confirmation.not_in_england")
+      record.errors.add :uprn_selection, I18n.t("validations.lettings.property.uprn_selection.not_in_england")
+      record.errors.add :startdate, I18n.t("validations.lettings.property.startdate.not_in_england")
+    elsif record.is_supported_housing?
+      return unless record.location
+      return if record.location.location_code.in?(LocalAuthority.england.pluck(:code))
```
RachaelBooth, line 69: Do we (or should we) validate this when locations are created?

RachaelBooth (approved):

app/models/validations/property_validations.rb
kosiakkatrina: I thought that, but it might be a bit difficult if the locations are used across different years

PR #2764, opened 2024-11-11
CLDC-3743: Allow coordinators to manage schemes for recently absorbed organisations

RachaelBooth:


RachaelBooth:
Aims:
* Data coordinators can view / create / edit / otherwise manage schemes at recently absorbed organisations, and similarly locations at these schemes
* Schemes are automatically deactivated when their organisation is merged
* Everything you'd expect to work given this does

kosiakkatrina (approved):

PR #2765, opened 2024-11-12
CLDC-3737 Route only to buyer questions for staircase transactions

kosiakkatrina:
We only want to route to these questions in household characteristics for 2025 if it's a staircase transaction:
image

### DIFF
--- a/app/models/form/sales/pages/buyer1_ethnic_group.rb
+++ b/app/models/form/sales/pages/buyer1_ethnic_group.rb
@@ -2,19 +2,21 @@ class Form::Sales::Pages::Buyer1EthnicGroup < ::Form::Page
   def initialize(id, hsh, subsection)
     super
     @id = "buyer_1_ethnic_group"
-    @depends_on = [
-      {
-        "buyer_has_seen_privacy_notice?" => true,
-      },
-      {
-        "buyer_not_interviewed?" => true,
-      },
-    ]
   end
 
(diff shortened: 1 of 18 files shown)


app/models/form/sales/pages/buyer1_ethnic_group.rb
```diff
@@ -2,19 +2,21 @@ class Form::Sales::Pages::Buyer1EthnicGroup < ::Form::Page
...
   end
+
+  def routed_to?(log, _current_user)
+    super && page_routed_to?(log)
+  end
+
+  def page_routed_to?(log)
+    return false if log.form.start_year_2025_or_later? && log.is_staircase?
+
+    log.buyer_has_seen_privacy_notice? || log.buyer_not_interviewed?
```
RachaelBooth, line 20: We still care about 2023 routing, I guess? Otherwise I think buyer_has_seen_privacy_notice is forced to be true for the section to be enabled and these could be simplified, but that wasn't true for 2023.

RachaelBooth (approved):

app/models/form/sales/pages/buyer1_ethnic_group.rb
kosiakkatrina: Hmm, good point, I think we care about it while we display 2023 logs, which would be until the end of crossover I think (?)

PR #2768, opened 2024-11-12
Handle flaky failing tests

Dinssa:
This branch uses `CGI.unescapeHTML` instead of `CGI.escapeHTML`, converting HTML entities back to their corresponding characters for accurate testing and matching. Previously, any randomly generated name with an apostrophe would fail in the tests. Now, `CGI.unescapeHTML` ensures that HTML entities in the `expected_title` are converted back to their original characters before comparison. This is crucial for verifying the page title, which may contain special characters.

Unlike in the service, the scheme names here are generated by Faker and not user-submitted data.

### DIFF
--- a/spec/factories/scheme.rb
+++ b/spec/factories/scheme.rb
@@ -1,6 +1,6 @@
 FactoryBot.define do
   factory :scheme do
-    service_name { Faker::Name.name }
+    service_name { "#{Faker::Name.name}'s Housing & Co." }
     sensitive { Faker::Number.within(range: 0..1) }
     registered_under_care_act { 1 }
     support_type { [0, 2, 3, 4, 5].sample }
--- a/spec/requests/locations_controller_spec.rb
+++ b/spec/requests/locations_controller_spec.rb
@@ -180,7 +180,7 @@ RSpec.describe LocationsController, type: :request do
 
     context "when signed in as a data coordinator user" do
       let(:user) { create(:user, :data_coordinator) }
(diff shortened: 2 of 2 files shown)


spec/factories/scheme.rb
```diff
@@ -1,6 +1,6 @@
 FactoryBot.define do
   factory :scheme do
-    service_name { Faker::Name.name }
+    service_name { "#{Faker::Name.name}'s Housing & Co." }
```
Dinssa, line 3: Still generate a random name, but also make sure it works for names with apostrophes and ampersands. Having the `'s Housing & Co.` hardcoded makes sure a developer has to use `CGI.unescapeHTML` when writing another similar test to those highlighted in another comment or it'd fail.

spec/requests/locations_controller_spec.rb
```diff
@@ -215,7 +215,7 @@
       end
 
       it "has correct title" do
-        expected_title = CGI.escapeHTML("#{scheme.service_name} - Submit social housing lettings and sales data (CORE) - GOV.UK")
+        expected_title = CGI.unescapeHTML("#{scheme.service_name} - Submit social housing lettings and sales data (CORE) - GOV.UK")
         expect(page).to have_title(expected_title)
```
Dinssa, line 219: In any future test written checking the title of the scheme might fail unless a developer follows the same approach taken here. Looks like `CGI.escapeHTML` was only used in this spec file.

RachaelBooth (approved):

PR #2774, opened 2024-11-13
CLDC-3738: Sales - Update household characteristics Q29 and Q38 (25/26)

Dinssa:


app/models/form/sales/pages/buyer2_relationship_to_buyer1_yes_no.rb
```diff
@@ -0,0 +1,23 @@
...
+      {
+        "joint_purchase?" => true,
+        "buyer_has_seen_privacy_notice?" => true,
+      },
+      {
+        "joint_purchase?" => true,
+        "buyer_not_interviewed?" => true,
+      },
+    ]
+  end
```
kosiakkatrina: I dont think `buyer_has_seen_privacy_notice` and `buyer_not_interviewed` is needed anymore if this page only exists on 2025. I think it's covered by privacy notice and interview questions being in the setup section

app/models/form/sales/pages/person_relationship_to_buyer_1_yes_no.rb
```diff
@@ -0,0 +1,15 @@
+class Form::Sales::Pages::PersonRelationshipToBuyer1YesNo < ::Form::Sales::Pages::Person
```
kosiakkatrina, line 1: Can there be more than 1 partner?

app/models/form/sales/questions/person_relationship_to_buyer_1_yes_no.rb
```diff
@@ -0,0 +1,32 @@
...
+      "value" => "Prefers not to say",
+    }]
+    @check_answers_card_number = person_index
+    @person_index = person_index
+    @question_number = question_number
+  end
+
+  def question_number
+    base_question_number = case form.start_date.year
+                           when 2023
```
kosiakkatrina: would it ever be 2023?

app/models/form/sales/pages/person_relationship_to_buyer_1_yes_no.rb
Dinssa: It's possible, perhaps we'd want to introduce a soft validation when there are more than one.

app/models/form/sales/questions/person_relationship_to_buyer_1_yes_no.rb
Dinssa: This is left over from when duplicating this question. I imagine we could update this when we have the right question number.

app/models/form/sales/pages/buyer2_relationship_to_buyer1_yes_no.rb
Dinssa: Updated to depend just on joint_purchase? => true

app/models/form/sales/questions/person_relationship_to_buyer_1_yes_no.rb
Dinssa: Updated this to more closely match how we're setting the question numbers elsewhere and just put a temp value for 2025 (same as 2024).

app/models/form/sales/pages/person_relationship_to_buyer_1_yes_no.rb
kosiakkatrina: Just had a look and we already have one!

kosiakkatrina (approved):

PR #2776, opened 2024-11-13
Add merge queue trigger to staging pipeline

RachaelBooth:


.github/workflows/staging_pipeline.yml
```diff
@@ -8,6 +8,7 @@ on:
     types:
       - opened
       - synchronize
+  merge_group:
```
RachaelBooth, line 11: I've deliberately not added this to the review pipeline, because I think once something is in the merge queue it's only the automated tests we want to run.

kosiakkatrina (approved):

PR #2781, opened 2024-11-14
CLDC-3757: Sales - Change values in validation check for Q84 (25/26)

Dinssa:


kosiakkatrina (approved):

app/models/validations/sales/sale_information_validations.rb
```diff
@@ -355,4 +358,11 @@ def over_tolerance?(expected, actual, tolerance, strict: false)
       (expected - actual).abs >= tolerance
     end
   end
+
+  def date_difference_in_years?(start_date, end_date, year)
+    year_difference = (end_date.year - start_date.year).abs
+    same_day_and_month = start_date.day == end_date.day && start_date.month == end_date.month
+
+    (end_date - start_date).abs >= year.years || (same_day_and_month && year_difference == year)
+  end
```
Dinssa: In review we found the dates did not correctly compare and you could enter a date 5 years in the past and not get an error. I suspect this is to do with leap years and the slight difference in the number of days versus what is returned by `5.years`. So I've kept the same logic mostly and added an additional check if the date is the same day and month to just compare the year part.

PR #2785, opened 2024-11-18
CLDC-3700 Update csv downloads

kosiakkatrina:
This PR adds a table that logs each csv download request
Instead of linking to a pre-signed url we link to our service and authenticate 
We might want to extend the expiration time given it's an authenticated link now
This PR also makes csv downloads (except for emails) flow work locally too

### DIFF
--- a/app/jobs/email_csv_job.rb
+++ b/app/jobs/email_csv_job.rb
@@ -22,6 +22,7 @@ class EmailCsvJob < ApplicationJob
 
     storage_service = Storage::S3Service.new(Configuration::EnvConfigurationService.new, ENV["BULK_UPLOAD_BUCKET"])
     storage_service.write_file(filename, BYTE_ORDER_MARK + csv_string)
+    CsvDownload.create!(user:, organisation: user.organisation, filename:, download_type: log_type)
 
     url = storage_service.get_presigned_url(filename, EXPIRATION_TIME)
 
--- a/app/jobs/email_csv_job.rb
+++ b/app/jobs/email_csv_job.rb
@@ -28,7 +28,7 @@ class EmailCsvJob < ApplicationJob
                       end
 
     storage_service.write_file(filename, BYTE_ORDER_MARK + csv_string)
-    csv_download = CsvDownload.create!(user:, organisation: user.organisation, filename:, download_type: log_type)
+    csv_download = CsvDownload.create!(user:, organisation: user.organisation, filename:, download_type: log_type, expiration_time: EXPIRATION_TIME)
 
     url = download_csv_download_path(csv_download.id)
 
--- a/app/jobs/email_csv_job.rb
+++ b/app/jobs/email_csv_job.rb
@@ -1,4 +1,5 @@
 class EmailCsvJob < ApplicationJob
+  include Rails.application.routes.url_helpers
   queue_as :default
 
   BYTE_ORDER_MARK = "\uFEFF".freeze # Required to ensure Excel always reads CSV as UTF-8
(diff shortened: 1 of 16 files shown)


app/policies/csv_download_policy.rb
```diff
@@ -0,0 +1,12 @@
...
+  attr_reader :current_user, :csv_download
+
+  def initialize(current_user, csv_download)
+    @current_user = current_user
+    @csv_download = csv_download
+  end
+
+  def download?
+    @current_user == @csv_download.user || @current_user.support? || @current_user.organisation == @csv_download.organisation
+  end
```
Dinssa, line 15: What do you think about also allowing users from affiliated organisations (to the download organisation) to also be able to download the link? E.g. if a managing agent shares the link with a stock owner.

kosiakkatrina: I feel like those organisations might have access to different org logs because they have different relationships, so it might potentially expose logs that wouldn't otherwise be visible

Dinssa (approved):

app/jobs/email_csv_job.rb
```diff
@@ -20,10 +21,16 @@ def perform(user, search_term = nil, filters = {}, all_orgs = false, organisatio
...
 
-    storage_service = Storage::S3Service.new(Configuration::EnvConfigurationService.new, ENV["BULK_UPLOAD_BUCKET"])
+    storage_service = if FeatureToggle.upload_enabled?
+                        Storage::S3Service.new(Configuration::EnvConfigurationService.new, ENV["BULK_UPLOAD_BUCKET"])
+                      else
+                        Storage::LocalDiskService.new
+                      end
+
     storage_service.write_file(filename, BYTE_ORDER_MARK + csv_string)
+    csv_download = CsvDownload.create!(user:, organisation: user.organisation, filename:, download_type: log_type)
```
RachaelBooth: It's an existing problem, so may be separate, but do we need to think about some error handling here? (if this failed, would the user just never get an email?)

app/models/csv_download.rb
```diff
@@ -0,0 +1,10 @@
+class CsvDownload < ApplicationRecord
+  enum download_type: { lettings: "lettings", sales: "sales", schemes: "schemes", locations: "locations", combined: "combined" }
+
+  belongs_to :user
+  belongs_to :organisation
+
+  def expired?
+    created_at < 24.hours.ago
```
RachaelBooth: Would it be worth setting an expiration time on the record (or a length and comparing that to created_at)? That would mean they could be different for different types of download, and we could enforce they matched what users were told by passing in the same constant when creating the object and when sending the email.

app/jobs/email_csv_job.rb
kosiakkatrina: I think sidekiq would retry the jobs by default (something like up to 25 times), I can't see us disabling that anywhere, so I suspect it would be fine and I assume we would get sentry errors too. But let me know if that's not sufficient

app/models/csv_download.rb
kosiakkatrina: Updated it now!

app/jobs/email_csv_job.rb
```diff
@@ -30,7 +30,8 @@ def perform(user, search_term = nil, filters = {}, all_orgs = false, organisatio
     storage_service.write_file(filename, BYTE_ORDER_MARK + csv_string)
     csv_download = CsvDownload.create!(user:, organisation: user.organisation, filename:, download_type: log_type, expiration_time: EXPIRATION_TIME)
 
-    url = download_csv_download_url(csv_download.id, host: ENV["APP_HOST"])
+    binding.pry
+    url = csv_download_url(csv_download.id, host: ENV["APP_HOST"])
```
Dinssa: Still can't see the page but can download. There's a binding.pry here

kosiakkatrina: Ah, sorry about that... Updated it!

PR #2790, opened 2024-11-18
CLDC-3766: Ensure copy for sales Q78 is found correctly for 2024 logs

RachaelBooth:


RachaelBooth:
Note: this is the fast version to get a fix out ASAP, will do a follow up with some testing/cleaning up other places that should use this instead of the id.

kosiakkatrina (approved):

PR #2793, opened 2024-11-19
CLDC-3760: Sales - firststair and numstair staircasing questions validations

Dinssa:


app/models/validations/sales/sale_information_validations.rb
```diff
@@ -348,6 +356,18 @@ def validate_mortgage_used_dont_know(record)
...
   end
 
+  def validate_number_of_staircase_transactions(record)
+    return unless record.numstair
+
+    if record.firststair == 1 && record.numstair >= 2 || record.firststair != 1 && record.numstair == 1
+      record.errors.add :numstair, I18n.t("validations.sales.sale_information.numstair.must_be_greater_than_one")
+      record.errors.add :firststair, I18n.t("validations.sales.sale_information.firststair.more_than_one_transaction")
+    end
+    if record.numstair > 10
```
kosiakkatrina: Wouldn't this get handled by max? Would it make sense to keep the error message consistent with whatever the numeric validation message is?

Dinssa: Yes it would by default be: `Number of staircasing transactions must be between 2 and 10.`

I can remove this and keep it consistent.

app/models/derived_variables/sales_log_variables.rb
```diff
@@ -75,7 +75,7 @@ def set_derived_fields!
     self.nationality_all = nationality_all_group if nationality_uk_or_prefers_not_to_say?
     self.nationality_all_buyer2 = nationality_all_buyer2_group if nationality2_uk_or_prefers_not_to_say?
 
-    self.numstair = 1 if is_firststair?
+    self.numstair = nil if is_firststair?
```
Dinssa: I changed this to nil, because when changing the response to `firststair` from yes to no, you could not save the form, I think this is because the `numstair` question has a minimum value of 2. However, the acceptance criteria (AC) stated that `numstair` should be inferred as 1 when it is `firststair`. Could we approach this differently?

app/models/validations/sales/sale_information_validations.rb
```diff
@@ -348,6 +356,20 @@ def validate_mortgage_used_dont_know(record)
...
+
+    if record.firststair == 2 && record.numstair < 2
+      record.errors.add :numstair, I18n.t("validations.sales.sale_information.numstair.must_be_greater_than_one")
+      record.errors.add :firststair, I18n.t("validations.sales.sale_information.firststair.cannot_be_no")
+    end
+
+    if record.firststair == 1 && record.numstair > 1
+      record.errors.add :numstair, I18n.t("validations.sales.sale_information.numstair.must_be_one")
+      record.errors.add :firststair, I18n.t("validations.sales.sale_information.firststair.cannot_be_yes")
+    end
```
Dinssa: Is this needed? You can't navigate to the questions to enter the values like this but you could in a bulk upload.

app/models/sales_log.rb
```diff
@@ -561,4 +561,8 @@ def is_resale?
   def is_firststair?
     firststair == 1
   end
+
+  def is_not_firststair?
+    firststair != 1
+  end
```
kosiakkatrina: Is this still used somewhere?

app/models/validations/sales/sale_information_validations.rb
kosiakkatrina: I think it would get inferred for BU as well, we do something similar with `beds` when it's a bedsit

app/models/sales_log.rb
Dinssa: No not anymore, removed it

app/models/validations/sales/sale_information_validations.rb
Dinssa: Okay, removed

kosiakkatrina (approved):

PR #2795, opened 2024-11-20
CLDC-3695 Update docs

kosiakkatrina:


docs/form/page.md
```diff
@@ -10,43 +10,44 @@ Pages sit below the [`Subsection`](subsection) level of a form definition.
...
+      Form::Sales::Questions::Question.new(nil, nil, self),
     ]
-  },
-}
+  end
+end
 ```
 
-In the above example the the subsection has the id `property_postcode`. This id is used for the url of the web page, but the underscore is replaced with a hash, so the url for this page would be `[environment-url]/logs/[log-id]/property-postcode` e.g. on staging this url might look like the following: `https://dluhc-core-staging.london.cloudapps.digital/logs/1234/property-postcode`.
+In the above example the the subsection has the id `property_postcode`. This id is used for the url of the web page, but the underscore is replaced with a hash, so the url for this page would be `[environment-url]/logs/[log-id]/property-postcode` e.g. on staging this url might look like the following: `https://staging.submit-social-housing-data.communities.gov.uk/logs/1234/property-postcode`.
```
Dinssa: Do you mean _dash_ instead of _hash_? in: "but the underscore is replaced with a hash"

kosiakkatrina: yep, good spot!

docs/app_api.md
```diff
@@ -12,4 +12,4 @@ In order to use the app as an API, you will need to configure requests to the AP
   - `Content-Type = application/json`
   - `Action = application/json` N.B. If you use `*/*` instead, the request won't be recognised as an API request`
 
-Currently only the logs controller is configured to accept and authenticate API requests, when the above API environment variables are set.
+Currently only the logs controller is configured to accept and authenticate API requests, when the above API environment variables are set. The API has not been maintained for a long time and may not work as expected.
```
RachaelBooth: Is it worth saying (assuming this is true) that it's not set up on any environment?

RachaelBooth (approved):

PR #2798, opened 2024-11-20
CLDC-3723: Add validation on location units question

Dinssa:
Makes sure the number of units at a location must be 1 or more.

### DIFF
--- a/spec/factories/location.rb
+++ b/spec/factories/location.rb
@@ -22,7 +22,7 @@ FactoryBot.define do
     end
 
     trait :incomplete do
-      mobility_type { nil }
+      units { nil }
       confirmed { false }
     end
 
--- a/spec/factories/location.rb
+++ b/spec/factories/location.rb
@@ -22,7 +22,7 @@ FactoryBot.define do
     end
 
     trait :incomplete do
-      units { nil }
+      mobility_type { nil }
       confirmed { false }
     end
 
--- a/spec/models/location_spec.rb
+++ b/spec/models/location_spec.rb
@@ -740,10 +740,38 @@ RSpec.describe Location, type: :model do
   describe "#units" do
     let(:location) { FactoryBot.build(:location) }
 
-    it "does add an error when the number of units is invalid" do
-      location.units = nil
-      location.valid?(:units)
-      expect(location.errors.count).to eq(2)
+    context "when the number of units is invalid" do
+      it "adds an error when units is nil" do
+        location.units = nil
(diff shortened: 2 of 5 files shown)


spec/factories/location.rb
```diff
@@ -22,7 +22,7 @@
     end
 
     trait :incomplete do
-      units { nil }
+      mobility_type { nil }
       confirmed { false }
     end
```
Dinssa: Use a different attribute in the location factory to make it incomplete

kosiakkatrina: do we need to? I mean Idm which attribute it is, just curious why it needs changing

spec/models/location_spec.rb
```diff
@@ -743,7 +743,7 @@
     it "does add an error when the number of units is invalid" do
       location.units = nil
       location.valid?(:units)
-      expect(location.errors.count).to eq(1)
+      expect(location.errors.count).to eq(2)
```
kosiakkatrina: Could we also add a tests specifically for the number range?

spec/models/validations/setup_validations_spec.rb
```diff
@@ -704,7 +704,7 @@
       end
 
       it "produces no error when location is completes" do
```
kosiakkatrina: there's a typo in "completes", not your change but if you can update it that'd be good

spec/factories/location.rb
Dinssa: No we don't I've changed it back, early on a test would fail because of this but not anymore.

spec/models/location_spec.rb
Dinssa: Yup added, a few to test invalid and valid values around the boundary (greater than or equal to 1) and others.

spec/models/location_spec.rb
```diff
@@ -740,10 +740,38 @@
...
+      it "adds an error when units is nil" do
+        location.units = nil
+        location.valid?(:units)
+        expect(location.errors.count).to eq(2)
+      end
+
+      it "adds an error when units is 0" do
+        location.units = 0
+        location.valid?(:units)
+        expect(location.errors.count).to eq(2)
```
kosiakkatrina: I think this should be 1, cause the presence error shouldn't trigger

Dinssa: Yup, changed

kosiakkatrina (approved):

PR #2801, opened 2024-11-20
CLDC-3344: Sales CSV download ecstat2 child label bug fix

Dinssa:
We had a bug where instead of the user friendly label to explain what is meant when `ecstat2 = 9` the value itself was returned in CSV downloads, unlike how it should appear (look to `ecstat3`).
Screenshot 2024-11-20 at 18 29 12

This bug only appears on person 2's working situation question (`ecstat2`) and only for when the value is `9` as the matching key-value pair in the answer options was missing. But was present on the generic working situation question for all people from person3, person4....onwards.

**Result:**
Screenshot 2024-11-20 at 18 32 20


app/models/derived_variables/sales_log_variables.rb
```diff
@@ -160,6 +160,8 @@ def child_under_16_constraints!
     (start_index..6).each do |idx|
       if age_under_16?(idx)
         self["ecstat#{idx}"] = 9
+      elsif self["ecstat#{idx}"] == 9
+        self["ecstat#{idx}"] = nil
       end
     end
```
Dinssa: I've added this so a user can change the age of a person that was first set as under 16 to another age above 16. At the moment should they want to do that they would get an error:
_Answer cannot be over 16 as person’s 2 working situation is ‘child under 16‘._

As the working situation question for a child is not one they can access they'd get stuck. Now should they change the age e.g. to 17, the next question will be to choose the working situation.

app/models/form/sales/questions/buyer2_working_situation.rb
```diff
@@ -26,6 +26,15 @@ def answer_options
         "6" => { "value" => "Not seeking work" },
         "7" => { "value" => "Full-time student" },
         "8" => { "value" => "Unable to work due to long term sick or disability" },
+        "9" => {
+          "value" => "Child under 16",
+          "depends_on" => [
+            { "saledate" => { "operator" => "<", "operand" => Time.zone.local(2024, 4, 1) } },
```
kosiakkatrina: If `form.start_year_2025_or_later?` the saledate can't be before 2024?

Maybe it's worth having answer_options and displayed_answer_options as we do on some other questions?

app/models/derived_variables/sales_log_variables.rb
kosiakkatrina: Can we add some tests for this to make sure it doesn't clear it when we don't want it to?

Dinssa: Sure, added

app/models/form/sales/questions/buyer2_working_situation.rb
```diff
@@ -26,6 +30,7 @@ def answer_options
         "6" => { "value" => "Not seeking work" },
         "7" => { "value" => "Full-time student" },
         "8" => { "value" => "Unable to work due to long term sick or disability" },
+        "9" => { "value" => "Child under 16" },
         "0" => { "value" => "Other" },
         "10" => { "value" => "Buyer prefers not to say" },
       }.freeze
```
Dinssa, line 36: Since we are not displaying the key “9” in 2024 or beyond, I added the value but omitted the depends_on conditions found on the generic person working situation page's answer options.

kosiakkatrina (approved):

PR #2803, opened 2024-11-21
CLDC-3768: Call process_postcode_changes when setting postcode_full from postcode_full_input

RachaelBooth:


app/models/log.rb
```diff
@@ -111,6 +111,7 @@ def process_address_change!
         self.town_or_city = nil
         self.county = nil
         self.postcode_full = postcode_full_input
+        process_postcode_changes!
```
kosiakkatrina, line 114: I ended up with the same solution here! Nice! https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/2805/files

kosiakkatrina (approved):

PR #2805, opened 2024-11-22
CLDC-3294 Add addresses flow feature tests and update the flow a bit

kosiakkatrina:


app/models/derived_variables/lettings_log_variables.rb
```diff
@@ -84,6 +84,15 @@ def set_derived_fields!
...
       self.uprn = nil
+      if uprn_known_was == 1
+        self.address_line1 = nil
+        self.address_line2 = nil
+        self.town_or_city = nil
+        self.county = nil
+        self.postcode_known = nil
+        self.postcode_full = nil
+        self.la = nil
+      end
```
kosiakkatrina, line 95: If the uprn changed from known and selected to not known I think it should go back to what the values would have been if not know was selected to begin with

app/models/log.rb
```diff
@@ -111,6 +111,7 @@ def process_address_change!
         self.town_or_city = nil
         self.county = nil
         self.postcode_full = postcode_full_input
+        process_postcode_changes!
```
kosiakkatrina: we want to infer the la from the newly set postcode

app/models/log.rb
```diff
@@ -125,9 +126,11 @@ def address_string
   end
 
   def address_options
-    return @address_options if @address_options
+    return @address_options if @address_options && @last_searched_address_string == address_string
```
kosiakkatrina, line 129: Sometimes the address input fields change and we don't get the new address options in time. So previously this situation would happen:
1. Search for address that returns some options
2. Land on upon-selection page
3. Go back to address search page and enter a query that wouldn't return any options
4. Next page is no-address-found, but because we still have old @address_options set, this page would not get routed to, so the flow would exit into the tasklist page

RachaelBooth (approved):

PR #2818, opened 2024-11-25
CLDC-3748 Check org name uniqueness

kosiakkatrina:


app/controllers/organisations_controller.rb
```diff
@@ -155,6 +155,7 @@ def update
         end
         redirect_to details_organisation_path(@organisation)
       else
+        @used_rent_periods = @organisation.lettings_logs.pluck(:period).uniq.compact.map(&:to_s)
```
kosiakkatrina, line 158: This is irrelevant, it's failing now if we try to update the org details to something invalid (like an empty name)

app/models/organisation.rb
```diff
@@ -58,6 +58,7 @@ def affiliated_stock_owners
   alias_method :la?, :LA?
 
   validates :name, presence: { message: I18n.t("validations.organisation.name_missing") }
+  validates :name, uniqueness: { case_sensitive: false, message: I18n.t("validations.organisation.name_not_unique") }
```
RachaelBooth, line 61: Should we be able to use a name that had been used by an organisation that is now deleted?

I suspect the answer is no, but that might be a bit weird if users get this message but can't find an org with that name.

spec/factories/organisation.rb
```diff
@@ -1,6 +1,6 @@
 FactoryBot.define do
   factory :organisation do
-    name { "MHCLG" }
+    name { "MHCLG#{SecureRandom.hex}" }
```
RachaelBooth: Might start hitting conflicts / overlaps with some of the things I did in https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/2834 (although some of the test fixes assumed we could call many orgs the same thing) - suggest merging main in ASAP to catch anything weird going on.

spec/features/lettings_log_spec.rb
```diff
@@ -270,6 +270,10 @@
     context "when visiting a subsection check answers page" do
       let(:lettings_log) { create(:lettings_log, :setup_completed) }
 
+      before do
+        lettings_log.owning_organisation.update!(name: "MHCLG")
```
RachaelBooth: Would it be better (fewer database updates) to do `let(:lettings_log) { create(:lettings_log, :setup_completed, owning_organisation: create(:organisation, name: "MHCLG")) }`?

Or even better make the test just use the name from whatever was created rather than comparing to a hardcoded string.

app/models/organisation.rb
kosiakkatrina: I see what you mean, from what I understand we don't have duplicates atm anyway and if we ever have a requirement that an org can be reinstated it might cause issues if we don't check deleted org names.

I don't think this would be a common scenario. Let me know if you disagree tho

spec/features/lettings_log_spec.rb
```diff
@@ -286,6 +286,10 @@
     context "when reviewing a complete log" do
       let(:lettings_log) { create(:lettings_log, :completed) }
 
+      before do
+        lettings_log.owning_organisation.update!(name: "MHCLG")
```
RachaelBooth: It looks like the test doesn't actually need this - does it work to remove it? (might be the same for the next few files too)

kosiakkatrina: Ah, yep, missed them!

RachaelBooth (approved):

PR #2826, opened 2024-11-26
CLDC-3769 Fix some check errors flows

kosiakkatrina:


app/controllers/form_controller.rb
```diff
@@ -100,9 +100,10 @@ def restore_error_field_values(questions)
     return unless questions
 
     questions.each do |question|
-      if question&.type == "date" && @log.attributes.key?(question.id)
-        @log[question.id] = @log.send("#{question.id}_was")
-      end
+      next unless question&.type == "date" && @log.attributes.key?(question.id)
+      next unless @log[question.id]&.year&.zero?
+
+      @log[question.id] = @log.send("#{question.id}_was")
```
kosiakkatrina: The reason we reset the date to what it was is because when it's an invalid date format, it gets reset to day: 1, month: 0, year: 0

We don't have to reset it if the format is valid to begin with, which this does. 

The reason this is relevant for compound validations is because when we click "see all related answers" we resubmit the form with whatever is entered for the questions on the current page in order to get all the validation errors.

If the date gets reset, we end up resubmitting a valid date, so there are no longer any validation errors and the list with error fields shows up empty.

app/controllers/form_controller.rb
```diff
@@ -419,7 +420,7 @@ def render_check_errors_page
       @log.valid?
       @log.reload
       error_attributes = @log.errors.map(&:attribute)
-      @questions = @log.form.questions.select { |q| error_attributes.include?(q.id.to_sym) }
+      @questions = @log.form.questions.select { |q| error_attributes.include?(q.id.to_sym) && q.page.routed_to?(@log, current_user) }
```
kosiakkatrina, line 441: Sometimes questions have 2 versions defined, but only of of them routed to, and we end up showing 2 rows of the same question on compound validations page

app/models/form/sales/pages/owning_organisation.rb
```diff
@@ -20,11 +20,13 @@ def routed_to?(log, current_user)
     if current_user.organisation.holds_own_stock?
       return true if current_user.organisation.absorbed_organisations.any?(&:holds_own_stock?)
       return true if stock_owners.count >= 1
+      return false if log.owning_organisation == current_user.organisation
 
       log.update!(owning_organisation: current_user.organisation)
```
kosiakkatrina, line 25: This feels like an irrelevant update if it's already set to that (was failing some tests)

app/models/validations/property_validations.rb
```diff
@@ -41,6 +41,8 @@ def validate_uprn(record)
 
   def validate_property_postcode(record)
     postcode = record.postcode_full
+    return unless postcode
```
kosiakkatrina, line 44: When we try to clear the postcode as part of the "clear all" for compound validations, we hit this validation.

The validation will still trigger fine when nothing is entered from the single log submission flow, cause postcode would come through as an empty string instead of a nil

RachaelBooth (approved):

PR #2829, opened 2024-11-27
CLDC-3335 Adjust BU errors for addresses

kosiakkatrina:
If there are issues when setting address from BU there are often several ways they can be fixed, but in the report we display errors on all the relevant fields and require the user to answer all of them. Updated error messages should make it clearer which fields need answering.


config/locales/validations/lettings/2024/bulk_upload.en.yml
```diff
@@ -50,6 +50,8 @@ en:
             invalid: "Age of person %{person_num} must be a number or the letter R"
           address:
             not_found: "We could not find this address. Check the address data in your CSV file is correct and complete, or select the correct address using the CORE site."
+            not_determined: "There are multiple matches for this address, you can choose to correct the log in the service and select the address manually or answer correct UPRN in the csv."
+            not_answered: "Enter either the URPR or the address."
```
RachaelBooth: Typo here for UPRN? (also same below)

RachaelBooth (approved):

config/locales/validations/lettings/2024/bulk_upload.en.yml
kosiakkatrina: Thanks!

PR #2836, opened 2024-11-27
CLDC-3740: Replace you didn't answer text with link to question

Dinssa:
This PR aims to change how we show users that a question needs to be answered from a text and "Change" button like this...
Screenshot 2024-12-16 at 19 36 04

... to now present a link with an appropriate prompt to answer the question:
Screenshot 2024-12-16 at 19 37 06

## Features
By default, it generates the link text for log questions based on their question type. Support users can customise this text by setting a value in the translation file for each question’s `check_answer_prompt` (it will have to be the full link text, not just the prefix). If no custom text is provided, it will add a prefix to the first available option in the following order: `error_label`, `check_answer_label`, `header`, or `id`. Note that `question_text` is intentionally excluded from this list. For questions with types checkbox, radio, or select, the default prefix is “Select”; for date types, it is “Set”; and for all other types, it is “Enter”.

Screenshot 2024-12-16 at 19 33 44

In this example we have a few custom set link text:

- "Enter UPRN if known" - This page is divided into two sections. First, it asks if you know the UPRN value. If you answer “Yes,” you will be prompted to enter the UPRN on the same page. I changed the prompt from a first draft of “Answer if UPRN known” to “Enter UPRN if known” to better reflect this two-part process.
- "Try find address" - As the check answer label here is not from a specific question but from a multi question page and without a custom `check_answer_prompt` would default to that of the first question "Address line 1"
- "Enter address line 1 and 2" - Although this looks like it is just the check answer label plus a default prefix it would have also defaulted to the first question on the page "Address line 1" and just said "Enter address line 1".
- "Answer if it's the first time being let as social-housing" - This would have by default been "Select first time being let as social-housing" as the question is a radio type. That didn't seem right so I've put in a draft custom check_answer_prompt as such. _The same for "Answer if property is built or adapted to wheelchair-user standards"._

## Non log pages
Unlike for logs where we have many questions, which can be changed by support users, other pages are set by us and as such I have for the most part written the link texts per attribute, with a default "Enter" prefix should anything change in the future but is in most places unused.

### Schemes example
Screenshot 2024-12-16 at 19 56 46

For some of these pages, it is not possible to create a resource without answering at least some of the questions. For example, when creating a scheme, you cannot complete the step and progress to the next page without answering the visible questions shown in the screenshot above. Therefore, I have not changed the behavior of those questions. The changes only apply to the questions you could skip in the initial stages.

## No permission to change values
We only need to show the link to enter a value if the user has the necessary permissions to make those changes. Therefore, it is still necessary to display a message indicating that the value has not been entered. The example below shows the view of a data provider when viewing an incomplete scheme.

Screenshot 2024-12-16 at 20 02 16


app/frontend/styles/_red-link.scss
```diff
@@ -0,0 +1,14 @@
...
+}
+
+.app-red-link:link,
+.app-red-link---no-visited-state:visited {
+  color: govuk-colour("red");
+}
+
+.app-red-link:hover {
+  color: govuk-shade(govuk-colour("red"), 20);
+}
```
Dinssa, line 14: This darkens the colour by adding 20% more black.

app/helpers/details_table_helper.rb
```diff
@@ -1,14 +1,42 @@
...
+    resource_class = resource.class.name
+
     if attribute[:format] == :bullet && attribute[:value].length > 1
       list = attribute[:value].map { |value| "<li>#{value}</li>" }.join
       simple_format(list, { class: "govuk-list govuk-list--bullet" }, wrapper_tag: "ul")
     else
-      return simple_format(attribute[:value].first.to_s, { class: "govuk-body" }, wrapper_tag: "p") if attribute[:value].is_a?(Array)
-
-      value = attribute[:value].presence || "<span class=\"app-!-colour-muted\">You didn’t answer this question</span>".html_safe
+      return simple_format(attribute[:value].first.to_s, { class: "govuk-body" }, wrapper_tag: "p") if attribute[:value].is_a?(Array) && attribute[:value].any?
```
Dinssa, line 7: Added the `attribute[:value].any?` to ensure that the value is not only an Array but also contains at least one element, preventing the code from attempting to access the first element of an empty array which would cause an error.

app/helpers/formatting_helper.rb
```diff
@@ -17,7 +16,7 @@ def ensure_sentence_ending(text)
   end
 
   def lowercase_first_letter(text)
-    return text if text.blank?
+    return text if text.blank? || text.split.first == text.split.first.upcase
```
Dinssa, line 20: Changed this to not lowercase the first letter if the entire first word is capitalised. Previously "UPRN known" would have been changed to "uPRN known".

app/helpers/locations_helper.rb
```diff
@@ -64,8 +64,25 @@ def location_edit_path(location, attribute)
...
+
+    paths = {
+      "postcode" => scheme_location_postcode_path(scheme, location, referrer: "details"),
+      "name" => scheme_location_name_path(scheme, location, referrer: "details"),
+      "units" => scheme_location_units_path(scheme, location, referrer: "details"),
+      "type_of_unit" => scheme_location_type_of_unit_path(scheme, location, referrer: "details"),
+      "mobility_standards" => scheme_location_mobility_standards_path(scheme, location, referrer: "details"),
+    }
+
+    paths[attr[:attribute]]
```
Dinssa, line 89: Moved most of this logic from the view.

app/helpers/merge_requests_helper.rb
```diff
@@ -2,10 +2,28 @@ module MergeRequestsHelper
...
+
+  def merge_request_details_prompt(page)
+    messages = {
+      "existing_absorbing_organisation" => "Answer if absorbing organisation is already active",
+      "helpdesk_ticket" => "Enter helpdesk ticket number",
+      "merging_organisations" => "Add merging organisations",
+      "merge_date" => "Set merge date",
+    }
+    messages[page] || "Enter #{lowercase_first_letter(page.humanize)}"
+  end
```
Dinssa: Unlike elsewhere in the app it was not the best solution to create the link text by adding a prefix infront of the page id and was best to custom write the entire message.

app/models/form/question.rb
```diff
@@ -58,6 +59,10 @@ def check_answer_label
     @check_answer_label ||= I18n.t("forms.#{form.start_date.year}.#{copy_key}.check_answer_label", default: "")
   end
 
+  def check_answer_prompt
+    @check_answer_prompt || I18n.t("forms.#{form.start_date.year}.#{copy_key}.check_answer_prompt", default: nil).presence || generate_check_answer_prompt
+  end
```
Dinssa, line 64: Added this property to the question model and translation files. If a value is set in the question, it will use that. Otherwise, it will use any value set in the translation file. Setting `default: nil` in the `I18n.t` method call ensures that if the translation is not found it will return `nil` instead of returning a fallback string. This ensures if neither is available, it will generate a default check answer prompt (link text).

config/locales/forms/2023/lettings/property_information.en.yml
```diff
@@ -6,99 +6,117 @@ en:
...
+            check_answer_prompt: ""
             hint_text: ""
             question_text: "What is the property’s local authority?"
 
           first_time_property_let_as_social_housing:
             page_header: ""
-            check_answer_label: "First time being let as social-housing?"
+            check_answer_label: "First time being let as social-housing"
+            check_answer_prompt: "Answer if it's the first time being let as social-housing"
             hint_text: ""
```
Dinssa: The `check_answer_prompt` text is subject to change after review by the content designer, as are some of the `check_answer_label` texts. Following my last discussion with the content designer, we are going to rewrite the check answer labels to avoid needing a question mark at the end.

app/helpers/details_table_helper.rb
```diff
@@ -1,14 +1,42 @@
...
+                                  when "Location"
+                                    location_value(attribute, resource)
+                                  when "Organisation"
+                                    organisation_value(attribute, resource)
+                                  when "Scheme"
+                                    scheme_value(attribute, resource)
+                                  else
+                                    "<span class=\"app-!-colour-muted\">No answer provided</span>".html_safe
+                                  end
+  end
```
kosiakkatrina: `attribute[:value].presence || send("#{resource_class.downcase}_value", attribute, resource) || "<span class=\"app-!-colour-muted\">No answer provided</span>".html_safe`

or early returns

kosiakkatrina: And maybe we can remove `No answer provided` from all the other methods

app/views/locations/check_answers.html.erb
```diff
@@ -23,14 +23,14 @@
...
+                <%= details_html(attr, @location) %>
                 <span class="govuk-!-font-weight-regular app-!-colour-muted"><%= formatted_local_authority_timeline(@location) %></span>
                 <% end %>
             <% else %>
-              <% row.with_value { details_html(attr) } %>
+              <% row.with_value { details_html(attr, @location) if user_can_edit_scheme?(current_user, @scheme) } %>
             <% end %>
             <% if LocationPolicy.new(current_user, @location).update? %>
-              <% row.with_action(text: action_text_helper(attr, @location), href: location_edit_path(@location, attr[:attribute])) %>
+              <% row.with_action(text: location_action_text_helper(attr, @location), href: location_edit_path(@location, attr[:attribute])) %>
```
kosiakkatrina: This is a confusing name for a method. Does it need `helper` at the end of it?

app/views/locations/show.html.erb
```diff
@@ -27,18 +27,18 @@
...
-              <% row.with_action(text: "Change", href: scheme_location_name_path(@scheme, @location, referrer: "details")) if attr[:attribute] == "name" %>
-              <% row.with_action(text: "Change", href: scheme_location_units_path(@scheme, @location, referrer: "details")) if attr[:attribute] == "units" && current_user.support? %>
-              <% row.with_action(text: "Change", href: scheme_location_type_of_unit_path(@scheme, @location, referrer: "details")) if attr[:attribute] == "type_of_unit" && current_user.support? %>
-              <% row.with_action(text: "Change", href: scheme_location_mobility_standards_path(@scheme, @location, referrer: "details")) if attr[:attribute] == "mobility_standards" && current_user.support? %>
+              <% %w[postcode name units type_of_unit mobility_standards].each do |attribute| %>
+                <% if attr[:attribute] == attribute %>
+                  <%= row.with_action(text: "Change", href: location_action_link(attr, @scheme, @location, current_user)) %>
+                <% end %>
+              <% end %>
```
kosiakkatrina: I think we end up looping the same amount of times, but is this at all clearer?`<%= row.with_action(text: "Change", href: location_action_link(attr, @scheme, @location, current_user)) if list_of_attributes.include?(attributes[:attribute])%>`

app/views/organisations/show.html.erb
```diff
@@ -22,8 +22,9 @@
           <% if can_edit_org?(current_user) && attr[:editable] %>
             <%= summary_list.with_row do |row| %>
               <% row.with_key { attr[:name] } %>
-              <% row.with_value { details_html(attr) } %>
+              <% row.with_value { details_html(attr, @organisation) } %>
               <% row.with_action(
+                text: organisation_action_text_helper(attr, @organisation),
```
kosiakkatrina: do we need `helper` in the method name?

app/helpers/schemes_helper.rb
```diff
@@ -79,6 +79,12 @@ def change_answer_link(scheme, question_id, user)
     end
   end
 
+  def change_link_text(question_id, scheme)
+    return "" if scheme.public_send(question_id).nil?
```
kosiakkatrina, line 83: If the text is "" do we still have the action and how would it impact screen readers?

app/views/locations/check_answers.html.erb
Dinssa: No I can remove it, main thing was that it's name was unique.

app/helpers/details_table_helper.rb
Dinssa: Nice i'll try that. ~~I can try remove `No answer provided` from all places regarding logs but the other pages (e.g. Users, Schemes...) might need to remain separate.~~ Removed from the other methods in the same helper module.

app/views/organisations/show.html.erb
Dinssa: Will remove it

app/helpers/schemes_helper.rb
Dinssa: Good point, the action would still be there. For screen readers it seems the link would still be read out, so I'll work to remove it.

Dinssa: I've removed the action link entirely unless a value is present. This also solved the issue with the "Change" link for support users on the first questions in a log.

app/views/locations/check_answers.html.erb
```diff
@@ -29,8 +29,10 @@
             <% else %>
               <% row.with_value { details_html(attr, @location) if user_can_edit_scheme?(current_user, @scheme) } %>
             <% end %>
-            <% if LocationPolicy.new(current_user, @location).update? %>
-              <% row.with_action(text: location_action_text_helper(attr, @location), href: location_edit_path(@location, attr[:attribute])) %>
+            <% if LocationPolicy.new(current_user, @location).update? && attr[:value] %>
+              <% row.with_action(text: location_action_text(attr, @location), href: location_edit_path(@location, attr[:attribute])) %>
+            <% else %>
+              <% row.with_action %>
```
kosiakkatrina, line 35: Does this need an else?

app/views/locations/show.html.erb
Dinssa: Yup good, idea, I've put this in.

Dinssa: `<%= row.with_action(text: "Change", href: location_action_link(attr, @scheme, @location, current_user)) if location_editable_attributes.include?(attr[:attribute]) %>`


With a method in the helper file that returns the list of attributes (`location_editable_attributes`)

app/views/locations/check_answers.html.erb
Dinssa: Oh so, this is just to place something in place of no action. It's done elsewhere in similar tables with action rows, but not everywhere. It's not needed at all but I thought might be worth adding to keep the structure the same. Again, makes no difference, can also just be removed.

kosiakkatrina: Whatever makes sense, if we do it elsewhere it might be fine. Up to you!

kosiakkatrina (approved):

PR #2838, opened 2024-11-28
CLDC-2141 Update form submit flow

kosiakkatrina:
Update the submit form flow to follow post - redirect - get instead of rendering on post, which should help make navigation smoother.
We now restore the errors and log question answers on show_page instead of submit.

The initial issue this is solving:
1. Make a question error
2. Answer question correctly, save and continue
3. Click browser back - this would resubmit with and display the error again, now it should just call a get

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,7 +10,7 @@ RUN apk add --update --no-cache tzdata && \
 # build-base: compilation tools for bundle
 # yarn: node package manager
 # postgresql-dev: postgres driver and libraries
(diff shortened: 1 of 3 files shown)


Dinssa (approved):

Dinssa (commented):
You can still use the browser back button to see the old error message, but the new correct answer will be displayed. If you refresh the page, the error will disappear as expected. This is an improvement over the previous behaviour, where you were prompted to resubmit the form, and the old value remained on the question.

PR #2839, opened 2024-11-28
CLDC-2128: Validate supplied year in bulk upload pages

RachaelBooth:


spec/requests/bulk_upload_lettings_logs_controller_spec.rb
```diff
@@ -73,5 +73,116 @@
...
+
+      context "when requesting the current year" do
+        it "succeeds" do
+          get "/lettings-logs/bulk-upload-logs/#{page_id}?form%5Byear%5D=#{current_collection_start_year}"
+
+          expect(response).to be_ok
+        end
+      end
+
+      if page_id != "prepare-your-file"
```
RachaelBooth, line 145: The prepare your file page has a year based switch statement for the view rather than just passing in the year, so we can expect this test to fail for that page whenever the next year hasn't been set up yet - which means we really don't want this test in that case.

app/controllers/bulk_upload_sales_logs_controller.rb
```diff
@@ -24,12 +25,26 @@ def update
...
     redirect_to sales_logs_path
   end
 
+  def validate_year!
+    return if params[:id] == "year"
+    return if params[:id] == "guidance" && params.dig(:form, :year).nil?
+
+    allowed_years = [current_year]
+    allowed_years.push(current_year - 1) if FormHandler.instance.sales_in_crossover_period?
+    allowed_years.push(current_year + 1) if FeatureToggle.allow_future_form_use?
```
RachaelBooth, line 40: We'd never have set up anything for years further ahead than this, so allowing just the one seems reasonable.

app/controllers/bulk_upload_lettings_logs_controller.rb
```diff
@@ -1,6 +1,7 @@
 class BulkUploadLettingsLogsController < ApplicationController
   before_action :authenticate_user!
-  before_action :validate_data_protection_agrement_signed!
+  before_action :validate_data_protection_agreement_signed!
```
kosiakkatrina, line 3: nice

app/controllers/bulk_upload_sales_logs_controller.rb
kosiakkatrina: Would it be useful extracting it into a FormHandler method at all? Something similar to `years_of_available_sales_forms`?

kosiakkatrina (approved):

PR #2849, opened 2024-12-04
Refactor exports to have years

kosiakkatrina:
We will want to start exporting sales collections, which would also be yearly, so it makes sense to have explicit year attached to the export

### DIFF
--- a/app/services/exports/xml_export_service.rb
+++ b/app/services/exports/xml_export_service.rb
@@ -11,9 +11,9 @@ module Exports
 
   private
 
-    def build_export_run(collection, base_number, full_update)
-      @logger.info("Building export run for #{collection}")
-      previous_exports_with_data = Export.where(collection:, empty_export: false)
+    def build_export_run(collection, base_number, full_update, year = nil)
+      @logger.info("Building export run for #{[collection, year].join(' ')}")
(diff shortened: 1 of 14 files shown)


app/services/exports/xml_export_service.rb
```diff
@@ -25,16 +25,16 @@ def build_export_run(collection, base_number, full_update)
...
       end
 
-      Export.new(collection:, started_at: @start_time, base_number:, increment_number:)
+      Export.new(collection:, year:, started_at: @start_time, base_number:, increment_number:)
     end
 
-    def write_export_archive(export, collection, recent_export, full_update)
-      archive = get_archive_name(collection, export.base_number, export.increment_number) # archive name would be the same for all logs because they're already filtered by year (?)
+    def write_export_archive(export, year, recent_export, full_update)
+      archive = get_archive_name(year, export.base_number, export.increment_number) # archive name would be the same for all logs because they're already filtered by year (?)
```
RachaelBooth: Is the comment still useful?

RachaelBooth (approved):

app/services/exports/xml_export_service.rb
kosiakkatrina: I don't understand what it means, so probably not

PR #2851, opened 2024-12-04
CLDC-3646 Do not update status for pending logs

kosiakkatrina:


app/models/log.rb
```diff
@@ -315,9 +315,11 @@ def plural_gender_for_person(person_num)
...
 
   def update_status!
-    return if skip_update_status
-
-    self.status = calculate_status
+    if status == "pending"
+      self.status_cache = calculate_status
+    else
+      self.status = calculate_status
+    end
```
kosiakkatrina, line 324: I can't see why we would ever not want to recalculate status_cache when updating a pending log, but let me know if there is such scenario

app/services/bulk_upload/lettings/log_creator.rb
```diff
@@ -16,9 +16,7 @@ def call
       row_parser.log.blank_invalid_non_setup_fields!
       row_parser.log.bulk_upload = bulk_upload
       row_parser.log.creation_method = "bulk upload"
-      row_parser.log.skip_update_status = true
       row_parser.log.status = "pending"
-      row_parser.log.status_cache = row_parser.log.calculate_status
```
kosiakkatrina, line 21: This should calculate on log.save! anyway

app/models/bulk_upload.rb
```diff
@@ -116,7 +116,6 @@ def prefix_namespace
 
   def unpend
     logs.find_each do |log|
-      log.skip_update_status = true
```
RachaelBooth: We might still want to be able to skip in this case, otherwise we'll be recalculating the status and may as well not have cached it.

On the other hand, maybe it's good to be recalculating in case e.g. some validation has changed.

I'm not really sure here - any thoughts?

kosiakkatrina: I think recalculating makes it more accurate and we would avoid such bugs where something is marked completed, but actually isn't. But I think the issue would be speed, cause I don't think we put this on a queue, I'll update it and will add this back in

kosiakkatrina: Updated now!

RachaelBooth (approved):

PR #2852, opened 2024-12-04
CLDC-3268: Update github action dependencies

RachaelBooth:


.github/workflows/aws_deploy.yml
```diff
@@ -41,19 +41,17 @@ jobs:
...
         with:
           aws-region: ${{ env.aws_region }}
           role-to-assume: ${{ env.app_repo_role }}
 
       - name: Login to Amazon ECR
         id: ecr-login
-        uses: aws-actions/amazon-ecr-login@v1
-        with:
-          mask-password: "true"
+        uses: aws-actions/amazon-ecr-login@v2
```
kosiakkatrina, line 54: Is mask password now default?

RachaelBooth: Yes it is

kosiakkatrina (approved):

PR #2853, opened 2024-12-05
Separate services tests to improve pipeline times

RachaelBooth:
This doesn't make as big a difference as I'd hoped (in most cases there weren't clear ways to actively speed up the tests themselves), but it does push it back under 10 minutes for tests.


.github/workflows/run_tests.yml
```diff
@@ -0,0 +1,411 @@
+name: Run Tests
```
RachaelBooth, line 1: This is pulled out of the staging pipeline, with the addition of the separation of the tests in spec/services.

PRs and the merge queue now run these tests directly instead of the staging pipeline having conditions about whether to do the deployment.

.github/workflows/staging_pipeline.yml
```diff
@@ -21,347 +16,13 @@ env:
...
-      - name: Audit
-        run: |
-          bundle exec bundler-audit
+  tests:
+    name: Run Tests
+    uses: ./.github/workflows/run_tests.yml
 
   aws_deploy:
     name: AWS Deploy
-    if: github.ref == 'refs/heads/main'
```
RachaelBooth, line 363: No longer needed as this pipeline only runs on pushes to main

spec/db/seeds_spec.rb
```diff
@@ -21,7 +21,8 @@
     allow(Rails.env).to receive(:review?).and_return(true)
   end
 
-  it "sets up correct data" do
+  # Doing this in one test should save ~2 minutes
```
kosiakkatrina, line 24: If this takes so long to run, would it make sense to create fewer resources in the seed for test env only?

RachaelBooth: We don't actually seed any data (other than possibly a list of local authorities) in test - but these tests are set up to run as if on a review env, where we do want there to be plenty.

kosiakkatrina: Oh, okay, of course

kosiakkatrina (approved):

PR #2854, opened 2024-12-06
CLDC-3789 Update frontend components

kosiakkatrina:
Use new `govuk_list` [List helper](https://govuk-components.netlify.app/helpers/list/) for our lists 
Use new [service navigation component](https://design-system.service.gov.uk/components/service-navigation/) - added some custom styles to keep it mostly the way it was
Our password fields will also be updated to look like this:
image

image


RachaelBooth (approved):
LGTM

I think this will update the royal arms in the footer as well (from govuk-frontend), which we want but is worth flagging to Alix etc so they know it's happened. (Although I don't think we can see that on review apps)

PR #2856, opened 2024-12-06
Update rails to 7.2.2

kosiakkatrina:


app/helpers/filters_helper.rb
```diff
@@ -165,9 +165,9 @@ def filters_applied_text(filter_type)
     applied_filters_count(filter_type).zero? ? "No filters applied" : "#{pluralize(applied_filters_count(filter_type), 'filter')} applied"
   end
 
-  def reset_filters_link(filter_type, path_params = {})
+  def reset_filters_link(filter_type, filter_path_params = {})
     if applied_filters_count(filter_type).positive?
-      govuk_link_to "Clear", clear_filters_path(filter_type:, path_params:)
+      govuk_link_to "Clear", clear_filters_path(filter_type:, filter_path_params:)
```
kosiakkatrina, line 170: Path_params kept on getting cleared, I think because of this: https://github.com/rails/rails/pull/43770

app/models/bulk_upload.rb
```diff
@@ -1,7 +1,7 @@
 class BulkUpload < ApplicationRecord
-  enum log_type: { lettings: "lettings", sales: "sales" }
-  enum rent_type_fix_status: { not_applied: "not_applied", applied: "applied", not_needed: "not_needed" }
-  enum failure_reason: { blank_template: "blank_template", wrong_template: "wrong_template" }
+  enum :log_type, { lettings: "lettings", sales: "sales" }
+  enum :rent_type_fix_status, { not_applied: "not_applied", applied: "applied", not_needed: "not_needed" }
+  enum :failure_reason, { blank_template: "blank_template", wrong_template: "wrong_template" }
```
kosiakkatrina, line 4: Our current syntax is deprecated in Rails 8

app/models/form/lettings/questions/offered.rb
```diff
@@ -7,7 +7,6 @@ def initialize(id, hsh, page)
     @check_answers_card_number = 0
     @max = 150
     @min = 0
-    @hint_text = I18n.t("hints.offered")
```
kosiakkatrina, line 10: This question is a 2023 question, but the translation for this is missing. We now get an exception because of `config.i18n.raise_on_missing_translations`

app/models/merge_request.rb
```diff
@@ -13,7 +13,9 @@ class MergeRequest < ApplicationRecord
     request_merged: "request_merged",
     deleted: "deleted",
   }.freeze
-  enum status: STATUS
+
+  attribute :status, :string
```
kosiakkatrina, line 17: Enums now have to be backed by either a db column or an attribute

config/locales/test.en.yml
```diff
@@ -0,0 +1,17 @@
+en:
```
kosiakkatrina, line 1: Don't know if this is the best place to put it, but as we raise all our missing translations, we have some that are only used in tests

db/seeds.rb
```diff
@@ -12,13 +12,13 @@ def find_or_create_user(organisation, email, name, role)
...
-  if LocalAuthority.count.zero?
-    la_path = "config/local_authorities_data/initial_local_authorities.csv"
-    service = Imports::LocalAuthoritiesService.new(path: la_path)
-    service.call
-  end
+if LocalAuthority.count.zero?
+  la_path = "config/local_authorities_data/initial_local_authorities.csv"
+  service = Imports::LocalAuthoritiesService.new(path: la_path)
+  service.call
+end
```
kosiakkatrina, line 19: enums can't be empty `{}` so this fixes location/scheme tests

config/locales/test.en.yml
RachaelBooth: This feels like a sign we should maybe get rid of them...although actually these are coming from test form definitions, right? Probably not too bad, and can't really think of anywhere better for them to go.

RachaelBooth (approved):
LGTM - feels like we might want to do a fair amount of testing though looking at what's needed to change.

PR #2864, opened 2024-12-11
CLDC-3801 Do not put error objects into cookies

kosiakkatrina:
This was causing some issues with cookie overflow (I suspect when there were many larger errors being put into flash). This PR filters out the errors to the ones present on the page and only keeps the attributes and messages in flash

### DIFF
--- a/app/controllers/form_controller.rb
+++ b/app/controllers/form_controller.rb
@@ -44,8 +44,8 @@ class FormController < ApplicationController
           end
         end
         flash[:log_data] = responses_for_page
-        question_ids = (log.errors.map(&:attribute) - [:base]).uniq
-        flash[:pages_with_errors_count] = question_ids.map { |id| log.form.get_question(id, log)&.page&.id }.compact.uniq.count
+        question_ids = (@log.errors.map(&:attribute) - [:base]).uniq
+        flash[:pages_with_errors_count] = question_ids.map { |id| @log.form.get_question(id, @log)&.page&.id }.compact.uniq.count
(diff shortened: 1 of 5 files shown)


app/controllers/form_controller.rb
```diff
@@ -116,7 +120,7 @@ def restore_previous_errors(previous_errors)
     return unless previous_errors
 
     previous_errors.each do |attribute, message|
-      @log.errors.add attribute, message.first
+      @log.errors.add attribute, message.html_safe
     end
   end
```
Dinssa, line 134: What is the change here for?

app/controllers/form_controller.rb
```diff
@@ -89,6 +95,7 @@ def show_page
           if flash[:errors].present?
             restore_previous_errors(flash[:errors])
             restore_error_field_values(flash[:log_data])
+            @questions_with_errors_count = flash[:questions_with_errors_count]
```
kosiakkatrina: Given that we now only recover the errors for page, we need to keep track of how many errors were actually added to the log to know if we need to display "see all related answers" flow

app/controllers/form_controller.rb
kosiakkatrina: Hmm, I think cause it's now coming through as a plain string as opposed to part of error object, it must have been handled within the error before in the individual cases we were setting it as html_safe.

I don't _think_ we're playing back any content that could mean this is harmful in our errors, and I think we already apply html_safe to all our error messages in BU error list, but please let me know if you can think of places where this could cause issues!

Dinssa (approved):

PR #2868, opened 2024-12-13
Update test file generation

kosiakkatrina:
- Extract test file generation into a separate controller
- Only create routes if feature flag is enabled
- Add a path to generate a BU file (2024 only atm)
- Display test file paths as links (buttons were getting a bit confusing)
image


app/controllers/test_data_controller.rb
```diff
@@ -0,0 +1,421 @@
...
+                         disposition: "attachment",
+                         after_send: lambda {
+                                       file.close
+                                       file.unlink
+                                     }
+  end
+
+private
+
+  def to_2024_lettings_row(log)
```
RachaelBooth: Is there a sensible way to share this code with the bits we use in tests for this?

kosiakkatrina: Hm, yeah, probably, I tried to avoid requiring files from tests and they weren't getting autoloaded. Realistically it probably is fine, but it doesn't feel right. 
Any ideas other than just requiring the test helper files?

RachaelBooth: Could they live somewhere around here (probably not this file) and then be required by the tests?

kosiakkatrina: I moved them to helpers, does that look ok?

RachaelBooth: I think that works - just should update the tests to use them too

kosiakkatrina: I think they are (?) it's very weird, but in the file changes it shows up as a rename, so the files in `spec/support/bullk_upload` no longer exist, at least when I pull it down

RachaelBooth: Huh, strange. Cool, lets get it in.

RachaelBooth (approved):

PR #2874, opened 2024-12-17
CLDC-3811: Upgrade postgres used in tests and docker compose to 13.18

RachaelBooth:


docker-compose.yml
```diff
@@ -5,7 +5,7 @@ volumes:
 
 services:
   db:
-    image: postgres:13.5-alpine
+    image: postgres:13.18-alpine
```
RachaelBooth, line 8: N.B. In the dockerfile we're already using the relevant postgres driver for this

Dinssa (approved):

PR #2875, opened 2024-12-18
Update some tests that would fail in 2025 collection

kosiakkatrina:


app/models/sales_log.rb
```diff
@@ -61,13 +61,14 @@ class SalesLog < Log
              [by_postcode, param_without_spaces, param_without_spaces])
   }
   scope :age1_answered, -> { where.not(age1: nil).or(where(age1_known: [1, 2])) }
+  scope :ecstat1_answered, -> { where.not(ecstat1: nil).or(where("saledate >= ?", Time.zone.local(2025, 4, 1))) }
```
kosiakkatrina, line 64: I don't think we ask this in 2025, so we'd get no duplicates

spec/factories/sales_log.rb
```diff
@@ -33,6 +33,7 @@
       noint { 2 }
       privacynotice { 1 }
       purchid { rand(999_999_999).to_s }
+      staircase { 1 }
```
kosiakkatrina, line 36: This is in setup for 2025

Dinssa (approved):

PR #2878, opened 2024-12-19
Use assume ssl setting

RachaelBooth:
Users are always accessing the site using ssl, we enforce that elsewhere, but that's terminated at the load balancer - it looks like we should have this setting on (it will e.g. ensure that cookies for analytics choice are set as secure cookies, which we ought to do).


kosiakkatrina (approved):
Makes sense

PR #2881, opened 2025-01-02
Fix soft validation unpend logs

Dinssa:


app/models/bulk_upload.rb
```diff
@@ -129,6 +129,7 @@ def fields_to_confirm(log)
   def unpend_and_confirm_soft_validations
     logs.find_each do |log|
       fields_to_confirm(log).each { |field| log[field] = 0 }
+      log.status = log.status_cache
```
kosiakkatrina, line 132: Maybe a test?

I assume at this point `status_cache` would be `in_progress` (cause it has the soft validation confirmations missing), I also assume it would get recalculated during `log.save!` and get set as `completed` but would be good to see that in test 👀 


When we unpend logs normally we set `log.skip_update_status = true` (line 119) https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/2851#discussion_r1871210990


How did this work before? I assume we were recalculating the statuses, cause I can't see any changes to this method in the PR I referenced. If so, it might be taking a little while to run those, but if it always has been, we can probably leave that for now

kosiakkatrina (approved):

PR #2883, opened 2025-01-06
Fix view bulk uploads and add another status label

Dinssa:
There are some errors in bulk uploads that are not currently accounted for as blank template or wrong template errors and are due to unexpected issues. This PR introduces a new status label, “Error processing CSV” to handle these cases and prevent showing an incorrect status of “Logs uploaded with no errors” (which is the catch all tag).

Screenshot 2025-01-06 at 15 48 48


app/services/bulk_upload/processor.rb
```diff
@@ -184,6 +185,8 @@ def handle_invalid_validator
       @bulk_upload.update!(failure_reason: "blank_template")
     elsif wrong_template_errors.any? { |error| validator.errors.full_messages.include?(error) }
       @bulk_upload.update!(failure_reason: "wrong_template")
+    else
+      @bulk_upload.update!(failure_reason: "invalid_upload")
     end
```
Dinssa: Adding the `invalid_upload` failure reason in this else statement wasn’t necessary to address the recent issue, but I included it to handle any unexpected cases that might arise in the future.

app/services/bulk_upload/processor.rb
```diff
@@ -61,6 +61,7 @@ def call
...
   rescue StandardError => e
     Sentry.capture_exception(e)
+    @bulk_upload.update!(failure_reason: "invalid_upload")
     send_failure_mail
```
Dinssa: If there are unexpected validation errors that are not handled by the `handle_invalid_validator` method or any other errors it will add this tag.

Dinssa (commented):
~What do you think about the error label? Should it be ‘Invalid upload’ or something else, like ‘Processing error’? I think ‘Processing error’ is more accurate in describing what happened.~

Updated the tag copy to "Error processing CSV" to match the email sent.

kosiakkatrina (commented):
Can we add a test? 👀

kosiakkatrina (approved):

PR #2886, opened 2025-01-07
CLDC-3822 Update date validations for 2025

kosiakkatrina:
Allow 20 year gap between start date and void date/mrcdate for 2025 lettings.

### DIFF
--- a/app/models/validations/date_validations.rb
+++ b/app/models/validations/date_validations.rb
@@ -11,14 +11,26 @@ module Validations::DateValidations
       record.errors.add :mrcdate, I18n.t("validations.lettings.date.mrcdate.not_first_let")
     end
 
-    if record["mrcdate"].present? && record["startdate"].present? && record["startdate"].to_date - record["mrcdate"].to_date > 3650
+    return unless record["mrcdate"].present? && record["startdate"].present?
+
+    if record.form.start_year_2025_or_later?
+      if record["startdate"].to_date - record["mrcdate"].to_date > 7300
+        record.errors.add :mrcdate, I18n.t("validations.lettings.date.mrcdate.twenty_years_before_tenancy_start")
+      end
+    elsif record["startdate"].to_date - record["mrcdate"].to_date > 3650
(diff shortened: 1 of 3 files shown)


app/models/validations/date_validations.rb
```diff
@@ -11,14 +11,26 @@ def validate_property_major_repairs(record)
...
     end
 
-    if record["mrcdate"].present? && record["startdate"].present? && record["startdate"].to_date - record["mrcdate"].to_date > 3650
+    return unless record["mrcdate"].present? && record["startdate"].present?
+
+    if record.form.start_year_2025_or_later?
+      if record["startdate"].to_date - record["mrcdate"].to_date > 7300
+        record.errors.add :mrcdate, I18n.t("validations.lettings.date.mrcdate.twenty_years_before_tenancy_start")
+      end
+    elsif record["startdate"].to_date - record["mrcdate"].to_date > 3650
```
RachaelBooth: I realise it's not a new thing, so not necessarily relevant to this PR, but should we be taking leap years into account in the comparison here? (in that 10 years will be slightly more than 3650 days, and similar for 20)

app/models/validations/date_validations.rb
```diff
@@ -42,18 +54,32 @@ def validate_startdate(record)
...
 private
 
   def is_rsnvac_first_let?(record)
     [15, 16, 17].include?(record["rsnvac"])
   end
+
+  def validate_startdate_against_mrc_and_void_dates(record)
+    if record.form.start_year_2025_or_later?
+      if record["voiddate"].present? && record["startdate"].to_date - record["voiddate"].to_date > 7300
+        record.errors.add :startdate, I18n.t("validations.lettings.date.startdate.twenty_years_after_void_date")
```
RachaelBooth: Would it be worth refactoring these to do the check once and add the error to both fields?

app/models/validations/date_validations.rb
kosiakkatrina: Yep!

app/models/validations/date_validations.rb
kosiakkatrina: It would be!

RachaelBooth (approved):

PR #2887, opened 2025-01-07
CLDC-3832: Move and update sheltered accomodation question

RachaelBooth:


app/models/form/lettings/questions/sheltered.rb
```diff
@@ -8,7 +8,14 @@ def initialize(id, hsh, page)
...
   def answer_options
-    if form.start_year_2024_or_later?
+    if form.start_year_2025_or_later?
+      { "1" => { "value" => "Yes – sheltered housing for tenants with low support needs" },
+        "2" => { "value" => "Yes – extra care housing" },
+        "7" => { "value" => "Yes - other" },
+        "3" => { "value" => "No" },
+        "divider" => { "value" => true },
+        "4" => { "value" => "Don’t know" } }
+    elsif form.start_year_2024_or_later?
```
kosiakkatrina: I think we should be able to remove the `else` and set this as the `else`, cause older logs can only be viewed and all of the answer options that exist in the current `else` also exist in this one and have the same wording

kosiakkatrina (approved):

PR #2888, opened 2025-01-08
CLDC-3826 Add LA changes for 2025

kosiakkatrina:
In this specific case small amount of postcodes are being moved from Barnsley to Sheffield.
Given that it's a small amount of postcodes, where we infer the LA for these we will treat it more like a change of the Ecode and ignore the area change. If there are any postcodes that would mistakenly get the wrong LA, they'd be fixed on case by case basis. 

Will also need to add new LAs when released:
Locally:`rbenv exec rake data_import:local_authorities\[config/local_authorities_data/2025_local_authorities.csv\]`
On aws: `overrides='{ "containerOverrides" : [{ "name" : "app", "command" : ["bundle", "exec", "rake", "data_import:local_authorities[config/local_authorities_data/2025_local_authorities.csv]"]}]}'`

And LA links for supported housing: 
Locally: `rbenv exec rake data_import:local_authority_links\[config/local_authorities_data/local_authority_links_2025.csv\]`
On aws: `overrides='{ "containerOverrides" : [{ "name" : "app", "command" : ["bundle", "exec", "rake", "data_import:local_authority_links[config/local_authorities_data/local_authority_links_2025.csv]"]}]}'`


app/models/log.rb
```diff
@@ -343,31 +343,27 @@ def reset_invalidated_dependent_fields!
...
 
   def get_inferred_la(postcode)
     result = PIO.lookup(postcode)
     location_code = result[:location_code] if result
-    if LA_CHANGES.key?(location_code) && form.start_date.year >= 2023
-      LA_CHANGES[location_code]
-    elsif !(LA_CHANGES.value?(location_code) && form.start_date.year < 2023)
+    if LA_CHANGES[form.start_date.year]&.key?(location_code)
+      LA_CHANGES[form.start_date.year][location_code]
+    elsif BACKWARDS_LA_CHANGES[form.start_date.year]&.key?(location_code)
```
kosiakkatrina, line 364: Specify the backwards LA changes explicity, so that it doesn't interfere with the `else` below

app/models/log.rb
```diff
@@ -343,31 +343,27 @@ def reset_invalidated_dependent_fields!
...
+      BACKWARDS_LA_CHANGES[form.start_date.year][location_code]
+    elsif !LA_CHANGES.value?(location_code)
```
kosiakkatrina, line 366: (This hasn't changed) Not all LA changes are backwards compatible, for example if a new LA is created from multiple LAs then it's unclear what to map back to, in that case the LA would just be nil and we'd ask for it to be entered explicitly in the form

app/models/log.rb
kosiakkatrina: Need to double check if this is actually backwards compatible tho ⌛

kosiakkatrina: Updated the description and this is okay to treat as backwards compatible

app/models/log.rb
```diff
@@ -343,31 +343,27 @@ def reset_invalidated_dependent_fields!
...
-    "E07000164" => "E06000065", # Hambleton => North Yorkshire
-    "E07000165" => "E06000065", # Harrogate => North Yorkshire
-    "E07000166" => "E06000065", # Richmondshire => North Yorkshire
-    "E07000167" => "E06000065", # Ryedale => North Yorkshire
-    "E07000168" => "E06000065", # Scarborough => North Yorkshire
-    "E07000169" => "E06000065", # Selby => North Yorkshire
-    "E07000187" => "E06000066", # Mendip => Somerset
-    "E07000188" => "E06000066", # Sedgemoor => Somerset
-    "E07000246" => "E06000066", # Somerset West and Taunton => Somerset
-    "E07000189" => "E06000066", # South Somerset => Somerset
```
kosiakkatrina, line 362: These would no longer get triggered, cause they're from 2023 and the APIs would now be up to date

Dinssa (approved):

PR #2890, opened 2025-01-08
CLDC-3830: Home page change quarterly deadlines in 25/26

Dinssa:


spec/features/home_page_spec.rb
```diff
@@ -0,0 +1,119 @@
...
+      end
+
+      scenario "displays correct current quarter as Q1" do
+        Timecop.freeze(Time.zone.local(current_collection_year, 4, 1)) do
+          visit root_path
+          expect(page).to have_selector("h1.govuk-heading-l + p.govuk-body.govuk-body-m", text: "Q1 - Friday 12 July 2024: Quarterly cut off date for tenancies and sales starting between 1 April 2024 and 30 June 2024.")
+          expect(page).to have_content("Q1 - Friday 12 July 2024")
+        end
+        Timecop.return
+      end
```
Dinssa: This tests that there is the current quarter deadline text directly after the heading "Upcoming deadlines". That's one way to test it, but the second expect might also have been enough since it includes the text we're looking for. And it doesn't expand the deadlines section so we can be sure that this text is directly after the heading. So in short, I should choose one over the other than have both here.

app/views/home/_upcoming_deadlines.html.erb
```diff
@@ -1,7 +1,6 @@
 <h1 class="govuk-heading-l">Upcoming deadlines</h1>
 
 <% open_lettings_form = FormHandler.instance.in_crossover_period? ? FormHandler.instance.previous_lettings_form : FormHandler.instance.current_lettings_form %>
-<% formatted_deadline = "#{open_lettings_form.submission_deadline.strftime('%A')} #{open_lettings_form.submission_deadline.to_formatted_s(:govuk_date)}" %>
```
Dinssa, line 4: Now a method in the helper file

app/helpers/collection_time_helper.rb
```diff
@@ -30,6 +30,10 @@ def current_collection_end_date
     Time.zone.local(current_collection_start_year + 1, 3, 31).end_of_day
   end
 
+  def current_collection_end_year
+    current_collection_start_year + 1
+  end
+
```
Dinssa, line 36: We do need to know the current collection start year plus 1, I could have just did that calculation on the fly in the view but chose to make this method to be self documenting.

app/helpers/collection_time_helper.rb
```diff
@@ -50,12 +54,56 @@ def archived_collection_start_year
...
-      { quarter: "Q1", cutoff_date: Time.zone.local(2024, 7, 12), start_date: Time.zone.local(2024, 4, 1), end_date: Time.zone.local(2024, 6, 30) },
-      { quarter: "Q2", cutoff_date: Time.zone.local(2024, 10, 11), start_date: Time.zone.local(2024, 7, 1), end_date: Time.zone.local(2024, 9, 30) },
-      { quarter: "Q3", cutoff_date: Time.zone.local(2025, 1, 10), start_date: Time.zone.local(2024, 10, 1), end_date: Time.zone.local(2024, 12, 31) },
+  def quarter_deadlines
+    Form::QUARTERLY_DEADLINES
+  end
+
+  def quarter_deadlines_for_year(year)
+    quarter_deadlines[year]
+  end
```
Dinssa: Could've been merged together into one method but left it separate incase it becomes useful anytime in the future.

kosiakkatrina: Given that the deadlines aren't form specific, would it make sense to put this somewhere else? Like form_handler?

app/views/home/_upcoming_deadlines.html.erb
Dinssa: This might be wrong, as I may not have thought through what should be shown on the when in a crossover period. I think I'll revert this change, what do you think?

app/helpers/collection_time_helper.rb
kosiakkatrina: And by this I mean the constant, commented on the wrong file..

kosiakkatrina: Changed my mind.
There are quite a few similarly named methods and it was a bit hard to wrap my head around at first, so it might make sense to either create a collection_deadlines helper and extract everything to there or move the constant into this file

app/views/home/_upcoming_deadlines.html.erb
kosiakkatrina: I think a helper method is fine, but we might want to pass in `open_lettings_form` to it and use it instead of `current_collection_start_year`, cause I think you're right that it'd be previous year in crossover.

Dinssa: I removed it, I change it to just use submission deadline once instead of twice (figuring out the day then adding the `govuk_date` formatted rest of it). After that, it seemed too small to extract to the helper. As a helper method with `open_lettings_form` you’d still need to declare a variable in the view, so there wouldn’t be much improvement.

`<% formatted_deadline = "#{open_lettings_form.submission_deadline.strftime("%A %-d %B %Y")}" %>`

kosiakkatrina (approved):

PR #2892, opened 2025-01-10
CLDC-3831 Update person relationship questions

kosiakkatrina:
Relationship question and options are changing slightly. Same as sales.

### DIFF
--- a/app/models/form/lettings/questions/person_partner.rb
+++ b/app/models/form/lettings/questions/person_partner.rb
@@ -1,0 +1,30 @@
+class Form::Lettings::Questions::PersonPartner < ::Form::Question
+  def initialize(id, hsh, page, person_index:)
+    super(id, hsh, page)
+    @id = "relat#{person_index}"
+    @type = "radio"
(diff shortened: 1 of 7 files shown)


app/models/form/lettings/questions/person_partner.rb
```diff
@@ -0,0 +1,30 @@
...
+    {
+      "P" => { "value" => "Yes" },
+      "X" => { "value" => "No" },
+      "R" => { "value" => "Tenant prefers not to say" },
+    }
+  end
+
+  def question_number
+    base_question_number = case form.start_date.year
+                           when 2023
```
RachaelBooth, line 22: Is this needed?

RachaelBooth (approved):

app/models/form/lettings/questions/person_partner.rb
kosiakkatrina: hmm, no, I thought I'd leave it until we get the question numbers

PR #2893, opened 2025-01-10
CLDC-3837: Fix bug with step validation for numeric fields with step 0.1

RachaelBooth:


app/models/validations/shared_validations.rb
```diff
@@ -54,14 +54,15 @@ def validate_numeric_step(record)
       field = question.check_answer_label || question.id
       incorrect_accuracy = (value.to_d * 100) % (question.step * 100) != 0
 
-      if question.step < 1 && incorrect_accuracy
-        record.errors.add question.id.to_sym, I18n.t("validations.shared.numeric.nearest_hundredth", field:)
-      elsif incorrect_accuracy || value.to_d != value.to_i    # if the user enters a value in exponent notation (eg '4e1') the to_i method does not convert this to the correct value
```
RachaelBooth, line 59: N.B. I don't this this `value.to_d != value.to_i` check is actually doing anything useful - anything entered as e.g. "4e1" is being caught by the numeric format validation, and that's a more relevant error message than these.

spec/models/validations/shared_validations_spec.rb
```diff
@@ -137,80 +141,96 @@
...
+        sales_log.income1 = 30_123.74
+        shared_validator.validate_numeric_step(sales_log)
+        expect(sales_log.errors[:income1]).to include I18n.t("validations.shared.numeric.nearest_step", field: "Buyer 1’s gross annual income", step: 0.001)
+      end
+    end
+  end
 
-        it "is valid when owning_organisation nil" do
-          log = build(log_type, owning_organisation: nil)
+  %i[sales_log lettings_log].each do |log_type|
```
RachaelBooth, line 202: This section is just being pulled out of the describe block that it should never have been in - I've not made any changes.

config/locales/validations/shared.en.yml
```diff
@@ -9,7 +9,9 @@ en:
         above_min: "%{field} must be at least %{min}."
         whole_number: "%{field} must be a whole number."
         nearest_ten: "%{field} must be given to the nearest ten."
+        nearest_tenth: "%{field} must be given to the nearest tenth."
         nearest_hundredth: "%{field} must be given to the nearest hundredth."
+        nearest_step: "${field} must be given to the nearest %{step}."
```
RachaelBooth, line 14: This seems a better backup for steps we've not put a nice word message in for than the current state.

kosiakkatrina (approved):

PR #2898, opened 2025-01-15
CLDC-3838: Unlock users on password reset

RachaelBooth:
Notes: 

I tried just enabling the email unlock strategy initially (which would make this happen automatically), but that also adds in various links to resend unlock instructions (which look confusing to me) and caused an app error when it tried to send that email (which is probably fixable, and may be a review app thing, but put me off anyway). 

So this just changes the check when resetting a password for whether to unlock.

### DIFF
--- a/app/controllers/auth/passwords_controller.rb
+++ b/app/controllers/auth/passwords_controller.rb
@@ -35,7 +35,7 @@ class Auth::PasswordsController < Devise::PasswordsController
     yield resource if block_given?
 
     if resource.errors.empty?
-      resource.unlock_access! if unlockable?(resource)
+      resource.unlock_access! if resource.respond_to?(:unlock_access!)
       if Devise.sign_in_after_reset_password
         set_flash_message!(:notice, password_update_flash_message)
         resource.after_database_authentication
--- a/config/initializers/devise.rb
+++ b/config/initializers/devise.rb
@@ -205,7 +205,7 @@ Devise.setup do |config|
   # :time  = Re-enables login after a certain amount of time (see :unlock_in below)
   # :both  = Enables both strategies
   # :none  = No unlock strategy. You should handle unlocking by yourself.
-  config.unlock_strategy = :both
+  config.unlock_strategy = :time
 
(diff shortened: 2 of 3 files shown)


app/controllers/auth/passwords_controller.rb
```diff
@@ -35,7 +35,7 @@ def update
     yield resource if block_given?
 
     if resource.errors.empty?
-      resource.unlock_access! if unlockable?(resource)
```
RachaelBooth, line 38: N.B. `unlockable?(resource)` is defined by devise and checks `resource.respond_to?(:unlock_access!) && resource.respond_to?(:unlock_strategy_enabled?) && resource.unlock_strategy_enabled?(:email)`.

kosiakkatrina: I'm confused why it needs email unlock strategy to be enabled. If it doesn't have a strategy set (I assume time doesn't work in this scenario(?)) to use to do the unlocking, what does it do? I assume if it has the email strategy it evokes it and sends the emails, does it evoke anything in this scenario? (also my assumption might be wrong)

Given that we're happy for the password change to unlock account, it's probably fine to do it without additional checks as long as we only do it here (which looks like the only place we explicitly call unlock_access)

Would it be worth adding a test?

RachaelBooth: Yeah, it kinda surprised me that `unlockable?` insisted on the email strategy being enabled - I don't really know why this is. I guess I see why you might want to force people to wait out the time, but I'm surprised it's not just a devise setting for whether or not to do so that we would be checking.

The unlock strategy thing is about what the normal route is for user accounts to be unlocked - email would be to send them an email, time is to wait a given length of time, or you can enable both (which I guess gives the time as a backup if they've not received/ignored the email). Calling `unlock_access!` doesn't invoke any checks, just sets the user to be unlocked (it assumes those checks have already happened - which at this point in the password reset journey we know that the password reset has been allowed, so we're saying that's enough).

kosiakkatrina (approved):

PR #2903, opened 2025-01-16
Remove git version from Dockerfile

kosiakkatrina:


Dockerfile
```diff
@@ -10,7 +10,7 @@ RUN apk add --update --no-cache tzdata && \
 # build-base: compilation tools for bundle
 # yarn: node package manager
 # postgresql-dev: postgres driver and libraries
-RUN apk add --no-cache build-base=0.5-r3 busybox=1.36.1-r7 nodejs-current=20.8.1-r0 yarn=1.22.19-r0 postgresql13-dev=13.18-r0 git=2.40.3-r0 bash=5.2.15-r5
+RUN apk add --no-cache build-base=0.5-r3 busybox=1.36.1-r7 nodejs-current=20.8.1-r0 yarn=1.22.19-r0 postgresql13-dev=13.18-r0 git bash=5.2.15-r5
```
RachaelBooth: Do we actually use git in this file? (or need it on the final image?)

kosiakkatrina: Good question. I've removed it in one of the runs and it didn't fail, but thought to keep it to be safe. If you can't think of a reason we might want it, I'll remove it

RachaelBooth: I'd definitely expect that if we were using it it would be in the get-and-build-code stage, so if we aren't it seems unnecessary.

RachaelBooth (approved):

PR #2909, opened 2025-01-20
Update exports documentation

kosiakkatrina:


docs/exports.md
```diff
@@ -4,16 +4,143 @@ nav_order: 7
...
+
+Typically, **one lettings logs collection (current)** is exported most of the time, and **two collections (current and previous)** are exported during the crossover period. Export objects are still created for all collections, but they are marked as **empty** when applicable, and no files are generated for these empty exports.
+
+**Users** and **organisations** are not year-specific exports, so there will never be more than one collection for each of these in a single partial export.
 
 Initially the application database field names and field types were chosen to match the existing CDS data as closely as possible to minimise the amount of transformation needed. This has led to a less than optimal data model though and increasingly we should look to transform at the mapping layer where beneficial for our application.
 
-We have a cron job triggering the export service daily at 5am.
+A **cron job** triggers the export service daily at **5 a.m.**.
```
RachaelBooth: Is this 5am UTC or 5am in the timezone that includes London? (I.e. what happens during daylight savings time?)

I suspect this will never matter, but might be useful to write down just in case something strange happens at any point.

RachaelBooth (approved):

PR #2910, opened 2025-01-22
CLDC-3850: Clear postcode_known when clearing postcode_full

RachaelBooth:


app/models/log.rb
```diff
@@ -258,6 +263,8 @@ def blank_compound_invalid_non_setup_fields!
       self.address_line2 = nil
       self.town_or_city = nil
       self.postcode_full = nil
+      self.postcode_known = nil if lettings?
+      self.pcodenk = nil if sales?
```
RachaelBooth, line 267: I'm struggling to get a unit test which actually isolates the problem here - suggest we try and merge this anyway (with manual testing) and I'll look at putting one in separately.

kosiakkatrina (approved):

PR #2914, opened 2025-01-23
Fix copy multiple matches for for this address

Dinssa:
This PR improves the copy returned in bulk uploads as follows:

Screenshot 2025-01-23 at 11 11 03

### DIFF
--- a/app/services/bulk_upload/lettings/year2024/row_parser.rb
+++ b/app/services/bulk_upload/lettings/year2024/row_parser.rb
@@ -617,7 +617,9 @@ private
   def validate_address_option_found
     if log.uprn.nil? && field_16.blank? && key_address_fields_provided?
       error_message = if log.address_options_present? && log.address_options.size > 1
-                        I18n.t("#{ERROR_BASE_KEY}.address.not_determined")
+                        I18n.t("#{ERROR_BASE_KEY}.address.not_determined.multiple")
+                      elsif log.address_options_present?
+                        I18n.t("#{ERROR_BASE_KEY}.address.not_determined.one")
                       else
                         I18n.t("#{ERROR_BASE_KEY}.address.not_found")
                       end
--- a/app/services/bulk_upload/lettings/year2024/row_parser.rb
+++ b/app/services/bulk_upload/lettings/year2024/row_parser.rb
@@ -616,7 +616,7 @@ private
 
   def validate_address_option_found
     if log.uprn.nil? && field_16.blank? && key_address_fields_provided?
-      error_message = if log.address_options_present?
+      error_message = if log.address_options_present? && log.address_options.size > 1
                         I18n.t("#{ERROR_BASE_KEY}.address.not_determined")
                       else
                         I18n.t("#{ERROR_BASE_KEY}.address.not_found")
--- a/config/locales/validations/lettings/2024/bulk_upload.en.yml
+++ b/config/locales/validations/lettings/2024/bulk_upload.en.yml
@@ -49,8 +49,10 @@ en:
           age:
             invalid: "Age of person %{person_num} must be a number or the letter R"
           address:
-            not_found: "We could not find this address. Check the address data in your CSV file is correct and complete, or find the correct address using the CORE site."
-            not_determined: "There are multiple matches for this address. Check the address data in your CSV file is correct and complete, or select the correct address using the CORE site."
+            not_found: "We could not find this address. Check the address data in your CSV file is correct and complete, or find the correct address in the service."
+            not_determined:
+              one: "There is a possible match for this address which doesn't look right. Check the address data in your CSV file is correct and complete, or confirm the address in the service."
+              multiple: "There are multiple matches for this address. Check the address data in your CSV file is correct and complete, or select the correct address in the service."
             not_answered: "Enter either the UPRN or the full address."
           nationality:
             invalid: "Select a valid nationality."
--- a/config/locales/validations/lettings/2024/bulk_upload.en.yml
+++ b/config/locales/validations/lettings/2024/bulk_upload.en.yml
@@ -17,7 +17,7 @@ en:
           owning_organisation:
             not_found: "The owning organisation code is incorrect."
             not_stock_owner: "The owning organisation code provided is for an organisation that does not own stock."
-            not_permitted: 
+            not_permitted:
               not_support: "You do not have permission to add logs for this owning organisation."
               support: "This owning organisation is not affiliated with %{org_name}."
           managing_organisation:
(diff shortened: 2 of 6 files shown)


config/locales/validations/lettings/2024/bulk_upload.en.yml
```diff
@@ -50,9 +50,9 @@ en:
             invalid: "Age of person %{person_num} must be a number or the letter R"
           address:
             not_found: "We could not find this address. Check the address data in your CSV file is correct and complete, or select the correct address using the CORE site."
-            not_determined: "There are multiple matches for this address. Either select the correct address manually or correct the UPRN in the CSV file."
+            not_determined: "There are multiple matches for this address. Either enter the correct address manually or correct the UPRN in the CSV file."
```
RachaelBooth: I know this reads better, but the user will actually (at the moment) first be asked to select the address from the search results right? Should this be wording that covers that rather than implying manual entry?

Dinssa: On the site, users would be prompted to “select an address.” However, in a bulk upload, they would have entered the address in their CSV file. From their perspective, they would need to go back to the CSV and enter different values.

Dinssa: This doesn't change the error message on the site, just on the bulk upload.

Dinssa: Perhaps **Either enter the correct address ~manually~ or correct the UPRN in the CSV file.** and expect both changes to be made in the CSV? I'm not sure when it says "select the correct address manually" it is clear that you're supposed to first choose the option to fix in service then select from the matches it has found. Maybe it's just me, what do you think?

RachaelBooth: I think I do read it as either "enter the correct address manually (in the service after selecting fix in the service)" or "put the correct UPRN in the csv file (and upload again)", since the user will already have entered address data they believe to be correct in the csv file if the search happened and gave multiple matches.

Not sure what the typical user would be thinking though.

Also, I think they only see this message if the uprn field was blank in the CSV, so "correct the uprn in the csv file" might also be misleading.

For not found (i.e. there were no matches at our minimum threshold) we include "Check the address data in your CSV file is correct and complete, or select the correct address using the CORE site." - if we're happy with that message, then something similar here might work? ("correct and complete" is probably the wrong wording, but "select the correct address using the CORE site" might be useful)

Dinssa: You're right it is only seen when uprn field was blank:
```
def validate_address_option_found
    if log.uprn.nil? && field_16.blank? && key_address_fields_provided?
      error_message = if log.address_options_present?
                        I18n.t("#{ERROR_BASE_KEY}.address.not_determined")
                      else
                        I18n.t("#{ERROR_BASE_KEY}.address.not_found")
                      end
      %i[field_17 field_18 field_19 field_20 field_21 field_22].each do |field|
        errors.add(field, error_message) if errors[field].blank?
      end
    end
  end
```

I think that suggestion will work nicely, I could change it to: "There are multiple matches for this address. Either enter the correct address **using the CORE site** or **enter a** UPRN in the CSV file."

I saw this error when there was only one address option returned so I think it's also misleading to say "select the correct address in the service". Perhaps in addition to checking if address_options is present we should check if it is more than 1 within the `validate_address_option_found` method.

Dinssa: Or, assuming the user did not enter a UPRN for a reason (because they didn't have one and would have if they did):

"There are multiple matches for this address. **Check the address data in your CSV file is correct and complete, or select the correct address using the CORE site.**"
I think this is the right scenario where you'd need to select the correct address in the site.

The other error for when the address is not found:
"We could not find this address. Check the address data in your CSV file is correct and complete, or select the correct address using the CORE site."

Does it make sense to say "select the correct address..." or more likely the user would need to go and find it using the find address feature, so should be:

"We could not find this address. Check the address data in your CSV file is correct and complete, or **find** the correct address using the CORE site."

RachaelBooth: That makes sense to me.

app/services/bulk_upload/lettings/year2024/row_parser.rb
```diff
@@ -616,7 +616,7 @@ def validate_uprn_exists_if_any_key_address_fields_are_blank
 
   def validate_address_option_found
     if log.uprn.nil? && field_16.blank? && key_address_fields_provided?
-      error_message = if log.address_options_present?
+      error_message = if log.address_options_present? && log.address_options.size > 1
```
RachaelBooth, line 619: Do we need a different message for exactly 1, or is the not found version good enough?

Dinssa: I think the not found error message is enough, although 1 was found it wasn't similar enough to be chosen. We should treat it as not found so a user can take the next steps.

config/locales/validations/lettings/2024/bulk_upload.en.yml
```diff
@@ -49,10 +49,10 @@ en:
           age:
             invalid: "Age of person %{person_num} must be a number or the letter R"
           address:
-            not_found: "We could not find this address. Check the address data in your CSV file is correct and complete, or select the correct address using the CORE site."
-            not_determined: "There are multiple matches for this address. Either select the correct address manually or correct the UPRN in the CSV file."
+            not_found: "We could not find this address. Check the address data in your CSV file is correct and complete, or find the correct address using the CORE site."
+            not_determined: "There are multiple matches for this address. Check the address data in your CSV file is correct and complete, or select the correct address using the CORE site."
```
kosiakkatrina: I know this has been through a few revisions now, but I wouldn't say `CORE` here, given it's the legacy system name and we should avoid using it.
Maybe `select correct address directly in the service` ?

spec/services/bulk_upload/sales/year2024/row_parser_spec.rb
```diff
@@ -1104,7 +1104,7 @@ def purch_id_field
               parser.valid?
               expect(parser.errors[:field_22]).to be_empty
               %i[field_23 field_24 field_25 field_26 field_27 field_28].each do |field|
-                expect(parser.errors[field]).to eql([I18n.t("validations.sales.2024.bulk_upload.address.not_determined")])
+                expect(parser.errors[field]).to eql([I18n.t("validations.sales.2024.bulk_upload.address.not_found")])
```
kosiakkatrina: Could we instead update this test to have 2 matches so that the not_determined path still gets tested?

config/locales/validations/lettings/2024/bulk_upload.en.yml
Dinssa: Good point I'll update it

app/services/bulk_upload/lettings/year2024/row_parser.rb
Dinssa: On second thoughts, updating this to add a different message for a single returned result.

spec/services/bulk_upload/sales/year2024/row_parser_spec.rb
Dinssa: I added an additional test to lettings and sales to test for when one result is returned and multiple. There will be slightly different error message to reflect what is going on.

What do you think?
```
not_determined:
  one: "There is a possible match for this address which doesn't look right. Check the address data in your CSV file is correct and complete, or confirm the address in the service."
  multiple: "There are multiple matches for this address. Check the address data in your CSV file is correct and complete, or select the correct address in the service."
```

kosiakkatrina (approved):

PR #2916, opened 2025-01-23
Enable managing collection resources on prod

kosiakkatrina:


app/services/feature_toggle.rb
```diff
@@ -28,7 +28,7 @@ def self.allow_future_resource_updates?
   end
 
   def self.managing_resources_enabled?
-    !Rails.env.production?
+    true
   end
```
Dinssa: Was this not working on production?

carolynbarker: Looks like it was disabled on production, was that expected?

Dinssa: I don't think that was expected. If we're enabling it on all environments, is there any need for this method?

kosiakkatrina: Deleted it! I think we just forgot tbh

Dinssa (approved):

PR #2920, opened 2025-01-29
CLDC-3852 Add some BU documentation

kosiakkatrina:


docs/bulk_upload.md
```diff
@@ -0,0 +1,47 @@
...
+- Uploads the file to S3
+- Schedules `ProcessBulkUploadJob`
+
+### Bulk upload service
+
+There are several outcomes to a bulk upload:
+
+- Successful upload
+- Partial upload: upload has errors but partial logs can be created, need a user approval
+- Errors in bulk upload: errors on important fields, or in the template. Logs can't be created and email with errors is sent to the user
```
Dinssa: This might imply that an email with errors is only sent when there are important errors, don't we also send emails for partial uploads. In both cases we do direct the user to an error report, maybe worth mentioning that? Also, the email doesn't actually include the errors just a message that there are some.

kosiakkatrina: Yeah, that's a good point. I updated it a bit now. 
I think in some cases we do send the errors directly in the email (if there's an issue with the template), I've updated it to include both

Dinssa (approved):

PR #2924, opened 2025-01-31
CLDC-3787 Autocomplete address search

Dinssa:
This PR creates a new way to add addresses to logs. 

## Before
Screenshot 2025-02-27 at 17 31 30

Currently we ask users if they know the property's UPRN, if they do they can enter it and continue. We'll derive the address and local authority from the UPRN.

Screenshot 2025-02-27 at 18 01 10

Alternatively a user can attempt to find the address by search, inputing an address line and postcode. The results who show on the next page and a user is asked to pick from the 10 available results. We'll derive the UPRN and local authority from their selection.

Screenshot 2025-02-27 at 18 01 26
Screenshot 2025-02-27 at 18 01 44

Should a user not find the address they're looking for they can enter it manually. We'd store the address but not derive a UPRN. If the postcode is valid then we'll still be able to derive a local authority but it's unlikely if the address was not found in the previous steps.
Screenshot 2025-02-27 at 18 02 02

## After
Screenshot 2025-02-27 at 17 54 32

In this PR we replace the URPN question and Find Address feature with a search box, returning results as the user types. We also show only one version of the address question at a time (search/manual).

Screenshot 2025-02-27 at 17 53 43
Screenshot 2025-02-27 at 17 54 12

This search box uses the same OS Places API to populate the results as we use in the old question but allows the user to make changes to their query dynamically. However, should they be unable to find their address they can switch between this search question and the manual entry question.

Screenshot 2025-02-27 at 17 52 59

### After: Users with JavaScript disabled
Should a user have JavaScript disabled the search box feature will not be available. We'd then add a message that the user should manually enter the address.
Screenshot 2025-02-27 at 18 05 35

Possible future improvements:
- We could retain the old find address feature, when JS is enabled replace it with this search box and display that instead when it is disabled.
- We could determine if JS is disabled, skip the search box question and go straight to the manual entry version.


app/controllers/address_search_controller.rb
```diff
@@ -0,0 +1,77 @@
...
+          presenter = AddressDataPresenter.new(result)
+          { address: presenter.address, uprn: presenter.uprn }
+        end
+        render json: formatted_results
+      end
+    end
+  end
+
+  def manual_input
+    log = params[:log_type] == "lettings_log" ? LettingsLog.find(params[:log_id]) : SalesLog.find(params[:log_id])
```
kosiakkatrina: Should this be `user.lettings_logs.find ...` so that we're only searching the logs available to the user?

app/controllers/address_search_controller.rb
```diff
@@ -0,0 +1,77 @@
...
+    log.update!(address_search_input: true)
+    redirect_to search_address_link(log)
+  end
+
+private
+
+  def manual_address_link(log)
+    base_url = send("#{log.log_type}_url", log)
+    "#{base_url}/address"
+  end
```
kosiakkatrina: Can we use route helpers here?

kosiakkatrina: Also would be nice to get that from the page somehow instead of hardcoding

app/frontend/controllers/address_search_controller.js
```diff
@@ -0,0 +1,57 @@
...
+
+export default class extends Controller {
+  connect () {
+    const searchUrl = JSON.parse(this.element.dataset.info).search_url
+    const selectEl = this.element
+
+    accessibleAutocomplete.enhanceSelectElement({
+      defaultValue: '',
+      selectElement: selectEl,
+      minLength: 1,
```
kosiakkatrina: Would it make sense to have a higher min length? I feel like it's unlikely that we'll have a single/two digit UPRN (?)

app/controllers/address_search_controller.rb
Dinssa: Sure good idea

spec/fixtures/files/sales_logs_csv_export_codes_24.csv
```diff
@@ -1,3 +1,3 @@
...
+,completed,,2024-05-01T00:00:00+01:00,2024-05-01T00:00:00+01:00,2024,1,,false,1,5,2024,MHCLG,MHCLG,billyboy@eyeklaud.com,billyboy@eyeklaud.com,,2,8,,,,1,1,2,1,1,"1, Test Street",,Test Town,,AA1 1AA,true,Westminster,E09000033,1,,Address line 1,SW1A 1AA,address line 1 as entered,address line 2 as entered,town or city as entered,county as entered,AB1 2CD,la as entered,2,1,1,1,30,X,17,17,826,1,1,P,35,X,17,,826,1,1,3,C,14,X,9,X,-9,X,3,R,-9,R,10,,,,,1,0,SW1A,1AA,1,E09000033,Westminster,1,1,1,1,,3,,1,4,5,1,1,0,13400,1,0,13400,1,4,1,,1,2,10,,,,,,,,,,,,,,,,,110000.0,,,1,20000.0,5,,10,1,80000.0,,,1,100.0,,,10000.0
```
Dinssa: In theses sales logs csv export fixture files, it now needed uprn_selection to equal 1 instead of nil. I believe this is because we now set uprn_selection to equal the uprn (which is also 1). Weirdly the same change was not needed for lettings. Thoughts?

app/controllers/address_search_controller.rb
```diff
@@ -0,0 +1,77 @@
...
+  def index
+    query = params[:query]
+
+    if query.match?(/\A\d+\z/) && query.length > 5
+      # Query is all numbers and greater than 5 digits, assume it's a UPRN
+      service = UprnClient.new(query)
+      service.call
+
+      if service.error.present?
+        render json: { error: service.error }, status: :unprocessable_entity
```
kosiakkatrina: What would the user see in this case?

app/controllers/address_search_controller.rb
```diff
@@ -0,0 +1,77 @@
...
+          presenter = AddressDataPresenter.new(result)
+          { address: presenter.address, uprn: presenter.uprn }
+        end
+        render json: formatted_results
+      end
+    end
+  end
+
+  def manual_input
+    log = params[:log_type] == "lettings_log" ? current_user.lettings_logs.find(params[:log_id]) : current_user.sales_logs.find(params[:log_id])
```
kosiakkatrina:
```suggestion
    log = current_user.send(params[:log_type]).find(params[:log_id])
```
?

spec/shared/shared_examples_for_derived_fields.rb
```diff
@@ -25,11 +25,14 @@
...
+      Timecop.freeze(Time.zone.local(2023, 4, 1)) do
+        log = FactoryBot.build(log_type, uprn_known: 0, uprn: nil, uprn_confirmed: 0)
+        allow(log.form).to receive(:start_year_2024_or_later?).and_return(false)
+        expect { log.set_derived_fields! }.to not_change(log, :uprn_known)
+                                          .and not_change(log, :uprn)
+                                          .and not_change(log, :uprn_confirmed)
+      end
+      Timecop.return
     end
   end
```
Dinssa, line 37: What we included in the derived files to only handle the `manual_address_entry_selected` logic behind the condition that the start year was 2024 or later prevents this test for failing only when the test itself is before 2024. Is this what we expect? I imagine we don't want to affect older logs at anytime not just before 2024, but it's not possible for a log from 2023 to trigger `set_derived_fields!` anymore so I don't think we should worry about this. Except is that wrong and is it possible to trigger `set_derived_fields!` if we ever had to bulk update previous logs to fix an error.

app/controllers/address_search_controller.rb
kosiakkatrina: Could also extract this into `before_action` for both `manual_input` and `search_input`

spec/shared/shared_log_examples.rb
```diff
@@ -156,7 +155,6 @@
                                               .and change(log, :uprn_confirmed).from(nil).to(1)
                                               .and change(log, :uprn).from(nil).to("UPRN")
                                               .and change(log, :uprn_known).from(nil).to(1)
-                                              .and change(log, :uprn_selection).from("UPRN").to(nil)
                                               .and change(log, :county).from("county").to(nil)
       end
```
Dinssa, line 159: We now set uprn_selection to also equal the UPRN value.

app/models/form/lettings/pages/address_search.rb
```diff
@@ -0,0 +1,23 @@
+class Form::Lettings::Pages::AddressSearch < ::Form::Page
+  def initialize(id, hsh, subsection)
+    super
+    @id = "address_search"
+    @copy_key = "sales.property_information.address_search"
```
kosiakkatrina, line 5: Would this not be the default key?

app/models/form/lettings/pages/address_search.rb
```diff
@@ -0,0 +1,23 @@
...
+    @questions ||= [
+      Form::Lettings::Questions::AddressSearch.new(nil, nil, self),
+    ]
+  end
+
+  def skip_href(log = nil)
+    return unless log
+
+    "/#{log.log_type.dasherize}s/#{log.id}/first-time-property-let-as-social-housing"
+  end
```
kosiakkatrina: Do we need this? Wouldn't it automatically choose the next routed to page?

app/models/form/lettings/questions/address_search.rb
```diff
@@ -0,0 +1,48 @@
...
+    return {} unless ActiveRecord::Base.connected?
+    return {} unless log&.address_options
+
+    answer_opts = {}
+
+    (0...[log.address_options.count, 10].min).each do |i|
+      answer_opts[log.address_options[i][:uprn]] = { "value" => log.address_options[i][:address] }
+    end
+
+    answer_opts
```
kosiakkatrina: What would the address options be here? I thought this would be empty/have only the selected option until someone triggers a search?

app/views/form/_select_question.html.erb
```diff
@@ -1,20 +1,23 @@
-
 <% selected = @log.public_send(question.id) || "" %>
```
kosiakkatrina, line 1: Is this file just formatting?

app/views/form/page.html.erb
```diff
@@ -75,6 +75,12 @@
         <%= f.hidden_field :check_errors, value: @check_errors %>
       <% end %>
 
+      <% if @page.id == "address" %>
+        <div class="govuk-button-group">
+          <%= govuk_link_to "Clear address and search instead", address_search_input_path(@log.log_type, @log.id), class: "govuk-button govuk-button--secondary" %>
+        </div>
+      <% end %>
```
kosiakkatrina: Could this be something similar to guidance_partial? Would probably be safer not to hardcode page IDs

db/migrate/20250203084427_add_address_search_to_both_logs.rb
```diff
@@ -0,0 +1,6 @@
+class AddAddressSearchToBothLogs < ActiveRecord::Migration[7.2]
```
kosiakkatrina: Can we merge the migrations?

db/schema.rb
```diff
@@ -245,14 +245,14 @@
     t.integer "hb"
     t.integer "hbrentshortfall"
     t.integer "property_relet"
-    t.datetime "mrcdate", precision: nil
+    t.datetime "mrcdate"
```
kosiakkatrina: These shouldn't be changing (?) Maybe worth double checking the version you're using when running migrations

app/models/form/lettings/pages/address_search.rb
Dinssa: Since the question id is now _uprn_, to get the page header we need to add the copy key. I do however need to remove from the translation files the old questions' copy, so could rename `address_search` to `uprn` but I think address_search is clearer so would be inclined to keep it as is.

app/models/form/lettings/pages/address_search.rb
Dinssa: Good point. This was originally from when we displayed both the manual entry and search entry options. However, since we’re now only showing one at a time, it’s no longer necessary. I'll remove.

app/views/form/_select_question.html.erb
Dinssa: The formatting is a little unintentional and I'll adjust the indentation. But the changes here are around handling the possibility of answers being empty:
```
  <% if answers.any? %>
    <% answers.each do |answer| %>
      <option value="<%= answer.id %>"
              data-synonyms="<%= answer_option_synonyms(answer.resource) %>"
              data-append="<%= answer_option_append(answer.resource) %>"
              data-hint="<%= answer_option_hint(answer.resource) %>"
              <%= question.answer_selected?(@log, answer) ? "selected" : "" %>><%= answer.name || answer.resource %></option>
    <% end %>
  <% else %>
    <option value="" disabled></option>
  <% end %>
```
The if condition and else block with a disabled blank option is new.

All of these changes are not actually need though, as you've pointed out before we do have search boxes with no options elsewhere and they've been handled differently. I can revert the changes in this file but it's also a nice to have?

app/controllers/address_search_controller.rb
Dinssa: Not this, they'd just see "No results found", I haven't implemented different error messages within the autocomplete box.

app/models/form/lettings/questions/address_search.rb
Dinssa: I'll update this to show the one single answer option which will exist when you've selected an address.

db/schema.rb
Dinssa: I've undone the changes to this file

db/migrate/20250203084427_add_address_search_to_both_logs.rb
Dinssa: Yup, deleted the old ones

app/views/form/page.html.erb
Dinssa: Yup, i've added both buttons on each address page to the bottom guidance partials.

app/controllers/address_search_controller.rb
```diff
@@ -0,0 +1,70 @@
...
+      service = AddressClient.new(query)
+      service.call
+
+      if service.error.present?
+        render json: { error: service.error }, status: :unprocessable_entity
+      else
+        results = service.result.map do |result|
+          presenter = AddressDataPresenter.new(result)
+          { text: presenter.address, value: presenter.uprn }
+        end
```
Dinssa, line 30: I've updated the controller here and the js controller so that the key for the address is now `text` and the uprn is now `value` (from `address` and `uprn`). This makes the address search controller js and existing search controller js more similar. They now only differ when determining which value to save and which to show. It is possible to switch out the address search js controller in the question view and use the search one, but as you type the result will show the uprn value not address. I think it's possible to resolve this in future improvements.

app/views/form/_select_question.html.erb
kosiakkatrina: I mean if it's redundant, could remove, but if there are situations where this would catch something then feel free to leave

app/controllers/address_search_controller.rb
```diff
@@ -18,7 +18,7 @@ def index
       end
     elsif query.match?(/[a-zA-Z]/)
       # Query contains letters, assume it's an address
-      service = AddressClient.new(query)
+      service = AddressClient.new(query, { minmatch: 0.2 })
```
kosiakkatrina, line 21: What does this do?

kosiakkatrina: The minmatch I mean

app/models/validations/property_validations.rb
```diff
@@ -50,7 +50,7 @@ def validate_property_postcode(record)
   end
 
   def validate_la_in_england(record)
-    return unless record.form.start_year_2025_or_later?
+    return unless record.form.start_year_2025_or_later? || (record.startdate_changed? && record.startdate_was.year == 2025 && record.startdate.year == 2024)
```
kosiakkatrina: Do we now also want to validate it for 2024?

Dinssa: No? I think we're not adding the LA validation in 2024 because it's mid year? This change just checks the address is in England in all of 2025 plus the one scenario when someone tries to change a 2024 log to a 2025 log, they'd be prevented from making that change. They'd see the same error message we use in 2025 and an all related answers page prompting them to clear one of the start date or address. Think this work, unless we want to also add the validation on LA to all logs in 2024 going forward from when this is released.

app/controllers/address_search_controller.rb
Dinssa: Ah forgot to talk about this. It will reduce the threshold for returning acceptable results, so OS Places via AddressClient will return results for search queries with fewer characters and less certainty, meaning as a user types they get a slightly better experience with results starting to popup sooner.

Dinssa: I was thinking of also increasing the number of results returned from the default, that is now possible via this `options` parameter should we want to do that. For example passing: `{ minmatch: 0.2, maxresults: 15 }`

Dinssa: What the `minmatch` change now means in practice is that where previously the ability to search for address and uprn at the same time was theoretically possible in practice there weren't any values of say less than 5 numeric characters where both the Address and UPRN api would return anything. In reality you'd end up for an small numeric input e.g. "134" just getting a single UPRN result back because the Address one had a high threshold and would not return anything yet. This way we now return the UPRN result for "134" first and also 10 other address results. I had thought about returning only 9 other address results if 1 is returned from UPRN to keep the overall results to 10 but did not make the change and it isn't that important, we can make that change though using `maxresults`.

app/models/validations/property_validations.rb
kosiakkatrina: Maybe I'm misreading this, but isn't this checking that it changed from 2025 (startdate_was) to 2024 (startdate)?

app/controllers/address_search_controller.rb
kosiakkatrina: Is there at all any sorting? Do we know that the best matches would always be in that top 10, or is there a chance that there will be a lot of not very close results purely because they get returned as first on the list?

Dinssa: I don’t perform any sorting, but OS Places does return a somewhat sorted list. For example, if you type in a full postcode, you’ll see all the addresses for that road in order. Since we’re passing the maxResults variable of 10 to OS Places, they return the top 10 best results on their end. If we were receiving 100 results and selecting the top 10 ourselves, there might be a chance of getting less relevant results. However, because OS Places handles the sorting, we can assume they are returning the top 10 best matches based on our request.

The results they send back could be better sorted and more accurate in matching. When there is sufficient information, such as a postcode and address line, we can rely on getting a good list. With less data, the results can be a bit random, but they are still fairly relevant to the query entered.

app/models/validations/property_validations.rb
Dinssa: You're right, this change is incorrect. The additional condition was not needed as the validation is already correctly triggering when switching between years. I've found the issue is to do with uprn_selection not being cleared and am working on a fix. I've removed the changes to these validations.

app/controllers/address_search_controller.rb
```diff
@@ -0,0 +1,70 @@
...
+        render json: { error: service.error }, status: :not_found
+      else
+        results = service.result.map do |result|
+          presenter = AddressDataPresenter.new(result)
+          { text: presenter.address, value: presenter.uprn }
+        end
+        render json: results
+      end
+    else
+      # Query is ambiguous, use both APIs and merge results
```
carolynbarker, line 34: So this happens if the search is 5 digits or fewer, and all digits? Doesn't this result in loads of matches? I'm wondering if there are too many results to be very useful - though I suspect it's nicer UX to have this even if so

app/services/bulk_upload/sales/year2025/row_parser.rb
```diff
@@ -961,6 +961,7 @@ def attributes_for_log
     attributes["address_line1_input"] = address_line1_input
     attributes["postcode_full_input"] = postcode_full
     attributes["select_best_address_match"] = true if field_16.blank?
+    attributes["manual_address_entry_selected"] = field_16.blank?
```
carolynbarker, line 964: Thanks!

config/locales/validations/sales/property_information.en.yml
```diff
@@ -7,7 +7,7 @@ en:
             joint_purchase: "Buyers’ last accommodation and discounted ownership postcodes must match."
             not_joint_purchase: "Buyer’s last accommodation and discounted ownership postcodes must match."
           invalid: "Enter a postcode in the correct format, for example AA1 1AA."
-          not_in_england: "It looks like you have entered a postcode outside of England - only submit Lettings forms for Lettings that occur in England"
+          not_in_england: "It looks like you have an entered a postcode outside of England. Only create logs for sales in England."
```
carolynbarker: I think this "an" is a typo

db/migrate/20250219122817_add_manual_address_entry_selected_to_logs.rb
```diff
@@ -0,0 +1,6 @@
+class AddManualAddressEntrySelectedToLogs < ActiveRecord::Migration[7.2]
+  def change
+    add_column :sales_logs, :manual_address_entry_selected, :boolean, default: false
+    add_column :lettings_logs, :manual_address_entry_selected, :boolean, default: false
```
carolynbarker, line 4: Why do we store this? Are there any specific requirements around it?

app/models/form/sales/questions/address_search.rb
```diff
@@ -0,0 +1,45 @@
...
+  def initialize(id, hsh, page)
+    super
+    @id = "uprn"
+    @type = "address_search"
+    @copy_key = "sales.property_information.address_search"
+    @plain_label = true
+    @bottom_guidance_partial = "address_search"
+    @question_number = QUESTION_NUMBER_FROM_YEAR[form.start_date.year] || QUESTION_NUMBER_FROM_YEAR[QUESTION_NUMBER_FROM_YEAR.keys.max]
+    @hide_question_number_on_page = true
+    @conditional_on = nil
```
carolynbarker: What does this do?

spec/shared/shared_log_examples.rb
carolynbarker: We could change this line to test that uprn_selection changes like uprn, and separately test the old behaviour for a lof from 2013

db/migrate/20250219122817_add_manual_address_entry_selected_to_logs.rb
Dinssa: We use it to determine if we show the address search or manual entry question. It'll be set to false by default so we show the address search page. Initially we tried to determine that by checking the values in urpn, uprn_confirmed, uprn_known, address_line1, postcode, etc... but there were a few scenarios where that would not be enough.

app/models/form/sales/questions/address_search.rb
Dinssa: Nice, the `@conditional_on = nil` is redundant, I'll remove it. It initially was needed when I kept the `uprn_known` page which only showed the `uprn` question if answered 'yes', adding this line stopped the old logic, but now I've removed the `uprn_known` question so it's no longer needed and you don't see it on the lettings version.

spec/shared/shared_log_examples.rb
Dinssa: This was changed for all years so we can't test it for pre-2024 separately. I could add: `.and not_change(log, :uprn_selection)`, but I it might make sense to keep it removed as its testing a field we no longer care about (`uprn_selection` is what we store from the old find address feature).

app/controllers/address_search_controller.rb
Dinssa: Yes, it’s for a scenario where someone has either typed a UPRN or house number. At most, we’d get 11 results per character typed (1 UPRN, 10 addresses). I do think it's nicer UX wise to get some results back sooner as they type, even if not yet what they're looking for. Most users likely won’t type in 5 digits without including an alphabetic character if they’re searching for an address. We could adjust this threshold to 4 digits, 5 is just arbitrary.

kosiakkatrina (approved):

PR #2927, opened 2025-02-03
Update sales factory and location tests

kosiakkatrina:


spec/factories/sales_log.rb
```diff
@@ -80,7 +80,6 @@
       noint { 2 }
       privacynotice { 1 }
       age1_known { 0 }
-      staircase { 1 }
```
kosiakkatrina, line 83: Completed log has `ownershipsch: 2` which is discounted ownership. It cannot be staircase anyway cause only shared ownership can be staircased, but setting staircase to 1 here, clears some of the other fields and this log ends up being not completed (at least in 2025)

spec/models/location_spec.rb
```diff
@@ -1016,15 +1016,16 @@
       end
 
       it "returns reactivating soon if the location has a future reactivation date" do
-        FactoryBot.create(:location_deactivation_period, deactivation_date: Time.zone.yesterday, reactivation_date: Time.zone.tomorrow, location:)
+        deactivation_period = FactoryBot.create(:location_deactivation_period, deactivation_date: Time.zone.yesterday, reactivation_date: Time.zone.tomorrow, location:)
         location.save!
+        expect(deactivation_period.deactivation_date).to eq(Time.zone.yesterday.to_date)
```
kosiakkatrina: The tests for this are flaky and instead of `location.status` being `reactivating_soon` it ends up being `active`. I initially thought it's a time issue, but couldn't trigger a failure when setting different times. I'm not sure why this is happening, there isn't an obvious way to reproduce it on the review apps, so adding this line just to check if the deactivation period does get created, but the status is still `active`

.github/workflows/run_tests.yml
```diff
@@ -369,7 +369,7 @@ jobs:
 
       - name: Create database
         run: |
-          bundle exec rake parallel:setup
+          bundle exec rake db:prepare
```
kosiakkatrina, line 372: These tests aren't being ran in parallel

Dinssa (approved):

PR #2930, opened 2025-02-04
CLDC-3863 Enable adding question numbers to page headers

Dinssa:


app/models/form/lettings/pages/address_fallback.rb
```diff
@@ -22,4 +23,6 @@ def questions
       Form::Lettings::Questions::PostcodeForFullAddress.new(nil, nil, self),
     ]
   end
+
+  QUESTION_NUMBER_FROM_YEAR = { 2024 => 13, 2025 => 13 }.freeze
```
kosiakkatrina, line 27: Is this meant to be 12 or was the number wrong in the translations? Both years?

config/locales/forms/2025/sales/property_information.en.yml
```diff
@@ -44,7 +44,7 @@ en:
             question_text: "Select the correct address"
 
           address:
-            page_header: "Q12 - What is the property's address?"
+            page_header: "What is the property's address?"
```
kosiakkatrina, line 47: Does this need an update in 2023 as well?

app/models/form/lettings/pages/address_fallback.rb
Dinssa: In 2023 it was 12 the UPRN question was 11, but in 2024 it should be 13 as the previous UPRN question is 12. At the moment the 2025 question number is just a placeholder.

config/locales/forms/2025/sales/property_information.en.yml
Dinssa: I left the 2023 translation file as is with the question number written in the translation file, to keep an example of how it was and illustrate that this change (adding question numbers to the page model) is just one way to achieve the same result. Also, we no longer show the 2023 page.

kosiakkatrina: Okay, I was thinking about CYA label, cause we still show 2023 review page, but realised that that's getting the number from the question instead. Thanks!

kosiakkatrina (approved):

PR #2931, opened 2025-02-05
CLDC-3858 Correctly set checkbox values

kosiakkatrina:
When some of the checkbox questions are routed to and some of the checkbox values are selected, all of the unselected values should be set to 0.

So far it's been happening on `submit_form`, in `form_controller`, so it only applied to single log submission logs.

This PR moves that out into `lettings_variables` which should now also apply to Bulk Uploaded logs.

This PR also corrects `reasonpref_reason` and `ilness_type` checkbox values for existing bulk uploaded logs.


app/controllers/form_controller.rb
```diff
@@ -155,9 +155,9 @@ def responses_for_page(page)
 
       next unless question_params
 
-      if %w[checkbox validation_override].include?(question.type)
```
kosiakkatrina, line 158: I don't think `validation_override` is a thing anymore (or was ever a thing)

app/controllers/form_controller.rb
```diff
@@ -155,9 +155,9 @@ def responses_for_page(page)
...
+      if %w[checkbox].include?(question.type)
         question.answer_keys_without_dividers.each do |option|
-          result[option] = question_params.include?(option) ? 1 : 0
+          result[option] = 1 if question_params.include?(option)
```
kosiakkatrina, line 160: 0s are now getting set in the derived variables

spec/fixtures/files/lettings_log_csv_export_codes_23.csv
```diff
@@ -1,3 +1,3 @@
...
+,completed,,s.port@jeemayle.com,s.port@jeemayle.com,false,2023-11-26T00:00:00+00:00,,2023-11-26T00:00:00+00:00,1,,,2023,MHCLG,MHCLG,1,7,0,2023-11-26,2,2,,,2,HIJKLMN,ABCDEFG,0,,,Address line 1,,London,,NW9 5LL,false,Barnet,E09000003,0,2,6,2,2,7,1,1,3,2023-11-24,1,,1,2023-11-25,,3,1,4,,2,,1,4,,1,4,0,0,2,35,,F,0,2,13,0,0,P,32,M,6,1,R,-9,R,10,0,R,-9,R,10,,,,,,,,,,,,,,,,,,,,,1,4,1,2,1,0,1,0,0,0,0,0,0,1,0,0,1,0,0,0,0,0,0,0,2,7,4,,6,1,0,TN23 6LZ,1,false,Ashford,E07000105,1,0,1,0,0,0,0,0,1,0,2,,0,0,1,268,,6,1,1,,0,2,,,,,200.0,100.0,,50.0,25.0,40.0,20.0,35.0,17.5,325.0,162.5,,,,1,0,12.0,6.0,,,,,,,,,,,,,,,,,,,,
```
kosiakkatrina, line 3: The CSV changes are just the nulls changing to 0s or "No" for checkbox values, because they're now getting set for checkbox questions even if the log isn't created via single log submission

app/controllers/form_controller.rb
carolynbarker: As there's only one type now we could go for just `if question.type == "checkbox"`

spec/lib/tasks/correct_checkbox_values_spec.rb
```diff
@@ -0,0 +1,216 @@
+require "rails_helper"
+require "rake"
+
+RSpec.describe "correct_checkbox_values" do
```
carolynbarker, line 4: We could potentially add tests for the case where more than one checkbox value is 1 (just to cover it, I don't expect any problem in that case)

carolynbarker (approved):
Looks good to me! Made a couple of minor suggestions

app/controllers/form_controller.rb
kosiakkatrina: Oh, good point!

spec/lib/tasks/correct_checkbox_values_spec.rb
kosiakkatrina: I think the second test (`"updates the reasonable preference reason values on a pending log"` ) has that scenario, so it might be enough, but let me know if you think we should test it explicitly

carolynbarker: It looks like that second test has reasonpref 1 and rp_homeless 1, but no other `rp_`s?

I think it might be nice to test it explicitly, to rule out some bug like finding the field that's 1 and setting everything else to 0, though to be fair I don't think there's actually much risk of that

app/controllers/form_controller.rb
kosiakkatrina: You're right! Updated it now

spec/lib/tasks/correct_checkbox_values_spec.rb
kosiakkatrina: You're right! Updated it now

PR #2936, opened 2025-02-10
CLDC-3806 Update question numbers for sales

kosiakkatrina:


app/models/form/sales/questions/living_before_purchase_years.rb
```diff
@@ -19,5 +19,6 @@ def suffix_label(log)
   QUESTION_NUMBER_FROM_YEAR_AND_OWNERSHIP = {
     2023 => { 1 => 75, 2 => 99 },
     2024 => { 1 => 77, 2 => 101 },
```
carolynbarker: Not part of your change I know, but while reviewing I noticed this doesn't look right for last year - it should be Q100 for 2024-2025 according to the paper form I'm looking at

app/models/form/sales/questions/person_gender_identity.rb
```diff
@@ -22,13 +22,9 @@ def initialize(id, hsh, page, person_index:)
     "R" => { "value" => "Person prefers not to say" },
   }.freeze
 
+  BASE_QUESTION_NUMBERS = { 2023 => 30, 2024 => 32, 2025 => 30 }.freeze
```
carolynbarker, line 25: Is this right? I think these numbers are for ethnicity - they are the same numbers as in `buyer2_ethnic_group.rb`

In the paper forms I have, person 2's gender identity is Q31 in 2024 and Q29 in 2025

app/models/form/sales/questions/person_relationship_to_buyer_1_yes_no.rb
```diff
@@ -25,5 +25,5 @@ def question_number
     base_question_number + (4 * @person_index)
   end
 
-  QUESTION_NUMBER_FROM_YEAR = { 2025 => 30 }.freeze
+  QUESTION_NUMBER_FROM_YEAR = { 2025 => 28 }.freeze
```
carolynbarker: I think Q28 is buyer 2's age and relationship is Q27

app/models/form/sales/questions/person_working_situation.rb
```diff
@@ -63,13 +63,9 @@ def answer_options
     end
   end
 
+  BASE_QUESTION_NUMBERS = { 2023 => 31, 2024 => 33, 2025 => 31 }.freeze
```
carolynbarker, line 66: Again these don't match what I'm expecting from the paper forms - in the 2024 one Q33 is buyer 2 ethnic background and Q35 is buyer 2 working situation, in the 2025 one (on the ticket) Q31 is buyer 2 ethnic background and Q33 is buyer 2 working situation

app/models/form/sales/questions/purchase_price.rb
```diff
@@ -15,6 +15,7 @@ def initialize(id, hsh, page, ownershipsch:)
   QUESTION_NUMBER_FROM_YEAR_AND_OWNERSHIP = {
     2023 => { 2 => 100, 3 => 110 },
     2024 => { 2 => 101, 3 => 111 },
+    2025 => { 2 => 103 },
```
carolynbarker, line 18: I think this is right for discounted ownership. Do we want to include the purchase price questions for shared ownership or staircasing as well?

app/models/form/sales/questions/value.rb
```diff
@@ -8,9 +8,11 @@ def initialize(id, hsh, page)
...
     @prefix = "£"
-    @question_number = QUESTION_NUMBER_FROM_YEAR[form.start_date.year] || QUESTION_NUMBER_FROM_YEAR[QUESTION_NUMBER_FROM_YEAR.keys.max]
+    @question_number = question_number_from_year[form.start_date.year] || question_number_from_year[question_number_from_year.keys.max]
     @top_guidance_partial = "financial_calculations_shared_ownership"
   end
 
-  QUESTION_NUMBER_FROM_YEAR = { 2023 => 88, 2024 => 89 }.freeze
+  def question_number_from_year
+    { 2023 => 88, 2024 => 89, 2025 => subsection.id == "shared_ownership_staircasing_transaction" ? 97 : 80 }
+  end
```
carolynbarker, line 17: Do we want this for discounted ownership?

app/models/form/sales/questions/person_gender_identity.rb
kosiakkatrina: I think we might be looking at different files. This is what I'm seeing for 2025 sales, and the way I'm reading it, person 2 gender is 38, so in `question_number` method below `base_question_number + (4 * @person_index)` == `30 + (4 * 2)` == 38 
image

app/models/form/sales/questions/person_relationship_to_buyer_1_yes_no.rb
kosiakkatrina: Ah, this should be named `BASE_QUESTION_NUMBERS` instead cause we use it for person numbers in `question_number` method above as well. So similarly to the gender number, it would get calculated, so for person 2 it would be 36, person 3 - 40 and so on. I think buyer has separate questions to person questions definitions. A bit confusing!

Updated the constant name tho

app/models/form/sales/questions/person_working_situation.rb
kosiakkatrina: I think same as above

app/models/form/sales/questions/purchase_price.rb
kosiakkatrina: I think this question only ever gets created for discounted ownership. Shared ownership uses `Form::Sales::Questions::Value` for the same question instead. Potentially cause they are/were slightly different questions or pages

app/models/form/sales/questions/value.rb
kosiakkatrina: Ah, this is the same question as `Form::Sales::Questions::PurchasePrice`, which is the discounted ownership one, so this would only show for shared ownership

app/models/form/sales/questions/living_before_purchase_years.rb
kosiakkatrina: Nice spot, updated it now!

app/models/form/sales/questions/person_gender_identity.rb
carolynbarker: Oh I see! Thanks - yes I agree this is correct.

app/models/form/sales/questions/person_relationship_to_buyer_1_yes_no.rb
carolynbarker: Thanks for explaining - yes this is correct too then! Good call on the constant name update

carolynbarker (approved):

PR #2940, opened 2025-02-12
CLDC-3829 Add guidance link and deadlines to log pages

kosiakkatrina:
On not started logs:
image

On incomplete logs with start date that is not overdue:
image

On  incomplete logs with start date that is overdue:
image


app/helpers/collection_deadline_helper.rb
```diff
@@ -61,11 +61,12 @@ def quarter_dates(year)
       first_quarter(year).merge(quarter: "Q1"),
       second_quarter(year).merge(quarter: "Q2"),
       third_quarter(year).merge(quarter: "Q3"),
+      fourth_quarter(year).merge(quarter: "Q4"),
```
kosiakkatrina, line 64: Not sure if there's a reason this wasn't here, but let me know if this messes with anything else that I might have not noticed

app/helpers/collection_deadline_helper.rb
```diff
@@ -61,11 +61,12 @@ def quarter_dates(year)
...
     ]
   end
 
   def quarter_for_date(date: Time.zone.now)
-    quarters = quarter_dates(current_collection_start_year)
+    quarters = quarter_dates(collection_start_year_for_date(date))
```
kosiakkatrina: This should still work the same with the default date, but in the cases where we want to get a quarter that is outside of this year, I think it makes sense to get the relevant collection year

app/frontend/styles/_red-link.scss
```diff
@@ -9,6 +9,10 @@
   color: govuk-colour("red");
 }
 
+.app-red-text {
+  color: govuk-colour("red");
+}
+
```
carolynbarker: this isn't really about links so I'm not sure it goes here - is there a file with more text/headers/warnings sort of styling that would be a better fit?

app/helpers/tasklist_helper.rb
```diff
@@ -70,6 +71,20 @@ def tasklist_link_class(status)
...
   end
 
+  def deadline_text(log)
+    return if log.completed?
+    return if log.startdate.nil?
+
+    log_quarter = quarter_for_date(date: log.startdate)
+    deadline_for_log = log_quarter.cutoff_date
+
+    if deadline_for_log > Time.zone.now
```
carolynbarker: What happens on the day itself? Do we have to worry about timezones?

spec/helpers/tasklist_helper_spec.rb
```diff
@@ -204,4 +204,42 @@
...
+      let(:log) { build(:sales_log, saledate: Time.zone.local(2025, 6, 1)) }
+
+      it "returns the deadline text" do
+        allow(Time.zone).to receive(:now).and_return(Time.zone.local(2025, 5, 7))
+        allow(Time.zone).to receive(:today).and_return(Time.zone.local(2025, 5, 7))
+        expect(deadline_text(log)).to include("Deadline: ")
+      end
+    end
+
+    context "when today is after the deadline for log with sale/start date" do
```
carolynbarker, line 243: I think it would be nice to have a test giving expected behaviour on the deadline day itself

carolynbarker (approved):

app/frontend/styles/_red-link.scss
kosiakkatrina: Yep, moved it to the tasklist.scss

app/helpers/tasklist_helper.rb
kosiakkatrina: Good point, I thought that we'd show the deadline (instead of overdue), but it appears that it's not what it was doing. I updated it now!

spec/helpers/tasklist_helper_spec.rb
kosiakkatrina: Added a test!

spec/helpers/tasklist_helper_spec.rb
```diff
@@ -204,4 +204,49 @@
...
+      it "returns the deadline text" do
+        allow(Time.zone).to receive(:today).and_return(Time.zone.local(2025, 5, 7))
+        expect(deadline_text(log)).to include("Q1 Deadline: 11 July 2025.")
+      end
+    end
+
+    context "when today is the deadline for log with sale/start date" do
+      let(:log) { build(:sales_log, saledate: Time.zone.local(2025, 2, 1)) }
+
+      it "returns the overdue text" do
```
carolynbarker, line 237: Tiny thing but I think this test name should be "returns the deadline text" not the overdue text? Confused me at first

app/helpers/tasklist_helper.rb
carolynbarker: Nice one! That's the behaviour I'd expect too

app/helpers/collection_deadline_helper.rb
Dinssa: On the home page we don't say the fourth quarter Q4 as a deadline but instead call it "End of year deadline", that's why it wasn't needed.

PR #2848, opened 2024-12-04
CLDC-3788 Export sales logs

kosiakkatrina:

PR #2943, opened 2025-02-13
CLDC-3788 Export sales logs

kosiakkatrina:
Copy of https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/2848 with some field changes


app/services/exports/sales_log_export_constants.rb
```diff
@@ -0,0 +1,143 @@
+module Exports::SalesLogExportConstants
+  MAX_XML_RECORDS = 10_000
+  LOG_ID_OFFSET = 300_000_000_000
+
+  EXPORT_MODE = {
+    xml: 1,
+    csv: 2,
+  }.freeze
+
+  EXPORT_FIELDS = Set["id",
```
carolynbarker: Do we have a final fields list? 
Rachel mentioned on the ticket she was adding `mrentprestaircasing` which I don't see here.

I also wonder whether we might be missing another couple of fields, by my count this is 123 fields +18+5 = 146 fields, compared to 148 on the spreadsheet on the ticket (which doesn't include `mrentprestaircasing`). Is this the service charge fields?

spec/factories/sales_log.rb
```diff
@@ -202,5 +202,109 @@
...
+      disabled { 1 }
+      deposit { 80_000 }
+      value { 110_000 }
+      value_value_check { 0 }
+      grant { 10_000 }
+      proplen { 10 }
+      pregyrha { 1 }
+      pregla { 1 }
+      pregother { 1 }
+      pregghb { 1 }
```
carolynbarker: I don't think we have these in 2025 (and we don't export sales logs from 2024 or earlier, so I think we can remove these)

spec/services/exports/export_service_spec.rb
```diff
@@ -23,7 +25,7 @@
     Timecop.return
   end
 
-  context "when exporting daily XMLs" do
+  context "when exporting daily XMLs before 2025" do
```
carolynbarker, line 28: Should we move the let(:start_time) from line 8 here?

spec/services/exports/export_service_spec.rb
```diff
@@ -330,4 +390,180 @@
...
+
+  context "with date after 2025-04-01" do
+    let(:start_time) { Time.zone.local(2025, 5, 1) }
+    let(:expected_master_manifest_filename) { "Manifest_2025_05_01_0001.csv" }
+    let(:lettings_logs_export_service) { instance_double("Exports::LettingsLogExportService", export_xml_lettings_logs: {}) }
+
+    context "when exporting daily XMLs" do
+      context "and no sales archives get created in sales logs export" do
+        let(:sales_logs_export_service) { instance_double("Exports::SalesLogExportService", export_xml_sales_logs: {}) }
+        let(:lettings_logs_export_service) { instance_double("Exports::LettingsLogExportService", export_xml_lettings_logs: {}) }
```
carolynbarker, line 403: This is already set on line 397 right?

spec/services/exports/export_service_spec.rb
kosiakkatrina: I duplicated it so that it's explicit, but we use this in other tests so left it on line 8 as well

spec/factories/sales_log.rb
kosiakkatrina: Good shout

spec/services/exports/export_service_spec.rb
kosiakkatrina: Yep, removed it now!

app/services/exports/sales_log_export_constants.rb
kosiakkatrina: The final list is still to be confirmed, so it might change!

kosiakkatrina: These have now been confirmed!

Dinssa (approved):

PR #2944, opened 2025-02-13
Remove invalid option causing migration errors

carolynbarker:
These class_name options cause errors in Rails 7.2 (and I think 7.1) - I think they were being ignored anyway so have been having no effect.


db/schema.rb
```diff
@@ -245,14 +245,14 @@
     t.integer "hb"
     t.integer "hbrentshortfall"
     t.integer "property_relet"
-    t.datetime "mrcdate"
+    t.datetime "mrcdate", precision: nil
```
carolynbarker, line 248: these are autogenerated when I migrate the db up - `nil` is the default so these should have no effect.

kosiakkatrina (approved):

PR #2945, opened 2025-02-13
CLDC-3802 sales bulk upload update 2025

carolynbarker:
Sales bulk upload review.

I'll check back through this myself as well and test it out, but keen to get any feedback in the meantime.


app/helpers/bulk_upload/sales_log_to_csv.rb
```diff
@@ -92,6 +100,15 @@ def to_2024_csv_row(seed: nil)
     end
   end
 
+  def to_2025_csv_row(seed: nil)
```
carolynbarker: I want to commonise some of these methods

app/services/bulk_upload/sales/year2025/csv_parser.rb
```diff
@@ -0,0 +1,124 @@
+require "csv"
+
+class BulkUpload::Sales::Year2025::CsvParser
+  include CollectionTimeHelper
+
+  FIELDS = 121
+  MAX_COLUMNS = 142
```
carolynbarker, line 7: This is the same as before. There is one column of row headers, so we need to allow at least 122 columns

app/views/bulk_upload_sales_logs/forms/prepare_your_file_2025.html.erb
carolynbarker, line 1: This is currently the same as prepare_your_file_2024.html.erb but with the year updated in the copy. Do we know yet whether there is new copy for this page in 2025?

config/locales/validations/sales/2025/bulk_upload.en.yml
carolynbarker, line 1: same as 2024

spec/services/bulk_upload/sales/log_creator_spec.rb
```diff
@@ -6,13 +6,13 @@
...
-  let(:row_parser) { instance_double(BulkUpload::Sales::Year2024::RowParser) }
+  let(:bulk_upload) { create(:bulk_upload, :sales, user:, year: 2025) }
+  let(:csv_parser) { instance_double(BulkUpload::Sales::Year2025::CsvParser) }
+  let(:row_parser) { instance_double(BulkUpload::Sales::Year2025::RowParser) }
   let(:log) { build(:sales_log, :completed, assigned_to: user, owning_organisation: owning_org, managing_organisation: owning_org) }
 
   before do
-    allow(BulkUpload::Sales::Year2024::CsvParser).to receive(:new).and_return(csv_parser)
+    allow(BulkUpload::Sales::Year2025::CsvParser).to receive(:new).and_return(csv_parser)
     allow(csv_parser).to receive(:row_parsers).and_return([row_parser])
```
carolynbarker: this doesn't look right to me, I actually want to add 2025 as a separate section and test both

spec/services/bulk_upload/sales/year2025/row_parser_spec.rb
carolynbarker, line 1: I've updated the fields from last year, but I haven't yet added tests for the new fields

app/services/bulk_upload/sales/year2025/row_parser.rb
```diff
@@ -0,0 +1,2063 @@
...
+
+  attribute :field_42, :integer
+  attribute :field_43, :string
+  attribute :field_44, :string
+  attribute :field_45, :integer
+  attribute :field_46, :integer
+  attribute :field_47, :string
+  attribute :field_48, :string
+  attribute :field_49, :integer
+  attribute :field_49, :integer
```
kosiakkatrina: I think this should be 50?

app/services/bulk_upload/sales/year2025/row_parser.rb
```diff
@@ -0,0 +1,2063 @@
...
+  attribute :field_85, :integer
+  attribute :field_86, :integer
+  attribute :field_87, :integer
+  attribute :field_88, :integer
+  attribute :field_89, :integer
+  attribute :field_90, :integer
+  attribute :field_91, :integer
+  attribute :field_92, :integer
+  attribute :field_93, :decimal
+
```
kosiakkatrina, line 249: Should there be 94 and 95 here?

app/services/bulk_upload/sales/year2025/csv_parser.rb
```diff
@@ -0,0 +1,124 @@
...
+  def correct_field_count?
+    valid_field_numbers_count = field_numbers.count { |f| f != "field_blank" }
+
+    valid_field_numbers_count == FIELDS
+  end
+
+private
+
+  def default_field_numbers
+    (1..FIELDS).map do |number|
```
kosiakkatrina, line 74: Nice

app/services/bulk_upload/sales/year2025/csv_parser.rb
```diff
@@ -0,0 +1,124 @@
...
+  def field_numbers
+    @field_numbers ||= if with_headers?
+                         rows[row_offset - 1][col_offset..].map { |number| number.to_s.match?(/^[0-9]+$/) ? "field_#{number}" : "field_blank" }
+                       else
+                         default_field_numbers
+                       end
+  end
+
+  def headers
+    @headers ||= ("field_1".."field_#{FIELDS}").to_a
```
kosiakkatrina, line 92: I think this is correct, but I wonder why in 2024 it was ..field_135 instead of ..field_131

app/services/bulk_upload/sales/year2025/row_parser.rb
```diff
@@ -0,0 +1,2063 @@
...
+  include InterruptionScreenHelper
+  include FormattingHelper
+
+  QUESTIONS = {
+    field_1: "Which organisation owned this property before the sale?",
+    field_2: "Which organisation is reporting this sale?",
+    field_3: "Username",
+    field_4: "What is the day of the sale completion date? - DD",
+    field_5: "What is the month of the sale completion date? - MM",
+    field_6: "What is the year of the sale completion date? - YY",
```
kosiakkatrina: If I'm looking at the correct file I think 1, 2, 3 and 4, 5, 6 should be swapped?

app/services/bulk_upload/sales/year2025/row_parser.rb
```diff
@@ -0,0 +1,2063 @@
...
+            presence: {
+              message: I18n.t("#{ERROR_BASE_KEY}.not_answered", question: "staircasing transaction."),
+              category: :setup,
+              if: :shared_ownership?,
+            },
+            on: :after_log
+
+  validates :field_11,
+            inclusion: {
+              in: [8, 14, 27, 9, 29, 21, 22],
```
kosiakkatrina: Can this also be 27 and 29?

app/services/bulk_upload/sales/year2025/row_parser.rb
```diff
@@ -0,0 +1,2063 @@
...
+            },
+            on: :after_log
+
+  validates :field_18,
+            presence: {
+              message: I18n.t("#{ERROR_BASE_KEY}.not_answered", question: "type of building."),
+              category: :setup,
+              if: :shared_or_discounted_but_not_staircasing?,
+            },
+            on: :after_log
```
kosiakkatrina: I don't think we need any of the new not answered validations here, cause they should be covered by `validate :validate_nulls`
We likely just want to keep the ones that we had, cause they're either in the setup section (`category: :setup`) or they have slightly different way of validating when coming through BU as opposed to validations we have on the log model.

spec/services/bulk_upload/sales/log_creator_spec.rb
kosiakkatrina: Yeah, it looks like it's not necessarily year specific, I guess the only thing it's testing in terms of the years is that it's actually creating the correct parser etc, we're stubbing everything else. So technically we could add tests for different years just to test that and leave the rest of the tests just for a single year

spec/services/bulk_upload/sales/year2025/row_parser_spec.rb
kosiakkatrina: I think that's fine unless the new fields have quite specific validations that we want to test. I'd probably leave it as is for now and see if anything comes out of testing

app/views/bulk_upload_sales_logs/forms/prepare_your_file_2025.html.erb
carolynbarker: We decided to use the existing copy (but with year updated) for now, and ask Rachel for any copy changes which would be a separate ticket

app/helpers/bulk_upload/sales_log_to_csv.rb
carolynbarker: I've done this but will raise as a separate PR

app/services/bulk_upload/sales/year2025/row_parser.rb
carolynbarker: Fixed, good spot!

app/services/bulk_upload/sales/year2025/row_parser.rb
carolynbarker: There should indeed - I've added them, thanks

app/services/bulk_upload/sales/year2025/row_parser.rb
carolynbarker: You were absolutely right, I've updated them. There were quite a few places I had to update these, let's keep an eye out for any places that look suspiciously swapped

app/services/bulk_upload/sales/year2025/row_parser.rb
carolynbarker: yes, but I think they're already included as options? I copied this from the 2024 row parser, but it does seem to match the 2025 template too

carolynbarker: oh wait maybe I see what you're saying - should 27 and 29 *not* be included any more? they're listed in the template, but I don't know right now what options they correspond to

app/services/bulk_upload/sales/year2025/row_parser.rb
carolynbarker: absolutely, thanks. I hadn't read through the validation logic in the row_parser.

It seems weird that field_115 is the only non setup validation here. It seems to throw a soft validation on the log model. Are there any other fields that would need a separate validation here?

app/services/bulk_upload/sales/year2025/row_parser.rb
kosiakkatrina: Oh, sorry, I think I just missed it cause it was out of order!

app/services/bulk_upload/sales/year2025/row_parser.rb
kosiakkatrina: Yeah, I think it wasn't getting passed on from the model for some reason so it got added here in CLDC-3318, maybe there was another way to do it. We have other non setup validations here, but they're all defined as methods, we could extract this as a method, but I don't think it makes much of a difference

app/services/bulk_upload/sales/year2025/row_parser.rb
carolynbarker: No worries! I've reordered it to be clearer

carolynbarker: I've put them in a more sensible order!

app/services/bulk_upload/sales/year2025/csv_parser.rb
carolynbarker: It seems to have just been copied from 2023, when there were 131 fields.

app/views/bulk_upload_sales_logs/forms/prepare_your_file_2025.html.erb
```diff
@@ -0,0 +1,36 @@
...
+  <div class="govuk-grid-column-two-thirds">
+    <%= form_with model: @form, scope: :form, url: bulk_upload_sales_log_path(id: "prepare-your-file"), method: :patch do |f| %>
+      <%= f.hidden_field :year %>
+      <%= f.hidden_field :organisation_id %>
+
+      <span class="govuk-caption-l">Upload sales logs in bulk (<%= @form.year_combo %>)</span>
+      <h1 class="govuk-heading-l">Prepare your file</h1>
+      <p class="govuk-body govuk-!-margin-bottom-2"><%= govuk_link_to "Read the full guidance", bulk_upload_sales_log_path(id: "guidance", form: { year: @form.year }, referrer: "prepare-your-file") %> before you start if you have not used bulk upload before.</p>
+
+      <h2 class="govuk-heading-s">Download template</h2>
```
Dinssa: Is the only difference between this prepare your file and the 2024 one the dates? Can this be turned into something reusable? Based off the current collection year selected we could do that.

carolynbarker: Good question! I initially left this separate because I wasn't sure whether there's any new copy planned for the prepare your file section. But I don't think we are, and it looks like the text was the same except for the dates in 2013 as well, so I agree let's make it reusable

app/services/bulk_upload/sales/year2025/row_parser.rb
```diff
@@ -373,6 +373,7 @@ class BulkUpload::Sales::Year2025::RowParser
               greater_than_or_equal_to: 0,
               less_than_or_equal_to: 70,
               if: :discounted_ownership?,
+              allow_blank: true,
```
carolynbarker, line 376: Currently, uploading a bulk upload row with a discounted sale under another right to buy scheme will fail with a critical error if discount field is empty. But we only require discount to be present when it's a right-to-buy-like discount scheme - for other discount schemes this is blanked if provided. So, this numericality validation should only apply when the field is present.

If the field is absent when discount should be provided, this will be caught by other validations.

I think the same applies for 2024 so have made this change there too.

app/views/bulk_upload_sales_logs/forms/prepare_your_file.html.erb
```diff
@@ -14,15 +14,15 @@
```
kosiakkatrina, line 14: Did we confirm these pages would now mostly be the same between the years?

kosiakkatrina (approved):

PR #2948, opened 2025-02-17
CLDC-3845 Save manually entered address in BU

kosiakkatrina:
This updates the BU flow for addresses to following:
- UPRN is given - prioritise and use UPRN (same as before)
- No UPRN, address with high enough confidence can be found - set the found address (same as before)
- No UPRN, no address over confidence threshold - use the manual entry fields
  - Some fields are missing - add error saying address fields are missing
  - All the manual entry fields exist - no error


Dinssa (approved):

app/models/log.rb
```diff
@@ -90,13 +90,27 @@ def process_address_change!
...
         os_match_threshold_for_bulk_upload = 0.7
         if presenter.match >= os_match_threshold_for_bulk_upload
           self.uprn_selection = presenter.uprn
         else
+          self.postcode_full = postcode_full_as_entered
+          self.address_line1 = address_line1_as_entered
+          self.address_line2 = address_line2_as_entered
+          self.county = county_as_entered
+          self.town_or_city = town_or_city_as_entered
+          self.manual_address_entry_selected = true
```
carolynbarker: we could factor this out as it's the same as above

app/services/feature_toggle.rb
```diff
@@ -4,7 +4,7 @@ def self.allow_future_form_use?
   end
 
   def self.bulk_upload_duplicate_log_check_enabled?
-    !Rails.env.staging?
+    !Rails.env.staging? && !Rails.env.review?
```
carolynbarker: I wondered whether we should just go `Rails.env.production?`, but I think it's probably best to keep if enabled for local dev initially as a reminder that it's there (and people can disable it here when needed, of course)

app/models/log.rb
```diff
@@ -90,13 +90,27 @@ def process_address_change!
...
+          return nil
+        end
 
         presenter = AddressDataPresenter.new(service.result.first)
         os_match_threshold_for_bulk_upload = 0.7
         if presenter.match >= os_match_threshold_for_bulk_upload
           self.uprn_selection = presenter.uprn
```
carolynbarker, line 101: Can `presenter.uprn` be "uprn_not_listed"? If so I think the address would be cleared later in this method (but I'm not convinced the UPRN client would return that string, it looks like something we're doing our side)

carolynbarker:
Looks good to me! Left a couple of tiny comments

carolynbarker (approved):

app/models/log.rb
kosiakkatrina: Yep! Good shout!

app/services/feature_toggle.rb
kosiakkatrina: Ah It's only added it for testing, I don't think we actually want it to be disabled on review apps. Removed it now!

app/models/log.rb
kosiakkatrina: I believe this is only done on our side, presenter should only return an actual uprn from the API for this

PR #2951, opened 2025-02-19
Update alpine and ruby versions

kosiakkatrina:


Dinssa (approved):

Dockerfile
```diff
@@ -10,7 +10,7 @@ RUN apk add --update --no-cache tzdata && \
 # build-base: compilation tools for bundle
 # yarn: node package manager
 # postgresql-dev: postgres driver and libraries
-RUN apk add --no-cache build-base=0.5-r3 busybox=1.36.1-r7 nodejs-current=20.8.1-r0 yarn=1.22.19-r0 postgresql13-dev=13.18-r0 bash=5.2.15-r5
+RUN apk add --no-cache build-base=0.5-r3 busybox=1.36.1-r29 nodejs=20.15.1-r0 yarn=1.22.22-r0 bash=5.2.26-r0 libpq-dev
```
kosiakkatrina, line 13: postgresql13-dev doesn't seem to be on alpine3.20 and above, but it needs libpq-dev or postgresql-devel to be installed (not sure if it matters which one).
Also, nodejs-current is 23.x.x, but I'm not sure if we care about current, so changed that to just nodejs

.ruby-version
```diff
@@ -1 +1 @@
-3.1.4
+3.1.6
```
kosiakkatrina, line 1: CVE-2025-26519 is only fixed on ruby alpine 3.20 and above, which only exists for ruby 3.1.6, so updated the ruby version slightly

PR #2955, opened 2025-02-20
CLDC-3860: Bulk upload add validation, prevent reasonpref dontknow being selected

Dinssa:


app/services/bulk_upload/lettings/year2024/row_parser.rb
carolynbarker, line 1: This looks good to me! I think we also want the same changes for 2025

spec/lib/tasks/recalculate_invalid_reasonpref_dontknow_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+    task.invoke
+    pre_2024_invalid_logs.each do |log|
+      log.reload
+      expect(log.reasonpref).to eq(1)
+      expect(log.rp_dontknow).to eq(1)
+      expect(log.rp_homeless).to eq(1)
+    end
+  end
+
+  it "does not update the logs with valid rp_dontknow values" do
```
carolynbarker, line 53: We can check the `updated_at` field here to verify that the logs haven't been updated

Dinssa: Good idea, i'll add that in

app/services/bulk_upload/lettings/year2024/row_parser.rb
Dinssa: Ye we do, once the 2025 PR is merged we can update it with this or vice versa

app/services/bulk_upload/lettings/year2024/row_parser.rb
```diff
@@ -748,6 +749,23 @@ def validate_reasonable_preference_homeless
...
   end
 
+  def validate_reasonable_preference_dont_know
+    other_reason_fields = %i[field_107 field_108 field_109 field_110]
+    if field_106 == 1
+      selected_reasons = other_reason_fields.select { |field| send(field) == 1 }
+      dont_know_selected = field_111 == 1
+
+      if selected_reasons.any? && dont_know_selected
+        block_log_creation!
```
kosiakkatrina: Do we really want to block log creation in this case? I assume we should still be able to create logs, but would clear the reasonable preference reasons?

Dinssa: Ok sure let's do that, will update it

app/services/bulk_upload/lettings/year2024/row_parser.rb
```diff
@@ -1288,11 +1280,11 @@ def attributes_for_log
...
-    attributes["rp_homeless"] = field_107
-    attributes["rp_insan_unsat"] = field_108
-    attributes["rp_medwel"] = field_109
-    attributes["rp_hardship"] = field_110
-    attributes["rp_dontknow"] = field_111
+    attributes["rp_homeless"] = field_107 unless rp_dontknow_conflict?
+    attributes["rp_insan_unsat"] = field_108 unless rp_dontknow_conflict?
+    attributes["rp_medwel"] = field_109 unless rp_dontknow_conflict?
+    attributes["rp_hardship"] = field_110 unless rp_dontknow_conflict?
+    attributes["rp_dontknow"] = field_111 unless rp_dontknow_conflict?
```
kosiakkatrina, line 1287: Do you know if "not answered" errors would also get added to these fields? I'd expect no (I think validate_nulls skips adding errors to attributes that already have errors), but might be worth checking.

kosiakkatrina (approved):

app/services/bulk_upload/lettings/year2024/row_parser.rb
Dinssa: It doesn't add the not answered error to the fields, just the Don't Know conflict error message and users can fix that in service. Just incase though, I've swapped the order of validate methods but even before that it didn't add the "not answered" error.

PR #2957, opened 2025-02-24
Use date instead of time in the status methods

kosiakkatrina:


spec/models/location_spec.rb
```diff
@@ -1018,14 +1018,12 @@
       it "returns reactivating soon if the location has a future reactivation date" do
         deactivation_period = FactoryBot.create(:location_deactivation_period, deactivation_date: Time.zone.yesterday, reactivation_date: Time.zone.tomorrow, location:)
         location.save!
-        expect(deactivation_period.deactivation_date).to eq(Time.zone.yesterday)
```
kosiakkatrina, line 1021: I've added these when trying to fix these tests a few weeks ago, so just removing them again

PR #2958, opened 2025-02-25
CLDC-3843: Lettings household situation change Q84 question format

Dinssa:
This PR updates the formatting of Q84 (“What was the source of referral for this letting?”). The new format presents a consistent question to all users, categorises referral reasons, and directs users to a relevant sub-question for further specification. This replaces the previous format, which displayed a single page with a long list of referral reasons that varied based on previous answers. The new format aligns more closely with the paper form.


app/models/form/lettings/pages/referral_general_needs.rb
```diff
@@ -0,0 +1,12 @@
...
+    super
+    @id = "referral_prp"
+    @copy_key = "lettings.household_situation.referral.general_needs.prp"
+    @depends_on = [{ "owning_organisation_provider_type" => "PRP", "needstype" => 1, "renewal" => 0 }]
+  end
+
+  def questions
+    @questions ||= [Form::Lettings::Questions::ReferralGeneralNeeds.new(nil, nil, self)]
+  end
+end
```
Dinssa: Renamed the old 2024 `referral_prp` page & question to `referral_general_needs` to reuse the name for the new 2025 version.

app/models/form/lettings/questions/referral_type.rb
```diff
@@ -0,0 +1,38 @@
...
+        "value" => "Police, probation, prison or youth offending team",
+      },
+      "7" => {
+        "value" => "Voluntary agency",
+      },
+      "16" => {
+        "value" => "Other",
+      },
+    }.freeze
+  end
```
Dinssa: The values 101-5 are intermediary values stored in referral_type to show the right sub page. I chose to start from 100 to throw the numbering well away from the numbers we'd expect for each referral type. 

7 & 16 are the actual numbers for each of those referral types, which do not have a sub page. It will be assigned to the referral attribute when in our derived variables.

app/models/derived_variables/lettings_log_variables.rb
```diff
@@ -208,6 +212,9 @@ def clear_inapplicable_derived_values!
     if form.start_year_2024_or_later? && (unittype_gn_changed? && unittype_gn_was == 2)
       self.beds = nil
     end
+    if referral_type_changed? && (referral_type_was == 7 || referral_type_was == 16)
+      self.referral = nil
+    end
```
kosiakkatrina: What would happen if it changed from 7 to 16? Can it be cleared by routing rules?

app/models/form/lettings/pages/referral_general_needs.rb
kosiakkatrina: But the other referral is also used for general needs? Maybe ReferralGeneralNeedsPrp?

app/models/form/lettings/questions/referral_general_needs.rb
```diff
@@ -0,0 +1,105 @@
+class Form::Lettings::Questions::ReferralGeneralNeeds < ::Form::Question
```
kosiakkatrina, line 1: I assume this is also just a rename?

app/models/form/lettings/questions/referral_type.rb
kosiakkatrina: Does referral type has to directly map onto referral?
Is there benefits/drawbacks to have these numbered independently (for example 1 - 7) and map when we're setting the referral where relevant?

app/models/derived_variables/lettings_log_variables.rb
kosiakkatrina: Is there a test for this?

app/models/form/lettings/pages/referral_general_needs.rb
Dinssa: Ok will change it to that

app/models/form/lettings/questions/referral_type.rb
Dinssa: Okay good idea, I'll change it up to work like this.

Dinssa: Updated to 1-7 and referral as 7 or 16 is derived if referral type is 6 or 7.

app/models/derived_variables/lettings_log_variables.rb
Dinssa: Reworked this to not check for 7 or 16 and just reset referral if referral type is changed.

app/models/form/lettings/questions/referral_general_needs.rb
Dinssa: Renamed this to ReferralGeneralNeedsPrp but also Referral to ReferralGeneralNeeds to match the naming of the SupportHousing versions.

app/models/form/lettings/subsections/household_situation.rb
```diff
@@ -21,11 +21,28 @@ def pages
...
+        Form::Lettings::Pages::ReferralJustice.new(nil, nil, self),
+      ]
+    else
+      [
+        Form::Lettings::Pages::ReferralGeneralNeeds.new(nil, nil, self),
+        Form::Lettings::Pages::ReferralGeneralNeedsPrp.new(nil, nil, self),
+        Form::Lettings::Pages::ReferralSupportedHousing.new(nil, nil, self),
+        Form::Lettings::Pages::ReferralSupportedHousingPrp.new(nil, nil, self),
+      ]
+    end
```
Dinssa, line 46: @kosiakkatrina ReferralGeneralNeedsPrp is what I previously called ReferralGeneralNeeds when you last reviewed it. I've also changed Referral to ReferralGeneralNeeds (it is for the LA version just like ReferralSupportedHousing).

app/models/derived_variables/lettings_log_variables.rb
Dinssa: Removed this, wasn't needed (I think because the values are cleared when the page is not routed to) & caused some other tests to fail.

app/models/form/lettings/pages/referral_type.rb
```diff
@@ -0,0 +1,10 @@
+class Form::Lettings::Pages::ReferralType < ::Form::Page
+  def initialize(id, hsh, subsection)
+    super
+    @id = "referral"
```
kosiakkatrina: nitpicky, but would it make sense for this ID to be `referral_type` so that it's not confused with the other `referral` page?

kosiakkatrina (approved):

PR #2959, opened 2025-02-26
CLDC-3876: Household characteristic ecstat of child under 16 bug fix when age changed

Dinssa:
This PR resolves an issue where users are unable to change a person’s age from under 16 to 16 or above. Currently, the form derives a working situation of “child under 16” for individuals under 16 and skips the working situation question. However, when users attempt to update the age to 16 or older, they encounter an error because the “child under 16” status is no longer valid. This fix ensures that users can successfully update the age and that the working situation question is appropriately prompted for individuals aged 16 and above.


carolynbarker (approved):
Looks good to me!

PR #2969, opened 2025-03-05
CLDC-3897 Re-export merged users

kosiakkatrina:
We use `update_all` to update users when they get moved into another organisation during a merge, which means `updated_at` values don't get updated.
We use `updated_at` values to determine which users need to be re-exported
This PR adds `values_updated_at` to track any manual updates (same as we do in lettings logs) and trigger the re-export

### DIFF
--- a/app/services/exports/user_export_service.rb
+++ b/app/services/exports/user_export_service.rb
@@ -28,7 +28,7 @@ module Exports
     def retrieve_resources(recent_export, full_update, _year)
       if !full_update && recent_export
         params = { from: recent_export.started_at, to: @start_time }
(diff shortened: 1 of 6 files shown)


carolynbarker:
Why do we have a separate values_updated_at field rather than setting updated_at?

carolynbarker (approved):
Looks good to me!

kosiakkatrina:
> Why do we have a separate values_updated_at field rather than setting updated_at?

updated_at is an active record timestamp, so it gets set automatically when .update gets called on the record. I don't think we should be manually setting it, cause it might make it confusing if we're doing any investigation into who updated what, so for manual updates we usually use values_updated_at

PR #2970, opened 2025-03-06
CLDC-3904 remove default from relat mapping

carolynbarker:
Adjust relation mapping which was mapping to R by default. This mapping happens before we validate that the answer is a valid radio option, so it should not do this. It now maps 1,2,3 correctly and gives a validation error for other inputs.


kosiakkatrina (approved):

carolynbarker:
Reopened as https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/2993

PR #2972, opened 2025-03-06
CLDC-3905 BU Sales 25/26 Buyer relationship validation error (student/child check)

Dinssa:
We changed the buyer relationship question from previously asking if the additional people were a 'Partner', 'Child', 'Other' or 'Prefers not to say' to just asking if the person is a partner 'yes' or 'no'. As we no longer know if the person is a child the validation checking that the person has a working situation of full time student against their status as a child is no longer needed and this PR removes that.


app/services/bulk_upload/sales/year2025/row_parser.rb
```diff
@@ -662,7 +662,7 @@ def field_mapping_for_errors
       relat2: %i[field_34],
       relat3: %i[field_42],
       relat4: %i[field_46],
-      relat5: %i[field_49],
+      relat5: %i[field_50],
```
carolynbarker, line 665: Good spot, thanks!

carolynbarker (approved):
Looks good to me!

PR #2977, opened 2025-03-10
CLDC-3913: Address search selected options bug

Dinssa:


app/models/form/lettings/questions/address_search.rb
```diff
@@ -12,11 +12,10 @@ def initialize(id, hsh, page)
 
   def answer_options(log = nil, _user = nil)
     return {} unless ActiveRecord::Base.connected?
-    return {} unless log&.address_options&.any?
+    return {} unless log&.address_search_options&.any?
 
-    log.address_options.each_with_object({}) do |option, hash|
-      hash[option[:uprn]] = { "value" => "#{option[:address]} (#{option[:uprn]})" }
-    end
+    option = log.address_search_options.first
```
kosiakkatrina: Is this a useless assignment?

kosiakkatrina (commented):
Can we add a test that shows what the bug is?

app/models/form/lettings/questions/address_search.rb
Dinssa: Kept this as a list but yes I can remove the assignment.

kosiakkatrina (approved):

PR #2981, opened 2025-03-10
Update address search results

kosiakkatrina:
`[uprn_service.result] || []` it would result in [nil] if it's empty and would attempt to join that `address_service.result` and fail

### DIFF
--- a/app/controllers/address_search_controller.rb
+++ b/app/controllers/address_search_controller.rb
@@ -40,7 +40,7 @@ class AddressSearchController < ApplicationController
       address_service.call
       uprn_service.call
 
-      results = (uprn_service.result.present? ? [uprn_service.result] : []) + (address_service.result.compact.any? ? address_service.result : [])
+      results = [uprn_service.result, *address_service.result].compact
 
       if address_service.error.present? && uprn_service.error.present?
         render json: { error: "Address and UPRN are not recognised." }, status: :not_found
--- a/app/controllers/address_search_controller.rb
+++ b/app/controllers/address_search_controller.rb
@@ -38,7 +38,7 @@ class AddressSearchController < ApplicationController
       address_service.call
       uprn_service.call
 
-      results = ([uprn_service.result] || []) + (address_service.result || [])
+      results = (uprn_service.result.present? ? [uprn_service.result] : []) + (address_service.result.compact.any? ? address_service.result : [])
 
       if address_service.error.present? && uprn_service.error.present?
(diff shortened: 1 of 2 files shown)


carolynbarker (approved):

spec/requests/address_search_controller_spec.rb
```diff
@@ -144,5 +144,45 @@
...
+
+        it "returns the address results" do
+          get "/address-search?query=100"
+
+          expect(response).to have_http_status(:ok)
+          expect(response.body).to eq([{ text: nil, value: "321" }].to_json)
+        end
+      end
+    end
   end
```
Dinssa: If there is no address returned by UPRN search then `uprn_body = { results: [{ DPA: { UPRN: "321" } }] }.to_json`
should be `uprn_body = { results: [{ DPA: { } }] }.to_json `and we'd not expect to have text or value.

kosiakkatrina: So I think this scenario tests that there is a UPRN result but not the address result. What would we expect if there is a uprn result?

Dinssa: If there is a UPRN result there will always be an address part to that result, from the UPRN Client.

Dinssa: Our issue was the joining of the two different clients (UPRN and ADDRESS) each of which either return full results with the address and uprn parts or none.

kosiakkatrina: Ah, okay, updated the tests now, thank you!

Dinssa: To model this, we'd have for example two different results, one result from the UPRN client and one from the ADDRESS client, if a URPN result comes back the first item in the results list would be an object with an address and uprn key, the second item would be an empty object (or this would just not be appended to the results list).

app/controllers/address_search_controller.rb
```diff
@@ -38,7 +38,7 @@ def index
       address_service.call
       uprn_service.call
 
-      results = ([uprn_service.result] || []) + (address_service.result || [])
+      results = (uprn_service.result.present? ? [uprn_service.result] : []) + (address_service.result.compact.any? ? address_service.result : [])
```
Dinssa: What do you think of this?
`results = [uprn_service.result, *address_service.result].compact`

compact removes the nil results, so should remove the nil uprn result we saw, before it is passed to the next step

kosiakkatrina: Yep, nice!

PR #2982, opened 2025-03-10
CLDC-3911 Limit CSV variables to those defined in or before year

carolynbarker:
Previously we could have eg a variable defined in 2023 and 2025 but not 2024, and the 2025 one would be used for 2024.

They are grouped by both name and description so this is not actually as important as if they'd been grouped only by name like I initially thought, but worth resolving anyway.

### DIFF
--- a/app/services/csv/lettings_log_csv_service.rb
+++ b/app/services/csv/lettings_log_csv_service.rb
@@ -284,8 +284,7 @@ module Csv
     end
 
     def lettings_log_definitions
-      CsvVariableDefinition.lettings.group_by { |record| [record.variable, record.definition] }
-                           .map { |_, options|
+      CsvVariableDefinition.lettings.group_by(&:variable).map { |_, options|
         exact_match = options.find { |definition| definition.year == @year }
         next exact_match if exact_match
 
--- a/app/services/csv/lettings_log_csv_service.rb
+++ b/app/services/csv/lettings_log_csv_service.rb
@@ -285,12 +285,12 @@ module Csv
 
     def lettings_log_definitions
       CsvVariableDefinition.lettings.group_by { |record| [record.variable, record.definition] }
-                           .map do |_, options|
+                           .map { |_, options|
         exact_match = options.find { |definition| definition.year == @year }
         next exact_match if exact_match
 
         options.select { |opt| opt.year < @year }.max_by(&:year)
-      end
+      }.compact
     end
 
     def insert_derived_and_related_attributes(ordered_questions)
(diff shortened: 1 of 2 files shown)


app/services/csv/lettings_log_csv_service.rb
```diff
@@ -285,12 +285,12 @@ def lettings_log_attributes
...
     def lettings_log_definitions
       CsvVariableDefinition.lettings.group_by { |record| [record.variable, record.definition] }
-                           .map do |_, options|
+                           .map { |_, options|
         exact_match = options.find { |definition| definition.year == @year }
         next exact_match if exact_match
 
-        options.max_by(&:year)
-      end
+        options.select { |opt| opt.year < @year }.max_by(&:year)
```
kosiakkatrina, line 291: Is there a situation where there are no descriptions for years before the one that we're downloading for? For example atm we can download 2023 logs, do we have descriptions for those or only the ones starting 2024?

Dinssa: The variable definitions start from 2023 so we have everything a user would need for that year.

kosiakkatrina (approved):

app/services/csv/lettings_log_csv_service.rb
```diff
@@ -285,12 +285,12 @@ def lettings_log_attributes
 
     def lettings_log_definitions
       CsvVariableDefinition.lettings.group_by { |record| [record.variable, record.definition] }
-                           .map do |_, options|
+                           .map { |_, options|
         exact_match = options.find { |definition| definition.year == @year }
         next exact_match if exact_match
```
Dinssa: I think your initial instinct was the right one, this method should have been just grouping by variable not the variable & definition. By using this functionality people will change the definition and we'll not actually group together the right records. Two definitions for the same variable could be returned by this method because they were grouped separately and it'd just be the first one found that gets outputted. When testing this I was not able to get the new 2025 definitions and kept getting the old ones until removing `record.definition` to:

```
    def sales_log_definitions
      CsvVariableDefinition.sales.group_by { |record| [record.variable] }
                           .map { |_, options|
        exact_match = options.find { |definition| definition.year == @year }
        next exact_match if exact_match

        options.select { |opt| opt.year < @year }.max_by(&:year)
      }.compact
    end
```
or:

```
def sales_log_definitions
  CsvVariableDefinition.sales.group_by(&:variable).map { |_, options|
    exact_match = options.find { |definition| definition.year == @year }
    next exact_match if exact_match

    options.select { |opt| opt.year < @year }.max_by(&:year)
  }.compact
end
```
same for lettings

Dinssa (commented):
I ran the rake task to add a startup list of all the variable/definition records you should be able to test this in the review app, download CSVs and add new records in the '/admin' page.

app/services/csv/lettings_log_csv_service.rb
carolynbarker: Great spot! Yes absolutely, this doesn't look intentional and it doesn't work with our output logic. I'll change it as discussed. Nice one Manny!

PR #2984, opened 2025-03-11
CLDC-3917: Sales BU numstair question range error message

Dinssa:


app/views/form/_date_question.html.erb
```diff
@@ -1,9 +1,10 @@
...
+<% legend = legend(question, page_header, conditional) %>
 <%= render partial: "components/date_picker", locals:
         {
           resource: @log,
           question_id: question.id,
-          legend: { text: legend(question, page_header, conditional)[:text], size: "l", caption: caption(caption_text, page_header, conditional) },
+          legend: { text: legend[:text], size: legend[:size], caption: caption(caption_text, page_header, conditional) },
           resource_type: @log.log_type,
           hint: (question.hint_text.blank? ? "" : (question.hint_text.html_safe + "</br></br>".html_safe)) + "For example, #{date_mid_collection_year_formatted(@log.startdate).tr(' ', '/')}",
           f:,
```
Dinssa, line 10: This improvement addresses an unrelated issue where the date question’s heading is always displayed in a large font, regardless of the page it’s on. When the date question is the only question on a page, the large font is appropriate. However, when it’s one of several questions on a page, the heading should be medium-sized. This change incorporates our existing calculation for the legend’s size. The issue is evident in the sales 2025 staircasing transaction section, where three questions ask about the staircasing history. Two of these questions are dates and appear comically large. The saledate question (single question on a single page) remains unchanged.

app/services/bulk_upload/sales/year2025/row_parser.rb
```diff
@@ -393,6 +393,15 @@ class BulkUpload::Sales::Year2025::RowParser
...
 
+  validates :field_103,
+            numericality: {
+              greater_than_or_equal_to: 2,
+              less_than_or_equal_to: 10,
+              message: I18n.t("#{ERROR_BASE_KEY}.numeric.within_range", field: "Number of staircasing transactions", min: "2", max: "10"),
+              allow_blank: true,
+            },
+            on: :before_log
+
```
kosiakkatrina, line 404: Why doesn't the min/max from question definition not trigger at `validate_numeric_min_max` ?

Dinssa: Hm, it looks like before the value gets to that method it is cleared and =nil instead of =1. Stuck a binding.pry in and checked the value. Another numeric question with min and max values is also added directly into the row parser like below, so I'm thinking we need to for values that do not fit the min/max within the question model and are cleared:

```
  validates :field_115,
            numericality: {
              message: I18n.t("#{ERROR_BASE_KEY}.numeric.within_range", field: "Percentage discount", min: "0%", max: "70%"),
              greater_than_or_equal_to: 0,
              less_than_or_equal_to: 70,
              if: :discounted_ownership?,
              allow_blank: true,
            },
            on: :before_log
```

kosiakkatrina (approved):

PR #2987, opened 2025-03-11
CLDC-3919: Sales BU Decimal numbers rounding errors

Dinssa:
This PR was initially intended to fix the issue where users were unable to upload bulk uploads with discount percentages to 1 decimal place. However, I discovered other fields that we currently allow in the service to be entered as decimals, which would also benefit from this fix. I used the scheme.rb and question models to verify that they accept decimals. I've separated out each group of changes should we need to revert any.

### DIFF
--- a/app/services/bulk_upload/sales/year2024/row_parser.rb
+++ b/app/services/bulk_upload/sales/year2024/row_parser.rb
@@ -252,8 +252,8 @@ class BulkUpload::Sales::Year2024::RowParser
   attribute :field_84, :integer
   attribute :field_85, :integer
   attribute :field_86, :integer
-  attribute :field_87, :integer
-  attribute :field_88, :integer
+  attribute :field_87, :decimal
+  attribute :field_88, :decimal
   attribute :field_89, :integer
 
   attribute :field_90, :integer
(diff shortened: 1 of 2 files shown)


kosiakkatrina (commented):
I think some of these should be full numbers, for example purchase price:
image

But I think this is correct, cause we probably want to surface this error, instead of just rounding (?)

Could we add some tests that check that the correct error gets added in these situations? I think fields like purchase price, mortgage amount and deposit amount should all be whole numbers.

Also, did we want to update that for 2024?

Dinssa:
> I think some of these should be full numbers, for example purchase price: image
> 
> But I think this is correct, cause we probably want to surface this error, instead of just rounding (?)
> 
> Could we add some tests that check that the correct error gets added in these situations? I think fields like purchase price, mortgage amount and deposit amount should all be whole numbers.
> 
> Also, did we want to update that for 2024?

Looks like the correct error does get added, as long as the "step" is defined on the question model the right error message will appear...

For a shared ownership scheme:
Screenshot 2025-03-13 at 13 21 59


For a discounted ownership scheme:
Screenshot 2025-03-13 at 13 17 27

I'll add the same changes for 2024.

carolynbarker:
These fields look right to me! I think it's worth checking that the step error is shown correctly for each of them. Particularly thinking about the staircasing fields which are new and might not have been tested (but probably best to go through the whole list anyway)

kosiakkatrina (approved):

PR #2989, opened 2025-03-12
Add error handling for JSON parsing in UPRN client

Dinssa:


carolynbarker (approved):

app/services/uprn_client.rb
```diff
@@ -20,7 +20,13 @@ def call
   end
 
   def result
-    @result ||= JSON.parse(response.body).dig("results", 0, "DPA") || JSON.parse(response.body).dig("results", 0, "LPI")
+    @result ||= begin
```
carolynbarker: We could check response.status before trying to parse - I'd expect the html error page to have an error status like 500

app/services/uprn_client.rb
```diff
@@ -20,7 +20,13 @@ def call
   end
 
   def result
-    @result ||= JSON.parse(response.body).dig("results", 0, "DPA") || JSON.parse(response.body).dig("results", 0, "LPI")
+    @result ||= if response.code == 200
```
carolynbarker: We can use `response.is_a?(Net::HTTPSuccess)` instead of checking the number directly

app/services/uprn_client.rb
carolynbarker: Ugh sorry just noticed this comment was still in pending!

app/services/uprn_client.rb
```diff
@@ -20,7 +20,13 @@ def call
...
+                  parsed_response = JSON.parse(response.body)
+                  parsed_response.dig("results", 0, "DPA") || parsed_response.dig("results", 0, "LPI")
+                else
+                  Rails.logger.error("Unexpected response code: #{response.code}")
```
carolynbarker: We could also log response.body to help debug

carolynbarker (approved):

PR #2995, opened 2025-03-13
CLDC 3899 fix deactivation period edge case bugs

carolynbarker:
Fixes a couple of bugs around deactivation periods, which are likely the culprits for the flaky tests we've been seeing.

- for status_at, we were only checking the most recently created deactivation period for `:reactivating_soon` - if during any other deactivation period it was `:active`
- similarly the inactive_status message referred to the most recently created deactivation period, not the soonest one
- I wrote a bunch of tests to document the behaviour around deactivation periods as it wasn't clear to me
- I added error handling to the /reactivate endpoints if there is no open deactivation period to be ended (currently it's a 500 server error). I made these 404s, on the grounds that no open deactivation period was found, but plausibly these could be 422s, opinions welcome.


app/controllers/schemes_controller.rb
```diff
@@ -88,13 +88,22 @@ def deactivate
   end
 
   def new_reactivation
-    @scheme_deactivation_period = @scheme.scheme_deactivation_periods.deactivations_without_reactivation.first
+    open_deactivations = @scheme.scheme_deactivation_periods&.deactivations_without_reactivation
```
kosiakkatrina, line 91: Any reason this needs `&.` and the locations doesn't?

spec/requests/schemes_controller_spec.rb
```diff
@@ -2967,7 +2992,7 @@
       end
 
       context "when there is a later open deactivation" do
-        let(:deactivation_date) { Time.zone.local(2022, 10, 10) }
+        # let(:deactivation_date) { Time.zone.local(2022, 10, 10) }
```
kosiakkatrina: Is this meant to be commented out?

spec/requests/schemes_controller_spec.rb
```diff
@@ -2980,6 +3005,216 @@
...
+      let(:deactivation_date) { Time.utc(2022, 10, 10) }
+      let(:startdate) { Time.utc(2022, 10, 11) }
+      let(:params) { { scheme_deactivation_period: { reactivation_date: "5/8/2023", reactivation_date_type: "other" } } }
+
+      let(:add_deactivations) {}
+
+      before do
+        allow(FormHandler.instance).to receive(:lettings_in_crossover_period?).and_return(true)
+        Timecop.freeze(Time.utc(2023, 10, 10))
+        # Singleton.__init__(FormHandler)
```
kosiakkatrina: Do we need this?

spec/models/scheme_deactivation_period_spec.rb
```diff
@@ -69,5 +73,83 @@
...
+
+          it "does not add an error" do
+            validator.validate(record)
+            expect(record.errors).to be_empty
+          end
+        end
+      end
+    end
+
+    context "when there is not an open deactivation period within six months" do # validate_deactivation
```
kosiakkatrina, line 116: Would it make sense to run `validator.validate_deactivation` or is there a reason we're not doing that?

kosiakkatrina (approved):

spec/requests/schemes_controller_spec.rb
carolynbarker: I think I suspected we didn't, and commented it out to test this - will tidy this up

spec/requests/schemes_controller_spec.rb
carolynbarker: I think I suspected we didn't, and commented it out to test this - will tidy this up

spec/models/scheme_deactivation_period_spec.rb
carolynbarker: I wanted to test that the correct method validate_deactivation or validate_activation was routed to - in some situations it's not the one I expected so I wanted to investigate and document, and it seemed more important to test the overall behaviour of the validator rather than how specifically it does that.

Maybe we should test where it's routed to as a separate test and then test validate_deactivation and validate_activation separately.

There's useful context in the scheme/location_controller that explains why the scheme/location_deactivation_periods are assumed to be activations or deactivations when they are. I think I'll add a helpful comment at the period validate logic. I'm tempted to refactor that to move the decision logic on when to validate_deactivation or validate_activation to the scheme/location controller tbh.

spec/requests/schemes_controller_spec.rb
carolynbarker: Yep removed it - it seems to have been unused

app/controllers/schemes_controller.rb
carolynbarker: I've added it to locations as well

PR #2998, opened 2025-03-17
Update address logic for queries with non-digit characters

Dinssa:


kosiakkatrina (approved):
Test?

app/controllers/address_search_controller.rb
```diff
@@ -18,8 +18,8 @@ def index
         presenter = UprnDataPresenter.new(service.result)
         render json: [{ text: presenter.address, value: presenter.uprn }]
       end
-    elsif query.match?(/[a-zA-Z]/)
-      # Query contains letters, assume it's an address
+    elsif query.match?(/\D/)
```
carolynbarker, line 21: Do we also want to check the length? I'm not sure if the UPRN client errors on single digits or very long numbers

Dinssa: I don't think so, if a user is entering a UPRN value it'll be all be numbers so caught within the previous condition:
```
    elsif query.match?(/\A\d+\z/) && query.length > 5
      # Query is all numbers and greater than 5 digits, assume it's a UPRN
      service = UprnClient.new(query)
      service.call

      if service.error.present?
        render json: { error: service.error }, status: :not_found
      else
        presenter = UprnDataPresenter.new(service.result)
        render json: [{ text: presenter.address, value: presenter.uprn }]
      end
```

It'll search for UPRN and if the value is above 12 digits we'd get the same error as you would if the UPRN doesn't exist `@error = "UPRN is not recognised. Check the number, or enter the address." `. We won't send single digit values because we don't start using the api until someone has entered in 2 digits, but even then a single digit UPRN does exist ("1" in Bristol) so we shouldn't mind if they enter a single digit. The reason why we only start searching with 2 digits despite there being a single digit option is due to the unlikelihood of a user ever needing to pick that one.

PR #3001, opened 2025-03-17
CLDC-3930 CLDC-3931 CLDC-3932 CLDC-3933 CLDC-3934 add staircasing validations

carolynbarker:
Adds validations to new staircasing fields, including date validations for initial purchase and last transaction dates, and a validation that the total equity percentages make sense.

The error message for that is quite long so any suggestions for how to make it clearer to the user would be particularly appreciated!


app/models/validations/sales/financial_validations.rb
```diff
@@ -118,6 +118,29 @@ def validate_equity_less_than_staircase_difference(record)
     end
   end
 
+  def validate_staircase_difference_enough_for_numstair(record)
+    return unless record.equity && record.stairbought && record.stairowned && record.numstair
+
+    percentage_left = record.stairowned - record.stairbought - record.equity
+    return if percentage_left.negative? # In this case we show the validate_equity_less_than_staircase_difference validation
```
kosiakkatrina: Any reason we wouldn't want to show both? For example on BU uploads?

app/models/validations/sales/sale_information_validations.rb
```diff
@@ -41,6 +41,28 @@ def validate_staircasing_initial_purchase_date(record)
     if record.initialpurchase < Time.zone.local(1980, 1, 1)
       record.errors.add :initialpurchase, I18n.t("validations.sales.sale_information.initialpurchase.must_be_after_1980")
     end
+
+    if record.saledate.present? && record.initialpurchase > record.saledate
+      record.errors.add :initialpurchase, I18n.t("validations.sales.sale_information.initialpurchase.must_be_before_saledate")
+      record.errors.add :saledate, I18n.t("validations.sales.sale_information.saledate.must_be_after_initial_purchase_date")
```
kosiakkatrina: We might want to add `:skip_bu_error` to the saledate errors, so that it doesn't block log creation

app/models/validations/sales/financial_validations.rb
```diff
@@ -118,6 +118,29 @@ def validate_equity_less_than_staircase_difference(record)
     end
   end
 
+  def validate_staircase_difference_enough_for_numstair(record)
```
kosiakkatrina: I assume this won't run for 2024 cause numstair is not a question in 2024?

kosiakkatrina (approved):
I think the error messages are good cause they're quite explicit, consistent with some of our other financial validations

app/models/validations/sales/financial_validations.rb
carolynbarker: Just seemed like better UX to me rather than have two similar errors (but very happy to be persuaded!). In particular for staircasing transactions this error would always appear whenever the validate_equity_less_than_staircase_difference one appeared, which swamps it with a lot of text and complexity. What do we do elsewhere? Happy to show both if that's clearer / more in line with expected practice / etc

app/models/validations/sales/sale_information_validations.rb
carolynbarker: absolutely, ta

app/models/validations/sales/financial_validations.rb
carolynbarker: yes exactly!

app/models/validations/sales/financial_validations.rb
kosiakkatrina: Hmm, good question, I'm not sure if we do this elsewhere, but if we're confident we won't accidentally skip a validation, then we can leave as is

carolynbarker: I think I wasn't as confident as I would have liked - the existing equity validation actually only ran if saledate was provided and if year_2024_or_later (which should be the case if numstair is provided, but it's less clear to have different conditions and can lead to weird behaviour with weird logs). So I've refactored both into one method - it's a little long, but it's clearer and means we don't have to worry about matching up the conditions when each triggers

PR #3003, opened 2025-03-17
CLDC-3928 Update logs with manually entered address showing the property information as incomplete

Dinssa:


spec/lib/tasks/update_manual_address_entry_selected_prexisting_logs_spec.rb
```diff
@@ -0,0 +1,103 @@
...
+      before do
+        lettings_log_address_manually_entered.manual_address_entry_selected = false
+        lettings_log_address_manually_entered.save!(validate: false)
+        sales_log_address_manually_entered.manual_address_entry_selected = false
+        sales_log_address_manually_entered.save!(validate: false)
+      end
+
+      it "updates logs with an address manually entered" do
+        expect(lettings_log_address_manually_entered.manual_address_entry_selected).to be false
+        expect(lettings_log_address_manually_entered.address_line1).to eq("1 Test Street")
```
Dinssa, line 94: This test exhibits some flaky behaviour, on line `expect(lettings_log_address_manually_entered.manual_address_entry_selected).to be false` this will fail as despite setting `manual_address_entry_selected` to `false` when building the log or in the before-do it'll sometimes still be returned as `true`.

lib/tasks/update_manual_address_entry_selected_prexisting_logs.rake
```diff
@@ -0,0 +1,40 @@
+namespace :bulk_update do
+  desc "Update logs with specific criteria and set manual_address_entry_selected to true"
+  task update_manual_address_entry_selected: :environment do
+    lettings_logs = LettingsLog.filter_by_year(2024)
```
kosiakkatrina, line 6: What if it's a new log that got created and the user just didn't get to the uprn question? Should we only set the manual address entry selected if there are address fields present?

Dinssa: Good shout, i'll update this

lib/tasks/update_manual_address_entry_selected_prexisting_logs.rake
```diff
@@ -1,14 +1,20 @@
...
   desc "Update logs with specific criteria and set manual_address_entry_selected to true"
   task update_manual_address_entry_selected: :environment do
+    updated_lettings_logs_count = 0
+    updated_sales_logs_count = 0
     lettings_logs = LettingsLog.filter_by_year(2024)
                                .where(status: %w[in_progress completed])
                                .where(needstype: 1, manual_address_entry_selected: false, uprn: nil)
 
     lettings_logs.find_each do |log|
+      next unless log.address_line1 || log.address_line2 || log.county || log.town_or_city || log.postcode_full
```
kosiakkatrina: Can we add this into the query? I think doing it this way might take longer.

Something like `lettings_logs = LettingsLog.filter_by_year(2024).where(status: %w[in_progress completed]).where(needstype: 1, manual_address_entry_selected: false, uprn: nil).where("(address_line1 IS NOT NULL AND address_line1 != "") OR (address_line2 IS NOT NULL AND address_line2 != "") OR ...")`

Dinssa: Yes, updated.

Semi-related: I've kept the counting logic (variable and tally), although the change was needed because of this `next unless` it seems more reliable regardless.

kosiakkatrina (approved):

PR #3004, opened 2025-03-19
CLDC-3928++ Update logs with manually entered address showing the property information as incomplete

Dinssa:


lib/tasks/update_manual_address_entry_selected_prexisting_logs.rake
```diff
@@ -38,11 +54,64 @@ namespace :bulk_update do
...
+  desc "Find logs to fix and update postcode_full if conditions are met"
+  task update_postcode_full_preexisting_manual_entry_logs: :environment do
+    updated_count = 0
+    fixed_count = 0
+    not_updated_count = 0
+    not_updated_ids = []
+    updated_but_not_fixed_ids = []
+
+    logs_to_fix = LettingsLog.filter_by_year(2024).where(manual_address_entry_selected: true, uprn: nil, status: "in_progress", postcode_full: nil, updated_at: Time.zone.parse("2025-03-19 16:00:00")..Time.zone.parse("2025-03-19 17:00:00"))
+
```
Dinssa, line 170: There are currently 2 other logs that could match this query without the `updated_at` part, they're new logs that user's are likely still working on.

lib/tasks/update_manual_address_entry_selected_prexisting_logs.rake
```diff
@@ -38,11 +54,64 @@ namespace :bulk_update do
...
+    logs_to_fix.find_each do |log|
+      previous_version = log.versions[-2]
+      previous_status = previous_version&.reify&.status
```
Dinssa, line 173: Get the previous log status before the last rake task was run to compare against and check if the `postcode_full` fix has resolved and returned it back to what is should be.

lib/tasks/update_manual_address_entry_selected_prexisting_logs.rake
```diff
@@ -17,12 +22,23 @@ namespace :bulk_update do
...
+        if log.postcode_full.nil? && log.address_line1 == log.address_line1_input
+          log.postcode_full = log.postcode_full_input
+          log.save!
+        end
+        if log.status == status_pre_change
+          lettings_postcode_fixed_count += 1
+        else
+          Rails.logger.info "Status changed from #{status_pre_change} to #{status_post_change} for lettings log #{log.id}"
+          lettings_status_changed_log_ids << log.id
+        end
```
Dinssa: We attempt to apply the `postcode_full` fix to any log that may need it before saving. If the status is kept the same we consider it fixed, if the status is not kept the same we'll log it's ID for further review.

lib/tasks/update_manual_address_entry_selected_prexisting_logs.rake
```diff
@@ -17,12 +22,23 @@ namespace :bulk_update do
...
-        Rails.logger.info "Status changed from #{status_pre_change} to #{status_post_change} for lettings log #{log.id}"
+      if status_pre_change != status_post_change
+        if log.postcode_full.nil? && log.address_line1 == log.address_line1_input
+          log.postcode_full = log.postcode_full_input
+          log.save!
+        end
+        if log.status == status_pre_change
+          lettings_postcode_fixed_count += 1
+        else
+          Rails.logger.info "Status changed from #{status_pre_change} to #{status_post_change} for lettings log #{log.id}"
```
kosiakkatrina: could `status_post_change` != `log.status` here at all?

lib/tasks/update_manual_address_entry_selected_prexisting_logs.rake
```diff
@@ -17,12 +22,23 @@ namespace :bulk_update do
         Rails.logger.info "Could not save changes to lettings log #{log.id}"
       end
       status_post_change = log.status
-      unless status_pre_change == status_post_change
-        Rails.logger.info "Status changed from #{status_pre_change} to #{status_post_change} for lettings log #{log.id}"
+      if status_pre_change != status_post_change
```
kosiakkatrina: I wonder if it would make sense to fix these for logs that stayed in progress, but still don't have the postcode?

Dinssa: If it was already in progress, perhaps the user had not yet entered the postcode. But I guess if the `address_line1` and `address_line1_input` match we could also put this in for them. Minimise logs that need user input later.

lib/tasks/update_manual_address_entry_selected_prexisting_logs.rake
```diff
@@ -17,12 +22,23 @@ namespace :bulk_update do
...
+        else
+          Rails.logger.info "Status changed from #{status_pre_change} to #{status_post_change} for lettings log #{log.id}"
+          lettings_status_changed_log_ids << log.id
+        end
       end
     end
 
     puts "#{updated_lettings_logs_count} lettings logs updated."
+    puts "Lettings logs with status changes: [#{lettings_status_changed_log_ids.join(', ')}]"
+    puts "Lettings logs where postcode fix maintained status: #{lettings_postcode_fixed_count}"
```
kosiakkatrina: Would it make sense to distinguish between the logs where we updated the postcode and they stayed in progress vs the logs where we didn't even attempt to update the postcode because the address lines didn't match?

lib/tasks/update_manual_address_entry_selected_prexisting_logs.rake
```diff
@@ -37,12 +73,97 @@ namespace :bulk_update do
...
+        postcode_fixed = true
+        log.save!
+      end
+
+      if log.postcode_full.nil? && log.creation_method == "bulk upload" && log.address_line1 == log.address_line1_as_entered
+        log.postcode_full = log.postcode_full_as_entered
+        sales_postcode_fixed_count += 1
+        postcode_fixed = true
+        log.save!
+      end
```
Dinssa: Maybe we should not check the creation method? We'd want to take the address_line1_as_entered if it exists, regardless

lib/tasks/update_manual_address_entry_selected_prexisting_logs.rake
```diff
@@ -21,25 +30,36 @@ namespace :bulk_update do
       else
         Rails.logger.info "Could not save changes to lettings log #{log.id}"
       end
+
+      postcode_fixed = false
+      if log.postcode_full.nil? && log.address_line1 == log.address_line1_input
+        log.postcode_full = log.postcode_full_input
+        lettings_postcode_fixed_count += 1
+        postcode_fixed = true
+        log.save!
```
kosiakkatrina: What if this fails with a validation?

lib/tasks/update_manual_address_entry_selected_prexisting_logs.rake
```diff
@@ -21,25 +30,36 @@ namespace :bulk_update do
...
+      end
+
+      if log.postcode_full.nil? && log.creation_method == "bulk upload" && log.address_line1 == log.address_line1_as_entered
+        log.postcode_full = log.postcode_full_as_entered
+        lettings_postcode_fixed_count += 1
+        postcode_fixed = true
+        log.save!
```
kosiakkatrina: Same here

lib/tasks/update_manual_address_entry_selected_prexisting_logs.rake
kosiakkatrina: I don't think it matters, we only get _as_entered fields if its BU

kosiakkatrina (approved):

PR #3010, opened 2025-03-21
CLDC-3939 update tests for new FY

carolynbarker:
A few different updates:

- update tests for some rake tasks run in 2024, which assumed a 2024 log but didn't specify
- update tests that assumed 2024 routing logic 
- one test assumes time passes so can't be in a Rubocop.freeze - I have made this Rubocop.return beforehand so that the freeze method of testing tests still works, but we could also just do Rubocop.travel to test it instead. In actual April time will still pass so the test itself will pass then either way.
- update a validation to return if a field is null instead of erroring
- update shared ownership sale information section to not have duplicate page ids in 2025+
- refactor documentation generation slightly to handle pages/subsections better


app/models/validations/financial_validations.rb
```diff
@@ -159,9 +159,11 @@ def validate_rent_period(record)
   end
 
   def validate_care_home_charges(record)
+    return unless record.period && record.chcharge
+
     if record.is_carehome?
       period = record.form.get_question("period", record).label_from_value(record.period).downcase
```
carolynbarker, line 165: This was erroring when record.period is nil

spec/lib/tasks/correct_checkbox_values_spec.rb
```diff
@@ -9,6 +9,7 @@
     let(:user) { create(:user, organisation:) }
 
     before do
+      Timecop.return
```
carolynbarker: One of the tests here tests that the updated at time has changed. This line makes it pass even if using Timecop.freeze around it

spec/models/bulk_upload_spec.rb
```diff
@@ -29,10 +29,10 @@
...
+        expect(bulk_upload.fields_to_confirm(log)).to match_array %w[value_value_check monthly_charges_value_check percentage_discount_value_check income1_value_check income2_value_check combined_income_value_check retirement_value_check old_persons_shared_ownership_value_check buyer_livein_value_check wheel_value_check mortgage_value_check savings_value_check deposit_value_check staircase_bought_value_check stairowned_value_check hodate_check shared_ownership_deposit_value_check extrabor_value_check grant_value_check discounted_sale_value_check deposit_and_mortgage_value_check multiple_partners_value_check partner_under_16_value_check]
```
carolynbarker, line 35: student_not_child_value_check is removed in 2025 (because we no longer capture whether they are buyer 1's child)

spec/spec_helper.rb
```diff
@@ -119,6 +119,14 @@
...
   end
 
+  config.around do |example|
+    Timecop.travel(Time.zone.local(2025, 4, 1))
+    Singleton.__init__(FormHandler)
+    example.run
+    Timecop.travel(Time.zone.local(2025, 4, 1))
+    Singleton.__init__(FormHandler)
+  end
+
```
carolynbarker: I'm not planning on committing this! This is here  so I can use Github's pipeline to run the tests for me. I'll remove this before merging

app/models/validations/financial_validations.rb
kosiakkatrina: I don't think we ask for care home info in 2025, should we not run this test if it's 2025 log or later?

spec/lib/tasks/correct_checkbox_values_spec.rb
kosiakkatrina: I think this entire file and the task is probably okay to delete, cause it only needed to be ran once

app/services/documentation_generator.rb
```diff
@@ -301,7 +301,7 @@ def save_bu_validation(result, meth, form, log_type, field_mapping_for_errors)
   end
 
   def save_soft_validation(form, page, validation_descriptions, log_type)
-    subsection_pages = form.subsection_for_page(page).pages
+    subsection_pages = page.subsection.pages
     page_index = subsection_pages.index(page)
     page_the_validation_applied_to = subsection_pages[page_index - 1]
```
carolynbarker, line 306: For 2025 this errored because `page` wasn't in `form.subsection_for_page(page).pages`. Why? Because `form.subsection_for_page(page)` uses the `page.id`, and there were multiple pages with the same id.

I have changed the ids to be different, which should solve it, but I also suspect we don't need the `form.subsection_for_page` method, so I've removed it and replaced it with `page.subsection` everywhere. Either of these changes should fix this.

app/models/validations/financial_validations.rb
carolynbarker: IIRC I was getting here with a 2025 log, if so it would have passed `is_carehome?`. I'll have a look

carolynbarker: I think this change is good for other years too though, if record.period is nil (and record.is_carehome?) then line 165 will error in any year

app/helpers/bulk_upload/sales_log_to_csv.rb
```diff
@@ -434,7 +434,7 @@ def to_2025_row
       log.nationality_all_group,
       log.ecstat1,
       log.buy1livein,
-      log.relat2,
+      { "P" => 1, "X" => 2, "R" => 3 }[log.relat2],
```
carolynbarker, line 437: This is another actual functionality change, a fix to the relat fields in sales_log_to_csv (which is used for testing bulk uploads). The 2025 bulk upload maps 1,2,3 to P,X,R so we have to reverse this to recreate the csv field

spec/lib/tasks/correct_checkbox_values_spec.rb
carolynbarker: I've deleted a few old rake tasks that looked specific to a previous year

app/helpers/bulk_upload/sales_log_to_csv.rb
kosiakkatrina: Is that what we want in the export? I assume up until now we were exporting P, X, R?

app/views/bulk_upload_shared/guidance.html.erb
```diff
@@ -34,8 +34,8 @@
 
         <p class="govuk-body">You can paste your data below the headers or copy the headers and insert them above the data in your file. The bulk upload fields start at column B. Leave column A blank.</p>
         <p class="govuk-body">Make sure that each column of data aligns with the corresponding question in the headers. We recommend ordering your data to match the headers, but you can also reorder the headers to match your data. When processing the file, we check what each column of data represents based on the headers above.</p>
-        <p class="govuk-body"><%= govuk_link_to "Download the lettings bulk upload template (2024 to 2025)", @form.lettings_template_path %></p>
-        <p class="govuk-body"><%= govuk_link_to "Download the sales bulk upload template (2024 to 2025)", @form.sales_template_path %></p>
+        <p class="govuk-body"><%= govuk_link_to "Download the lettings bulk upload template (#{@form.year} to #{@form.year + 1})", @form.lettings_template_path %></p>
+        <p class="govuk-body"><%= govuk_link_to "Download the sales bulk upload template (#{@form.year} to #{@form.year + 1})", @form.sales_template_path %></p>
```
kosiakkatrina, line 38: nice

app/helpers/bulk_upload/sales_log_to_csv.rb
carolynbarker: Is this used for the export? I think this is used in the test data controller to produce the test file, and to create temporary test files for testing bulk upload. I'd expect both of those to update to the new 2025 input format, is there anywhere else where this is used where we don't want to change it?

kosiakkatrina: Omg you're right! I thought this was the csv export 🤦 Ignore me!

kosiakkatrina (approved):

app/helpers/bulk_upload/sales_log_to_csv.rb
carolynbarker: No problem! Always better to check

PR #3011, opened 2025-03-21
CLDC-3945 Update validation generation

kosiakkatrina:


lib/tasks/generate_lettings_documentation.rake
```diff
@@ -1,57 +1,61 @@
 namespace :generate_lettings_documentation do
   desc "Generate documentation for hard lettings validations"
-  task describe_lettings_validations: :environment do
+  task :describe_lettings_validations, %i[year] => :environment do |_task, args|
+    form_year = args[:year]&.to_i
+    raise "Usage: rake generate_lettings_documentation:describe_lettings_validations['year']" if form_year.blank?
+
+    form = FormHandler.instance.forms[FormHandler.instance.form_name_from_start_year(form_year, "lettings")]
+    raise "No form found for given year" if form.blank?
+
```
Dinssa, line 9: Would it be good to not require the year and use the `current_collection_year` method from the collection time helper as the default year?

app/services/documentation_generator.rb
```diff
@@ -69,9 +73,9 @@ def describe_bu_validations(client, form, row_parser_class, all_validation_metho
     end
   end
 
-  def describe_soft_validations(client, all_validation_methods, all_helper_methods, log_type)
+  def describe_soft_validations(client, form, all_validation_methods, all_helper_methods, log_type)
     validation_descriptions = {}
-    all_validation_methods[0..5].each do |meth|
```
carolynbarker, line 74: Why were we truncating this?

app/services/documentation_generator.rb
```diff
@@ -87,20 +91,57 @@ def describe_soft_validations(client, all_validation_methods, all_helper_methods
...
+    interruption_screen_pages = form.pages.select { |page| page.questions.first.type == "interruption_screen" }
+    interruption_screen_pages_grouped_by_question = interruption_screen_pages.group_by { |page| page.questions.first.id }
+    interruption_screen_pages_grouped_by_question.each do |_question_id, pages|
+      pages.map do |page|
+        save_soft_validation(form, page, validation_descriptions, log_type)
       end
     end
   end
 
+  def get_all_sales_methods
```
carolynbarker: We could commonise this with get_all_lettings_methods by passing in the list of validation classes to one common get_methods function

lib/tasks/generate_lettings_documentation.rake
kosiakkatrina: I thought about that, but decided to leave it as mandatory to avoid any accidents and it's pretty simple to add the year argument

lib/tasks/generate_lettings_documentation.rake
```diff
@@ -1,57 +1,61 @@
...
-                   Validations::DateValidations,
-                   Validations::LocalAuthorityValidations].map { |x| x.instance_methods + x.private_instance_methods }.flatten
-    all_helper_methods = all_methods - all_validation_methods
-
-    DocumentationGenerator.new.describe_hard_validations(client, all_validation_methods, all_helper_methods, "lettings")
+
+    documentation_generator = DocumentationGenerator.new
+    all_validation_methods, all_helper_methods = documentation_generator.get_all_lettings_methods
+
+    documentation_generator.describe_hard_validations(client, form, all_validation_methods, all_helper_methods, "lettings")
```
carolynbarker: Nice to take all this out of the rake task!

spec/lib/tasks/generate_sales_documentation_spec.rb
```diff
@@ -6,14 +6,24 @@
...
       task.reenable
     end
 
+    after do
+      Timecop.return
+      Singleton.__init__(FormHandler)
+    end
+
     context "when the rake task is run" do
       it "creates new validation documentation records" do
```
carolynbarker, line 23: Should we move this test to the documentation_generator_spec?

carolynbarker: (and "skips if the validation already exists in the database")

app/services/documentation_generator.rb
kosiakkatrina: No idea, but it doesn't make sense, cause we'd not be getting all the validations

spec/lib/tasks/generate_sales_documentation_spec.rb
kosiakkatrina: I think for numeric validations we just do it all in the task, it's pretty straightforward

app/services/documentation_generator.rb
kosiakkatrina: Updated

Dinssa (approved):
Looks good!

carolynbarker (approved):
Looks good to me!

PR #3015, opened 2025-03-24
CLDC-3857 Add new group and profit status questions to organisation setup

Dinssa:
This PR adds new questions to an organisation, allowing support users to answer:
1) If the organisation is part of a group. And if it is, to set which group it is part of.
2) If the organisation is For-profit, Non-profit or neither because its a Local authority.

Screenshot 2025-03-25 at 14 18 05

There is some JS functionality added to improve the user experience by preselecting options linked to other questions, this is optional and the page remains accessible should JS not be available. Validations have been put in place to prevent invalid options being selected as a fallback.


app/services/exports/organisation_export_service.rb
```diff
@@ -63,8 +63,7 @@ def apply_cds_transformation(organisation)
       attribute_hash["provider_type"] = organisation.provider_type_before_type_cast
       attribute_hash["merge_date"] = organisation.merge_date&.iso8601
       attribute_hash["available_from"] = organisation.available_from&.iso8601
-      attribute_hash["profit_status"] = nil # will need update when we add the field to the org
-      attribute_hash["group"] = nil # will need update when we add the field to the org
+      attribute_hash["profit_status"] = organisation.profit_status_before_type_cast
       attribute_hash["status"] = organisation.status
```
Dinssa, line 67: `group` should be passed in as one of the attributes on the organisation & I reassign `profit_status` just like with `provider_type`.

app/views/organisations/new.html.erb
```diff
@@ -48,6 +48,8 @@
         :id,
         :name,
         label: { text: "Organisation type", size: "m" },
+        "data-controller": "organisations",
+        "data-action": "change->organisations#updateProfitStatusOptions",
```
Dinssa, line 52: Depending on the user's selection for the provider type question we will filter out invalid options from the profit status question using js. If js is disabled there are validations on the organisation model to prevent adding invalid options.

app/views/organisations/new.html.erb
```diff
@@ -56,6 +58,31 @@
...
+        <%= f.govuk_radio_button :group_member, false, label: { text: "No" } %>
+      <% end %>
+
+      <%= f.govuk_collection_select :profit_status,
+          profit_status_options,
+          :id,
+          :name,
+          label: { text: "Is the organisation for-profit?", size: "m" },
+          options: { disabled: [""], selected: @organisation.profit_status || "" } %>
+
```
Dinssa: This adds a Yes/No question of whether the organisation is part of a group or not. If Yes is selected a additional question for find and select another organisation that is part of the same group is asked. In the options organisations with groups will have the group appended to their name e.g. "MHCLG (Group2)", those without will just have their name e.g. "MHCLG". User's just need to select any organisation already in the group, should there be multiple.

app/views/organisations/edit.html.erb
```diff
@@ -33,6 +33,33 @@
...
+             "data-action": "click->conditional-question#displayConditional",
+             "data-info": { conditional_questions: { group: [true] }, type: "organisation" }.to_json do %>
+            <%= f.govuk_collection_select :group_member_id,
+                group_organisation_options,
+                :id,
+                :name,
+                label: { text: "Search for an organisation that is part of the same group as this organisation", size: "m" },
+                options: { disabled: [""], selected: @organisation.oldest_group_member&.id || "" },
+                "data-controller": %w[accessible-autocomplete conditional-filter] %>
+          <% end %>
```
Dinssa, line 51: In the org edit page, should a org be part of a group we'll show the search box to find an organisation. I chose to make the organisation shown in the box the oldest organisation within the group, rather than a random organisation or the org a user may have selected when creating the org. The idea being that the oldest would be sort of the "leader" and to keep things the same when viewing the editable question across multiple orgs in a group. This is only visible to support users who are the only users that can change a group.

app/controllers/organisations_controller.rb
```diff
@@ -385,4 +389,19 @@ def get_duplicate_schemes_and_locations
...
+  def assign_group_number(current_org_id, selected_org_id)
+    selected_org = Organisation.find_by(id: selected_org_id)
+    if selected_org&.group.present?
+      selected_org.group
+    else
+      next_group_number = next_available_group_number
+      selected_org.update!(group_member: true, group_member_id: current_org_id, group: next_group_number) if selected_org
+      next_group_number
+    end
+  end
```
Dinssa, line 402: When creating or editing an org we select another org that the org should be in the same group with. If the selected org has a group number we'll take that. If not, we'll give them both the next available group number.

app/controllers/organisations_controller.rb
```diff
@@ -91,6 +91,7 @@ def create
     selected_rent_periods = rent_period_params[:rent_periods].compact_blank
     @organisation = Organisation.new(org_params)
     if @organisation.save
+      @organisation.update!(group: assign_group_number(@organisation.id, org_params[:group_member_id])) if org_params[:group_member]
```
kosiakkatrina, line 94: Could we just add the group to the org_params? Something similar to what we have in locations_controller > location_params ?
That way we'd avoid saving the org twice

app/controllers/organisations_controller.rb
```diff
@@ -127,6 +128,9 @@ def reactivate
 
   def update
     if (current_user.data_coordinator? && org_params[:active].nil?) || current_user.support?
+      if org_params[:group_member] && org_params[:group_member_id]
+        @organisation.group = assign_group_number(@organisation.id, org_params[:group_member_id])
+      end
```
kosiakkatrina: Same here, this could probably be removed if we add group to org_params

kosiakkatrina: What if we select no? Should it remove the group number from the other org?

app/controllers/organisations_controller.rb
```diff
@@ -385,4 +389,19 @@ def get_duplicate_schemes_and_locations
...
     end
   end
+
+  def assign_group_number(current_org_id, selected_org_id)
+    selected_org = Organisation.find_by(id: selected_org_id)
+    if selected_org&.group.present?
+      selected_org.group
+    else
+      next_group_number = next_available_group_number
+      selected_org.update!(group_member: true, group_member_id: current_org_id, group: next_group_number) if selected_org
```
kosiakkatrina, line 399: What would happen if we fail to save the current org because of a different validation? Would that still set the group number on the selected org?

app/views/organisations/edit.html.erb
kosiakkatrina: It's kinda weird tbh that we're selecting a single org, but might be joining a group with many

kosiakkatrina: If there's lots of orgs in 1 group is there any way to know what all of the orgs in that group are?

app/controllers/organisations_controller.rb
Dinssa: This approach allows us to use the `assign_group_number` method to find the next available group number and simultaneously add the group number to the other selected organisation. This is primarily to ensure the group number is added to the other selected organisation after a user has confirmed that this is the group they want to select .ie. after saving. Performing this action in the view would be premature before the user has chosen to create the organisation.

app/controllers/organisations_controller.rb
Dinssa: If you select "No," the other organisation selection question should be empty, as it will be disabled, and the other organisation should not be affected. On the edit page, if you change an organisation to not be part of a group, there is logic to clear the group value, but only for that specific log.

app/views/organisations/edit.html.erb
Dinssa: The search result will return the organisation name with the group in brackets at the end, if present. You can search by the text within the brackets. For example, if the organisation MHCLG is in group 2, the search box will store it as "MHCLG (GROUP2)." You can then search for either "MHCLG" or "GROUP2" - to retrieve all organisations that are also in group 2 in the dropdown list.

app/controllers/organisations_controller.rb
kosiakkatrina: Sorry, I mean what if you select yes, choose an org and then select no afterwards?

app/controllers/organisations_controller.rb
Dinssa: @samyou-softwire I didn't get a chance to have a look into this, could you please investigate.

app/controllers/organisations_controller.rb
Dinssa: It should not remove the group number from the other org, if someone did select no after choosing an org, we should not assign any group number to the new organisation and I don't think we will as it's currently implemented. Please check @samyou-softwire

lib/tasks/update_organisations_group_profit_status.rake
```diff
@@ -0,0 +1,29 @@
...
+
+    CSV.foreach(csv_path, headers: true) do |row|
+      organisation = Organisation.find_by(id: row["id"].to_i)
+      if organisation
+        organisation.skip_group_member_validation = true
+        organisation.update!(
+          profit_status: row["profit_status"].to_i,
+          group_member: true,
+          group: row["group"].to_i,
+        )
```
Dinssa: @samyou-softwire Had a look at the CSV file shared and I released this task does not account for updating organisations without a group, where we still want to update the profit status. I think this needs to be amended to conditionally update the `group` and `group_member` fields only if a group number is provided, for example we do not want to be setting `group_member` to `true` for each of the logs in the file.

Additionally, I would remove the `core_registration_number` and `core_type` from the CSV before adding it in to the branch. The `core_registration_number` is not needed, nor is `core_type`. However, there is a validation that checks the profit status against the core_type (e.g., profit status must be local authority if the `core_type` is LA). This validation does not require the column to be included in the CSV, it should already be set.

app/controllers/organisations_controller.rb
samyou-softwire: yup the `if org_params[:group_member]` on that line (and on the save line) should prevent an association being created

app/controllers/organisations_controller.rb
samyou-softwire: found that:
- on create `assign_group_number` is guarded by a `.save` call so will not associate if validation fails
- on update `assign_group_number` was outside the update check so would associate if a validation fails. this has been resolved

lib/tasks/update_organisations_group_profit_status.rake
samyou-softwire: have imported the csv accordingly

samyou-softwire:
pipeline seems stuck on an old commit, reopening this

PR #3017, opened 2025-03-25
Explain types of logs

carolynbarker:
I think this is useful introductory context but couldn't see it anywhere - I think this looks like a good place, alongside the user and organisation info.

(As always) feedback or suggestions are very welcome!

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -4,11 +4,11 @@ nav_order: 1
 
 # Overview
 
-All lettings and and sales of social housing in England need to be logged with the Ministry of Housing, Communities and Local Government (MHCLG). This is done by data providing organisations: Local Authorities and Private Registered Providers (PRPs, i.e. housing associations).
+All lettings and sales of social housing in England need to be logged with the Ministry of Housing, Communities and Local Government (MHCLG). This is done by data providing organisations: Local Authorities and Private Registered Providers (PRPs, i.e. housing associations).
 
 Data is collected via a form that runs on an annual data collection window basis. Form changes are made annually to add new questions, remove any that are no longer needed, or adjust wording or answer options etc.
 
-Each data collection window runs from 1 April to 1 April the following year (plus an extra 3 months to allow for any late submissions). This means that between April and June, 2 collection windows are open simultaneously and logs can be submitted for either.
+Each data collection window runs from 1 April to 1 April the following year (plus an extra 3 months to allow for any late submissions). This means that between April and June, 2 collection windows are open simultaneously and logs can be submitted for either. This is called the **crossover period**.
 
 ADD (Analytics & Data Directorate) statisticians are the other primary users of the service. The data collected is transferred to MHCLG's consolidated data store (CDS) via nightly XML exports to an S3 bucket. CDS ingests and transforms this data, ultimately storing it in a MS SQL database and exposing it to analysts and statisticians via Amazon Workspaces.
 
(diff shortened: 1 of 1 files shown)


docs/index.md
```diff
@@ -56,9 +56,33 @@ Taking the relationships from the above diagram, and looking at which logs each
 
 ![User log access permissions](https://raw.githubusercontent.com/communitiesuk/submit-social-housing-lettings-and-sales-data/main/docs/images/user_log_permissions.png)
 
-## Supported housing schemes
+## Logs
 
-A supported housing scheme (or service) provides shared or self-contained housing for a particular client group, for example younger or vulnerable people. A scheme can be run at multiple locations, and a single location may contain multiple units (for example bedrooms in shared houses or a bungalow with 3 bedrooms).
+There are two types of logs: sales logs and lettings logs.
+Sales and lettings logs have some validations and fields in comon, but many more are specific to either sales or to lettings.
```
kosiakkatrina: `common`

docs/index.md
```diff
@@ -56,9 +56,33 @@ Taking the relationships from the above diagram, and looking at which logs each
...
+**Shared ownership** transactions may include either shared ownership initial transactions or staircasing transactions.
+
+In 2024 and before, we also collected data on **outright sales** - full property sales with no discount scheme (ie any other property sale).
+
+### Lettings logs
+
+There are two types of lettings logs: general needs and supported housing.
+
+- **General needs** refers to a letting that is not part of a supported housing scheme.
+- **Supported housing** lettings are those belonging to a **scheme**, as described in the following section.
```
kosiakkatrina, line 81: We have this in user guidance, feels a bit redundant. Maybe linking to the guidance page from here instead might be less overwhelming?

carolynbarker: Where's the user guidance page?

I was looking for somewhere to put this in some sort of project intro for developers. "Core: Overview" didn't seem right as it's more about key facts and contacts etc, in a standard format, and the "Zero to Hero" wasn't right as that's setup instructions. This page already has types of users and types of organisations, which are useful context but less essential to understand for most dev tasks 

If the user guidance page already has similar stuff, that sounds potentially useful, happy to link to that / move bits accordingly

kosiakkatrina: https://submit-social-housing-data.communities.gov.uk/guidance 
We link to this from homepage as well

carolynbarker: Thanks! There's lots of detail there and that will be great to link.

I still think an intro explanation at this level would be useful for new devs - I didn't realise for a while that eg general needs was just the opposite of supported housing, which isn't mentioned in the user guidance (because users would have that domain knowledge already). And the list of lettings types in the guidance is huge and quite intimidating!

carolynbarker: Thinking about it, the types of user and the notion of a scheme are systems that we've defined within CORE, so we definitely need to explain them here. Whereas the difference between a sales and lettings log and the types of each are existing distinctions in the wider sector which CORE only reflects rather than defining.

carolynbarker: So arguably this could be a place for only explaining the decisions we've made in CORE, and the broader overview of "what are logs" could be a separate page? Perhaps in support confluence?

kosiakkatrina: I think there's some user facing schemes guidance that might be useful: submit-social-housing-data.communities.gov.uk/scheme-changes

If you think having logs explanation here would benefit new people then that's fine, we can leave it in

Yeah there might be something on confluence or Sharepoint, especially from alpha times

kosiakkatrina (approved):

docs/index.md
carolynbarker: Cool, if you're happy then I'll leave it in for now. I definitely think it's useful for new people, and of course future generations can iterate as they see fit

I'll mention the confluence/sharepoint docs on the "sort out confluence" ticket - I thought we had one but can't see it, I'll write one up

PR #3018, opened 2025-03-25
CLDC-3938 Log invalid BU row numbers

carolynbarker:
This will help debug the alerts we've been seeing when bulk uploads fails with rows that are still invalid.

### DIFF
--- a/app/services/bulk_upload/lettings/validator.rb
+++ b/app/services/bulk_upload/lettings/validator.rb
@@ -55,8 +55,10 @@ class BulkUpload::Lettings::Validator
     row_parsers.each do |row_parser|
       row_parser.log.blank_invalid_non_setup_fields!
     end
-    if any_logs_invalid?
+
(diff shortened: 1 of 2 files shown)


app/services/bulk_upload/lettings/validator.rb
```diff
@@ -101,8 +103,8 @@ def validate_duplicate_rows
     end
   end
 
-  def any_logs_invalid?
-    row_parsers.any? { |row_parser| row_parser.log.invalid? }
+  def invalid_row_numbers
+    row_parsers.each_with_index.map { |row_parser, index| index + row_offset + 1 if row_parser.log.invalid? }.compact
```
kosiakkatrina, line 107: Only thought is that `any?` will stop as soon as it finds a match, so might be a bit more performant, but given this very rarely returns true, it's probably okay to change it to help with the logging

kosiakkatrina (approved):

PR #3023, opened 2025-03-27
CLDC-3949 Add dividers before don't know answer options

kosiakkatrina:


app/models/form/lettings/questions/hb.rb
```diff
@@ -12,8 +12,8 @@ def initialize(id, hsh, page)
     "1" => { "value" => "Housing benefit" },
     "6" => { "value" => "Universal Credit housing element" },
     "9" => { "value" => "Neither" },
-    "divider" => { "value" => true },
     "3" => { "value" => "Don’t know" },
+    "divider" => { "value" => true },
     "10" => { "value" => "Tenant prefers not to say" },
   }.freeze
```
Dinssa: We'd want Don't know on the other side of this divider too and probably "Tenant prefers not to say" above it

app/models/form/sales/questions/buyer_still_serving.rb
```diff
@@ -11,7 +11,8 @@ def initialize(id, hsh, page)
     "4" => { "value" => "Yes" },
     "5" => { "value" => "No" },
     "6" => { "value" => "Buyer prefers not to say" },
-    "7" => { "value" => "Don't know" },
+    "divider" => { "value" => true },
+    "7" => { "value" => "Don’t know" },
   }.freeze
```
Dinssa, line 16: > "Tenant prefers not to say" above it"

Like here

app/models/form/lettings/questions/hb.rb
kosiakkatrina: Ah, nice catch, updated!

Dinssa (commented):
It looks like there is a few more places this can be added:
In lettings...
joint.rb (here it's is just written differently with quotes, think it'll still show but not sure)

In sales...
buyer1_mortgage.rb
buyer2_mortgage.rb
buyer_previous.rb
fromprop.rb
mortgage_lender.rb

kosiakkatrina:
> mortgage_lender.rb

Ah, yes, updated the joint one, for the other questions, we never display the `Don't know` options, they're either not in `displayed_answer_options` or we filter them out

Dinssa (approved):

PR #3029, opened 2025-03-31
Add error to firststair question and test date changes that cross collection years

Dinssa:


app/models/validations/sales/financial_validations.rb
```diff
@@ -122,7 +122,7 @@ def validate_staircase_difference(record)
     elsif record.numstair
       # We must use the lowest possible percentage for a staircasing transaction of any saletype, any year since 1980
       minimum_percentage_per_staircasing_transaction = 1
-      previous_staircasing_transactions = record.numstair - 1
+      previous_staircasing_transactions = record.numstair - 2
```
Dinssa, line 126: In addition to the current staircasing transaction being recored, the initial purchase (equity) also counts as a staircasing transaction in number of staircasing transactions.

spec/models/sales_log_spec.rb
```diff
@@ -1087,14 +1089,14 @@ def previous_postcode_fields
...
+      LocalAuthority.find_by(code: "E08000003").update!(end_date: Time.zone.yesterday)
     end
 
     it "removes the LA" do
-      sales_log.update!(saledate: Time.zone.yesterday, la: "E08000003")
+      sales_log.update!(saledate: 2.days.ago, la: "E08000003")
       expect(sales_log.reload.la).to eq("E08000003")
 
-      sales_log.update!(saledate: Time.zone.tomorrow)
+      sales_log.update!(saledate: Time.zone.today)
```
Dinssa: i'll change this

app/helpers/collection_time_helper.rb
```diff
@@ -53,4 +53,10 @@ def previous_collection_start_date
   def archived_collection_start_year
     current_collection_start_year - 2
   end
+
+  def generate_different_date_within_collection_year(date, start_date_override = nil, end_date_override = nil)
+    start_date = [start_date_override || collection_start_date(date).to_date, collection_start_date(date).to_date].max.to_date
+    end_date = [end_date_override || collection_end_date(date).to_date, collection_end_date(date).to_date].min.to_date
+    ((start_date..end_date).to_a - [date.to_date]).sample
+  end
 end
```
Dinssa: Allows you to override the start and end dates, but still truncates the range to be within the collection year.

app/helpers/collection_time_helper.rb
```diff
@@ -53,4 +53,10 @@ def previous_collection_start_date
   def archived_collection_start_year
     current_collection_start_year - 2
   end
+
+  def generate_different_date_within_collection_year(date, start_date_override = nil, end_date_override = nil)
+    start_date = [start_date_override || collection_start_date(date).to_date, collection_start_date(date).to_date].max.to_date
+    end_date = [end_date_override || collection_end_date(date).to_date, collection_end_date(date).to_date].min.to_date
```
carolynbarker: We could go eg `[start_date_override, collection_start_date(date).to_date].compact.max.to_date` to avoid repeating collection_start_date(date).to_date

spec/helpers/merge_requests_helper_spec.rb
```diff
@@ -163,7 +165,7 @@
     let(:organisation) { create(:organisation, name: "Org 1") }
     let(:merging_organisation) { create(:organisation, name: "Org 2") }
     let(:merging_organisation_2) { create(:organisation, name: "Org 3") }
-    let(:merge_request) { create(:merge_request, absorbing_organisation: organisation, merge_date: Time.zone.today) }
+    let(:merge_request) { create(:merge_request, absorbing_organisation: organisation, merge_date: Time.zone.today - 2.days) }
```
carolynbarker: Is there anything special about this date? would it shift any problems 2 days later?

app/helpers/collection_time_helper.rb
```diff
@@ -53,4 +53,10 @@ def previous_collection_start_date
   def archived_collection_start_year
     current_collection_start_year - 2
   end
+
+  def generate_different_date_within_collection_year(date, start_date_override = nil, end_date_override = nil)
```
carolynbarker: It would be nice to go `start_date_override: nil, end_date_override: nil` so that it's clear what each argument is when we call this method

carolynbarker: It would also let us call with an end date override without having to pass `nil` as well

spec/models/sales_log_spec.rb
```diff
@@ -1085,16 +1087,18 @@ def previous_postcode_fields
 
   context "when form year changes and LA is no longer active" do
     let!(:sales_log) { create(:sales_log) }
+    let(:date) { generate_different_date_within_collection_year(Time.zone.now, nil, Time.zone.now + 14.days) }
+    let(:date_after) { generate_different_date_within_collection_year(date, Time.zone.now, Time.zone.now + 14.days) }
```
carolynbarker: I think the syntax change I suggested for generate_different_date_within_collection_year would make this clearer to read

carolynbarker: Is `date_after` actually always after `date`? I think it might just be a *different* date in the next 14 days?

carolynbarker: ah, but I think that's fine for the test - slightly confusing name though

spec/requests/merge_requests_controller_spec.rb
```diff
@@ -416,7 +416,7 @@
         context "when merge date set to a date more than 1 year in the future" do
           let(:merge_request) { MergeRequest.create!(requesting_organisation: organisation) }
           let(:params) do
-            { merge_request: { page: "merge_date", "merge_date": [(Time.zone.now.day + 1).to_s, Time.zone.now.month.to_s, (Time.zone.now.year + 1).to_s].join("/") } }
+            { merge_request: { page: "merge_date", "merge_date": (Time.zone.now + 1.year + 1.day).strftime("%d/%m/%Y") } }
```
carolynbarker, line 419: Wouldn't the old test have sometimes given us a day of 32? 😨 

Nice fix!

app/helpers/collection_time_helper.rb
Dinssa: That's better will update it

spec/helpers/merge_requests_helper_spec.rb
Dinssa: Updated this to generate a merge date which we'll use to base all other dates off. It was to keep the sale/start dates within the same area but you're right on 2nd April it would fail. The update should work on any date.

spec/models/sales_log_spec.rb
Dinssa: Updated the naming

app/helpers/collection_time_helper.rb
Dinssa: Updated with this

spec/models/sales_log_spec.rb
```diff
@@ -376,7 +378,7 @@
 
     context "when there is a log with a different sale date" do
       before do
-        create(:sales_log, :duplicate, saledate: Time.zone.tomorrow)
+        create(:sales_log, :duplicate, saledate: Time.zone.yesterday)
```
carolynbarker: Will this be ok when yesterday is in a different collection year?

spec/models/sales_log_spec.rb
carolynbarker: Sorry, I think I was wrong that it's fine for the rest - we do need `:date_after_end_date` to be later than `:end_date` for the test to pass right? But if `:end_date` is in the next 14 days then `:date_after_end_date` can end up being later than it. 

We could try using `end_date` as a `start_date_override` for `date_after_end_date`, but then we'd have to be careful about there being enough days afterwards in the collection year.
   
I think that might be overcomplicating this, maybe it's best to just hard code two dates in the 2024 collection year - that will still test the functionality we're testing here

carolynbarker: Or hardcode the days and months and use the current year

spec/helpers/merge_requests_helper_spec.rb
```diff
@@ -198,10 +201,10 @@
     context "when merging organisations have logs" do
       before do
         create(:lettings_log, owning_organisation: organisation)
-        create(:lettings_log, owning_organisation: merging_organisation, startdate: Time.zone.tomorrow)
+        create(:lettings_log, owning_organisation: merging_organisation, startdate: generate_different_date_within_collection_year(merge_date, start_date_override: merge_date, end_date_override: Time.zone.now + 14.days))
```
carolynbarker, line 204: What happens if the merge date is the last day of the year?

app/helpers/collection_time_helper.rb
```diff
@@ -53,4 +53,10 @@ def previous_collection_start_date
   def archived_collection_start_year
     current_collection_start_year - 2
   end
+
+  def generate_different_date_within_collection_year(date, start_date_override: nil, end_date_override: nil)
+    start_date = [start_date_override, collection_start_date(date).to_date].compact.max.to_date
+    end_date = [end_date_override, collection_end_date(date).to_date].compact.min.to_date
+    ((start_date..end_date).to_a - [date.to_date]).sample
```
carolynbarker: Actually I think we need to handle the case here where start_date_override is the last day or end_date_override is the first day, or they're both the same day (or even start_date_override > end_date_override). In those cases there can be no dates left to choose from, ((start_date..end_date).to_a can be empty

spec/models/sales_log_spec.rb
Dinssa: Updated

app/helpers/collection_time_helper.rb
carolynbarker: It may be simplest to remove the overrides for now, or the whole method if unused. Happy to think more about it later but don't want it to block the release!

spec/helpers/merge_requests_helper_spec.rb
Dinssa: I changed this to hardcode the merge date as 1st Jan 2025 and kept the generate different dates method for the logs.

spec/models/sales_log_spec.rb
```diff
@@ -1087,8 +1087,8 @@ def previous_postcode_fields
 
   context "when form year changes and LA is no longer active" do
     let!(:sales_log) { create(:sales_log) }
-    let(:end_date) { generate_different_date_within_collection_year(Time.zone.now, end_date_override: Time.zone.now + 14.days) }
-    let(:date_after_end_date) { generate_different_date_within_collection_year(end_date, start_date_override: Time.zone.now, end_date_override: Time.zone.now + 14.days) }
+    let(:end_date) { Time.zone.local(2025, 3, 30) }
+    let(:date_after_end_date) { Time.zone.local(2025, 3, 31) }
```
carolynbarker, line 1091: Beautiful, nice and safe and clear!

I think this test is really about the LA having ended, despite the name referring to the end of the year. So not worried about the fact we aren't really changing year.

app/helpers/collection_time_helper.rb
Dinssa: Updated to return nil when 1) the start date after the end date and 2) when there are no available dates to choose from. When writing tests or using this method it should return nil if invalid values are passed. We should be able to identify that at the time of creating or running the tests.

Dinssa (commented):
> Sorry, I think I was wrong that it's fine for the rest - we do need :date_after_end_date to be later than :end_date for the test to pass right? But if :end_date is in the next 14 days then :date_after_end_date can end up being later than it.
> 
> We could try using end_date as a start_date_override for date_after_end_date, but then we'd have to be careful about there being enough days afterwards in the collection year.
> 
> I think that might be overcomplicating this, maybe it's best to just hard code two dates in the 2024 collection year - that will still test the functionality we're testing here

I hardcoded some dates, which isn't perfect. I set the end date of the LA code to 30th March and the log's sale date to 31st March. It's does not look possible to update a log to another collection year in the test, but this should sufficiently check that there is no LA code on the log after the LA code is no longer available.

spec/helpers/merge_requests_helper_spec.rb
carolynbarker: Yep that seems safe to me

carolynbarker (approved):

PR #3031, opened 2025-04-01
Update Test Log Creation: Ensure dates fall within CY and 14 day range

Dinssa:


app/controllers/test_data_controller.rb
```diff
@@ -20,7 +22,7 @@ def create_setup_test_lettings_log
       return render_not_found unless FeatureToggle.create_test_logs_enabled?
 
       file = Tempfile.new("#{year}_test_lettings_log.csv")
-      log = FactoryBot.create(:lettings_log, :completed, assigned_to: current_user, ppostcode_full: "SW1A 1AA", startdate: Time.zone.local(year.to_i, rand(4..12), rand(1..28)))
+      log = FactoryBot.create(:lettings_log, :completed, assigned_to: current_user, ppostcode_full: "SW1A 1AA", startdate: generate_different_date_within_collection_year(Time.zone.local(year.to_i, 4, 1), end_date_override: Time.zone.now + 14.days))
```
carolynbarker, line 25: I'd like to add some unit tests for `generate_different_date_within_collection_year` - just to check that it behaves as expected including on the first and last days, and with an end date override.

If we keep start_date_override we should test that too, but I wonder if that can be removed? That would add confidence that there's not going to be any awkward date combinations with no results. If we do keep it we would need to handle start date > end date, or having no days between them, or only one day which is the excluded day, or any other edge cases we can think of. It adds a lot of complexity and options that are only taken in edge cases so won't be well covered. And I don't think we currently use it - so I'm hopeful we can get rid of it.

We also seem to always use end_date_override with a 14 day period - so we could consider splitting `generate_different_date_within_collection_year` into two methods, one `generate_different_date_within_collection_year` which has no overrides and one `generate_different_date_within_collection_year_within_fourteen_days` which always assumes an end date override of 14 days. I think I'm leaning towards not splitting it out though

app/helpers/collection_time_helper.rb
```diff
@@ -58,7 +58,7 @@ def generate_different_date_within_collection_year(date, start_date_override: ni
     start_date = [start_date_override, collection_start_date(date).to_date].compact.max.to_date
     end_date = [end_date_override, collection_end_date(date).to_date].compact.min.to_date
 
-    return nil if start_date > end_date
+    return nil if start_date > end_date || start_date == end_date
```
carolynbarker: We could just go start_date >= end_date

carolynbarker: We could just go start_date >= end_date

carolynbarker: We could just go start_date >= end_date

spec/helpers/collection_time_helper_spec.rb
```diff
@@ -106,4 +106,72 @@
       end
     end
   end
+
+  describe "#generate_different_date_within_collection_year" do
```
carolynbarker, line 110: These look good!

spec/helpers/collection_time_helper_spec.rb
```diff
@@ -106,4 +106,72 @@
...
+        expect(result).to be_nil
+      end
+    end
+
+    context "when end_date_override is the first day" do
+      it "returns nil" do
+        result = generate_different_date_within_collection_year(date, end_date_override: Time.zone.local(2024, 4, 1))
+        expect(result).to be_nil
+      end
+    end
```
carolynbarker, line 133: We could also test that the override is ignored if outside the collection year.

spec/helpers/collection_time_helper_spec.rb
```diff
@@ -106,4 +106,72 @@
...
+
+    context "when valid overrides are provided" do
+      it "returns a different date within the overridden range" do
+        result = generate_different_date_within_collection_year(date, start_date_override: Time.zone.local(2024, 5, 1), end_date_override: Time.zone.local(2024, 7, 1))
+        expect(result).not_to eq(date)
+        expect(result).to be_between(Time.zone.local(2024, 5, 1), Time.zone.local(2024, 7, 1))
+      end
+    end
```
carolynbarker: We could test the overrides individually too

spec/helpers/collection_time_helper_spec.rb
```diff
@@ -106,4 +106,72 @@
...
+
+    context "when the date is at the end of the collection year" do
+      let(:date) { Time.zone.local(2025, 3, 31) }
+
+      it "returns a different date within the collection year" do
+        result = generate_different_date_within_collection_year(date)
+        expect(result).not_to eq(date)
+        expect(result).to be_between(Time.zone.local(2024, 4, 1), Time.zone.local(2025, 3, 31))
+      end
```
carolynbarker, line 166: We could also test these with one or both overrides

carolynbarker (approved):

app/helpers/collection_time_helper.rb
carolynbarker: Sorry, I think my internet had a wobble, didn't mean to send this three times!

carolynbarker (approved):

PR #3032, opened 2025-04-01
Don't block BU creation with invalid rent range

Dinssa:
Skips BU errors on scheme, location, rent type, and needstype, keep on period, postcode_known, la, uprn, beds, brent.

### DIFF
--- a/app/models/validations/financial_validations.rb
+++ b/app/models/validations/financial_validations.rb
@@ -218,10 +218,10 @@ private
         record.errors.add :uprn, I18n.t("validations.lettings.financial.uprn.rent_below_hard_min")
         record.errors.add :la, I18n.t("validations.lettings.financial.la.rent_below_hard_min")
         record.errors.add :postcode_known, I18n.t("validations.lettings.financial.postcode_known.rent_below_hard_min")
-        record.errors.add :scheme_id, I18n.t("validations.lettings.financial.scheme_id.rent_below_hard_min")
-        record.errors.add :location_id, I18n.t("validations.lettings.financial.location_id.rent_below_hard_min")
-        record.errors.add :rent_type, I18n.t("validations.lettings.financial.rent_type.rent_below_hard_min")
-        record.errors.add :needstype, I18n.t("validations.lettings.financial.needstype.rent_below_hard_min")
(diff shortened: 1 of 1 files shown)


app/models/validations/financial_validations.rb
```diff
@@ -218,10 +218,10 @@ def validate_rent_range(record)
...
-        record.errors.add :scheme_id, I18n.t("validations.lettings.financial.scheme_id.rent_below_hard_min")
-        record.errors.add :location_id, I18n.t("validations.lettings.financial.location_id.rent_below_hard_min")
-        record.errors.add :rent_type, I18n.t("validations.lettings.financial.rent_type.rent_below_hard_min")
-        record.errors.add :needstype, I18n.t("validations.lettings.financial.needstype.rent_below_hard_min")
+        record.errors.add :scheme_id, :skip_bu_error, message: I18n.t("validations.lettings.financial.scheme_id.rent_below_hard_min")
+        record.errors.add :location_id, :skip_bu_error, message: I18n.t("validations.lettings.financial.location_id.rent_below_hard_min")
+        record.errors.add :rent_type, :skip_bu_error, message: I18n.t("validations.lettings.financial.rent_type.rent_below_hard_min")
+        record.errors.add :needstype, :skip_bu_error, message: I18n.t("validations.lettings.financial.needstype.rent_below_hard_min")
         record.errors.add :period, I18n.t("validations.lettings.financial.period.rent_below_hard_min")
```
carolynbarker, line 225: If the rent is below hard min, won't this error on `period` block bulk upload anyway?

(likewise with the hard_max one)

Dinssa: Period is not a setup question / important question, it'll get the error and be displayed in the report but it won't block creation and will allow the user to fix in service. We block all logs where there is an issue with the setup questions and ask a user to reupload, not giving them a chance to fix in service.

carolynbarker: ah, of course, we only need to add `skip_bu_error` on setup errors. thanks!

carolynbarker (approved):

carolynbarker (approved):

PR #3055, opened 2025-04-16
CLDC-3987: Change BU validation message for charge fields related to missing charge

Dinssa:


app/services/bulk_upload/lettings/year2024/row_parser.rb
```diff
@@ -945,9 +945,10 @@ def validate_all_charges_given
       errors.add(field, I18n.t("#{ERROR_BASE_KEY}.charges.missing_charges", sentence_fragment: charge))
     end
 
-    other_charge_fields.each do |field, _charge|
-      blank_charge_fields.each do |_blank_field, blank_charge|
-        errors.add(field, I18n.t("#{ERROR_BASE_KEY}.charges.missing_charges", sentence_fragment: blank_charge))
+    # if there were any errors, add an error to all the remaining present charges to make this entire section invalid
+    if blank_charge_fields.any?
+      other_charge_fields.each_key do |field|
+        errors.add(field, I18n.t("#{ERROR_BASE_KEY}.charges.related_to_missing_charge"))
```
davidm-m, line 951: It looks like this will still generate the three extra errors reported in the ticket. They'll have better error messages, but I don't see why we need them at all.

samyou-softwire: I'll double check this one with Rachel

samyou-softwire: looking into this more, it seems that there may a reason for this. there's some specific code that prevents any charge fields being written to the csv if any are missing, see `app/.../2024/row_parser.rb#1290`

```rb
attributes["brent"] = field_125 if all_charges_given?
attributes["scharge"] = field_126 if all_charges_given?
attributes["pscharge"] = field_127 if all_charges_given?
attributes["supcharg"] = field_128 if all_charges_given?
```

if I remove the code in this PR then we get some unhelpful errors in the log as these other present charges are listed as 'missing' (despite not being so), so looks like these errors are here to overwrite this behaviour and make it clearer for users @davidm-m

davidm-m: Ah ok that makes sense. Thanks for checking!

davidm-m (approved):

PR #3074, opened 2025-05-15
CLDC-4025: Move Sales Log organisation to first question

samyou-softwire:
this is to resolve [CLDC-4025](https://mhclgdigital.atlassian.net/browse/CLDC-4025), where we agreed to move organisation back to being the first question

a little extra work needed as we don't want to reorder the CSV as we've already released this years form, so introduces a function to order it back

[CLDC-4025]: https://mhclgdigital.atlassian.net/browse/CLDC-4025?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/services/csv/sales_log_csv_service.rb
```diff
@@ -321,6 +322,21 @@ def order_address_fields_for_support(attributes)
       end
     end
 
+    # as part of CLDC-3719 it was decided to move the saledate question to be first in the form
+    # this caused issues reported in CLDC-4025 where the user only enter saledates for their active organisation
+    # we decided to move the organisation question back to being first
+    # however, we did not want to reorder the CSV export as this would disrupt existing users' data pipelines
+    # so, this function reorders questions back when exporting CSVs
+    # next year, we can remove this function as we will be reordering the csv fields anyway
+    def order_saledate_question_before_owning_organisation_question(ordered_questions)
```
samyou-softwire, line 331: happy to hear thoughts on where best to place this code, did consider making this a form method instead

app/models/form/sales/questions/owning_organisation_id.rb
```diff
@@ -93,5 +93,5 @@ def merged_organisation_label(name, merge_date)
     "#{name} (inactive as of #{merge_date.to_fs(:govuk_date)})"
   end
 
-  QUESTION_NUMBER_FROM_YEAR = { 2023 => nil, 2024 => 1, 2025 => 2 }.freeze
+  QUESTION_NUMBER_FROM_YEAR = { 2023 => nil, 2024 => 1 }.freeze
```
davidm-m: I think it makes sense to leave in 2025 and set it to 1 instead of removing it. Is there a specific reason you got rid of it completely?

app/models/form/sales/questions/sale_date.rb
```diff
@@ -6,5 +6,5 @@ def initialize(id, hsh, page)
     @question_number = QUESTION_NUMBER_FROM_YEAR[form.start_date.year] || QUESTION_NUMBER_FROM_YEAR[QUESTION_NUMBER_FROM_YEAR.keys.max]
   end
 
-  QUESTION_NUMBER_FROM_YEAR = { 2023 => 1, 2024 => 3, 2025 => 1 }.freeze
+  QUESTION_NUMBER_FROM_YEAR = { 2023 => 1, 2024 => 3 }.freeze
```
davidm-m: Same here about getting rid of the explicit 2025 value.

app/services/csv/sales_log_csv_service.rb
davidm-m: I think this makes the most sense. It only applies to the csv export so it doesn't need to be on the form itself.

davidm-m (changes requested):

app/models/form/sales/questions/owning_organisation_id.rb
samyou-softwire: looking around it largely seems to be the dominant pattern for the project to not re-specify the year if it doesn't change between years, it's also what the question_number initialiser is set up for
```rb
@question_number = QUESTION_NUMBER_FROM_YEAR[form.start_date.year] || QUESTION_NUMBER_FROM_YEAR[QUESTION_NUMBER_FROM_YEAR.keys.max]
```

davidm-m: 🫤
The problem I see with it is that then next year if the question order changes we'll either add it in again (for 0 net benefit) or forget to and break this year's form. I don't see what benefit we get from removing it.

samyou-softwire: ah makes sense 👍, I've changed this now

davidm-m (approved):
Looks good, thanks!

PR #3075, opened 2025-05-19
CLDC-3857 add new questions to organisation setup

samyou-softwire:
see description and comments of #3015 

reopening PR as the tests didn't seem to be running on the correct commit

difference of note:
- finish the raketask to import the new organisation data
- update the xml tests to not care about file line order, as XML is unordered so there's no point updating tests if the ordering changes. happy to hear alternate views on this


app/controllers/organisations_controller.rb
```diff
@@ -385,4 +389,19 @@ def get_duplicate_schemes_and_locations
...
     end
   end
+
+  def assign_group_number(current_org_id, selected_org_id)
+    selected_org = Organisation.find_by(id: selected_org_id)
+    if selected_org&.group.present?
+      selected_org.group
+    else
+      next_group_number = next_available_group_number
+      selected_org.update!(group_member: true, group_member_id: current_org_id, group: next_group_number) if selected_org
```
davidm-m, line 399: What's the intention of group member id?

app/frontend/controllers/organisations_controller.js
```diff
@@ -0,0 +1,27 @@
...
+
+    profitStatusSelect.disabled = false
+    localAuthorityOption.hidden = false
+    nonProfitOption.hidden = false
+    profitOption.hidden = false
+
+    if (providerType === 'LA') {
+      profitStatusSelect.value = 'local_authority'
+      nonProfitOption.hidden = true
+      profitOption.hidden = true
```
davidm-m, line 21: Since the idea is this is the only option, I'd probably disable the input to make that clear.

app/frontend/controllers/organisations_controller.js
```diff
@@ -0,0 +1,27 @@
...
+    } else if (providerType === 'PRP') {
+      profitStatusSelect.value = ''
+      localAuthorityOption.hidden = true
+    }
```
davidm-m, line 25: This means that any time the provider type changes we're clearing this. I think that's fine since there are only two provider types, but can you think of any issues that might cause?

app/services/exports/organisation_export_service.rb
```diff
@@ -74,8 +74,7 @@ def apply_cds_transformation(organisation)
       attribute_hash["provider_type"] = organisation.provider_type_before_type_cast
       attribute_hash["merge_date"] = organisation.merge_date&.iso8601
       attribute_hash["available_from"] = organisation.available_from&.iso8601
-      attribute_hash["profit_status"] = nil # will need update when we add the field to the org
-      attribute_hash["group"] = nil # will need update when we add the field to the org
+      attribute_hash["profit_status"] = organisation.profit_status_before_type_cast
```
davidm-m, line 77: Are we deliberately not exporting the group?

app/views/organisations/edit.html.erb
```diff
@@ -29,6 +29,33 @@
...
+             label: { text: "Yes" },
+             "data-controller": "conditional-question",
+             "data-action": "click->conditional-question#displayConditional",
+             "data-info": { conditional_questions: { group: [true] }, type: "organisation" }.to_json do %>
+            <%= f.govuk_collection_select :group_member_id,
+                group_organisation_options,
+                :id,
+                :name,
+                label: { text: "Search for an organisation that is part of the same group as this organisation", size: "m" },
+                options: { disabled: [""], selected: @organisation.oldest_group_member&.id || "" },
```
davidm-m: Concerned that this name doesn't match to what's being selected here - the user is just picking _any_ org in the same group, not necessarily the oldest one.

db/migrate/20250219122817_add_manual_address_entry_selected_to_logs.rb
davidm-m, line 1: Migrations shouldn't be deleted as a rule. Why have we done it here?

config/locales/en.yml
```diff
@@ -220,6 +220,10 @@ en:
       name_missing: "Enter the name of the organisation."
       name_not_unique: "An organisation with this name already exists. Use the organisation that already exists or add a location or other identifier to the name. For example: Organisation name (City)."
       provider_type_missing: "Select the organisation type."
+      group_missing: "Select a group member."
+      profit_status:
+        must_be_LA: "This organisation is a local authority, it's profit status must also be local authority."
+        must_not_be_LA: "This organisation is a private registered provider, it's profit status cannot be local authority."
```
davidm-m: Grammar nitpick: `its` is possessive here and shouldn't have an apostrophe.

config/csv/organisations_structure/20250515_organisation_profit_status_and_group.csv
davidm-m, line 1: I probably wouldn't commit this? It's only going to be needed for a single run of the rake task and it being on the ticket is fine for that.

davidm-m (changes requested):
Got a few comments but I'm particularly concerned about the `group_member_id` and `oldest_group_member` things - they're not actually used for anything as far as I can tell and it seems like they're going to be a random mix of ids that could cause problems with data consistency down the line.

db/migrate/20250219122817_add_manual_address_entry_selected_to_logs.rb
samyou-softwire: not sure; it was deleted in 812787fd which also added a migration to delete this column so seems accidental. have now added it back

config/locales/en.yml
samyou-softwire: good spot; have removed the apostrophe

config/csv/organisations_structure/20250515_organisation_profit_status_and_group.csv
samyou-softwire: after merging this in we'll need to run a script that has access to this data. if it's included here as it'll be available on the deployed environment without needing to find a way to upload. should we look into alternative ways to upload?

app/frontend/controllers/organisations_controller.js
samyou-softwire: looks like this was on purpose; disabled html elements don't submit when the form does. could perhaps look into some disabled value workaround?

davidm-m: I feel like I should know that already... No need for a workaround in that case, let's just stick with this.

config/csv/organisations_structure/20250515_organisation_profit_status_and_group.csv
davidm-m: What's the script that we're running? If we do need this file on the server then I'm fine for it to be committed but generally in rails, if you're running rake task on a remote server you only need any files like this locally.

samyou-softwire: `update_organisations_group_profit_status` https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/3075/files#diff-00a85dd54353e90fd050ee8127f98b68c7973be4ea01edb09905e542e88ff204

app/controllers/organisations_controller.rb
samyou-softwire: as I understand it `group_member_id` stems from the way we input groups, by inputting an organisation in that group
image
this input submits the value `group_member_id` (the org you pick) which is saved to the org and used to figure out which group to add to/create

I did see if it was possible to remove this field from the model, but there does seem to be logic that depends on this field being on organisation.

we have the following validation on `Organisation`
```rb
validates :group_member_id, presence: { message: I18n.t("validations.organisation.group_missing") }, if: -> { group_member? }
```
this is used to verify that the user selects a group after pressing that yes the org is in a group. if `group_member_id` was removed, we wouldn't have anything to validate here, since we only begin associating/create a group if this first validation check succeeds.

we could move this validation logic elsewhere, but that feels like against the spirit of these validations. they're typically used for form validation

app/views/organisations/edit.html.erb
samyou-softwire: this is used as an autocomplete on the org edit page for the group, so it most importantly must be an organisation in the current group

although, considering we do store the value that the user initially picked, we could use that if applicable. I've added this now
image

app/frontend/controllers/organisations_controller.js
samyou-softwire: I wasn't able to break it. considering these types come from existing org structure feels unlikely to need to add a new one in the near future

app/services/exports/organisation_export_service.rb
samyou-softwire: presumably this will come from the `organisation.attributes` setter on the top line, [organisation.xml](https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/3075/files#diff-197ff17081bb17a55966f5e39203567dd8d3b99b3c8f49abc9f5a590cdfbbfa8) suggests this will be exported

config/csv/organisations_structure/20250515_organisation_profit_status_and_group.csv
davidm-m: Yeah, I'm pretty sure that's looking for a file locally rather than on the server. Rake tasks connect to a specific database but run your local code rather than hooking into the remote ruby server.

samyou-softwire: for reference the way the docs describe to run a rake task on prod is to use `aws ecs run-task <details for a runner pointing to this task>`. I'm not sure if we would be able to get the csv to the task runner without having it as part of the codebase (& therefore part of the tasks that run in the cluster), so I think it may need to be here

davidm-m: Ooh alright yeah I'm used to running rake locally but that does sound like it needs to be on the task runner then. That's fun. Okay let's leave this committed then.

app/services/exports/organisation_export_service.rb
davidm-m: Nice, thanks for checking.

app/controllers/organisations_controller.rb
davidm-m: I wonder whether this is a rare occasion where having a controller level validation makes sense, because I don't think we should be persistently storing this id at all. What we care about is the group id, which we just use the member id to calculate at save time.

What do you think? And do you have an idea of how much extra effort it would be to 1) remove the `group_member_id` column from the database and 2) move that validation to the controller?

samyou-softwire: not too much more, I got most of the way yesterday before getting stuck at the validation. I'll give it a go now and we can see how it turns out

app/views/organisations/edit.html.erb
davidm-m: If we're getting rid of `group_member_id` we'll want to change this back to `oldest_group_member` 😅

samyou-softwire:
I'll close and reopen this, should hopefully fix the migrations failing

davidm-m (approved):
Nice, thanks for making those changes.

davidm-m (approved):

PR #3076, opened 2025-05-20
CLDC-2587: Remove deactivated user from log owner typeahead

samyou-softwire:
when generating candidate users for the 'Created by' question, ensure that deactivated users are included in that list

this does not change validation - there are valid reasons a log should be associated with an inactive user (users join/leave the service all the time but logs hang around). just makes it clearer for users on the site to not show inactive users in their organisation

blocked on some confirmation from MHCLG around whether this should apply for service managers too


davidm-m (approved):
Lovely, happy for this to be merged once you've confirmed about service managers.

davidm-m (approved):
Nice, and thanks for the comment explaining why we have two different active statuses.

PR #3077, opened 2025-06-03
CLDC-4038: Specify correct questions related to incorrect housingneeds

samyou-softwire:
for the housingneeds question, specifies which fields in the data export relate to it, in this case it should be fields 79-84

image


davidm-m (approved):

samyou-softwire:
rebasing onto main

PR #3078, opened 2025-06-03
CLDC-2594: Deactivated user login custom error

samyou-softwire:
# Description
adds a check upon login for whether they are an active user or not

uses agreed wording

# Screenshots
image


davidm-m (approved):

samyou-softwire:
rebasing onto main

PR #3080, opened 2025-06-06
CLDC-4039: Fix merge functionality not working

samyou-softwire:
catches a case where the location could be edited from a non nil value to a nil one, as the guard clause only checks if scheme is null

adds a verifying test

### DIFF
--- a/spec/services/merge/merge_organisations_service_spec.rb
+++ b/spec/services/merge/merge_organisations_service_spec.rb
@@ -715,12 +715,16 @@ RSpec.describe Merge::MergeOrganisationsService do
             create(:location, scheme:)
             incomplete_lettings_log = build(:lettings_log, scheme:, owning_organisation: merging_organisation, startdate: Time.zone.today)
             incomplete_lettings_log.save!(validate: false)
-            expect(Rails.logger).not_to receive(:error)
 
+            # if the location is overwritten with the nil one above, it will fail validation
+            # since a rollback will occur incomplete_lettings_log will not change so there's nothing to verify later
+            # so instead we verify that no rollback occurs
+            expect(Rails.logger).not_to receive(:error)
             merge_organisations_service.call
 
             incomplete_lettings_log.reload
 
+            # also ensure it wasn't overwritten with a valid location
             expect(incomplete_lettings_log.location).to be_nil
           end
 
--- a/spec/services/merge/merge_organisations_service_spec.rb
+++ b/spec/services/merge/merge_organisations_service_spec.rb
@@ -708,7 +708,7 @@ RSpec.describe Merge::MergeOrganisationsService do
           end
 
           it "does not change the lettings log location" do
-            create(:scheme, owning_organisation: merging_organisation)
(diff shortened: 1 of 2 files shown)


spec/services/merge/merge_organisations_service_spec.rb
```diff
@@ -1578,6 +1600,8 @@
           expect(new_absorbing_organisation.available_from.to_date).to eq(Time.zone.today)
         end
       end
+
+      context "and "
```
davidm-m: I assume this was a mistake? 😄

spec/services/merge/merge_organisations_service_spec.rb
```diff
@@ -707,6 +707,28 @@
...
+            let!(:incomplete_location) { create(:location, scheme:, name: nil, postcode: nil) }
+            let!(:complete_location) { create(:location, scheme:) }
+            let!(:complete_location2) { create(:location, scheme:) }
+
+            context "with lettings log with scheme set but no location" do
+              let!(:incomplete_lettings_log) { build(:lettings_log, scheme:, owning_organisation: merging_organisation, startdate: Time.zone.today) }
+
+              it "does not change the lettings log location" do
+                incomplete_lettings_log.save!(validate: false)
+                expect(Rails.logger).not_to receive(:error)
```
davidm-m: Feels like this is testing something different? Doesn't need to be removed I guess but a little strange to be part of this test.

davidm-m (changes requested):
One bit of code accidentally left in, otherwise looks great.

spec/services/merge/merge_organisations_service_spec.rb
samyou-softwire: yup

spec/services/merge/merge_organisations_service_spec.rb
samyou-softwire: the difficulty was that a letting logs with a location in this state is hard to get as an output, as it fails validation checks. if it fails validation it will error here instead of being nil. could make this more clear?

davidm-m: Does there being an error in validation not fail the test anyway? What I'm really questioning is having the `expect` there, makes it seem like part of the test is checking that lettings logs can be saved without validating them which isn't the purpose of this test.

Sidenote, not saying it isn't tricky but I feel like a lettings log in this state _shouldn't_ be hard to get because we save lettings logs in incomplete states all the time.

samyou-softwire: oh right I can restructure this as the layout is a bit confusing
- `incomplete_lettings_log.save!(validate: false)` yeah we're not testing this line. just much easier to setup a log like this & not have to set a bunch of unrelated properties to satisfy the validation rules
- the expect call has to go before we call the service. its not verifying the line above

have reworked things to be a little clearer

davidm-m: Yeah much clearer now, thanks!

davidm-m (approved):

samyou-softwire:
rebasing onto main

PR #3081, opened 2025-06-16
CLDC-4049: Ensure that if an Organisation merge fails due to a validation error it will be marked as incomplete

samyou-softwire:
no longer raise a rollback on merge validation fail

this caused the error to not be bubbled up correctly, meaning the catch in process_merge_request_job was not being triggered, and so not setting the merge back to ready

raising any exception will trigger a rollback


davidm-m (approved):
Nice, and thanks for checking for the new behaviour in all those tests too

PR #3082, opened 2025-06-17
CLDC-4030: Add some documentation on running rake tasks

samyou-softwire:
this is not included in our technical documentation right now, so wrote up some notes on the process


davidm-m (approved):
This is really helpful, thanks. Might even be worth making a bash script for it rather than having to copy and paste multiple lines of stuff but this will do for now.

If you can fix the linting then happy for this to be merged.

PR #3083, opened 2025-06-18
CLDC-4035: Improve our handling of commas in price inputs

samyou-softwire:
add the option to ignore commas in a price field (or other numeric/text fields)

happy to hear alterative implementations for this - there is no unifying currency type for questions so settled on a general parameter for this behaviour, though could look into making a specific currency type that ignores commas by default

### DIFF
--- a/app/controllers/form_controller.rb
+++ b/app/controllers/form_controller.rb
@@ -160,6 +160,9 @@ private
           result[option] = 1 if question_params.include?(option)
         end
       elsif question.type != "date"
+        if question.strip_commas
+          question_params = question_params.delete(",")
+        end
         result[question.id] = question_params
       end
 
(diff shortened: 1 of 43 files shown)


davidm-m:
I think this way is fine. The alternative in my mind would have been to automatically strip commas from all number fields - doesn't have to be an amount of money for people to put commas in. But this works too and it's clear what's going on.

davidm-m (approved):

samyou-softwire:
> The alternative in my mind would have been to automatically strip commas from all number fields - doesn't have to be an amount of money for people to put commas in

ah good point yeah I think that could work too and wouldn't be too tricky to add. I could run it by CORE to see if they want to apply this to all numbers too. would be nice to have this behaviour by default as it feels more natural for numbers

samyou-softwire:
core have confirmed that they want this for only price inputs

davidm-m:
> core have confirmed that they want this for only price inputs

Great, happy for this to be merged whenever.

PR #3084, opened 2025-06-24
CLDC-NONE: Document release process

samyou-softwire:
this is currently not included in our documentation as far as I can see


davidm-m (approved):

samyou-softwire:
checked and looks like this is documented already, so have improved this existing documentation a little instead

davidm-m (approved):

PR #3088, opened 2025-07-09
CLDC-4048: Add an exception for UPRN -> address matching for Sage Homes in 2024

samyou-softwire:
they've been having issues with address changing on upload. addresses are changed on an UPRN lookup happening so this is disabled for them

TBC from CORE whether we need to release this


samyou-softwire:
no longer required as we were able to upload manually fine

PR #3089, opened 2025-07-14
CLDC-4033: Improve flow for unconfirmed users

samyou-softwire:
send confirmation email if user is unconfirmed rather than password reset email

show text provided by CORE in this case as the regular password reset screen may be unclear

solves [CLDC-4033](https://mhclgdigital.atlassian.net/browse/CLDC-4033)

image

[CLDC-4033]: https://mhclgdigital.atlassian.net/browse/CLDC-4033?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/controllers/auth/passwords_controller.rb
+++ b/app/controllers/auth/passwords_controller.rb
@@ -66,7 +66,7 @@ protected
     resource.need_two_factor_authentication?(request) ? :updated_2FA : :updated
   end
 
-  def after_sending_reset_password_instructions_path_for(_resource)
+  def after_sending_reset_password_instructions_path_for(resource)
     account_password_reset_confirmation_path(email: params.dig("user", "email"), unconfirmed: resource.initial_confirmation_sent && !resource.confirmed?)
   end
 
--- a/app/controllers/auth/passwords_controller.rb
+++ b/app/controllers/auth/passwords_controller.rb
@@ -67,7 +67,7 @@ protected
   end
 
   def after_sending_reset_password_instructions_path_for(_resource)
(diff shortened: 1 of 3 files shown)


app/controllers/auth/passwords_controller.rb
```diff
@@ -66,7 +67,7 @@ def password_update_flash_message
   end
 
   def after_sending_reset_password_instructions_path_for(_resource)
-    account_password_reset_confirmation_path(email: params.dig("user", "email"))
+    account_password_reset_confirmation_path(email: params.dig("user", "email"), unconfirmed: !resource.confirmed?)
```
samyou-softwire: happy to hear on if this is the clearest way of doing this. did consider adding a new page potentially as well?

app/controllers/auth/passwords_controller.rb
```diff
@@ -66,7 +67,7 @@ def password_update_flash_message
   end
 
   def after_sending_reset_password_instructions_path_for(_resource)
```
davidm-m: The `_` prefix is used to indicate an argument that isn't being used, we should rename it because it's now used.

davidm-m (approved):

samyou-softwire:
@davidm-m on this which failed client QA, turns out the `_resource` variable that you renamed to `resource` was actually badly named itself as it is passed a resource name symbol (eg `:user`). calling `initial_confirmation_sent` was causing the error rachel saw. I've renamed this back and changed the name to `_resource_name` to make this more clear

davidm-m (approved):

PR #3095, opened 2025-08-06
CLDC-4074: Update tests for 2023 edit closing

samyou-softwire:
[CLDC-4074](https://mhclgdigital.atlassian.net/browse/CLDC-4074)

split into a separate PR from #3093 for clarity

after the edit window closed for 2024 logs this made 2023 logs inaccessible in many tests

these references have been removed and tests made more dynamic where possible

[CLDC-4074]: https://mhclgdigital.atlassian.net/browse/CLDC-4074?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


spec/lib/tasks/recalculate_invalid_reasonpref_dontknow_spec.rb
```diff
@@ -2,12 +2,19 @@
...
 RSpec.describe "recalculate_invalid_reasonpref_dontknow" do
+  include CollectionTimeHelper
+
   subject(:task) { Rake::Task["recalculate_invalid_rpdontknow"] }
 
   before do
     Rake.application.rake_require("tasks/recalculate_invalid_reasonpref_dontknow")
     Rake::Task.define_task(:environment)
     task.reenable
+    Timecop.freeze(previous_collection_end_date)
```
samyou-softwire, line 13: fixing these rake tasks is tricky because they specifically only work for 2024 logs. we can use timecop to go back but these will break again next year. I think by that time we can just remove this rake task & its associated tests

spec/models/log_spec.rb
```diff
@@ -62,12 +64,12 @@
...
 
     context "when a non setup field is invalid for a lettings log" do
-      subject(:model) { build_stubbed(:lettings_log, :completed, startdate: Time.zone.local(2023, 12, 12), offered: 234) }
+      subject(:model) { build_stubbed(:lettings_log, :completed, startdate: current_collection_start_date, age1: 234) }
 
       it "blanks it" do
         model.valid?
         model.blank_invalid_non_setup_fields!
-        expect(model.offered).to be_nil
+        expect(model.age1).to be_nil
```
samyou-softwire, line 72: `offered` turned out to no longer be on the route by default so swapped it

spec/models/validations/household_validations_spec.rb
```diff
@@ -50,17 +50,6 @@
         expect(record.errors["reasonother"]).to be_empty
       end
 
-      context "when form year is before 2024" do
```
samyou-softwire, line 53: most tests to do with pre 2024 logic were removed

spec/models/validations/sales/sale_information_validations_spec.rb
```diff
@@ -68,17 +70,29 @@
...
 
-          expect(record.errors[:hodate]).to be_present
+          expect(record.errors[:hodate]).to be_empty
+          expect(record.errors[:saledate]).to be_empty
         end
       end
     end
 
+    context "when hodate 5 or more years before saledate" do
+      let(:record) { build(:sales_log, hodate: current_collection_start_date - 5.years, saledate: current_collection_start_date) }
```
samyou-softwire, line 86: there's some calcs going on with the dates here, wonder if this is the most resilient way?

spec/requests/duplicate_logs_controller_spec.rb
```diff
@@ -64,11 +54,11 @@
             end
 
             it "displays check your answers for each log with correct questions" do
-              expect(page).to have_content("Q5 - Tenancy start date", count: 3)
-              expect(page).to have_content("Q7 - Tenant code", count: 3)
-              expect(page).to have_content("Q32 - Lead tenant’s age", count: 3)
-              expect(page).to have_content("Q33 - Lead tenant’s gender identity", count: 3)
-              expect(page).to have_content("Q37 - Lead tenant’s working situation", count: 3)
+              expect(page).to have_content("- Tenancy start date", count: 3)
```
samyou-softwire, line 57: decided to omit question numbers here as it makes the tests more brittle than they need to be

samyou-softwire:
did also update the [docs](https://mhclgdigital.atlassian.net/wiki/spaces/MC/pages/477593604/Check+the+automated+tests+will+pass+after+1st+April) to mention this edge case to check

spec/lib/tasks/recalculate_invalid_reasonpref_dontknow_spec.rb
davidm-m: Yeah, if a task is for a specific year it should be deleted when that collection closes.

spec/models/validations/sales/household_validations_spec.rb
```diff
@@ -1,279 +1,121 @@
...
   let(:validator_class) { Class.new { include Validations::Sales::HouseholdValidations } }
   let(:record) { build(:sales_log, saledate:) }
   let(:saledate) { Time.zone.now }
 
   describe "#validate_partner_count" do
-    let(:saledate) { Time.zone.local(2023, 4, 1) }
+    let(:saledate) { current_collection_start_date }
 
-    it "validates that only 1 partner exists" do
+    it "does not validate" do
```
davidm-m: Not a very helpful test name. Does `validate_partner_count` not do anything anymore?

spec/models/validations/sales/property_validations_spec.rb
```diff
@@ -1,80 +1,21 @@
...
-      it "when postcodes do not match an error is added for non joint purchase" do
-        record.postcode_full = "SW1A 1AA"
-        record.ppostcode_full = "SW1A 0AA"
-        record.jointpur = 2
-        property_validator.validate_postcodes_match_if_discounted_ownership(record)
-        expect(record.errors["postcode_full"]).to include("Buyer’s last accommodation and discounted ownership postcodes must match.")
-        expect(record.errors["ppostcode_full"]).to include("Buyer’s last accommodation and discounted ownership postcodes must match.")
-        expect(record.errors["ownershipsch"]).to include("Buyer’s last accommodation and discounted ownership postcodes must match.")
-      end
+    it "is not validated" do
```
davidm-m: Again not a very helpful test name, it should have some kind of description of why this case isn't being validated

spec/models/validations/sales/sale_information_validations_spec.rb
davidm-m: Yeah this looks fine to me. Very clear what we're doing and why, and unless the validation is weird (which imo would be a bug with the validation) should be resilient.

davidm-m (changes requested):
Phew, there's a lot here and you've done some great work making it more resilient when form years change! Got a couple of places I wanted to ask about but this is mostly really good.

spec/models/validations/sales/household_validations_spec.rb
samyou-softwire: for forms post 2024 no
```ruby
  def validate_partner_count(record)
    return if record.form.start_year_2024_or_later?

    ...
  end

spec/models/validations/sales/property_validations_spec.rb
samyou-softwire: also not validated after 2024. I can update the test names to account for this
```ruby
  def validate_postcodes_match_if_discounted_ownership(record)
    return unless record.saledate && !record.form.start_year_2024_or_later?

    ...
  end

davidm-m (approved):
Nice! Thanks for changing those test names.

PR #3097, opened 2025-08-13
CLDC-4055: Use name property for filtering organisations

samyou-softwire:
[CLDC-4055](https://mhclgdigital.atlassian.net/browse/CLDC-4055)

uses ruby code instead to do the organisation filtering. this will be more costly per search, though the org amount is 982 so this shouldn't be too costly

[CLDC-4055]: https://mhclgdigital.atlassian.net/browse/CLDC-4055?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/models/organisation.rb
+++ b/app/models/organisation.rb
@@ -38,7 +38,13 @@ class Organisation < ApplicationRecord
     Organisation.where(id: ids)
   end
 
-  scope :search_by_name, ->(name) { where("name ILIKE ?", "%#{name}%") }
+  scope :search_by_name, lambda { |name|
+    # we can't use the activerecord queries to query the DB as the `name` col doesn't update when the org changes name.
+    # this is due to the info being stored in the organisation_name_changes table.
+    # so we load all and filter in memory. the number of organisations is small so this should be fine.
+    matched_ids = all.filter { |org| org.name.downcase.include?(name.downcase) }.map(&:id)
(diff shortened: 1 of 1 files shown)


app/models/organisation.rb
```diff
@@ -38,7 +38,13 @@ def affiliated_stock_owners
     Organisation.where(id: ids)
   end
 
-  scope :search_by_name, ->(name) { where("name ILIKE ?", "%#{name}%") }
+  scope :search_by_name, lambda { |name|
+    # we can't use the activerecord queries to query the DB as the `name` col doesn't update when the org changes name.
+    # this is due to the info being stored in the organisation_name_changes table.
+    # so we load all and filter in memory. the number of organisations is small so this should be fine.
+    matched_ids = all.filter { |org| org.name.downcase.include?(name.downcase) }.map(&:id)
+    where(id: matched_ids)
```
davidm-m, line 46: I think this is actually possible to keep on the db side, probably by making `name` itself a scope instead of a function, but I agree this isn't going to have a performance impact with so few orgs.

samyou-softwire:
@davidm-m is this good to go?

davidm-m (approved):

PR #3102, opened 2025-09-02
CLDC-4082: Allow for an unresolved log to redirect to the correct location page

samyou-softwire:
closes https://mhclgdigital.atlassian.net/browse/closes CLDC-4082

previously would always go to location, causing issues for schemes with many locations which should send to location_search page

### DIFF
--- a/app/models/form/lettings/pages/scheme.rb
+++ b/app/models/form/lettings/pages/scheme.rb
@@ -6,7 +6,7 @@ class Form::Lettings::Pages::Scheme < ::Form::Page
         "needstype" => 2,
       },
     ]
-    @next_unresolved_page_id = "location"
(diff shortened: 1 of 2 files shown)


app/models/form/lettings/pages/scheme.rb
```diff
@@ -6,7 +6,7 @@ def initialize(_id, hsh, subsection)
         "needstype" => 2,
       },
     ]
-    @next_unresolved_page_id = "location"
+    @next_unresolved_page_id = :location_or_location_search
```
davidm-m, line 9: Just to double check, this is the only place where we redirect to location?

samyou-softwire: yes
core usages

davidm-m (approved):

PR #3103, opened 2025-09-08
CLDC-4044: Allow unconfirmed but logged in (reconfirmed) users to be reinvited

samyou-softwire:
closes [CLDC-4044](https://mhclgdigital.atlassian.net/browse/CLDC-4044)

this is to ensure that the code considers them a fresh user

specifically, ensures that the "Resend invite link" button will show

[CLDC-4044]: https://mhclgdigital.atlassian.net/browse/CLDC-4044?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -167,6 +167,11 @@ class User < ApplicationRecord
     update!(
       active: true,
       reactivate_with_organisation: false,
+      # resetting these fields ensures that the 'resend confirmation instructions' button shows
+      # we have this button be based on sign in date than confirmation status to ensure that the
+      # user has successfully completed the entire login flow before we hide the button
+      last_sign_in_at: nil,
+      last_sign_in_ip: nil,
     )
   end
 
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -167,11 +167,6 @@ class User < ApplicationRecord
     update!(
       active: true,
       reactivate_with_organisation: false,
-      # resetting these fields ensures that the 'resend confirmation instructions' button shows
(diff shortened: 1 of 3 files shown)


davidm-m (approved):

samyou-softwire:
@davidm-m prompted by the failing test took another look at this ticket. it looks like CORE does have functional differences for users that are unconfirmed but have logged in before; it uses this to provide custom email templates to "reconfirmed" users

if we reset login date we lose the ability to track reconfirmed users vs unconfirmed users & to avoid adding a new db column have gone for the more simple solution to accept this as intended & allow for invites to be sent if you're unconfirmed or never logged in

have added a comment inline about this, lmk if there's a better place to put this

davidm-m:
> have added a comment inline about this, lmk if there's a better place to put this

@samyou-softwire I'm not seeing a comment about this?

It does make sense to me though and in many ways it's good not to lose that last sign in info anyway! Happy with this approach.

samyou-softwire:
> > have added a comment inline about this, lmk if there's a better place to put this
> 
> @samyou-softwire I'm not seeing a comment about this?
> 
> It does make sense to me though and in many ways it's good not to lose that last sign in info anyway! Happy with this approach.

ah apologies pushed

davidm-m (approved):

davidm-m:
Perfect, thanks!

PR #3104, opened 2025-09-11
CLDC-4066: Increase our resilience to an OS Places outage

samyou-softwire:
closes [CLDC-4066](https://mhclgdigital.atlassian.net/browse/CLDC-4066)

helps the app be much more resilient when OS Places API has periods of slow responses

adds a debounce to the address search. requests will wait 0.5 seconds and only the last request will be sent. previously a request would be sent for every keypress which was quite wasteful

reduces the max wait time for OS Places API requests from 90 seconds to 45

both of these changes should help the app not be susceptible to all threads being stuck waiting on OS Places API requests which could cause an outage

[CLDC-4066]: https://mhclgdigital.atlassian.net/browse/CLDC-4066?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/services/uprn_client.rb
+++ b/app/services/uprn_client.rb
@@ -39,7 +39,7 @@ private
     client.use_ssl = true
     client.verify_mode = OpenSSL::SSL::VERIFY_PEER
     client.max_retries = 3
-    client.read_timeout = 30 # seconds
+    client.read_timeout = 20 # seconds
     client
   end
 
--- a/app/services/uprn_client.rb
+++ b/app/services/uprn_client.rb
@@ -39,7 +39,7 @@ private
     client.use_ssl = true
     client.verify_mode = OpenSSL::SSL::VERIFY_PEER
     client.max_retries = 3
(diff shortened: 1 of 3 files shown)


app/services/uprn_client.rb
```diff
@@ -39,7 +39,7 @@ def http_client
     client.use_ssl = true
     client.verify_mode = OpenSSL::SSL::VERIFY_PEER
     client.max_retries = 3
-    client.read_timeout = 30 # seconds
+    client.read_timeout = 20 # seconds
```
davidm-m: Why are we making this different to the address client timeout?

samyou-softwire: ah good spot thanks thats my mistake; intended for them to both be 15 (unless we have a want for them to be different)

davidm-m (approved):

PR #3105, opened 2025-09-16
CLDC-4028: Ensure changes to dependent objects are included in export

samyou-softwire:
closes [CLDC-4028](https://mhclgdigital.atlassian.net/browse/CLDC-4028)

currently in most cases in the app a record will be included in an export if one of its own values updates (as this changes the `updated_at` value). this meant in cases of joins if a joined value updated this would not count as an update in the base record and so it would not be included in an export

this PR resolves this by expanding the scope of the export query code to specifically check if a related value is updated. this seemed most sensible as changing the model itself to account for this would be difficult as there are many places where values can update

restructures the code a little to cut down on duplication between the closed & open ended range exports

[CLDC-4028]: https://mhclgdigital.atlassian.net/browse/CLDC-4028?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/services/exports/lettings_log_export_service.rb
+++ b/app/services/exports/lettings_log_export_service.rb
@@ -30,14 +30,32 @@ module Exports
       "core_#{year}_#{year + 1}_apr_mar_#{base_number_str}_#{increment_str}".downcase
     end
 
-    def retrieve_resources(recent_export, full_update, year)
-      if !full_update && recent_export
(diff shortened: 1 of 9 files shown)


davidm-m (approved):
Awesome, all looks good.

PR #3106, opened 2025-09-17
CLDC-NONE: Improve README

samyou-softwire:
specify rbenv exec

remove confusing/outdated steps

include steps for RubyMine

### DIFF
--- a/docs/setup.md
+++ b/docs/setup.md
@@ -230,6 +230,12 @@ bundle exec rspec ./spec/path/to/file.rb
 
 or run individual files/tests from your IDE
 
+If you have made database changes, you may need to run the migrations for the test database:
+
+```bash
+bundle exec rake db:migrate RAILS_ENV=test
+```
+
 ### Feature toggles
 
 Feature toggles can be found in `app/services/feature_toggle.rb`
--- a/docs/setup.md
+++ b/docs/setup.md
@@ -121,19 +121,19 @@ Also ensure you have firefox installed
 2. Install the dependencies:
 
    ```bash
-   rbenv exec bundle install && yarn install
+   bundle install && yarn install
    ```
 
 3. Create the database & run migrations:
 
    ```bash
-   rbenv exec bundle exec rake db:create db:migrate
+   bundle exec rake db:create db:migrate
    ```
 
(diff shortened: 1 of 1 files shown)


docs/setup.md
```diff
@@ -84,25 +88,10 @@ We recommend using [nvm](https://github.com/nvm-sh/nvm) to manage NodeJS version
...
    ```
 
-   or you could run it without specifying the version and it should use the version from .nvmrc
-
-   ```bash
-   nvm install
-   nvm use
-   brew install yarn
-   ```
-
```
samyou-softwire, line 94: this step appears before cloning the repo so isn't very relevant

docs/setup.md
```diff
@@ -84,25 +88,10 @@ We recommend using [nvm](https://github.com/nvm-sh/nvm) to manage NodeJS version
...
 
    ```bash
-   curl -sL https://deb.nodesource.com/setup_20.x | sudo bash -
-   sudo apt -y install nodejs
-   mkdir -p ~/.npm-packages
-   npm config set prefix ~/.npm-packages
-   echo 'NPM_PACKAGES="~/.npm-packages"' >> ~/.bashrc
-   echo 'export PATH="$PATH:$NPM_PACKAGES/bin"' >> ~/.bashrc
-   source ~/.bashrc
-   npm install --location=global yarn
```
samyou-softwire, line 105: this seems like a mistake as to why we'd install node again. replaced with the much simpler yarn install command

docs/setup.md
```diff
@@ -132,42 +121,50 @@ Also ensure you have firefox installed
 2. Install the dependencies:
 
    ```bash
-   bundle install && yarn install
+   rbenv exec bundle install && yarn install
```
davidm-m: Why do we need to rbenv exec this? Because we have a `.ruby-version` file rbenv will automatically detect and use the correct version.
image
See me testing this by switching between two codebases with different ruby versions here.

samyou-softwire: ah thats a great point! I remember being told rbenv exec when I first ramped up but looking online it seems that this is unnecessary here. have removed all the references now

davidm-m (approved):

PR #3109, opened 2025-09-29
CLDC-4028: Ensure logs will update when an organisation name change is created

samyou-softwire:
closes [CLDC-4028](https://mhclgdigital.atlassian.net/browse/CLDC-4028)

this could lead to redundant exports if the name change does not impact the log at its startdate (which we put into the log), but it'd be more difficult to query this way so fine to be safe here

```ruby
attribute_hash["owningorgid"] = lettings_log.owning_organisation.old_visible_id || (lettings_log.owning_organisation.id + LOG_ID_OFFSET)
attribute_hash["owningorgname"] = lettings_log.owning_organisation.name(date: lettings_log.startdate)
attribute_hash["hcnum"] = lettings_log.owning_organisation.housing_registration_no
```

could look into further improving the query if the name change would cause a name change at the start date of the log but this feels quite tricky to do (needs to be both before log start date but not overwritten by a future name change)

also query name changes by updated_at rather than created_at. shouldn't really matter as we don't expect name changes to be updated but worth being safe in case we need to update a name change for maintenance reasons

[CLDC-4028]: https://mhclgdigital.atlassian.net/browse/CLDC-4028?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/services/exports/lettings_log_export_service.rb
+++ b/app/services/exports/lettings_log_export_service.rb
@@ -58,10 +58,10 @@ module Exports
 
       # these must be separate as activerecord struggles to join to two different name change tables in the same query
       ids.concat(
-        relation.left_joins(owning_organisation: :organisation_name_changes).where(owning_organisation: { organisation_name_changes: { created_at: range } }).pluck(:id),
+        relation.left_joins(owning_organisation: :organisation_name_changes).where(owning_organisation: { organisation_name_changes: { updated_at: range } }).pluck(:id),
       )
       ids.concat(
-        relation.left_joins(managing_organisation: :organisation_name_changes).where(managing_organisation: { organisation_name_changes: { created_at: range } }).pluck(:id),
+        relation.left_joins(managing_organisation: :organisation_name_changes).where(managing_organisation: { organisation_name_changes: { updated_at: range } }).pluck(:id),
       )
 
       LettingsLog.where(id: ids)
--- a/app/services/exports/lettings_log_export_service.rb
+++ b/app/services/exports/lettings_log_export_service.rb
@@ -31,7 +31,8 @@ module Exports
     end
 
     def retrieve_resources_from_range(range, year)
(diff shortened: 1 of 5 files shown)


app/services/exports/lettings_log_export_service.rb
```diff
@@ -55,6 +55,14 @@ def retrieve_resources_from_range(range, year)
         )
         .pluck(:id)
 
+      # these must be separate as activerecord struggles to join to two different name change tables in the same query
+      ids.concat(
```
samyou-softwire, line 60: could look into seeing if this can be done in a single query though at a quick glance the only options I could find call for writing raw SQL

app/services/exports/lettings_log_export_service.rb
```diff
@@ -55,6 +56,14 @@ def retrieve_resources_from_range(range, year)
...
+        relation.left_joins(owning_organisation: :organisation_name_changes).where(owning_organisation: { organisation_name_changes: { created_at: range } }).pluck(:id),
+      )
+      ids.concat(
+        relation.left_joins(managing_organisation: :organisation_name_changes).where(managing_organisation: { organisation_name_changes: { created_at: range } }).pluck(:id),
```
davidm-m: These are both using `created_at` instead of `updated_at`.

davidm-m (changes requested):
Just the one place where we're still using `created_at`, otherwise looks sensible to me.

app/services/exports/lettings_log_export_service.rb
samyou-softwire: good spot thanks, forgot to update this one

davidm-m (approved):

PR #3114, opened 2025-10-08
CLDC-4090: BU validation of LA ECode

samyou-softwire:
closes [CLDC-4090](https://mhclgdigital.atlassian.net/browse/CLDC-4090)

adds new validation rules that the start date of a log is after the start date of its LA and before its end date, if one is present

this means logs can only be associated with the appropriate logs

[CLDC-4090]: https://mhclgdigital.atlassian.net/browse/CLDC-4090?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/models/validations/property_validations.rb
+++ b/app/models/validations/property_validations.rb
@@ -76,4 +76,39 @@ module Validations::PropertyValidations
       record.errors.add :startdate, I18n.t("validations.lettings.property.startdate.location_not_in_england")
     end
   end
+
+  def validate_la_is_active(record)
+    return unless record.form.start_year_2025_or_later?
+
+    if record.is_general_needs?
+      return unless record.la
+
+      la = LocalAuthority.england.find_by(code: record.la)
+
+      # will be caught by the not in england validation
+      return if la.nil?
+      # only compare end date if it exists
+      return if record.startdate >= la.start_date && (la.end_date.nil? || record.startdate <= la.end_date)
+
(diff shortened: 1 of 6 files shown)


davidm-m (approved):
Looks good, I think this will resolve the issue. It feels quite duplicated though. Is there any way to link together the sales and lettings validations? Either centralise the actual check somewhere or just mention that if you're changing one you should check the other?

samyou-softwire:
> Looks good, I think this will resolve the issue. It feels quite duplicated though. Is there any way to link together the sales and lettings validations? Either centralise the actual check somewhere or just mention that if you're changing one you should check the other?

it's possible by creating a new class that is imported into both sales & lettings logs, though this'd split the validations & the validation works differently between sales & lettings as only one needs to consider the location postcode, so have opted to comment

davidm-m (approved):

samyou-softwire:
found by CORE on the ticket; the errors weren't working sales BU. looked to be due to adding errors to a field that isn't used anymore. this also appears to impact the existing LA in UK check which also wasn't working on sales uploads.

have removed the error on both validations and it seems to work now https://review.submit-social-housing-data.communities.gov.uk/3114/sales-logs/bulk-upload-results/11 https://review.submit-social-housing-data.communities.gov.uk/3114/sales-logs/bulk-upload-results/12

davidm-m (approved):

PR #3124, opened 2025-11-27
CLDC-4029: Add additional user filters

kwunkiuma:
Add new filters for "role type" and "additional responsibilities" to user views.

Users view
{33E45B5A-1B7D-46F1-BA03-43811AB9822B}

Filtered by role
{22321C24-7997-401E-9088-A5710688DE8C}

Filtered by role and additional responsibilities
{F75B5A8C-DB27-4828-8608-34E439FF3D40}

As a support user
{07837D51-030D-4022-949F-638FA43994A3}

Filtered by support role
{A9A9369F-0667-4DFB-BC35-7B1B3710460D}


app/models/user.rb
```diff
@@ -81,6 +81,29 @@ class User < ApplicationRecord
...
+    filtered_records = all
+    scopes = []
+
+    additional_responsibilities.each do |responsibility|
+      case responsibility
+      when "key_contact"
+        scopes << send("is_key_contact")
+      when "data_protection_officer"
+        scopes << send("is_data_protection_officer")
+      end
```
samyou-softwire, line 95: `send` is only required when the string you pass is dynamic (its a function that works a lot like eval in other languages). since we're using a constant string we can just reference the scope directly
```suggestion
      when "key_contact"
        scopes << is_key_contact
      when "data_protection_officer"
        scopes << is_data_protection_officer
      end
```

samyou-softwire (changes requested):
one usage of `send` we can remove otherwise LGTM

davidm-m (approved):
Approved! (So long as you do the markups Sammy's asked for)

app/models/user.rb
samyou-softwire: I've fixed this as we've now handed over

spec/services/filter_manager_spec.rb
```diff
@@ -94,4 +94,51 @@
       expect(described_class.filter_schemes(Scheme.all, nil, {}, nil, nil)).to eq(alphabetical_order_schemes)
     end
   end
+
+  describe "filter_users" do
```
natdeanlewissoftwire, line 98: could we test that the support user can be selected and and returned in the reponse ONLY when the user is a support user? I think we should add to the "when filtering users" tests in `user_spec.rb` to do this.

app/helpers/filters_helper.rb
```diff
@@ -52,6 +52,22 @@ def user_status_filters
...
   end
 
+  def user_role_type_filters(include_support: false)
+    roles = {
+      "1" => "Data provider",
+      "2" => "Coordinator",
+    }
+    roles["99"] = "Support" if include_support
+    roles.freeze
+  end
```
natdeanlewissoftwire, line 62:
```suggestion
  def user_role_type_filters(include_support: false)
    roles = {
      "data_provider" => "Data provider",
      "data_coordinator" => "Data coordinator",
    }
    roles["support"] = "Support" if include_support
    roles.freeze
  end
```

natdeanlewissoftwire: I think we can just used `data_provider` as the key here etc. and it won't make any difference to rails as it infers from the numeric values anyway.
I think it's more readable with the snake case strings, and we're sending them in the tests anyway so we may as well just use them here imo.

Then we're visually consistent with `user_additional_responsibilities_filters` below too

natdeanlewissoftwire: Note - I also think this should be "Data coordinator", not "Coordinator" for consistency with the language we already use in the UI. I don't think that should hold up PO but maybe worth adding a comment to the ticket for Rachel if you agree?

natdeanlewissoftwire (commented):
LGTM! Some small ideas but seems solid as is.

spec/services/filter_manager_spec.rb
samyou-softwire: ah nice, I've added tests for this

app/helpers/filters_helper.rb
samyou-softwire: agree on both, I'll check in with Rachel. there's not particular hurry on this so might as well get it confirmed

natdeanlewissoftwire: Sounds good, approving pending the outcome of that copy change :)

natdeanlewissoftwire (approved):
LGTM!

samyou-softwire (approved):

PR #3128, opened 2025-12-11
CLDC-NONE: Split deployments page from infrastructure in docs

samyou-softwire:
makes these vital steps easier to find

docs page


docs/rake.md
```diff
@@ -1,5 +1,5 @@
 ---
-nav_order: 11
+nav_order: 13
```
carolynbarker, line 2: So this and bulk_upload both used to be 11? 😮

carolynbarker: [seems that ](https://just-the-docs.com/docs/navigation/main/order/) multiple pages having the same number isn't a disaster, but is best avoided
>The order of pages with equal nav_order parameters is unstable: it may change with each build

docs/infrastructure.md
```diff
@@ -57,68 +57,5 @@ Where to find the Infrastructure?
...
-  - You may need to manually re-enter secret values into Secrets Manager at some point. When you do, just paste the secret value as plain text (don’t enter a key name, or format it as JSON).
-- ECS
-  - Sometimes task definitions don’t get deleted. You may need to manually delete them.
-  - After destroying the db, you’ll need to make sure the ad hoc ECS task which seeds the database gets run in order to set up the database correctly.
-- SNS
-  - When creating an email subscription in an environment, Terraform will look up the email to use as the subscription endpoint from Secrets Manager. If you haven’t already created this (e.g. by running terraform apply -target="module.monitoring" -var="create_secrets_first=true") then this will lead to the subscription creation erroring, because it can’t retrieve the value of the secret (because it doesn’t exist yet). If this happens, remember you’ll need to go to Secrets Manager in the console and enter the desired email (as plaintext, no quotation marks or anything else required) as the value of the secret (which is most likely called MONITORING_EMAIL). Then run another apply with Terraform and this time it should succeed.
-
```
carolynbarker, line 122: This bit isn't in the other page, was it intentionally deleted?

docs/infrastructure.md
```diff
@@ -57,68 +57,5 @@ Where to find the Infrastructure?
...
-- S3
-  - Terraform won’t let you delete buckets that have objects in them.
-- Secrets
-  - If you destroy secrets, they will actually be marked as ‘scheduled to delete’ which will take effect after a minimum of 7 days. You can’t recreate secrets with the same name during this period. If you want to destroy immediately, you need to do it from the command line (using your staging developer role, rather than your MHCLG-wide role used to apply Terraform) with this command: aws secretsmanager delete-secret --force-delete-without-recovery --secret-id <secret-arn>. (Note that if a secret is marked as scheduled to delete, you can undo this in the console to make it an ‘active’ secret again.)
```
carolynbarker, line 115: The bit about destroying immediately was deleted, is that deliberate?

carolynbarker: happy to remove bits that aren't very relevant, but I wonder if this is something it would be useful to know when finding out that secrets will otherwise take 7 days to delete

samyou-softwire: the initial part about secrets taking 7 days to delete should still be in the new documentation, I removed the latter bit about immediate deletion as we can't use terraform commands any more

docs/infrastructure.md
samyou-softwire: yeah, it's not something we can do any more as we can't run terraform commands

docs/infrastructure.md
carolynbarker: do we need terraform for the secretsmanager command? it sounds to me like it's contrasting with terraform so the secretsmanager bit would still apply (except for the reference to Terraform which we can lose)

> If you want to destroy immediately, you need to do it from the command line (using your staging developer role, rather than your MHCLG-wide role used to apply Terraform) with this command: aws secretsmanager delete-secret --force-delete-without-recovery --secret-id <secret-arn>. (Note that if a secret is marked as scheduled to delete, you can undo this in the console to make it an ‘active’ secret again.)

samyou-softwire: ah good point this is still doable. have added it back

carolynbarker (approved):

PR #3133, opened 2025-12-19
CLDC-3499: Clean up old code

katlan-sw:
# CLDC-3499: Clean up old code

A partial code clean-up of old code. Things removed:

- Old form schemas (no longer used to generate forms)
- Old form locale files
- Old form validation locale files

Potential future work:

- Slimming down the dictionaries mapping years to question numbers. OR refactoring it to be calculated differently.
- Cleaning up methods like "sales_or_start_year_after_2022" and replacing usage instances with a logical `true`.


samyou-softwire:
it looks like some of the test seed data is still based in 2023 so it's failing to start the tests, think we'll need to resolve this

these changes seem fine though there may be a task of removing 2023 and before logs from CORE as part of this so we can be sure these code paths aren't going to be used

katlan-sw:
Is it worth spending more time on this or should I just focus on the sprint tickets?

samyou-softwire:
As mentioned at refinement, let's park this one for now. Perhaps let's close the PR (to deprovision the resources) and note the branch you've made changes to on the ticket?

PR #3134, opened 2025-12-22
CLDC-4162: Set up Bulk Uploads for 2026

samyou-softwire:
closes [CLDC-4162](https://mhclgdigital.atlassian.net/browse/CLDC-4162)

following our docs as listed on ticket, makes the necessary changes to allow for 2026 logs to be uploaded via bulk upload

adds necessary tests

updates existing test files too. integrated the collection time helpers into a couple to stop them being year specific

[CLDC-4162]: https://mhclgdigital.atlassian.net/browse/CLDC-4162?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/controllers/test_data_controller.rb
```diff
@@ -17,12 +17,13 @@ def create_setup_test_lettings_log
...
   end
 
-  %w[2024 2025].each do |year|
+  %w[2025 2026].each do |year|
     define_method("create_#{year}_test_lettings_bulk_upload") do
       return render_not_found unless FeatureToggle.create_test_logs_enabled?
 
       file = Tempfile.new("#{year}_test_lettings_log.csv")
-      log = FactoryBot.create(:lettings_log, :completed, assigned_to: current_user, ppostcode_full: "SW1A 1AA", startdate: generate_different_date_within_collection_year(Time.zone.local(year.to_i, 4, 1), end_date_override: Time.zone.now + 14.days))
+      end_date_override = year == current_collection_start_year ? Time.zone.now + 14.days : collection_start_date(Time.zone.local(year.to_i, 4, 1)) + 14.days
```
carolynbarker: what's this logic doing?

I think I see why `Time.zone.now + 14.days` isn't right - that would fail for the year 2026.

Do we actually need an end_date_override, or would `generate_different_date_within_collection_year` give something sensible even with no override? I would have hoped it would, but the fact that we've added an end_date_override in the first places makes me wonder if something goes wrong without it

spec/services/bulk_upload/lettings/validator_spec.rb
```diff
@@ -165,8 +169,8 @@
         expect(error.tenant_code).to eql(log.tenancycode)
         expect(error.property_ref).to eql(log.propcode)
         expect(error.row).to eql("2")
-        expect(error.cell).to eql("CY2")
-        expect(error.col).to eql("CY")
+        expect(error.cell).to eql("CX2")
+        expect(error.col).to eql("CX")
```
carolynbarker: is this going to change each year?

spec/services/bulk_upload/sales/log_creator_spec.rb
```diff
@@ -1,187 +1,171 @@
 require "rails_helper"
 
 RSpec.describe BulkUpload::Sales::LogCreator do
+  include CollectionTimeHelper
+
   subject(:service) { described_class.new(bulk_upload:, path: "") }
 
+  let(:year) { current_collection_start_year }
```
carolynbarker, line 8: shall we test the previous_collection_start_year as well? I'm thinking particularly of the crossover period when we would want to be sure both years' logs were working

carolynbarker: although, this isn't actually the year specific logic, is it, so maybe not necessary - have I got that right?

spec/services/bulk_upload/sales/validator_spec.rb
```diff
@@ -151,14 +139,14 @@
...
 
-        expect(error.field).to eql("field_1")
+        expect(error.field).to eql("field_4")
         expect(error.error).to eql("You must answer owning organisation.")
         expect(error.purchaser_code).to eql(log.purchaser_code)
         expect(error.row).to eql("2")
-        expect(error.cell).to eql("B2")
-        expect(error.col).to eql("B")
+        expect(error.cell).to eql("E2")
+        expect(error.col).to eql("E")
```
carolynbarker: again these seem year specific, I'm keen that we think carefully about how to handle that. Would these start failing on 1st April?

I don't love hardcoding years but that would at least make it clear that the test is year-specific - maybe we can think of a nicer solution though

app/helpers/bulk_upload/lettings_log_to_csv.rb
```diff
@@ -91,6 +91,15 @@ def default_2025_field_numbers
     (1..129).to_a
   end
 
+  def default_2026_field_numbers
+    (1..129).to_a
```
carolynbarker, line 96: this may also need a todo, unless we actually do coincidentally work out to exactly the same number of questions as last year?

app/helpers/bulk_upload/sales_log_to_csv.rb
```diff
@@ -69,6 +69,8 @@ def default_field_numbers_for_year(year)
       (1..131).to_a
     when 2025
       (1..121).to_a
+    when 2026
+      (1..121).to_a
```
carolynbarker, line 74: this may also need a todo, unless we actually do coincidentally work out to exactly the same number of questions as last year?

app/services/bulk_upload/lettings/year2026/csv_parser.rb
```diff
@@ -0,0 +1,122 @@
+require "csv"
+
+class BulkUpload::Lettings::Year2026::CsvParser
+  include CollectionTimeHelper
+
+  FIELDS = 129
```
carolynbarker, line 7: FIELDS and @cols both here and in the sales one might need updating. if we don't have them yet that can be done later, but maybe worth adding a comment to avoid confusion in the meantime.

(unless you've checked and there are going to be 129 fields again, in which case this is fine!)

oscar-richardson-softwire:
It seems like this branch is quite out of date with `main` and could do with a rebase (and then https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/3136 which targets this branch).

app/controllers/test_data_controller.rb
samyou-softwire: I looked into this a bit further, I think there's a clearer way to do this

we only need to care about the 14 days log limit when the future forms feature flag is turned _off_, which it will be after we release the forms

eg in `setup_validations.rb` for lettings logs validates startdate with:
```ruby
return unless record.startdate && date_valid?("startdate", record) && !FeatureToggle.allow_future_form_use?
```

so have replaced this line to turn off the end date override when we have this feature flag on

app/helpers/bulk_upload/lettings_log_to_csv.rb
samyou-softwire: have updated

app/helpers/bulk_upload/sales_log_to_csv.rb
samyou-softwire: have updated

app/services/bulk_upload/lettings/year2026/csv_parser.rb
samyou-softwire: have added a todo

spec/services/bulk_upload/lettings/validator_spec.rb
samyou-softwire: depends on the year, though I think broadly yes as the specific col for field_4 moves around

spec/services/bulk_upload/sales/log_creator_spec.rb
samyou-softwire: these tests don't seem to be year specific

spec/services/bulk_upload/sales/validator_spec.rb
samyou-softwire: I think it'd be the most straightforward, using metadata as mentioned we can make it more manageable to monitor which tests need updating per year

I've split these out to be year specific and use metadata to mark as such

app/controllers/test_data_controller.rb
carolynbarker: yep discussed and happy with this

app/controllers/test_data_controller.rb
```diff
@@ -53,12 +54,13 @@ def create_setup_test_sales_log
...
   end
 
-  %w[2024 2025].each do |year|
+  %w[2025 2026].each do |year|
     define_method("create_#{year}_test_sales_bulk_upload") do
       return render_not_found unless FeatureToggle.create_test_logs_enabled?
 
       file = Tempfile.new("#{year}_test_sales_log.csv")
-      log = FactoryBot.create(:sales_log, :completed, assigned_to: current_user, value: 180_000, deposit: 150_000, county: "Somerset", saledate: generate_different_date_within_collection_year(Time.zone.local(year.to_i, 4, 1), end_date_override: Time.zone.now + 14.days))
+      end_date_override = year == current_collection_start_year ? Time.zone.now + 14.days : collection_start_date(Time.zone.local(year.to_i, 4, 1)) + 14.days
```
carolynbarker: this can be ` FeatureToggle.allow_future_form_use? ? nil : Time.zone.now + 14.days` to match the other one

carolynbarker (approved):
looks good to me!

PR #3137, opened 2026-01-06
CLDC-NONE: Add a manual runner pipeline that can redeploy review app code

samyou-softwire:
will give us a reliable method to redeploy code

this will push a docker image to ECR which can fix issues when the image expires

replaces rerunning the review app pipeline which is sometimes not possible

### DIFF
--- a/.github/workflows/manual_review_code_pipeline.yml
+++ b/.github/workflows/manual_review_code_pipeline.yml
@@ -1,12 +1,12 @@
 name: Manual review app code pipeline
 
 concurrency:
-  group: review-${{ inputs.key }}
+  group: review-${{ inputs.review_app_key }}
 
 on:
   workflow_dispatch:
     inputs:
-      review-app-key:
+      review_app_key:
         required: true
         type: string
         description: "The review app ID to deploy code for."
@@ -22,8 +22,8 @@ jobs:
     with:
       aws_account_id: 837698168072
       aws_role_prefix: core-dev
-      aws_task_prefix: core-review-${{ inputs.key }}
-      concurrency_tag: ${{ inputs.key }}
+      aws_task_prefix: core-review-${{ inputs.review_app_key }}
+      concurrency_tag: ${{ inputs.review_app_key }}
       environment: review
     permissions:
       id-token: write
--- a/.github/workflows/manual_review_code_pipeline.yml
+++ b/.github/workflows/manual_review_code_pipeline.yml
@@ -1,0 +1,29 @@
+name: Manual review app code pipeline
+
+concurrency:
+  group: review-${{ inputs.key }}
+
+on:
+  workflow_dispatch:
+    inputs:
+      review-app-key:
+        required: true
+        type: string
+        description: "The review app ID to deploy code for."
+
+defaults:
+  run:
+    shell: bash
+
+jobs:
+  code:
+    name: Deploy review app code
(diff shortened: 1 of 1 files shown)


.github/workflows/manual_review_code_pipeline.yml
```diff
@@ -0,0 +1,29 @@
+name: Manual review app code pipeline
+
+concurrency:
+  group: review-${{ inputs.key }}
+
+on:
+  workflow_dispatch:
+    inputs:
+      review-app-key:
```
carolynbarker: good name but I think I'd expect snake case `review_app_key`

.github/workflows/manual_review_code_pipeline.yml
```diff
@@ -0,0 +1,29 @@
+name: Manual review app code pipeline
+
+concurrency:
+  group: review-${{ inputs.key }}
```
carolynbarker: I think `inputs.key` needs to match the input variable name on line 9, i.e. `inputs.review-app-key` currently or `inputs.review_app_key` if you change to snake case

.github/workflows/manual_review_code_pipeline.yml
```diff
@@ -0,0 +1,29 @@
...
+
+jobs:
+  code:
+    name: Deploy review app code
+    uses: ./.github/workflows/aws_deploy.yml
+    with:
+      aws_account_id: 837698168072
+      aws_role_prefix: core-dev
+      aws_task_prefix: core-review-${{ inputs.key }}
+      concurrency_tag: ${{ inputs.key }}
```
carolynbarker: these also want to match the input variable name

.github/workflows/manual_review_code_pipeline.yml
carolynbarker: good idea to add the concurrency group btw!

.github/workflows/manual_review_code_pipeline.yml
```diff
@@ -0,0 +1,29 @@
...
+      environment: review
+    permissions:
+      id-token: write
```
carolynbarker, line 29: this job is the same as the one in review_pipeline.yml, so I wonder whether it's worth commonising it. I'm not sure whether it's worth it though.

I think there's a risk if we ever edited one review deploy job but not the other that we'd have different concurrency tags and so we could potentially end up running two conflicting deployments of the same review app if we ever ran one manually while an automated pipeline was running. But that seems pretty unlikely and also low impact.

.github/workflows/manual_review_code_pipeline.yml
samyou-softwire: sure

.github/workflows/manual_review_code_pipeline.yml
samyou-softwire: good spot, have fixed

.github/workflows/manual_review_code_pipeline.yml
samyou-softwire: see above

.github/workflows/manual_review_code_pipeline.yml
samyou-softwire: all this really does is call `./.github/workflows/aws_deploy.yml` so it's mostly commonised already. agree doubt we'd ever want to change the concurrency tag

carolynbarker (approved):

PR #3141, opened 2026-01-13
CLDC-4145: Don't ask partner for children

samyou-softwire:
closes [CLDC-4145](https://mhclgdigital.atlassian.net/browse/CLDC-4145)

mostly copies the flow for inferring ecstat

sets relat to X (No) if age under 16

clears relat if age changes, forcing user to re answer if age changes back above 16

marks question as 'derived' as appropriate

removes the age soft validation post 2026. it wouldnt be shown as we infer an answer thatll prevent it from being visible but may as well make it clear when it's not used

[CLDC-4145]: https://mhclgdigital.atlassian.net/browse/CLDC-4145?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

age question


app/models/form/lettings/questions/person_partner.rb
```diff
@@ -27,4 +27,8 @@ def question_number
 
     base_question_number + (4 * @person_index)
   end
+
+  def derived?(log)
+    form.start_year_2026_or_later? && log.is_partner_inferred?(@person_index)
+  end
```
samyou-softwire, line 33: one part of the code that's been vexing me is that there seems to be three ways to hide a question:
- remove it from the question list
- use depends on
- use derived? method

I'm not sure if there's a difference between these I should be aware of other than semantics

samyou-softwire:
need to fix failing tests (& maybe add some new ones); feel free to ignore this PR for now

app/models/derived_variables/lettings_log_variables.rb
```diff
@@ -246,8 +246,13 @@ def get_totelder
...
   def get_totchild
-    relationships = [relat2, relat3, relat4, relat5, relat6, relat7, relat8]
-    relationships.count("C")
+    if form.start_year_2025_or_later? then
+      ages = [age1, age2, age3, age4, age5, age6, age7, age8]
+      ages.count { |x| !x.nil? && x < 16 }
+    else
+      relationships = [relat2, relat3, relat4, relat5, relat6, relat7, relat8]
+      relationships.count("C")
+    end
```
samyou-softwire, line 255: interested in thoughts on this change

`Form::Lettings::Questions::PersonRelationship` is a question that defines the following options, stored in relat
```ruby
  def answer_options
    if form.start_year_2024_or_later?
      {
        "P" => { "value" => "Partner" },
        "C" => { "value" => "Child" },
        "X" => { "value" => "Other" },
        "R" => { "value" => "Person prefers not to say" },
      }
    else
      {
        "P" => { "value" => "Partner" },
        "C" => {
          "value" => "Child",
          "hint" => "Must be eligible for child benefit: under age 16 or under 20 if still in full-time education.",
        },
        "X" => { "value" => "Other" },
        "R" => { "value" => "Person prefers not to say" },
      }
    end
  end
```

however, past 2025 this question isn't used. in `household_characteristics.rb`
```ruby
  def relationship_question(person_index:)
    if form.start_year_2025_or_later?
      Form::Lettings::Pages::PersonLeadPartner.new(nil, nil, self, person_index:)
    else
      Form::Lettings::Pages::PersonRelationshipToLead.new(nil, nil, self, person_index:)
    end
  end
```

at a quick test on staging for these answers on a 2025 log

answers with tenant 2 as 8

I get the following export, which incorrectly shows 0 children

excel results 0 totchild

so I replaced it with a calculation that figures out total children correctly for post 2025 logs when the question was changed

worth letting rachel know we've fixed this or handling in a different way?

app/services/feature_toggle.rb
```diff
@@ -1,6 +1,6 @@
 class FeatureToggle
   def self.allow_future_form_use?
-    Rails.env.development? || Rails.env.review? || Rails.env.staging?
+    Rails.env.development? || Rails.env.review? || Rails.env.staging? || Rails.env.test?
```
samyou-softwire: I will probably raise a direct PR to main for this - tests should be fine to try 2026 forms

config/locales/forms/2026/lettings/soft_validations.en.yml
```diff
@@ -39,15 +39,6 @@ en:
...
 
-          partner_under_16_value_check:
-            page_header: ""
-            check_answer_label: "Partner under 16 confirmation"
-            check_answer_prompt: "Confirm partner’s age"
-            hint_text: ""
-            question_text: "Are you sure this is correct?"
-            title_text: "You told us this person is aged %{age} years and has 'Partner' relationship to the lead tenant."
-            informative_text: "Are you sure this is correct?"
-
```
samyou-softwire, line 50: this copy is not used in 2026

spec/models/form/lettings/subsections/household_characteristics_spec.rb
```diff
@@ -17,141 +17,6 @@
     expect(household_characteristics.section).to eq(section)
   end
 
-  context "with start year before 2024" do
```
samyou-softwire, line 20: pre 2024 logic checks can be deleted

spec/models/lettings_log_derived_fields_spec.rb
```diff
@@ -2,9 +2,12 @@
 require "shared/shared_examples_for_derived_fields"
 
 RSpec.describe LettingsLog, type: :model do
+  include CollectionTimeHelper
+
   let(:organisation) { build(:organisation, name: "derived fields org") }
   let(:user) { build(:user, organisation:) }
-  let(:log) { build(:lettings_log, :startdate_today, assigned_to: user) }
+  let(:startdate) { current_collection_start_date }
```
samyou-softwire, line 9: split out so I can override the startdate easier later

app/helpers/collection_time_helper.rb
```diff
@@ -4,6 +4,10 @@ def collection_start_year_for_date(date)
     date < window_end_date ? date.year - 1 : date.year
   end
 
+  def collection_start_date_for_year(year)
+    Time.zone.local(year, 4, 1)
+  end
+
```
samyou-softwire, line 10: I'm surprised this function doesn't already exist. can raise a PR to add this directly to main if it'd be useful to have available everywhere else?

app/services/feature_toggle.rb
samyou-softwire: never mind lots of tests rely on forms being unavailable. looks like I'll need to instead use timecop to work around this

app/models/form/lettings/questions/person_partner.rb
oscar-richardson-softwire: I've also been wondering about this. As far as I can tell, the only difference is semantic.

app/models/derived_variables/lettings_log_variables.rb
oscar-richardson-softwire: That looks good to me. Maybe we should @ Rachel on the ticket so this change is recorded somewhere and she's made aware of it (and we can always bring it up when we next have a call with her as well). Would you be able to @ her on a comment on the ticket explaining this change?

spec/models/lettings_log_derived_fields_spec.rb
```diff
@@ -111,10 +114,10 @@
     before do
       log.assign_attributes(
         relat2: "X",
-        relat3: "C",
+        relat3: "X",
         relat4: "X",
-        relat5: "C",
-        relat7: "C",
+        relat5: "X",
+        # relat7 is derived
```
oscar-richardson-softwire: I think we should leave in a test case for 2024 to check that the pre-2025 counting logic still works

spec/models/lettings_log_derived_fields_spec.rb
```diff
@@ -1213,23 +1232,128 @@
...
+      end
+
+      context "and it is 2025", metadata: { year: 25 } do
+        let(:startdate) { collection_start_date_for_year(2025) }
+
+        around do |example|
+          Timecop.freeze(collection_start_date_for_year(2025)) do
+            example.run
+          end
+        end
```
oscar-richardson-softwire, line 1332: Why is it necessary to freeze time in these tests (but not in the `"deriving household member fields"` tests above?). Isn't `let(:startdate) { collection_start_date_for_year(2025) }` sufficient for `form.start_year_2026_or_later?` to return `false`?

spec/models/form/lettings/subsections/household_characteristics_spec.rb
```diff
@@ -498,13 +363,11 @@
           person_2_known
           person_2_age
           person_2_lead_partner
-          relationship_2_partner_under_16_value_check
           relationship_2_multiple_partners_value_check
           no_females_pregnant_household_person_2_age_value_check
           females_in_soft_age_range_in_pregnant_household_person_2_age_value_check
           age_2_under_retirement_value_check
           age_2_over_retirement_value_check
-          age_2_partner_under_16_value_check
```
oscar-richardson-softwire, line 507: I know this isn't directly related to the work done in this ticket but this has got me questioning if there's a reason for these two different checks. As far as I can tell they're both identical, just with different `id`s?

app/helpers/collection_time_helper.rb
oscar-richardson-softwire: That would be good, thank you!

app/models/form/lettings/subsections/household_characteristics.rb
```diff
@@ -43,7 +43,7 @@ def pages
                                                                                              person_index: 2),
       Form::Lettings::Pages::PersonUnderRetirementValueCheck.new("age_2_under_retirement_value_check", nil, self, person_index: 2),
       Form::Lettings::Pages::PersonOverRetirementValueCheck.new("age_2_over_retirement_value_check", nil, self, person_index: 2),
-      (Form::Lettings::Pages::PartnerUnder16ValueCheck.new("age_2_partner_under_16_value_check", nil, self, person_index: 2) if form.start_year_2024_or_later?),
+      (Form::Lettings::Pages::PartnerUnder16ValueCheck.new("age_2_partner_under_16_value_check", nil, self, person_index: 2) if form.start_year_2024_or_later? && !form.start_year_2026_or_later?),
```
oscar-richardson-softwire, line 46: Are both `relationship_x_partner_under_16_value_check` and `age_2_partner_under_16_value_check` needed? They appear to be the same thing with different IDs

spec/fixtures/files/lettings_log_csv_export_codes_26.csv
```diff
@@ -1,3 +1,3 @@
...
-,in_progress,,s.port@jeemayle.com,s.port@jeemayle.com,false,2026-04-01T00:00:00+01:00,,2026-04-01T00:00:00+01:00,1,2026,,MHCLG,MHCLG,1,7,0,2026-04-01,2,2,,,2,HIJKLMN,ABCDEFG,1,0,2,6,2,,Address line 1,,London,,NW9 5LL,false,Barnet,E09000003,0,,,,,address line 1 as entered,address line 2 as entered,town or city as entered,county as entered,AB1 2CD,la as entered,7,1,3,2026-03-30,1,,1,2026-03-31,,,3,1,4,,2,4,,1,4,0,0,2,35,,F,0,2,36,0,0,32,P,,,M,6,1,-9,R,R,10,0,-9,R,R,10,,,,,,,,,,,,,,,,,,,,,1,4,1,2,1,0,1,0,0,0,0,0,0,1,0,0,1,0,0,0,0,0,0,0,2,7,4,,,6,1,0,TN23 6LZ,1,false,Ashford,E07000105,1,0,1,0,0,0,0,0,1,0,0,,,,0,0,1,268,,6,1,1,,,2,200.0,100.0,,50.0,25.0,40.0,20.0,35.0,17.5,325.0,162.5,,,,1,0,12.0,6.0,,,,,,,,,,,,,,,,,,,,
```
oscar-richardson-softwire, line 3: Why this change?

spec/fixtures/files/lettings_log_csv_export_labels_26.csv
```diff
@@ -1,3 +1,3 @@
...benefits, will they still need to pay for rent and charges?",Can you estimate the outstanding amount?,Estimated outstanding amount,Weekly total rent shortfall charge for tenant receiving housing benefit,What scheme does this letting belong to?,"From scheme code, we map to the scheme name",Does the scheme contain confidential information?,"What is this type of scheme? (Direct access hostel), Foyer, Housing for older people or Other supported housing",Is this scheme registered under the Care Standards Act 2000?,Which organisation owns the housing stock for this scheme?,What client group is this scheme intended for?,Does this scheme provide for another client group?,What is the other client group?,What support does this scheme provide?,Intended length of stay,Date scheme was created,Which location is this letting for?,What is the postcode for this location?,What is the name of this location?,How many units are at this location?,What is the most common type of unit at this location?,What are the mobility standards for the majority of the units in this location?,What is the local authority of this postcode?,When did the first property in this location become available under this scheme?
```
oscar-richardson-softwire, line 1: I'm struggling to parse what has changed in this file, could you point out what's changed please?

app/models/derived_variables/lettings_log_variables.rb
samyou-softwire: have done so

app/helpers/collection_time_helper.rb
samyou-softwire: see #3146

spec/models/lettings_log_derived_fields_spec.rb
samyou-softwire: have added some tests for 2024

spec/models/lettings_log_derived_fields_spec.rb
samyou-softwire: these tests use `create` to create a log. this means all of the validations will run as the log is written to the test db. without the timecop wrapper, the validations for 2026 will fail as the 2026 startdate is not valid (its not open).

the above tests use `build` which do not run any additional validations

spec/models/form/lettings/subsections/household_characteristics_spec.rb
samyou-softwire: believe the point is so no matter which order you answer the questions in the check will always appear next. it'll also show next to each relevant question in the CYA page

app/models/form/lettings/subsections/household_characteristics.rb
samyou-softwire: see above

spec/fixtures/files/lettings_log_csv_export_codes_26.csv
samyou-softwire: there is a row in the export for the soft validation confirmation "The following soft validation was confirmed: You said that [person X]'s relationship to lead tenant is partner, and that their age is [AGEX]. Are you sure this is correct?"

spec/fixtures/files/lettings_log_csv_export_labels_26.csv
samyou-softwire: see above. former contains a "The following soft validation was confirmed: You said that [person X]'s relationship to lead tenant is partner, and that their age is [AGEX]. Are you sure this is correct?" row. latter has this removed

oscar-richardson-softwire (approved):
LGTM

spec/models/form/lettings/pages/person_lead_partner_spec.rb
```diff
@@ -4,9 +4,14 @@
...
 
   let(:page_definition) { nil }
-  let(:subsection) { instance_double(Form::Subsection, form: instance_double(Form, start_date: Time.zone.local(2024, 4, 1), start_year_2025_or_later?: true)) }
+  let(:form) { instance_double(Form, start_date: Time.zone.local(2024, 4, 1)) }
+  let(:subsection) { instance_double(Form::Subsection, form:) }
   let(:person_index) { 2 }
 
+  before do
+    allow(form).to receive(:start_year_2026_or_later?).and_return(false)
+  end
```
oscar-richardson-softwire: I don't think this top level before do is needed

app/models/form/lettings/pages/person_lead_partner.rb
```diff
@@ -9,4 +9,21 @@ def initialize(id, hsh, subsection, person_index:)
   def questions
     @questions ||= [Form::Lettings::Questions::PersonPartner.new(nil, nil, self, person_index: @person_index)]
   end
+
+  def depends_on
```
oscar-richardson-softwire, line 12: I know it works as is anyway (including setting the inferred answer to the partner question to `No` if it is answered `Yes` before age is answered and then age is answered < 16) as the `@depends_on` in `initialize` gets overwritten, but I think we should remove the `@depends_on` in `initialize` for clarity.

spec/models/form/lettings/pages/person_lead_partner_spec.rb
```diff
@@ -24,10 +29,37 @@
       expect(page.id).to eq("person_2_lead_partner")
     end
 
-    it "has correct depends_on" do
-      expect(page.depends_on).to eq(
-        [{ "details_known_2" => 0 }],
-      )
+    context "with start year >= 2025", metadata: { year: 25 } do
```
oscar-richardson-softwire: Perhaps `with start year < 2026` would be more accurate to what we're testing here?

spec/models/form/lettings/pages/person_lead_partner_spec.rb
```diff
@@ -42,10 +74,37 @@
       expect(page.id).to eq("person_3_lead_partner")
     end
 
-    it "has correct depends_on" do
-      expect(page.depends_on).to eq(
-        [{ "details_known_3" => 0 }],
-      )
+    context "with start year >= 2025", metadata: { year: 25 } do
```
oscar-richardson-softwire: Perhaps `with start year < 2026` would be more accurate to what we're testing here?

spec/models/form/lettings/pages/person_lead_partner_spec.rb
samyou-softwire: sure

spec/models/form/lettings/pages/person_lead_partner_spec.rb
samyou-softwire: see above

app/models/form/lettings/pages/person_lead_partner.rb
samyou-softwire: ah good spot, missed this

spec/models/form/lettings/pages/person_lead_partner_spec.rb
samyou-softwire: agree

app/helpers/collection_time_helper.rb
oscar-richardson-softwire: Approved that PR

oscar-richardson-softwire (approved):
Approved, needs rebasing though

PR #3142, opened 2026-01-14
CLDC-4156: Update Care Standards Act values

samyou-softwire:
closes [CLDC-4156](https://mhclgdigital.atlassian.net/browse/CLDC-4156)

maps the values as requested on the ticket

includes a migration to map existing scheme values

[CLDC-4156]: https://mhclgdigital.atlassian.net/browse/CLDC-4156?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

new care standard options


app/models/scheme.rb
```diff
@@ -262,10 +261,10 @@ def check_details_attributes
     ]
   end
 
-  def care_acts_options_with_hints
+  def self.care_acts_options_with_hints
```
samyou-softwire: as far as I can tell this function was never called, the view pages would recalculate this

have replaced the view calls to now use this function

app/models/scheme.rb
```diff
@@ -262,10 +261,10 @@ def check_details_attributes
...
     hints = { "Yes – part registered as a care home": "A proportion of units are registered as being a care home." }
 
-    Scheme.registered_under_care_acts.keys.map { |key, _| OpenStruct.new(id: key, name: key.to_s.humanize, description: hints[key.to_sym]) }
```
samyou-softwire, line 268: humanize breaks the "CQC" part on the options. the strings are already human readable anyway

db/migrate/20260114160036_update_scheme_care_standards_values.rb
samyou-softwire, line 1: happy to hear whether this is the best way of doing the recode. a migration will apply automatically, though the other broad option is to include a rake task and manually run it after deploying. this will need a bit more work but does mean we can test the script

I went with a migration as the script is very simple so probably wouldn't need testing anyway

app/models/scheme.rb
carolynbarker: much nicer, thanks

db/schema.rb
```diff
@@ -84,7 +84,7 @@
     t.datetime "last_accessed"
     t.datetime "created_at", null: false
     t.datetime "updated_at", null: false
-    t.check_constraint "log_type::text = ANY (ARRAY['lettings'::character varying, 'sales'::character varying]::text[])", name: "log_type_check"
+    t.check_constraint "log_type::text = ANY (ARRAY['lettings'::character varying::text, 'sales'::character varying::text])", name: "log_type_check"
```
carolynbarker: any idea why this has changed? it looks like possibly different syntax for the same thing

db/migrate/20260114160036_update_scheme_care_standards_values.rb
carolynbarker: I'd prefer a rake task, as you say it would let us test it beforehand. Rails [advise against](https://guides.rubyonrails.org/active_record_migrations.html#execute-sql) modifying data in migrations, they recommend only using them to make changes to the schema.

carolynbarker: Also it's safer to avoid using models in migrations so that they can still be run if a model is renamed

carolynbarker: there's an existing rake task `correct_referral_value.rake` which I think will look similar

samyou-softwire: have reimplemented as a rake

db/schema.rb
samyou-softwire: not sure, this was an automated inclusion from adding the migration. it's been removed now

lib/tasks/correct_scheme_registered_under_care_value.rake
carolynbarker, line 1: nitpick but I wouldn't say we're correcting the scheme, just changing the options, so something like `update` might be a better verb

lib/tasks/correct_scheme_registered_under_care_value.rake
```diff
@@ -0,0 +1,4 @@
+desc "Alter registered under care act values for lettings logs in the database to 5 if they are 3 or 4, as these options are being deprecated"
+task correct_scheme_registered_under_care_value: :environment do
+  Scheme.where(registered_under_care_act: [3, 4]).update!(registered_under_care_act: 5)
```
carolynbarker: We could use `update_all` which doesn't change the updated_at fields (and I think it's also faster for bulk changes)

carolynbarker: I think this is the sort of change where we'd prefer not to set all the updated_at fields, but shout if you disagree

app/models/scheme.rb
```diff
@@ -262,10 +261,10 @@ def check_details_attributes
     ]
   end
 
-  def care_acts_options_with_hints
+  def self.care_acts_options_with_hints
     hints = { "Yes – part registered as a care home": "A proportion of units are registered as being a care home." }
```
oscar-richardson-softwire: Is this hint still needed (now that this option has been removed)?

app/models/scheme.rb
oscar-richardson-softwire: This is more of a question than a comment - I can't see where `registered_under_care_acts` is defined. Any ideas where this comes from?

lib/tasks/correct_scheme_registered_under_care_value.rake
```diff
@@ -0,0 +1,4 @@
+desc "Alter registered under care act values for lettings logs in the database to 5 if they are 3 or 4, as these options are being deprecated"
```
oscar-richardson-softwire: Isn't this for schemes, not lettings logs?

lib/tasks/correct_scheme_registered_under_care_value.rake
oscar-richardson-softwire: Agreed, maybe `update_scheme_registered_under_care_act_values.rake`

oscar-richardson-softwire (changes requested):
Looks like a question wording change from `"Is this scheme registered under the Care Standards Act 2000?"` to `"Is the scheme regulated by the Care Quality Commission (CQC)?"` is also part of this ticket.

app/models/scheme.rb
samyou-softwire: the short answer is that `schema.rb` defines it as a database column. ruby magic makes the database cols of this model available as variables in the model file without defining them

lib/tasks/correct_scheme_registered_under_care_value.rake
samyou-softwire: good spot, forgot to replace when I cloned

app/models/scheme.rb
samyou-softwire: ah yup, have remove the hint & dictionary

lib/tasks/correct_scheme_registered_under_care_value.rake
samyou-softwire: ah yeah I see other rakes do this too. have replaced

lib/tasks/correct_scheme_registered_under_care_value.rake
samyou-softwire: have updated the rake task name, file, and ticket post release steps

app/models/scheme.rb
```diff
@@ -251,7 +250,7 @@ def check_details_attributes
       { name: "Status", value: status, id: "status" },
       { name: "Confidential information", value: sensitive, id: "sensitive", edit: true },
       { name: "Type of scheme", value: scheme_type, id: "scheme_type", edit: true },
-      { name: "Registered under Care Standards Act 2000", value: registered_under_care_act, id: "registered_under_care_act", edit: true },
```
samyou-softwire, line 254: will check in with CORE on this change
scheme attribute list view

app/models/scheme.rb
```diff
@@ -262,9 +262,7 @@ def check_details_attributes
   end
 
   def self.care_acts_options_with_hints
```
oscar-richardson-softwire: We should probably remove `_with_hints` from the function name now that there are no hints

samyou-softwire: have fixed

oscar-richardson-softwire (commented):
Needs rebasing and an answer from CORE RE the wording question above, but then happy to approve

samyou-softwire:
have rebased and CORE has approved the wording @oscar-richardson-softwire

oscar-richardson-softwire (approved):
LGTM :) Happy for you to move the ticket to PO Review and assign to Joana. Remember to post a link to the review app in the ticket description

samyou-softwire:
was PO approved

PR #3143, opened 2026-01-15
CLDC-4114: Silence DPO warning if DSA is signed

samyou-softwire:
closes [CLDC-4114](https://mhclgdigital.atlassian.net/browse/CLDC-4114)

ftoad DSA = data sharing agreement, DPO = data protection officer

initially we agreed to change the warning but looking into the logic I think this is incorrect. all we want to do is hide the 'No DPO' warning if a DSA has been signed.

the flow now works like:

- On initial creation of the org, there is no DPO. Show the ‘There is no DPO’ wording
no DPO

- On assigning a DPO, they need to sign the DSA. Show a ‘DSA needs signing’ warning
DSA needs signing

After signing the DSA then the warning hides.

restructure redundant logic in the file. the only logic change is the addition of `&& !dsa_signed?` to the no DPO check

[CLDC-4114]: https://mhclgdigital.atlassian.net/browse/CLDC-4114?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


katlan-sw (approved):

app/components/data_protection_confirmation_banner_component.rb
```diff
@@ -51,34 +51,42 @@ def data_protection_officers_text
...
       data_sharing_agreement_organisation_path(org_or_user_org)
     end
   end
 
-  def dpo_required?
-    org_or_user_org.data_protection_officers.empty?
+  def show_no_dpo_message?
+    # it is fine if an org has a DSA and the DPO has moved on
+    # CORE staff do this sometimes as a single DPO covers multiple 'orgs' that exist as branches of the same real world org
+    # so, they move the DPO to all the mini orgs and have the sign each DSA
```
oscar-richardson-softwire:
```suggestion
    # so, they move the DPO to all the mini orgs and have them sign each DSA
```

spec/components/data_protection_confirmation_banner_component_spec.rb
```diff
@@ -23,13 +23,18 @@
       organisation.users.where(is_dpo: true).destroy_all
```
oscar-richardson-softwire, line 23: I think these tests could generally do with a tidy up - e.g., there are two separate `"when org does not have a DPO"` context blocks, one with the signed DSA case and one with the unsigned DSA case, there's an `"when org does not have a DPO"` context block inside a `"when org has a DPO"` block etc.

samyou-softwire: have removed a lot of the redundant code

oscar-richardson-softwire: This is much cleaner thanks, but I think we're now missing the `"when org does not have a DPO"` -> `"when org has a signed data sharing agreemeent"` case. Could you add this back in please?

spec/components/data_protection_confirmation_banner_component_spec.rb
```diff
@@ -81,7 +86,7 @@
       end
     end
 
-    context "when org has a signed data sharing agremeent" do
+    context "when org has a signed data sharing agreemeent" do
```
oscar-richardson-softwire:
```suggestion
    context "when org has a signed data sharing agreement" do
```

spec/components/data_protection_confirmation_banner_component_spec.rb
samyou-softwire: good spot, have added this now

oscar-richardson-softwire (approved):
LGTM

PR #3144, opened 2026-01-16
CLDC-4177: New Sex At Birth question for sales logs

katlan-sw:
Closes [CLDC-4177](https://mhclgdigital.atlassian.net/browse/CLDC-4177), with some disclaimers (see "left to do").

## Summary

Adds a new question before each gender question that asks about what sex was assigned to the person when they were born.

Applies to Sales logs only, not to Lettings logs.

This means that there is a new question for:
- Buyer 1.
- The joint buyer, if they exist.
- Any additional household members, up to 6 people total.

## Changes made

- Added new question for Buyer 1 for sex assigned at birth.
- Added new page for Buyer 1 for sex assigned at birth.
- Added new question for Buyer 2 for sex assigned at birth.
- Added new page for Buyer 2 for sex assigned at birth.
- Added new question for Person 2 - 6 for sex assigned at birth.
- Added new page for Person 2 - 6 for sex assigned at birth.
- Added new properties to the database.
- Added new pages to the "household characteristics" subsection of the sales form.
- Added appropriate unit tests.

## Testing

- Created a new sales log for a household with a joint buyer and 4 additional people.
  - Saw a new question for Buyer 1 asking about their sex at birth. ✅
  - Saw a new question for Buyer 2 asking about their sex at birth. ✅ image
  - Saw a new question for People 3 - 6 asking about their sex at birth. ✅
  - Was able to save the form after each new question. ✅
  - Could see the new question appear on the summary page for each person. ✅ image
  - Exporting the log as CSV includes the answers for the new question. ✅ \* See "left to do".
- Created a new sales log for a household with a single buyer and 1 additional person.
  - Saw a question for Person 2, rather than for Buyer 2. ✅

## Left to do

- Although the new questions appear in the CSV, I have not managed to get it to output the answers in plain english. For parity with the existing gender question, it might be nice to change this. Could potentially create a new ticket for this? image
- The new questions are left with question number 0, as we will be sorting out question numbers all at once later.

[CLDC-4177]: https://mhclgdigital.atlassian.net/browse/CLDC-4177?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/models/form/sales/pages/person_sex_registered_at_birth.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+
+class Form::Sales::Pages::PersonSexRegisteredAtBirth < ::Form::Page
+  def initialize(id, hsh, subsection, person_index:)
+    super(id, hsh, subsection)
+    @copy_key = "sales.household_characteristics.sex2.person" if person_index == 2
```
oscar-richardson-softwire: What does `@copy_key` do in a `Page` (rather than a `Question`) and why is this `sales.household_characteristics.sex2.person` rather than `sales.household_characteristics.sexRAB2.person`?

app/models/form/sales/pages/sex_registered_at_birth2.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+
+class Form::Sales::Pages::SexRegisteredAtBirth2 < ::Form::Page
+  def initialize(id, hsh, subsection)
+    super
+    @id = "buyer_2_sex_registered_at_birth"
+    @copy_key = "sales.household_characteristics.sex2.buyer"
```
oscar-richardson-softwire: What does `@copy_key` do in a `Page` (rather than a `Question`) and why is this `sales.household_characteristics.sex2.buyer` rather than `sales.household_characteristics.sexRAB2.buyer`?

app/services/exports/sales_log_export_service.rb
```diff
@@ -151,7 +151,8 @@ def apply_cds_transformation(sales_log, _export_mode)
     end
 
     def is_omitted_field?(field_name, _sales_log)
-      !EXPORT_FIELDS.include?(field_name)
+      !EXPORT_FIELDS.include?(field_name) ||
+        (!_sales_log.form.start_year_2026_or_later? && POST_2026_EXPORT_FIELDS.include?(field_name))
```
oscar-richardson-softwire: If the start year is 2026, doesn't this function return `true` for the `SEXRAB#{index}` fields, as EXPORT_FIELDS doesn't include them?

config/locales/forms/2026/sales/household_characteristics.en.yml
```diff
@@ -16,6 +16,13 @@ en:
               hint_text: ""
               question_text: "Age"
 
+          sexRAB1:
+            page_header: ""
+            check_answer_label: "Buyer 1’s sex registered at birth"
+            check_answer_prompt: ""
+            hint_text: "This is the sex that was registered at birth. The next question will ask about the buyer's gender identity."
+            question_text: "What was buyer 1's sex at birth?"
```
oscar-richardson-softwire, line 24: This is likely a non-issue (and I actually think it is an improvement) but it looks like the question text itself in the ticket omits `at birth`. Not sure how much this matters for e.g., matching the paper form. Same for the other question text in this file.

db/schema.rb
```diff
@@ -373,8 +373,8 @@
     t.integer "partner_under_16_value_check"
     t.integer "multiple_partners_value_check"
     t.bigint "created_by_id"
-    t.integer "referral_type"
     t.boolean "manual_address_entry_selected", default: false
+    t.integer "referral_type"
```
oscar-richardson-softwire: Doesn't seem like it matters either way but why this change?

db/schema.rb
```diff
@@ -504,7 +504,7 @@
     t.date "discarded_at"
     t.datetime "created_at", null: false
     t.datetime "updated_at", null: false
-    t.index ["organisation_id", "startdate"], name: "index_org_name_changes_on_org_id_and_startdate", unique: true
+    t.index ["organisation_id", "startdate", "discarded_at"], name: "index_org_name_changes_on_org_id_startdate_discarded_at", unique: true
```
oscar-richardson-softwire: Why this change?

db/schema.rb
```diff
@@ -894,6 +900,7 @@
   add_foreign_key "local_authority_links", "local_authorities"
   add_foreign_key "local_authority_links", "local_authorities", column: "linked_local_authority_id"
   add_foreign_key "locations", "schemes"
+  add_foreign_key "organisation_name_changes", "organisations"
```
oscar-richardson-softwire, line 904: Why this change?

db/schema.rb
katlan-sw: Please see [this PR](https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/3148). It seems to be a leftover from a previous migration that was only partially run.

oscar-richardson-softwire (commented):
Great work! :) Just a few minor comments

app/models/form/sales/pages/person_sex_registered_at_birth.rb
katlan-sw: To be honest, I don't know exactly what it does, apart from it being related to differentiating between person 2 being a joint buyer or now. I cribbed this from the source files for 'person gender' - hence why it's sex instead of sexRAB (an oversight).

I will have investigate exactly where it's used in the morning.

app/services/exports/sales_log_export_service.rb
katlan-sw: It's a 2026 log and therefore we want to include SexRAB, thus "is omitted field" should return false.

- EXPORT_FIELDS does not include SexRAB
- POST_2026_EXPORT_FIELDS does include SexRAB

Thus, we return
```
!false || (!true && true)
= true || (false && true)
= true || false
= true
```

You are correct.

So, the logic should actually be:

A field is omitted iff:
- It is not in the base export list
- AND it's in the 2026 export list and it's not 2026

katlan-sw: But the tests didn't pass and my focus is flagging from its already precarious perch. I will have to tackle it afresh tomorrow.

oscar-richardson-softwire: > A field is omitted iff:
> 
> It is not in the base export list
> AND it's in the 2026 export list and it's not 2026

I don't think this is quite right. This would mean that a field that wasn't in the base export list or the 2026 export list would not be omitted, regardless of what year it is.

I think the logic we want is
`!EXPORT_FIELDS.include?(field_name) && (!_sales_log.form.start_year_2026_or_later? || !POST_2026_EXPORT_FIELDS.include?(field_name))`

Then not being in the base export list is always a requirement for omission, but not being in the 2026 export list is only a requirement for omission if the year is 2026 or later (deffo sanity check this though as my brain hurts from this).

db/schema.rb
oscar-richardson-softwire: Nice, approved that PR

app/services/exports/sales_log_export_service.rb
katlan-sw: I've done a refactor to make this easier to parse.

config/locales/forms/2026/sales/household_characteristics.en.yml
katlan-sw: Ah, I know from time in LGBT spaces that this is the more inclusive language, so I just added it on autopilot.

app/services/bulk_upload/sales/year2026/row_parser.rb
```diff
@@ -135,6 +135,13 @@ class BulkUpload::Sales::Year2026::RowParser
     field_119: "Does this include any extra borrowing?",
     field_120: "How much was the cash deposit paid on the property?",
     field_121: "What are the total monthly leasehold charges for the property?",
+
+    field_122: "Buyer 1's sex, as registered at birth",
+    field_123: "Buyer 2's sex, as registered at birth",
```
oscar-richardson-softwire: Not sure how best to indicate that this field may be buyer 2 or person 2 in BU - `"Buyer/person 2's sex, as registered at birth"` maybe?

app/models/form/sales/pages/person_sex_registered_at_birth.rb
oscar-richardson-softwire: I see you've corrected this to `sexrab` in both cases so I'm happy to resolve this thread - out of interest, did you figure out what this is used for?

app/services/exports/sales_log_export_service.rb
oscar-richardson-softwire: Really nice refactor. Now wondering whether we should add a quick test case for this in `sales_log_export_service_spec.rb` (although probably not worth it if this is going to be something non-trivial to test)

app/services/bulk_upload/sales/year2026/row_parser.rb
katlan-sw: Yeah, makes sense.

app/services/exports/sales_log_export_service.rb
katlan-sw: Not trivial to isolate the new logic. But there is a test that fails if in 2026 the new fields aren't included, so I think it's fine.

oscar-richardson-softwire (approved):
LGTM :) Happy for you to move the ticket to PO Review and assign to Joana. Remember to post a link to the review app in the ticket description

db/schema.rb
```diff
@@ -787,6 +790,12 @@
     t.datetime "lasttransaction"
     t.datetime "initialpurchase"
     t.boolean "manual_address_entry_selected", default: false
+    t.string "sexrab1"
```
samyou-softwire, line 791: this should be reflected in `spec/factories/sales_log.rb` in the `:completed` trait. fine to add a value regardless of the year (if its not asked it'll be nulled automatically).

if we don't have this creating `:completed` logs for 2026 won't work

db/schema.rb
```diff
@@ -375,6 +375,9 @@
     t.bigint "created_by_id"
     t.boolean "manual_address_entry_selected", default: false
     t.integer "referral_type"
+    t.integer "referral_register"
+    t.integer "referral_noms"
+    t.integer "referral_org"
```
samyou-softwire: should this be here? https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/3150/changes#diff-cbf8b25d6d853acf58fa9057841f407d29ffe90649c75cf9e3f51b2d6e3aa1d3R378 includes this already

samyou-softwire (changes requested):

db/schema.rb
oscar-richardson-softwire: Good spot, have removed

db/schema.rb
oscar-richardson-softwire: Added

samyou-softwire (approved):
LGTM assuming tests pass

db/schema.rb
```diff
@@ -375,9 +375,7 @@
     t.bigint "created_by_id"
     t.boolean "manual_address_entry_selected", default: false
     t.integer "referral_type"
-    t.integer "referral_register"
-    t.integer "referral_noms"
-    t.integer "referral_org"
+    t.integer "working_situation_illness_check"
```
oscar-richardson-softwire, line 378: I get why the `referral_` fields were removed (they should've have been added to `main` yet) but do we know why `working_situation_illness_check` was added? Was that accidentally removed on `main`?

natdeanlewissoftwire: `working_situation_illness_check` wasn't yet on main, so it's included here because it was added on main, and that's been merged in, but shouldn't be showing up as added in the full diff:

Image

I'm guessing you saw this in a diff for a specific commit? I've been fiddling around with db schema to get it to sync correctly (so there may have been some incorrect states in the interim that showed the wrong fields) but latest state should be correct.

oscar-richardson-softwire: Ah yep, was looking at a specific commit range 🤦 All good in that case! 😄

oscar-richardson-softwire (approved):
LGTM

PR #3145, opened 2026-01-16
CLDC-4159: Address / UPRN for supported housing - implementation

oscar-richardson-softwire:
Closes https://mhclgdigital.atlassian.net/browse/CLDC-4159.

### DIFF
--- a/app/models/lettings_log.rb
+++ b/app/models/lettings_log.rb
@@ -190,7 +190,8 @@ class LettingsLog < Log
     location.linked_local_authorities.active(form.start_date).first&.code || location.location_code
   end
 
-  def postcode_full # TODO: CLDC-4119: Beware! This method may cause issues when testing supported housing log duplicate detection after postcode is added, as it can return `location.postcode` instead of the actual `postcode_full` stored on the log record (`super`). If this happens, investigate why it isn't returning `super`, as it should when `form.start_year_2026_or_later? && super`.
+  # TODO: CLDC-4119: Beware! This method may cause issues when testing supported housing log duplicate detection after postcode is added, as it can return `location.postcode` instead of the actual `postcode_full` stored on the log record (`super`). If this happens, investigate why it isn't returning `super`, as it should when `form.start_year_2026_or_later? && super`.
+  def postcode_full
     return super unless location
     return super if form.start_year_2026_or_later? && super
 
--- a/app/models/lettings_log.rb
+++ b/app/models/lettings_log.rb
@@ -184,7 +184,7 @@ class LettingsLog < Log
   end
 
   def la
-    if (!form.start_year_2026_or_later? && location) || (form.start_year_2026_or_later? && !super)
+    if location && (!form.start_year_2026_or_later? || (form.start_year_2026_or_later? && !super))
       location.linked_local_authorities.active(form.start_date).first&.code || location.location_code
     else
       super
@@ -192,7 +192,7 @@ class LettingsLog < Log
   end
 
   def postcode_full
-    if (!form.start_year_2026_or_later? && location) || (form.start_year_2026_or_later? && !super)
+    if location && (!form.start_year_2026_or_later? || (form.start_year_2026_or_later? && !super))
       location.postcode
     else
       super
--- a/app/models/lettings_log.rb
+++ b/app/models/lettings_log.rb
@@ -184,19 +184,17 @@ class LettingsLog < Log
   end
 
   def la
-    if location && (!form.start_year_2026_or_later? || (form.start_year_2026_or_later? && !super))
-      location.linked_local_authorities.active(form.start_date).first&.code || location.location_code
-    else
-      super
-    end
+    return super unless location
+    return super if form.start_year_2026_or_later? && super
+
+    location.linked_local_authorities.active(form.start_date).first&.code || location.location_code
   end
 
   def postcode_full
-    if location && (!form.start_year_2026_or_later? || (form.start_year_2026_or_later? && !super))
-      location.postcode
-    else
-      super
-    end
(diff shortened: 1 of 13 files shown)


app/models/lettings_log.rb
```diff
@@ -184,15 +184,15 @@ def self.editable_fields
   end
 
   def la
-    if location
+    if location && (!form.start_year_2026_or_later? || (form.start_year_2026_or_later? && !super))
```
samyou-softwire: I think this could do with a comment. I think I get the intent
- pre 26 just check if location has been set
- post 26 use super `la` if its set

or maybe theres a clearer way to set this out. ruby is a big fan of guard clauses `return x if y` so that could be a more natural way to express the guards

app/models/validations/property_validations.rb
samyou-softwire, line 1: in the below errors itll still add errors to location if the entered address is wrong
```ruby
elsif record.is_supported_housing?
  record.errors.add :location_id, I18n.t("validations.lettings.property.location_id.not_in_england")
  record.errors.add :scheme_id, I18n.t("validations.lettings.property.scheme_id.not_in_england")
  record.errors.add :startdate, I18n.t("validations.lettings.property.startdate.location_not_in_england")
end
```

have we covered all usages of `record.is_general_needs?`/`record.is_supported_housing?`

app/models/lettings_log.rb
carolynbarker: we could use a guard clause like `return super if form.start_year_2026_or_later? && super`

carolynbarker: or split it out another way. agree it feels a bit wordy at the moment, could be clearer to break up across multiple lines

app/models/validations/property_validations.rb
```diff
@@ -50,6 +50,16 @@ def validate_property_postcode(record)
...
+
+    return unless record.form.start_year_2026_or_later?
+
+    location_postcode = record.location&.postcode
+    return unless location_postcode
+
+    if postcode != location_postcode
+      record.errors.add :uprn, I18n.t("validations.lettings.property.uprn.postcode_does_not_match_scheme_location_postcode")
+      record.errors.add :postcode_full, I18n.t("validations.lettings.property.postcode_full.does_not_match_scheme_location_postcode")
+    end
```
carolynbarker, line 68: we could pull this out into a separate method for validate_property_location_postcodes_match or something

app/models/validations/property_validations.rb
```diff
@@ -50,6 +50,16 @@ def validate_property_postcode(record)
...
       record.errors.add :postcode_full, :wrong_format, message: error_message
     end
+
+    return unless record.form.start_year_2026_or_later?
+
+    location_postcode = record.location&.postcode
+    return unless location_postcode
+
+    if postcode != location_postcode
+      record.errors.add :uprn, I18n.t("validations.lettings.property.uprn.postcode_does_not_match_scheme_location_postcode")
```
carolynbarker, line 66: do we also want to add the error on `location`?

spec/models/lettings_log_derived_fields_spec.rb
```diff
@@ -1446,4 +1446,146 @@
       expect { log.set_derived_fields! }.to change(log, :beds).to nil
     end
   end
+
+  describe "resetting address fields and LA" do
+    # log.read_attribute is used in these tests to read the underlying attribute (which is what
+    # is being reset) rather than the getter method, which may override the underlying attribute.
```
carolynbarker, line 1452: this makes sense, but I'm wondering if we also want to test the getter method behaviour somewhere as we'll be relying on its behaviour

app/models/lettings_log.rb
oscar-richardson-softwire: Lovely suggestions, I've rewritten both the `la` and `postcode_full` methods as suggested.

app/models/validations/property_validations.rb
oscar-richardson-softwire: Good shout, added

app/models/validations/property_validations.rb
oscar-richardson-softwire: Done

spec/models/lettings_log_derived_fields_spec.rb
oscar-richardson-softwire: Added tests in `spec/models/lettings_log_spec.rb`.

app/models/validations/property_validations.rb
oscar-richardson-softwire: ~~I've added `&& !record.read_attribute(:la)` to this condition. The first time it'll try to validate the LA is in England/is active on a supported housing log is when the location is entered in the setup questions (and `&& !record.read_attribute(:la)` - i.e., has the LA question in the address questions not been answered - will return true). When it tries to validate it again (when the address is entered), this will return false and prevent these errors from being added, but this case will be covered by the postcode not matching the location anyway.~~

I realised this won't work for BU (if someone enters a scheme location in Scotland, say, and also a matching LA in Scotland in the LA question, this validation won't kick in). I've updated the logic so that all of the fields that receive errors in a general needs log _plus_ the location-related fields receive errors in a supported housing log (in the web form you'd realistically only see these errors when you first try to enter a location outside of England as if this was done at the point of entering address this would be caught by the non-matching postcode validation as by this point you must have entered a scheme location in England).

oscar-richardson-softwire: > have we covered all usages of record.is_general_needs?/record.is_supported_housing?

I've doubled checked this as well

samyou-softwire (approved):
LGTM, though needs a lint fix

PR #3147, opened 2026-01-19
CLDC-4105: Add new working situation and long-term illness validation

samyou-softwire:
closes [CLDC-4105](https://mhclgdigital.atlassian.net/browse/CLDC-4105)

adds soft validation for working situation

soft validation in main flow

bulk upload warning

cya in household needs

cya in household characteristics

</details>

[CLDC-4105]: https://mhclgdigital.atlassian.net/browse/CLDC-4105?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/models/form/lettings/subsections/household_characteristics.rb
```diff
@@ -31,6 +31,7 @@ def pages
       Form::Lettings::Pages::LeadTenantWorkingSituation.new(nil, nil, self),
       Form::Lettings::Pages::LeadTenantUnderRetirementValueCheck.new("working_situation_lead_tenant_under_retirement_value_check", nil, self),
       Form::Lettings::Pages::LeadTenantOverRetirementValueCheck.new("working_situation_lead_tenant_over_retirement_value_check", nil, self),
+      (Form::Lettings::Pages::WorkingSituationIllnessCheckLead.new("working_situation_lead_tenant_long_term_illness_check", nil, self) if form.start_year_2026_or_later?),
```
samyou-softwire, line 34: note; initially I wasn't sure why each question used a different ID. turns out CORE routing gets very confused if there are multiple pages with the same ID

app/models/form/lettings/pages/working_situation_illness_check.rb
samyou-softwire, line 1: the layout here does seem to follow the established consensus in CORE
- this base file is the most basic version of the question used once after the illness question
- the `_lead` file lets an ID be input so it can differ from persons
- the `_person` file lets a person ID be input
though maybe this it too much machinery

db/schema.rb
samyou-softwire, line 1: always seems like a couple extra migration lines are added when I generate migrations. can revert the unexpected if we don't want it

spec/services/bulk_upload/lettings/year2026/row_parser_spec.rb
```diff
@@ -197,7 +197,7 @@
             field_82: "1",
             field_83: "0",
 
-            field_85: "2",
+            field_85: "3",
```
samyou-softwire, line 200: needed since the previous BU questions would've triggered this soft validation. this template should be a 'complete' log with no errors

spec/fixtures/variable_definitions/lettings_download_26_27.csv
samyou-softwire, line 1: needed so I could add a sensible new line to the export labels csv

spec/fixtures/files/lettings_log_csv_export_codes_26.csv
samyou-softwire, line 1: new col is `working_situation_illness_check`

app/models/form/lettings/pages/working_situation_illness_check.rb
carolynbarker: it does seem a bit repetitive, but I think it's probably more important to be clear. We could most easily combine the base and _lead files, but then I'd worry it could surprise people that the persons page wasn't used for the lead.

carolynbarker: it doesn't quite feel right to have three versions of the same question though! 🤔

app/models/validations/soft_validations.rb
```diff
@@ -282,4 +286,21 @@ def partner_under_16?(person_num)
 
     age < 16 && relationship == "P"
   end
+
+  def at_least_one_person_working_situation_is_illness?
+    return unless hhmemb
+
+    person_count = hhmemb || 8
```
carolynbarker: do we need the `|| 8`? I think L291 should guarantee that hhmemb is truthy

app/models/validations/soft_validations.rb
```diff
@@ -282,4 +286,21 @@ def partner_under_16?(person_num)
...
+
+    (1..person_count).any? do |n|
+      ecstat = public_send("ecstat#{n}")
+      ecstat == 8
```
carolynbarker: I think I'd find this a bit clearer on one line rather than introducing a variable for ecstat

db/schema.rb
carolynbarker: @katlan-sw and I found the same the other day - we think it is changes from the last previous migration from April that weren't included in the schema. Happy to include them here

carolynbarker (approved):

db/schema.rb
samyou-softwire: have rebased to include these as they were merged in

app/models/validations/soft_validations.rb
samyou-softwire: agree, have removed

carolynbarker (approved):
LGTM, though I think the pipeline will need rerunning

samyou-softwire:
rebase onto main

oscar-richardson-softwire (approved):
LGTM

PR #3150, opened 2026-01-26
CLDC-4151: Split Referral into three questions

samyou-softwire:
closes [CLDC-4151](https://mhclgdigital.atlassian.net/browse/CLDC-4151)

adds `referral_register`, `referral_noms`, `referral_org` to lettings logs

adds questions and pages for each. answers are left as A and B for now

remove the referral value check for now

note on the old referral questions when they were deprecated. do we want to do more?

update tests

skip a couple tests we'll cover in [CLDC-4191](https://mhclgdigital.atlassian.net/browse/CLDC-4191)

temporary Q84
temporary CYA
</details>

[CLDC-4151]: https://mhclgdigital.atlassian.net/browse/CLDC-4151?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[CLDC-4191]: https://mhclgdigital.atlassian.net/browse/CLDC-4191?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/helpers/bulk_upload/lettings_log_to_csv.rb
```diff
@@ -93,12 +93,162 @@ def default_2025_field_numbers
 
   def default_2026_field_numbers
     # TODO: CLDC-4162 Replace with actual field numbers when 2026 format is known
-    (1..129).to_a
+    (1..132).to_a
   end
 
   def to_2026_row
     # TODO: CLDC-4162: Implement when 2026 format is known
-    to_2025_row
```
samyou-softwire: a copy of to_2025_row with `referral_x` fields included

app/models/form/lettings/pages/referral_direct.rb
samyou-softwire, line 1: interested if we think this is the appropriate way to document this. could it be a model attribute and also maybe validated accordingly? or fine as just a comment?

app/models/form/lettings/subsections/household_situation.rb
```diff
@@ -22,26 +22,33 @@ def pages
       Form::Lettings::Pages::ReasonablePreferenceReason.new(nil, nil, self),
       Form::Lettings::Pages::AllocationSystem.new("allocation_system", nil, self),
       referral_questions,
-      Form::Lettings::Pages::ReferralValueCheck.new(nil, nil, self),
```
samyou-softwire, line 25: TBC whether we want this check for 2026. I've asked on the ticket

app/services/bulk_upload/lettings/year2026/row_parser.rb
```diff
@@ -120,7 +120,7 @@ class BulkUpload::Lettings::Year2026::RowParser
     field_113: "Was the letting made under the Common Allocation Policy (CAP)?",
     field_114: "Was the letting made under the Common Housing Register (CHR)?",
     field_115: "Was the letting made under the Accessible Register?",
-    field_116: "What was the source of referral for this letting?",
+    field_116: "What was the source of referral for this letting? - LA properties",
```
samyou-softwire, line 123: these changes are based on the new BU form layout
bulk upload referral source qs

four fields in total, one used for LAs the other for PRPs

this PR includes code to add the fields and assign them to logs, but no validations on them yet

spec/models/form/lettings/subsections/household_situation_spec.rb
```diff
@@ -10,16 +10,18 @@
...
+    allow(form).to receive(:start_year_2024_or_later?).and_return(false)
+    allow(form).to receive(:start_year_2025_or_later?).and_return(false)
+    allow(form).to receive(:start_year_2026_or_later?).and_return(false)
   end
 
   it "has correct section" do
     expect(household_situation.section).to eq(section)
   end
 
-  context "with form year before 2024" do
```
samyou-softwire, line 19: before 2024 tests removed

spec/services/bulk_upload/lettings/validator_spec.rb
```diff
@@ -190,8 +190,8 @@
           expect(error.tenant_code).to eql(log.tenancycode)
           expect(error.property_ref).to eql(log.propcode)
           expect(error.row).to eql("2")
-          expect(error.cell).to eql("CX2")
-          expect(error.col).to eql("CX")
+          expect(error.cell).to eql("DA2")
```
samyou-softwire: theres a lot of these changes to excel columns, though the tests randomise the column order so this seems naturally brittle. are these actually useful tests?

app/helpers/bulk_upload/lettings_log_to_csv.rb
```diff
@@ -93,12 +93,162 @@ def default_2025_field_numbers
...
+      log.period,
+      log.brent,
+      log.scharge,
+      log.pscharge,
+      log.supcharg,
+      log.hbrentshortfall,
+      log.tshortfall,
+      owning_organisation&.prp? ? log.referral_register : nil,
+      log.referral_noms,
+      log.referral_org, # 132
```
oscar-richardson-softwire: (May delete this comment later as still wrapping my head around the requirements): Don't we want to make these conditional as well?

app/helpers/bulk_upload/lettings_log_to_csv.rb
```diff
@@ -93,12 +93,162 @@ def default_2025_field_numbers
...
 
   def to_2026_row
     # TODO: CLDC-4162: Implement when 2026 format is known
-    to_2025_row
+    owning_organisation = if overrides[:organisation_id]
+                          then Organisation.find(overrides[:organisation_id])
+                          else
+                            log.owning_organisation
+                          end
+    [
```
oscar-richardson-softwire: It looks like this logic is slightly difference from 2025 (namely if `!if overrides[:organisation_id]` then we use `log.owning_organisation.id`, whereas 2025 uses `log.owning_organisation.old_visible_id`). Just checking if this is intentional.

app/models/form/lettings/pages/referral_direct.rb
oscar-richardson-softwire: Given we have a lot of work to get through, I'm inclined to say let's not add to our dev burden right now and that this is fine as just comments for the time being. If we do decide we want this as a model attribute down the line, we can add a tech debt ticket for this post-April.

app/models/form/lettings/pages/referral_general_needs.rb
```diff
@@ -1,3 +1,4 @@
+# removed in 2025
```
oscar-richardson-softwire: Just double checking this is right (2025 not 2026?).

app/services/bulk_upload/lettings/year2026/csv_parser.rb
```diff
@@ -4,8 +4,8 @@ class BulkUpload::Lettings::Year2026::CsvParser
   include CollectionTimeHelper
 
   # TODO: CLDC-4162: Update when 2026 format is known
-  FIELDS = 129
-  MAX_COLUMNS = 130
+  FIELDS = 132
+  MAX_COLUMNS = 133
```
oscar-richardson-softwire: Why is `MAX_COLUMNS` one more than `FIELDS`?

app/services/bulk_upload/lettings/year2026/row_parser.rb
```diff
@@ -1684,21 +1692,29 @@ def rp_dontknow_conflict?
...
+    else
+      field_130
     end
+  end
 
-    0
+  def referral_noms
+    return unless owning_organisation
+
+    if owning_organisation.la?
```
oscar-richardson-softwire:
```suggestion
    if owning_organisation.prp?
```

app/services/bulk_upload/lettings/year2026/row_parser.rb
```diff
@@ -1684,21 +1692,29 @@ def rp_dontknow_conflict?
...
+      field_131
+    end
+  end
+
+  def referral_org
+    return unless owning_organisation
+
+    if owning_organisation.la?
```
oscar-richardson-softwire:
```suggestion
    if owning_organisation.prp?
```

spec/models/form/lettings/subsections/household_situation_spec.rb
```diff
@@ -69,20 +72,23 @@
...
           allocation_system
-          referral
+          referral_type
+          referral_direct
+          referral_la
           referral_prp
-          referral_supported_housing
-          referral_supported_housing_prp
+          referral_hsc
+          referral_justice
```
oscar-richardson-softwire, line 80: Why has this changed? In the code it looked like nothing had changed for 2025.

spec/services/bulk_upload/lettings/validator_spec.rb
oscar-richardson-softwire: Ah, I think I get it - when we add N new fields, the columns in these tests will shift an unpredictable amount (by between 0 and N depending on how many of the new fields end up to the left of the field in question) - is this right?

I guess we have these tests because we want to check that error handling still works in BU CSVs that don't match the column order, so there's probably still value to them in that sense? They are annoying to maintain, though.

app/helpers/bulk_upload/lettings_log_to_csv.rb
samyou-softwire: don't think we do; a valid log should already have these last two as nil if they're not asked

app/helpers/bulk_upload/lettings_log_to_csv.rb
samyou-softwire: ah good catch 👍

app/services/bulk_upload/lettings/year2026/csv_parser.rb
samyou-softwire: this came from the previous years' validator, though I can't see any reason for it to be 1 lower. it works if I keep them the same, so will decrease them

spec/models/form/lettings/subsections/household_situation_spec.rb
samyou-softwire: this test used to check 2024 logic. suppose the diff is a bit unhelpful here

spec/services/bulk_upload/lettings/validator_spec.rb
samyou-softwire: yeah suppose if these failed for another reason we'd be interested in knowing. I've added a comment to make this a little clearer

app/helpers/bulk_upload/lettings_log_to_csv.rb
oscar-richardson-softwire: I may well be missing something here but it's still not obvious to me that we're doing the same thing as 2025 here (if that is indeed the intention). It looks like in 2025, if `overrides[:organisation_id]` is truthy then we use this value, whereas in 2026 we use `Organisation.find(overrides[:organisation_id])&.old_visible_id`. Is this definitely/always the same thing?

samyou-softwire: so looked a little further, the only usage of this override is on the BU Test File buttons 
screenshot of bu test buttons

it's only ever passed as "ORG#{id}", which is how orgs are presented in the BU, and is always the owning org ID. I think the later section doesn't need to depend on this override, looks to be an override for presentation purposes of the same org

I've changed it to be like this now

app/services/bulk_upload/lettings/year2026/row_parser.rb
samyou-softwire: turns out this change is a little problematic; orgs in tests are by default LAs and so this will be false. however we don't have the flow setup yet to allow for the LA flow to ignore referral_noms so tests start failing

I've made these fields always be returned for now, will make truly correct in later PRs

samyou-softwire:
rebase back onto [CLDC-4115](https://mhclgdigital.atlassian.net/browse/CLDC-4115)

[CLDC-4115]: https://mhclgdigital.atlassian.net/browse/CLDC-4115?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

oscar-richardson-softwire (approved):
LGTM

samyou-softwire:
rebase after merging [CLDC-4115](https://mhclgdigital.atlassian.net/browse/CLDC-4115)

[CLDC-4115]: https://mhclgdigital.atlassian.net/browse/CLDC-4115?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

samyou-softwire:
rebased in the [CLDC-4188](https://mhclgdigital.atlassian.net/browse/CLDC-4188) and [CLDC-4189](https://mhclgdigital.atlassian.net/browse/CLDC-4189) branches after they were PO approved.

waiting on [CLDC-4190](https://mhclgdigital.atlassian.net/browse/CLDC-4190) and [CLDC-4191](https://mhclgdigital.atlassian.net/browse/CLDC-4191) to be approved before merging this branch in to add the new package of work

[CLDC-4188]: https://mhclgdigital.atlassian.net/browse/CLDC-4188?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[CLDC-4189]: https://mhclgdigital.atlassian.net/browse/CLDC-4189?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[CLDC-4190]: https://mhclgdigital.atlassian.net/browse/CLDC-4190?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[CLDC-4191]: https://mhclgdigital.atlassian.net/browse/CLDC-4191?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

app/models/lettings_log.rb
```diff
@@ -544,6 +550,12 @@ def is_prevten_la_general_needs?
     [30, 31].any?(prevten)
   end
 
+  def is_prevten_general_needs?
+    return false unless prevten
+
+    ![30, 31, 32, 33, 35, 38, 6].include?(prevten)
```
samyou-softwire: note that if [CLDC-4149](https://mhclgdigital.atlassian.net/browse/CLDC-4149) is merged before this we'll need to add a code 40 here

app/models/form/lettings/questions/referral_noms.rb
```diff
@@ -6,7 +6,7 @@ def initialize(id, hsh, page, referral_register)
     @copy_key = "lettings.household_situation.referral.noms"
     @type = "radio"
     @check_answers_card_number = 0
-    @question_number = QUESTION_NUMBER_FROM_YEAR[form.start_date.year] || QUESTION_NUMBER_FROM_YEAR[QUESTION_NUMBER_FROM_YEAR.keys.max]
+    @question_number = QUESTION_NUMBER_FROM_YEAR[form.start_date.year]
```
oscar-richardson-softwire: Just wondering whether it makes sense to leave this as is to save us from having to make a change in a bunch of places just to get stuff working initially when we set up the 2027 forms?

oscar-richardson-softwire: And same elsewhere this was changed

samyou-softwire: perhaps this is more a comment to go on #3169 as it contains most of these changes

the thinking behind removing this is that:
- we update most questions numbers every year
- the current logic is clever but not very helpful. 
  - removing a val for a year will not assume the next one but instead the latest, this can be a surprising result
  - this can make support hard
- removing the fallback enforces that a new question number is specified every year

for what its worth I could refine the script a little that copilot generated that adds a new 2027 (or other year) value to all the question dictionaries, which should cut down the amount of work needed by a lot

samyou-softwire:
rebase onto main

app/models/lettings_log.rb
samyou-softwire: have done so

app/models/form/lettings/questions/referral_noms.rb
oscar-richardson-softwire: Ah gotcha, yeah happy with this reasoning

oscar-richardson-softwire (approved):
Edit: Happy for this to be merged once the failing test is addressed

PR #3151, opened 2026-01-27
CLDC-4450: Update local development docs

katlan-sw:
## Context

Given the amount of faffing around that I did trying to work out what to do for creating a new question (CLDC-4177), I thought it valuable to document the steps that were necessary. This should save time for future updates.

I also elected to update the existing "Question" page as it was very out of date.

There are undoubtedly more things that need an update, but I have resisted the urge to change them at this time.

### Pictures

Taken at a low zoom level to give a better overview.

#### The "new question" guide

image

#### The "question" page

image


docs/dev_tasks/new_question.md
```diff
@@ -0,0 +1,119 @@
+---
+parent: Common dev tasks
+nav_order: 1
+---
+
+# New Sales Log Questions
```
samyou-softwire: could this be generic between sales and lettings questions? don't think anything here is sales specific

docs/form/question.md
```diff
@@ -6,81 +6,106 @@ nav_order: 4
...
+Let's take a look at the properties in the `initialize` function.
 
-The `check_answer_label` contains the text that will be displayed in the label of the table on the check answers page.
+<dl>
+<dt>id</dt>
+<dd>The name of the field. This should correspond to a column in the database. In the example, the id is 'ppcodenk'.</dd>
 
-The header is text that is displayed for the question.
+<dt>copy_key</dt>
+<dd>Unknown; requires investigation.</dd>
```
samyou-softwire: believe this is what specifies which entry in the locale file to pull

docs/dev_tasks/new_question.md
```diff
@@ -0,0 +1,119 @@
...
+- Add the new field to `field_mapping_for_errors` with the name of the field in the database.
+- Add the new field to `attributes_for_log` with the name of the field in the database.
+
+You may also have to add some additional validation rules in this file.
+
+Validation for ensuring that the value uploaded is one of the permitted options is handled automatically, using the question class as the original source of truth.
+
+### 9. Update unit tests
+
+Create new test files for any new classes you have created. Update any test files for files that you have edited.
```
samyou-softwire: one other task we should do; add a description for the new question to the relevant `lettings_download_year.csv`, for instance: https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/blob/e04e95b029abfb06e2fbe37fd97d8190650abfef/spec/fixtures/variable_definitions/lettings_download_26_27.csv

not essential but means the log download tests can be more realistic

docs/dev_tasks/new_question.md
katlan-sw: I will let you know if I encounter anything different. I didn't want to assume and it be wrong.

docs/form/question.md
katlan-sw: That would make sense!

docs/dev_tasks/new_question.md
samyou-softwire: fwiw I've been adding a question to lettings and the advice all looks relevant to me

natdeanlewissoftwire: Have renamed this single file and added a note that it's generic for both question types.

docs/dev_tasks/new_question.md
```diff
@@ -0,0 +1,121 @@
...
+
+### 5. Add new page to an existing subsection
+
+Without this step, your new page will not be inserted into the form!
+
+Subsections can be found in `app/models/form/sales/subsections`.
+
+You will want to add your new page to the appropriate place in the list returned by `def pages`.
+
+To make your new page only appear in the forms for the upcoming year, you wrap the page class in parentheses and add a conditional expression to the end, like so:
```
samyou-softwire, line 77: worth making clear here, for the @id attribute of a page
- its what's shown in URLs
- for a user journey, page IDs cannot be repeated, as in:
  - if the user sees two pages, they cannot share an ID
  - if the user sees only one (eg 25 vs 26 questions), they can share an ID

docs/form/question.md
natdeanlewissoftwire: Have added this :)

docs/dev_tasks/new_question.md
natdeanlewissoftwire: think I've captured this - lmk if I've misunderstood :)

samyou-softwire: ftoad this isn't strictly speaking the logic. when we define subsections and add pages to it, pages in the list must have a unique ID. if we want two pages to have the same ID, one must be removed from the list of pages

note that if we use other mechanism to show/hide pages (eg `routed_to?` method), this doesn't count and they can never share an ID.

the only place where we remove pages from the subsection list is when we split on year, so the logic as you've put it is effectively correct. I think the whole picture is a bit tricky to explain

natdeanlewissoftwire: Ah I see - in that case I think this is probably a good starting point docs wise but given the complexity I think the source of truth for all the details just being the code is probably best anyway  - but equally if you'd rather this be rephrased in the docs happy to take your lead on a better wording :)

samyou-softwire: yeah let's leave it here then

samyou-softwire:
I'm happy for this to be reviewed now. see updates made to the page, question and new question pages

page docs

question docs

new question docs

natdeanlewissoftwire (approved):

PR #3152, opened 2026-01-28
CLDC-4146 infer no other partners

carolynbarker:
Closes https://mhclgdigital.atlassian.net/browse/CLDC-4146

For lettings only, we want to only ask partner until it has been answered "yes" once, then infer "no" for unanswered questions. And also infer "no" for any other relationship questions previously answered "yes", but not for "prefers not to say".

If there's multiple tenants marked as a partner we intentionally give priority to the lower numbered tenants, as it is much more common for people to input the lead tenant's partner second rather than to enter children or other housemates first and the partner later.


app/models/lettings_log.rb
```diff
@@ -873,6 +873,10 @@ def get_lettype
     end
   end
 
+  def partner_numbers
+    (2..8).select { |i| public_send("relat#{i}") == "P" }
+  end
+
```
carolynbarker, line 953: part of my thinking here was that we can commonise this with another place it's used, but the other usage actually looks like it's pre-2024 only

app/models/form/lettings/subsections/household_characteristics.rb
```diff
@@ -34,16 +34,16 @@ def pages
       Form::Lettings::Pages::PersonKnown.new(nil, nil, self, person_index: 2),
       (Form::Lettings::Pages::PersonAge.new(nil, nil, self, person_index: 2) if form.start_year_2026_or_later?),
       relationship_question(person_index: 2),
-      (Form::Lettings::Pages::PartnerUnder16ValueCheck.new("relationship_2_partner_under_16_value_check", nil, self, person_index: 2) if form.start_year_2024_or_later? && !form.start_year_2026_or_later?),
-      (Form::Lettings::Pages::MultiplePartnersValueCheck.new("relationship_2_multiple_partners_value_check", nil, self, person_index: 2) if form.start_year_2024_or_later?),
+      (Form::Lettings::Pages::PartnerUnder16ValueCheck.new("relationship_2_partner_under_16_value_check", nil, self, person_index: 2) unless form.start_year_2026_or_later?),
+      (Form::Lettings::Pages::MultiplePartnersValueCheck.new("relationship_2_multiple_partners_value_check", nil, self, person_index: 2) unless form.start_year_2026_or_later?),
```
carolynbarker: I think `form.start_year_2024_or_later?` is now always true

spec/models/lettings_log_derived_fields_spec.rb
```diff
@@ -1446,4 +1446,156 @@
...
+    context "when 2025", metadata: { year: 25 } do
+      before do
+        Timecop.freeze(Time.zone.local(2025, 5, 10))
+        Singleton.__init__(FormHandler)
+      end
+
+      after do
+        Timecop.return
+        Singleton.__init__(FormHandler)
+      end
```
carolynbarker, line 1602: because I'm using `changed`, I want to actually `create` (or save) rather than just `build` the log, so the time has to be within two weeks of the log start date. Open to other approaches if anyone has suggestions!

app/models/derived_variables/lettings_log_variables.rb
```diff
@@ -290,6 +294,23 @@ def clear_child_constraints_for_age_changes!
     end
   end
 
+  def infer_at_most_one_relationship!
+    new_partner_numbers = partner_numbers.select { |i| public_send("relat#{i}_changed?") }
+    if new_partner_numbers.any?
+      infer_only_partner!(new_partner_numbers.first)
+    elsif partner_numbers.any?
+      infer_only_partner!(partner_numbers.first)
+    end
```
oscar-richardson-softwire: I may be wrong, but I don't think we need this branching logic, as I think `new_partner_numbers.first == parter_numbers.first` in all cases (`new_partner_numbers.first` couldn't be < `partner_numbers.first` as `new_partner_numbers` is a subset of `partner_numbers` and `new_partner_numbers.first` couldn't be > `partner_numbers.first` as we wouldn't have been able to answer a later partner question as `P`).

app/models/form/lettings/questions/person_partner.rb
```diff
@@ -7,6 +7,12 @@ def initialize(id, hsh, page, person_index:)
     @answer_options = answer_options
     @person_index = person_index
     @question_number = question_number
+    @hidden_in_check_answers = {
+      "depends_on" => (2...person_index).map do |i|
+        { "relat#{i}" => "P", "relat#{person_index}" => "X" }
+      end,
+    }
```
oscar-richardson-softwire: Currently, we have no way of telling if a 'No' to the partner question was inferred or answered manually. Hence, the logic generally treats 'No's that come after a 'Yes' as inferred and 'No's that come before a 'Yes' as answered manually, which is usually the case, but not always.* Because of this, 'No's that come after a 'Yes' are hidden from the CYA and reset to nil if the 'Yes' answer changes (even though they may have been answered manually), and 'No's that come before a 'Yes' are shown in the CYA and not reset (even though they may have been inferred). I'm working on the assumption that this behaviour is acceptable.

*For instance, a user could answer 'No' and then go back to an earlier partner question and answer 'Yes' (i.e., a manually answered 'No' that comes after a 'Yes) or they could skip a partner question, then answer 'Yes' to a later partner question so that the skipped question is inferred as 'No' (i.e., an inferred 'No' that comes before a 'Yes').

app/models/derived_variables/lettings_log_variables.rb
```diff
@@ -243,6 +247,17 @@ def clear_inapplicable_derived_values!
...
+      (2..8).each do |i|
+        if send("relat#{i}_changed?") && send("relat#{i}_was") == "P"
+          (i + 1..8).each do |j|
+            if self["relat#{j}"] == "X"
+              self["relat#{j}"] = nil
+            end
+          end
+        end
+      end
+    end
```
oscar-richardson-softwire: Currently, we have no way of telling if a 'No' to the partner question was inferred or answered manually. Hence, the logic generally treats 'No's that come after a 'Yes' as inferred and 'No's that come before a 'Yes' as answered manually, which is usually the case, but not always.* Because of this, 'No's that come after a 'Yes' are reset to nil if the 'Yes' answer changes (even though they may have been answered manually), and 'No's that come before a 'Yes' are not reset (even though they may have been inferred). I'm working on the assumption that this behaviour is acceptable.

*For instance, a user could answer 'No' and then go back to an earlier partner question and answer 'Yes' (i.e., a manually answered 'No' that comes after a 'Yes) or they could skip a partner question, then answer 'Yes' to a later partner question so that the skipped question is inferred as 'No' (i.e., an inferred 'No' that comes before a 'Yes').

app/models/derived_variables/lettings_log_variables.rb
```diff
@@ -295,6 +302,38 @@ def clear_child_constraints_for_age_changes!
...
+    person_count = hhmemb || 8
+    (2..person_count).each do |i|
+      if send("relat#{i}_changed?") && send("relat#{i}_was") == "P"
+        ((i + 1)..person_count).each do |j|
+          if self["relat#{j}"] == "X"
+            self["relat#{j}"] = nil
+          end
+        end
+      else
+        next
```
samyou-softwire: does this do anything?

app/models/derived_variables/lettings_log_variables.rb
samyou-softwire: think this sounds fine, though suppose there is the thought of when we can't tell if the answer is inferred or not, we should always err on the side of resetting the answer. this avoids the possibility of a log being submitted with info the user that isn't entered or currently inferred. but then suppose it could get annoying to reset answers the user did give

worth a check on the ticket?

app/models/derived_variables/lettings_log_variables.rb
```diff
@@ -295,6 +302,38 @@ def clear_child_constraints_for_age_changes!
...
+
+  def infer_only_partner!(partner_number)
+    other_partner_numbers = partner_numbers.reject { |x| x == partner_number }
+    other_partner_numbers.each { |i| self["relat#{i}"] = "X" }
+
+    return unless hhmemb
+
+    unanswered_partner_questions = (2..hhmemb).select { |i| public_send("relat#{i}").nil? }
+
+    unanswered_partner_questions.each { |i| self["relat#{i}"] = "X" }
```
samyou-softwire: I'm not sure why we have two loops here on different ranges.

I think only the latter one is needed, for each member in household
- if relatx is P and its not partner number, set to x
- else set to x

feels a bit more straightforward to parse if we only loop once

app/models/derived_variables/lettings_log_variables.rb
```diff
@@ -295,6 +302,38 @@ def clear_child_constraints_for_age_changes!
...
+    end
+  end
+
+  def infer_only_partner!(partner_number)
+    other_partner_numbers = partner_numbers.reject { |x| x == partner_number }
+    other_partner_numbers.each { |i| self["relat#{i}"] = "X" }
+
+    return unless hhmemb
+
+    unanswered_partner_questions = (2..hhmemb).select { |i| public_send("relat#{i}").nil? }
```
samyou-softwire: in any case, think this variable should be `unanswered_partner_numbers`

app/models/form/lettings/pages/person_lead_partner.rb
```diff
@@ -18,8 +18,11 @@ def depends_on
             "operator" => ">=",
             "operand" => 16,
           },
+          **(2...@person_index).map { |i| ["relat#{i}", { "operator" => "!=", "operand" => "P" }] }.to_h,
```
samyou-softwire: this logic seems odd to me.
- we infer _all_ relationships if one is P
- we hide _later_ relationships if one is P

feels like we should decide one way or the other; are we inferring all or just the later ones?

app/models/form/lettings/questions/person_partner.rb
```diff
@@ -7,6 +7,7 @@ def initialize(id, hsh, page, person_index:)
     @answer_options = answer_options
     @person_index = person_index
     @question_number = question_number
+    @disable_clearing_if_not_routed_or_dynamic_answer_options = form.start_year_2026_or_later?
```
samyou-softwire: oh haven't seen this before, I've always ran on the assumption that the log will clear this value and then reinfer it every time, though nice that we can signal to CORE to skip this

spec/models/form/lettings/pages/person_lead_partner_spec.rb
```diff
@@ -75,6 +75,10 @@
         allow(form).to receive(:start_year_2026_or_later?).and_return(false)
       end
 
+      it "has correct questions" do
+        expect(page.questions.map(&:id)).to eq(%w[relat3])
```
samyou-softwire: why is this now year specific?

samyou-softwire (changes requested):

app/models/derived_variables/lettings_log_variables.rb
oscar-richardson-softwire: Good shout, have asked on the ticket

app/models/derived_variables/lettings_log_variables.rb
oscar-richardson-softwire: No, but Rubocop was complaining about it. I tried the suggested fix instead which is the more 'idiomatic Ruby' use of a guard clause, so have gone with that

app/models/derived_variables/lettings_log_variables.rb
oscar-richardson-softwire: Good shout, I've refactored to only use one loop

app/models/derived_variables/lettings_log_variables.rb
oscar-richardson-softwire: Removed this variable as part of the refactor

app/models/form/lettings/pages/person_lead_partner.rb
oscar-richardson-softwire: Agreed. Rereading it, this does seem to be the implied desired behaviour, but it is unclear and I would question if this is what we really want. Will ask Rachel on the ticket.

spec/models/form/lettings/pages/person_lead_partner_spec.rb
oscar-richardson-softwire: Because the `initialize` method of the question now depends on `form.start_year_2026_or_later?`, because we added:

`@disable_clearing_if_not_routed_or_dynamic_answer_options = form.start_year_2026_or_later?`

spec/models/lettings_log_derived_fields_spec.rb
oscar-richardson-softwire: @samyou-softwire just noticed this comment again and realised I'm still not quite sure what Carolyn meant here. Don't suppose you have any ideas?

spec/models/form/lettings/pages/person_lead_partner_spec.rb
samyou-softwire: ah that's interesting, the other tests pass as we only init question upon calling `questions`. makes sense 👍

samyou-softwire (approved):
LGTM, though we'll need clarity from Rachel on what the precise inferring logic is here before moving to PO review

spec/models/lettings_log_derived_fields_spec.rb
samyou-softwire: I think the point is that
- the inference uses `changed?` and `_was` attributes to work
- these come from active record
- they need a real value in the database
- `create` does this
- creating a log will run all the validations
- one validation is the one on startdate
- future logs are turned off while testing so this validation runs
- the log must be 2025 for the test to make sense
- so we must move time back to 2025 so the startdate validation passes

app/models/form/lettings/pages/person_lead_partner.rb
oscar-richardson-softwire: As per [Rachel's comment](https://mhclgdigital.atlassian.net/browse/CLDC-4146?focusedCommentId=798420), we now hide _all_ relationship questions if one is P (although not from the CYA page).

app/models/form.rb
```diff
@@ -323,7 +324,12 @@ def depends_on_met(depends_on, log)
         if value.is_a?(Hash) && value.key?("operator")
           operator = value["operator"]
           operand = value["operand"]
-          log[question]&.send(operator, operand)
+
+          if operator == "!=" # This branch is needed as `nil` does not behave as expected with the default logic (`nil&.send("!=", operand)` => `nil` i.e., `false`).
+            log[question] != operand
+          else
+            log[question]&.send(operator, operand)
+          end
```
oscar-richardson-softwire: I don't think this is actually needed anymore since I removed the `depends_on` but I'm inclined to suggest we leave it (at the risk of being YAGNI) because this behaviour is so clearly what we want and it was surprising to me (and took a bit of figuring out) when it didn't behave like this

spec/models/lettings_log_derived_fields_spec.rb
```diff
@@ -1588,4 +1588,269 @@
...
+        it "does not reset any answers when the relationship is removed" do
+          log.relat3 = "X"
+          expect { log.set_derived_fields! }
+            .to not_change(log, :relat2)
+            .and not_change(log, :relat4)
+            .and not_change(log, :relat5)
+            .and not_change(log, :relat6)
+            .and not_change(log, :relat7)
+            .and not_change(log, :relat8)
+        end
```
oscar-richardson-softwire: I don't think Rachel has explicitly confirmed that this is what she wants but given that we are now mostly treating answers of 'No' as if they were answered manually (except from hiding inferred 'No's in the initial form flow) and Rachel's comment of "my starting principle is to keep as much manually entered info as possible", I think it's fairly safe to assume that this is the desired behaviour. I've asked for confirmation on the ticket though

oscar-richardson-softwire:
Updated as per comments https://mhclgdigital.atlassian.net/browse/CLDC-4146?focusedCommentId=798025 and https://mhclgdigital.atlassian.net/browse/CLDC-4146?focusedCommentId=798420

app/models/derived_variables/lettings_log_variables.rb
```diff
@@ -325,6 +329,27 @@ def clear_child_constraints_for_age_changes!
     end
   end
 
+  def infer_at_most_one_relationship!
+    new_partner_numbers = partner_numbers.select { |i| public_send("relat#{i}_changed?") }
+    if new_partner_numbers.any?
+      infer_only_partner!(new_partner_numbers.first)
+    elsif partner_numbers.any?
+      infer_only_partner!(partner_numbers.first)
```
samyou-softwire: will this path ever be used?

app/models/form/lettings/questions/person_partner.rb
samyou-softwire: do we still need this?

spec/models/form/lettings/pages/person_lead_partner_spec.rb
```diff
@@ -81,7 +118,6 @@
 
     context "with start year >= 2026", metadata: { year: 26 } do
       let(:start_year_2026_or_later?) { true }
-      let(:person_question_count) { 5 }
```
samyou-softwire, line 84: seems odd that we were able to remove these?

spec/models/form_spec.rb
```diff
@@ -23,6 +23,21 @@
...
+      let(:next_page) { form.get_page("person_1_gender") }
+
+      before do
+        allow(next_page).to receive(:skip_page_in_form_flow?)
+          .with(lettings_log)
+          .and_return(true)
+      end
+
+      it "returns the page after next, given the previous" do
+        puts next_page.skip_page_in_form_flow?(lettings_log)
```
samyou-softwire: think this can be removed

spec/models/lettings_log_derived_fields_spec.rb
```diff
@@ -1588,4 +1588,269 @@
...
+        let(:log) { create(:lettings_log, :completed, relat2: "X", relat3: "X", relat4: "R", relat5: "X") }
+
+        before do
+          log.relat2 = "P"
+          log.relat3 = "P"
+        end
+
+        it "does not infer no to any relationship answers" do
+          expect { log.set_derived_fields! }
+            .to not_change(log, :relat4)
```
samyou-softwire: worth affirming that neither relat2 not relat3 will change?

spec/models/lettings_log_derived_fields_spec.rb
```diff
@@ -1588,4 +1588,269 @@
...
+            .and not_change(log, :relat6)
+            .and not_change(log, :relat7)
+            .and not_change(log, :relat8)
+        end
+      end
+    end
+
+    context "when 2026", metadata: { year: 26 } do
+      before do
+        Timecop.freeze(Time.zone.local(2026, 5, 10))
```
samyou-softwire: `CollectionTimeHelper` has a method `collection_start_date_for_year(year)` which can get an appropriate date without needing to write one out manually

spec/models/lettings_log_derived_fields_spec.rb
```diff
@@ -1588,4 +1588,269 @@
...
+
+          it "does not change relationship answers of no or prefer not to say" do
+            expect { log.set_derived_fields! }
+              .to not_change(log, :relat3)
+              .and not_change(log, :relat5)
+          end
+        end
+
+        it "does not change other relationship values if no is changed to prefer not to say" do
+          log.relat4 = "R"
```
samyou-softwire: could these be set in a before? makes it consistent with the above

samyou-softwire (changes requested):
LGTM on the logic, few qs around other areas

app/models/derived_variables/lettings_log_variables.rb
oscar-richardson-softwire: Great catch, don't think it will, removed

app/models/form/lettings/questions/person_partner.rb
oscar-richardson-softwire: Good spot, removed

spec/models/lettings_log_derived_fields_spec.rb
oscar-richardson-softwire: Rachel has confirmed that this is the desired behaviour.

spec/models/form/lettings/pages/person_lead_partner_spec.rb
oscar-richardson-softwire: So my thinking was that these are redundant in these tests as this is the page (which doesn't have a question number) rather than the question, but I've realised now that we don't currently test page number in the question either (`spec/models/form/lettings/questions/person_partner_spec.rb`) so I've added these tests.

spec/models/form_spec.rb
oscar-richardson-softwire: Removed

spec/models/lettings_log_derived_fields_spec.rb
oscar-richardson-softwire: Good point, added similar assertions everywhere that's relevant.

spec/models/lettings_log_derived_fields_spec.rb
oscar-richardson-softwire: Tests now use this

spec/models/lettings_log_derived_fields_spec.rb
oscar-richardson-softwire: Changed everywhere that's applicable

spec/factories/lettings_log.rb
```diff
@@ -14,6 +14,17 @@
...
+    after(:build) do |log|
+      if log.hhmemb
+        (2..log.hhmemb).each do |i|
+          setter = "details_known_#{i}="
+          getter = "details_known_#{i}"
+
+          log.public_send(setter, 0) if log.public_send(getter).nil?
+        end
+      end
+    end
```
oscar-richardson-softwire: Tests started failing after I removed `@disable_clearing_if_not_routed_or_dynamic_answer_options` (I still agree that we don't need it) - this is because `details_known_{i}` was `nil` for `i > 2`, so when we `create` a lettings log from the factory, we end up clearing `relat{i}` for `i > 2` as the question is considered not routed to.

I added this as I figured this is probably the most useful way of doing it (if we want the details not to be known for a member of the household, we are more likely to think to set this explicitly). Shout if you disagree.

All of the tests still passed so clearly nothing was relying on the previous behaviour.

samyou-softwire: this feels a bit odd to do to me. not 100% what tests were failing here, but it feels more natural to either:
- update the tests so they set up details_known appropriately for their case
- update the trait to set up details_known appropriately for what makes sense for that trait

setting details_known to 0 (yes) globally across all tests sounds a bit surprising and easy to miss the mechanism that sets this when debugging if you don't know about this edge case

app/models/lettings_log.rb
```diff
@@ -445,10 +445,16 @@ def bedsit_changed_to_not_bedsit?
     unittype_gn_changed? && unittype_gn_was == 2
   end
 
-  def is_partner_inferred?(person_index)
+  def is_person_under_16(person_index)
```
samyou-softwire: really minor, a ? would be appropriate here `is_person_under_16?`

app/models/lettings_log.rb
```diff
@@ -445,10 +445,16 @@ def bedsit_changed_to_not_bedsit?
...
     public_send("age#{person_index}") && public_send("age#{person_index}") < 16
   end
 
+  def is_another_person_partner?(person_index)
+    partner_numbers.delete(person_index)
+
+    !partner_numbers.empty?
```
samyou-softwire: looking over it again, this function seemed a bit confusing to me. specifically, the `partner_numbers.delete(person_index)` line
- at the point of calling this function from `skip_page_in_form_flow?`, you've not answered the question yet, so person at person_index cannot be a partner. I'm not sure when the delete would come in handy
- partner_numbers is a function so I'm not sure calling `delete` on it has any impact

samyou-softwire (changes requested):
just one final comment (and a fixup)

app/models/form/page.rb
```diff
@@ -55,6 +55,10 @@ def interruption_screen?
 
   def skip_href(log = nil); end
 
+  def skip_page_in_form_flow?(_log)
```
samyou-softwire: one thought, I think this should be a property of a question and not a page. it's most close to what we want to do here (skip the question because it's already got an answer). we would only go ahead and skip the page if all questions are skipped. setting it on the page feels like a shortcut that blurs the line a little between a page & question's responsibilities

it doesn't make much difference since most pages have one question so don't mind whether we go ahead or not but wonder your thoughts on this

app/models/lettings_log.rb
oscar-richardson-softwire: Great spot, you're right on both points. I've rewritten this method as `is_any_person_partner?` with the appropriate logic.

app/models/form/page.rb
oscar-richardson-softwire: Agreed, I've moved this method to the question class

app/models/lettings_log.rb
oscar-richardson-softwire: Done

samyou-softwire (approved):
LGTM

PR #3153, opened 2026-01-29
CLDC-4188: Q84 Referral LA path

samyou-softwire:
closes [CLDC-4188](https://mhclgdigital.atlassian.net/browse/CLDC-4188)

sets up LA flow through Q84, as of now uses only the `referral_register` question

sets up inference for renewals

adds validation between this and Q77 prevten, as well as disabling ones from previous years

q84
CYA
q84 validation
q77 validation

[CLDC-4188]: https://mhclgdigital.atlassian.net/browse/CLDC-4188?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/models/derived_variables/lettings_log_variables.rb
```diff
@@ -177,7 +177,54 @@ def set_derived_fields!
...
-  DEPENDENCIES = [
+  def dependencies
+    if form.start_year_2026_or_later?
+      DEPENDENCIES_2026
+    else
+      DEPENDENCIES_PRE_2026
+    end
+  end
+
+  DEPENDENCIES_2026 = [
```
samyou-softwire: wonder if there is something clever we could do with merging here to avoid duplicating the dependencies list, though since this is an array of objects it felt too tricky to be worthwhile

app/models/form/lettings/pages/referral_noms.rb
```diff
@@ -8,4 +8,8 @@ def initialize(id, hsh, subsection)
   def questions
     @questions ||= [Form::Lettings::Questions::ReferralNoms.new(nil, nil, self)]
   end
+
+  def routed_to?(log, _current_user)
+    log.owning_organisation&.prp?
+  end
```
samyou-softwire, line 14: disables them for this route, to be set up properly in [CLDC-4189](https://mhclgdigital.atlassian.net/browse/CLDC-4189)

app/models/form/lettings/pages/referral_register_la.rb
```diff
@@ -0,0 +1,15 @@
+# added in 2026
```
samyou-softwire, line 1: its not easy to have conditional logic in these page files depend on log properties.

we want different question options but the `question_answers` field on the question cannot access the log.

the only method that does here is `routed_to`, so we define a page for LAs and one for PRPs and route to it as required

we don't need to duplicate the question, just tell it which page its being shown on

config/locales/validations/lettings/household.en.yml
```diff
@@ -98,8 +98,12 @@ en:
           internal_transfer: "Answer cannot be %{prevten} as this tenancy is an internal transfer."
           la_general_needs:
             internal_transfer: "Answer cannot be a fixed-term or lifetime local authority general needs tenancy as it’s an internal transfer and a private registered provider is on the tenancy agreement."
+          general_needs:
+            internal_transfer: "Answer cannot be %{prevten} as this tenancy is an internal transfer. Internal transfers are for existing social tenants only, your answer to where the household was immediately before this letting shows the tenant was not in the social sector immediately prior to this letting."
```
samyou-softwire: have left a comment on the ticket to confirm this wording as its not exactly specified in the ticket

spec/services/merge/merge_organisations_service_spec.rb
```diff
@@ -56,6 +56,7 @@
 
       it "rolls back if there's an error" do
         allow(Organisation).to receive(:find).with(merging_organisation_ids).and_return(Organisation.find(merging_organisation_ids))
+        allow(Organisation).to receive(:find).with(merging_organisation.id).and_return(merging_organisation)
```
samyou-softwire, line 59: funny how this test started failing - this is due to the org changing triggering the `should_reset_referral_register?` function which tries to find the old org

config/locales/validations/lettings/household.en.yml
oscar-richardson-softwire: Also this is a real nitpick (and I know you've copied the second sentence from a ticket but I think this is grammatically incorrect (comma splice). I think we want "Internal transfers are for existing social tenants only; your answer to where..."

app/models/derived_variables/lettings_log_variables.rb
oscar-richardson-softwire: I guess we could just define a `COMMON_DEPENDENCIES` array, then do e.g.,

`DEPENDENCIES_2026 = [
  { conditions: { renewal: 1 }, derived_values: { referral_register: 1, waityear: 2, offered: 0, rsnvac: 14, first_time_property_let_as_social_housing: 0 } },
  *COMMON_DEPENDENCIES
].freeze`

We'd just have to remember to move a condition out of the common dependencies if we need to add overrides for that condition.

app/models/validations/household_validations.rb
```diff
@@ -172,7 +172,13 @@ def validate_previous_housing_situation(record)
...
     return unless record.owning_organisation
 
-    if record.is_internal_transfer? && record.owning_organisation.provider_type == "PRP" && record.is_prevten_la_general_needs?
+    if record.form.start_year_2026_or_later?
+      if record.is_internal_transfer? && record.owning_organisation.la? && ![30, 31, 32, 33, 35, 38, 6].include?(record.prevten)
+        label = record.form.get_question("prevten", record).present? ? record.form.get_question("prevten", record).label_from_value(record.prevten) : ""
+        record.errors.add :prevten, message: I18n.t("validations.lettings.household.prevten.general_needs.internal_transfer", prevten: label)
+        record.errors.add :referral_register, message: I18n.t("validations.lettings.household.referral.general_needs.internal_transfer", prevten: label)
+      end
+    elsif record.is_internal_transfer? && record.owning_organisation.provider_type == "PRP" && record.is_prevten_la_general_needs?
```
oscar-richardson-softwire:
```suggestion
    elsif record.is_internal_transfer? && record.owning_organisation.prp? && record.is_prevten_la_general_needs?
```

oscar-richardson-softwire: Just for style consistency

app/models/validations/household_validations.rb
```diff
@@ -172,7 +172,13 @@ def validate_previous_housing_situation(record)
   def validate_referral(record)
     return unless record.owning_organisation
 
-    if record.is_internal_transfer? && record.owning_organisation.provider_type == "PRP" && record.is_prevten_la_general_needs?
+    if record.form.start_year_2026_or_later?
+      if record.is_internal_transfer? && record.owning_organisation.la? && ![30, 31, 32, 33, 35, 38, 6].include?(record.prevten)
+        label = record.form.get_question("prevten", record).present? ? record.form.get_question("prevten", record).label_from_value(record.prevten) : ""
+        record.errors.add :prevten, message: I18n.t("validations.lettings.household.prevten.general_needs.internal_transfer", prevten: label)
```
oscar-richardson-softwire, line 178: I'm confused about why these validation messages are nested under `general_needs`.

samyou-softwire: I think the error is that the prevten seems to suggest general needs, and this option must be for social tenants, so I've labelled it as `general_needs`. happy to change the name to something clearer

config/locales/validations/lettings/household.en.yml
samyou-softwire: good spot, will ask on the ticket

app/models/derived_variables/lettings_log_variables.rb
samyou-softwire: sounds good; have implemented this

app/models/derived_variables/lettings_log_variables.rb
```diff
@@ -224,44 +212,37 @@ def dependencies
...
-  DEPENDENCIES_PRE_2026 = [
+  DEPENDENCIES_2026 = [
     {
       conditions: {
         renewal: 1,
       },
       derived_values: {
-        referral: 1,
-        referral_type: 3,
+        referral_register: 1,
```
oscar-richardson-softwire, line 221: Just thought - should we explicitly state `referral_noms: nil`, `referral_org: nil` so that we set these if `renewal` is changed to ` and they are non-nil?

app/models/validations/household_validations.rb
oscar-richardson-softwire: Ah gotcha. I think what confused me was the the condition was `record.owning_organisation.la?` rather than `record.is_supported_housing?`. Do these tend to go hand in hand?

samyou-softwire: `![30, 31, 32, 33, 35, 38, 6].include?(record.prevten)` is what's implying general needs. fwiw the method in #3154 to be a bit clearer

app/models/derived_variables/lettings_log_variables.rb
samyou-softwire: this should be handled for us, CORE will nil any questions no longer asked by default

spec/models/validations/household_validations_spec.rb
```diff
@@ -488,6 +490,60 @@
...
+        { code: 24, label: "Home Office Asylum Support" },
+        { code: 37, label: "Host family or similar refugee accommodation" },
+        { code: 23, label: "Mobile home or caravan" },
+        { code: 21, label: "Refuge" },
+        { code: 9, label: "Residential care home" },
+        { code: 4, label: "Tied housing or rented with job" },
+        { code: 25, label: "Any other accommodation" },
+      ].each do |prevten|
+        it "prevten cannot be #{prevten[:code]}" do
+          record.referral_register = 2
```
oscar-richardson-softwire: Could we move this (and in the block below) to a before/do?

oscar-richardson-softwire: Also I can see that the organisation factory creates an org with provider type LA by default but I think it might be clearer what this test is testing if we set that explicitly

spec/models/validations/household_validations_spec.rb
```diff
@@ -593,16 +649,33 @@
...
-          expect(record.errors["prevten"])
-            .to include(match I18n.t("validations.lettings.household.prevten.internal_transfer", prevten: label))
-          expect(record.errors["referral"])
-            .to include(match I18n.t("validations.lettings.household.referral.prevten_invalid", prevten: ""))
+        context "when year is 2025" do
+          let(:startdate) { collection_start_date_for_year(2025) }
+
+          it "prevten cannot be #{prevten[:code]}" do
+            record.referral_type = 3
+            record.referral = 1
```
oscar-richardson-softwire: Could this be moved to a before/do please. Also do we need `record.referral_type = 3`?

spec/models/validations/household_validations_spec.rb
```diff
@@ -593,16 +649,33 @@
...
+              .to include(match I18n.t("validations.lettings.household.referral.prevten_invalid", prevten: ""))
+          end
+        end
+
+        context "when year is 2026" do
+          let(:startdate) { collection_start_date_for_year(2026) }
+
+          it "prevten can be #{prevten[:code]}" do
+            record.referral_type = 3
+            record.referral = 1
```
oscar-richardson-softwire:
```suggestion
            record.referral_register = 2
```

spec/models/validations/household_validations_spec.rb
samyou-softwire: have done so

spec/models/validations/household_validations_spec.rb
samyou-softwire: have restructured this section as well to set the correct args

samyou-softwire:
rebase onto [CLDC-4151](https://mhclgdigital.atlassian.net/browse/CLDC-4151)

[CLDC-4151]: https://mhclgdigital.atlassian.net/browse/CLDC-4151?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

spec/models/validations/household_validations_spec.rb
samyou-softwire: should be handled by the above restructures

spec/models/validations/household_validations_spec.rb
oscar-richardson-softwire: I meant, isn't `record.referral_type = 3` redundant here? `is_internal_transfer?` only seems to check `referral == 1`

samyou-softwire: I think the point behind this is that `referral_type == 3` is a required answer for `referral == 1`, so the log is more 'valid' this way. though functionally it doesn't make much of a difference

have removed it for now though happy to put it back

samyou-softwire:
rebase after merging [CLDC-4115](https://mhclgdigital.atlassian.net/browse/CLDC-4115)

[CLDC-4115]: https://mhclgdigital.atlassian.net/browse/CLDC-4115?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

oscar-richardson-softwire (approved):
LGTM

PR #3156, opened 2026-02-02
CLDC-4141: New sex registered at birth question for lettings logs

natdeanlewissoftwire:
Closes [CLDC-4141](https://mhclgdigital.atlassian.net/browse/CLDC-4141).

Note - base branch is not main so the diff is clearer, the original commits here were based off 4177, so I think it's simpler to just wait for that to be merged (currently in PO review). If that proves slow I'll rebase. Currently test failures are due to that PR, not this one (as of 6th Feb)

Summary

Adds a new question before each gender question that asks about what sex was assigned to the person when they were born.

Applies to Lettings logs only [CLDC-4177](https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/3144) is for sales

This means that there is a new question for:

- Lead Tenant
- Any additional household members, up to 8 people total.

Changes made

- Added new question + page for Lead Tenant for sex registered at birth.
- Added new question + page for Person 2 - 6 for sex assigned at birth.
- Added new properties to the database.
- Added new pages to the "household characteristics" subsection of the sales form.
- Added appropriate unit tests.

Testing

- Created a new lettings log log for a household with a lead tenant and 7 additional people.
	- Saw a new question for Lead Tenant 1 asking about their sex at birth. ✅
	- Saw a new question for People 2 - 8 asking about their sex at birth. ✅
	- Was able to save the form after each new question. ✅
	- Could see the new question appear on the summary page for each person. ✅
	- Exporting the log as CSV includes the answers for the new question. ✅

Screenshots of new pages + changes (note Q number is 0 as in the equivalent sales ticket, these will be updated later as part of a separate package of work).

image
image
image

CSV download:
image

[CLDC-4141]: https://mhclgdigital.atlassian.net/browse/CLDC-4141?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[CLDC-4177]: https://mhclgdigital.atlassian.net/browse/CLDC-4177?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/services/bulk_upload/lettings/year2026/csv_parser.rb
```diff
@@ -4,8 +4,8 @@ class BulkUpload::Lettings::Year2026::CsvParser
   include CollectionTimeHelper
 
   # TODO: CLDC-4162: Update when 2026 format is known
-  FIELDS = 129
-  MAX_COLUMNS = 130
+  FIELDS = 137
```
natdeanlewissoftwire, line 7: These changes were required to test BU without `wrong_template` errors - but these changes don't seem to be in the equivalent sales PR. Is this something we're not committing for now?

app/models/form/lettings/questions/person_sex_registered_at_birth.rb
```diff
@@ -0,0 +1,26 @@
+# frozen_string_literal: true
```
samyou-softwire: I don't think we need this flag

app/services/exports/lettings_log_export_service.rb
```diff
@@ -162,16 +162,38 @@ def add_location_fields!(location, attribute_hash)
...
+
+      if field_name.starts_with?(details_known_prefix) || pattern_age.match(field_name)
+        false
+      end
+
+      fields_for_form.include?(field_name)
+    end
+
+    def get_fields_for_form(lettings_log)
+      included_fields = Set[]
```
samyou-softwire, line 171: we'll need a little care around this. the new logic looks good though it does work a little differently to previously. when I added a new post 2026 field I needed to add it to both `EXPORT_FIELDS` and `POST_2026_EXPORT_FIELDS`. with this new logic it looks like I'd only need to add it to `POST_2026_EXPORT_FIELDS`. does this remain compatible with the previous `EXPORT_FIELDS` lists?

natdeanlewissoftwire: Ah yes I meant to add a comment here - I don't love this personally as it diverges from the sales behaviour. I take it from your comment this was't a pre-agreed approach (this was added before I took over the PR) so I think I'll look at refactoring this :)

config/locales/forms/2026/lettings/household_characteristics.en.yml
```diff
@@ -116,6 +123,13 @@ en:
               hint_text: "Answer 1 for children aged under 1 year old"
               question_text: "Age"
 
+          sexrab2:
+            page_header: ""
+            check_answer_label: "Person 2’s sex registered at birth"
+            check_answer_prompt: ""
+            hint_text: "This is the sex that was registered at birth. The next question will ask about their gender identity."
+            question_text: "What was Person 2's sex at birth?"
```
samyou-softwire: should we capitalise person here? the other questions don't seem to

natdeanlewissoftwire: Good spot will fix

app/models/form/lettings/questions/person_sex_registered_at_birth.rb
natdeanlewissoftwire: agreed, will remove here and in the associated sales PR

app/services/exports/lettings_log_export_service.rb
samyou-softwire: agree I think this proposed logic is better, just want to be sure this won't break anything for previous years. tests should pick up if there is a change in what cols are exported I suppose

samyou-softwire: although, do we have any cases where cols are only around from for instance to 2025 only (not 2024 or 2026)? since this function is only additive I'm not sure how we would model a single year inclusion.

maybe itd be more straightforward to have a list of cols that are always included, then `2026_EXPORT_FIELDS` which gets merged in. would lead to a little repetition but easier to think about what's included where

app/models/form/lettings/questions/person_sex_registered_at_birth.rb
```diff
@@ -0,0 +1,24 @@
...
+  def initialize(id, hsh, page, person_index:)
+    super(id, hsh, page)
+    @type = "radio"
+    @check_answers_card_number = person_index
+    @inferred_check_answers_value = [{
+      "condition" => {
+        id => "R",
+      },
+      "value" => "Person prefers not to say",
+    }]
```
samyou-softwire: I'm not really sure what this section does; its label matches the answer options and I think CYA will draw from there

I don't think this question is ever inferred?

natdeanlewissoftwire: Hm, agreed - it looks like the standard in these cases is to say "Prefers not to say" in the CYA page - is there approved copy somewhere I can confirm this for this ticket (and the sales equivalent which has the same thing)? If not I'll probably just remove these `@inferred_check_answers_value`s

samyou-softwire: yeah don't see a specific request. we can ask on the ticket. if CORE does use a custom copy for CYA prefers not to say think it makes sense to be consistent

oscar-richardson-softwire: @samyou-softwire are you okay to put a comment on the ticket checking this with Rachel?

db/schema.rb
```diff
@@ -375,9 +375,14 @@
     t.bigint "created_by_id"
     t.boolean "manual_address_entry_selected", default: false
     t.integer "referral_type"
-    t.integer "referral_register"
-    t.integer "referral_noms"
-    t.integer "referral_org"
+    t.string "sexrab1"
```
samyou-softwire, line 379: this should be reflected in `spec/factories/lettings.rb` in the `:completed` trait. fine to add a value regardless of the year (if its not asked it'll be nulled automatically).

if we don't have this creating `:completed` logs for 2026 won't work

samyou-softwire (changes requested):

db/schema.rb
```diff
@@ -84,7 +84,7 @@
     t.datetime "last_accessed"
     t.datetime "created_at", null: false
     t.datetime "updated_at", null: false
-    t.check_constraint "log_type::text = ANY (ARRAY['lettings'::character varying, 'sales'::character varying]::text[])", name: "log_type_check"
+    t.check_constraint "log_type::text = ANY (ARRAY['lettings'::character varying::text, 'sales'::character varying::text])", name: "log_type_check"
```
samyou-softwire: don't think this should be here. migrate seems to add this if you run it last on the test db rather than local dev

app/services/exports/lettings_log_export_service.rb
oscar-richardson-softwire: > a list of cols that are always included

Maybe we could rename this something like `ALL_YEARS_EXPORT_FIELDS` to make this even clearer

app/services/bulk_upload/lettings/year2026/row_parser.rb
samyou-softwire, line 1: we'll also need to update the `person_2_present?` etc methods in this class to include the new fields

samyou-softwire (changes requested):

app/services/bulk_upload/lettings/year2026/row_parser.rb
```diff
@@ -134,6 +134,15 @@ class BulkUpload::Lettings::Year2026::RowParser
     field_127: "What is the support charge?",
     field_128: "After the household has received any housing-related benefits, will they still need to pay for rent and charges?",
     field_129: "What do you expect the outstanding amount to be?",
+
+    field_130: "Lead tenant's sex, as registered at birth",
```
samyou-softwire, line 138: I think there's a couple other changes/checks here to make sure the bulk upload works
- the parent `csv_parser.rb` defines `FIELDS` and `MAX_COLUMNS` consts. in #3150 we [decided they can just be the same as eachother](https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/3150#discussion_r2746629069), which will be the field count
- in /spec/fixtures/files there is `2026_27_lettings_bulk_upload.csv`. I'm not sure it's used in any tests but it is useful as a reference for an example BU file to try that'll work for this ticket. can it be updated? fwiw the only important rows in each col are the field number and the value. all others can be left blank for now
- does the 26/27 BU test file button generate a file valid for BU? the conversion from log to BU file is mostly handled by `lettings_log_to_csv.rb`
test homepage

samyou-softwire (changes requested):

app/services/bulk_upload/lettings/year2026/row_parser.rb
natdeanlewissoftwire: Yep, I had made that fields update in a previous commit to confirm BU worked, I've been trying to keep in line with the sales changes on 4177 but it seems that one wasn't actually finished despite being in PO review? Will mention in su, but basically it seems like the scope of this ticket is larger than I had thought so have moved back to draft/in progress for now :)

app/services/bulk_upload/lettings/year2026/row_parser.rb
natdeanlewissoftwire: Good point, done!

spec/services/bulk_upload/lettings/validator_spec.rb
```diff
@@ -11,7 +11,7 @@
   let(:user) { create(:user, organisation:) }
   let(:log) { build(:lettings_log, :completed, period: 2, assigned_to: user) }
   let(:log_to_csv) { BulkUpload::LettingsLogToCsv.new(log:) }
-  let(:bulk_upload) { create(:bulk_upload, user:, year: log.collection_start_year) }
+  let(:bulk_upload) { create(:bulk_upload, :lettings, user:, year:) }
```
natdeanlewissoftwire, line 14: This was previously sometimes passing in non-lettings bulk uploads, and wasn't correctly testing future year bulk uploads

app/services/exports/lettings_log_export_service.rb
natdeanlewissoftwire: Yeah I've tried a couple of things and think this is the simplest - it's a bit fiddly as there are already fields e.g. `carehome_charges_value_check` which have been included one year, and then removed the next - I've checked and sales is a lot simpler so the simple merge behaviour there works fine for now.

This new approach is definitely repetitive const-wise as Sammy mentions but I think avoiding the PRE/POST behaviour is the way to reduce risk if this regressing in future years. Also shows a nice timeline documenting the various changes!

Now when fields are removed, we literally are just removing them from any year const we like rather than adding them to a PRE set, and when fields are added, we are literally just adding them to the new year consts.

db/schema.rb
natdeanlewissoftwire: Yeah I mean this is a non-functional change but agree I don't want to be adding/removing every PR - this is definitely running on `development` `RAILS_ENV` - so probably related to psql versioning diffs in our local setups I think. Happy to remove from the PR manually for now but I'll look into it. Can I ask what PSQL version you are both running @samyou-softwire @oscar-richardson-softwire?

app/services/bulk_upload/lettings/year2026/row_parser.rb
natdeanlewissoftwire: The test BU is valid yep, and have updated the fixture :)

app/helpers/bulk_upload/lettings_log_to_csv.rb
```diff
@@ -92,13 +92,22 @@ def default_2025_field_numbers
...
+      [
+        log.sexrab1, # 130
+        log.sexrab2,
+        log.sexrab3,
+        log.sexrab4,
+        log.sexrab5,
+        log.sexrab6,
+        log.sexrab7,
+        log.sexrab8, # 137
+      ]
```
samyou-softwire: I think we can define the entire row without depending on the 2025 row. it's something we'll end up doing anyway

app/models/form/lettings/questions/person_sex_registered_at_birth.rb
samyou-softwire: have done so

app/services/bulk_upload/lettings/year2026/csv_parser.rb
samyou-softwire: as discussed over call we're going to remove `MAX_COLUMNS` as it doesn't seem to have a practical benefit (looks like you've done this already)

app/models/form/lettings/questions/person_sex_registered_at_birth.rb
natdeanlewissoftwire: FTAOD, have removed these inferrals for now as they're redundant, can always add back in an inferred value in future if needed :)

db/schema.rb
samyou-softwire: postgresql 17.5

spec/fixtures/variable_definitions/lettings_download_26_27.csv
samyou-softwire, line 1: this should be 26_27 I think? the file is additive so it's only necessary to add new variable definitions for this year (all variable definitions will be loaded)

app/services/bulk_upload/lettings/year2026/row_parser.rb
```diff
@@ -541,6 +559,7 @@ def spreadsheet_duplicate_hash
       !supported_housing? ? "field_24" : nil,  # postcode # TODO: CLDC-4119: add postcode to hash for supported housing
       "field_42", # age1
       "field_43",  # sex1
+      "field_130", # sexrab1
```
samyou-softwire, line 562: I think we were going to wait till [CLDC-4119](https://mhclgdigital.atlassian.net/browse/CLDC-4119) until adding the duplicate logs upgrades. is there more work needed for it in terms of the sexrab question?

spec/fixtures/variable_definitions/lettings_download_26_27.csv
natdeanlewissoftwire: You're right... which is odd because there are no variable definition files for last year. So are they really needed?

natdeanlewissoftwire: Have updated the name for now either way

app/helpers/bulk_upload/lettings_log_to_csv.rb
natdeanlewissoftwire: Agree it'll happen in the future, I was thinking that would be part of CLDC-4162 - and this way it's a lot easier to see what changes we're actually making for this particular year while we're doing them, so I'd rather keep this unless you feel v strongly?

app/services/bulk_upload/lettings/year2026/row_parser.rb
natdeanlewissoftwire: Hm, none that I'm aware of but I've not confirmed this thoroughly, just added alongside as it clearly needs to go there. Happy to leave actual thorough testing of this, QA etc to 4119. Equally if it's simpler just to revert this change team-process wise lmk :)

spec/fixtures/variable_definitions/lettings_download_26_27.csv
samyou-softwire: > You're right... which is odd because there are no variable definition files for last year. So are they really needed?

they're not needed particularly, they just mean when a new dev ramps up on the project they get labels in their CSV export, as well as review apps and the tests

app/helpers/bulk_upload/lettings_log_to_csv.rb
samyou-softwire: [CLDC-4151](https://mhclgdigital.atlassian.net/browse/CLDC-4151) #3150 replaces an existing BU field so needs to expand the row. maybe we should put a note to whoever makes the `to_2027_row` function to define the new row explicitly the first time, then we get nice diffs on the PRs

suppose we can leave as is for now and hash this out in merge conflicts

natdeanlewissoftwire: Ah no, fair enough I thought all our changes were additive - I'll make this explicit

PR #3159, opened 2026-02-04
CLDC-NONE: Add a test to ensure future logs are completed

samyou-softwire:
adds a test that the factory traits for future logs work as expected

noticed this has been something we've been missing a lot when adding new questions. feels best to make sure this is tested as we go along, otherwise it'll cause a headache at form release time

### DIFF
--- a/spec/shared/shared_log_examples.rb
+++ b/spec/shared/shared_log_examples.rb
@@ -8,14 +8,14 @@ RSpec.shared_examples "shared log examples" do |log_type|
     [
       {
         start_date: :current_collection_start_date,
-        label: "current year",
+        label: "current",
       },
       {
         start_date: :next_collection_start_date,
-        label: "next year",
+        label: "next",
       },
     ].each do |scenario|
-      context "when creating a log for #{scenario[:label]}" do
+      context "when creating a log for #{scenario[:label]} year" do
         around do |example|
           start_date = Timecop.return { send(scenario[:start_date]) }
 
@@ -30,8 +30,8 @@ RSpec.shared_examples "shared log examples" do |log_type|
         let(:completed_log) { create(log_type, :completed) }
 
         it "is set to not started for an empty #{log_type} log" do
-          expect(empty_log.not_started?).to be(true)
           expect(empty_log.in_progress?).to be(false)
+          expect(empty_log.not_started?).to be(true)
           expect(empty_log.completed?).to be(false)
           expect(empty_log.deleted?).to be(false)
         end
--- a/spec/shared/shared_log_examples.rb
+++ b/spec/shared/shared_log_examples.rb
@@ -2,30 +2,57 @@ require "rails_helper"
 
 # rubocop:disable RSpec/AnyInstance
 RSpec.shared_examples "shared log examples" do |log_type|
+  include CollectionTimeHelper
+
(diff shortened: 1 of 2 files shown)


spec/shared/shared_log_examples.rb
```diff
@@ -2,30 +2,57 @@
...
-    it "is set to completed for a completed #{log_type} log" do
-      expect(completed_log.in_progress?).to be(false)
-      expect(completed_log.not_started?).to be(false)
-      expect(completed_log.completed?).to be(true)
-      expect(completed_log.deleted?).to be(false)
+  describe "status" do
+    [
+      {
+        start_date: :current_collection_start_date,
+        label: "current year",
```
natdeanlewissoftwire: These labels feel confusing to me - these are the collection start dates not the collection start years, maybe just `current/next collection start date` would be clearer

spec/shared/shared_log_examples.rb
```diff
@@ -2,30 +2,57 @@
...
+            example.run
+          end
+        end
+
+        let(:empty_log) { create(log_type) }
+        let(:in_progress_log) { create(log_type, :in_progress) }
+        let(:completed_log) { create(log_type, :completed) }
+
+        it "is set to not started for an empty #{log_type} log" do
+          expect(empty_log.not_started?).to be(true)
```
natdeanlewissoftwire: Tiny one: can we keep these 4 expectations in the same order in each of the 3 tests for readability?

spec/shared/shared_log_examples.rb
```diff
@@ -2,30 +2,57 @@
...
+        it "is set to in progress for a started #{log_type} log" do
+          expect(in_progress_log.in_progress?).to be(true)
+          expect(in_progress_log.not_started?).to be(false)
+          expect(in_progress_log.completed?).to be(false)
+          expect(in_progress_log.deleted?).to be(false)
+        end
+
+        # if this test starts failing, likely because a new question has been added
+        # and the :completed trait in spec/factories/lettings_log.rb (or sales_log.rb)
+        # needs to be updated to set the database col with that questions answer
```
natdeanlewissoftwire, line 48: Possibly a good idea to add that this step is required to the [docs update PR](https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/3151/changes). Happy to do myself unless that's already in the changes you're going to push?

natdeanlewissoftwire (approved):
lgtm! tiny thoughts only so happy to approve

spec/shared/shared_log_examples.rb
samyou-softwire: yeah I was gonna add it there too, on my list for when I get a mo to look into that PR again 👍

spec/shared/shared_log_examples.rb
samyou-softwire: good spot thanks 😅

spec/shared/shared_log_examples.rb
samyou-softwire: the label that this is used in comes out as "when creating a log for current year", though agree it seems a little contradictory. I've reduced the label to just "current"

oscar-richardson-softwire (approved):
LGTM

PR #3161, opened 2026-02-05
CLDC-4142: New question on gender same as sex - lettings

samyou-softwire:
closes [CLDC-4142](https://mhclgdigital.atlassian.net/browse/CLDC-4142)

the conditional part was surprisingly simple, just needs a field in the parent question

adds some explicit inferring to make sure the description is nil if the gender same as sex question is not answered 2. by default it gets set to "" from the form submission

adds to the BU as well, nothing custom needed here beyond registering the fields

restructures the household characteristics field to remove a lot of the redundant code. as far as I can tell persons 2-7 were duplicates

checked
- form flow
- question number matches paper form
- bulk upload reference file updated & can be uploaded fine
- auto generated bulk upload file updated & can be uploaded fine
- csv download

question yes
question yes
question no
question no
full cya page
full cya page
</details>

[CLDC-4142]: https://mhclgdigital.atlassian.net/browse/CLDC-4142?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/services/bulk_upload/lettings/year2026/csv_parser.rb
```diff
@@ -4,8 +4,8 @@ class BulkUpload::Lettings::Year2026::CsvParser
   include CollectionTimeHelper
 
   # TODO: CLDC-4162: Update when 2026 format is known
-  FIELDS = 129
-  MAX_COLUMNS = 130
+  FIELDS = 153
+  MAX_COLUMNS = 153
```
samyou-softwire: as discussed this will probably change if we remove MAX_COLUMNS

app/services/exports/lettings_log_export_constants.rb
```diff
@@ -207,5 +207,7 @@ module Exports::LettingsLogExportConstants
   POST_2026_EXPORT_FIELDS = Set[]
   (1..8).each do |index|
     POST_2026_EXPORT_FIELDS << "sexrab#{index}"
+    POST_2026_EXPORT_FIELDS << "gender_same_as_sex#{index}"
+    POST_2026_EXPORT_FIELDS << "gender_description#{index}"
```
samyou-softwire: this will probably change as well depending on how we do export fields

spec/services/exports/lettings_log_export_service_spec.rb
```diff
@@ -497,7 +497,39 @@ def replace_duplicate_set_id(export_file)
...
+            address_line1_as_entered: "address line 1 as entered",
+            address_line2_as_entered: "address line 2 as entered",
+            town_or_city_as_entered: "town or city as entered",
+            county_as_entered: "county as entered",
+            postcode_full_as_entered: "AB1 2CD",
+            la_as_entered: "la as entered",
+            manual_address_entry_selected: false,
+            uprn: "1",
+            uprn_known: 1,
+            gender_description2: "Non-binary",
```
samyou-softwire, line 532: the meaningful change is this line, the rest is spreading out this function call across lines

spec/spec_helper.rb
```diff
@@ -141,10 +141,14 @@ def fix_xml_content_order(xml_file_content)
   end
 
   failure_message do |actual|
-    "expected that unordered #{fix_xml_content_order(actual)} would be equal to unordered #{fix_xml_content_order(expected)}"
+    actual_unique_lines = fix_xml_content_order(actual).split("\n") - fix_xml_content_order(expected).split("\n")
+    expected_unique_lines = fix_xml_content_order(expected).split("\n") - fix_xml_content_order(actual).split("\n")
+    "expected that unordered #{fix_xml_content_order(actual)} would be equal to unordered #{fix_xml_content_order(expected)}. Unique lines in actual: #{actual_unique_lines.join("\n")}. Unique lines in expected: #{expected_unique_lines.join("\n")}"
```
samyou-softwire, line 146: was struggling a lot with this matcher (which I wrote!) since it doesn't say what lines were different. updated it so the output is more explicit on whats wrong

app/models/form.rb
```diff
@@ -357,4 +357,12 @@ def start_year_2025_or_later?
   def start_year_2026_or_later?
     start_date && start_date.year >= 2026
   end
+
+  def person_question_count
+    if start_year_2026_or_later?
+      5
+    else
+      4
+    end
```
samyou-softwire, line 366: this change is from [CLDC-4202](https://mhclgdigital.atlassian.net/browse/CLDC-4202) #3169, including it here so I can make the question numbers correct

spec/fixtures/variable_definitions/lettings_download_26_27.csv
```diff
@@ -1,222 +1,3 @@
...
-scheme_intended_stay,Intended length of stay
-scheme_created_at,Date scheme was created
-location_code,Which location is this letting for?
-location_postcode,What is the postcode for this location?
-location_name,What is the name of this location?
-location_units,How many units are at this location?
-location_type_of_unit,What is the most common type of unit at this location?
-location_mobility_type,What are the mobility standards for the majority of the units in this location?
-location_local_authority,What is the local authority of this postcode?
-location_startdate,When did the first property in this location become available under this scheme?
```
samyou-softwire, line 219: we don't need to include labels from previous years so I removed these

app/models/form/lettings/subsections/household_characteristics.rb
```diff
@@ -33,170 +34,42 @@ def pages
...
+      *person_questions(person_index: 3),
+      *person_questions(person_index: 4),
+      *person_questions(person_index: 5),
+      *person_questions(person_index: 6),
+      *person_questions(person_index: 7),
+      *person_questions(person_index: 8),
     ].compact
   end
 
+  def person_questions(person_index:)
```
natdeanlewissoftwire, line 47: 😍

spec/spec_helper.rb
natdeanlewissoftwire: Ah nice, I was copy pasting debug output to get around this 😅

app/models/derived_variables/lettings_log_variables.rb
```diff
@@ -418,6 +420,16 @@ def get_irproduct
     return 3 if rent_type == 5
   end
 
+  def clear_gender_description!
```
natdeanlewissoftwire: Can we rename this so it's clear it's conditional on gender_same_as_sex being 'No'
e.g.
`clear_gender_descriptions_if_gender_same_as_sex!` or similar?
We could probably get rid of the comment then imo but happy if you'd rather keep it

spec/fixtures/variable_definitions/lettings_download_26_27.csv
natdeanlewissoftwire: Ah I think I added these in - It looked to me like previous years had all the fields but agree they're not always needed. Could you clean up sales too so we're consistent going forwards, doesn't matter to me which approach we go for. Or would you rather move that to a sales ticket?

spec/factories/lettings_log.rb
```diff
@@ -73,6 +74,8 @@
       age2 { Faker::Number.within(range: 25..45) }
       sexrab2 { %w[F M R].sample }
       sex2 { %w[F M X R].sample }
+      gender_same_as_sex2 { 2 }
+      gender_description2 { Faker::Gender.type }
```
natdeanlewissoftwire, line 78: cool!

spec/factories/lettings_log.rb
natdeanlewissoftwire, line 44: do we want to add `gender_same_as_sex1` and possibly `gender_description1` here too? Possible the answer is no - but I've added `sexrab1` because `sex1` was there already

natdeanlewissoftwire (commented):
nice!

natdeanlewissoftwire (approved):

spec/fixtures/variable_definitions/lettings_download_26_27.csv
samyou-softwire: ah good point; I'll just update it now

app/models/derived_variables/lettings_log_variables.rb
samyou-softwire: sounds good, I've updated the comment a bit too to detail exactly why we do this as well

spec/factories/lettings_log.rb
samyou-softwire: I think it's in there for duplicate log reasons. suppose we'll update it in [CLDC-4119](https://mhclgdigital.atlassian.net/browse/CLDC-4119). I'll note this on the ticket

oscar-richardson-softwire (approved):
LGTM

PR #3164, opened 2026-02-09
CLDC-4149: Prevten new option

samyou-softwire:
closes [CLDC-4149](https://mhclgdigital.atlassian.net/browse/CLDC-4149)

adds the new option for 2026

updates tests

image

[CLDC-4149]: https://mhclgdigital.atlassian.net/browse/CLDC-4149?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/models/form/lettings/questions/previous_tenure.rb
```diff
@@ -5,7 +5,13 @@ def initialize(id, hsh, page)
     @copy_key = "lettings.household_situation.prevten.not_renewal"
     @type = "radio"
     @check_answers_card_number = 0
-    @answer_options = form.start_year_2025_or_later? ? ANSWER_OPTIONS_2025 : ANSWER_OPTIONS
+    @answer_options = if form.start_year_2026_or_later?
+                        ANSWER_OPTIONS_2026
+                      elsif form.start_year_2025_or_later?
```
natdeanlewissoftwire: Now we've restructured this, would it be more explicit just to have this case be `form.start_date.year == 2025`?

spec/models/form/lettings/questions/previous_tenure_spec.rb
```diff
@@ -56,14 +63,53 @@
...
+        "37" => { "value" => "Host family or similar refugee accommodation" },
+        "25" => { "value" => "Any other accommodation" },
+      })
+    end
+  end
+
+  context "with 2026 logs" do
+    let(:year) { 2026 }
+    let(:start_year_2025_or_later?) { true }
+    let(:start_year_2026_or_later?) { true }
```
natdeanlewissoftwire, line 104: If we need to mock this, can we just mock `form.start_date.year` instead? Not the end of the world tbf

natdeanlewissoftwire (approved):
Looks good, only nit comments really so approving at this point :)

app/models/form/lettings/questions/previous_tenure.rb
samyou-softwire: I think it'd be nicer if this didn't break immediately on 2027 starting hence using the `_or_later?` calls, though I've renamed the variable to make this a bit clearer

spec/models/form/lettings/questions/previous_tenure_spec.rb
samyou-softwire: would be nice, but instance doubles don't allow for any 'real' function calls. everything must be mocked and we must respecify `start_year_2025_or_later?`

natdeanlewissoftwire: Ah okay, I guess I'm not really sure why these are needed - we've set year to 2026, the form starts in that year, why do we need to enforce this at all? the form should return true for these already given year is 2026 right?

EDIT - ignore that, have just been doing some research to remind myself about how ruby instance doubles work, this makes sense now!

app/models/form/lettings/questions/previous_tenure.rb
natdeanlewissoftwire: Ah no I just mean specifically for 2025, `elsif form.start_year_2025_or_later?` could be `elsif form.start_date.year == 2025`

Agree 2026 needs to stay as `form.start_year_2026_or_later?` for use in 2027 onwards.

samyou-softwire: ah sure can tweak this

PR #3165, opened 2026-02-09
CLDC-4150: Update hint text for rent period

samyou-softwire:
closes [CLDC-4150](https://mhclgdigital.atlassian.net/browse/CLDC-4150)

standard new hint text

[CLDC-4150]: https://mhclgdigital.atlassian.net/browse/CLDC-4150?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/config/locales/forms/2026/lettings/income_and_benefits.en.yml
+++ b/config/locales/forms/2026/lettings/income_and_benefits.en.yml
@@ -48,7 +48,7 @@ en:
             page_header: ""
             check_answer_label: "Frequency of household rent and charges"
             check_answer_prompt: ""
-            hint_text: "Select how often the household is charged. This may be different to how often they pay."
+            hint_text: "Select how often the household is charged. This may be different to how often they pay.<br><br>Only options relevant to your organisation are shown below. To change these, go to the 'About your organisation' page or contact our helpdesk using the link at the bottom of this page."
             question_text: "How often does the household pay rent and other charges?"
 
           care_home:


natdeanlewissoftwire (approved):
FWIW using <br> like this is baaaad practice (https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/br)

But I don't think that would justify a) diverging from the rest of the codebase at this point or
b) going through and restructuring now to use p tags 

So pragmatically this is fine by me 👍

samyou-softwire:
> FWIW using like this is baaaad practice (https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/br)
> 
> But I don't think that would justify a) diverging from the rest of the codebase at this point or b) going through and restructuring now to use p tags
> 
> So pragmatically this is fine by me 👍

I've raised [CLDC-4228](https://mhclgdigital.atlassian.net/browse/CLDC-4228) to the backlog to investigate this further

[CLDC-4228]: https://mhclgdigital.atlassian.net/browse/CLDC-4228?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #3167, opened 2026-02-09
CLDC-4155: Update reason for leaving hint text

samyou-softwire:
closes [CLDC-4155](https://mhclgdigital.atlassian.net/browse/CLDC-4155)

updates hint text as required on ticket

q76 reason for leaving hint text

[CLDC-4155]: https://mhclgdigital.atlassian.net/browse/CLDC-4155?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/config/locales/forms/2026/lettings/household_situation.en.yml
+++ b/config/locales/forms/2026/lettings/household_situation.en.yml
@@ -35,7 +35,7 @@ en:
               reason:
                 check_answer_label: "Reason for leaving last settled home"
                 check_answer_prompt: ""
-                hint_text: "The tenant’s ‘last settled home’ is their last long-standing home. For tenants who were in temporary accommodation, sleeping rough or otherwise homeless, their last settled home is where they were living previously."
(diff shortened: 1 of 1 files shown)


natdeanlewissoftwire (commented):
Can we add a test for this, I think one should have already existed but it seems not

samyou-softwire:
> Can we add a test for this, I think one should have already existed but it seems not

do you know examples we have of tests for the hint text? since it's on the page as part of a locale we'd need to render it in a feature test, which I've not really seen much before on CORE

natdeanlewissoftwire:
@samyou-softwire actually we've got a test that covers this in a broad sense in `question_spec.rb`, I think we're okay to continue to rely on that here rather than adding a new suite of tests at this point. Will approve on that basis :)
image

natdeanlewissoftwire (approved):

PR #3168, opened 2026-02-09
CLDC-4200: Add household characteristics household members CYA label

samyou-softwire:
closes [CLDC-4200](https://mhclgdigital.atlassian.net/browse/CLDC-4200)

adds scope in CORE to have a custom label for a CYA section

precise wording TBC

household characteristics cya

doesn't impact 2025 logs or sales

[CLDC-4200]: https://mhclgdigital.atlassian.net/browse/CLDC-4200?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #3172, opened 2026-02-11
CLDC-1980: Add income and benefits CYA label

natdeanlewissoftwire:
Closes [CLDC-1980](https://mhclgdigital.atlassian.net/browse/CLDC-1980)

This copies the behaviour added in CLDC-4200, so that'll disappear from the diff when that's merged in.

Otherwise this just adds an "All buyers" label to the relevant questions.

the only other point of note is that the svaings and deposit value checks are included in the "All buyers" section currently. I've flagged with Rachel as it wasn't clear on the ticket whether this should be the case, but it's my best guess that they should be grouped together.

Screenshot of new "All buyers" label:
image

[CLDC-1980]: https://mhclgdigital.atlassian.net/browse/CLDC-1980?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/components/check_answers_summary_list_card_component.html.erb
+++ b/app/components/check_answers_summary_list_card_component.html.erb
@@ -1,8 +1,8 @@
 <% if applicable_questions.first %>
   <div class="govuk-summary-card govuk-!-margin-bottom-6">
-    <% if applicable_questions.first.check_answers_card_number != 0 && applicable_questions.first.check_answers_card_number.present? %>
+    <% if check_answers_card_label(applicable_questions.first).present? %>
       <div class="govuk-summary-card__title-wrapper">
-        <h3 class="govuk-summary-card__title"><%= check_answers_card_title(applicable_questions.first) %></h3>
+        <h3 class="govuk-summary-card__title"><%= check_answers_card_label(applicable_questions.first) %></h3>
       </div>
     <% end %>
 
--- a/app/components/check_answers_summary_list_card_component.rb
+++ b/app/components/check_answers_summary_list_card_component.rb
@@ -22,7 +22,10 @@ class CheckAnswersSummaryListCardComponent < ViewComponent::Base
     [question.question_number_string, question.check_answer_label.to_s.presence || question.header.to_s].compact.join(" - ")
   end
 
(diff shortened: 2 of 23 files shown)


app/components/check_answers_summary_list_card_component.html.erb
natdeanlewissoftwire, line 1: These changes are from [CLDC-4200](https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/3168), sharing that behaviour here.

app/components/check_answers_summary_list_card_component.rb
natdeanlewissoftwire, line 1: These changes are also from [CLDC-4200](https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/3168), sharing that behaviour here.

app/models/form/sales/questions/deposit_value_check.rb
```diff
@@ -18,6 +18,6 @@ def initialize(id, hsh, page, joint_purchase:)
         },
       ],
     }
-    @check_answers_card_number = 0
+    @check_answers_card_number = 0 unless form.start_year_2026_or_later?
```
oscar-richardson-softwire, line 21: I might be missing something obvious here, but how does this work? I.e., why does not setting a `check_answers_card_number` in 2026 means this ends up in the 'All Buyers' section?

natdeanlewissoftwire: So previously the validations appeared in a separate block of their own on the CYA page, but now we want them to be in the "All buyers" section (confirmed with Rachel on the ticket). So this is just to make sure they aren't considered a separate answer card (and hence aren't displayed outside of the "All buyers section" from 2026 onwards - the actual behaviour for displaying "All buyers" is what I've copied from CLDC-4200, this was just overriding that previously because of the `question.check_answers_card_number != 0` condition

oscar-richardson-softwire (approved):
LGTM

PR #3173, opened 2026-02-11
CLDC-NONE: Fix flaky test

samyou-softwire:
fixes https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/actions/runs/21908583229/job/63255596575

### DIFF
--- a/spec/features/bulk_upload_sales_logs_spec.rb
+++ b/spec/features/bulk_upload_sales_logs_spec.rb
@@ -23,6 +23,7 @@ RSpec.describe "Bulk upload sales log" do
   context "when during crossover period" do
     before do
       Timecop.freeze(2024, 5, 1)
+      Singleton.__init__(FormHandler)
     end
 
     after do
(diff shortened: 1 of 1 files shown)


oscar-richardson-softwire (approved):
LGTM, assuming tests pass

spec/features/bulk_upload_sales_logs_spec.rb
```diff
@@ -23,6 +23,7 @@
   context "when during crossover period" do
     before do
       Timecop.freeze(2024, 5, 1)
+      Singleton.__init__(FormHandler)
```
natdeanlewissoftwire: could we just do
``` 
    before do
      allow(FormHandler.instance).to receive(:sales_in_crossover_period?).and_return(true)
    end
```

like we do in `bulk_upload_lettings_logs_spec.rb` instead? That way I think we should keep testing the latest form rather than stay stuck in 2024

samyou-softwire: good point, I've reworked this file to use this method & not use timecop

natdeanlewissoftwire (approved):

PR #3174, opened 2026-02-12
CLDC-4139: Tenancy type other free text validation

oscar-richardson-softwire:
Closes https://mhclgdigital.atlassian.net/browse/CLDC-4139.


config/locales/forms/2026/lettings/soft_validations.en.yml
```diff
@@ -57,6 +57,16 @@ en:
             title_text: "You told us that the tenant’s main reason for leaving their last settled home was %{reasonother}."
             informative_text: "The reason you have entered looks very similar to one of the existing response categories. Please check the categories and select the appropriate one. If the existing categories are not suitable, please confirm here to move onto the next question."
 
+          tenancyother_value_check:
```
samyou-softwire, line 60: this is the soft validation page we show the user
tenancyother value check page

I find this hard to understand what is being asked of me, most other soft validation pages use wording like "Please check the categories and select the appropriate one." or "Are you sure this person isn’t retired?" rather than a direct command.

feels like the wording here needs a pass, I think when Rachel wrote these it was with the idea that a soft validation was just a single line error, like a hard validation, rather than being a whole page. can this be checked?

config/locales/forms/2026/lettings/soft_validations.en.yml
```diff
@@ -57,6 +57,16 @@ en:
...
+            page_header: ""
+            check_answer_label: "Tenancy other confirmation"
+            check_answer_prompt: "Confirm tenancy type"
+            hint_text: ""
+            question_text: ""
+            title_text: "You told us that the type of tenancy was %{tenancyother}."
+            informative_text: "Give the type of tenancy that the tenants rolls onto after any introductory or starter period."
+
+
```
samyou-softwire: there's a double space here
```suggestion

```

samyou-softwire (changes requested):

config/locales/forms/2026/lettings/soft_validations.en.yml
oscar-richardson-softwire: Updated as per Rachel's comments on the ticket:

{8A3ABB4E-32C4-4894-AB2B-8B4FD26A9441}

Note Q27 now shows under 'Make sure these answers are correct', as requested

samyou-softwire (approved):
LGTM, had a little scare as I noticed that this soft validation shows no matter what value you put for Q27 'is it a starter' but reading the soft validation again I think this is on purpose, the request is to say that the user should never write 'starter' under any circumstances because we already know if its a starter or not

this how you understand it too?

oscar-richardson-softwire:
> LGTM, had a little scare as I noticed that this soft validation shows no matter what value you put for Q27 'is it a starter' but reading the soft validation again I think this is on purpose, the request is to say that the user should never write 'starter' under any circumstances because we already know if its a starter or not
> 
> this how you understand it too?



> LGTM, had a little scare as I noticed that this soft validation shows no matter what value you put for Q27 'is it a starter' but reading the soft validation again I think this is on purpose, the request is to say that the user should never write 'starter' under any circumstances because we already know if its a starter or not
> 
> this how you understand it too?

Yeah that's how I understand it

samyou-softwire:
happy for this to go to PO then 👍

PR #3175, opened 2026-02-12
Bump faraday version

oscar-richardson-softwire:


Gemfile
```diff
@@ -117,3 +117,6 @@ gem "cssbundling-rails"
 gem "tzinfo-data", platforms: %i[mingw mswin x64_mingw jruby]
 
 gem "excon", "~> 0.111.0"
+
+# faraday is a transitive dependency, but we need to force the version to 2.14.1 to avoid a vulnerability.
+gem "faraday", ">= 2.14.1"
```
oscar-richardson-softwire, line 122: A quick google seemed to suggest this is the preferred way to handle this situation

config/initializers/ostruct.rb
samyou-softwire, line 1: see https://guides.rubyonrails.org/v2.3/configuring.html#using-initializers and https://github.com/ruby/ostruct

samyou-softwire (approved):
looks to deploy correctly now after requiring ostruct

PR #3176, opened 2026-02-12
CLDC-4141: use prefers not to say text in cya page

natdeanlewissoftwire:
Final tweak for this ticket after getting confirmation from Rachel, we now uses the generic "Prefers not to say" text in all CYA pages (and csv downloads) rather than "Tenant/Person prefers not to say".

Screenshots:
image

image


app/services/csv/lettings_log_csv_service.rb
natdeanlewissoftwire, line 1: This wasn't previously using the inferred refused labels when the refused code was actually given for various person fields, only when the details weren't known at all. This seems inconsistent to me so I've made updates to fix that here.

samyou-softwire: I think this is on purpose, it's a specific override case to say that we should use this copy if the details of the person are set to unknown, regardless of the specific value (which assumedly is nil anyway)

if we want to customise the label based on its value, there are other paths around this file that can do that. I think most flexible is `question#label_from_value`, which is used in most places to generate a display label. this includes the CYA page, so by setting this up we don't need `@inferred_check_answers_value`. for instance in `person_sex_registered_at_birth.rb`
```ruby
def label_from_value(value, _log = nil, _user = nil)
  return "Prefers not to say" if value == "R"

  super
end
```

this does mean the other person fields won't be consistent, which feels like an oversight. perhaps this can be future work? probably worth checking. in any case they're not consistent as it stands right now as the others don't have the CYA override

it looks like the sales person questions are consistent and use "Prefers not to say"

spec/fixtures/files/lettings_log_csv_export_labels_25.csv
samyou-softwire, line 1: think in general without explicitly clearing we shouldn't be making any changes to old years' function

samyou-softwire (changes requested):
feels like we've found an area of CORE that potentially isn't working quite right

app/services/csv/lettings_log_csv_service.rb
natdeanlewissoftwire: Yep agree the override case is intentional - what I think isn't intentional is that we're not overriding what's included in the csv download when R is actually selected, like we do in the CYA, which is what Rachel has confirmed (though this isn't currently the case for all qs as you mention).

`label_from_value` sounds like a good way of doing this from your description. But agree - I guess what I'm really asking [here](https://softwire.slack.com/archives/C0A0V75PAAY/p1770903975356849) is whether it's more pragmatic to include this or not when I put this PR up for review, feels a shame not to take the chance to fix it now but happy if the standard approach is to prioritise not changing previous years when we see things like this?

samyou-softwire: I think we should either
- update only this question to be consistent between CYA and referral exporter
- update all person questions to be consistent for all years

as these mean the logic is at least consistent specific to the question

I think the other questions not working correctly is a bug but we'll want to get clarity from CORE on what they actually want to happen here, which is probably not worth delaying this ticket over. could another ticket be raised?

natdeanlewissoftwire: Agree! What's the process for raising bugs with CORE? Rachel again? Or do I make a ticket and pass to Joana to get prioritisation?

samyou-softwire: we're fine to just raise them yeah. can pass by Joana so we can get this in the refinement pipeline

natdeanlewissoftwire: Sweet, will do, cheers
ticket: https://mhclgdigital.atlassian.net/browse/CLDC-4230

samyou-softwire (approved):
LGTM

oscar-richardson-softwire (approved):
LGTM

PR #3177, opened 2026-02-12
CLDC-4140: remove old gender question lettings

natdeanlewissoftwire:
Closes CLDC-4140.

This PR hides the old gender question (`sex1` etc.) from 2026/27 logs. 

Also included is an update to the `refused` behaviour to include `sexrab` in the place of `sex` in future years.


app/models/form/question.rb
```diff
@@ -441,6 +441,14 @@ def get_question_number_from_hash(hash)
...
     sex7: %w[R],
     sex8: %w[R],
+    sexrab1: %w[R],
+    sexrab2: %w[R],
+    sexrab3: %w[R],
+    sexrab4: %w[R],
+    sexrab5: %w[R],
+    sexrab6: %w[R],
+    sexrab7: %w[R],
+    sexrab8: %w[R],
```
samyou-softwire, line 451: do we know what these do? I can see this is used in `question#value_is_refused?` but then the trail goes cold

app/services/exports/lettings_log_export_constants.rb
```diff
@@ -235,7 +249,6 @@ module Exports::LettingsLogExportConstants
     "referral_noms",
     "referral_org",
   ]
-
```
samyou-softwire, line 238: I think this is fine to stay?

spec/services/exports/lettings_log_export_service_spec.rb
```diff
@@ -83,7 +83,7 @@ def replace_record_number(export_template, record_number)
     end
 
     context "and one lettings log is available for export" do
-      let!(:lettings_log) { FactoryBot.create(:lettings_log, :completed, assigned_to: user, age1: 35, sex1: "F", age2: 32, sex2: "M", propcode: "123", ppostcode_full: "SE2 6RT", postcode_full: "NW1 5TY", town_or_city: "London", tenancycode: "BZ737", startdate: Time.zone.local(2022, 2, 2, 10, 36, 49), voiddate: Time.zone.local(2019, 11, 3), mrcdate: Time.zone.local(2020, 5, 5, 10, 36, 49), tenancylength: 5, underoccupation_benefitcap: 4) }
+      let!(:lettings_log) { FactoryBot.create(:lettings_log, :completed, assigned_to: user, age1: 35, sex1: "F", sexrab1: nil, age2: 32, sex2: "M", sexrab2: nil, propcode: "123", ppostcode_full: "SE2 6RT", postcode_full: "NW1 5TY", town_or_city: "London", tenancycode: "BZ737", startdate: Time.zone.local(2022, 2, 2, 10, 36, 49), voiddate: Time.zone.local(2019, 11, 3), mrcdate: Time.zone.local(2020, 5, 5, 10, 36, 49), tenancylength: 5, underoccupation_benefitcap: 4) }
```
samyou-softwire, line 86: could we have answers filled in here?

do we need the previous sex question as well?

app/models/derived_variables/lettings_log_variables.rb
```diff
@@ -300,7 +300,7 @@ def get_totadult
   end
 
   def get_refused
-    return 1 if details_unknown? || age_refused? || sex_refused? || relat_refused? || ecstat_refused?
+    return 1 if details_unknown? || age_refused? || sex_refused? || sexrab_refused? || relat_refused? || ecstat_refused?
```
natdeanlewissoftwire, line 303: I think this change (and the related changes below) sort of fell through the gaps between the sexrab question and the validations updates, so I've added it here. The impact of it is that now refused is correctly set when sexrab is refused, not just sex

app/services/exports/lettings_log_export_constants.rb
natdeanlewissoftwire: I've removed for consistency with the above, just to group changes by year visually. Pretty subjective though, do you think it's more standard/readable in the context of this particular codebase to separate all of these subsections rather than grouping by year?

spec/services/exports/lettings_log_export_service_spec.rb
natdeanlewissoftwire: 1. Not really, these tests are for past year logs (2022), so I don't think we want to include sexrab there. This log is used to generate the expected output, which won't include sexrab. However, it's still useful to have the new fields in the factory for newly created logs (and override them here) imo, rather than having separate factories for separate years as that would be a much larger refactor.

2. Similarly , yes we do want the sex questions as these are pre-2026 exports - they're specified here to override the random value that is given in the factory to make sure the output is consistent.

app/helpers/bulk_upload/lettings_log_to_csv.rb
```diff
@@ -240,33 +240,25 @@ def to_2026_row
...
       log.gender_description6,
-      log.gender_same_as_sex7, # 150
+      log.gender_same_as_sex7,
       log.gender_description7,
       log.gender_same_as_sex8,
       log.gender_description8,
       log.owning_organisation.prp? ? log.referral_register : nil,
       log.referral_noms,
-      log.referral_org, # 156
+      log.referral_org, # 148
```
oscar-richardson-softwire, line 262: Nitpick but we've ended up with some weird spacing from `#120` onwards

app/services/bulk_upload/lettings/year2026/row_parser.rb
```diff
@@ -328,18 +320,10 @@ class BulkUpload::Lettings::Year2026::RowParser
   attribute :field_143, :string
   attribute :field_144, :integer
   attribute :field_145, :string
```
oscar-richardson-softwire: Double check these, doesn't look consistent

app/services/bulk_upload/lettings/year2026/row_parser.rb
```diff
@@ -1255,30 +1236,22 @@ def field_mapping_for_errors
...
+      gender_same_as_sex7: %i[field_136],
+      gender_same_as_sex8: %i[field_137],
+      gender_description1: %i[field_138],
+      gender_description2: %i[field_139],
+      gender_description3: %i[field_140],
+      gender_description4: %i[field_141],
+      gender_description5: %i[field_142],
+      gender_description6: %i[field_143],
+      gender_description7: %i[field_144],
+      gender_description8: %i[field_145],
```
oscar-richardson-softwire, line 1254: This looks wrong, looks like we're alternating between `gender_same_as_sex` and `gender_description` everywhere else but this has them sequentially

docs/Gemfile.lock
```diff
@@ -216,11 +216,11 @@ GEM
...
     net-http (0.9.1)
       uri (>= 0.11.1)
-    nokogiri (1.18.9-arm64-darwin)
+    nokogiri (1.19.1-arm64-darwin)
       racc (~> 1.4)
-    nokogiri (1.18.9-x86_64-darwin)
+    nokogiri (1.19.1-x86_64-darwin)
       racc (~> 1.4)
-    nokogiri (1.18.9-x86_64-linux-gnu)
+    nokogiri (1.19.1-x86_64-linux-gnu)
```
oscar-richardson-softwire: Why this bump?

spec/services/exports/lettings_log_export_service_spec.rb
```diff
@@ -127,7 +127,7 @@ def replace_record_number(export_template, record_number)
     end
 
     context "and one lettings log with unknown user details is available for export" do
-      let!(:lettings_log) { FactoryBot.create(:lettings_log, :completed, details_known_2: 1, assigned_to: user, age1: 35, sex1: "F", propcode: "123", ppostcode_full: "SE2 6RT", postcode_full: "NW1 5TY", town_or_city: "London", tenancycode: "BZ737", startdate: Time.zone.local(2022, 2, 2, 10, 36, 49), voiddate: Time.zone.local(2019, 11, 3), mrcdate: Time.zone.local(2020, 5, 5, 10, 36, 49), tenancylength: 5, underoccupation_benefitcap: 4) }
+      let!(:lettings_log) { FactoryBot.create(:lettings_log, :completed, details_known_2: 1, assigned_to: user, age1: 35, sex1: "F", sexrab1: nil, propcode: "123", ppostcode_full: "SE2 6RT", postcode_full: "NW1 5TY", town_or_city: "London", tenancycode: "BZ737", startdate: Time.zone.local(2022, 2, 2, 10, 36, 49), voiddate: Time.zone.local(2019, 11, 3), mrcdate: Time.zone.local(2020, 5, 5, 10, 36, 49), tenancylength: 5, underoccupation_benefitcap: 4) }
```
oscar-richardson-softwire, line 130: Do we not need to override `sexrab2` as `nil` here?

app/services/exports/lettings_log_export_constants.rb
samyou-softwire: ah apologies I see now, agree looks better to remove this line

spec/services/exports/lettings_log_export_service_spec.rb
samyou-softwire: I see, though I suppose I don't see the merit of overriding `sexrab1: nil` here, the tests for 2025 and before shouldn't export sexrab so setting them to nil is slightly more accurate but doesn't change how the test runs

spec/services/bulk_upload/lettings/year2026/row_parser_spec.rb
```diff
@@ -2084,22 +2076,22 @@
...
+          expect(parser.errors.where(:field_132, category: :soft_validation)).to be_present
         end
 
         it "does not add errors to fields that are not routed to" do
           parser.valid?
-          expect(parser.errors.where(:field_133, category: :soft_validation)).not_to be_present
-          expect(parser.errors.where(:field_144, category: :soft_validation)).not_to be_present
-          expect(parser.errors.where(:field_134, category: :soft_validation)).not_to be_present
-          expect(parser.errors.where(:field_146, category: :soft_validation)).not_to be_present
+          expect(parser.errors.where(:field_157, category: :soft_validation)).not_to be_present
```
oscar-richardson-softwire:
```suggestion
          expect(parser.errors.where(:field_57, category: :soft_validation)).not_to be_present
```

spec/services/bulk_upload/lettings/year2026/row_parser_spec.rb
```diff
@@ -2084,22 +2076,22 @@
...
+          expect(parser.errors.where(:field_136, category: :soft_validation)).not_to be_present
+          expect(parser.errors.where(:field_161, category: :soft_validation)).not_to be_present
```
oscar-richardson-softwire:
```suggestion
          expect(parser.errors.where(:field_61, category: :soft_validation)).not_to be_present
```

app/helpers/bulk_upload/lettings_log_to_csv.rb
natdeanlewissoftwire: No good spot, I thought these were grouped by category rather than just tens. Fixed!

app/models/form/question.rb
natdeanlewissoftwire: Ah I forgot to comment here - I think this is actually entirely stale code, but I didn't want to _not_ update in case it's called by some magic ruby helper function I'm not aware of. I also can't see it actually being used anywhere though. Happy for me to make a tech debt ticket to investigate later? This doesn't feel like priority to clean up at this point.

spec/services/exports/lettings_log_export_service_spec.rb
natdeanlewissoftwire: You're right that `sexrab` is not exported - but we still need to explicitly set them as nil here, because otherwise the _expected_ output will include values for sexrab. The creation of expected outputs is not year dependent in its current form.

spec/services/exports/lettings_log_export_service_spec.rb
natdeanlewissoftwire: Nope! `details_known_2` being 1 ("No" 🙃 ) means all the person 2 values (`sex2` and `sexrab2` included) are nil regardless

docs/Gemfile.lock
natdeanlewissoftwire: I'm not actually sure which commit this was from, but regardless these changes aren't in the latest diff - and they definitely shouldn't be! So I think we're good :)

app/services/bulk_upload/lettings/year2026/row_parser.rb
natdeanlewissoftwire: Ah yes good spot, I'll fix this (and the other field number updates) when I update to the latest main

natdeanlewissoftwire: Done!

app/services/bulk_upload/lettings/year2026/row_parser.rb
natdeanlewissoftwire: Good spot, done :)

app/models/form/question.rb
oscar-richardson-softwire: That sounds good to me @natdeanlewissoftwire, please could you raise a ticket for that

docs/Gemfile.lock
oscar-richardson-softwire: Ah sorry my bad

oscar-richardson-softwire (approved):
LGTM

app/models/form/question.rb
samyou-softwire: @natdeanlewissoftwire looking through this diff again - was this done?

PR #3178, opened 2026-02-13
CLDC-4177: use prefers not to say text in cya page

natdeanlewissoftwire:
Final tweak for this ticket after getting confirmation from Rachel, we now uses the generic "Prefers not to say" text in all CYA pages (and csv downloads) rather than "Tenant/Person prefers not to say".

Screenshots:
image

image

image

image


oscar-richardson-softwire (approved):
LGTM

PR #3179, opened 2026-02-13
CLDC-4119: Amend lettings duplicate log check

samyou-softwire:
closes [CLDC-4119](https://mhclgdigital.atlassian.net/browse/CLDC-4119)

adding this logic to the bulk upload parser was fairly simple as we just pass different fields and it'll automatically only run when adding 2026 logs. no need to consider how it interacts with previous years as they will already not match those as they'll have different start dates

the main site logic is more tricky. essentially:
- we want to update it to no longer duplicate check on location for 2026 logs. two logs with the same address but different location should still match. this would happen if one location was nil and the other wasn't
- in previous years we would ignore address for support logs and only use location. overrides in place mean in 2026 location is always ignored and address is always used
- this would change the duplicate log checking for 2025 to now use address, but I think this is fine as the original ticket was to fix duplicate log check not using this
- we can just add the new question, sexrab1, for free. this is because all previous years will have it as nil so it won't impact previous years

to write this in another way:
- 2025
  - now check the necessary address fields
  - no other changes
- 2026
  - always check the address fields regardless of log type
  - introduce sexrab1
  - no other changes

if what I've done here seems to contradict the above, let me know!

also needs some changes to the duplicate logs display page to make sure it's showing the right fields, see below screenshots

<details>
2025 with address search
2025 with address search

2025 with manual address
2025 with manual address

2026 with address search
2026 with address search

2026 with manual address
2026 with manual address
</details>

[CLDC-4119]: https://mhclgdigital.atlassian.net/browse/CLDC-4119?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/models/lettings_log.rb
```diff
@@ -99,33 +108,62 @@ class LettingsLog < Log
...
+        "COUNT(*) > 1",
+      )
 
     if assigned_to_id
       scope = scope.having("MAX(CASE WHEN assigned_to_id = ? THEN 1 ELSE 0 END) >= 1", assigned_to_id)
     end
     scope.pluck("ARRAY_AGG(id)")
   }
 
+  scope :duplicate_sets_2026_and_later, lambda { |assigned_to_id = nil|
```
samyou-softwire, line 141: this isn't the best but I've found combining scopes doesn't normally work very well, as they have different shapes. happy to combine if we can find a neat way.

this needs to be a different section as 2026 logs specifically need to never check the location

oscar-richardson-softwire:
As noted in https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/3161#discussion_r2793841426, do we want to update the factories as part of this ticket?

oscar-richardson-softwire:
I believe there's some more BU logic/tests that need to be updated as part of this ticket, if you do a search for `TODO: CLDC-4119` that should bring them up

samyou-softwire:
> As noted in [#3161 (comment)](https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/3161#discussion_r2793841426), do we want to update the factories as part of this ticket?

have done so

> I believe there's some more BU logic/tests that need to be updated as part of this ticket, if you do a search for `TODO: CLDC-4119` that should bring them up

took a look and resolved the remaining `CLDC-4119` comments

also updated the ticket as I've now accounted for the duplicate log display page. see updated code in d5f1dd6

app/services/bulk_upload/lettings/year2026/row_parser.rb
```diff
@@ -993,11 +993,9 @@ def validate_if_log_already_exists
       errors.add(:field_9, error_message) # startdate
       errors.add(:field_10, error_message) # startdate
       errors.add(:field_13, error_message) # tenancycode
-      errors.add(:field_6, error_message) if !general_needs? && :field_6.present? # location # TODO: CLDC-4119: remove location from error fields
-      errors.add(:field_5, error_message) if !general_needs? && :field_6.blank? # add to Scheme field as unclear whether log uses New or Old CORE ids
```
oscar-richardson-softwire, line 1004: I think we still want to add an error to the scheme field if the location is blank

app/services/bulk_upload/lettings/year2026/row_parser.rb
```diff
@@ -993,11 +993,9 @@ def validate_if_log_already_exists
...
-      errors.add(:field_23, error_message) unless supported_housing? # postcode_full # TODO: CLDC-4119: add postcode to error fields for supported housing
-      errors.add(:field_24, error_message) unless supported_housing? # postcode_full # TODO: CLDC-4119: add postcode to error fields for supported housing
-      errors.add(:field_25, error_message) unless supported_housing? # la # TODO: CLDC-4119: add LA to error fields for supported housing
+      errors.add(:field_23, error_message) # postcode_full
+      errors.add(:field_24, error_message) # postcode_full
+      errors.add(:field_25, error_message) # la
       errors.add(:field_42, error_message) # age1
       errors.add(:field_130, error_message) # sexrab1
       errors.add(:field_43, error_message) # sex1
```
oscar-richardson-softwire: I think we want to remove `sex1` now it's not used in the duplicate checks

app/services/bulk_upload/lettings/year2026/row_parser.rb
```diff
@@ -993,11 +993,9 @@ def validate_if_log_already_exists
...
       errors.add(:field_10, error_message) # startdate
       errors.add(:field_13, error_message) # tenancycode
-      errors.add(:field_6, error_message) if !general_needs? && :field_6.present? # location # TODO: CLDC-4119: remove location from error fields
-      errors.add(:field_5, error_message) if !general_needs? && :field_6.blank? # add to Scheme field as unclear whether log uses New or Old CORE ids
-      errors.add(:field_23, error_message) unless supported_housing? # postcode_full # TODO: CLDC-4119: add postcode to error fields for supported housing
-      errors.add(:field_24, error_message) unless supported_housing? # postcode_full # TODO: CLDC-4119: add postcode to error fields for supported housing
-      errors.add(:field_25, error_message) unless supported_housing? # la # TODO: CLDC-4119: add LA to error fields for supported housing
+      errors.add(:field_23, error_message) # postcode_full
+      errors.add(:field_24, error_message) # postcode_full
+      errors.add(:field_25, error_message) # la
```
oscar-richardson-softwire, line 1009: We probably want to add errors to the `address_line1` and `uprn` fields too

spec/services/bulk_upload/lettings/year2026/row_parser_spec.rb
```diff
@@ -368,141 +370,10 @@
...
-                expect(parser.errors[field]).to include(error_message)
-              end
-
-              expect(parser.errors[:field_23]).not_to include(error_message)
-              expect(parser.errors[:field_24]).not_to include(error_message)
-              expect(parser.errors[:field_25]).not_to include(error_message)
-            end
-          end
-
-          context "with old core scheme and location ids" do
```
oscar-richardson-softwire, line 387: Are we confident we no longer need this test case and the one below it? Do we know why these were added?

spec/factories/lettings_log.rb
```diff
@@ -35,6 +35,7 @@
       setup_completed
       status { 1 }
       tenancycode { "same tenancy code" }
+      address_line1 { "same address line 1" }
```
oscar-richardson-softwire, line 38: Should we add uprn as well?

app/helpers/duplicate_logs_helper.rb
```diff
@@ -73,16 +75,26 @@ def duplicate_log_answer_label(question, log)
...
-    if question.id == "uprn"
+    if log.form.start_year_2026_or_later?
+      case question.id
+      when "uprn"
+        value = [
+          log.address_line1,
+          log.postcode_full,
+        ].select(&:present?)
+
+        "\n\n#{value.join("\n")}"
```
oscar-richardson-softwire: Struggling to understand precisely what this function is used for and why we want this new format, could you explain please?

app/helpers/duplicate_logs_helper.rb
```diff
@@ -73,16 +75,26 @@ def duplicate_log_answer_label(question, log)
...
+      else
+        question.get_extra_check_answer_value(log)
+      end
+    elsif question.id == "uprn" && !log.form.start_year_2026_or_later?
```
oscar-richardson-softwire: Also I think we now don't return anything if `question.id != "uprn"` and `!log.form.start_year_2026_or_later?`, not sure this is what we want

app/models/lettings_log.rb
```diff
@@ -99,33 +108,62 @@ class LettingsLog < Log
...
+    scope = visible
+      .filter_by_year_or_later(2026)
+      # separate function as location needs to be fully ignored in 2026
+      .group(*DUPLICATE_LOG_ATTRIBUTES, :postcode_full, :uprn, :address_line1)
+      .where.not(startdate: nil)
+      .where.not(sex1: nil)
```
oscar-richardson-softwire: Should this not use `sexrab1`?

app/models/lettings_log.rb
```diff
@@ -85,9 +88,15 @@ class LettingsLog < Log
   scope :age1_answered, -> { where.not(age1: nil).or(where(age1_known: 1)) }
   scope :tcharge_answered, -> { where.not(tcharge: nil).or(where(household_charge: 1)).or(where(is_carehome: 1)) }
   scope :chcharge_answered, -> { where.not(chcharge: nil).or(where(is_carehome: [nil, 0])) }
-  scope :location_for_log_answered, ->(log) { where(location_id: log.location_id).or(where(needstype: 1)) }
-  scope :postcode_for_log_answered, ->(log) { where(postcode_full: log.postcode_full).or(where(needstype: 2)) }
-  scope :location_answered, -> { where.not(location_id: nil).or(where(needstype: 1)) }
+  # once 2025 logs are removed this logic can be simplified
+  # pre 2025, match on location if supported, or address if general needs
+  # post 2026, match on address only
```
oscar-richardson-softwire:
```suggestion
  # 2025 and before, match on location if supported, or address if general needs
  # 2026 and after, match on address only
```

lib/tasks/handle_unpended_logs.rake
oscar-richardson-softwire: Should this be `sexrab1` when >= 2026?

app/services/bulk_upload/lettings/year2026/row_parser.rb
samyou-softwire: this method is around adding errors if there was a duplicate, so even if location is blank we shouldn't add an error to scheme as the scheme wasn't used for determining if the log was a dupe

app/services/bulk_upload/lettings/year2026/row_parser.rb
samyou-softwire: ah yup good spot

app/services/bulk_upload/lettings/year2026/row_parser.rb
samyou-softwire: agree, have added these

spec/services/bulk_upload/lettings/year2026/row_parser_spec.rb
samyou-softwire: ah apologies would've been better to comment here

these ones don't seem to be adding much, checking the duplicate checks around entering different kinds of scheme & location ID. we not longer check what these are set to so these tests won't validate anything of use

spec/factories/lettings_log.rb
samyou-softwire: I think safer not too, all the dupe checks account for a UPRN not being found. I'll push a commit that adds a UPRN but may revert if tests start failing

app/helpers/duplicate_logs_helper.rb
samyou-softwire: ah yeah, this is because its based off of `get_extra_check_answer_value` in `address_search.rb`
```ruby
value = [
  log.address_line1,
  log.address_line2,
  log.town_or_city,
  log.county,
  log.postcode_full,
  (LocalAuthority.find_by(code: log.la)&.name if log.la.present?),
].select(&:present?)
```

but agree out of that context it doesn't need all this. I've refactored

app/helpers/duplicate_logs_helper.rb
samyou-softwire: I've refactored this function to be clearer on the code paths

app/models/lettings_log.rb
samyou-softwire: ah yup will need to be

app/models/lettings_log.rb
samyou-softwire: thanks, have applied

lib/tasks/handle_unpended_logs.rake
samyou-softwire: good spot, I've added a new scope that accounts for this

oscar-richardson-softwire:
Has failing tests and needs rebasing onto main, but apart from that happy with the changes 👍

samyou-softwire:
as agreed have removed UPRN from duplicate log factory as it's been causing tests to fail

oscar-richardson-softwire (approved):
LGTM

PR #3129, opened 2025-12-18
CLDC-4158: Create initial 26/27 forms - sales and lettings

samyou-softwire:
closes [CLDC-4158](https://mhclgdigital.atlassian.net/browse/CLDC-4158)

re-enables the future form use on pre-prod environments

adds new locales files, copied from 2025

adds deadlines provided by Rachel

adds the vital 2026 helper for use in other tickets

tested locally, can create a 2026 log now

2026 lettings log page

[CLDC-4158]: https://mhclgdigital.atlassian.net/browse/CLDC-4158?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #3180, opened 2026-02-16
CLDC-4231: add create test log buttons for 26/27 logs

natdeanlewissoftwire:
Closes [CLDC-4231](https://mhclgdigital.atlassian.net/browse/CLDC-4231)

Adds new test log buttons for setup only, and completed 26/27 logs for use in mobbing/testing.

Start date is randomly generated within the correct collection year for 26/27 logs using an existing function that does this (`generate_different_date_within_collection_year`)

[CLDC-4231]: https://mhclgdigital.atlassian.net/browse/CLDC-4231?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/components/create_log_actions_component.html.erb
```diff
@@ -14,13 +14,25 @@
         <span class="govuk-body govuk-body-s">These tools can only be seen and used in testing environments.</span>
         <div>
           <%= govuk_button_link_to create_test_log_href, class: "govuk-button" do %>
-            New test log
+            New <%= current_collection_year_label %> test log
+            <svg class="govuk-button__start-icon" xmlns="http://www.w3.org/2000/svg" width="17.5" height="19" viewBox="0 0 33 40" aria-hidden="true" focusable="false">
+              <path fill="currentColor" d="M0 0h13l20 20-20 20H0l20-20z"></path>
+            </svg>
+          <% end %>
+          <%= govuk_button_link_to create_next_year_test_log_href, class: "govuk-button" do %>
```
samyou-softwire: I think it'd make more sense if this button only showed if the future forms feature toggle was enabled; support members in the middle of the year will have no use for the button (and pressing it could have unexpected consequences?)

natdeanlewissoftwire: Sure, nice idea. Will do!

natdeanlewissoftwire: Actually on second thought - this is already restricted by `FeatureToggle.create_test_logs_enabled?` which restricts us to test and review anyway. This has made me realise we probably want to extend this to staging for the mobbing though so I've done that, but either way actual support users won't see it on prod, it's environment specific rather than user role specific.

samyou-softwire: good call with the test log buttons being turned on for staging

in normal times the `allow_future_form_use?` toggle will be disabled, we turned it on for this year to do the build. see https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/3129/changes#diff-5c1b2afd6bd653e926116b3b9d4199a18af93c971d12ad3b4bb01f528a42ee70R3

after we're done we'll set it back to false and it'll no longer be possible to create future logs on any environment, so we should definitely make this button vanish too at the same time. it'll only be on pre-prod environments though could lead to confusion for future devs is the idea

natdeanlewissoftwire: Yeah, sounds smart. Is there a ticket for reverting the `allow_future_form_use?` toggle? If not I can make a ticket to change both feature toggles back at the same time post-mobbing

samyou-softwire: not that I'm aware of, please do 👍

oscar-richardson-softwire (approved):
LGTM, happy for this to go to PO review but can I just check you've raised a ticket for reverting the toggle and hiding the button?

app/components/create_log_actions_component.html.erb
natdeanlewissoftwire: Sweet, here it is https://mhclgdigital.atlassian.net/browse/CLDC-4235

PR #3183, opened 2026-02-17
CLDC-4143: Update gender-related validations

samyou-softwire:
closes [CLDC-4143](https://mhclgdigital.atlassian.net/browse/CLDC-4143)

updates the soft validation for gender and retirement to use the new sexrab and gender_same_as_sex questions for 2026

add a new single soft validation that shows wherever the two previous ones would have. this is since the wording is too vague to be specific to one scenario so is reused for both

adds some new tests

touches both the all male and all females not in age range validations. in 2026 we count anybody that is not definitely male as a potential female. this is different to 2025 which assumed nothing if gender identity was non-binary

soft validation

</details>

[CLDC-4143]: https://mhclgdigital.atlassian.net/browse/CLDC-4143?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/models/validations/soft_validations.rb
```diff
@@ -221,25 +203,52 @@ def at_least_one_working_situation_is_sickness_and_household_sickness_is_no?
...
+      person_in_expected_pregnancy_age_range(n, min, max) && person_is_non_male(n)
     end
   end
 
-  def females_in_the_household?
+  def non_males_in_the_household?
     person_count = hhmemb || 8
 
     (1..person_count).any? do |n|
-      public_send("sex#{n}") == "F" || public_send("sex#{n}").nil?
```
samyou-softwire, line 242: strictly speaking this could be a functional change, but we always verify all sex questions are given before running this check so I think the ` || public_send("sex#{n}").nil?` was unnecessary

app/models/validations/soft_validations.rb
```diff
@@ -221,25 +203,52 @@ def at_least_one_working_situation_is_sickness_and_household_sickness_is_no?
 
 private
 
+  def all_tenants_age_and_gender_information_completed?
```
samyou-softwire, line 210: moved these to private as they are never called outside this class

app/services/bulk_upload/lettings/year2026/row_parser.rb
```diff
@@ -1212,6 +1212,22 @@ def field_mapping_for_errors
       sexrab6: %i[field_135],
       sexrab7: %i[field_136],
       sexrab8: %i[field_137],
+      gender_same_as_sex1: %i[field_138],
```
samyou-softwire, line 1266: think these were missed when we added the question. will make sure its documented

app/models/form/lettings/pages/females_in_soft_age_range_in_pregnant_household_lead_age_value_check.rb
```diff
@@ -19,6 +19,10 @@ def questions
   end
 
   def interruption_screen_question_ids
-    %w[preg_occ sex1 sex2 sex3 sex4 sex5 sex6 sex7 sex8 age1 age2 age3 age4 age5 age6 age7 age8]
+    if form.start_year_2026_or_later?
+      %w[preg_occ age1 sexrab1 gender_same_as_sex1 age2 sexrab2 gender_same_as_sex2 age3 sexrab3 gender_same_as_sex3 age4 sexrab4 gender_same_as_sex4 age5 sexrab5 gender_same_as_sex5 age6 sexrab6 gender_same_as_sex6 age7 sexrab7 gender_same_as_sex7 age8 sexrab8 gender_same_as_sex8]
```
natdeanlewissoftwire: Could we order these by type rather than number  like in the existing array for readability?

natdeanlewissoftwire: We should probably save them as a CONST also given how much they're repeated. Or just add this function to the page model?

app/models/validations/soft_validations.rb
natdeanlewissoftwire: Nice!

app/models/validations/soft_validations.rb
natdeanlewissoftwire: hm, yeah I agree, looks redundant. I think we're okay to trust any future implementation to confirm this before calling `non_males_in_the_household?` if required

app/models/validations/soft_validations.rb
```diff
@@ -249,10 +260,36 @@ def all_male_tenants_in_the_household?
...
+    else
+      public_send("sex#{person_number}") == "M"
     end
   end
 
+  def person_is_non_male(person_number)
+    if form.start_year_2026_or_later?
+      !person_is_male(person_number)
+    else
+      # non male wording only came later, originally was specific to female gender
```
natdeanlewissoftwire: maybe worth making this clearer for future readers even if future collection years make further tweaks, something like:
`the "non-male" wording was introduced in the 26/27 collection year, before that this behaviour was limited to female gender only`

app/models/validations/soft_validations.rb
```diff
@@ -76,29 +76,11 @@ def rent_in_soft_max_range?
...
-    end
-  end
-
-  def all_tenants_gender_information_completed?
-    person_count = hhmemb || 8
-
-    (1..person_count).all? do |n|
-      public_send("sex#{n}").present? && details_known_or_lead_tenant?(n)
-    end
+  def non_males_in_pregnant_household_in_soft_validation_range?
```
natdeanlewissoftwire: You've pluralised `non_males` here, whereas the original method was just `female_...`- is that intentional? I think non_males is actually better tbh, as below:

More to the point, I think this was badly named before - what it really checks is 
`no_non_males_in_pregnant_household_in_soft_validation_age_range?`, maybe that's a better name?

natdeanlewissoftwire (commented):
Looks ok, couple of thoughts - but one larger thing I think this PR should also include updates to the retirement soft validation right? Is that still to come? And possibly some hard validation updates (?)

samyou-softwire:
> Looks ok, couple of thoughts - but one larger thing I think this PR should also include updates to the retirement soft validation right? Is that still to come? And possibly some hard validation updates (?)

@natdeanlewissoftwire as far as I can tell this is all, as mentioned on the ticket the current retirement validation doesn't differ  on gender and I couldn't find any hard validation that reference `sex{n}`

natdeanlewissoftwire:
> > Looks ok, couple of thoughts - but one larger thing I think this PR should also include updates to the retirement soft validation right? Is that still to come? And possibly some hard validation updates (?)
> 
> @natdeanlewissoftwire as far as I can tell this is all, as mentioned on the ticket the current retirement validation doesn't differ on gender and I couldn't find any hard validation that reference `sex{n}`

Ah thanks @samyou-softwire, I'd missed those comments. All good on that front then :)

app/models/form/lettings/pages/females_in_soft_age_range_in_pregnant_household_lead_age_value_check.rb
samyou-softwire: I ended up consolidating the pages which should resolve the latter

on the former I put them in this order as I think it reads better on the site as it'll put the qs in question order, though can double check on the ticket
related answers view

app/models/validations/soft_validations.rb
samyou-softwire: yeah I think the wording here I think is that the 'soft validation range' means 'range where we apply the soft validation' means 'NOT 16-50', though yeah that makes two of us that didn't find this clear

I've updated the wording to not mention the soft validation range

samyou-softwire:
also rebased this onto main

app/models/validations/soft_validations.rb
samyou-softwire: sounds good, have updated to this 👍

samyou-softwire:
no longer draft as rachel has agreed wording

app/models/form/lettings/pages/females_in_soft_age_range_in_pregnant_household_lead_age_value_check.rb
natdeanlewissoftwire: Ah gotcha, yeah that's fair, fine as is then :)

natdeanlewissoftwire (commented):
Nice, those look resolved to me :)

natdeanlewissoftwire (approved):

PR #3185, opened 2026-02-18
CLDC-4142: Add CYA override for gender_same_as_sex

samyou-softwire:
improvement for [CLDC-4142](https://mhclgdigital.atlassian.net/browse/CLDC-4142)

it's clearer for 'No' here as the below question in CYA talks above gender identity description

as requested by client

household characteristics cya page person 2

[CLDC-4142]: https://mhclgdigital.atlassian.net/browse/CLDC-4142?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/models/form/lettings/questions/gender_same_as_sex.rb
+++ b/app/models/form/lettings/questions/gender_same_as_sex.rb
@@ -7,6 +7,7 @@ class Form::Lettings::Questions::GenderSameAsSex < ::Form::Question
     @conditional_for = { "gender_description#{person_index}" => [2] }
     @person_index = person_index
     @question_number = question_number
+    @inferred_check_answers_value = [{ "condition" => { "gender_same_as_sex#{person_index}" => 2 }, "value" => "No" }]
   end
 
   def answer_options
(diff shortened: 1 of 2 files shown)


app/models/form/lettings/questions/gender_same_as_sex.rb
```diff
@@ -7,6 +7,7 @@ def initialize(id, hsh, page, person_index:)
     @conditional_for = { "gender_description#{person_index}" => [2] }
     @person_index = person_index
     @question_number = question_number
+    @inferred_check_answers_value = [{ "condition" => { "gender_same_as_sex#{person_index}" => 2 }, "value" => "No" }]
```
natdeanlewissoftwire, line 10: Do we want to use `label_from_value` like we do for the gender questions, so the "No" text also shows in csv downloads?

samyou-softwire: I think in this case no, as the request here is specific to the CYA page and is about users getting confused about this being a clickable link

in the csv download I think it'll be clear how the questions relate

natdeanlewissoftwire: Good enough for me, will approve ✅

natdeanlewissoftwire (approved):

oscar-richardson-softwire (approved):
LGTM

PR #3186, opened 2026-02-19
CLDC-4173: add new building height question sales

natdeanlewissoftwire:
Closes [CLDC-4173](https://mhclgdigital.atlassian.net/browse/CLDC-4173).

New question has been added for 26/27 logs. Field has been added at the end of the bu field list, will be reordered later as we are doing for lettings.

BU works, CSV download works. I've also added tests for 26/27 sales downloads (including testing the new field), as we didn't have these before.

[CLDC-4173]: https://mhclgdigital.atlassian.net/browse/CLDC-4173?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


spec/services/csv/sales_log_csv_service_spec.rb
natdeanlewissoftwire, line 1: the diff here's a little confusing because I've also reordered the tests so they're in chronological order. They were all mixed up previously. But the only actual change is adding the 2026 tests

spec/fixtures/files/2026_27_sales_bulk_upload.csv
samyou-softwire, line 1: not a change needed, but worth noting that one of the jobs in finalising the BU fields is setting all the titles, descriptions, formats etc correctly. the only necessary field for testing is the BU field number, and suppose the question title is good for sanity

feel free to omit setting the format
format field present in example bu file

app/services/bulk_upload/sales/year2026/row_parser.rb
```diff
@@ -842,6 +845,7 @@ def attributes_for_log
     attributes["sexrab4"] = field_125
     attributes["sexrab5"] = field_126
     attributes["sexrab6"] = field_127
+    attributes["buildheightclass"] = field_128
```
samyou-softwire, line 848: to sanity check, the ticket lists some specific requirements for how BU should work here in relation to Q16. will this code by default fulfil them?

spec/models/form/sales/pages/building_height_class_spec.rb
```diff
@@ -0,0 +1,34 @@
+require "rails_helper"
+
+RSpec.describe Form::Sales::Pages::BuildingHeightClass, type: :model do
+  subject(:page) { described_class.new(page_id, page_definition, subsection) }
+
+  let(:page_id) { nil }
+  let(:page_definition) { nil }
+  let(:subsection) { instance_double(Form::Subsection, form: instance_double(Form, start_date: Time.zone.local(2026, 4, 1))) }
```
samyou-softwire: there's a method in `CollectionTimeHelper` smt like `collection_start_date_for_year()` which we can use instead of specific dates

spec/models/form/sales/questions/building_height_class_spec.rb
```diff
@@ -0,0 +1,37 @@
+require "rails_helper"
+
+RSpec.describe Form::Sales::Questions::BuildingHeightClass, type: :model do
+  subject(:question) { described_class.new(question_id, question_definition, page) }
+
+  let(:question_id) { nil }
+  let(:question_definition) { nil }
+  let(:page) { instance_double(Form::Page, subsection: instance_double(Form::Subsection, form: instance_double(Form, start_date: Time.zone.local(2026, 4, 1)))) }
```
samyou-softwire: also here

spec/models/form/sales/subsections/property_information_spec.rb
```diff
@@ -100,6 +100,36 @@
         )
       end
     end
+
+    context "when 2026" do
+      let(:start_date) { Time.utc(2026, 2, 8) }
+
+      before do
+        allow(form).to receive_messages(start_year_2024_or_later?: true, start_year_2025_or_later?: true, start_year_2026_or_later?: true)
```
samyou-softwire: not super opinionated on this, but when doing tests I've normally had `start_year_2024_or_later?` as let() variables and set them to true or false in the inner contexts rather than setting up the messages again. puts the smallest amount of setup in each inner config

samyou-softwire (approved):
few small comments, LGTM overall

app/services/bulk_upload/sales/year2026/row_parser.rb
natdeanlewissoftwire: Yep, it's all there by default - there's no explicit validation/error message if you enter a `buildheightclass` for an invalid `proptype` in a BU, but the ticket just says that this should be "ignored". The log uploads correctly with a nil value for `buildheightclass` in this case as the question isn't routed to.

spec/models/form/sales/pages/building_height_class_spec.rb
natdeanlewissoftwire: Ah nice I forgot about that thanks, have done :)

spec/models/form/sales/subsections/property_information_spec.rb
natdeanlewissoftwire: Have done this too, good point

PR #3188, opened 2026-02-23
 CLDC-4178: Add gender same as sex question for sales

natdeanlewissoftwire:
Closes [CLDC-4178](https://mhclgdigital.atlassian.net/browse/CLDC-4178).

Adds new fields for gender same as sex and gender description for buyer 1-2 and/or person 2-6.

Adds new pages and questions for gender same as sex and gender decsripition.

Includes these new fields in exports, csv download, bulk upload.

Screenshots:

Example of new page and questions (same for person, buyer 2 etc. with relevant copy changes):
image

CYA:
image
image

CSV download codes and labels:
image
image

[CLDC-4178]: https://mhclgdigital.atlassian.net/browse/CLDC-4178?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/models/derived_variables/sales_log_variables.rb
```diff
@@ -254,6 +256,18 @@ def prevten_was_social_housing?
...
+    # default submit will send a "" for gender description. this ensure it's nil in this case
+    # as well as blanking it if the user writes it in mistakenly in bulk upload
+    (1..6).each do |person_index|
+      gender_same_as_sex = public_send("gender_same_as_sex#{person_index}")
+      if gender_same_as_sex.present? && gender_same_as_sex != 2
+        self["gender_description#{person_index}"] = nil
+      end
+    end
+  end
+
```
samyou-softwire: think these are the same between lettings and sales, or at least almost. can we extract this to `log.rb`?

app/models/form/sales/questions/gender_same_as_sex1.rb
samyou-softwire, line 1: these 3 question files look really very similar to eachother, differing on indexes mostly. I think there is precedent for this on CORE, but I've not been convinced there's any particular advantage to doing it this way vs having a single question file we pass person index into and it figures out the rest. it leads to a lot of duplication otherwise. same for `gender_same_as_sexn`

thoughts on this?

spec/fixtures/exports/sales_log_26_27.xml
```diff
@@ -95,6 +95,18 @@
     <SEXRAB5>M</SEXRAB5>
     <SEXRAB6/>
     <BUILDHEIGHTCLASS>2</BUILDHEIGHTCLASS>
+    <GENDER_SAME_AS_SEX1/>
+    <GENDER_DESCRIPTION1/>
```
samyou-softwire: seems a bit odd that these are all nil. could we add some example values?

spec/models/form/sales/pages/buyer1_gender_same_as_sex_spec.rb
```diff
@@ -0,0 +1,29 @@
+require "rails_helper"
+
+RSpec.describe Form::Sales::Pages::Buyer1GenderSameAsSex, type: :model do
+  subject(:page) { described_class.new(page_id, page_definition, subsection) }
+
+  let(:page_id) { nil }
+  let(:page_definition) { nil }
+  let(:subsection) { instance_double(Form::Subsection, form: instance_double(Form, start_date: Time.zone.local(2026, 4, 1))) }
```
samyou-softwire: if this isn't year specific we should use `current_collection_start_date`. likewise for all the other files

spec/services/bulk_upload/sales/year2026/row_parser_spec.rb
```diff
@@ -119,6 +119,8 @@
       field_126: "R",
       field_127: "R",
       field_128: "1",
+      field_129: "1",
```
samyou-softwire: could we set a gender_description field to test with?

samyou-softwire (changes requested):
LGTM, though think we should consider if all these question/page files can be compacted

app/models/derived_variables/sales_log_variables.rb
natdeanlewissoftwire: Sure! Will just need a switch for the number of people but agree this should be shared.

app/models/form/sales/questions/gender_same_as_sex1.rb
natdeanlewissoftwire: Yeah I was a bit on the fence here when working on this, tbh I'd also prefer this to be shared so as it sounds like we're on the same page there I'll DRY these new files up a bit.

app/models/form/sales/questions/gender_same_as_sex.rb
samyou-softwire, line 1: does seem a little odd to me that the sales questions are `PersonGenderSameAsSex` but the lettings questions are just `GenderSameAsSex`, can we make these consistent?

config/locales/forms/2026/sales/household_characteristics.en.yml
```diff
@@ -144,6 +158,20 @@ en:
               hint_text: "This is the sex that was registered at birth. The next question will ask about the person's gender identity."
               question_text: "What was person 2's sex at birth?"
 
+          gender_same_as_sex2:
+            page_header: ""
+            check_answer_label: "Buyer/Person 2's gender identity same as registered at birth"
```
samyou-softwire: looking around, most of the other questions seem to either have custom copy depending on whether it's a buyer or seller, or just write buyer. feels like we should be doing the former here rather than using a /

spec/factories/sales_log.rb
```diff
@@ -303,6 +311,9 @@
       ethnic_group { 17 }
       sexrab2 { "R" }
       sex2 { "X" }
+      gender_same_as_sex1 { 1 }
```
samyou-softwire: could this go with the person 1 lines?

spec/factories/sales_log.rb
samyou-softwire, line 323: could we get a gender_same_as_sex as 3 down here to test with? seems odd to test just two

samyou-softwire (approved):
last couple comments

app/models/form/sales/questions/gender_same_as_sex.rb
natdeanlewissoftwire: Yeah good shout :)

config/locales/forms/2026/sales/household_characteristics.en.yml
natdeanlewissoftwire: Ah thanks meant to update this, yep will do

spec/factories/sales_log.rb
natdeanlewissoftwire: FWIW this is just the export factory, 3 is tested higher up in the completed log, but I'll add here anyway as it's easily done :)

PR #3192, opened 2026-02-24
CLDC-4239: Widen age range for pregnancy soft validation

samyou-softwire:
closes [CLDC-4239](https://mhclgdigital.atlassian.net/browse/CLDC-4239)

update age range for pregnancy soft validation for 2026

needs to only impact `no_household_member_likely_to_be_pregnant?`, without changing the previous year's `non_males_in_pregnant_household_not_in_pregnancy_range?` soft validation

[CLDC-4239]: https://mhclgdigital.atlassian.net/browse/CLDC-4239?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


spec/models/validations/soft_validations_spec.rb
```diff
@@ -328,31 +328,55 @@
...
+          record.sexrab1 = "F"
+          record.gender_same_as_sex1 = 1
+        end
+
+        it "does not show the interruption screen" do
+          expect(record.no_household_member_likely_to_be_pregnant?).to be false
+        end
+      end
+
+      context "when female tenants are under 13" do
```
oscar-richardson-softwire, line 369: I feel like it would make more sense for these tests to test boundary cases,

i.e., `"when female tenants are 13 or over"` with `record.age1 = 13` and then `"when female tenants are under 13"`, and the same for the upper bound

spec/models/validations/soft_validations_spec.rb
oscar-richardson-softwire:
```suggestion
      context "when non binary tenants are under 13" do
```

oscar-richardson-softwire:
```suggestion
context "when non binary tenants are over 55" do
before do
record.age1 = 56
```

spec/models/validations/soft_validations_spec.rb
samyou-softwire: sounds good to me, I've reworked the tests to test the boundary conditions only

oscar-richardson-softwire: Sorry, could you do this for the 2025 tests as well please

samyou-softwire: have added some boundary tests for 2025 as well

spec/models/validations/soft_validations_spec.rb
oscar-richardson-softwire: Sorry, only just spotted this, could we change this (and the same case for 2026) to something like "when all tenants are male and there are more than 8 household members". Currently it sounds a bit like it's saying all household members are older than 8 years old

samyou-softwire: have updated

oscar-richardson-softwire (approved):
LGTM

PR #3193, opened 2026-02-24
CLDC-4241: Fix empty check values page bug

natdeanlewissoftwire:
Closes [CLDC-4241](https://mhclgdigital.atlassian.net/browse/CLDC-4241).

Turns out this bug only appears when hitting Enter, not clicking the submit button. This is because the "See all related answers" link that appears when there are multiple errors is actually a form submission styled like a link. As it appears earlier in the DOM than the actual submit button, hitting enter when it's present simulates clicking this link rather than the submit button. This is nonstandard and surprising. I initially wanted to fix this by to changing this to actually be a link, but actually the form submission is required to run the validations, which then dislpay all the error messages across all related fields in the next page. So, the cleanest solution to the navigation bug is to add a hidden "Save and continue" submit button above the show related fields submit. As this is a little surprising to someone reading the file for the first time, I've added a comment explaining why the hidden button is there. There is of course no visual change:

image

The only other change is renaming an error class that I came across for consistency, non-functional change.

[CLDC-4241]: https://mhclgdigital.atlassian.net/browse/CLDC-4241?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/controllers/form_controller.rb
```diff
@@ -91,13 +92,14 @@ def show_page
       @pages_with_errors_count = 0
       if @page.routed_to?(@log, current_user) || is_referrer_type?("interruption_screen") || adding_answer_from_check_errors_page?
         if updated_answer_from_check_errors_page?
-          @questions = request.params["related_question_ids"].map { |id| @log.form.get_question(id, @log) }
+          @questions = request.params["related_question_ids"].map { |id| @log.form.get_question(id, @log) }.compact
```
samyou-softwire: to check, why does this change need to be made? doesn't seem directly related to the flash mechanism we've added

natdeanlewissoftwire: Good Q , I think this should have been here before but it's only become apparent since this bug fix (error was thrown on visiting the "See all related answers" page otherwise). Basically we're passing in things to `related_question_ids` that don't necessarily correspond to a question. In this case, `postcode_known` is included but returns nil form `get_question`. So we need compact here to avoid having a nil value in our `@questions` output. To be fair, another solution might be to filter these out at the point of flashing but I personally think that's a bit more bloated. 

i.e.
 ```
        flash[:related_question_ids] = question_ids.select { |id| @log.form.get_question(id, @log).present? }.compact.uniq
```
I'd probably still like to keep the `.compact` on `related_question_ids` anyway for safety though personally.

Happy with either though, do you have a preference?

Comment (samyou-softwire) [reply app/controllers/form_controller.rb]: that sounds fine, though does make me wonder why we are adding errors that don't relate to question IDs, is `postcode_known` from like an old form?

Comment (natdeanlewissoftwire) [reply app/controllers/form_controller.rb]: Hm, I think so - my guess is it was a previous question "do you know the postcode?" - but a (presumably newer) derived method with the same name is used in some location validation logic, possibly elsewhere too. From what I can see it looks like dead code to me, and removing it from the errors (or rather replacing with `postcode_known` which seems like the direct replacement) seems to work...

There are quite a lot of references to it in the codebase, so maybe replacing it with `postcode_full` (possibly after confirming w Rachel that's the right approach if it indeed was an old question) should be a separate ticket?

Comment (samyou-softwire) [reply app/controllers/form_controller.rb]: sounds good to me

Comment (natdeanlewissoftwire) [reply app/controllers/form_controller.rb]: Update to this, it turned out using a link rather than a form submit meant errors weren't passed through, so the new simplest solution is actually just a hidden submit in the DOM. So, no flashing etc. is needed! I've added the new change and a comment as it's a bit surprising, but I'll make a ticket to investigate removing `postcode_known` anyway as that's probably a good idea regardless.

Comment (natdeanlewissoftwire) [reply app/controllers/form_controller.rb]: https://mhclgdigital.atlassian.net/browse/CLDC-4245

File: app/views/form/page.html.erb [hunk 2]
```diff
@@ -76,6 +76,8 @@
       <% end %>
 
       <% if @pages_with_errors_count > 1 %>
+        <%# Hidden form submit ensures pressing Enter triggers "Save and continue" rather than "See all related answers" as the user would expect %>
```
oscar-richardson-softwire:
```suggestion
        <%# Hidden form submit ensures pressing Enter triggers "Save and continue" as the user would expect rather than "See all related answers"%>
```

natdeanlewissoftwire: Agree that's clearer, have committed :)

oscar-richardson-softwire (approved):
LGTM

PR #3194, opened 2026-02-24
CLDC-4166: How many live in property updates

natdeanlewissoftwire:
Closes [CLDC-4166](https://mhclgdigital.atlassian.net/browse/CLDC-4166)

Tweaks some copy and validations relating to `householdmemb`, and importantly doesn't take into account number of buyers when inferring this value. It's now a combined total for the whole household.

[CLDC-4166]: https://mhclgdigital.atlassian.net/browse/CLDC-4166?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


spec/models/form/sales/pages/person_known_spec.rb
```diff
@@ -3,11 +3,13 @@
...
-    let(:page_id) { "person_2_known" }
-    let(:page_definition) { nil }
-    let(:subsection) { instance_double(Form::Subsection) }
-    let(:person_index) { 2 }
+  let(:page_id) { "person_2_known" }
+  let(:page_definition) { nil }
+  let(:person_index) { 2 }
+
+  context "with 2024 form" do
+    let(:form) { Form.new(nil, 2024, [], "sales") }
```
oscar-richardson-softwire: Interesting - for some reason it seems that while the convention in the page tests for lettings is to mock forms, it seems like it lots (but not all) of the equivalent tests for sales we're using real forms. I think personally I lean towards using mock forms here, for better isolation, but please shout if you disagree (and/or know of a reason why some of the sales tests seem to use the approach of using real forms rather than mock ones).

oscar-richardson-softwire:
Needs rebasing

spec/models/form/sales/pages/person_known_spec.rb
natdeanlewissoftwire: Hm, yep from a search through the tests I agree it does seem pretty sporadic what approach different files go for, I'm not aware of any reason we shouldn't just use mocks wherever we can - happy to proceed on that basis :)

spec/services/csv/sales_log_csv_service_spec.rb
```diff
@@ -243,9 +243,21 @@
       let(:now) { collection_start_date_for_year(2026) }
       let(:year) { 2026 }
       let(:fixed_time) { collection_start_date_for_year(2026) }
-
       before do
-        log.update!(nationality_all: 36, manual_address_entry_selected: false, uprn: "1", uprn_known: 1)
+        log.update!(
```
natdeanlewissoftwire, line 256: These updates are required since we've changed how `hholdcount` works - the data for person 4 is cleared when the log with hholdcount 3 is instantiated at the top of the file, so specifying here is required as the 2026 equivalent of `hholdcount` 3  with two buyers is actually `hholdcount` 5

spec/fixtures/files/sales_logs_csv_export_labels_26.csv
natdeanlewissoftwire, line 1: I've noticed as part of these changes that `"Besides the buyer(s), how many other people live or will live in the property?"` is used as question text for the export of `hhtype`, which _seems_ wrong to me. But it also seems to be explicitly overrided to be this, so maybe it is a genuine requirement? Seems like this is happening in real exports rather than a test mismatch to me, maybe it's something we should confirm as it seems surprising

spec/models/form/sales/pages/number_of_others_in_property_spec.rb
```diff
@@ -1,13 +1,19 @@
...
+  include CollectionTimeHelper
+
   subject(:page) { described_class.new(page_id, page_definition, subsection, joint_purchase:) }
 
   let(:page_id) { "number_of_others_in_property" }
   let(:page_definition) { nil }
   let(:joint_purchase) { false }
-  let(:form) { Form.new(nil, 2024, [], "sales") }
-  let(:subsection) { instance_double(Form::Subsection, form:, enabled?: true, depends_on: nil) }
+  let(:form) { instance_double(Form, start_date: current_collection_start_date, start_year_2026_or_later?: true, start_year_2025_or_later?: false, depends_on_met: true) }
```
oscar-richardson-softwire: I don't think `depends_on_met: true` is needed anywhere in this file

oscar-richardson-softwire: Also it seems odd that we set `start_year_2026_or_later?: true, start_year_2025_or_later?: false` here

spec/models/form/sales/pages/number_of_others_in_property_spec.rb
```diff
@@ -1,13 +1,19 @@
...
+  let(:subsection) { instance_double(Form::Subsection, enabled?: true, depends_on: nil) }
```
oscar-richardson-softwire: Don't think we need `enabled?: true` here

spec/models/form/sales/pages/number_of_others_in_property_spec.rb
oscar-richardson-softwire: Could we change this to something more descriptive like `when the page expects a joint purchase`? At the moment this is identical to a context block description nested within it but they represent different things.

spec/models/form/sales/pages/number_of_others_in_property_spec.rb
```diff
@@ -39,6 +45,8 @@
       end
 
       context "with 2024 logs" do
+        let(:form) { instance_double(Form, start_date: collection_start_date_for_year(2024), start_year_2026_or_later?: false, start_year_2025_or_later?: false, depends_on_met: true) }
+
         context "with joint purchase" do
```
oscar-richardson-softwire: Could we change this (and equivalent context-blocks) to something like `when the sales log is a joint purchase`?

spec/models/form/sales/pages/number_of_others_in_property_spec.rb
oscar-richardson-softwire: Everything in this block needs to be changed to `does not route to the page`

spec/models/form/sales/pages/number_of_others_in_property_spec.rb
```diff
@@ -109,7 +117,27 @@
       end
 
       context "with 2025 logs" do
```
oscar-richardson-softwire, line 119: For this and the block below it could we change to more descriptive language (e.g., "and the sales log is (not) a staircasing transaction" rather than "and staircase is `n`"

spec/services/csv/sales_log_csv_service_spec.rb
oscar-richardson-softwire: Understood, but wouldn't it be easier just to change `hholdcount` to 5 in the log at the top of the file?

spec/fixtures/files/sales_logs_csv_export_labels_26.csv
oscar-richardson-softwire: Good spot, yeah could you check with Rachel on the ticket please 👍

spec/services/csv/sales_log_csv_service_spec.rb
natdeanlewissoftwire: Nope, then I'd have to fundamentally change all the tests for previous years, as they'd then need 2 more household members. Would require either expanding the factory which I'd rather not do as it's good to have a balance of filled and not-filled users imo for coverage, or hard coding loads of values for these extra users for previous years just to get the tests to pass again, which also feels a bit risky.

natdeanlewissoftwire:
Nice catches @oscar-richardson-softwire, all actioned!

oscar-richardson-softwire (approved):
LGTM!

PR #3195, opened 2026-02-25
CLDC-4167, CLDC-4168: Add don't know option to mortgage length questions

samyou-softwire:
closes [CLDC-4167](https://mhclgdigital.atlassian.net/browse/CLDC-4167) & [CLDC-4168](https://mhclgdigital.atlassian.net/browse/CLDC-4168) as they are the same question

reworks the mortgage length question to work like the age question where the user is presented with an initial 'do you know' option with a conditional number. this exact mechanism is to be confirmed with CORE

adds another mortgage length page so the user will be either shown the page with just the number box or the conditional one depending on the specifics of the log

updates the BU parser to allow "R" for field_118 or field_90, depending on the flow. adds the specific error message wanted by CORE if the field is R but the buyer was not interviewed.

adds tests

cya page for q84
cya page for q84

q84 with interview
q84 with interview

q84 without interview
q84 without interview

cya page for q108
cya page for q108

q108 with inteview
q108 with inteview

q108 without interview
q108 without interview

</details>

[CLDC-4167]: https://mhclgdigital.atlassian.net/browse/CLDC-4167?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[CLDC-4148]: https://mhclgdigital.atlassian.net/browse/CLDC-4148?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[CLDC-4168]: https://mhclgdigital.atlassian.net/browse/CLDC-4168?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/models/form/sales/questions/mortgage_length_known.rb
```diff
@@ -0,0 +1,24 @@
+class Form::Sales::Questions::MortgageLengthKnown < ::Form::Question
+  def initialize(id, hsh, page, ownershipsch:)
+    super(id, hsh, page)
+    @id = "mortgage_length_known"
+    @type = "radio"
+    @answer_options = ANSWER_OPTIONS
+    @conditional_for = { "mortlen" => [0] }
+    @hidden_in_check_answers = {
```
samyou-softwire, line 8: this makes it so in CYA if mortgage length is given hide this question, there's no reason to show both

app/models/form/question.rb
```diff
@@ -340,7 +340,10 @@ def conditional_on
   end
 
   def evaluate_condition(condition, log)
-    case page.questions.find { |q| q.id == condition[:from] }.type
+    conditional_question = page.questions.find { |q| q.id == condition[:from] }
+    return true unless conditional_question
```
samyou-softwire, line 344: this should be an improvement, previously we'd have an error if a question was on the page but not the conditional question that should show it

without fixing this we'd need a lot of mechanism to ensure the MortgageLengthKnown question is always with MortgageLength

app/services/bulk_upload/sales/year2026/row_parser.rb
samyou-softwire, line 1520: as far as I can see this was added but never validated

app/models/form/sales/questions/mortgage_length_known.rb
natdeanlewissoftwire: Yep nice, this is the standard pattern I've seen elsewhere

lib/tasks/lint.rake
```diff
@@ -3,6 +3,11 @@ task rubocop: :environment do
   sh "bundle exec rubocop"
 end
 
+desc "Run Rubocop Autocorrect"
+task rubocop_autocorrect: :environment do
```
natdeanlewissoftwire, line 7: I've just been using the direct commands for this, but I do think this is nice to have. As long as people aren't using it blindly, agree it shouldn't be included in the lint command or anythign pipeline related ofc

spec/models/form/sales/subsections/discounted_ownership_scheme_spec.rb
```diff
@@ -1,57 +1,21 @@
...
+  include CollectionTimeHelper
+
   subject(:discounted_ownership_scheme) { described_class.new(subsection_id, subsection_definition, section) }
 
   let(:subsection_id) { nil }
   let(:subsection_definition) { nil }
-  let(:form) { instance_double(Form, start_date: Time.zone.local(2024, 4, 1), start_year_2025_or_later?: false) }
+  let(:start_year_2025_or_later?) { true }
+  let(:start_year_2026_or_later?) { true }
+  let(:form) { instance_double(Form, start_date: current_collection_start_date, start_year_2025_or_later?: start_year_2025_or_later?, start_year_2026_or_later?: start_year_2026_or_later?) }
```
natdeanlewissoftwire, line 12: Small one,  I think the more common convention in the codebase is to use 
```
allow(form).to receive(start_year_...).and_return(...)
```
here, do you think we should be using that convention here instead as we're touching this file anyway?

app/models/form/sales/pages/mortgage_length_interviewed.rb
```diff
@@ -0,0 +1,16 @@
+class Form::Sales::Pages::MortgageLengthInterviewed < ::Form::Page
+  def initialize(id, hsh, subsection, ownershipsch:)
+    super(id, hsh, subsection)
+    @ownershipsch = ownershipsch
+    @depends_on = [{
+      "mortgageused" => 1, "buyer_not_interviewed?" => false
```
natdeanlewissoftwire, line 6: My reading of the ticket is that the don't know option should only be allowed it `buyer_not_interviewed?` is true. i.e. you can only say you don't know the mortgage length if you haven't spoken to the buyer.

app/services/bulk_upload/sales/year2026/row_parser.rb
```diff
@@ -1156,6 +1175,16 @@ def mortlen
     field_118 if discounted_ownership?
   end
 
+  def mortgage_length_known
+    return nil if field_14 == 1
```
natdeanlewissoftwire: similarly I think we want this condition to be `field_14 == 2`, have seen a couple of other occurences of this sort of thing also

db/migrate/20260224141705_add_mortgage_length_known_to_sales_logs.rb
```diff
@@ -0,0 +1,5 @@
+class AddMortgageLengthKnownToSalesLogs < ActiveRecord::Migration[7.2]
+  def change
+    add_column :sales_logs, :mortgage_length_known, :integer
```
natdeanlewissoftwire: This is a good name. However, I think it'd actually be clearer to call this the arguably worse name `mortlen_known` to tie it to `mortlen` more explicitly :)

spec/models/form/sales/subsections/discounted_ownership_scheme_spec.rb
natdeanlewissoftwire: Since writing the above comment I've seen you've put a comment on one of my PRs about me doing this sort of thing differently also - I don't massively mind which we go for but I think we'd both prefer a convention - I'm happy with your approach as standard so might just use this style going forwards, sound good?

app/models/sales_log.rb
```diff
@@ -390,6 +390,10 @@ def bedsit_changed_to_not_bedsit?
     proptype_changed? && proptype_was == 2
   end
 
+  def buyer_not_interviewed_changed_to_interviewed_and_mortgage_length_known?
+    noint_changed? && noint_was == 1 && !buyer_not_interviewed? && mortlen.present?
```
natdeanlewissoftwire: Not a massive fan of `!buyer_not_interviewed?`, do you think it's worth just defining a separate `buyer_interviewed?` method?

This might invert anyway tbf pending possibly inverting these conditions as mentioned in other comments

natdeanlewissoftwire (commented):
Nice! Mostly small things, lmk what you think!

Other misc thoughts:

The PR description links to CLDC-4148 rather than CLDC-4168, assume that's just a typo?

Also this approach is waiting confirmation from Sammy Pitch AFAICT from the ticket right?

app/services/bulk_upload/sales/year2026/row_parser.rb
```diff
@@ -1523,14 +1552,10 @@ def valid_nationality_options
     %w[0] + GlobalConstants::COUNTRIES_ANSWER_OPTIONS.keys # 0 is "Prefers not to say"
   end
 
-  def validate_relat_fields
```
natdeanlewissoftwire, line 1531: Did you mean to delete this? It does look like dead code to me, just checking :)

app/models/form/sales/questions/mortgage_length_known.rb
```diff
@@ -0,0 +1,24 @@
...
+        { "mortgage_length_known" => 0 },
+      ],
+    }
+    @question_number = QUESTION_NUMBER_FROM_YEAR_AND_OWNERSHIP.fetch(form.start_date.year, QUESTION_NUMBER_FROM_YEAR_AND_OWNERSHIP.max_by { |k, _v| k }.last)[ownershipsch]
+  end
+
+  ANSWER_OPTIONS = { "0" => { "value" => "Yes" }, "1" => { "value" => "No" } }.freeze
+
+  QUESTION_NUMBER_FROM_YEAR_AND_OWNERSHIP = {
+    2023 => { 1 => 93, 2 => 106, 3 => 114 },
```
natdeanlewissoftwire: I don't think we need question numbers for old years for this q right?

app/models/form/sales/questions/mortgage_length.rb
```diff
@@ -19,5 +19,6 @@ def suffix_label(log)
     2023 => { 1 => 93, 2 => 106, 3 => 114 },
     2024 => { 1 => 94, 2 => 107, 3 => 114 },
     2025 => { 1 => 84, 2 => 108 },
```
natdeanlewissoftwire, line 21: Why don't we need a q number for ownership scheme 3 from 2025 onwards (also applies to `mortage_length_known`)?

spec/models/form/sales/subsections/discounted_ownership_scheme_spec.rb
samyou-softwire: yeah I like it a little more this way; has less noise per test

app/models/form/sales/pages/mortgage_length_interviewed.rb
samyou-softwire: ah really good catch, I had this the wrong way around. I'll go and flip all the conditions

samyou-softwire:
closing to rerun migrations

db/migrate/20260224141705_add_mortgage_length_known_to_sales_logs.rb
samyou-softwire: sounds sensible, have renamed

app/models/sales_log.rb
samyou-softwire: this got inverted yea

app/services/bulk_upload/sales/year2026/row_parser.rb
samyou-softwire: yeah this was on purpose

app/models/form/sales/questions/mortgage_length.rb
samyou-softwire: at a quick look ownership scheme 3 hasn't been used since 2024
`ownership_scheme.rb`
```ruby
def answer_options
  if form.start_year_2025_or_later?
    {
      "1" => { "value" => "Shared Ownership", "hint" => "When the purchaser buys an initial share of up to 75% of the property value and pays rent to the Private Registered Provider (PRP) on the remaining portion, or a subsequent staircasing transaction" },
      "2" => { "value" => "Discounted Ownership" },
    }.freeze
  elsif form.start_year_2024_or_later?
    {
      "1" => { "value" => "Yes - a shared ownership scheme", "hint" => "When the purchaser buys an initial share of up to 75% of the property value and pays rent to the Private Registered Provider (PRP) on the remaining portion, or a subsequent staircasing transaction" },
      "2" => { "value" => "Yes - a discounted ownership scheme" },
      "3" => { "value" => "No - this is an outright or other sale" },
    }.freeze
  else
    {
      "1" => { "value" => "Yes - a shared ownership scheme" },
      "2" => { "value" => "Yes - a discounted ownership scheme" },
      "3" => { "value" => "No - this is an outright or other sale" },
    }.freeze
  end
end
```

app/models/form/sales/questions/mortgage_length_known.rb
samyou-softwire: ah good point, have removed

app/services/bulk_upload/sales/year2026/row_parser.rb
samyou-softwire: have updated this

app/models/form/sales/questions/mortgage_length.rb
natdeanlewissoftwire: Hm, okay - but we're passing in `ownershipsch: 3` in `outright_sale.rb`. I guess those updates aren't necessary? 

Seems like it won't make a difference in functionality, but to avoid confusion I think we should revert the change to add the new mortgage length pages in that subsection if they're not relevant to 2026 logs.

natdeanlewissoftwire (approved):
Nice, this all looks great! Just one other piece of cleanup I think we can do but I'll approve now pending that + test updates you're presumably working on atm :)

samyou-softwire:
need to update migration again

app/models/form/sales/questions/mortgage_length.rb
samyou-softwire: ah good point; I checked if the subsection itself had a year check but yeah it's dependent on ownershipsch of 3 which doesn't appear post 2024

have reverted the changes to `outright_sale.rb` and added a comment

PR #3199, opened 2026-02-25
CLDC-4226: Index logs by org and ID

natdeanlewissoftwire:
Fixes the bug raised in support ticket [CLDC-4226](https://mhclgdigital.atlassian.net/browse/CLDC-4226).

Cause: on prod, for small orgs Postgresql sometimes makes very bad (inefficient) decisions as to how to search for logs. In this case, it was scanning by index from largest to smallest through all 1million+ logs in the database. Our data is unevenly distributed, so this turned out to be significantly worse than other approaches it could take (can take around 40 seconds when it could take around a ms). This PR adds an index that allows a more efficient scan of logs, which should enable a significantly faster loading time for these small orgs.

In the reporting org's case, they have less than 20 logs visible, so the scan has to go through every log in the database before we can render the page, hence the timeouts we've seen.

Not reproducible on staging sadly as the data is totally different in distribution and having searched all possible candidate orgs, none experience this issue. So, this fix should be confirmed to be fixed on prod (by loading e.g. Tamil Housing, the org that repeated this issue).

[CLDC-4226]: https://mhclgdigital.atlassian.net/browse/CLDC-4226?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


oscar-richardson-softwire (approved):
LGTM

PR #3200, opened 2026-02-25
CLDC-4175:  add new service charge question sales

natdeanlewissoftwire:
Closes CLDC-4175.

Adds a new service charge question after the existing one in staircasing logs. Adds appropriate validations for when the new value is actually the same as the old value.

Changes reflected in UI, CSV, BU, export.

New question in UI:
image

Validation error:
image

New question in CYA:
image

New fields in CSV download:
image


config/locales/forms/2026/sales/sale_information.en.yml
```diff
@@ -285,6 +285,19 @@ en:
...
+            has_service_charges_changed:
+              check_answer_label: "Service charge will change"
+              check_answer_prompt: "Tell us if the service charge will change"
+              hint_text: "This includes any charges for day-to-day maintenance and repairs, building insurance, and any contributions to a sinking or reserved fund. It does not include estate management fees."
+              question_text: "Will the service charge change after this staircasing transaction takes place?"
+            new_service_charges:
+              check_answer_label: "New monthly service charges"
+              check_answer_prompt: ""
+              hint_text: ""
+              question_text: "Yes, enter monthly amount:"
```
samyou-softwire: this question text doesn't seem to fit in with other questions


for instance, the previous service charge question has the following


so I would expect something like "Enter monthly amount" or even for consistency "Enter the new total monthly charge"

spec/models/form/sales/pages/service_charge_changed_spec.rb
```diff
@@ -0,0 +1,29 @@
+require "rails_helper"
+
+RSpec.describe Form::Sales::Pages::ServiceChargeChanged, type: :model do
+  subject(:page) { described_class.new(page_id, page_definition, subsection) }
+
+  let(:page_id) { nil }
+  let(:page_definition) { nil }
+  let(:subsection) { instance_double(Form::Subsection, form: instance_double(Form, start_date: Time.zone.local(2026, 4, 1))) }
```
samyou-softwire: we can use `collection_date_for_year(2026)`

spec/models/form/sales/questions/has_service_charges_changed_spec.rb
```diff
@@ -0,0 +1,54 @@
+require "rails_helper"
+
+RSpec.describe Form::Sales::Questions::HasServiceChargesChanged, type: :model do
+  subject(:question) { described_class.new(question_id, question_definition, page) }
+
+  let(:question_id) { nil }
+  let(:question_definition) { nil }
+  let(:subsection) { instance_double(Form::Subsection, form: instance_double(Form, start_date:)) }
+  let(:page) { instance_double(Form::Page, subsection:) }
+  let(:start_date) { Time.utc(2026, 5, 1) }
```
samyou-softwire: same here

spec/models/form/sales/questions/new_service_charges_spec.rb
```diff
@@ -0,0 +1,51 @@
+require "rails_helper"
+
+RSpec.describe Form::Sales::Questions::NewServiceCharges, type: :model do
+  subject(:question) { described_class.new(question_id, question_definition, page) }
+
+  let(:question_id) { nil }
+  let(:question_definition) { nil }
+  let(:subsection) { instance_double(Form::Subsection, form: instance_double(Form, start_date:)) }
+  let(:page) { instance_double(Form::Page, subsection:) }
+  let(:start_date) { Time.utc(2026, 5, 1) }
```
samyou-softwire: and here

spec/models/validations/sales/financial_validations_spec.rb
```diff
@@ -478,4 +478,57 @@
       expect(record.errors).to be_empty
     end
   end
+
+  describe "#validate_newservicecharges_different_from_mscharge" do
```
samyou-softwire, line 482: I think these tests would be better generated with a fixture list, something like
```ruby
[
  [nil, 100, 100, false],
  ...
]
.each |hasservicheschargeschanged, newservicecharges, mscharge, adderrors|
  context "when hasservicechargeschanged is #{hasservicheschargeschanged] ... "
     it "#{ternary for does/does not} add errors"
       ...
     end
  end
end
```

to save repetition and make it a little clearer what each test is doing

spec/services/exports/sales_log_export_service_spec.rb
```diff
@@ -474,16 +468,93 @@ def replace_mscharge_and_shared_ownership_values(export_file)
...
+              export_file.sub!("<MORTLEN1>10</MORTLEN1>", "<MORTLEN1/>")
+              export_file.sub!("<NATIONALITYALL1>826</NATIONALITYALL1>", "<NATIONALITYALL1/>")
+              export_file.sub!("<NATIONALITYALL2>826</NATIONALITYALL2>", "<NATIONALITYALL2/>")
+              export_file.sub!("<PREVOWN>1</PREVOWN>", "<PREVOWN/>")
+              export_file.sub!("<PREVSHARED>2</PREVSHARED>", "<PREVSHARED/>")
+              export_file.sub!("<PREVTEN>1</PREVTEN>", "<PREVTEN/>")
+              export_file.sub!("<SAVINGSNK>1</SAVINGSNK>", "<SAVINGSNK/>")
+              export_file.sub!("<WCHAIR>1</WCHAIR>", "<WCHAIR/>")
+              export_file.sub!("<WHEEL>1</WHEEL>", "<WHEEL/>")
+              export_file.sub!("<HASSERVICECHARGESCHANGED/>", "<HASSERVICECHARGESCHANGED>1</HASSERVICECHARGESCHANGED>")
```
samyou-softwire, line 541: seems odd that we do this replacement work but the `sales_log_26_27.xml` doesn't contain a `<HASSERVICECHARGESCHANGED>1</HASSERVICECHARGESCHANGED>`

app/models/validations/sales/financial_validations.rb
```diff
@@ -139,6 +139,15 @@ def validate_staircase_difference(record)
     end
   end
 
+  def validate_newservicecharges_different_from_mscharge(record)
+    return unless record.hasservicechargeschanged == 1 && record.newservicecharges && record.mscharge
```
samyou-softwire: this condition seems a bit odd. if `hasservicechargeschanged` isn't 1 then `newservicecharges` will be `nil`, so we don't really need both. if we want both anyway for clarity then could we do the same for `mscharge` and `has_mscharge`?

config/locales/validations/sales/financial.en.yml
```diff
@@ -48,6 +48,10 @@ en:
         mscharge:
           monthly_leasehold_charges:
             not_zero: "Monthly leasehold charges cannot be £0 if the property has monthly charges."
+          same_as_new: "You said that the service charge will change and you entered the same amount as the new service charge. If the service charge will not change, answer 'No' to that question before entering the service charge here."
```
samyou-softwire, line 51: good spot this'll need a diff error message, though perhaps worth getting this wording confirmed

app/models/validations/sales/financial_validations.rb
```diff
@@ -139,6 +139,15 @@ def validate_staircase_difference(record)
...
+
+    if record.newservicecharges == record.mscharge
```
samyou-softwire, line 145: direct equality checking between floats can be dangerous in some cases due to floating point differences. do we want to round here, or use a bigdecimal?

for instance `a.to_d.round(2) == b.to_d.round(2)`, though I've not looked too deeply and this may not be an issue as we don't do sums on this decimal

app/services/csv/sales_log_csv_service.rb
```diff
@@ -108,6 +108,14 @@ def prepare_csv(logs)
         labels: %i[mscharge_value],
         codes: %i[mscharge_value],
       },
+      hasservicechargeschanged: {
+        labels: %i[hasservicechargeschanged_label],
```
samyou-softwire: do we know what this does? looks like its just getting the label from `hasservicechargeschanged`

samyou-softwire (changes requested):
logic looks good, couple suggestions for fixups

config/locales/validations/sales/financial.en.yml
natdeanlewissoftwire: Thanks, already done, Sammy's confirmed this on the ticket

app/services/csv/sales_log_csv_service.rb
natdeanlewissoftwire: Good point, was added for consistency with the other service charges but since removing the condition tbf it's just a default so I'll confirm the labels still download correctly without and remove if so.

spec/models/form/sales/pages/service_charge_changed_spec.rb
natdeanlewissoftwire: ah thanks, started on this PR ages ago and forgot to update with this paradigm

config/locales/forms/2026/sales/sale_information.en.yml
natdeanlewissoftwire: Yeah that's true. I'll change to "Enter the new total monthly charge" and get Sammy to confirm on the ticket.

spec/services/exports/sales_log_export_service_spec.rb
natdeanlewissoftwire: We're replacing `<HASSERVICECHARGESCHANGED/>` with `<HASSERVICECHARGESCHANGED>1</HASSERVICECHARGESCHANGED>`. The export has a `<HASSERVICECHARGESCHANGED/>` rather than a value of 1 because it's a non-staircasing log (which is the cause of why all the empty values we replace empty, because they're not routed to).

spec/models/validations/sales/financial_validations_spec.rb
natdeanlewissoftwire: I'm not sure I agree tbh, I think that approach is a lot less explicit and increases the mental load when reading the tests for me at least given what's important is the relative values, having to cross ref back and forth as to what each value is for is less nice than all the vars being instantiated explicilty. Test names are more explanatory my way too imo because they can highlight the important vars in each test rather than all of them. Also I'm also not just checking the presence of errors, I'm checking that specific errors appear in the final test. I'm pretty strongly in favour of the existing approach here but v happy to hear any counterarguments!

app/models/validations/sales/financial_validations.rb
natdeanlewissoftwire: Good thought, but I'm pretty sure we're okay here as these are stored as `BigDecimal` not `Float`. And no sums on these values either fwiw

app/models/validations/sales/financial_validations.rb
natdeanlewissoftwire: Yeah I suppose this assymetry is surprising, I slightly prefer with both so I'll add it for mscharge. Shouldn't ever make a difference ofc but better safe than sorry :)

spec/models/validations/sales/financial_validations_spec.rb
samyou-softwire: to me the idea is that the defined list of cases is the important part of the case which you'd read first. the fixture list approach keeps all of the cases we test and results we expect in one place. below that is the less interesting boiler plate to actually setup the log and check the errors.

suppose the idea is that you wouldn't need to cross reference to understand the test since the fixture list should tell you everything you'd need to know (mscharge is this, hasservicecharges is that, and we expect errors). I found these cases a little difficult to reason with as it's quite spread out

for ref the fixture list we can use more friendly structures than a list of values (a hash perhaps) and we can add labels to it as well

app/services/csv/sales_log_csv_service.rb
```diff
@@ -108,6 +108,10 @@ def prepare_csv(logs)
         labels: %i[mscharge_value],
         codes: %i[mscharge_value],
       },
+      newservicecharges: {
+        labels: %i[newservicecharges],
+        codes: %i[newservicecharges],
+      },
```
samyou-softwire: is this needed as well? the hash is `CUSTOM_CALL_CHAINS` and this doesn't seem to be custom

samyou-softwire (approved):
one last functional change. put some comment re those tests though happy to approve either way

app/services/csv/sales_log_csv_service.rb
natdeanlewissoftwire: I swear I made that change, will do now 😅

spec/models/validations/sales/financial_validations_spec.rb
natdeanlewissoftwire: Yeah with a hash would be better than without imo, but I think I'll leave this consistent with the rest of the file for now. Will bear in mind going forwards though I can definitely imagine finding it a valuable approach :)

PR #3201, opened 2026-02-26
CLDC-4176: Remove gender identity question sales

samyou-softwire:
closes [CLDC-4176](https://mhclgdigital.atlassian.net/browse/CLDC-4176)

removes sexn from web flow, bulk upload, csv export and xml export

adds tests for a couple cases that were not previously tested

refactors the household characteristics page to use a loop, similar to the refactor we made for lettings logs

[CLDC-4176]: https://mhclgdigital.atlassian.net/browse/CLDC-4176?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


config/locales/forms/2026/lettings/household_characteristics.en.yml
```diff
@@ -44,13 +44,6 @@ en:
             hint_text: ""
             question_text: "Enter gender identity"
 
-          sex1:
```
natdeanlewissoftwire, line 47: Great find! Thanks for picking these up for lettings too :)

spec/fixtures/files/sales_logs_csv_export_codes_23.csv
natdeanlewissoftwire, line 1: Are we okay to remove these? Even if so, this feels like it might be better in a different PR to me, but don't massively mind.

spec/fixtures/files/sales_logs_csv_export_labels_26.csv
natdeanlewissoftwire, line 1: I'm guessing as there no changes to the `sales_logs_csv_export_non_support_X_26` files that they're not on this branch - I've added them in CLDC-4173 (and made relevant test updates) if you wanted to copy those changes over to include here?

natdeanlewissoftwire (approved):
Looks good to me, just a couple of thoughts but happy to approve now :)

spec/fixtures/files/sales_logs_csv_export_labels_26.csv
samyou-softwire: agree; think at this point we're happy that [CLDC-4173](https://mhclgdigital.atlassian.net/browse/CLDC-4173) will most likely go in first so think I'll rebase this branch onto that one

spec/fixtures/files/sales_logs_csv_export_codes_23.csv
samyou-softwire: I removed the 2023 tests as well; we don't need to support 2023 any more so the tests won't be of help

agree could be separate though with constrains on number of PRs feels like it'd be inconvenient to split out. don't think this is essential to expedite to main

spec/fixtures/files/sales_logs_csv_export_codes_24.csv
samyou-softwire, line 1: the change here is just that it now uses the collection start date. same for any other 24/25 file changes

spec/services/csv/sales_log_csv_service_spec.rb
```diff
@@ -197,25 +199,10 @@
       expect(la_label_value).to eq "Westminster"
     end
 
-    context "when the requested form is 2023" do
```
samyou-softwire, line 200: the changes here are a lot but should be
- remove any 2023 tests
- use collection_start_date_for_year
- use `expect(csv).to eq expected_content` rather than skipping the first row since we're now adding variable definitions

app/services/bulk_upload/sales/year2026/row_parser.rb
```diff
@@ -1046,23 +1018,23 @@ def age1_known?
   end
 
   def person_2_present?
-    field_35.present? || field_36.present? || field_34.present? || field_123.present?
+    field_35.present? || field_36.present? || field_34.present?
```
natdeanlewissoftwire, line 1052: I think these should be renumbered rather than being removed - so e.g. this should include `field_36.present?`

spec/fixtures/files/sales_logs_csv_export_codes_23.csv
natdeanlewissoftwire: Sure, happy with that reasoning :)

app/services/bulk_upload/sales/year2026/row_parser.rb
natdeanlewissoftwire: Edit - it already does, ignore :)

samyou-softwire:
force push to rebase onto main

PR #3206, opened 2026-03-02
CLDC-4176: include new case insensitive fields in field renumbering

natdeanlewissoftwire:
Fixes some test failures on main since a couple of conflicting merges


samyou-softwire (approved):
LGTM, wonder why this didn't trip the merge queue pipeline

PR #3209, opened 2026-03-04
CLDC-4226: Unionise org filter scope

natdeanlewissoftwire:
Closes [CLDC-4226](https://mhclgdigital.atlassian.net/browse/CLDC-4226).

The migration that the previous PR for this ticket applied has been successfully applied, but the prod db is still unable to use it. I've done some further investigation and the POSTGRES is still making the inefficient scanning choice it was making before this migration, which comes down to the `filter_by_organisations` scope. This has an `.or` on managing and owning orgs, which is the key restriction preventing the new indices from being used. So, the solution is to use a UNION instead. Our current version of rails doesn't allow `.union` so I've done this in raw SQL but it's a simple enough query.

```
org = Organisation.find(1032)
orgs = org.absorbed_organisations + [org]
years = [2024, 2025, 2026]

# OLD (OR)
old_scope = LettingsLog.visible
  .where(owning_organisation: orgs).or(LettingsLog.visible.where(managing_organisation: orgs))
  .filter_by_years_or_nil(years.dup)
  .order(id: :desc).limit(20)

puts "=== OLD (OR) ==="
t = Time.now
old_ids = old_scope.pluck(:id)
old_time = Time.now - t
puts "IDs: #{old_ids}"
puts "Time: #{old_time}s"

# NEW (ANY(ARRAY(UNION))) — hardcoded
owned   = LettingsLog.where(owning_organisation: orgs).select(:id)
managed = LettingsLog.where(managing_organisation: orgs).select(:id)
new_scope = LettingsLog.visible
  .where("id = ANY(ARRAY(#{owned.to_sql} UNION #{managed.to_sql}))")
  .filter_by_years_or_nil(years.dup)
  .order(id: :desc).limit(20)

puts
puts "=== NEW (ANY(ARRAY(UNION))) ==="
t = Time.now
new_ids = new_scope.pluck(:id)
new_time = Time.now - t
puts "IDs: #{new_ids}"
puts "Time: #{new_time}s"

puts
puts "IDs match: #{old_ids == new_ids}"
puts "Speedup: #{(old_time / new_time).round(0)}x"
```
Returns the following, so we can be confident this will work on prod. I've also confirmed the IDs returned are the same; the outputs of the old vs new scope are identical.

```
=== OLD (OR) ===
IDs: [1013652, 1013569, 1013063, 983308, 789628, 789590, 789464, 789338, 789221, 788895]
Time: 43.271396378s

=== NEW (ANY(ARRAY(UNION))) ===
IDs: [1013652, 1013569, 1013063, 983308, 789628, 789590, 789464, 789338, 789221, 788895]
Time: 0.024607612s

IDs match: true
Speedup: 1758x
```

[CLDC-4226]: https://mhclgdigital.atlassian.net/browse/CLDC-4226?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/models/log.rb
+++ b/app/models/log.rb
@@ -52,8 +52,8 @@ class Log < ApplicationRecord
   scope :imported_2023, -> { imported.filter_by_year(2023) }
   # TODO: CLDC-4273: use .union in filter_by_organisation rather than raw SQL
   scope :filter_by_organisation, lambda { |orgs, _user = nil|
-    owned = where(owning_organisation: orgs).select(:id)
-    managed = where(managing_organisation: orgs).select(:id)
+    owned = unscoped { where(owning_organisation: orgs).select(:id) }
+    managed = unscoped { where(managing_organisation: orgs).select(:id) }
 
     where("#{table_name}.id = ANY(ARRAY(#{owned.to_sql} UNION #{managed.to_sql}))")
   }
--- a/app/models/log.rb
+++ b/app/models/log.rb
@@ -50,6 +50,7 @@ class Log < ApplicationRecord
   scope :has_old_form_id, -> { where.not(old_form_id: nil) }
   scope :imported_2023_with_old_form_id, -> { imported.filter_by_year(2023).has_old_form_id }
   scope :imported_2023, -> { imported.filter_by_year(2023) }
+  # TODO: CLDC-4273: use .union in filter_by_organisation rather than raw SQL
   scope :filter_by_organisation, lambda { |orgs, _user = nil|
     owned   = unscoped.where(owning_organisation: orgs).select(:id)
     managed = unscoped.where(managing_organisation: orgs).select(:id)
--- a/app/models/log.rb
+++ b/app/models/log.rb
@@ -50,7 +50,12 @@ class Log < ApplicationRecord
   scope :has_old_form_id, -> { where.not(old_form_id: nil) }
   scope :imported_2023_with_old_form_id, -> { imported.filter_by_year(2023).has_old_form_id }
   scope :imported_2023, -> { imported.filter_by_year(2023) }
-  scope :filter_by_organisation, ->(org, _user = nil) { where(owning_organisation: org).or(where(managing_organisation: org)) }
+  scope :filter_by_organisation, lambda { |orgs, _user = nil|
+    owned   = unscoped.where(owning_organisation: orgs).select(:id)
+    managed = unscoped.where(managing_organisation: orgs).select(:id)
+
+    where("#{table_name}.id = ANY(ARRAY(#{owned.to_sql} UNION #{managed.to_sql}))")
+  }
   scope :filter_by_owning_organisation, ->(owning_organisation, _user = nil) { where(owning_organisation:) }
   scope :filter_by_managing_organisation, ->(managing_organisation, _user = nil) { where(managing_organisation:) }
   scope :filter_by_user_text_search, ->(param, user) { where(assigned_to: User.visible(user).search_by(param)) }
--- a/app/models/log.rb
+++ b/app/models/log.rb
@@ -52,10 +52,10 @@ class Log < ApplicationRecord
   scope :imported_2023, -> { imported.filter_by_year(2023) }
   # TODO: CLDC-4273: use .union in filter_by_organisation rather than raw SQL
   scope :filter_by_organisation, lambda { |orgs, _user = nil|
-    owned   = unscoped.where(owning_organisation: orgs).select(:id)
-    managed = unscoped.where(managing_organisation: orgs).select(:id)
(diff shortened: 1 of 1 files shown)


samyou-softwire:
> Our current version of rails doesn't allow .union

I don't think we have an open ticket to update rails. looks like support for 7.2.2 which we use will expire august 9th 2026 https://rubyonrails.org/maintenance. would you be able to raise that ticket and mention that we should replace this code with a union?

natdeanlewissoftwire:
> > Our current version of rails doesn't allow .union
> 
> I don't think we have an open ticket to update rails. looks like support for 7.2.2 which we use will expire august 9th 2026 [rubyonrails.org/maintenance](https://rubyonrails.org/maintenance). would you be able to raise that ticket and mention that we should replace this code with a union?

Yeah, great shout - I've made [CLDC-4273](https://mhclgdigital.atlassian.net/browse/CLDC-4273) and added a TODO above the scope linked to that ticket number

[CLDC-4273]: https://mhclgdigital.atlassian.net/browse/CLDC-4273?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

app/models/log.rb
```diff
@@ -50,7 +50,13 @@ class Log < ApplicationRecord
   scope :has_old_form_id, -> { where.not(old_form_id: nil) }
   scope :imported_2023_with_old_form_id, -> { imported.filter_by_year(2023).has_old_form_id }
   scope :imported_2023, -> { imported.filter_by_year(2023) }
-  scope :filter_by_organisation, ->(org, _user = nil) { where(owning_organisation: org).or(where(managing_organisation: org)) }
+  # TODO: CLDC-4273: use .union in filter_by_organisation rather than raw SQL
+  scope :filter_by_organisation, lambda { |orgs, _user = nil|
+    owned   = unscoped.where(owning_organisation: orgs).select(:id)
```
samyou-softwire: not sure the spaces add much here, especially as owning and managing aren't the same length

```suggestion
    owned = unscoped.where(owning_organisation: orgs).select(:id)
```

samyou-softwire: the use of `unscoped` here seems odd. if I was chaining `filter_by_organisation` to an existing scope I'd want it to filter that scope. looks like unscoped here would reset the scope. think just `where` rather than `unscoped.where` would do it. thoughts?

app/models/log.rb
```diff
@@ -50,7 +50,13 @@ class Log < ApplicationRecord
...
+    managed = unscoped.where(managing_organisation: orgs).select(:id)
+
+    where("#{table_name}.id = ANY(ARRAY(#{owned.to_sql} UNION #{managed.to_sql}))")
```
samyou-softwire, line 58: does the `#{table_name}` do anything here? should be unambigious as we're selecting from lettings_logs/sales_logs table only

samyou-softwire (changes requested):
LGTM overall and works locally, just want to check in on that `unscoped`

app/models/log.rb
natdeanlewissoftwire: It's not _strictly_ necessary, but there's a small reduction in this needing changing in the future if we ever join on another table. Seems unlikely though admittedly, and given there's no actual reason for it at the moment and it's surprised you enough to ask about it I think we should just clean it up, will remove :)

app/models/log.rb
natdeanlewissoftwire: Ah nice catch, I've just run a test and it's not needed. Was part of a multi pronged attack to get this to work but testing it in isolation we can do without it.

app/models/log.rb
natdeanlewissoftwire: Actually scratch that, I've just removed it and the tests complain when we join on bulk upload as it's ambiguous. This should be left unchanged.

app/models/log.rb
natdeanlewissoftwire: Actually, similar to the other change I'm not making, this is needed too so the ANY ARRAY doesn't get polluted with invalid sql when chaining scopes. But I've put it in block form so it only applies to the inner scope - so scope chaining will still work as expected.

samyou-softwire: not sure I understand the usage of `unscoped` here or how not using it can lead to invalid SQL, though looking back now the final where that will correctly respect the scope. suppose `owned` & `unowned` could contain more IDs than expected.

so yeah think this makes sense. we check if our current scoped ID matches any that are in the org

samyou-softwire (approved):
comments resolved, LGTM

PR #3210, opened 2026-03-04
CLDC-4260: Sentry migration

natdeanlewissoftwire:


samyou-softwire:
could we rename the PR to reference this is an empty commit on purpose? or when merging make sure the commit title includes this. would feel odd to come across this commit later that is called migration but doesn't do anything 😅

natdeanlewissoftwire:
This was just for review app testing, will close :)

PR #3211, opened 2026-03-04
CLDC-4266: Update manual address clear copy

samyou-softwire:
closes [CLDC-4266](https://mhclgdigital.atlassian.net/browse/CLDC-4266)

uses the 'search UPRN' copy if the property is a newbuild and you see this flow

q17 address manual search with clear button references UPRN

[CLDC-4266]: https://mhclgdigital.atlassian.net/browse/CLDC-4266?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/models/lettings_log.rb
+++ b/app/models/lettings_log.rb
@@ -815,20 +815,12 @@ class LettingsLog < Log
     "lettings_log"
   end
 
-  def newbuild?
-    rsnvac == 15
-  end
-
-  def was_newbuild?
-    rsnvac_was == 15
-  end
-
   def changed_to_newbuild?
-    newbuild? && !was_newbuild?
+    rsnvac == 15 && rsnvac_was != 15
   end
 
   def changed_from_newbuild?
-    !newbuild? && was_newbuild?
+    rsnvac != 15 && rsnvac_was == 15
   end
 
   def is_address_asked?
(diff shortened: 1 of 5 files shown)


app/models/lettings_log.rb
```diff
@@ -815,12 +815,16 @@ def log_type
     "lettings_log"
   end
 
+  def newbuild?
+    rsnvac == 15
+  end
+
   def changed_to_newbuild?
-    rsnvac == 15 && rsnvac_was != 15
+    newbuild? && rsnvac_was != 15
```
natdeanlewissoftwire: Small one, but I think this and the below method are less readable now to a newcomer to them. Could you also add a `def was_newbuild?` and use that instead so at least these changed_... methods don't mix whether they're using a helper or referring to the field directly?

natdeanlewissoftwire (approved):
Just a small refactor idea but LGTM

app/models/lettings_log.rb
samyou-softwire: sounds good to me 👍

natdeanlewissoftwire (approved):
LGTM :)

PR #3213, opened 2026-03-05
CLDC-4271: Rework BU error hiding logic

samyou-softwire:
# pending CORE confirmation on how the BU parser should work

closes [CLDC-4271](https://mhclgdigital.atlassian.net/browse/CLDC-4271)

make the BU parser logic consistent across sales & lettings

right now it works like:
- lettings: only add soft validation error to that question if there are no other errors
- sales: only add soft validation error to that question if there are no other soft validation errors

not sure that either of these two approaches are quite correct. in the following report it's sensible to have two soft validations on the same question:

bu with two errors on ap2 lead tenants age

the code below will only skip adding the error if there's already one on the same question with the same text. this should make all meaningful soft validation errors shown

[CLDC-4271]: https://mhclgdigital.atlassian.net/browse/CLDC-4271?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/services/bulk_upload/lettings/year2025/row_parser.rb
+++ b/app/services/bulk_upload/lettings/year2025/row_parser.rb
@@ -679,10 +679,17 @@ private
         next if log.form.questions.none? { |q| q.id == interruption_screen_question_id && q.page.routed_to?(log, nil) }
 
         field_mapping_for_errors[interruption_screen_question_id.to_sym]&.each do |field|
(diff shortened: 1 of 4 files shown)


samyou-softwire:
closed as ticket was closed

PR #3214, opened 2026-03-06
CLDC-4212, CLDC-4213, CLDC-4214: Add 0.1% tolerance to all calculations that involve %s of a value

samyou-softwire:
closes [CLDC-4212](https://mhclgdigital.atlassian.net/browse/CLDC-4212), [CLDC-4213](https://mhclgdigital.atlassian.net/browse/CLDC-4213), [CLDC-4214](https://mhclgdigital.atlassian.net/browse/CLDC-4214)

as far as I can tell there are two validations to edit here:

- SO initial purchase: value, equity, deposit, mortgage
- discounted ownership: value, discount, deposit, mortgage

SO staircasing is not edited at all since as far as I can see this validation is unused since 2025. in 2025 we stopped asking deposit for staircasing. see [CLDC-3758](https://mhclgdigital.atlassian.net/browse/CLDC-3758) 128ac5d57ad719d36705cefaa6936d38dd15dfcc

these now explicitly run a range between the calculation if equity was 0.1% higher and 0.1% lower. I think there could be a way to reuse the existing tolerance calculation but this felt like it'd become hard to understand later. better to do explicitly.

initial purchase within 0.1%
initial purchase within 0.1%

initial purchase over bounds
initial purchase over bounds

discounted ownership within 0.1%
discounted ownership within 0.1%

discounted ownership over bounds
discounted ownership over bounds
</details>

[CLDC-4212]: https://mhclgdigital.atlassian.net/browse/CLDC-4212?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[CLDC-4213]: https://mhclgdigital.atlassian.net/browse/CLDC-4213?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[CLDC-4214]: https://mhclgdigital.atlassian.net/browse/CLDC-4214?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[CLDC-3758]: https://mhclgdigital.atlassian.net/browse/CLDC-3758?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/models/sales_log.rb
+++ b/app/models/sales_log.rb
@@ -268,10 +268,20 @@ class SalesLog < Log
     value * equity / 100
   end
 
-  def expected_shared_ownership_deposit_value_range
-    return unless value && equity
+  def expected_shared_ownership_deposit_value_tolerance
+    return 1 unless value && equity
 
-    (value * ((equity - 0.1) / 100)..value * ((equity + 0.1) / 100))
+    # we found that a simple tolerance was not quite what we wanted here.
+    # CORE wants it so if a user say, has a 66.6% equity then can enter either 66.6% or 66.7%
+    # so in 2026 we base our tolerance off of a discount 0.1% higher or lower
+    if form.start_year_2026_or_later?
+      lower_bound = value * ((equity - 0.1) / 100)
+      upper_bound = value * ((equity + 0.1) / 100)
+
+      (upper_bound - lower_bound) / 2
+    else
+      1
+    end
   end
 
   def stairbought_part_of_value
@@ -474,13 +484,20 @@ class SalesLog < Log
     value - discount_amount
   end
 
-  def value_with_discount_range
-    return if value.blank?
-    return (value..value) if discount.nil?
+  def value_with_discount_tolerance
+    return 1 if value.blank? || discount.nil?
 
-    discount_amount_lower_bound = value * (discount - 0.1) / 100
-    discount_amount_upper_bound = value * (discount + 0.1) / 100
-    (value - discount_amount_upper_bound..value - discount_amount_lower_bound)
+    # we found that a simple tolerance was not quite what we wanted here.
+    # CORE wants it so if a user say, has a 66.6% discount then can enter either 66.6% or 66.7%
+    # so in 2026 we base our tolerance off of a discount 0.1% higher or lower
+    if form.start_year_2026_or_later?
+      discount_amount_lower_bound = value * (discount - 0.1) / 100
+      discount_amount_upper_bound = value * (discount + 0.1) / 100
+
+      (discount_amount_upper_bound - discount_amount_lower_bound) / 2
+    else
+      discount ? value * 0.05 / 100 : 1
+    end
   end
 
   def mortgage_deposit_and_grant_total
--- a/app/models/sales_log.rb
+++ b/app/models/sales_log.rb
@@ -271,8 +271,8 @@ class SalesLog < Log
   def expected_shared_ownership_deposit_value_tolerance
     return 1 unless value && equity
 
-    # we found that a simple tolerance was not quite what we wanted here.
-    # CORE wants it so if a user say, has a 66.6% equity then can enter either 66.6% or 66.7%
+    # we found that a fixed tolerance was not quite what we wanted here.
+    # CORE wants it so if a user say, has a 66.666% equity they can enter either 66.6% or 66.7% (or 66.5%)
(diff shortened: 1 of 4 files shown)


app/models/sales_log.rb
```diff
@@ -468,6 +474,15 @@ def value_with_discount
     value - discount_amount
   end
 
+  def value_with_discount_range
+    return if value.blank?
+    return (value..value) if discount.nil?
```
natdeanlewissoftwire: Just checking, do we not want to have a value +-0.1% range if there is no discount?

app/models/validations/sales/sale_information_validations.rb
```diff
@@ -396,6 +393,25 @@ def validate_number_of_staircase_transactions(record)
...
+    else
+      over_tolerance?(total, record.expected_shared_ownership_deposit_value, 1)
+    end
+  end
+
+  def mortgage_deposit_and_grant_total_is_over_discount_tolerance?(record)
+    if record.form.start_year_2026_or_later?
+      !record.value_with_discount_range.cover?(record.mortgage_deposit_and_grant_total)
+    else
+      # we found that this tolerance still wasn't working for us, so for post 2026 we just calculate a range for 0.1% higher and lower.
```
natdeanlewissoftwire: Could you give a bit more detail here on what "wasn't working for us" means?

natdeanlewissoftwire (commented):
Some thoughts but looks good pending the confirmations we're waiting for on the ticket that a) this is everything and b) we should/shoudn't deprecate the soft validation here too

app/models/sales_log.rb
samyou-softwire: think so, as in the conceit behind the ticket is to allow for fudging the discount column. if they haven't entered it then there's no need to apply fudging

in practice I don't think this value will be called without discount being set

natdeanlewissoftwire: Cool, makes sense!

app/models/validations/sales/sale_information_validations.rb
samyou-softwire: have gone into some more detail

app/models/sales_log.rb
```diff
@@ -268,6 +268,22 @@ def expected_shared_ownership_deposit_value
     value * equity / 100
   end
 
+  def expected_shared_ownership_deposit_value_tolerance
+    return 1 unless value && equity
+
+    # we found that a simple tolerance was not quite what we wanted here.
+    # CORE wants it so if a user say, has a 66.6% equity then can enter either 66.6% or 66.7%
```
natdeanlewissoftwire: This is definitely +/- 0.1% right? so should be (also note I have added a missing "they"):
```
CORE wants it so if a user say, has a 66.6% equity then they can enter 66.5%, 66.6% or 66.7%
```

app/models/sales_log.rb
```diff
@@ -268,6 +268,22 @@ def expected_shared_ownership_deposit_value
     value * equity / 100
   end
 
+  def expected_shared_ownership_deposit_value_tolerance
+    return 1 unless value && equity
+
+    # we found that a simple tolerance was not quite what we wanted here.
```
natdeanlewissoftwire: Maybe also worth clarifying here that by simple tolerance we mean `tolerance by a constant value`. Can't really find a standard term for this, some people say absolute/fixed tolerance though fwiw so happy if you prefer one of those instead

natdeanlewissoftwire (approved):
Just minor stuff, will approve now and leave with you :)

app/models/sales_log.rb
samyou-softwire: the intent was to just describe these two bounds, since that's really what CORE are looking for, though can understand how this may be unclear. have updated the comment

app/models/sales_log.rb
samyou-softwire: went with fixed tolerance

PR #3216, opened 2026-03-09
CLDC-4236: trigger review app deploys manually

natdeanlewissoftwire:
Closes [CLDC-4236](https://mhclgdigital.atlassian.net/browse/CLDC-4236).

I've looked a few options here and what I've landed on is:
On PR creation, github bot comments this comment on a PR:
image
The dev then comments `/deploy-review` when they want to deploy the review app (probably just before PO)
image

I've kept the workflow dispatch trigger too as a manual fallback, but that requires going to the actions tab, selecting the "Review app pipeline" one, selecting the right branch and then making sure to enter the correct PR number manually so the method explained above is a bit less overhead for the dev given we'll be doing this a lot.

The review app is then provisioned, and a link added to the PR as normal. Review apps are still auto-closed as normal too, all that's unchanged.

[CLDC-4236]: https://mhclgdigital.atlassian.net/browse/CLDC-4236?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

I'll confirm this is all working as expected once we merge to main, but the manual fallback definitely does (that's testable on any branch) so worst case nobody will be blocked.


.github/workflows/review_pipeline.yml
```diff
@@ -1,57 +1,74 @@
...
-      - synchronize
-      - reopened
-  workflow_dispatch:
+  issue_comment:
+    types: [created]
 
-defaults:
-  run:
-    shell: bash
+permissions: {}
```
natdeanlewissoftwire, line 13: Added to address this warning I saw without it:

Image

.github/workflows/review_pipeline.yml
```diff
@@ -1,57 +1,84 @@
...
       pull-requests: write
 
     steps:
       - name: Comment on PR with URL
-        uses: unsplash/comment-on-pr@v1.3.0
-        env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        uses: actions/github-script@v7
         with:
-          msg: "Created review app at https://review.submit-social-housing-data.communities.gov.uk/${{ github.event.pull_request.number }}. Note that the review app will be automatically deprovisioned after 30 days and will need the review app pipeline running again."
```
oscar-richardson-softwire, line 55: Nice, looks like that `unsplash/comment-on-pr@v1.3.0` action hasn't been updated in 4 years, so think this move is very wise

oscar-richardson-softwire (approved):
LGTM

.github/workflows/review_pipeline.yml
natdeanlewissoftwire: Nice, this was necessitated anyway now we're not adding the review app link comment on PR creation but that's definitely a bonus!

PR #3217, opened 2026-03-09
CLDC-4248: change waityear and layear response copy

natdeanlewissoftwire:
Closes CLDC-4248.

Changes the copy in `layear` and `waityear` for "Less than 1 year" to "Under 1 year" for consistency with the BU forms. This discrepancy exists in both the 25/26 and 26/27 forms so both are updated in this PR.

image

image

image


app/models/form/lettings/questions/layear.rb
```diff
@@ -8,7 +8,20 @@ def initialize(id, hsh, page)
...
+        "7" => { "value" => "1 year but under 2 years" },
+        "8" => { "value" => "2 years but under 3 years" },
+        "9" => { "value" => "3 years but under 4 years" },
+        "10" => { "value" => "4 years but under 5 years" },
+        "11" => { "value" => "5 years but under 10 years" },
+        "12" => { "value" => "10 years or more" },
+        "divider" => { "value" => true },
+        "6" => { "value" => "Don’t know" },
+      }
+    elsif form.start_year_2024_or_later?
```
samyou-softwire, line 24: seems a bit odd that we're implicitly no longer allowing code 12 for 2024 logs here. don't think these are 100% deprecated yet so we should probably avoid invalidating them

perhaps a ternary on code 2?

samyou-softwire (changes requested):
one question otherwise LGTM

app/models/form/lettings/questions/layear.rb
samyou-softwire: has been pointed out to me that this branch is in addition to the one below so all should be fine

samyou-softwire (approved):

PR #3224, opened 2026-03-10
CLDC-4157: Remove nursing care from scheme level of support

samyou-softwire:
closes [CLDC-4157](https://mhclgdigital.atlassian.net/browse/CLDC-4157)

removes support type option 5 from the enum. this removes it from the schemes page automatically

include a script to automatically set all impacted schemes to invalid and all impacted logs to incomplete, ready for LAs to fill in again

[CLDC-4157]: https://mhclgdigital.atlassian.net/browse/CLDC-4157?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/views/schemes/support.html.erb
+++ b/app/views/schemes/support.html.erb
@@ -12,23 +12,15 @@
       <%= render partial: "organisations/headings", locals: { main: "What support does this scheme provide?", sub: @scheme.service_name } %>
       <%= govuk_inset_text(text: "Only update a scheme if you’re fixing an error. If the scheme is changing,  create a new scheme.") if @scheme.confirmed? %>
 
-      <% support_level_options_hints = { "Low level": "Staff visiting once a week, fortnightly or less.", "Medium level": "Staff on site daily or making frequent visits with some out-of-hours cover.", "High level": "Intensive level of staffing provided on a 24-hour basis." } %>
-
-      <% support_level_options_with_hints = Scheme.support_types.keys.excluding("Missing").excluding("Floating support").map { |key, _| OpenStruct.new(id: key, name: key.to_s.humanize, description: support_level_options_hints[key.to_sym]) } %>
-
       <%= f.govuk_collection_radio_buttons :support_type,
-                                           support_level_options_with_hints,
+                                           Scheme.support_level_options_with_hints,
                                            :id,
                                            :name,
                                            :description,
                                            legend: { text: "Level of support given", size: "m" } %>
 
-      <% intended_length_of_stay_options_hints = { "Very short stay": "Up to one month.", "Short stay": "Up to one year.", "Medium stay": "More than one year but with an expectation to move on.", "Permanent": "Provides a home for life with no requirement for the tenant to move." } %>
-
-      <% intended_length_of_stay_options_with_hints = Scheme.intended_stays.keys.excluding("Missing").map { |key, _| OpenStruct.new(id: key, name: key.to_s.humanize, description: intended_length_of_stay_options_hints[key.to_sym]) } %>
-
       <%= f.govuk_collection_radio_buttons :intended_stay,
-                                           intended_length_of_stay_options_with_hints,
+                                           Scheme.intended_length_of_stay_options_with_hints,
                                            :id,
                                            :name,
                                            :description,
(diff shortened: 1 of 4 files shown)


app/views/schemes/support.html.erb
```diff
@@ -12,23 +12,15 @@
       <%= render partial: "organisations/headings", locals: { main: "What support does this scheme provide?", sub: @scheme.service_name } %>
       <%= govuk_inset_text(text: "Only update a scheme if you’re fixing an error. If the scheme is changing,  create a new scheme.") if @scheme.confirmed? %>
 
-      <% support_level_options_hints = { "Low level": "Staff visiting once a week, fortnightly or less.", "Medium level": "Staff on site daily or making frequent visits with some out-of-hours cover.", "High level": "Intensive level of staffing provided on a 24-hour basis." } %>
```
samyou-softwire, line 15: there was an issue with this on the other support pages as well, despite `support_level_options_with_hints` being defined identically in `scheme.rb` it was redefined here. I've refactored to use that usage in `scheme.rb`

lib/tasks/update_schemes_logs_impacted_by_nursing_level_removal.rake
```diff
@@ -0,0 +1,10 @@
+desc "Update schemes that had a level of support of 'Nursing' to be incomplete, and mark all logs that use that scheme as incomplete"
+task update_scheme_registered_under_care_value: :environment do
+  ActiveRecord::Base.transaction do
+    impacted_schemes = Scheme.where(support_type: 5)
+    impacted_logs = LettingsLog.filter_by_year_or_later(2025).where(scheme: impacted_schemes)
```
oscar-richardson-softwire, line 5: Didn't know Rails could do this, I've been plucking IDs this whole time 😆 Cool 🆒

lib/tasks/update_schemes_logs_impacted_by_nursing_level_removal.rake
```diff
@@ -0,0 +1,10 @@
+desc "Update schemes that had a level of support of 'Nursing' to be incomplete, and mark all logs that use that scheme as incomplete"
+task update_scheme_registered_under_care_value: :environment do
```
oscar-richardson-softwire: Looks like the task name needs to be updated

lib/tasks/update_schemes_logs_impacted_by_nursing_level_removal.rake
```diff
@@ -0,0 +1,10 @@
...
+  ActiveRecord::Base.transaction do
+    impacted_schemes = Scheme.where(support_type: 5)
+    impacted_logs = LettingsLog.filter_by_year_or_later(2025).where(scheme: impacted_schemes)
+
+    impacted_schemes.update!(support_type: nil, confirmed: false)
+    impacted_logs.update!(scheme: nil)
```
oscar-richardson-softwire, line 8: Does this also blank the location?

lib/tasks/update_schemes_logs_impacted_by_nursing_level_removal.rake
oscar-richardson-softwire, line 5: Why are we filtering to 2025 or later? Is the rational that we don't care about earlier logs because they can in theory be removed from CORE now?

lib/tasks/update_schemes_logs_impacted_by_nursing_level_removal.rake
samyou-softwire: ah good spot, have fixed

lib/tasks/update_schemes_logs_impacted_by_nursing_level_removal.rake
samyou-softwire: yes, since unasked questions are blanked

lib/tasks/update_schemes_logs_impacted_by_nursing_level_removal.rake
samyou-softwire: yeah they're only there for reference and won't be exported any more so doesn't seem necessary to fix these schemes. would want to update as little as we need to

oscar-richardson-softwire (approved):
LGTM

PR #3230, opened 2026-03-11
CLDC-4270: Fix text being coerced to 0 on integer fields

samyou-softwire:
closes [CLDC-4270](https://mhclgdigital.atlassian.net/browse/CLDC-4270)

when generating the hash for the row parser, check that every field marked as wanting to be an integer can actually be made into an integer

without this, entering a string will end up with the attribute as 0

has to be csv parser logic as by the time the row parser is constructed it's already too late, though needs a little extra mechanism so the row parser can report which fields were invalid when checked by the validator

should be duplicate changes for lettings 2025, lettings 2026, sales 2025, sales 2026

[CLDC-4270]: https://mhclgdigital.atlassian.net/browse/CLDC-4270?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/services/bulk_upload/sales/year2025/row_parser.rb
```diff
@@ -288,7 +288,7 @@ class BulkUpload::Sales::Year2025::RowParser
 
   attribute :field_112, :integer
   attribute :field_113, :decimal
-  attribute :field_114, :integer
+  attribute :field_114, :decimal
```
samyou-softwire, line 291: so this and the change for 2026 are a specific change. in short I think this should have always been decimal, but I'm also not convinced decimal is correct.

the issue here was surfaced to me by the test in `validator_spec.rb` for sales that generates a log, converts to csv, and attempts to parse it. field_114 relates to the `grant` db column.

- `grant` is a decimal column in the db
- despite this, CORE validates that it is always a whole number
- when the test generates the CSV, it writes for instance `10000.0` as it is a decimal
- the new code in this PR wants to validate that `10000.0` is an integer, but this fails the parse
- so we get a failing test as it expects the log to be valid

what I want to do is make it so `grant` is not a decimal in the DB (it only holds whole numbers) but this feels like a bigger change than is wanted right now. for the meantime, I think it's fine to mark this as a decimal in the BU. later validations will ensure it actually is a whole number



though happy to hear other viewpoints on this. I've raised [CLDC-4299](https://mhclgdigital.atlassian.net/browse/CLDC-4299) to investigate why these are decimals in the db in the first place

natdeanlewissoftwire: Yep sounds fine to me - we're covered by the hard validation, and agree DB changes shouldn't be included here.

spec/services/bulk_upload/lettings/year2026/csv_parser_spec.rb
```diff
@@ -251,4 +251,24 @@
...
+    let(:seed) { rand }
+    let(:log_to_csv) { BulkUpload::LettingsLogToCsv.new(log:) }
+    let(:field_numbers) { log_to_csv.default_2026_field_numbers }
+    let(:field_values) { log_to_csv.to_2026_row }
+
+    before do
+      field_46_index = field_numbers.index(46)
+      field_values[field_46_index] = "GBR" # should be a 3 digit code
+
+      file.write(log_to_csv.custom_field_numbers_row(seed:, field_numbers:))
```
natdeanlewissoftwire: why do we seed here but not in the 2025 spec?

app/services/bulk_upload/lettings/year2025/csv_parser.rb
```diff
@@ -110,6 +131,20 @@ def normalised_string
     @normalised_string
   end
 
+  # this is needed as a string passed to an int attribute is by default mapped to '0'.
+  # this is bad as some questions will accept a '0'. so you could enter something invalid and not be told about it
+  def value_is_valid_for_field(field, value)
```
natdeanlewissoftwire: I think ruby convention is to name bool methods to end in ?, maybe `value_valid_for_field?`

app/services/bulk_upload/lettings/year2026/csv_parser.rb
```diff
@@ -32,11 +34,30 @@ def row_parsers
     @row_parsers ||= body_rows.map { |row|
       next if row.empty?
 
+      invalid_fields = []
       stripped_row = row[col_offset..]
 
-      hash = Hash[field_numbers.zip(stripped_row)]
+      hash_rows = field_numbers
```
natdeanlewissoftwire, line 40: As this isn't a hash, maybe something like `valid_fields` would be a better name

natdeanlewissoftwire (approved):
Can we share some of the duplicated methods?
I think most/all of:
`value_is_valid_for_field`
`add_invalid_field`
`invalid_fields`
`add_errors_for_invalid_fields` can be shared, maybe in a separate helper? To be fair this probably applies to a lot of our bulk upload logic so if it's actually more confusing to do that at this point and it would be a larger refactor, I'm happy if raising a tech debt ticket is the more pragmatic thing to do :)

app/services/bulk_upload/lettings/year2026/csv_parser.rb
samyou-softwire: this is the list of rows that will become the hash. maybe `rows_for_hash` would be clearer?

spec/services/bulk_upload/lettings/year2026/csv_parser_spec.rb
samyou-softwire: ah good point, this was originally in the test but I didn't think it added much so I took it out. have removed

app/services/bulk_upload/lettings/year2025/csv_parser.rb
samyou-softwire: good spot, I'll add this

samyou-softwire:
have raised [CLDC-4317](https://mhclgdigital.atlassian.net/browse/CLDC-4317)

[CLDC-4317]: https://mhclgdigital.atlassian.net/browse/CLDC-4317?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

app/services/bulk_upload/lettings/year2026/csv_parser.rb
natdeanlewissoftwire: Yeah agree that's clearer, still could be a bit more descriptive of what it actually contains I think but I'm happy with `rows_for_hash`

natdeanlewissoftwire (approved):

PR #3238, opened 2026-03-13
CLDC-4236: Deploy review apps manually using PR labels

natdeanlewissoftwire:
Closes [CLDC-4236](https://mhclgdigital.atlassian.net/browse/CLDC-4236).

[The previous PR](https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/3216) for this was based off adding a `/deploy-review` comment that then kicked off the review app. But on subsequent updates that added a lot of complexity + rewriting due to needing to then check if the comment had been added and only updating review apps in that case on pushes to open PRs.

So, the much simpler approach is just adding a `review-app` label. On PR open, a hint comment is added to the PR:
image

Then the user adds the label and the review app is deployed from here on out.

image

Review apps are deployed on adding this label, and on push IF this label exists on the PR. This makes the code change required minimal, and keeps our review apps up to date once they've been deployed.

If the label is removed then review app deployments will be skipped again until it's re-added, or the PR is closed (which will initiate teardown, as in the current process). Adding teardown on label removal is not needed and would add more points of failure so I have left that process as it is.

This PR also makes a few other misc updates to workflows which I've commented on in the diff view.

[CLDC-4236]: https://mhclgdigital.atlassian.net/browse/CLDC-4236?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.github/workflows/production_pipeline.yml
```diff
@@ -3,7 +3,6 @@ name: Production CI/CD Pipeline
 on:
   release:
     types: [released]
-  workflow_dispatch:
```
natdeanlewissoftwire, line 6: Discussed w Sammy, and searched docs - using `workflow_dispatch` to release is not part of our prod release process, so this should be removed.

.github/workflows/review_deploy.yml
```diff
@@ -48,10 +51,21 @@ jobs:
 
     steps:
       - name: Comment on PR with URL
-        uses: unsplash/comment-on-pr@v1.3.0
```
natdeanlewissoftwire, line 51: Though this change is not technically needed in the new approach, `comment-on-pr` hasn't been updated in years, moving to use github's default way of adding comments is a good idea going forwards. see https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/3216#discussion_r2916591640 from the previous iteration of this PR

.github/workflows/review_teardown_pipeline.yml
```diff
@@ -7,7 +7,6 @@ on:
   pull_request:
     types:
       - closed
-  workflow_dispatch:
```
natdeanlewissoftwire, line 10: This was broken anyway as workflow_dispatch doesn't provide the PR number. If we need this in future we could add it as an input the user has to manually enter, but it's not been needed thus far so just cleaning this up.

.github/workflows/manual_review_code_pipeline.yml
```diff
@@ -1,4 +1,4 @@
-name: Manual review app code pipeline
+name: "Review App: Manual Code Redeploy"
```
natdeanlewissoftwire: Unifying the naming convention for all our separate review app related workflows (4 of them now)

samyou-softwire: I think this would be more useful as

`Review App: Deploy (ECS Image only)`

to make it clearer the difference between this and a full deploy.

can pipelines have descriptions? it'd be useful to be able to mention why we'd want to run this (to manually push an image if it's expired from ECS)

otherwise, please make sure the internal docs on running these tasks is up to date https://softwiretech.atlassian.net/wiki/x/_QBWCgU

.github/workflows/review_deploy.yml
samyou-softwire, line 1: is `review_pipeline` still an accurate name for this file?

.github/workflows/review_deploy.yml
```diff
@@ -18,6 +18,7 @@ defaults:
 jobs:
   infra:
     name: Deploy review app infrastructure
+    if: contains(github.event.pull_request.labels.*.name, 'review-app')
```
samyou-softwire, line 21: contains seems a little odd here, is there an equals operator?

.github/workflows/review_teardown_pipeline.yml
samyou-softwire, line 10: could we bring this back into use? right now it'd only be possible to tear down a review app by closing the PR which would be a bit inconvenient

don't mind if the flow is remove the label > run this pipeline manually. do think it is important though that we do have some sort of process for this

.github/workflows/review_deploy.yml
```diff
@@ -48,10 +51,21 @@ jobs:
...
-        env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        uses: actions/github-script@v7
         with:
-          msg: "Created review app at https://review.submit-social-housing-data.communities.gov.uk/${{ github.event.pull_request.number }}. Note that the review app will be automatically deprovisioned after 30 days and will need the review app pipeline running again."
-          check_for_duplicate_msg: true
-          duplicate_msg_pattern: Created review app at*
+          script: |
+            const prNumber = context.issue.number;
+            const msg = `Created review app at https://review.submit-social-housing-data.communities.gov.uk/${prNumber}. Note that the review app will be automatically deprovisioned after 30 days and will need the review app pipeline running again.`;
```
samyou-softwire, line 58: regardless of what we go with, could this comment also detail how to turn off a review app if wanted, even if briefly

samyou-softwire (changes requested):
generally wonder if there's something we can do for turning off review apps. if it's going to be difficult then don't mind if we have to drop it, though I think manual teardown pipeline will be useful in general

.github/workflows/review_deploy.yml
natdeanlewissoftwire: I think contains is right, if another label gets added we still want the `review-app` label to have the same effect

.github/workflows/review_teardown_pipeline.yml
natdeanlewissoftwire: We could, but it feels unnecessary to me. Thus far we've been fine with review apps closing after 30 days or on PR close, and I'm not aware of any cases we really needed to be able to do this - the real issue we've had has just been creating too many, which will be fixed once this PR is merged. So I think we should do the minimal thing here and not add extra process until it's actually needed (as my guess is it never will be!).

The current process of closing a PR to force teardown of a review app will still work of course :) I'm happy to update the docs to highlight that that process is the way to go if a review app needs to be torn down urgently if you're happy with leaving things like this?

.github/workflows/review_deploy.yml
natdeanlewissoftwire: see other thread, I'm not convinced this is actually needed, but will let you reply there :)

.github/workflows/manual_review_code_pipeline.yml
natdeanlewissoftwire: Sure, happy with that name!
I don't think gh workflows can have descriptions but I've added a comment to the YAML to explain usage.
Those docs look up to date to me as is but the comment is useful to have too imo

.github/workflows/review_deploy.yml
natdeanlewissoftwire: This feels okay to me, but I've changed to review_deploy (and review_deploy_image_only) for the above which is better though I think, so fair point!

.github/workflows/review_teardown_pipeline.yml
natdeanlewissoftwire: Or is this something that you've wanted to have in support? If so I guess it feels not the purpose of this ticket's scope to me, but I suppose it wouldn't be too much extra effort to tack on if you really think it's needed?

.github/workflows/review_deploy.yml
samyou-softwire: ohh gotcha, thought this meant true if any label name contains 'review-app', that makes more sense

.github/workflows/review_teardown_pipeline.yml
samyou-softwire: we've had cases in the past where it'd be good to be able to manually deprovision a review app if it failed to deploy, though agree this is getting out of scope. maybe a new ticket?

samyou-softwire (approved):
LGTM after one comment

.github/workflows/review_teardown_pipeline.yml
natdeanlewissoftwire: Sure, happy with that: https://mhclgdigital.atlassian.net/browse/CLDC-4322

PR #3239, opened 2026-03-16
CLDC-4215: Don't run validate_discounted_ownership_value if mortgageused is 3

samyou-softwire:
I don't think this should have ever been turned on. if we don't know the mortgage we can't assume info in this check

technically this could be a breaking change as it is not gated to 26/27 but:
- 24/25 logs are read only. this change only makes existing invalid logs valid, but there shouldn't be any invalid 24/25 logs anyway. it won't break any existing logs
- 25/26 logs do not have a don't know option visible for discounted ownership
- 26/27 logs are fine to update

so I think it's better to not add a 2026 check if its not needed. keeps the logic cleaner

cya page where the numbers dont add up but its fine as mortgage isn't known


oscar-richardson-softwire (approved):
LGTM

PR #3240, opened 2026-03-17
CLDC-4297: Align new service charges with bu spec

natdeanlewissoftwire:
Closes [CLDC-4297](https://mhclgdigital.atlassian.net/browse/CLDC-4297).

This PR adds bulk upload behaviour for the new service charge question `newservicecharges`, inferring `hasservicechargeschanged` as required. This mostly mirrors `mscharge`, except `mscharge` is present in all three flows, not just staircasing.

This PR also tests that R is a valid value for both bulk upload fields, mapping to "No" in both cases.

[CLDC-4297]: https://mhclgdigital.atlassian.net/browse/CLDC-4297?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


spec/services/bulk_upload/sales/year2023/row_parser_spec.rb
```diff
@@ -1334,7 +1334,7 @@ def purch_id_field
     context "when mscharge is given, but is set to 0 for shared ownership" do
       let(:attributes) { valid_attributes.merge(field_114: "0") }
 
-      it "does not override variables correctly" do
```
natdeanlewissoftwire, line 1337: having "correctly" here was just confusing and didn't add anything meaningful - before, it sounded like this was testing that this was NOT working correctly

app/services/bulk_upload/sales/year2026/row_parser.rb
```diff
@@ -1389,6 +1398,33 @@ def infer_soctenant_from_prevten_and_prevtenbuy2
     end
   end
 
+  def validate_service_charge_fields
```
natdeanlewissoftwire, line 1424: needed a separate validation method here to force block_log_creation!, otherwise a nil value triggers an error but defaults to "there are no service charges" which is not what we want if the field uploaded is empty.

app/services/bulk_upload/sales/year2026/row_parser.rb
```diff
@@ -950,8 +957,8 @@ def attributes_for_log
     attributes["gender_same_as_sex6"] = field_68
     attributes["gender_description6"] = field_69
 
-    attributes["hasservicechargeschanged"] = field_125
-    attributes["newservicecharges"] = field_126
+    attributes["newservicecharges"] = field_126.to_d if field_126.present? && field_126 != "R" && field_126.to_d.positive?
```
natdeanlewissoftwire: This ensures "R" maps to no charges, and there are similar updates for the older mscharge questions below

app/services/bulk_upload/sales/year2026/row_parser.rb
```diff
@@ -1389,6 +1396,33 @@ def infer_soctenant_from_prevten_and_prevtenbuy2
     end
   end
 
+  def validate_service_charge_fields
+    service_charge_format = /\A(\d+(\.\d+)?|R)\z/
+    message = I18n.t("#{ERROR_BASE_KEY}.mscharge.invalid")
+
+    if shared_ownership_initial_purchase? && field_107.present? && !field_107.match?(service_charge_format)
+      block_log_creation!
```
samyou-softwire: we should be careful about using `block_log_creation!`, the error it sends to users is 


which isn't particularly accurate to the situation here.

fwiw I believe the flow is that adding the error is enough as long as the eventual log that is made isn't complete. easiest is to make sure mscharge isn't set to anything if the user inputs an invalid value. this way they can go correct the log on the CORE site and add in the valid value if they want

maybe a comment on `block_log_creation!` also would make it more clear what it does

samyou-softwire (changes requested):
a little tricky to review this right now due to the diff, though I think we need a change in approach. ideally if you put in something like "hmm" it should show a single error message using the normal 'there are errors in the BU' email. believe there should be precedent in the file from other R fields on how to do this

happy to pair as appreciate this isn't the best documented so we can figure this out

app/services/bulk_upload/sales/year2026/row_parser.rb
natdeanlewissoftwire: Ah thanks you're absolutely right, `block_log_creation!` is not what I want here. The issue was that invalid values are currently defaulting to "no service charges" in both questions, leading to an otherwise valid log.

Have rebased again since this comment too so the diff will be clearer too (depending when you see this)

app/services/bulk_upload/sales/year2026/row_parser.rb
samyou-softwire: is this still accurate? suppose it can't hurt being a separate function

app/services/bulk_upload/sales/year2026/row_parser.rb
samyou-softwire: lovely stuff

app/services/bulk_upload/lettings/year2025/row_parser.rb
```diff
@@ -525,6 +525,7 @@ def log
     @log ||= LettingsLog.new(attributes_for_log)
   end
 
+  # Will send a "Bulk upload failed" email rather than an "Errors in bulk upload" email
```
samyou-softwire: perhaps worth being clear the body of the email is that there were errors in the setup section - so we can only use this if there's an error in the setup section

app/services/bulk_upload/sales/year2026/row_parser.rb
samyou-softwire: although it is a little inconsistent. we use `validates` methods elsewhere in the parser

```ruby
validates :field_133,
          if: :discounted_ownership?,
          format: {
            with: /\A(\d+|R)\z/,
            message: I18n.t("#{ERROR_BASE_KEY}.mortlen.invalid"),
          },
          on: :after_log
```

I'm not particularly opinionated on this but feels like we shouldn't be mixing approaches in the same file

samyou-softwire (changes requested):

app/services/bulk_upload/lettings/year2025/row_parser.rb
natdeanlewissoftwire: Sure, happy to

app/services/bulk_upload/sales/year2026/row_parser.rb
natdeanlewissoftwire: Yeah, I considered using validates but I think this is much clearer this way to group these errors more clearly and show how they're similar/different to each other. We already mix between validates and def validate... in this file so it's not really any more inconsistency imo.

spec/services/bulk_upload/sales/year2026/row_parser_spec.rb
```diff
@@ -1957,13 +1957,103 @@ def purch_id_field
...
+        log = parser.log
+        expect(log["newservicecharges"]).to eq(150)
+        expect(log["hasservicechargeschanged"]).to eq(1)
+      end
+    end
+
+    context "when newservicecharges is set to 0" do
+      let(:attributes) { valid_attributes.merge(field_126: "0") }
+
+      it "does not set newservicecharges and sets hasservicechargeschanged to no" do
```
samyou-softwire: should it? I thought you should input "R" for no

spec/services/bulk_upload/sales/year2026/row_parser_spec.rb
```diff
@@ -1957,13 +1957,103 @@ def purch_id_field
...
+
+    context "when newservicecharges is an invalid string" do
+      let(:attributes) { valid_attributes.merge(field_126: "S") }
+
+      it "adds a validation error" do
+        parser.valid?
+        expect(parser.errors[:field_126]).to include(I18n.t("validations.sales.2026.bulk_upload.newservicecharges.invalid"))
+      end
+
+      it "does not set hasservicechargeschanged" do
```
samyou-softwire: think these last 3 test names for nil should reference both variables

samyou-softwire (changes requested):
one last question on the logic here

spec/services/bulk_upload/sales/year2026/row_parser_spec.rb
natdeanlewissoftwire: Ah good shout, this is actually assymmetric to mscharge we want to allow 0 here.

spec/services/bulk_upload/sales/year2026/row_parser_spec.rb
```diff
@@ -1944,23 +1944,283 @@ def purch_id_field
...
+
+          it "does not set has_mscharge or mscharge" do
+            log = parser.log
+            expect(log["has_mscharge"]).to be_nil
+            expect(log["mscharge"]).to be_nil
+          end
+        end
+
+        context "when negative" do
+          let(:attributes) { valid_attributes.merge(field_125: "-100") }
```
samyou-softwire, line 2213: is there a reason field_106 doesn't have an equivalent here?

spec/services/bulk_upload/sales/year2026/row_parser_spec.rb
```diff
@@ -1944,23 +1944,283 @@ def purch_id_field
...
+          end
+        end
+      end
+
+      context "when newservicecharges equals mscharge (field_125 == field_126)" do
+        let(:attributes) { valid_attributes.merge(field_125: "200", field_126: "200") }
+
+        it "adds validation errors to both fields" do
+          parser.valid?
+          expect(parser.errors[:field_125]).to be_present
```
samyou-softwire: could we specific on the error here?

samyou-softwire (approved):
LGTM, only one other test to mention is that the sales 2026 `row_parser_spec` has a test for the case insensitive fields at line 304 that'll need updating

```ruby
context "and case insensitive fields are set to lowercase" do
  let(:case_insensitive_fields) { %w[field_30 field_39 field_49 field_55 field_61 field_67] }
  let(:case_insensitive_integer_fields_with_r_option) { %w[field_29 field_38 field_48 field_54 field_60 field_66 field_77 field_88 field_83 field_85 field_103 field_133] }
  let(:attributes) do
    valid_attributes
      .merge(case_insensitive_fields.each_with_object({}) { |field, h| h[field.to_sym] = valid_attributes[field.to_sym]&.downcase })
      .merge(case_insensitive_integer_fields_with_r_option.each_with_object({}) { |field, h| h[field.to_sym] = "r" })
  end

  it "is still valid" do
    expect(parser).to be_valid
  end
end
```

spec/services/bulk_upload/sales/year2026/row_parser_spec.rb
natdeanlewissoftwire: field_106 isn't related to this changeset - and there are already negative tests for all 4 fields:
107, 125, 126, 136. Guessing you were just looking for the wrong field number?

spec/services/bulk_upload/sales/year2026/row_parser_spec.rb
natdeanlewissoftwire: Sure, good shout will do

app/services/bulk_upload/sales/year2026/row_parser.rb
```diff
@@ -1387,6 +1397,29 @@ def infer_soctenant_from_prevten_and_prevtenbuy2
     end
   end
 
+  def validate_service_charge_fields
+    message = I18n.t("#{ERROR_BASE_KEY}.mscharge.invalid")
+
+    if shared_ownership_initial_purchase? && field_107.present? && (!field_107.match?(SERVICE_CHARGE_FORMAT) || (field_107 != "R" && field_107.to_d.zero?))
```
samyou-softwire: this logic is getting a little confusing right now as we include all these cases

I notice part of it will prevent a '0' being sent to the log. though my understanding is that if we were to do that the validations on the log (that the web flow use) would kick in and reject the 0, therefore rejecting the BU. adding it here as well seems to be duplicating the logic. can it be removed?

samyou-softwire (changes requested):

app/services/bulk_upload/sales/year2026/row_parser.rb
natdeanlewissoftwire: Not quite though I agree this is quite convoluted now. The way bulk upload (and the mscharge q) are set up atm, 0 is mapped to "No" for `has_mscharge`. Without this check, otherwise valid logs are auto-uploaded with "Errors fixed on site", not rejected, even though this is an invalid value. We could add this check to line 1036 instead (the `attributes["has_mscharge"]` bit) but it needs to be somewhere.

samyou-softwire: could it be reworked so 0 (and anything that can be parsed to a number) goes to "yes" for has_mscharge. then only "R" would go to "no" (and blank to nil I suppose). hope this would mean it'll trigger the site validations (as if the user answered 'yes' and wrote 0)

natdeanlewissoftwire: Yes! I've done this and refactored out into a few simpler methods which hopefully help with the readability.

app/services/bulk_upload/sales/year2026/row_parser.rb
```diff
@@ -152,6 +152,7 @@ class BulkUpload::Sales::Year2026::RowParser
   }.freeze
 
   ERROR_BASE_KEY = "validations.sales.2026.bulk_upload".freeze
+  SERVICE_CHARGE_FORMAT = /\A(\d+(\.\d+)?|R)\z/i
```
samyou-softwire: I wonder if we can name this better, other fields may want to use this kind of format (decimal or R)

config/locales/validations/sales/2026/bulk_upload.en.yml
```diff
@@ -44,6 +44,10 @@ en:
             not_answered: "Enter either the UPRN or the full address."
           nationality:
             invalid: "Select a valid nationality."
+          mscharge:
+            invalid: "Service charge must be a positive number or the letter R."
+          newservicecharges:
+            invalid: "New service charge must be a number or the letter R."
```
samyou-softwire, line 50: is there a reason why these two differ?

spec/services/bulk_upload/sales/year2026/row_parser_spec.rb
```diff
@@ -1944,23 +1944,472 @@ def purch_id_field
...
+
+        context "when negative" do
+          let(:attributes) { valid_attributes.merge(field_10: "2", field_107: "-100") }
+
+          it "does not set has_mscharge or mscharge" do
+            log = parser.log
+            expect(log["has_mscharge"]).to be_nil
+            expect(log["mscharge"]).to be_nil
+          end
+        end
```
samyou-softwire, line 2085: notice we don't verify errors here and in the other fields' blank and negative tests

samyou-softwire (approved):
LGTM, a couple final comments

config/locales/validations/sales/2026/bulk_upload.en.yml
natdeanlewissoftwire: Yep, because `newservicecharges` doesn't have to be positive - it can be 0 as well. I don't think we should explicitly mention negatives here fwiw (even though we do validate against them of course), that's overkill imo.

spec/services/bulk_upload/sales/year2026/row_parser_spec.rb
natdeanlewissoftwire: Yeah, happy to add them too for consistency

app/services/bulk_upload/sales/year2026/row_parser.rb
natdeanlewissoftwire: Sure, will go for `NUMBER_OR_R_FORMAT` as that includes decimals AND whole numbers which is a bit clearer on first impression I think

PR #3241, opened 2026-03-17
Revert "CLDC-4179: Amend sales duplicate log check"

samyou-softwire:
Reverts communitiesuk/submit-social-housing-lettings-and-sales-data#3207 [CLDC-4179](https://mhclgdigital.atlassian.net/browse/CLDC-4179)

not able to be tested in time for release so pulling, will go in as part of next release

[CLDC-4179]: https://mhclgdigital.atlassian.net/browse/CLDC-4179?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #3207, opened 2026-03-03
CLDC-4179: Amend sales duplicate log check

samyou-softwire:
closes [CLDC-4179](https://mhclgdigital.atlassian.net/browse/CLDC-4179)

see also #3179 and #3205 for lettings log equivalent changes

makes the same set of changes for sales logs

- use new sexrab1 question for 2026. keep using sex1 for 2025
- add uprn and address_line1 to duplicate fields

in both the frontend UI and BU

<details><summary>screenshots</summary>
bulk upload duplicate error
bulk upload duplicate error

duplicate logs 2025 manual address
duplicate logs 2025 manual address

duplicate logs 2025 uprn entered
duplicate logs 2025 uprn entered

duplicate logs 2026 manual address
duplicate logs 2026 manual address

duplicate logs 2026 uprn entered
duplicate logs 2026 uprn entered
</details>

[CLDC-4179]: https://mhclgdigital.atlassian.net/browse/CLDC-4179?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #3245, opened 2026-03-18
CLDC-4179: Fix duplicate logs check

samyou-softwire:
fix for [CLDC-4179](https://mhclgdigital.atlassian.net/browse/CLDC-4179)

original code #3207 

reverted in #3241 

the original code looks fine, noticed that on staging BU dupe checking was turned off entirely which was preventing us from checking this during mobbing

have turned this on everywhere. to be confirmed on staging whether this is of issue

[CLDC-4179]: https://mhclgdigital.atlassian.net/browse/CLDC-4179?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/models/sales_log.rb
+++ b/app/models/sales_log.rb
@@ -18,6 +18,7 @@ class SalesLog < Log
   include Validations::Sales::SoftValidations
   include Validations::SoftValidations
   include MoneyFormattingHelper
+  include CollectionTimeHelper
 
   self.inheritance_column = :_type_disabled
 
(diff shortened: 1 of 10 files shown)


natdeanlewissoftwire (approved):
LGTM

Is the plan to turn the duplicate logs check off again after we've confirmed it works on staging?

I'm not 100% sure of the reason it's disabled on staging atm, but I'm guessing maybe there was a want to be able to bulk upload lots of copies of the same log for load testing or similar. Looks like it's been disabled since sept 2024 so last year's updates don't seem to have included confirming the duplicate bulk upload checks on staging either.

My instinct would be we should leave the toggle as true until we become aware of a definite need for it to be false again to make our environments more aligned. Unless you're aware of that reason already ofc in which case let's revert ASAP!

samyou-softwire:
> My instinct would be we should leave the toggle as true until we become aware of a definite need for it to be false again to make our environments more aligned. Unless you're aware of that reason already ofc in which case let's revert ASAP!

yeah I'm not aware of a reason either. #2635 turned it on last year and #2639 turned it off again soon after but there's no clear comment on why they did this. when we test the duplicate log check we'll find out if there's any reason to turn it off afterwards, though I'm happy to default to not

PR #3253, opened 2026-03-19
CLDC-4312: Fix broken purchase price soft validation

natdeanlewissoftwire:
Closes [CLDC-4312](https://mhclgdigital.atlassian.net/browse/CLDC-4312).

This bug was present because for bedsits, beds is a derived value (1).
But the beds question was not marked as derived in this case, so on submission, `beds` was having its value automatically cleared by `reset_free_user_input_questions_if_not_routed`.

This then meant the soft vaildation question (`value_value_check`) was not routed, so its value was also cleared (as its `depends_on` requires beds to be present).

`set_derived_fields!` then runs after and re-derives beds = 1.

This led to the behaviour where it looked like nothing was happening when confirming the soft validaiton. The good news is the beds value being re-derived means logs won't have lost data if they hit this state. I've confirmed this is the case with the rails console and a test staging log.

The equivalent lettings beds q is already marked as derived so this bug was sales-only.

[CLDC-4312]: https://mhclgdigital.atlassian.net/browse/CLDC-4312?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/models/form/sales/questions/property_number_of_bedrooms.rb
+++ b/app/models/form/sales/questions/property_number_of_bedrooms.rb
@@ -11,5 +11,9 @@ class Form::Sales::Questions::PropertyNumberOfBedrooms < ::Form::Question
     @question_number = get_question_number_from_hash(QUESTION_NUMBER_FROM_YEAR)
   end
 
+  def derived?(log)
+    log.is_bedsit?
+  end
+
   QUESTION_NUMBER_FROM_YEAR = { 2023 => 11, 2024 => 18, 2025 => 17, 2026 => 18 }.freeze
 end
--- a/spec/models/form/sales/questions/property_number_of_bedrooms_spec.rb
+++ b/spec/models/form/sales/questions/property_number_of_bedrooms_spec.rb
@@ -19,8 +19,22 @@ RSpec.describe Form::Sales::Questions::PropertyNumberOfBedrooms, type: :model do
     expect(question.type).to eq("numeric")
   end
 
-  it "is not marked as derived" do
-    expect(question.derived?(nil)).to be false
+  describe "#derived?" do
+    context "when the log is a bedsit" do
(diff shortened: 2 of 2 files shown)


samyou-softwire:
I did notice that if the log is a bedsit it leaves the soft validation page without a reference to number of bedrooms


which does seem a bit confusing. normally it'd show number of bedrooms here. should we make it list the property type as 'bedsit' here?

natdeanlewissoftwire:
> I did notice that if the log is a bedsit it leaves the soft validation page without a reference to number of bedrooms 
> 
> which does seem a bit confusing. normally it'd show number of bedrooms here. should we make it list the property type as 'bedsit' here?

That's a good point @samyou-softwire , I suggest we merge this fix in and then let Rachel/Sammy decide the prioritisation of that if that sounds good? I'm not sure what the standard is here around whether fields that lead to the derivation of soft val fields are included or not in soft validation pages. Can see arguments both ways - it's a bit confusing to show bedsit as what we really care about for the validation is the number of beds. But we shouldn't really show beds as it's not routed/changeable without changing bedsit first. So in summary I think we should not change this here - I'll raise a ticket  as updating the interruption screen question ids is definitely lower prio than getting this bug fix in imo as it's a live bug affecting current collection year logs.

Alternatively if you're aware of a standard for including derivation-related fields then I guess I could just add property type in? Happy either way

samyou-softwire (approved):
there's no fixed standard but it's in essence any question you can change that will hide the soft validation. in this case we want to tell users "your property that has this price, this address and is a bedsit doesn't look right for that price"

a ticket sounds good

natdeanlewissoftwire:
Ticket made: https://mhclgdigital.atlassian.net/browse/CLDC-4336

PR #3257, opened 2026-03-19
CLDC-4318: Set min 1 for proplen

natdeanlewissoftwire:
Closes [CLDC-4318](https://mhclgdigital.atlassian.net/browse/CLDC-4318).

New minimum in action:
image

[CLDC-4318]: https://mhclgdigital.atlassian.net/browse/CLDC-4318?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/models/form/sales/questions/living_before_purchase_years.rb
+++ b/app/models/form/sales/questions/living_before_purchase_years.rb
@@ -4,7 +4,7 @@ class Form::Sales::Questions::LivingBeforePurchaseYears < ::Form::Question
     @id = "proplen"
     @copy_key = "sales.sale_information.living_before_purchase.#{joint_purchase ? 'joint_purchase' : 'not_joint_purchase'}.proplen"
     @type = "numeric"
-    @min = 0
+    @min = form.start_year_2026_or_later? ? 1 : 0
     @max = 80
     @step = 1
     @width = 5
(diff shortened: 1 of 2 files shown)


oscar-richardson-softwire (approved):
LGTM

app/models/form/sales/questions/living_before_purchase_years.rb
```diff
@@ -16,6 +16,10 @@ def suffix_label(log)
     " #{'year'.pluralize(log[id])}"
   end
 
+  def error_suffix_label(_log, value)
+    " #{'year'.pluralize(value)}"
+  end
+
```
samyou-softwire: I think we'd want to implement this differently. seems a bit odd to me that I need to write out this logic twice for different inputs. ideally I'd write that I want "year/years" once.

let's split this out into a separate ticket

samyou-softwire (changes requested):
see comment, I think fixing pluralisation for errors wants more than a quick fix

app/models/form/sales/questions/living_before_purchase_years.rb
natdeanlewissoftwire: Yep this was the cleaner of the two simple options I considered - but fine to separate and have a wider discussion first as to other approaches, this isn't as high prio as the other changes here so don't want to hold up this PR for those changes :)

natdeanlewissoftwire: https://mhclgdigital.atlassian.net/browse/CLDC-4341

samyou-softwire (approved):

PR #3268, opened 2026-03-24
CLDC-4340: Account for nil location name on location select

samyou-softwire:
closes [CLDC-4340](https://mhclgdigital.atlassian.net/browse/CLDC-4340)

there was a bug in the sort function that would always try and sort by the name which can be nil

this is only used visually for ordering the options on the page so doesn't need to be very precise

use postcode as a fallback, which should never be nil

locations select page with a location without a name set

[CLDC-4340]: https://mhclgdigital.atlassian.net/browse/CLDC-4340?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/models/form/lettings/questions/location_id.rb
+++ b/app/models/form/lettings/questions/location_id.rb
@@ -30,16 +30,8 @@ class Form::Lettings::Questions::LocationId < ::Form::Question
 
     scheme_location_ids = lettings_log.scheme.locations.visible.confirmed.pluck(:id)
     answer_options.select { |k, _v| scheme_location_ids.include?(k.to_i) }
-                  .sort_by { |_, v|
-                    if v["hint"].present?
-                      name = v["hint"]
-                      name_text = name&.match(/[a-zA-Z].*/).to_s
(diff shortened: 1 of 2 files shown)


natdeanlewissoftwire (commented):
Has using the postcode as a fallback been agreed w Rachel? I can't see that on the ticket if so - I'm just thinking if users have a name in mind for a location (even if it's not been assigned one in the service) it might be easier to put the nameless locations at the bottom/top of the list instead so they stand out a bit more rather than ordering these by multiple fields. Changes look ok though!

samyou-softwire:
> Has using the postcode as a fallback been agreed w Rachel?

ah good question, hadn't considered other sorts. I'll check in on the ticket. I've pushed a commit for if we want to group all the postcodes at the end

samyou-softwire:
agreed to sort by postcode, good call Nat

PR #3274, opened 2026-03-25
CLDC-4308: Align hb answer option copy

natdeanlewissoftwire:
Closes [CLDC-4308](https://mhclgdigital.atlassian.net/browse/CLDC-4308).

Simple copy update to align with the paper form.
New copy:
image

[CLDC-4308]: https://mhclgdigital.atlassian.net/browse/CLDC-4308?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/models/form/sales/questions/housing_benefits.rb
+++ b/app/models/form/sales/questions/housing_benefits.rb
@@ -12,7 +12,7 @@ class Form::Sales::Questions::HousingBenefits < ::Form::Question
   ANSWER_OPTIONS = {
     "2" => { "value" => "Housing benefit" },
     "3" => { "value" => "Universal Credit housing element" },
-    "1" => { "value" => "Neither housing benefit or Universal Credit housing element" },
+    "1" => { "value" => "Neither" },
     "divider" => { "value" => true },
     "4" => { "value" => "Don’t know " },
   }.freeze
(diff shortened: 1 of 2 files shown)


app/models/form/sales/questions/housing_benefits.rb
```diff
@@ -12,7 +12,7 @@ def initialize(id, hsh, page, joint_purchase:)
   ANSWER_OPTIONS = {
     "2" => { "value" => "Housing benefit" },
     "3" => { "value" => "Universal Credit housing element" },
-    "1" => { "value" => "Neither housing benefit or Universal Credit housing element" },
+    "1" => { "value" => "Neither" },
```
natdeanlewissoftwire, line 15: Have checked the 25 paper form and it also just has "Neither" too, so we should just update this in both years. Simple!

samyou-softwire: I wouldn't be super comfortable with this, changes to live forms are normally done quite carefully as it could, say, impact how users answer the question. I'd want to change 2026 only, especially if we want to keep this as safe as possible

we could also pop a question on the ticket and park it, don't think this'll make it through PO in any case

natdeanlewissoftwire: Last we asked Rachel she explicitly said to change 25 as well in cases like this if it's fixing something that's incorrect (rather than adding/removing fields like our other 26 form changes) so I do think we should do this for both years. It was in one of the refinement calls if you remember? Or did you leave that with a different impression?

samyou-softwire: ah yeah sure let's fix this then

samyou-softwire (approved):

PR #3278, opened 2026-03-26
CLDC-4304: Update person known page copy for 2026

samyou-softwire:
closes [CLDC-4304](https://mhclgdigital.atlassian.net/browse/CLDC-4304)

to account for this question now being total person count

I'm pretty certain what we want now is that the logic is simpler, person 3 page is shown for the 3rd person regardless of if it's a joint purchase or not. joint purchase only impacts seeing the person 2 page or not

person-2-known
person-3-known
person-4-known
person-5-known
person-6-known
</details>

[CLDC-4304]: https://mhclgdigital.atlassian.net/browse/CLDC-4304?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/views/form/headers/_person_2_known_page.erb
+++ b/app/views/form/headers/_person_2_known_page.erb
@@ -1,1 +1,5 @@
-You have given us the details for 0 of the <%= log.hholdcount %> other people in the household
+<% if log.form.start_year_2026_or_later? %>
+  You have given us the details for 1 of the <%= log.hholdcount %> people in the household
+<% else %>
+  You have given us the details for 0 of the <%= log.hholdcount %> other people in the household
+<% end %>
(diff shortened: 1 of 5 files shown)


natdeanlewissoftwire:
This diff looks right to me. A couple of wider Qs @samyou-softwire :

1. t looks like lettings has a different approach and copy to sales but we've made the same changes to how we track household count there for 26/27 I believe - Lettings doesn't  reference total household count at all. Should we be aligning the approach here? My guess is Rachel/Sammy P aren't aware of this divergence
image

2. Even wider but more nitty, I don't think "You've given us details for X of the Y people in the household" is necessarily correct - the user can answer person 2/3 onwards before actually giving previous people's details. To be fair this feels unlikely to happen, so if you're happy with this we can probably leave as is, just raising in case you're aware of a scenario where this could reasonably confuse a user - I can't really think of any :)

samyou-softwire:
> This diff looks right to me. A couple of wider Qs @samyou-softwire :

good questions. I've asked Sammy on the ticket. I think we'll likely end up raising these as separate tasks as to not block this bug fix

natdeanlewissoftwire:
Sounds reasonable to me @samyou-softwire I'll approve these changes on that basis so we can get this merged in as required :)

natdeanlewissoftwire (approved):

PR #3293, opened 2026-03-31
CLDC-4351: only teardown review apps for prs with review apps

natdeanlewissoftwire:
When a review app doesn't exist, we don't want to run the teardown - hence the label check on PR close.

However, if the label was added then removed, we DO want that to invoke a teardown (as it now wouldn't be torn down on PR close).

This label-based approach is in line with the deploy approach, and avoids needing to check whether infra exists or not (and hence avoids potential race conditions that would incur).


.github/workflows/review_teardown_pipeline.yml
```diff
@@ -54,10 +58,33 @@ jobs:
...
+    steps:
+      - name: Comment on PR
+        uses: actions/github-script@v7
+        with:
+          script: |
+            await github.rest.issues.createComment({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              issue_number: context.issue.number,
+              body: 'Review app has been torn down. To redeploy, reapply the `review-app` label.',
```
samyou-softwire, line 89: could we add to the deploy comment that undeploying is possible?

samyou-softwire (approved):
do you have any test PRs where you've ran this pipeline?

natdeanlewissoftwire:
> do you have any test PRs where you've ran this pipeline?

Yep! I've made a few at varying states. I'll make a clean one of the final iteration now though and link here to show the full process.
Edit: see https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/3298
which is based off the latest state of this branch. You can see the review app label spinning up a review app with the new deploy success comment, followed by removal of the label spinning down the review app successfully
image

PR #3297, opened 2026-03-31
CLDC-4298: Ensure value must be a whole number

samyou-softwire:
closes [CLDC-4298](https://mhclgdigital.atlassian.net/browse/CLDC-4298)

the `@step` property for purchase_price was incorrectly set to 0.01, allowing for non whole numbers to be input

this is fixed for 2026

we should also run the attached script post release to round all logs with an incorrect value value to the nearest whole number

image

[CLDC-4298]: https://mhclgdigital.atlassian.net/browse/CLDC-4298?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/tasks/round_value_for_2026_sales_logs.rake
+++ b/lib/tasks/round_value_for_2026_sales_logs.rake
@@ -1,0 +1,12 @@
+desc "Rounds and value for sales logs in the database if they are not a whole number"
+task round_value_for_2026_sales_logs: :environment do
+  ids = SalesLog.filter_by_year(2026).where("value % 1 != 0").pluck(:id)
+  puts "Correcting #{ids.count} sales logs, #{ids}"
+
+  # find all values of mortgage that are not a whole number
+  SalesLog.filter_by_year(2026).where("value % 1 != 0").find_each do |log|
+    log.update(value: log.value.round)
+  end
+
+  puts "Done"
+end
--- a/lib/tasks/round_value_for_2026_sales_logs.rake
+++ b/lib/tasks/round_value_for_2026_sales_logs.rake
@@ -1,10 +1,9 @@
-desc "Rounds and value for sales logs in the database if they are not a whole number"
+desc "Rounds purchase price (the 'value' field) for sales logs in the database if not a whole number"
 task round_value_for_2026_sales_logs: :environment do
-  ids = SalesLog.filter_by_year(2026).where("value % 1 != 0").pluck(:id)
-  puts "Correcting #{ids.count} sales logs, #{ids}"
+  logs = SalesLog.filter_by_year(2026).where("value % 1 != 0")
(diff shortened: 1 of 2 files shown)


lib/tasks/round_value_for_2026_sales_logs.rake
```diff
@@ -0,0 +1,12 @@
+desc "Rounds and value for sales logs in the database if they are not a whole number"
+task round_value_for_2026_sales_logs: :environment do
+  ids = SalesLog.filter_by_year(2026).where("value % 1 != 0").pluck(:id)
```
oscar-richardson-softwire: Minor but could we pull out `SalesLog.filter_by_year(2026).where("value % 1 != 0")` into a variable to save us running this query twice

lib/tasks/round_value_for_2026_sales_logs.rake
```diff
@@ -0,0 +1,12 @@
+desc "Rounds and value for sales logs in the database if they are not a whole number"
```
oscar-richardson-softwire:
```suggestion
desc "Rounds purchase price (the 'value' field) for sales logs in the database if not a whole number"
```

oscar-richardson-softwire: Found it a bit confusing that this field was called 'value' when reading this so might be worth being explicit

lib/tasks/round_value_for_2026_sales_logs.rake
```diff
@@ -0,0 +1,12 @@
...
+task round_value_for_2026_sales_logs: :environment do
+  ids = SalesLog.filter_by_year(2026).where("value % 1 != 0").pluck(:id)
+  puts "Correcting #{ids.count} sales logs, #{ids}"
+
+  # find all values of mortgage that are not a whole number
```
oscar-richardson-softwire:
```suggestion
  # find all values of purchase price that are not a whole number
```

samyou-softwire: this was an LLM hint I forgot to remove, think the description of the rake should explain what's going on here

lib/tasks/round_value_for_2026_sales_logs.rake
samyou-softwire: ah yeah this is better, the value field is definitely confusing

lib/tasks/round_value_for_2026_sales_logs.rake
samyou-softwire: have done so

oscar-richardson-softwire (approved):
LGTM

PR #3300, opened 2026-04-01
CLDC-NONE: Fix tests failing in 2026

samyou-softwire:
this was missed as this file already uses a lot of time travel

hope to remove the time travel from this file when we remove all tests for 2022 logic


oscar-richardson-softwire (approved):
LGTM

PR #3302, opened 2026-04-02
CLDC-3205: Income soft validation not always being shown

samyou-softwire:
closes [CLDC-3205](https://mhclgdigital.atlassian.net/browse/CLDC-3205)

adds the soft validation to the missing spots

this impacts the csv export for a log since it's based on the ordering of questions and the check is moved earlier so these test files needed updating

needs to update the depends_on and have some logic so the check is not shown for age8 if hhmemb is 2

note the logic behind showing the check is not changed so this should not invalidate any existing logs

income soft validation

[CLDC-3205]: https://mhclgdigital.atlassian.net/browse/CLDC-3205?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/models/form/lettings/subsections/household_characteristics.rb
```diff
@@ -50,28 +52,36 @@ def pages
   def person_questions(person_index:)
     [
       Form::Lettings::Pages::PersonKnown.new(nil, nil, self, person_index:),
+
       (Form::Lettings::Pages::PersonAge.new(nil, nil, self, person_index:) if form.start_year_2026_or_later?),
+      (Form::Lettings::Pages::NetIncomeValueCheck.new("person_#{person_index}_age_net_income_value_check", nil, self, person_index:) if form.start_year_2026_or_later?),
```
oscar-richardson-softwire:
```suggestion
      (Form::Lettings::Pages::NetIncomeValueCheck.new("age_#{person_index}_net_income_value_check", nil, self, person_index:) if form.start_year_2026_or_later?),
```

samyou-softwire: ah yep this doesn't match the 2025 usage, have fixed

oscar-richardson-softwire (approved):
LGTM

PR #3305, opened 2026-04-09
CLDC-4332: Check that a support user can only be in certain orgs

samyou-softwire:
closes [CLDC-4332](https://mhclgdigital.atlassian.net/browse/CLDC-4332)

adds a validation to user.rb that ensures support users can only be in org ID 1

left the ID in featuretoggles. it could potentially be in the DB too but I'm happy that this number will change so infrequently we can handle it changing with a release

note that the logic around changing an organisation is complex and makes this work a bit difficult. namely I'm aware of two potential issues:

- user is non support non mhclg. update to support mhclg. this should be valid but you will see an error
- user is non support mhclg. update to support non mhclg. this will be blocked but you will not see an error

fixing these however is quite difficult as changing an organisation is its own special flow where you reassign logs and we must ensure the user is never invalid at any point on that flow. for the primary use case (creating support users in the wrong org) it works fine. I'm cautious against overcomplicating the system and potentially introducing bugs that leaving this reliable check which should always block invalid data

happy to hear thoughts on this before presenting to client

[CLDC-4332]: https://mhclgdigital.atlassian.net/browse/CLDC-4332?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -224,7 +224,7 @@ private
     if user_params.key?(:organisation_id)
       if user_params[:organisation_id].blank?
         @user.errors.add :organisation_id, :blank
-      elsif !@user.role_is_allowed_to_be_in_organisation?(override_organisation_id: user_params[:organisation_id].to_i)
+      elsif !@user.role_is_allowed_to_be_in_organisation?(override_organisation_id: user_params[:organisation_id].to_i) && @user.id != nil
         # this will also be flagged by the validation in user.rb.
         # for convenience we show the error early before they go through the change org flow (involves reassigning logs).
         @user.errors.add :organisation_id, I18n.t("validations.user.support_user_in_wrong_organisation.change_organisation")
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -221,8 +221,14 @@ private
       @user.errors.add :phone
     end
 
-    if user_params.key?(:organisation_id) && user_params[:organisation_id].blank?
-      @user.errors.add :organisation_id, :blank
+    if user_params.key?(:organisation_id)
+      if user_params[:organisation_id].blank?
+        @user.errors.add :organisation_id, :blank
+      elsif !@user.role_is_allowed_to_be_in_organisation?(override_organisation_id: user_params[:organisation_id].to_i)
+        # this will also be flagged by the validation in user.rb.
+        # for convenience we show the error early before they go through the change org flow (involves reassigning logs).
+        @user.errors.add :organisation_id, I18n.t("validations.user.support_user_in_wrong_organisation.change_organisation")
+      end
     end
   end
 
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -224,7 +224,7 @@ private
     if user_params.key?(:organisation_id)
       if user_params[:organisation_id].blank?
         @user.errors.add :organisation_id, :blank
-      elsif !@user.role_is_allowed_to_be_in_organisation?(override_organisation_id: user_params[:organisation_id].to_i) && @user.id != nil
+      elsif !@user.role_is_allowed_to_be_in_organisation?(override_organisation_id: user_params[:organisation_id].to_i) && @user.id.present?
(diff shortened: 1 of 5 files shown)


oscar-richardson-softwire:
> namely I'm aware of two potential issues:
> 
> user is non support non mhclg. update to support mhclg. this should be valid but you will see an error
> user is non support mhclg. update to support non mhclg. this will be blocked but you will not see an error

Please could you explain why these are the case?

app/controllers/users_controller.rb
```diff
@@ -221,8 +221,14 @@ def validate_attributes
...
 
-    if user_params.key?(:organisation_id) && user_params[:organisation_id].blank?
-      @user.errors.add :organisation_id, :blank
+    if user_params.key?(:organisation_id)
+      if user_params[:organisation_id].blank?
+        @user.errors.add :organisation_id, :blank
+      elsif !@user.role_is_allowed_to_be_in_organisation?(override_organisation_id: user_params[:organisation_id].to_i)
+        # this will also be flagged by the validation in user.rb.
+        # for convenience we show the error early before they go through the change org flow (involves reassigning logs).
+        @user.errors.add :organisation_id, I18n.t("validations.user.support_user_in_wrong_organisation.change_organisation")
```
oscar-richardson-softwire, line 230: Do we not want to make this error text dynamic somehow? If we create a new user and we've tried to create a support user in a non-MHCLG org, won't this `validate_attributes` function run? In that case, it wouldn't make much sense to show 'You cannot move a user with a support account to a non-MHCLG organisation.'

app/models/user.rb
```diff
@@ -407,6 +414,16 @@ def organisation_not_merged
     end
   end
 
+  def support_user_is_in_correct_organisation
+    return if role_is_allowed_to_be_in_organisation?
+
+    if role_changed?
+      errors.add :organisation_id, I18n.t("validations.user.support_user_in_wrong_organisation.change_role")
```
oscar-richardson-softwire: Is this the right field to add this error to?

app/controllers/users_controller.rb
samyou-softwire: have changed this to only run on edit specifically. making a new will run the code in `user.rb`

app/models/user.rb
samyou-softwire: suppose it can go either way, just that we show an error. I agree though looking back now it'll be likely the user swapped this field so will add it to the role field

samyou-softwire:
> Please could you explain why these are the case?

sure. the notable part is that editing an org has its own flow. namely
- make edits on a user (in this example to both role and org)
- submit
- role edit is saved immediately
- user is sent to an org move confirmation page to deal with logs that may move
- confirm
- organisation edit is saved

> user is non support non mhclg. update to support mhclg. this should be valid but you will see an error

- make edits on a user
- submit
- role edit is saved immediately
- user is now a non mhclg support
- this is invalid
- abort

> user is non support mhclg. update to support non mhclg. this will be blocked but you will not see an error

- make edits on a user
- submit
- role edit is saved immediately
- user is now a support mhclg
- user is sent to an org move confirmation page to deal with logs that may move
- confirm
- organisation edit is saved
- user is now a support non mhclg
- this is invalid
- the user saves but the route after confirming an organisation never expects any errors
- so the save will be blocked but no error will be shown

I think it is possible to fix these but neither feels very intuitive too and I'm wary of overcomplicating this to fix a couple niche cases may not be worthwhile

oscar-richardson-softwire (approved):
LGTM

samyou-softwire:
# readme

so we can test on the review app I've added a temporary feature toggle to turn it on for review apps

after getting this approved on review apps, push a commit to turn it on for staging with organisation name "MHCLG 2" so it can be tested there, if testing on staging is desired

if not, revert the review app line and merge. it should work when deployed to production

the single model test fail will be fixed when this commit is reverted

samyou-softwire:
PO done on review app, no staging testing needed. will queue for merge, expecting pipeline to be fixed

PR #3307, opened 2026-04-10
CLDC-4300: Update sale date staircasing date validation

samyou-softwire:
closes [CLDC-4300](https://mhclgdigital.atlassian.net/browse/CLDC-4300)

updates this validation to what the ticket requests for 2026 logs

adds a rake task to set existing invalid logs back to in progress

q103 q104 validations

[CLDC-4300]: https://mhclgdigital.atlassian.net/browse/CLDC-4300?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


samyou-softwire:
there is more work to be done here in regards to Q103 last staircasing date

config/locales/validations/sales/sale_information.en.yml
```diff
@@ -25,11 +25,11 @@ en:
         initialpurchase:
           must_be_after_1980: "The initial purchase date must be after January 1, 1980."
           must_be_before_last_transaction: "The initial purchase date must be before the last staircasing transaction date."
-          must_be_before_saledate: "The initial purchase date must be before the date of this sale."
+          must_be_before_saledate: "The initial staircasing transaction date must be before the sale date of this staircasing transaction."
         lasttransaction:
           must_be_after_1980: "The last staircasing transaction date must be after January 1, 1980."
           must_be_after_initial_purchase: "The last staircasing transaction date must be after the initial purchase date."
-          must_be_before_saledate: "The last staircasing transaction date must be before the date of this sale."
+          must_be_before_saledate: "The initial staircasing transaction date must be before the sale date of this staircasing transaction."
```
oscar-richardson-softwire: Have we okayed this wording change with CORE?

lib/tasks/fix_sales_logs_with_invalid_initialpurchase_lasttransaction.rake
```diff
@@ -0,0 +1,14 @@
+desc "We tightened the validation between initial purchase date and sale date so the two can no longer be equal. To avoid invalid logs we clear initialpurchase if it equals saledate"
+task fix_sales_logs_with_invalid_initialpurchase_lasttransaction: :environment do
+  initial_purchase_equal_saledate_logs = SalesLog.filter_by_year_or_later(2025).where("initialpurchase = saledate")
+  initial_purchase_equal_lasttransaction_logs = SalesLog.filter_by_year_or_later(2025).where("initialpurchase = lasttransaction")
```
oscar-richardson-softwire: Shouldn't we be identifying logs where `"initialpurchase = lasttransaction"` and logs where `"lasttransaction = saledate"`? Think that covers everything, as when `"initialpurchase = saledate"` we must have both of the other conditions. At the moment I think we're missing logs where only `"lasttransaction = saledate"`.

lib/tasks/fix_sales_logs_with_invalid_initialpurchase_lasttransaction.rake
```diff
@@ -0,0 +1,14 @@
+desc "We tightened the validation between initial purchase date and sale date so the two can no longer be equal. To avoid invalid logs we clear initialpurchase if it equals saledate"
```
oscar-richardson-softwire: Think this description isn't quite complete as it should reference the updated relationship between all three fields.

samyou-softwire: have updated this to cover the other case

lib/tasks/fix_sales_logs_with_invalid_initialpurchase_lasttransaction.rake
samyou-softwire: > Shouldn't we be identifying logs where "initialpurchase = lasttransaction"

this is in the code

> and logs where "lasttransaction = saledate"

I don't think this is necessary. if lasttransaction =/= saledate then either lasttransaction =/= initialpurchase or saledate =/= initialpurchase, since initialpurchase can't be two values at the same time. or at least I think this logic holds

config/locales/validations/sales/sale_information.en.yml
samyou-softwire: this was originally in the ticket description but I think as a a mistake as it was assumed this validation didn't already exist. should be fine to leave as-is as CORE okayed the new description without this change, so I've reverted

samyou-softwire (commented):
I checked ticket comments and realised this actually should only apply to 26 and later. I've updated the code to account for this

lib/tasks/fix_sales_logs_with_invalid_initialpurchase_lasttransaction.rake
oscar-richardson-softwire: Will make change as discussed on call

samyou-softwire: done, see comment on commit

lib/tasks/fix_sales_logs_with_invalid_initialpurchase_lasttransaction.rake
```diff
@@ -0,0 +1,13 @@
+desc "We tightened the validation between initial purchase date in 2026, last transaction date and sale date so the two can no longer be equal. To avoid invalid logs we clear initialpurchase if it equals saledate and if initialpurchase = lasttransaction we clear both"
```
oscar-richardson-softwire:
```suggestion
desc "We tightened the validation in 2026 between initial purchase date, last transaction date and sale date so that no two can be equal and initial purchase date < last transaction date < sale date. To avoid invalid logs we clear lasttransaction if it equals saledate and if initialpurchase = lasttransaction we clear both"
```

oscar-richardson-softwire (approved):
LGTM

oscar-richardson-softwire (approved):
LGTM

PR #3308, opened 2026-04-13
CLDC-4373: Fix pregnancy check appearing twice

samyou-softwire:
closes [CLDC-4373](https://mhclgdigital.atlassian.net/browse/CLDC-4373)

the `NoHouseholdMemberLikelyToBePregnantCheck` after the lead tenant age was missing a person_index set, this would cause it to appear outside of any person box at the top of the section

old household characteristics with double pregnancy check

setting the person index places this second `NoHouseholdMemberLikelyToBePregnantCheck` correctly after lead tenant age

new household characteristics with a pregnancy check after age and only one after household count

[CLDC-4373]: https://mhclgdigital.atlassian.net/browse/CLDC-4373?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/models/form/lettings/pages/no_household_member_likely_to_be_pregnant_check.rb
+++ b/app/models/form/lettings/pages/no_household_member_likely_to_be_pregnant_check.rb
@@ -2,7 +2,8 @@ class Form::Lettings::Pages::NoHouseholdMemberLikelyToBePregnantCheck < ::Form::
   def initialize(id, hsh, subsection, person_index: 0)
     super(id, hsh, subsection)
     @copy_key = "lettings.soft_validations.pregnancy_value_check.no_household_member_likely_to_be_pregnant_check"
-    @depends_on = [{ "no_household_member_likely_to_be_pregnant?" => true }]
+    @person_index = person_index
+    @depends_on = depends_on
     @title_text = {
       "translation" => "forms.#{form.start_date.year}.#{@copy_key}.title_text",
       "arguments" => [],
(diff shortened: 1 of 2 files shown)


natdeanlewissoftwire (approved):
Change looks good to me!

It looks like this bug happened because we're defaulting person_index to 0 in the question for this value check (which looks to be needed so that it shows in the right answers card in the household needs subsection). I wonder if it's worth a tech debt ticket to separate this check answers card behaviour out from the person index behaviour? Then we can remove the default from this and similar validations, and missing it would throw an error rather than just producing the wrong behaviour silently as we saw here.
Doing a search for `@check_answers_card_number = person_index` shows quite a few cases where we're conflating these two concepts atm so there may be other cases where this is a bit confusing.

samyou-softwire:
> Change looks good to me!
> 
> It looks like this bug happened because we're defaulting person_index to 0 in the question for this value check (which looks to be needed so that it shows in the right answers card in the household needs subsection). I wonder if it's worth a tech debt ticket to separate this check answers card behaviour out from the person index behaviour? Then we can remove the default from this and similar validations, and missing it would throw an error rather than just producing the wrong behaviour silently as we saw here. Doing a search for `@check_answers_card_number = person_index` shows quite a few cases where we're conflating these two concepts atm so there may be other cases where this is a bit confusing.

sure, I think `check_answers_card_title` we can be using in pretty much all cases. [CLDC-4428](https://mhclgdigital.atlassian.net/browse/CLDC-4428)

[CLDC-4428]: https://mhclgdigital.atlassian.net/browse/CLDC-4428?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

samyou-softwire:
fixed issue around this validation also showing in every user regardless of whether their details are given 
household characteristics page with 2 members and soft validation only shown for these 2

natdeanlewissoftwire (approved):

PR #3310, opened 2026-04-14
CLDC-4400: Buyer still serving answer remapping

natdeanlewissoftwire:
Closes [CLDC-4400](https://mhclgdigital.atlassian.net/browse/CLDC-4400).

Aligns 2025 answer options for `hhregresstill` with the new 2026 answer options and updates spec. (the new option 10 is hidden in the UI as it's just used for the remapping of old values.

Adds a rake task for remapping values entered in 2025 logs using the UI. This rake task needs to be run after release with the release datetime passed as an arg, so only 2025 logs created BEFORE the release goes live are updated. We could just use a date and check for edge cases (and probably there would be 0) but the datetime just avoids this extra check entirely so I've gone for that.

2025 question with updated options:
image
2025 question with value of 10 showing correctly as "No" in the CYA page (Q62):
image

[CLDC-4400]: https://mhclgdigital.atlassian.net/browse/CLDC-4400?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lib/tasks/remap_2025_hhregresstill_values.rake
```diff
@@ -0,0 +1,17 @@
+desc "Remaps hhregresstill values for manually created 2025/26 sales logs"
+task :remap_2025_hhregresstill_values, %i[before_date] => :environment do |_task, args|
+  usage_message = "Usage: rake remap_2025_hhregresstill_values['before_datetime']. before_datetime must be in format YYYY-MM-DDTHH:MM:SS"
+  raise usage_message if args[:before_date].blank?
+
+  before_datetime = Time.zone.parse(args[:before_date])
+  raise usage_message if before_datetime.nil?
```
samyou-softwire: I'm getting some mixed messages here between `before_date` and `before_datetime`

lib/tasks/remap_2025_hhregresstill_values.rake
```diff
@@ -0,0 +1,17 @@
+desc "Remaps hhregresstill values for manually created 2025/26 sales logs"
+task :remap_2025_hhregresstill_values, %i[before_date] => :environment do |_task, args|
+  usage_message = "Usage: rake remap_2025_hhregresstill_values['before_datetime']. before_datetime must be in format YYYY-MM-DDTHH:MM:SS"
```
samyou-softwire, line 3: what would the release plan for this look like? can we get a datetime for when a release happened?

could we get this noted on the ticket?

samyou-softwire, line 3: do we know how this would interact with daylight savings?

lib/tasks/remap_2025_hhregresstill_values.rake
```diff
@@ -0,0 +1,17 @@
...
+  raise usage_message if args[:before_date].blank?
+
+  before_datetime = Time.zone.parse(args[:before_date])
+  raise usage_message if before_datetime.nil?
+
+  logs = SalesLog.filter_by_year(2025).where(bulk_upload_id: nil).where(hhregresstill: [5, 6, 7]).where("created_at < ?", before_datetime)
+  puts "Updating #{logs.count} sales logs"
```
samyou-softwire, line 10: do we have an idea how many logs this will touch? I've been adding a `#{logs.map(&:id)}` to the print statement so we have a record of what we changed

lib/tasks/remap_2025_hhregresstill_values.rake
```diff
@@ -0,0 +1,17 @@
...
+
+  logs.where(hhregresstill: 5).update_all(hhregresstill: 10)
```
samyou-softwire: I've been generally favouring using `update!` for this kind of work. it's slower (since it'll run validations) but it means we can be confident we didn't create invalid logs, since `update_all` skips validations

samyou-softwire (changes requested):
couple qs on how we'd use this rake task, main site changes look good

lib/tasks/remap_2025_hhregresstill_values.rake
natdeanlewissoftwire: Ah thanks, datetime is correct - hadn't pushed the full update

lib/tasks/remap_2025_hhregresstill_values.rake
natdeanlewissoftwire: Yep, currently it should affect 60 logs.

lib/tasks/remap_2025_hhregresstill_values.rake
natdeanlewissoftwire: Great suggestion, will do

lib/tasks/remap_2025_hhregresstill_values.rake
natdeanlewissoftwire: Yep, `Time.zone.parse` uses the current time zone (London). So we'll pass in the time of release completion in UK time, which is what we'll see in github. `created_at` is in UTC but ActiveRecord handles the conversion automatically based on the time zone.

natdeanlewissoftwire: Sure, see the ticket :)

natdeanlewissoftwire: FWIW the staging pipeline has an extra performance step so the time will be a minute off (which should be fine) - but for prod we should be (closer to) bang on. I'll add a note to check for any discrepancies if there are any as there will still be a small window of possible erroneously included logs.

samyou-softwire (approved):

PR #3312, opened 2026-04-15
CLDC-4429 and CLDC-4242: dependency updates

natdeanlewissoftwire:
Closes [CLDC-4249](https://mhclgdigital.atlassian.net/browse/CLDC-4249) AND [CLDC-4242](https://mhclgdigital.atlassian.net/browse/CLDC-4242).

As part of the investigations into the changes for CLDC-4249, I tried upgrading our alpine image (including upgrading ruby to 3.4.9 and node to 24). While not strictly necessary for these changes (alpine 21 has actually receive these security patches despite being in EOL), it made sense to combine all the changes in this PR for efficiency (and confidence that the fixes made will still apply after these upgrades are made too.

Vulnerabilities in current image used on main:
image

All vulnerabilities are cleared as a result of these changes:
image

[CLDC-4249]: https://mhclgdigital.atlassian.net/browse/CLDC-4249?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[CLDC-4242]: https://mhclgdigital.atlassian.net/browse/CLDC-4242?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Dockerfile
```diff
@@ -1,7 +1,10 @@
-FROM ruby:3.4.4-alpine3.20 as base
+FROM ruby:3.4.9-alpine3.23 as base
 
 WORKDIR /app
 
+# Upgrade base packages to pick up latest security patches
+RUN apk upgrade --no-cache
```
natdeanlewissoftwire, line 6: This is the line that actually addresses the vulnerabilities - AWS scanner was triggering because it thinks the base images reference vulnerable unpatched versions of `musl libc` etc., even though they're implicitly upgraded anyway as a result of our `RUN apk add ...` command on line 16. So we were never really vulnerable in the final image state as far as I could see - but doing this upgrade at the top of the file is required for AWS scanner to be happy that we're really using the latest version, which I suppose is fair enough otherwise we might be doing something unsafe before making the upgrade.

Dockerfile
```diff
@@ -10,7 +13,7 @@ RUN apk add --update --no-cache tzdata && \
 # build-base: compilation tools for bundle
 # yarn: node package manager
 # postgresql-dev: postgres driver and libraries
-RUN apk add --no-cache build-base=0.5-r3 busybox=1.36.1-r29 nodejs=20.15.1-r0 yarn=1.22.22-r0 bash=5.2.26-r0 libpq-dev yaml-dev linux-headers
+RUN apk add --no-cache build-base busybox nodejs yarn bash libpq-dev yaml-dev linux-headers
```
natdeanlewissoftwire, line 16: These pins aren't needed, we're already pinning the base version of the image. Hence I haven't added new pins since changing the base image.

.github/workflows/run_tests.yml
```diff
@@ -38,7 +38,6 @@ jobs:
 
     env:
       RAILS_ENV: test
-      GEMFILE_RUBY_VERSION: 3.1.1
```
natdeanlewissoftwire, line 41: Not used anywhere, and was out of date anyway.

Dockerfile
```diff
@@ -40,14 +43,14 @@ RUN bundle config set without ""
...
-RUN apk add firefox
+RUN apk add firefox=145.0-r0
 RUN wget https://github.com/mozilla/geckodriver/releases/download/v0.31.0/geckodriver-v0.31.0-linux64.tar.gz \
     && tar -xvzf geckodriver-v0.31.0-linux64.tar.gz \
     && rm geckodriver-v0.31.0-linux64.tar.gz \
     && chmod +x geckodriver \
     && mv geckodriver /usr/local/bin/
 
-CMD bundle exec rake parallel:setup && bundle exec rake parallel:spec
+CMD ["sh", "-c", "bundle exec rake parallel:setup && bundle exec rake parallel:spec"]
```
natdeanlewissoftwire, line 53: wiz prefers this

Dockerfile
```diff
@@ -40,14 +43,14 @@ RUN bundle config set without ""
 RUN bundle install --jobs=4 --no-binstubs --no-cache
 
 # Install gecko driver for Capybara tests
-RUN apk add firefox
+RUN apk add firefox=145.0-r0
```
natdeanlewissoftwire, line 46: pinning here satisfies wiz, I can see both sides myself but fine with this

Dockerfile
samyou-softwire: I think these are useful, it means we can be confident exactly what versions we use and that nothing will silently break

I don't think pinning the docker tag guarantees they won't push a new tag (especially since our tag doesn't specify a minor version)

samyou-softwire (changes requested):
~could you deploy a review app?~ never mind I see there is one

Dockerfile
natdeanlewissoftwire: Yes agreed, let me restate - it feels like a lot of maintenance to me to have all these individual versions pinned. We won't receive patches or updates to these versions if we pin them (which means we may experience exactly the issue that caused us to have to ensure `musl-libc` was upgraded again for these packages!). I can go through and find the latest versions of all these pinned libraries (though we're only pinning half of them as it stands, so maybe I should do all of them in that case?). On balance I don't think it's worth it personally - but if we do I should probably pin to the latest versions of all these relevant libraries for the new alpine version. Thoughts?

samyou-softwire: thinking about it again, the scary situation I want to avoid is our app suddenly breaking when a dependent library updates without our knowledge. though, pinning doesn't really solve this either, since the base image could update and drop the pinned version from its package list. a crystal ball will tell us which is more likely to happen here...

happy to leave them unpinned then for simplicity

natdeanlewissoftwire: Yeah, I think this has always been a risk to a greater or lesser extent to be fair - share your concerns but agree pinning here doesn't massively reduce it (and has its own cons) so will leave as is :)

samyou-softwire (approved):

PR #3313, opened 2026-04-15
CLDC-4331: Make downloads traceable

samyou-softwire:
closes [CLDC-4331](https://mhclgdigital.atlassian.net/browse/CLDC-4331)

adds a new DB table to track any time a download is created for auditing purposes

for sales, lettings, users, schemes, and their respective pages in the per-organisation view

database view

[CLDC-4331]: https://mhclgdigital.atlassian.net/browse/CLDC-4331?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


db/migrate/20260415101455_create_download_record.rb
```diff
@@ -0,0 +1,13 @@
+class CreateDownloadRecord < ActiveRecord::Migration[7.2]
+  def change
+    create_table :download_records do |t|
+      t.integer :download_type, null: false
+      t.string :download_filters, null: false
```
natdeanlewissoftwire, line 5: Just to check - can this definitely never be null? Do all flows default to send all filters when there are none selected?

app/models/download_record.rb
```diff
@@ -0,0 +1,26 @@
+# Not used functionally, but used to allow auditing of what users downloaded what info
```
natdeanlewissoftwire: I think "not used functionally" is a bit confusing personally - can you remove that bit? I think the second half of this line makes it clear what you mean here

app/models/download_record.rb
```diff
@@ -0,0 +1,26 @@
...
+class DownloadRecord < ApplicationRecord
+  belongs_to :user
+  belongs_to :user_organisation, class_name: "Organisation"
+
+  DOWNLOAD_TYPE = {
+    user: 0,
+    lettings_log: 1,
+    sales_log: 2,
+    organisation: 3,
+    scheme: 4,
```
natdeanlewissoftwire: I'm assuming this also includes location downloads - do we want to distinguish between these?

app/models/download_record.rb
```diff
@@ -0,0 +1,26 @@
+# Not used functionally, but used to allow auditing of what users downloaded what info
+# Caches some info about the user at the time of download
+class DownloadRecord < ApplicationRecord
+  belongs_to :user
+  belongs_to :user_organisation, class_name: "Organisation"
+
+  DOWNLOAD_TYPE = {
```
natdeanlewissoftwire, line 7: Small one, is the order of these just pulled from the ticket? Might be more helpful to order the same as the tabs in the service imo

app/models/download_record.rb
```diff
@@ -0,0 +1,26 @@
+# Not used functionally, but used to allow auditing of what users downloaded what info
+# Caches some info about the user at the time of download
+class DownloadRecord < ApplicationRecord
```
natdeanlewissoftwire, line 3: Can this link to a specific download - i.e. what's the process for identifying which record in s3 a given DownloadRecord corresponds to? Do you think it might be helpful if we can include the filename at least in this record, or is the plan for this to be constructed by a dev at time of audit from the download type + timestamp (which won't be quite correct but I suppose will be close enough)?

I know for users at least we're not saving files in s3 - though I think there's a ticket to change that so will be relevant for those downloads soon as well

app/models/download_record.rb
```diff
@@ -0,0 +1,26 @@
...
+  belongs_to :user
+  belongs_to :user_organisation, class_name: "Organisation"
+
+  DOWNLOAD_TYPE = {
+    user: 0,
+    lettings_log: 1,
+    sales_log: 2,
+    organisation: 3,
```
natdeanlewissoftwire: This download type isn't used anywhere - and I'm not sure we have the option for downloading orgs anyway, I think we can just remove it unless we're planning to add this functionality in the near future

db/migrate/20260415101455_create_download_record.rb
samyou-softwire: we call `to_s` before saving the filters in the db so I think that should put us in the clear. in any case it looks like it defaults to `{}`, see method `deserialize_filters_from_session`

app/models/download_record.rb
samyou-softwire: sounds good, have reworded

app/models/download_record.rb
samyou-softwire: I don't think there's a specific location download; when you download schemes it bundles the locations in the same download

app/models/download_record.rb
samyou-softwire: funnily enough minus the unused organisation this already is in order
CORE header

app/models/download_record.rb
samyou-softwire: yeah for simplicity I think it's fine for devs to match up downloads from the logs. there shouldn't be too many

app/models/download_record.rb
samyou-softwire: ah good spot, will remove

app/models/download_record.rb
natdeanlewissoftwire: Just confirmed in staging: there's an option to do just schemes, just locations or schemes AND locations, and there's a `download_type` query param in the url for each case, so imo we may as well retain this distinction unless it's a nightmare to do?
image

app/models/download_record.rb
natdeanlewissoftwire: Ah very good I was looking at the non-support tabs, yeah agree this is fine; support is more the perspective of anyone looking at this table :)

app/models/download_record.rb
natdeanlewissoftwire: (if it is a nightmare - at the very least I think we should rename the download type `scheme` to something like `scheme_or_location`)

samyou-softwire: ah apologies, thought I read only one option. I've stored this in the download filters

natdeanlewissoftwire: Ah okay yeah that's fair, happy with that for audit purposes - could you rename  `DOWNLOAD_TYPE` `scheme` to `scheme_or_location` though as it could be a locations download with no schemes? Or are you thinking of Download type as representing the page the download came from? if so we should probably rename that instead imo 😅 

Separately the scheme `DOWNLOAD_TYPE` should have value 3 now we've removed orgs

samyou-softwire: I've renamed it to `scheme_location`, since it could contain schemes, locations, or both. have fixed the number

natdeanlewissoftwire (approved):
LGTM!

PR #3317, opened 2026-04-17
CLDC-4389: BU validation prioritisation

natdeanlewissoftwire:
Closes [CLDC-4389](https://mhclgdigital.atlassian.net/browse/CLDC-4389).

Ensures all soft validations are added to the error report only after all hard validations have been added, and do not override existing hard validations that have already been added.

e.g. field_128 now only shows in "critical errors", not "confirmation needed" when it is is affected by hard and soft validations, previously was only showing in "confirmation needed":

image
image

[CLDC-4389]: https://mhclgdigital.atlassian.net/browse/CLDC-4389?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/services/bulk_upload/sales/year2026/row_parser.rb
+++ b/app/services/bulk_upload/sales/year2026/row_parser.rb
@@ -492,7 +492,6 @@ class BulkUpload::Sales::Year2026::RowParser
   validate :validate_assigned_to_when_support, on: :after_log
   validate :validate_managing_org_related, on: :after_log
   validate :validate_relevant_collection_window, on: :after_log
-  validate :validate_incomplete_soft_validations, on: :after_log
 
   validate :validate_uprn_exists_if_any_key_address_fields_are_blank, on: :after_log
   validate :validate_address_fields, on: :after_log
@@ -559,6 +558,8 @@ class BulkUpload::Sales::Year2026::RowParser
         end
       end
     end
+
+    validate_incomplete_soft_validations
 
     add_errors_for_invalid_fields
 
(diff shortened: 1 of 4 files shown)


app/services/bulk_upload/sales/year2026/row_parser.rb
```diff
@@ -1654,7 +1655,7 @@ def validate_incomplete_soft_validations
         next if log.form.questions.none? { |q| q.id == interruption_screen_question_id && q.page.routed_to?(log, nil) }
 
         field_mapping_for_errors[interruption_screen_question_id.to_sym]&.each do |field|
-          if errors.none? { |e| e.options[:category] == :soft_validation && field_mapping_for_errors[interruption_screen_question_id.to_sym].include?(e.attribute) }
+          if errors.none? { |e| field_mapping_for_errors[interruption_screen_question_id.to_sym].include?(e.attribute) }
```
samyou-softwire, line 1658: for the avoidance of doubt, we're happy that there's no reason for sales to work different?

samyou-softwire (approved):

app/services/bulk_upload/sales/year2026/row_parser.rb
natdeanlewissoftwire: Hard to be 100% certain, but this was removed from lettings when the 2023 parser was made, and it wasn't removed from sales then, which I think was just an oversight. I've had a dig through git history and tickets and found no justification anywhere there either so I think we're okay to do this.

PR #3319, opened 2026-04-20
CLDC-4328: Update wrong field numbers error

samyou-softwire:
close [CLDC-4328](https://mhclgdigital.atlassian.net/browse/CLDC-4328)

CORE already has the functionality to add a description for critical failures. we will need to update notify to start showing them again. see `BulkUploadMailer#send_bulk_upload_failed_service_error_mail`

updates wording as requested

[CLDC-4328]: https://mhclgdigital.atlassian.net/browse/CLDC-4328?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/config/locales/validations/lettings/2025/bulk_upload.en.yml
+++ b/config/locales/validations/lettings/2025/bulk_upload.en.yml
@@ -12,7 +12,7 @@ en:
           wrong_template:
             wrong_template: "Incorrect start dates, please ensure you have used the correct template."
             no_headers: "Your file does not contain the required header rows. Add or check the header rows and upload your file again. [Read more about using the template headers](%{guidance_link})."
-            wrong_field_numbers_count: "Incorrect number of fields, please ensure you have used the correct template."
+            wrong_field_numbers_count: "Incorrect number of fields, please ensure you have used the template for the correct type (lettings or sales) and year."
             over_max_column_count: "Too many columns, please ensure you have used the correct template."
           owning_organisation:
             not_found: "The owning organisation code is incorrect."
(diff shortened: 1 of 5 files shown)


config/locales/validations/lettings/2025/bulk_upload.en.yml
```diff
@@ -12,7 +12,7 @@ en:
           wrong_template:
             wrong_template: "Incorrect start dates, please ensure you have used the correct template."
             no_headers: "Your file does not contain the required header rows. Add or check the header rows and upload your file again. [Read more about using the template headers](%{guidance_link})."
-            wrong_field_numbers_count: "Incorrect number of fields, please ensure you have used the correct template."
+            wrong_field_numbers_count: "Incorrect number of fields, please ensure you have used the template for the correct type (lettings or sales) and year."
```
natdeanlewissoftwire, line 15: FWIW I'd have thought this should say "...correct log type...", do you agree/think it's worth checking? Happy to just take what's in the ticket as gospel but thought I'd flag in case you think this reads a bit odd too

natdeanlewissoftwire (approved):
Code changes look good - re: notify, can this error trigger if there aren't any explicit errors? i.e. in a true service error/unexpected problem case, what happens in the template? We should test that the email still sends even if there's not an explicit error passed (though if it doesn't that's a pre-existing issue, not something introduced here).

config/locales/validations/lettings/2025/bulk_upload.en.yml
samyou-softwire: I'll double check, this was specific wording in a comment from CORE though I'll mention it as it would be very easier (almost harder not to)

samyou-softwire:
> Code changes look good - re: notify, can this error trigger if there aren't any explicit errors? i.e. in a true service error/unexpected problem case, what happens in the template? We should test that the email still sends even if there's not an explicit error passed (though if it doesn't that's a pre-existing issue, not something introduced here).

notify is quite useful and will omit that section
processing error email without errors list

config/locales/validations/lettings/2025/bulk_upload.en.yml
samyou-softwire: rachel agreed to use specific messages, have changed this now

PR #3321, opened 2026-04-20
CLDC-4330: Stop login count resetting on deactivation

samyou-softwire:
closes [CLDC-4330](https://mhclgdigital.atlassian.net/browse/CLDC-4330)

rather than piggybacking on existing devise properties, use our own flag

this will handle users reset under the old system and the new system

[CLDC-4330]: https://mhclgdigital.atlassian.net/browse/CLDC-4330?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/controllers/auth/confirmations_controller.rb
+++ b/app/controllers/auth/confirmations_controller.rb
@@ -5,8 +5,8 @@ class Auth::ConfirmationsController < Devise::ConfirmationsController
     yield resource if block_given?
 
     if resource.errors.empty?
-      # previously we reset sign in count to indicate that a user was deactivated and so needs to reset their password on confirming their email post reactivation.
-      # now we have a specific flag for this.
+      # previously we reset sign_in_count to indicate that a user was deactivated and so needs to reset their password on confirming their email post reactivation.
+      # now we have a specific flag for this as resetting sign in count was difficult for auditing.
       # though for backwards compatability we need to ensure previous users with a reset sign in count still will see the password reset screen
       if resource.reset_password_on_confirmation || resource.sign_in_count.zero?
         token = resource.send(:set_reset_password_token)
--- a/app/controllers/auth/confirmations_controller.rb
+++ b/app/controllers/auth/confirmations_controller.rb
@@ -9,7 +9,7 @@ class Auth::ConfirmationsController < Devise::ConfirmationsController
       # this would force a password reset both if it was your very first log in, and on your first login after reactivation.
       # now we have a specific flag for the latter case as resetting sign_in_count was difficult for auditing.
       # note that some deactivated users will have a sign_in_count of 0 and not have this flag set if they were deactivated before we made this change.
-      if resource.reset_password_on_confirmation || resource.sign_in_count.zero?
+      if resource.force_reset_password_on_confirmation || resource.sign_in_count.zero?
         token = resource.send(:set_reset_password_token)
         redirect_to "#{edit_user_password_url}?reset_password_token=#{token}&confirmation=true"
       else
--- a/app/controllers/auth/confirmations_controller.rb
+++ b/app/controllers/auth/confirmations_controller.rb
@@ -5,7 +5,10 @@ class Auth::ConfirmationsController < Devise::ConfirmationsController
     yield resource if block_given?
 
     if resource.errors.empty?
-      if resource.sign_in_count.zero?
+      # previously we reset sign in count to indicate that a user was deactivated and so needs to reset their password on confirming their email post reactivation.
+      # now we have a specific flag for this.
+      # though for backwards compatability we need to ensure previous users with a reset sign in count still will see the password reset screen
+      if resource.reset_password_on_confirmation || resource.sign_in_count.zero?
         token = resource.send(:set_reset_password_token)
         redirect_to "#{edit_user_password_url}?reset_password_token=#{token}&confirmation=true"
       else
--- a/app/controllers/auth/confirmations_controller.rb
+++ b/app/controllers/auth/confirmations_controller.rb
@@ -5,9 +5,10 @@ class Auth::ConfirmationsController < Devise::ConfirmationsController
     yield resource if block_given?
 
     if resource.errors.empty?
-      # previously we reset sign_in_count to indicate that a user was deactivated and so needs to reset their password on confirming their email post reactivation.
-      # now we have a specific flag for this as resetting sign in count was difficult for auditing.
-      # though for backwards compatability we need to ensure previous users with a reset sign in count still will see the password reset screen
+      # previously we reset sign_in_count on deactivation and had only the .zero? check here.
+      # this would force a password reset both if it was your very first log in, and on your first login after reactivation.
+      # now we have a specific flag for the latter case as resetting sign_in_count was difficult for auditing.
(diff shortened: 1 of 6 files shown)


app/controllers/auth/confirmations_controller.rb
```diff
@@ -5,7 +5,10 @@ def show
     yield resource if block_given?
 
     if resource.errors.empty?
-      if resource.sign_in_count.zero?
+      # previously we reset sign in count to indicate that a user was deactivated and so needs to reset their password on confirming their email post reactivation.
```
natdeanlewissoftwire: Can we say `sign_in_count` not `sign in count` for clarity?

app/models/user.rb
```diff
@@ -179,7 +179,7 @@ def deactivate!(reactivate_with_organisation: false)
     update!(
       active: false,
       confirmed_at: nil,
-      sign_in_count: 0,
+      reset_password_on_confirmation: true,
```
natdeanlewissoftwire, line 182: I find naming this by its usage a bit confusing, what we mean here is that 
if `reset_password_on_confirmation` is true, it's the first time a user has confirmed their account adn they've never been deactivated before, and if it's false then they're reconfirming after being reactivated right?

I wonder if it would be clearer to instead track an attribute like `deactivated_at` or something to check whether a user has been deactivated before or not, what do you think?

natdeanlewissoftwire (commented):
I notice we're using `sign_in_count` in `resend_invitation_emails.rake`, I think that needs updating too in case it's used in the future.#

Also, do we need to include this new user field in user downloads?

app/models/user.rb
samyou-softwire: this is not how I've been tracking it. when reactivated a user should be prompted to change their password after signing in

there are other reasons to receive this confirmation email; if you change your email address you need to but you will not need to change your password

so, if:
- you are a new user (sign in count is 0)
- you are a returning user after being deactivated (`reset_password_on_confirmation` is set)

you must reset your password. unless I've been misinterpreting this flow?

samyou-softwire:
> I notice we're using `sign_in_count` in `resend_invitation_emails.rake`, I think that needs updating too in case it's used in the future.#

I don't mind, it's not clear what exactly this rake is for, though the original ticket [CLDC-2484](https://mhclgdigital.atlassian.net/browse/CLDC-2484) makes it sound like theyre interested in users who haven't logged in to CORE at all, which I think probably doesn't include reactivated users

it'll be up to whoever runs the script to decide, I'd rather leave it as the less surprising behaviour that this sends to users who have never signed in at all

> Also, do we need to include this new user field in user downloads?

could do, though it's not a request from CORE and it doesn't feel super useful

[CLDC-2484]: https://mhclgdigital.atlassian.net/browse/CLDC-2484?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

app/models/user.rb
samyou-softwire: we could name it after the specific purpose here `user_has_not_logged_in_after_reactivation`, though limits us if we want some other flow to force a password reset when you confirm your email

app/controllers/auth/confirmations_controller.rb
samyou-softwire: good spot, have update this comment line and the line below to make this a little clearer

app/controllers/auth/confirmations_controller.rb
```diff
@@ -5,7 +5,10 @@ def show
     yield resource if block_given?
 
     if resource.errors.empty?
-      if resource.sign_in_count.zero?
+      # previously we reset sign_in_count to indicate that a user was deactivated and so needs to reset their password on confirming their email post reactivation.
+      # now we have a specific flag for this as resetting sign in count was difficult for auditing.
```
natdeanlewissoftwire: couple more `sign_in_count`s to add here and below

app/models/user.rb
natdeanlewissoftwire: Ah okay, yep that makes more sense - I'd personally find `force_password_reset_on_confirmation`, `trigger_password_reset_on_confirmation` etc.  a bit clearer (not really sure why) but I'm happy that the current name does describe what's going on so will leave with you

app/controllers/auth/passwords_controller.rb
```diff
@@ -37,6 +37,7 @@ def update
 
     if resource.errors.empty?
       resource.unlock_access! if resource.respond_to?(:unlock_access!)
+      resource.reset_password_on_confirmation = false if resource.reset_password_on_confirmation
```
natdeanlewissoftwire: Can this line just be `resource.reset_password_on_confirmation = false`? Personally prefer it but also nitty so don't massively mind which you prefer :)

app/controllers/auth/confirmations_controller.rb
```diff
@@ -5,7 +5,10 @@ def show
...
+      # though for backwards compatability we need to ensure previous users with a reset sign in count still will see the password reset screen
+      if resource.reset_password_on_confirmation || resource.sign_in_count.zero?
```
natdeanlewissoftwire: Having looked at this again, couldn't we just avoid this weirdness by setting all users with `sign_in_count` of 0 to have `reset_password_on_confirmation` true and not need to bother including `sign_in_count` in this condition (or having the big descriptive comment explaining the weird behaviour)?

samyou-softwire: I think the sign in count check will always need to be here; when you sign in for the first time ever you will need to set a password

samyou-softwire: I've taken another pass at this wording, let me know if this makes sense now

app/models/user.rb
samyou-softwire: like that, have gone with `force_reset_password_on_confirmation`

app/controllers/auth/confirmations_controller.rb
samyou-softwire: should be fixed

app/controllers/auth/passwords_controller.rb
samyou-softwire: can do, I added it to make it clear that if this is false we do nothing (even if this is naturally the case), though it is redundant

natdeanlewissoftwire: yeah agree it's a bit more explicit this way if at the cost of verbosity, fine to leave :)

app/controllers/auth/confirmations_controller.rb
natdeanlewissoftwire: Hm, I still think we could probably do this more cleanly but it would require some data updates (and admittedly expanding the default behaviour of sign_in_count that devise would expect) - so given those drawbacks I think it's alright that we keep this explicit, the comment's clearer now too so anyone seeing this should hopefully take caution which will be a good thing rather than hiding the complexity :)

natdeanlewissoftwire (approved):
LGTM!

PR #3322, opened 2026-04-20
CLDC-4432: Skip invalid field errors if the question is not routed to

samyou-softwire:
closes [CLDC-4332](https://mhclgdigital.atlassian.net/browse/CLDC-4332)

adds an additional check when adding invalid field errors that at least one question related to that field is routed to

since the BU success depends on the log being marked as complete & not error presence, this should not be a functional change. it should only mean it such that fewer errors are shown on the error report

[CLDC-4332]: https://mhclgdigital.atlassian.net/browse/CLDC-4332?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/spec/services/bulk_upload/sales/year2025/row_parser_spec.rb
+++ b/spec/services/bulk_upload/sales/year2025/row_parser_spec.rb
@@ -339,6 +339,7 @@ RSpec.describe BulkUpload::Sales::Year2025::RowParser do
 
         describe "invalid fields" do
           context "when a field has been marked as invalid" do
+            # field_34 nationality is only shown if field_10 staircasing is no
             let(:attributes) { setup_section_params.merge({ field_10: 2, field_31: 0 }) }
 
             before do
--- a/spec/services/bulk_upload/sales/year2025/row_parser_spec.rb
+++ b/spec/services/bulk_upload/sales/year2025/row_parser_spec.rb
@@ -338,9 +338,9 @@ RSpec.describe BulkUpload::Sales::Year2025::RowParser do
         end
 
         describe "invalid fields" do
-          let(:attributes) { setup_section_params.merge({ field_31: 0 }) }
-
           context "when a field has been marked as invalid" do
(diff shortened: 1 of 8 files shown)


spec/services/bulk_upload/sales/year2026/row_parser_spec.rb
```diff
@@ -344,9 +344,9 @@ def purch_id_field
         end
 
         describe "invalid fields" do
-          let(:attributes) { setup_section_params.merge({ field_34: 0 }) }
-
           context "when a field has been marked as invalid" do
+            let(:attributes) { setup_section_params.merge({ field_10: "2", field_34: 0 }) }
```
samyou-softwire: turns out that this test was always slightly wrong, by default field_10 is 1 (staircasing) which hides the nationality question

oscar-richardson-softwire: Cool, I think maybe a short comment explaining this (and the fact that setting field_10 as 1 hides the nationality question in the test case below) might be helpful

spec/services/bulk_upload/sales/year2025/row_parser_spec.rb
```diff
@@ -338,9 +338,9 @@ def purch_id_field
         end
 
         describe "invalid fields" do
-          let(:attributes) { setup_section_params.merge({ field_31: 0 }) }
-
           context "when a field has been marked as invalid" do
+            let(:attributes) { setup_section_params.merge({ field_10: 2, field_31: 0 }) }
```
oscar-richardson-softwire, line 343: Same as my comments on `spec/services/bulk_upload/sales/year2026/row_parser_spec.rb` below but for this file

oscar-richardson-softwire (approved):
LGTM

PR #3329, opened 2026-04-22
CLDC-3676: Remove all pre 2025 validations

samyou-softwire:
first audit done by claude opus 4.6

removes all hard validations that run in 2023 and before

'for all 2024 and later' logic assumed as true

removes 3 soft validations that only showed in 2023 and earlier. left 2024 soft validations in since theyre technically 'questions' and we don't want to remove 2024 questions quite yet


spec/models/form/lettings/pages/net_income_value_check_spec.rb
samyou-softwire, line 1: claude fixed this file originally. I'm not sure what prompted it to but the fixes all look good so included it

natdeanlewissoftwire (approved):

PR #3332, opened 2026-04-23
CLDC-4269: Raise max hhmemb

natdeanlewissoftwire:
Closes [CLDC-4269](https://mhclgdigital.atlassian.net/browse/CLDC-4269).

Increases the max of `hhmemb` in lettings logs to 15. Equivalent sales question `hholdcount` already has max 15.

Also updates any behaviour that assumed `hhmemb` would cap out at 8. I'm pretty sure I've found everything but whoever reviews this, please do check my thinking/work there!

New max of 15 is a valid value:
image

Validation now allows values 1-15:
image

Still routes correct other person questions:
image
And only routes 8 max:
image

[CLDC-4269]: https://mhclgdigital.atlassian.net/browse/CLDC-4269?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/models/derived_variables/lettings_log_variables.rb
```diff
@@ -339,7 +339,7 @@ def infer_at_most_one_relationship!
   def infer_only_partner!(partner_number)
     return unless hhmemb
 
-    (2..hhmemb).each do |i|
+    (2..[hhmemb, 8].min).each do |i|
```
samyou-softwire: considering how many times we use this same segment, could it become a lettings log method? perhaps for the more general case `[hhmemb || 8, 8].min`

app/models/form/lettings/questions/hhmemb.rb
```diff
@@ -5,7 +5,7 @@ def initialize(id, hsh, page)
     @type = "numeric"
     @width = 2
     @check_answers_card_number = 0
-    @max = 8
+    @max = 15
```
samyou-softwire, line 8: for all years?

app/models/validations/soft_validations.rb
samyou-softwire: I think this method will need an edit, this was originally implemented to avoid an error checking ecstat9 if you entered 9 for household members, since it'll run all validations on submit (even if its invalid). as it stands setting hhmemb to 9 would disable the working situation validation

fwiw the rest of the similar calls here I think are fine, they're in methods that are 'all_xyz'. we can't assume all household members are xyz if we don't have all the info recorded. this one though is an `any_xyz` method so we can do something here

samyou-softwire (changes requested):

app/models/form/lettings/questions/hhmemb.rb
natdeanlewissoftwire: Yep, all years in general - and furthermore the support ticket is specifically about a 25/26 log so it's more than just a nice to have

app/models/derived_variables/lettings_log_variables.rb
natdeanlewissoftwire: Hm, good idea - I'll have a count and look at doing this for sales too as this copies that pattern.

app/models/validations/soft_validations.rb
natdeanlewissoftwire: Ah yes good catch, I've intentionally left the rest as you say due to the "all_xyz" thinking. Missed this one, will update :)

app/models/sales_log.rb
```diff
@@ -104,6 +104,7 @@ class SalesLog < Log
 
   OPTIONAL_FIELDS = %w[purchid othtype buyers_organisations].freeze
   DUPLICATE_LOG_ATTRIBUTES = %w[owning_organisation_id purchid saledate age1_known age1 sex1 sexrab1 ecstat1 postcode_full uprn address_line1].freeze
+  MAX_PEOPLE_WITH_DETAILS = 6
```
samyou-softwire: does this do anything? I'm not super opinionated on changing how sales works right now but if it's unused a comment mentioning this may be of use to not confuse people

app/models/form/lettings/pages/person_known.rb
```diff
@@ -2,11 +2,18 @@ class Form::Lettings::Pages::PersonKnown < ::Form::Page
   def initialize(id, hsh, subsection, person_index:)
     super(id, hsh, subsection)
     @id = "person_#{person_index}_known"
-    @depends_on = (person_index..8).map { |index| { "hhmemb" => index } }
+    @depends_on = depends_on
     @person_index = person_index
```
samyou-softwire: these two lines make me suspicious. as I understand it the `depends_on` method will run immediately, but `@person_index` is not set until a line later. the tests seem to suggest it works but the local debugger suggests it may not
debugger screenshot showing operand is nil

fix would be to swap these two lines over

samyou-softwire (changes requested):
requesting mostly for the latter comment

app/models/sales_log.rb
natdeanlewissoftwire: Yep it's referenced by `max_people_with_details` in `log.rb`

app/models/form/lettings/pages/person_known.rb
natdeanlewissoftwire: Yes - if this was the other way around it didn't work for this reason. For safety I'm happy to just put it after person_index but I've confirmed this works regardless :)

natdeanlewissoftwire: Not sure why the debugger disagrees though, and that concerns me too! Best be safe, have pushed that update

app/models/sales_log.rb
samyou-softwire: as in, sales validations don't use `max_people_with_details`

natdeanlewissoftwire: ah gotcha, sure okay. the equivalent lettings isn't used everywhere either, feels out of scope so I'll put comments

samyou-softwire (approved):

PR #3333, opened 2026-04-23
CLDC-3315: Add dynamic error fields for type

samyou-softwire:
closes [CLDC-3315](https://mhclgdigital.atlassian.net/browse/CLDC-3315)

most of the field fixups were made in #2313, only type needed doing. based on #2311 originally

[CLDC-3315]: https://mhclgdigital.atlassian.net/browse/CLDC-3315?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/services/bulk_upload/sales/year2025/row_parser.rb
```diff
@@ -1191,6 +1191,13 @@ def mortgageused
     field_109 if staircasing?
   end
 
+  def type_fields
```
natdeanlewissoftwire: nit - but there's also a related method `sale_type` earlier in the file - I think we should use either `type` or `sale_type` for both, fine with renaming that to just `type` rather than changing this one :) (for both 25 an 26 row parsers)

natdeanlewissoftwire (approved):
Could you add https://github.com/communitiesuk/submit-social-housing-lettings-and-sales-data/pull/2313 to the description for posterity too please :)

app/services/bulk_upload/sales/year2025/row_parser.rb
samyou-softwire: good spot, I like `sale_type_fields` more so I'll normalise that way

PR #3338, opened 2026-04-27
CLDC-4365: Reduce size of soft validation success box

samyou-softwire:
closes [CLDC-4365](https://mhclgdigital.atlassian.net/browse/CLDC-4365)

adds an option to make the notification be two thirds in desired cases

don't want this to be universal as not all pages are two thirds

pregnancy soft validation with two thirds success message

I'm happy enough leaving this as the only fix for now. there are other two third pages we could touch but none with such an obvious two thirds box. don't think it'd be worth the time to research further

[CLDC-4365]: https://mhclgdigital.atlassian.net/browse/CLDC-4365?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -38,7 +38,7 @@ module ApplicationHelper
       success: true,
       title_heading_level: 3,
       title_id: "flash-notice",
-      role: "alert"
+      role: "alert",
     ) do |notification_banner|
       notification_banner.with_heading(text: flash.notice.html_safe)
       if flash[:notification_banner_body]
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -32,6 +32,20 @@ module ApplicationHelper
     !request.path.match?(/\/notifications\/\d+$/) && (authenticated_user_has_notifications? || unauthenticated_user_has_notifications?)
   end
 
+  def notification_banner
+    govuk_notification_banner(
+      title_text: "Success",
+      success: true, title_heading_level: 3,
+      title_id: "flash-notice",
+      role: "alert"
+    ) do |notification_banner|
+      notification_banner.with_heading(text: flash.notice.html_safe)
+      if flash[:notification_banner_body]
+        tag.p flash[:notification_banner_body]&.html_safe
+      end
+    end
+  end
+
 private
 
   def paginated_title(title, pagy)
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -35,7 +35,8 @@ module ApplicationHelper
   def notification_banner
     govuk_notification_banner(
       title_text: "Success",
-      success: true, title_heading_level: 3,
+      success: true,
+      title_heading_level: 3,
       title_id: "flash-notice",
       role: "alert"
     ) do |notification_banner|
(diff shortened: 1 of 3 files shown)


app/helpers/application_helper.rb
```diff
@@ -32,6 +32,20 @@ def notifications_to_display?
     !request.path.match?(/\/notifications\/\d+$/) && (authenticated_user_has_notifications? || unauthenticated_user_has_notifications?)
   end
 
+  def notification_banner
+    govuk_notification_banner(
+      title_text: "Success",
+      success: true, title_heading_level: 3,
```
natdeanlewissoftwire: can we put these on two lines?

app/views/layouts/application.html.erb
```diff
@@ -120,17 +120,15 @@
...
-          end %>
+          <% if flash[:notification_banner_two_thirds] %>
+            <div class="govuk-grid-row">
+              <div class="govuk-grid-column-two-thirds-from-desktop">
+                <%= notification_banner %>
+              </div>
+            </div>
+          <% else %>
+            <%= notification_banner %>
+          <% end %>
```
natdeanlewissoftwire: Maybe e.g. this would be tidier (and more explicit that it's a full row otherwise)?
```suggestion
          <div class="govuk-grid-row">
            <div class="<%= flash[:notification_banner_two_thirds] ? 'govuk-grid-column-two-thirds-from-desktop' : 'govuk-grid-column-full' %>">
              <%= notification_banner %>
            </div>
          </div>
```

natdeanlewissoftwire (commented):
LGTM (with some small suggestions)! Small thing, can you fix the ticket link in the description please it links to [CLDC-4395](https://mhclgdigital.atlassian.net/browse/CLDC-4395) instead

Just re other pages that might exhibit the same behaviour, I'm happy for this not to be included in the scope of this ticket, I think they're much rarer and a separate refactor, but are you aware of anywhere else we might want to consider doing something similar? I think at least the first of these would be worth a tech debt ticket, possibly in combination with any other areas you've come across as part of this work :) But approving this now as far as the scope of this ticket goes

Image

Image

samyou-softwire:
force push to correct numbers on commits

app/helpers/application_helper.rb
samyou-softwire: have done so

app/views/layouts/application.html.erb
samyou-softwire: good idea, checked locally and this works too

samyou-softwire:
> Just re other pages that might exhibit the same behaviour, I'm happy for this not to be included in the scope of this ticket, I think they're much rarer and a separate refactor, but are you aware of anywhere else we might want to consider doing something similar? I think at least the first of these would be worth a tech debt ticket, possibly in combination with any other areas you've come across as part of this work :) But approving this now as far as the scope of this ticket goes

I've raised [CLDC-4452](https://mhclgdigital.atlassian.net/browse/CLDC-4452) to cover that first example and one other place I noticed

[CLDC-4452]: https://mhclgdigital.atlassian.net/browse/CLDC-4452?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

natdeanlewissoftwire (approved):

PR #3341, opened 2026-04-30
CLDC-4300: Update sale date staircasing date validation

samyou-softwire:
closes [CLDC-4300](https://mhclgdigital.atlassian.net/browse/CLDC-4300)

fixed version of #3307, reverts #3339 with an extra commit

fixes the rake script to work in cases where all 3 dates are equal. previously it'd try to run the first update!, fail as the other two dates were still equal and not complete the rake

now it'll work successfully as it runs the update to reset both dates first, which will always succeed

[CLDC-4300]: https://mhclgdigital.atlassian.net/browse/CLDC-4300?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


oscar-richardson-softwire (approved):
LGTM

PR #3349, opened 2026-05-29
CLDC-4470: Update vulnerable packages

natdeanlewissoftwire:
In updating some package vulnerabilities, some restyling of the navbar was required that we should run by product owners. 
Basically:
Vulnerable packages needed updating -> the secure version is only compatible with govuk 6.x components -> govuk 6.x has breaking changes e.g. in navbar styling (and a few other minor places)

Minor changes:
- some colour label tweaks
- referencing helpers requires `helpers.` prefix in `/components` `.rb` files
- font size 14 is not allowed anymore, 16 is the new small.

Screenshots of new navbar styling on various pages:

Screenshot 2026-06-01 at 15 37 20
Screenshot 2026-06-01 at 15 38 12
Screenshot 2026-06-01 at 15 37 44
Screenshot 2026-06-01 at 15 37 40
Screenshot 2026-06-01 at 15 37 24
Screenshot 2026-06-01 at 15 40 03


app/components/bulk_upload_error_row_component.rb
```diff
@@ -2,9 +2,8 @@ class BulkUploadErrorRowComponent < ViewComponent::Base
   attr_reader :bulk_upload_errors
 
   def initialize(bulk_upload_errors:)
+    super()
     @bulk_upload_errors = bulk_upload_errors
-
-    super
```
oscar-richardson-softwire, line 7: Why this change?

app/views/layouts/rails_admin/_navigation.html.erb
```diff
@@ -1,9 +1,13 @@
 <%= govuk_header(
-      classes: "app-header app-header--orange",
+      classes: "app-header",
       homepage_url: Rails.application.routes.url_helpers.root_path,
-      navigation_classes: "govuk-header__navigation--end",
```
oscar-richardson-softwire, line 4: Was `govuk-header__navigation--end` a non-existent class? I can't see it being removed anywhere in this PR but I also can't see it in the codebase.

spec/helpers/application_helper_spec.rb
```diff
@@ -8,18 +8,42 @@
   let(:pagy) { nil }
   let(:current_user) { FactoryBot.create(:user) }
 
-  describe "govuk_header_classes" do
-    context "with external user" do
-      it "shows the standard app header" do
-        expect(govuk_header_classes(current_user)).to eq("app-header")
+  describe "govuk_service_navigation_classes" do
```
oscar-richardson-softwire, line 11: This is a nitpick but I feel the context blocks in these tests could perhaps be structured more cleanly.

- "with non-support user"
  - "when no notifications are displayed"
  - "when notifications are displayed"

- "with support user"
  - "when no notifications are displayed"
  - "when notifications are displayed"

These are the same cases as you already had but just spells it out a bit more clearly.

webpack.config.js
```diff
@@ -40,8 +40,8 @@ module.exports = {
   },
   resolve: {
     alias: {
-      'govuk-frontend-styles': path.resolve(__dirname, 'node_modules/govuk-frontend/dist/govuk/all.scss'),
-      'govuk-prototype-styles': path.resolve(__dirname, 'node_modules/@x-govuk/govuk-prototype-components/x-govuk/all.scss'),
+      'govuk-frontend-styles': path.resolve(__dirname, 'node_modules/govuk-frontend/dist/govuk/index.scss'),
+      'govuk-prototype-styles': path.resolve(__dirname, 'node_modules/@x-govuk/govuk-prototype-components/dist/govuk-prototype-components.scss'),
```
oscar-richardson-softwire, line 44: Why these changes?

spec/requests/users_controller_spec.rb
```diff
@@ -76,13 +76,28 @@
     end
 
     describe "title link" do
-      it "routes user to the home page" do
+      it "has GOV.UK header and service navigation both linking to home page for non-support user" do
```
oscar-richardson-softwire: Again nitpick but feel this would be cleaner with "for a non-support user" and "for a support user" context blocks.

app/components/bulk_upload_error_row_component.rb
natdeanlewissoftwire: The new version of ViewComponent::Base has no initializer so we can't just reference `super` anymore. Moving it to the top is a Ruby convention, doesn't functionally make a difference though. It didn't do anything before, and it won't do anything now, it's just Ruby convention to call super when overriding initialize

spec/helpers/application_helper_spec.rb
natdeanlewissoftwire: Sure thing, have done!

spec/requests/users_controller_spec.rb
natdeanlewissoftwire: Yep fair, this is definitely cleaner split up a bit more

webpack.config.js
natdeanlewissoftwire: These files have moved in the latest version's node_modules files, so we need to update our aliases to keep the styling working correctly

app/views/layouts/rails_admin/_navigation.html.erb
natdeanlewissoftwire: Nope, it was just in a package, `node_modules/govuk-frontend`

oscar-richardson-softwire: Do we still want to apply this styling in the new navigation component?

natdeanlewissoftwire: Nope, there's not really an equivalent anymore - we're using the service navigation bar rather than a header for our service name and sign in link. Rachel's approved the designs as is so I think we're good to go here :)

oscar-richardson-softwire (approved):
LGTM

PR #3351, opened 2026-06-04
CLDC-4473: Add pipeline steps for SBOM tracking and upload

natdeanlewissoftwire:
Closes [CLDC-4473](https://mhclgdigital.atlassian.net/browse/CLDC-4473).

Adds a CI pipeline step for SBOM tracking and upload to https://deps.softwire.com

Summary from our claude skill that generated this:
```
Changes

  New file — .github/workflows/upload-sbom.yml (reusable workflow)
  - Generates a CycloneDX SBOM with Syft v1.45.0 (install script pinned to commit 9673f867…), which auto-detects both Gemfile.lock (Ruby) and yarn.lock (Node) in one scan.
  - Uploads to api-deps.softwire.com with projectname: CORE, parentname: Support, autocreate: true, and the projectversion passed in by the caller.
  - Syft binary is cached by version.
  
  staging_pipeline.yml — added sbom job: needs: [aws_deploy], projectversion: staging.

  production_pipeline.yml — added sbom job: needs: [aws_deploy], projectversion: prod.

  Each runs only after its deploy succeeds, so SBOMs aren't uploaded on PR branches or failed deploys.                                                                                       
```

[CLDC-4473]: https://mhclgdigital.atlassian.net/browse/CLDC-4473?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.claude/settings.json
+++ b/.claude/settings.json
@@ -1,0 +1,5 @@
+{
+  "enabledPlugins": {
+    "dependency-track@skillwire": true
+  }
+}
--- a/.github/workflows/production_pipeline.yml
+++ b/.github/workflows/production_pipeline.yml
@@ -20,3 +20,12 @@ jobs:
       release_tag: ${{ needs.test.outputs.releasetag }}
     permissions:
       id-token: write
+
+  sbom:
+    name: Upload SBOM
+    needs: [aws_deploy]
+    uses: ./.github/workflows/upload-sbom.yml
+    with:
+      projectversion: prod
+    secrets:
+      DTRACK_API_KEY: ${{ secrets.DTRACK_API_KEY }}
--- a/.github/workflows/production_pipeline.yml
+++ b/.github/workflows/production_pipeline.yml
@@ -24,6 +24,8 @@ jobs:
   sbom:
     name: Upload SBOM
     needs: [aws_deploy]
+    permissions:
+      contents: read
     uses: ./.github/workflows/upload-sbom.yml
     with:
       projectversion: prod
--- a/.github/workflows/staging_pipeline.yml
+++ b/.github/workflows/staging_pipeline.yml
@@ -32,6 +32,15 @@ jobs:
     permissions:
       id-token: write
 
+  sbom:
+    name: Upload SBOM
+    needs: [aws_deploy]
+    uses: ./.github/workflows/upload-sbom.yml
+    with:
+      projectversion: staging
+    secrets:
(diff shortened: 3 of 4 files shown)


.claude/settings.json
```diff
@@ -0,0 +1,5 @@
+{
+  "enabledPlugins": {
+    "dependency-track@skillwire": true
```
natdeanlewissoftwire, line 3: enables this for all users of the repo. We shouldn't need to use it again (I think!) but good to have this consistent if we do

natdeanlewissoftwire:
Have added the required DTRACK_API_KEY to staging and prod actions environments

.github/workflows/staging_pipeline.yml
```diff
@@ -32,6 +32,15 @@ jobs:
...
+  sbom:
+    name: Upload SBOM
+    needs: [aws_deploy]
+    uses: ./.github/workflows/upload-sbom.yml
+    with:
+      projectversion: staging
+    secrets:
+      DTRACK_API_KEY: ${{ secrets.DTRACK_API_KEY }}
+
   performance:
```
natdeanlewissoftwire, line 44: This is just about our github token on our staging env. And there's already a permissinos block for the performance job anyway so have marked as a false positive

.github/workflows/production_pipeline.yml
```diff
@@ -20,3 +20,14 @@ jobs:
...
       id-token: write
+
+  sbom:
+    name: Upload SBOM
+    needs: [aws_deploy]
+    permissions:
+      contents: read
+    uses: ./.github/workflows/upload-sbom.yml
+    with:
+      projectversion: prod
```
oscar-richardson-softwire, line 31: Just wondering if `environment` would be a better name for this input? `projectversion` makes me think it's the release version number or something

natdeanlewissoftwire: Yeah I did think this was a little odd too - but it's called this to be in line with https://github.com/marketplace/actions/upload-bom-to-dependency-track which is what we're using under the hood. Their `projectversion` is 'master' in their examples, so I think using it as the env label is correct (we upload prod and staging SBOMs separately given they run in different pipelines, which will allow us to track when prod is out of sync w staging due to not being released to).

oscar-richardson-softwire (approved):
LGTM

PR #3360, opened 2026-07-13
CLDC-4497: Allow users with no role to download data

natdeanlewissoftwire:
When creating our download_records table there was an assumption that role would be non-null. This is not true - multiple users are just DPOs, and have no role otherwise.

This PR allows these users to download records without causing an ActiveRecord null violation


oscar-richardson-softwire (approved):
LGTM

